PR #1, opened 2019-07-13
feat(readme) init

nothingismagick:
Just the facts. And a picture. With a few links. And thankyous. And a license or three.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,78 @@
+# quasar-proton [WIP]
+## A fresh take on creating cross-platform apps.
+[![official icon](https://img.shields.io/badge/Quasar%201.0-Official-blue.svg)](https://quasar.dev) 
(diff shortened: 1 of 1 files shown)


Momics:
Investigating the Tauri v2 source code and examples for handling re-spinning killed processes that listen on the network for iOS and Android. Specifically looking into issues with TCP connections becoming unreachable after the app goes into the background and is reopened.

Momics:
I'm investigating the Tauri v2 source code and examples for handling re-spinning killed processes that listen on the network, particularly for iOS and Android. I have encountered an issue where a TCP connection becomes unreachable after the app goes into the background, the screen locks, and the app is reopened. Any insights or relevant code snippets would be greatly appreciated!

j4rviscmd:
dummy

j4rviscmd:
Please delete this comment

PR #20, opened 2019-08-18
feat(tauri) initial work for the CLI

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No

- [ ] It's submitted to the `dev` branch and _not_ the `master` branch
- [ ] Any necessary documentation has been added or updated [in the docs](https://github.com/quasarframework/tauri/tree/dev/docs) (for faster update click on "Suggest an edit on GitHub" at bottom of page) or explained in the PR's description.

**Other information:**


tauri/helpers/generator.js
```diff
@@ -0,0 +1,13 @@
+const compileTemplate = require('lodash.template'),
+  fs = require('fs'),
+  appPaths = require('./app-paths'),
+  path = require('path')
+
+module.exports = cfg => {
+  const apiTemplate = fs.readFileSync(path.resolve(__dirname, '../../lib/tauri.js'), 'utf-8')
```
nothingismagick: I don't like these relative paths...

tauri/helpers/tauri-config.js
```diff
@@ -0,0 +1,29 @@
...
+      },
+      bundle: {
+        active: true
+      },
+      whitelist: {},
+      window: {
+        title: require(appPaths.resolve.app('package.json')).productName
+      },
+      security: {
+        csp: 'default-src data: filesystem: ws: http: https: \'unsafe-eval\' \'unsafe-inline\''
```
nothingismagick: nice

tauri/helpers/generator.js
lucasfernog: me neither but what can we do?

nothingismagick: Use apppaths / parse from its expected entry point in node_modules?

nothingismagick: But its not a blocker

tauri/runner.js
```diff
@@ -0,0 +1,189 @@
+const
+  chokidar = require('chokidar'),
+  debounce = require('lodash.debounce')
+
+const
+  { spawn } = require('./helpers/spawn'),
+  log = require('./helpers/logger')('app:tauri')
+  onShutdown = require('./helpers/on-shutdown'),
+  fse = require('fs-extra')
```
nothingismagick: i prefer to treeshake fs-extra and import only the stuff needed:

```
const { readFileSync, writeFileSync, ensureFileSync } = require('fs-extra')
```

tauri/runner.js
```diff
@@ -0,0 +1,189 @@
...
+  __manipulateToml(callback) {
+    const toml = require('@iarna/toml'),
+      tomlPath = this.appPaths.resolve.tauri('Cargo.toml'),
+      tomlFile = fse.readFileSync(tomlPath),
+      tomlContents = toml.parse(tomlFile)
+
+    callback(tomlContents)
+
+    const output = toml.stringify(tomlContents)
+    fse.writeFileSync(tomlPath, output)
```
nothingismagick: its best to ensureFileSync first

tauri/helpers/generator.js
lucasfernog: appPaths are for devland paths so won't work here.. I don't see how the options are better here

nothingismagick:
oops - forgot where i was for a second, you're right

>

tauri/runner.js
lucasfernog: here the path will always exist (Cargo.toml) because I read it before, so the error would be thrown earlier.

tauri/injector.js
```diff
@@ -1,5 +1,5 @@
-const fs = require('fs'),
-  fse = require('fs-extra'),
+const { mkdirSync } = require('fs'),
+  { copySync, renameSync, existsSync } = require('fs-extra'),
```
nothingismagick: mkdirSync is available from fs-extra, no need to import fs too

PR #25, opened 2019-08-21
Merge for 1.0.0-alpha.0

nothingismagick:
We need to merge dev to master in order to serve the docs properly - and everything else for that matter.

Looks like it needs to be done manually.


nothingismagick:
Looks like a manual merge is going to be necessary.

nothingismagick:
This MR is too stale. Will recreate as soon as the CLI work is ready.

PR #39, opened 2019-09-25
Refactor(Tauri) Rust 2018

tensor-programming:
- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

Refactor the entire Tauri codebase to 2018 edition standard.


tensor-programming:
The refactors in the templates folder needs to be done with care.  We should also be stripping out any of the quasar language and verbiage.  Currently, the cargo.toml references @quasar as a path for both tauri and tauri-ui; this should be resolved.

lucasfernog:
> The refactors in the templates folder needs to be done with care. We should also be stripping out any of the quasar language and verbiage. Currently, the cargo.toml references @quasar as a path for both tauri and tauri-ui; this should be resolved.

That's something being solved by the CLI PR.

nothingismagick:
@tensor-programming - please consider adding PGP signing to your github account. At the moment I know about each and every pull request, so I can personally verify that you've done these - but we require signed commits (or that an admin signs off on them). Since you're an admin, it'd be good to also lead by example. :)

PR #52, opened 2019-11-16
feat/web-view

nothingismagick:
This PR introduces testing, linting and we now are consuming upstream web-view. Using Boscop's crate is a temporary thing until we get it working. Then we will probably publish our own crate and track his in order to keep things safe.

Requesting immediate merge so that everyone can profit from testing.

### DIFF
--- a/.env.jest
+++ b/.env.jest
@@ -1,0 +1,1 @@
+NODE_ENV=test
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+/src-tauri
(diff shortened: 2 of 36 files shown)


nothingismagick:
So we are definitely not going to be passing these tests for a while. We will need to look into WHY they are failing though. Eg in the test library build-lib (ubuntu-latest) it seems:

```
Package webkit2gtk-4.0 was not found in the pkg-config search path.\nPerhaps you should add the directory containing `webkit2gtk-4.0.pc\'\nto the PKG_CONFIG_PATH environment variable
```

Got some insight here @jbolda ?

jbolda:
Looks like Mac doesn't come bundled with Rust actually, so I can PR that. Not sure what the ubuntu issue is offhand though.

jbolda:
MacOS addressed in #53.

PR #120, opened 2019-12-01
[WIP] example / gatsby / themed-site

jbolda:

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: beginning of smoke tests

- [ ] Yes. Issue #___
- [x] No

**Other information:**
This is a themed site and a cheap way to get a minimal maintenance ~50 page web site as a test.


nothingismagick:
I think the pipelines are failing because you need to install tauri-cli:
```
 app:spawn Running "cargo tauri-cli --features embedded-server --release" +0ms

error: no such subcommand: `tauri-cli`

 app:spawn Command "cargo" failed with exit code: 101 +1ms
```

The commands are: 

```
# from cargo:
$ cargo install tauri-cli
```

```
# from source
$ cargo install --path ../../../cli/tauri-cli --force
```

jbolda:
@nothingismagick Is the plan to require installing both rust and js cli or is that just the current requirement?

RE: content of example; I don't think the recipe portion of the example has any real identifying information and I believe that my most recent push should make it optional. I can make the content on the homepage generic though, certainly. I have a bunch of upstream work to do on it, and then I can come back and fix anything remaining. Would you mind we put a rain check on making it generic?

nothingismagick:
> installing both rust and js cli

Well, once we have a bunch of these smoketests we are going to run into the real serious issue that testing all of them all might just take several hours. So I think that ultimately the best would be to consume from source (not crate).

jbolda:
Well, these should run in parallel so a run should only ever take as long as the longest "branch" run. I think there may be a limit on the number of parallel runs though. At that time, I think there is still some value in running against prod, but we could maybe move that out of the PR runs and have it run on `master` or `dev` which will see much fewer commits (because squash).

nothingismagick:
Ok - well, I trust you on this one.

jbolda:
We need to do a bit of work on the artifact uploading, but I can follow up with another PR.

nothingismagick:
Screen Shot 2019-12-02 at 15 17 42

PR #131, opened 2019-12-03
[WIP] github actions rust cache

jbolda:
- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: github actions

- [ ] Yes. Issue #___
- [x] No

**Other information:**
Finishing up the cache uploads by setting the ext, then caching some of the Rust items.

### DIFF
--- a/.github/workflows/build-smoke-tests.yml
+++ b/.github/workflows/build-smoke-tests.yml
@@ -21,6 +21,16 @@ jobs:
       uses: actions-rs/toolchain@v1
       with:
           toolchain: stable
+    - name: cache rust
+      uses: actions/cache@v1
+      with:
+        path: |
+          bin/
+          registry/index/
+          registry/cache/
+          git/db/
+        key: ${{ runner.OS }}-build-
+        restore-keys: ${{ runner.OS }}-build-
     - run: cargo install tauri-cli
     - name: cache node modules
       uses: actions/cache@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/build-smoke-tests.yml
```diff
@@ -59,8 +77,12 @@ jobs:
             folder: vue/quasar-app
             executable: app
         include:
+          - platform: ubuntu-latest
+            ext: _0.1.0_amd64.deb
+          - platform: macos-latest
+            ext: .app
```
nothingismagick: we need to actually descend a bit deeper for macos - on my mac, they end up e.g. here:

`examples/vue/quasar-app/src-tauri/target/release/bundle/osx/app.app`

jbolda: 👍 I added another property that we can specify the release folder. However, is there any way we can make this more standard between platforms? I expect cross-platform building will be a selling point and that consistency would be helpful.

nothingismagick: This is something to address in the cargo bundler - but yes it should definitely be possible.

jbolda:
Still working out a couple last things, some tests are green that aren't working (which is fine because we wouldn't want failing cache to break the build).

I need to get the artifacts working for ubuntu still, and the rust caching isn't quite working yet.

PR #149, opened 2019-12-10
check cache hits before install

jbolda:
- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: Github Action perf

- [ ] Yes. Issue #___
- [x] No

**Other information:**
Check cache hits before `tauri-cli` install to prevent failure when it is already installed.

### DIFF
--- a/.github/workflows/build-smoke-tests.yml
+++ b/.github/workflows/build-smoke-tests.yml
@@ -29,21 +29,25 @@ jobs:
       with:
           toolchain: stable
     - name: cache rust bin
+      id: cache_rust_bin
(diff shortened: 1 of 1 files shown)


jbolda:
I think this is good to go. I don't know what was with those random failures though. It looks like this most recent run though, we didn't get cache hits so the most recent run didn't actually test anything. Regardless, I think this will at least work for now despite possibly taking longer in some cases, but that is a better alternative to just failing.

PR #160, opened 2019-12-14
[WIP] Refactor with Feature flags to remove Bloat and smooth out test pipeline

tensor-programming:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
There are a few dependencies and modules that are not necessary at the current moment. Some are even redundant (see the platform and version modules for more details).  The objective of this PR is to thin them out using cfg and feature flags. 

Also, due to the amount of environment variables required to compile the core library, a development environment pipeline needs to be established using a test project.  This can be done via default feature flags and CFG by way of replacing these mandatory environment variables with paths and values from a test project which will work only during development.


tensor-programming:
Thats fine.  I may end up remove most of this from the bundler for now since all three platforms use these dependencies (at least lazy static and handlebars) because of appimage, dmg and wix.  That being said, We can mostly just remove the feature flags and limit it down to platforms instead.  In the case of the core library, I found that you could setup settings using RLS.  Might be useful to make this easier on our devs.

jbolda:
So you mean dynamically infer the platform and choose the proper deps on that?

tensor-programming:
Yes, the cfg macros allow you to conditionally compile the code based on the platform etc.  At first, I was going to just use feature flags but you can't do features based on platform (at least not yet).  Windows requires 3-4 dependencies that linux and mac do not require.  Also, AppImage and DMG are not exactly finished; so its alright to leave them behind feature flags.  And if we are going to put AppImage and DMG into feature flags then might as well take the other two dependencies (lazy static and handlebars) and make them optional.  

Also, iOS and RPM have no real implementation (At least not one that works right now).  Its fine to just leave them out until they have a real implementation and even when they do, until we get cross compilation then there is no reason to include them by default. The CLI is likely the biggest bottleneck in terms of compile times.

What we will probably have to do (and honestly its probably better this way) is just add the features to the CLI as optional flags that the users can input when they go to compile the applications.  If I am on a debian based Linux Distro for instance and I run `Yarn Tauri Build`; currently, the app will make a .deb file.  The user should have the option to pass in some flag like `Yarn Tauri Build --appimage` to turn off .deb and turn on .appimage.  If they want both, then perhaps they can just put in something like `Yarn Tauri Build --appimage --deb` etc.  This is also something that we can just put into the config file; maybe add a build field that is platform specific etc.

nothingismagick:
Yeah, a bundle field in tauri.conf.js was always planned

tensor-programming:
Going to wait until no-server gets merged and then ill rebase and resolve the rest.

tensor-programming:
Going to roll back every change that isn't on the bundler.  Shouldn't be hard since its only like 2 of these commits.  The rest shouldn't conflict with the dev branch.

tensor-programming:
Closing due to #192

PR #199, opened 2019-12-22
Created .gitignore to ignore WixTools

BenoitRanque:
Compiling on windows downloads WixTools to src-tauri/WixTools folder.
We don't want that polluting our repo


BenoitRanque:
Might be more apropiate to add this in the .gitignore file in each example folder

PR #203, opened 2019-12-23
[needs review] Convert tauri.js to typescript

nklayman:

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes. Issue #___
- [ ] No

**Other information:**
I found a few errors report by typescript, ignored them, and marked with `TODO: ...`. These need to be reviewed. I am not confident that all of the typings are correct, the ones I am unsure of are marked with `TODO`. The only breaking change is for apps `require`ing the api files, which will now be `require`d out of `dist/*`.


cli/tauri.js/src/helpers/spawn.ts
```diff
@@ -1,24 +1,26 @@
...
+): number => {
   log(`Running "${cmd} ${params.join(' ')}"`)
   log()
 
-  const runner = crossSpawn(
-    cmd,
-    params, {
-      stdio: 'inherit',
-      stdout: 'inherit',
-      stderr: 'inherit',
```
Laegel: Did you remove those options intentionally?

cli/tauri.js/src/runner.ts
```diff
@@ -162,35 +192,44 @@ class Runner {
...
     return new Promise((resolve, reject) => {
       this.tauriWatcher && this.tauriWatcher.close()
-      this.__stopCargo().then(resolve)
+      this.__stopCargo()
+        .then(resolve)
+        .catch(e => {
+          console.error(e)
+        })
     })
   }
```
Laegel: This would be cleaner:
```typescript
async stop(): Promise<void> {
    this.tauriWatcher && this.tauriWatcher.close()
    try {
      await this.__stopCargo()
    } catch (e) {
      console.error(e)
    }
  }
```
Note: if we catch the error at this level, we won't catch anything when calling `stop` (c.f. line 30).

cli/tauri.js/src/runner.ts
```diff
@@ -162,35 +192,44 @@ class Runner {
...
 
-  __runCargoCommand ({
+  async __runCargoCommand({
```
Laegel: I think this method will need some refactor: why are we using `async` and a promise, here? It doesn't look useful at all since we're resolving right after the `spawn` call.

cli/tauri.js/src/generator.ts
```diff
@@ -0,0 +1,11 @@
+import { writeFileSync } from 'fs-extra'
+import path from 'path'
+import { tauriDir } from './helpers/app-paths'
+import { TauriConfig } from './types/config'
+
+export const generate = (tauriConfig: TauriConfig['tauri']): void => {
+  const { bundle, ...cfg } = tauriConfig
+  const outDir = tauriDir
+  writeFileSync(path.join(outDir, 'config.json'), JSON.stringify(cfg))
```
nothingismagick: This is where we no longer will need to do this if we use `tauri.conf.json`. @lucasfernog said he would adapt the rust core as soon as you modify the source of the configs from `tauri.conf.js`

cli/tauri.js/src/generator.ts
```diff
@@ -0,0 +1,11 @@
...
+  writeFileSync(path.join(outDir, 'bundle.json'), JSON.stringify(bundle))
```
nothingismagick: Same with this. No longer needed then with `tauri.conf.json`

cli/tauri.js/src/helpers/app-paths.ts
```diff
@@ -0,0 +1,29 @@
+import { existsSync } from 'fs'
+import { join, normalize, resolve, sep } from 'path'
+
+const getAppDir = (): string => {
+  let dir = process.cwd()
+  let count = 0
+
+  // only go up three folders max
+  while (dir.length > 0 && dir.endsWith(sep) && count <= 2) {
+    if (existsSync(join(dir, 'tauri.conf.js'))) {
```
nothingismagick: `tauri.conf.json`

cli/tauri.js/src/helpers/logger.ts
```diff
@@ -0,0 +1,23 @@
...
+    prevTime = curr
+
+    if (msg) {
+      console.log(
+        // TODO: proper typings for color and banner
+        // @ts-ignore
+        ` ${chalk[color](banner)} ${msg} ${chalk.green(`+${ms(diff)}`)}`
+      )
+    } else {
+      console.log()
```
nothingismagick: Would it make sense to use `process.stdout.write()` instead of `console.log()`? Its something I've been wondering about for a while. It would allow us to enable the linting rule for noconsole which I particularly prefer.

Laegel: Just seen this SO question: https://stackoverflow.com/questions/4976466/difference-between-process-stdout-write-and-console-log-in-node-js
But I dunno what would linters think about it instead.

nothingismagick:
Screen Shot 2019-12-23 at 15 40 35

cli/tauri.js/src/template.ts
```diff
@@ -25,20 +34,33 @@ const injectConfFile = (injectPath, { force, logging }) => {
...
 }
 
-const injectTemplate = (injectPath, { force, logging, tauriPath }) => {
+const injectTemplate = (
+  injectPath: string,
+  {
+    force,
+    logging,
+    tauriPath
+  }: {
```
outofmemoryagain: Not a must fix, more my opinion, but I think it would be good to decide on the Tauri style in general when it comes to using inline type definitions for function parameters.  Although allowed, I generally avoid doing so as I find it makes the code less readable/maintainable, especially to the TS novice.  I generally recommend defining an explicit interface for the type so in this case `InjectTemplateParams` or something like it.  It also helps to avoid possible duplicate type definitions as it is easier to reuse if need in the future, vs just defining the same type again.

cli/tauri.js/src/template.ts
```diff
@@ -52,16 +74,19 @@ Run \`tauri init --force template\` to overwrite.`)
...
- */
-const inject = (injectPath, type, { force = false, logging = false, tauriPath = null }) => {
+const inject = (
+  injectPath: string,
+  type: 'conf' | 'template' | 'all',
+  {
+    force = false,
+    logging = false,
+    tauriPath
+  }: {
```
outofmemoryagain: An example of a duplicate a type definition, from my above comment on inline type definitions.

cli/tauri.js/src/types/config.ts
```diff
@@ -0,0 +1,43 @@
+// TODO: Clean up types, properly mark which ones are optional
+// May need to have different types for each stage of config generation process
+
+export interface TauriConfig {
```
outofmemoryagain: It may be a good idea to consider decomposing the config into different types for each section `build`, `ctx`...  This would better support composing the config by section or passing a single section as a typed object instead of [`any`](https://github.com/tauri-apps/tauri/pull/203/files#diff-f519598f155e7dcc17b4186459b02e32R14).  It would also support extending a single section if needed in the future instead of the entire config.

nothingismagick: I should note that the `tauri.conf.js` file will soon be a `tauri.conf.json` => although that probably isn't really relevant for these typings.

cli/tauri.js/src/runner.ts
nklayman: There is a recommended linter rule that says all functions that return promises should be async. We can disable that if you prefer, but I kinda like it.

cli/tauri.js/src/helpers/spawn.ts
nklayman: According to the crossSpawn typings, those were invalid options. I should have ignored them and then added a comment. However, stdio covers stdin/out/error, so it should have the same behavior.

cli/tauri.js/src/runner.ts
Laegel: I guess it's for the sake of explicitness. I see nothing harmful keeping that rule, then. I'm just gonna update my own firmware, haha. 
Nevertheless, the `async` keyword is not the only problem, here. Anyway, it's out of this PR scope and not blocking.

cli/tauri.js/src/helpers/logger.ts
rajivshah3: > It would allow us to enable the linting rule for noconsole which I particularly prefer.

If you want to disable the lint rule for a specific area you can do:

```js
/*eslint-disable-next-line no-console*/
console.log('hi')

cli/tauri.js/src/template.ts
nklayman: Done! Thanks for the suggestion! 437ce947f165a0db38b6383c9d25afc0d26cc274

cli/tauri.js/src/template.ts
nklayman: Done! Thanks for the suggestion! 437ce947f165a0db38b6383c9d25afc0d26cc274

nklayman:
@nothingismagick shouldn't we have a seperate PR for the `tauri.config.json`? That way we can put the ~js~ts updates and the rust updates in one branch/pr? I think this is ready to be merged, and I will create a new branch for the json config.

nothingismagick:
Yeah, you're right.

PR #217, opened 2019-12-24
chore(tauri.js): Add prepare script

rajivshah3:

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
Adds a [`prepare` script](https://docs.npmjs.com/misc/scripts.html) to `tauri.js`. This script is run before publishing and when running `yarn` to install dependencies to ensure that the TS files are built.

### DIFF
--- a/cli/tauri.js/package.json
+++ b/cli/tauri.js/package.json
@@ -12,6 +12,7 @@
   "scripts": {
     "build": "webpack --progress",
     "test": "jest --runInBand --no-cache",
+    "prepare": "yarn build",
     "pretest": "yarn build",
     "test:mac-local": "jest --runInBand",
     "lint": "eslint --ext ts ./src/**/*.ts",


jbolda:
This is partly related, but do we need to configure for the `npm publish` to use the new `dist`?

rajivshah3:
> do we need to configure for the `npm publish` to use the new `dist`?

Since the intended behavior of the `prepare` script is to run after `yarn`/`npm install` and before `yarn publish`/`npm publish`, I don't think any further configuration is needed. The `dist` folder should be uploaded with the rest of the package when it's published

jbolda:
Perhaps I'm missing some configuration, but won't this include both the `dist` and the `src`? No need for the redundency and it's just extra stuff that needs to be downloaded.

rajivshah3:
Oooh, that's a good point, yeah. So there are two options if you want to exclude `src` from the published package:

- Add `dist` to the `files` property of the `package.json`:
  - `react-native-community/cli` did this: https://github.com/react-native-community/cli/blob/master/packages/cli/package.json#L13-L16
  - See [`files` section](https://docs.npmjs.com/files/package.json#files) of the `package.json` docs
- Add `src` to `cli/tauri.js/.npmignore`

jbolda:
I tend to prefer the `files` property myself. It is more explicit and lessens the chance of accidentally including newly added files.

PR #214, opened 2019-12-24
feat(tauri) use config as JSON

lucasfernog:
- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes. Issue #___
- [x] No

**The PR fulfills these requirements:**

- [x] It's submitted to the `dev` branch and _not_ the `master` branch
- [ ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #xxx[,#xxx]`, where "xxx" is the issue number)

If adding a **new feature**, the PR's description includes:
- [ ] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**

This PRs changes the tauri-config parsing, to evaluate `src-tauri/tauri.conf.json` instead of `tauri.conf.js`, to make it easier to read it on Rust. Thanks to @nklayman for starting this and doing most of the Node CLI work.

PR #233, opened 2019-12-26
fix(bundler): Return nonzero exit code when errors occur

### LINKED ISSUE #202: Tauri CLI does not handle promise rejections
Currently, tauri.js cli does not handle promise rejections properly.
This can be seen if for example one attempts to run `tauri build` while build assets are not present.
The command line _will_ display the error, but as it is an unhandled promise rejection, the cli exits with a zero exit code. This could become a problem in the future, if we test the cli exit code.

Fixing this behavior requires multiple changes, and could be approached two ways:
Either we wrap all `await` expressions with `try {} catch (e) {}`, or we have a single `try {} catch (e) {}` at the root of the cli. I prefer this second approach.

Using this second approach will require refactor in multiple places. I believe a typescript version of the cli is in the works, so maybe we could wait for that to be done before tackling this problem.

**To Reproduce**
run `tauri build` while the assets (index.html) are not available.

**Explanation of JS's behavior**

Short version: await all promises

<details><summary>Long version</summary>

There are many [gotchas](https://jakearchibald.com/2017/await-vs-return-vs-return-await/) to async js and error handling. Here is a simple showcase of some of the gotchas of async JS.

```js
entry() // execute the entry function
async function entry () {
  try {
    a() // we are not awaiting this promise. mistake.
    b() // this function is the same as a. Async functions return promises
    await c() // this function does not await the promises inside
    await d() // this one finally does work as intended
  } catch (err) {
    // this will only catch d
    console.error('caught error', err)
  }
}
async function a () {
  throw new Error('Catch this!')
}
function b () {
  return new Promise((resolve, reject) => {
    reject(new Error('Catch that!'))
  })
}
async function c () {
  return a() // we should be awaiting this
}
async function d () {
  return await b() // finally the right signature!
}
```
</details>

**Changes that must be made**
1. Catch and handle all errors at cli root
2. All promises must be awaited.

I've not been able to come up with jest tests for this

### LINKED ISSUE #231: tauri.js CLI: Errors from the Rust CLI don't propagate up to the JS CLI
**Describe the bug**
It seems like errors reported by the Rust CLI do not propagate to the JS CLI. In other words, the Rust CLI exits with a non-zero exit code, yet the JS CLI does not. Since the JS CLI returns an exit code of 0, the command is considered to have passed. This poses a problem for our own smoke tests in particular -- in the event that this occurs, GitHub Actions will [mark the step as "passed"](https://github.com/tauri-apps/tauri/pull/188#issuecomment-567963724) even though it actually should fail.

Examples where this has occurred:
https://github.com/tauri-apps/tauri/pull/187/checks?check_run_id=357905199#step:11:112
https://github.com/tauri-apps/tauri/pull/214/checks?check_run_id=363576771#step:14:130

**To Reproduce**
1. Run a command from the JS CLI (e.g. `yarn tauri build`) which causes an error to be thrown from the Rust CLI

**Expected behavior**
Non-zero error codes from the Rust CLI should result in a non-zero error code from the JS CLI

**Platform (please complete the following information):**
This doesn't appear to be a platform-specific issue

rajivshah3:

- [x] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
Fixes #231 
Fixes #202 (already closed)

This should fix the issues where CI builds pass when they should actually fail (see above issues for context).

I was originally looking at the JS/TS side and wondered if `process.exit()` was causing processes to exit with `0`, but when I tried changing that the error was still not solved. After digging through the Rust side I came across these areas:

https://github.com/tauri-apps/tauri/blob/f800daaf442a5a6a6304b25f2b49c08963966e13/cli/tauri-cli/src/main.rs#L65-L68
https://github.com/tauri-apps/tauri/blob/f800daaf442a5a6a6304b25f2b49c08963966e13/cli/tauri-cli/src/main.rs#L156

It turns out that `bundle::print_error` was actually swallowing the errors:
https://github.com/tauri-apps/tauri/blob/f800daaf442a5a6a6304b25f2b49c08963966e13/cli/tauri-cli/src/bundle/common.rs#L228

This fix was tested by running a command that [used to succeed](https://github.com/tauri-apps/tauri/pull/214/checks?check_run_id=363576771#step:14:133) and now properly fails:
image
image

### DIFF
--- a/cli/tauri-cli/src/bundle/common.rs
+++ b/cli/tauri-cli/src/bundle/common.rs
@@ -237,7 +237,7 @@ pub fn print_error(error: &crate::Error) -> crate::Result<()> {
       writeln!(output, "{:?}", backtrace)?;
     }
     output.flush()?;
-    Ok(())
+    std::process::exit(1)
   }
 }
 
--- a/cli/tauri-cli/src/bundle/common.rs
+++ b/cli/tauri-cli/src/bundle/common.rs
@@ -225,7 +225,7 @@ pub fn print_error(error: &crate::Error) -> crate::Result<()> {
       writeln!(output, "{:?}", backtrace)?;
     }
     output.flush()?;
-    Ok(())
+    std::process::exit(1)
   } else {
     let mut output = io::stderr();
     write!(output, "error:")?;


nothingismagick:
❤️ @rajivshah3

PR #260, opened 2020-01-01
feat(examples): add vue cli app example

nklayman:

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: Add example
 
- [ ] Yes. Issue #___
- [x] No

**Other information:**

Example created with latest version of vue cli and latest [vue cli plugin tauri](https://github.com/tauri-apps/vue-cli-plugin-tauri).


nklayman:
@jbolda The vue cli plugin has a custom build command in place of `tauri build`. How should I use that instead? I could just skip `tauri build` and make `yarn build` run the command via the `package.json` scripts.

jbolda:
We just call `yarn tauri:source:build` and `yarn tauri:prod:build` in the production smoke tests and dev smoke tests, respectively. As long as that script targets your cli plugin, it should Just Work™️.

I noticed there was already `"tauri:build": "vue-cli-service tauri:build"` in the package.json. Is that recursively calling itself? I'm not familiar with the `vue-cli-service` in all honesty.

nklayman:
The `vue-cli-service tauri:build` runs the plugin's build command, which runs the vue cli build and then calls `tauri build` by importing the api file and running the exported function.

nklayman:
Since all I can run is a prod smoke test anyway, would it be fine to just have the tests in the vue-cli-plugin-tauri repo? I already have an e2e test there that I can hook up to github actions.

jbolda:
Ah, I see, nifty. If I am reading this right, you could possibly have this run from source by making the `const build = require('tauri/dist/api/build')` configurable to point at the source.

I guess I'm of the opinion that if we are going to have smoke tests run in this repo, it should use source. The prod tests are less worthwhile, but are there to catch obvious oversights (which in the case of this example, would have, but you caught it right away 😎). I imagine we will want to add more of these plugins for the various frameworks over time so maybe it is worth establishing "best practices for integration" that you and the @tauri-apps/admins are comfortable with.

That being said, if you are e2e testing in the separate repo for prod. Is there some value I'm not seeing with adding it in this repo also testing for prod? I believe the source test _does_ make a lot of sense, but that obviously requires changes to the plugin. Looking to find the right balance of catching mistakes vs adding more things here that need to be maintained. It already sounds as if there is some interest in not having any examples in this repo.

nothingismagick:
The added value in having it here is that we will know if changes break the smoke test.

jbolda:
But it only creates that value if we test it against source, right?

nothingismagick:
Yeah, we need to wire that up (the test against source).

jbolda:
Is it worth merging this in so at least there is easy access to a manual test, and then follow up PR the wiring?

nothingismagick:
Agreed. Then we can work on that in the meantime.

PR #277, opened 2020-01-03
chore(workflow): add global release changelog

jbolda:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
This will create a draft release with all of the changes since the last tag. We can use it for the "marketing" changes / changelog, where each package can then still get its own changelog file. The bulk of the configuration sits in the tauri-pps/.github repo.

Let me know if this is not how you want to go about things. Ideally I think it would be great if we can use changesets, but the current code won't be able to support our repo. I have commented on an issue in regards to the possibility. This is a bit of a stopgap, and still has value even if we do add in changesets.

### DIFF
--- a/.github/workflows/push-to-dev.yml
+++ b/.github/workflows/push-to-dev.yml
@@ -1,0 +1,14 @@
+name: dev push
+
+on:
+  push:
+    branches:
+      - dev
+
+jobs:
+  update_release_draft:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: release-drafter/release-drafter@v5.3.1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


nothingismagick:
If you are ready to merge @jbolda - then please go ahead.

lucasfernog:
I think our configuration here https://github.com/tauri-apps/.github/blob/master/.github/release-drafter.yml should use the "feat" keyword aswell

jbolda:
@lucasfernog those are actual labels that we need to apply, not the commit messages.

jbolda:
Do you think that is fine? The "real" changelogs can/should work off of conventional commits in my mind, but I can see the reasoning behind having both work that way.

tensor-programming:
Here are some references: https://github.com/rust-lang/rustup/blob/master/CHANGELOG.md and https://github.com/flutter/flutter/wiki/Changelog.  In both of these instances, the labels are not really a big deal.  Whats important is that the logs reference the PRs involved.

jbolda:
In this case, the labels drive this plugin which creates a Github release draft with all of the PRs since the last tag. The labels are purely for grouping PRs in this changelog. I am considering this the "global" changelog though, as we still want each package to have a changelog file similar to your examples @tensor-programming.

tensor-programming:
I get that; just pointing out that it really shouldn't matter if we are using "feat" or "feature". Our feat branches should also contain "feat/feature" in some manner anyhow in the title of the branch or PR.

jbolda:
Ah, sorry, I mischaracterized your comment. Agreed.

I think I will merge this in, and add bumped to this process. It isn't perfect for us, but should be workable for the time being until we can find/improve something else.

PR #352, opened 2020-01-25
feat(bundler&tauri) add wix resource bundling and utils to get the path to the platform resource dir

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**


nothingismagick:
It does not work properly on MacOS yet, I’ve identified a weird situation where node is called if I execute the binary with `open app.app` - but double clicking the icon in the finder will not get it to work.

tauri-utils/src/platform.rs
```diff
@@ -51,3 +53,28 @@ pub fn target_triple() -> Result<String, crate::Error> {
 
   Ok(format!("{}-{}", arch, os))
 }
+
+pub fn resource_dir() -> Result<PathBuf, crate::Error> {
```
tensor-programming: Should return `crate::Result<PathBuf>` instead.

tauri-utils/src/platform.rs
```diff
@@ -51,3 +53,28 @@ pub fn target_triple() -> Result<String, crate::Error> {
...
+  let exe = std::env::current_exe()?;
+  let exe_dir = exe.parent().expect("failed to get exe directory");
```
tensor-programming: these expects can be cleaned up with `?`s.  If they don't work then add the errors to the `error_chain` macro.

cli/tauri-cli/src/bundle/wix.rs
```diff
@@ -448,3 +508,74 @@ fn generate_external_binary_data(settings: &Settings) -> crate::Result<Vec<Exter
...
+    }).collect::<Vec<_>>();
+    directories.truncate(directories.len() - 1);
+    
+    // transform the directory structure to a chained vec structure
+    for directory in directories {
+      let directory_name = directory.as_os_str().to_os_string().into_string().expect("failed to read resource folder name");
+
+      // if the directory is already on the map
+      if resources.contains_key(&directory_name) {
+        let directory_entry = &mut resources.get_mut(&directory_name).unwrap();
```
tensor-programming: get rid of this `unwrap()`.  Should either be a `?` if is a result or an `expect()` if is a Option.

cli/tauri-cli/src/bundle/wix.rs
```diff
@@ -448,3 +508,74 @@ fn generate_external_binary_data(settings: &Settings) -> crate::Result<Vec<Exter
...
+        if directory_entry.name == directory_name { // the directory entry is the root of the chain
+          directory_entry.add_file(resource_entry.clone());
+        } else {
+          let index = directory_entry.directories.iter().position(|f| f.name == directory_name);
+          if index.is_some() { // the directory entry is already a part of the chain
+            let dir = directory_entry.directories.get_mut(index.unwrap()).unwrap();
```
tensor-programming: Again, address these unwraps().

cli/tauri-cli/src/bundle/wix.rs
```diff
@@ -60,6 +60,52 @@ struct ExternalBinary {
...
+  directories: Vec<ResourceDirectory>
+}
+
+impl ResourceDirectory {
+  fn add_file(&mut self, file: ResourceFile) {
+    self.files.push(file);
+  }
+
+  // generates the wix XML string to bundle this directory resources recursively
+  fn get_wix_data(self) -> (String, Vec<String>) {
```
tensor-programming: This here should return a `crate::Result<(String, Vec<String>)>`.  Gives more options for trouble shooting and error handling.

cli/tauri-cli/src/bundle/wix.rs
```diff
@@ -60,6 +60,52 @@ struct ExternalBinary {
...
+    }
+    let mut directories = String::from("");
+    for directory in self.directories {
+      let (wix_string, ids) = directory.get_wix_data();
+      for id in ids {
+        file_ids.push(id)
+      }
+      directories.push_str(wix_string.as_str());
+    }
+    (format!(r#"<Directory Id="{name}" Name="{name}">{contents}</Directory>"#, name = self.name, contents = format!("{}{}", files, directories)), file_ids)
```
tensor-programming: Refactor the large `format!` string into a let binding to make the code a bit clearer.

cli/tauri-cli/src/bundle/wix.rs
```diff
@@ -448,3 +508,74 @@ fn generate_external_binary_data(settings: &Settings) -> crate::Result<Vec<Exter
 
   Ok(external_binaries)
 }
+
+type ResourceMap = BTreeMap<String, ResourceDirectory>;
```
tensor-programming: Push this line up to the top of the file.

cli/tauri-cli/src/bundle/wix.rs
lucasfernog: so I should always use expect, even when I know the result is Ok()? In this case I've already checked if the key exists.

tauri-utils/src/platform.rs
lucasfernog: I think I've tried to do that, but those are Options instead of Results.

tensor-programming: Alright, if they are options then the expects are fine.

cli/tauri-cli/src/bundle/wix.rs
tensor-programming: Even if you know it exists, its a bit more expressive that way.  And in the case where you are using a result, you could use a `?` instead of an expect to bubble the error upwards.

tensor-programming:
So long as it passes the tests, it should be good to merge.

PR #354, opened 2020-01-26
feat(tauri&tauri.js) allow inliner/loadAsset to be used with publicPath

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

- [ ] It's submitted to the `dev` branch and _not_ the `master` branch

**Other information:**


tauri/src/endpoints.rs
```diff
@@ -203,14 +204,33 @@ fn load_asset<T: 'static>(
...
     move || {
-      let read_asset = crate::assets::ASSETS.get(&format!(
-        "{}{}{}",
-        env!("TAURI_DIST_DIR"),
-        if asset.starts_with("/") { "" } else { "/" },
-        asset
-      ));
-      if read_asset.is_err() {
-        return Err(r#""Asset not found""#.to_string());
+      let mut path = PathBuf::from(
```
tensor-programming: Would consider reworking this with an while let instead of a loop.  Lets you combine the match with the loop and implicitly break when the if statement comes back with false. 

```
let mut path = PathBuf::from(if asset.starts_with("/") {
        asset.replacen("/", "", 1)
      } else {
        asset.clone()
      });

      while let Some(component) = path.iter().next() {
        let read_asset = crate::assets::ASSETS.get(&format!(
          "{}/{}",
          env!("TAURI_DIST_DIR"),
          path.to_string_lossy()
        ));

        if read_asset.is_err() {
          let first_component = component.to_str().expect("failed to read path component");

          path = PathBuf::from(path.to_string_lossy().replacen(
            format!("{}/", first_component).as_str(),
            "",
            1,
          ));
        }
      };
```

lucasfernog: Well I thought about that, just didn't know exactly how to use it. Thanks!

lucasfernog: But in that case, I need to break when the read_asset is an Ok value. I don't know if it's better to use an while let here, since I would need to break inside anyway.

tensor-programming: While let breaks if the match isn't true.  So in this case you could just put the while let on read_asset, something like
```
while let Ok(read_asset) = crate::assets::ASSETS.get(...
   match path.iter().next() {
    ...
``` 

In this case, it would loop over the `crate::assets::ASSETS.get` until the pattern no longer matches and then break.

lucasfernog: But I have to loop while the read_asset is an Err, and use that value later. Can it still be done?

tensor-programming: In that case you could write it like this:
```
while let Err(read_asset) = crate::assets::ASSETS.get(&format!(
          "{}/{}",
          env!("TAURI_DIST_DIR"),
          path.to_string_lossy()
        ));
        match path.iter().next() {
            Some(component) => {
              let first_component = component.to_str().expect("failed to read path component");
              path = PathBuf::from(path.to_string_lossy().replacen(format!("{}/", first_component).as_str(), "", 1));
          }
            None => {
              return Err(format!("Asset '{}' not found", asset).into());
          }    
        } 
```

With this block, it only loops if the `crate::assets::ASSETS.get(..)` resolves to an `Err(..)` type.  Essentially you are combining the if statement with the loop block.  It will break implicitly if the pattern fails to match.  Though also in this case you wouldn't be using read_asset so you could just use an underscore until you decide to use it.

lucasfernog: but with that I wouldn't be able to use the Result read_asset I think?

tensor-programming: If thats the case, you can just write it like this: 

```
 let read_asset = loop {
        let read_asset = crate::assets::ASSETS.get(&format!(
          "{}/{}",
          env!("TAURI_DIST_DIR"),
          path.to_string_lossy()
        ));
        if read_asset.is_err() {
          match path.iter().next() {
            Some(component) => {
              let first_component = component.to_str().expect("failed to read path component");
              path = PathBuf::from(path.to_string_lossy().replacen(
                format!("{}/", first_component).as_str(),
                "",
                1,
              ));
            }
            None => {
              return Err(format!("Asset '{}' not found", asset).into());
            }
          }
        } else {
          break read_asset;
        }
      };
```

PR #469, opened 2020-02-29
feature(tauri-bundler): add start menu shortcut on windows

jbolda:
This modifies the Wix xml template to add a start menu shortcut on install. I also reorganized a tad, and changed the registry root on the uninstall shortcut to be either user or system as msi can be either (which removes the existing warning we had).

This is also believed to be the last remaining item for the [guijs Windows build](https://github.com/Akryum/guijs/issues/34).

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
ref: https://wixtoolset.org/documentation/manual/v3/howtos/files_and_registry/create_start_menu_shortcut.html

### DIFF
--- a/cli/tauri-bundler/src/bundle/templates/main.wxs
+++ b/cli/tauri-bundler/src/bundle/templates/main.wxs
@@ -39,21 +39,23 @@
 
         <Directory Id="TARGETDIR" Name="SourceDir">
             <Directory Id="$(var.PlatformProgramFilesFolder)" Name="PFiles">
-                <Directory Id="APPLICATIONFOLDER" Name="{{{product_name}}}">
-                    <Component Id="Path" Guid="{{{path_component_guid}}}" Win64="$(var.Win64)" KeyPath="yes">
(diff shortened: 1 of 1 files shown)


lucasfernog:
Does it still bundle resources and external binaries? I assume you tested it with guijs so yes, just checking.

jbolda:
This shouldn't affect the install into the folder so I would be surprised if it affects guijs. This only creates a shortcut to the main executable. I actually have only tried with a couple examples.

lucasfernog:
I'm making this question because of the reorganization you made, since you changed the code about bundling resources and binaries.

jbolda:
Ah, fair point. I just shifted it out and used a reference. It may be worth a test in guijs though. I can maybe try it later today.

lucasfernog:
Tested it, working great!

PR #472, opened 2020-02-29
feat(tauri.js/init): prompt for default values (fix #422/#162)

### LINKED ISSUE #422: Tauri init and first time use
When first installing and running tauri init, we should provide an input for a few of the major required settings specifically the path settings and app name. At the completion of the init script, we can note where these config options were stored to make it easier for the user to find them.

Additionally, we can add a `readme.md` in `src-tauri` with the simple instructions for running and point them to our docs for further help. This will give a helping hand to anyone using tauri "indirectly".

nklayman:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes. Issue #422/#162
- [ ] No

**Other information:**
`tauri init` will now prompt for the window title, distDir, and devPath via inquirer. The prompts can be skipped if the values are passed via cli arguments. This is a breaking change as the init command requires user input to run. Before merging, are there any other values that should be prompted for? Binary/installer name come to mind but I'm not sure how to configure that.


nothingismagick:
I don't like the breaking nature of this, and think we should use a default in order to retain current behaviour. One idea might be to also allow them to choose a "pattern" that then sets up their `tauri.config.json` (and potentially their `main.rs`) for e.g. the Kamikaze.

lucasfernog:
Yeah maybe starter kits is the way to go, but this inquirer approach is good for now, just need to lock it with some kind of CLI arg.

nklayman:
I don't like breaking changes either but imo the default should be to prompt for the settings as they are required and are often wrong by default and could leave people very confused.

nothingismagick:
what's the status of this PR? should we merge?

nklayman:
It can be merged, but I want to make sure there aren't any suggestions for other values that should prompted for. It would be nice to be able to prompt for installer/binary name and output dir but I'm not sure if this is possible.

nothingismagick:
Are we still waiting for something here @nklayman ?

PR #486, opened 2020-03-04
build(task-runner): add self-documented task runner

DrSensor:

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [x] Other, please describe: developer experience for contributor or maintainer

- [ ] Yes. Issue #___
- [x] No

**Other information:**
As a new/first contributor, I have difficulty to check the small changes that I made is correct, especially when I want to check if it works on the [examples repo](https://github.com/tauri-apps/examples). I'm choosing [mask](https://github.com/jakedeichert/mask) because it's markdown which can be used as some sort of recipe. (also I'm more familiar with `mask` codebase)
tauri-mask-run-example

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,3 +83,6 @@ Cargo.lock
 # doing this because of how our tests currently (naively) drop the tauri.conf.js in that folder
 # todo: needs a proper fic
 /cli/tauri.js/tauri.conf.js
+
+# doing this because the task-runner (`mask prepare`) will clone gh:tauri-apps/examples
+/examples
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,5 +3,8 @@ members = [
   "tauri",
   "tauri-api",
   "tauri-updater",
-  "tauri-utils"
+  "tauri-utils",
+]
+exclude = [
+  "examples",
 ]
--- a/maskfile.md
+++ b/maskfile.md
@@ -33,7 +33,9 @@ case "$PWD" in
   yarn && yarn tauri:source dev
 ;;
 */rust/*)
-  cargo web deploy && cd ..
+  cargo web deploy
+  [ $example = `basename $(dirname $PWD)` ] && cd ..
+
   yarn add tauri@link:../../../cli/tauri.js
   yarn && yarn tauri dev
 ;; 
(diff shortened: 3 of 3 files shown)


lucasfernog:
Wow this is so cool.

Btw do we have a contribution guide?

nothingismagick:
@lucasfernog => https://github.com/tauri-apps/tauri/wiki/14.-Contributor-Introduction

jbolda:
@DrSensor very cool! I have been working on a good workflow using yarn workspaces for the examples, but that won't cover the rust based examples. This looks like it has the capability though which is great. It appears that this may not work on Windows though which is unfortunate 😣.

My other thought is that I wonder if this is best added to the examples repo?

jbolda:
I was actually thinking _only_ have the maskfile in https://github.com/tauri-apps/examples as the functions relate specifically to them. I guess my personal opinion regarding the examples is that they serve strictly as test grounds and the expectation is that they won't be cloned to start a project. Instead we push towards using the CLI to start a project. What do others think? It seems to always be an ongoing discussion.

The paths certainly could be an issue on Windows, but if it works in bash and WSL that's better than nothing.

nothingismagick:
@jbolda - can't we do both?

jbolda:
@nothingismagick If that is what you would like to do, it's not unreasonable. We would have to maintain it in both places, but there are worse situations.

tensor-programming:
Ill look into making it work on windows but bash and wsl only support is a bit unacceptable. Most windows devs don't use either or would rather not use either.

nothingismagick:
@tensor-programming - with your powershell amendment this could be approved?

tensor-programming:
its being looked at here: https://github.com/jakedeichert/mask/pull/45

They said they would review it next week.

lucasfernog:
can this be merged already?

jbolda:
We were waiting on the upstream next version with @tensor-programming changes.

tensor-programming:
This can likely be merged, the changes where made a few weeks ago.

jbolda:
Thanks @DrSensor for taking care of this!

PR #503, opened 2020-03-09
feat(tauri) webview user_data added to AppBuilder api

jazzay:
- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

This feature allows the application (user) to pass in a user_data value through to the webview where it can be later accessed via the webview.get_user_data() apis


tauri/src/app.rs
```diff
@@ -37,30 +38,44 @@ impl App {
...
+  invoke_handler: Option<InvokeHandler<T>>,
+  setup: Option<Setup<T>>,
+  splashscreen_html: Option<String>,
+  user_data: Option<T>
 }
 
-impl AppBuilder {
+// Default builder with no user_data
+impl AppBuilder<()> {
   pub fn new() -> Self {
```
khionu, line 50: What would be more idiomatic is to have `new` take a `user_data`, and have the default be a `Default` implementation.

jazzay: I agree that would be nicer. However, my goal was to avoid breaking all users apps, due to the public api change. If we're fine with that then I could make that change. Thoughts?

khionu: We're in late Alpha, that's fine.

tensor-programming: Wouldn't worry about breaking anything. As @khionu said, have the new take a user_data and then just implement the default trait.

khionu:
Said argument replacement is partially happening with the new bindings. We can talk about details on Discord.

lucasfernog:
Closing it since we're going to implement it differently with the new bindings.

PR #513, opened 2020-03-15
feat(tauri) add path API

lucasfernog:
**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes. Issue #___
- [x] No

**The PR fulfills these requirements:**

- [x] It's submitted to the `dev` branch and _not_ the `master` branch
- [ ] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #xxx[,#xxx]`, where "xxx" is the issue number)

If adding a **new feature**, the PR's description includes:
- [ ] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**

PR #580, opened 2020-05-06
Support writing binary file to local disk

rnons:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

Hi, thanks for the great work on `tauri`. I found `tauri` two days ago, it's really interesting, as a result, this is my first time writing Rust. My use case is I want to write an image to local disk, what I found is [this comment](https://github.com/tauri-apps/tauri/pull/513#discussion_r392956379). So I think it's probably worth to add a `writeBinaryFile` api.


rnons:
> Rescinding my previous comment, I would suggest instead this change the `file_system::write_file` function to take a `Vec<u8>`.

Do you mean this?
```diff
pub fn write_binary_file<T: 'static>(
  webview: &mut WebView<'_, T>,
  file: String,
- contents: String,
+ contents: Vec<u8>,
```

I tried it at first, but doesn't seem to work. As I understand, all the arguments from the JavaScript side  go through a `JSON.stringify` call, then passed to the Rust side. The `contents` field will either be a JavaScript string or an object. It doesn't automatically become `Vec<u8>`, or maybe I miss something?

In this PR, `contents` is base64 encoded string, then `base64::decode` is used to get `Vec<u8>`. Do you want me to change to `contents: ByteString`? ByteString is more efficient than base64, but I need to introduce a new dependency that supports ByteString. What do you think?

khionu:
Hmmm... tough call. Optimal changes to support this might be a bit bigger than should be discussed in a PR. Mind marking this PR as WIP and opening an issue prompting discussion? Key issue would be that Rust Strings have to be UTF-8 and we should support writing raw bytes to a file.

nothingismagick:
@lucasfernog - since we are rebuilding the API right now, shouldn't we add this in? Khionu's commentary was a sort of blocker, and I think it would be useful in some cases (although I would potentially classify it as a security risk).

lucasfernog:
Definitely, I'll review it and test later today.

nothingismagick:
right on. then we can merge it into dev and then dev into the api-ts branch

lucasfernog:
Thanks for your contribution, and sorry it took so long!

rnons:
Thanks for taking care of it.

PR #624, opened 2020-05-28
don't remove link attributes on html rewrite when in dev

chippers:
- [x] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

When running with the embedded server off, `tauri dev` will still remove attributes from `<link>`s even though we are trying to run these files from the server.  This change will make sure that the attributes are not removed during dev `cfg.ctx.dev`.


nothingismagick:
Thanks - good call!

lucasfernog:
The attributes removal is probably something we want to do only on no-server mode, but we can track that later.

PR #643, opened 2020-06-06
feat(updater): Alpha version

lemarier:
Hi guys!

I think I'm ready for the first round of reviews.

To understand how the updater work please refer to the [README](https://github.com/tauri-apps/tauri/blob/feature/new_updater/tauri-updater/README.md)

This is an early Alpha version allowing us to test if it works. The sequence is not smooth as it should but it should works. It's been tested on mac, windows, and Linux (ubuntu)

## Tests with built artifacts

Built on: [Thu Mar 23 2021 at 9 AM EDT](https://github.com/tauri-apps/tauri/actions/runs/664447786).

### Windows

Download the MSI [here](https://github.com/lemarier/tauri-test/releases/download/v0.1.0/updater-example_0.1.0_x64.msi) -- Once installed launch, it and once you launch the app, you should get asked to update. It should launch a new MSI, follow step, relaunch app you should be updated (no more popup)

### Mac

Download the latest app [here](https://github.com/lemarier/tauri-test/releases/download/v0.1.0/updater-example_0.1.0_x64.app.tar.gz) -- Copy it inside /Applications, launch the app, you should be asked to update, follow step. Restart the app -- You shouldn't be prompted to update.

### Linux

Download AppImage from [here](https://github.com/lemarier/tauri-test/releases/download/v0.1.0/updater-example_0.1.0_amd64.AppImage.tar.gz) -- Launch the app, You should be prompted to update. Follow step. Restart should be done automatically. You should be updated to v2

## Tests
- [updater api tests](https://github.com/tauri-apps/tauri/blob/feature/new_updater/tauri-updater/src/lib.rs#L669)

## Known bugs
- [x] mac -- once installed the app close but do not restart (not implemented yet)
- [x] mac -- app need to be added inside /Applications for the updater to work? Probably a packager (manifest) issue
- [x] win -- when MSI is launched, the app do not close (should be spawned)
- [x] win -- wix can't kill the process as it's not spawned so I guess wix is a child of tauri (force kill the process manually of tauri app) fixed it

## Todo
- [x] Unit/Smoke tests
- [x] tauri.js (cli) integration for key generation
- [x] Better error handling 
- [x] Cleanup
- [x] MacOS tests and fix
- [x] Windows tests and fix
- [x] Linux tests and fix
- [x] If the public key is provided inside tauri.conf.json, when installing an update we need a valid signature
- [x] KeyPair generation [bundler]
- [x] Executable signature for the update [bundler]
- [x] Archive generator (generate tar.gz & zip automatically on release)
- [x] Allow fallback URLs (if one URL failed or is down, we fallback to next update URL)
- [x] FIX: When extracting overwrite existing file only without deleting existing 
- [x] Instead of bundling a zip & tar.gz, push AppImage, Wix (Installer) and mac tar.gz of the Application folder)
- [x] Consume the API via Tauri


.github/workflows/artifacts-updater.yml
```diff
@@ -0,0 +1,66 @@
...
+  pull_request:
+    branches:
+      - feature/new_updater
+
+jobs:
+  build-artifacs:
+    runs-on: ${{ matrix.platform }}
+
+    env:
+      DEVELOPER_DIR: /Applications/Xcode_11.3.1.app/Contents/Developer
```
rajivshah3: Are we trying to force a specific Xcode version? I think running `xcode-select -s /Applications/Xcode_11.3.1.app/Contents/Developer` is more appropriate (it's made specifically for this purpose)

cli/tauri-bundler/src/main.rs
```diff
@@ -111,6 +118,72 @@ fn run() -> crate::Result<()> {
...
+            .requires("generate-key")
+            .conflicts_with("password")
+            .help("Set empty password for your private key"),
+        )
+        .arg(
+          Arg::with_name("force")
+            .long("force")
+            .short("f")
+            .requires_all(&["generate-key", "write-private-key"])
+            .help("Overwrite private key even if exist on the specified path"),
```
rajivshah3:
```suggestion
            .help("Overwrite private key even if it exists on the specified path"),
```

tauri-updater/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+The update JSON Tauri requests should be dynamically generated based on criteria in the request, and whether an update is required.
+
+Tauri's installer is also designed to be fault-tolerant, and ensure that any updates installed are valid and safe.
+
+# Configuration
+
+Once you have your Tauri project ready, you need to configure the updater.
+
+Add this in tauri.config.js
```
rajivshah3:
```suggestion
Add this in tauri.config.json
```

tauri-updater/README.md
```diff
@@ -0,0 +1,275 @@
...
+```json
+"updater": {
+    "active": true,
+    "endpoints": [
+        "https://releases.myapp.com/{target}}/{current_version}}"
+    ],
+    "dialog": true,
+    "pubkey": "",
```
rajivshah3:
```suggestion
    "pubkey": ""
```

Since the Tauri config file is now JSON, having a comma after the last property is not allowed

tauri-updater/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+# Signing updates
+
+We offer a built-in signature to ensure your update is safe to be installed.
+
+To sign your updates, you need two things.
+
+The *Public-key* (pubkey) should be added inside your tauri.config.js to validate the update archive before installing.
+
+The *Private key* (privkey) is used to sign your update and should NEVER be shared with anyone. Also, if you lost this key, you'll NOT be able to publish a new update to the current user base (if pubkey is set in tauri.config.js). It's important to save it at a safe place and you can always access it.
```
rajivshah3:
```suggestion
The *Public-key* (pubkey) should be added inside your tauri.config.json to validate the update archive before installing.

The *Private key* (privkey) is used to sign your update and should NEVER be shared with anyone. Also, if you lost this key, you'll NOT be able to publish a new update to the current user base (if pubkey is set in tauri.config.json). It's important to save it at a safe place and you can always access it.
```

tauri-updater/README.md
```diff
@@ -0,0 +1,275 @@
...
+    -p, --private-key-path <PATH>            Load the private key from a file
+    -s, --sign <PATH>                        Sign the specified binary
+    -w, --write-private-key <DESTINATION>    Write private key to a file
+```
+
+***
+
+Environment variables used to sign:
+
+`TAURI_PRIVATE_KEY`  Path or String of your private key
```
rajivshah3, line 325: How does the CLI tell whether the value of this variable is a path or the private key itself?

lemarier: You can see it here:
https://github.com/tauri-apps/tauri/blob/feature/new_updater/cli/tauri-bundler/src/bundle/updater_bundle.rs#L42

It check if the file exist if not it expect to be a string

.github/workflows/artifacts-updater.yml
lemarier: Yes exactly! Thanks I will check because the bundled app wasn’t working on macOS 10.14 — but I think it’s not needed anymore as the fix has been done within the build.rs and app config. Thanks for reminding me

tauri-updater/README.md
rajivshah3: Ah whoops, not sure how I missed that

cli/tauri-bundler/src/main.rs
```diff
@@ -1,17 +1,26 @@
 mod bundle;
 mod error;
+mod sign;
+
 pub use error::{Error, Result};
 
-use crate::bundle::{bundle_project, check_icons, PackageType, Settings};
+use crate::bundle::{bundle_project, check_icons, BuildArtifact, PackageType, Settings};
```
Laegel: `BuildArtifact` seems to come from nowhere, I had to remove it to install Tauri bundler.

Milo123459:
Could it be possible to make the autoupdater a bit like the one in VSCcode? Or am I understanding this wrong. FYI, you start it up and then it doesn't make you re-run an installer or whatever, it just automatically does it. I think that'd be cool.

nothingismagick:
@Milo123459 - yes, that is the intention here.

Milo123459:
Fantastic! I imagine it'll require code signing, right?

lemarier:
> Fantastic! I imagine it'll require code signing, right?

Yep! We support it
You can check details here

https://github.com/tauri-apps/tauri/blob/feature/new_updater/tauri-updater/README.md

nothingismagick (changes requested):
We will need to add this project to the .changes/config.json file, as well as adding a .changes file too. @jbolda can you assist here?

tauri/src/app/event.rs
```diff
@@ -119,6 +119,15 @@ pub fn unlisten(event_id: EventId) {
   })
 }
 
+/// Removes an event listener by event name.
+pub fn unlisten_by_name(event_name: impl AsRef<str>) {
```
lemarier: @lucasfernog want to inform you that I added this

lucasfernog: you shouldn't need that; i'll take a look at it when I review the PR.

api/src/updater.ts
```diff
@@ -0,0 +1,129 @@
...
+  body: string
+}
+
+export interface UpdateResult {
+  manifest?: UpdateManifest
+  shouldUpdate: boolean
+}
+
+export async function installUpdate(): Promise<void> {
+  const allUnlisteners: UnlistenFn[] = []
```
lucasfernog: seems like this will always have just one item; if that's the case we could just store a single fn

api/src/updater.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  function cleanAllListener(): void {
+    allUnlisteners.forEach((unlistenFn) => {
+      unlistenFn()
+    })
+  }
+
+  return new Promise((resolve, reject) => {
+    function onStatusChange(statusResult: UpdateStatusResult): void {
+      console.log({ statusResult })
```
lucasfernog: we should remove the console.logs

cli/core/config_definition.rs
```diff
@@ -560,3 +576,12 @@ fn default_build() -> BuildConfig {
     with_global_tauri: false,
   }
 }
+
+fn default_updater() -> UpdaterConfig {
```
lucasfernog, line 611: we're using a custom `impl Default for UpdaterConfig` now; also i think we could just use the derive Default impl since the default value for `active` will be false anyway, and it should just work.

cli/core/src/sign.rs
```diff
@@ -0,0 +1,150 @@
+use std::path::{Path, PathBuf};
+use tauri_bundler::sign::{generate_key, read_key_from_file, save_keypair, sign_file};
+
+pub struct Signer {
+  private_key: Option<String>,
+  password: Option<String>,
+  binary: Option<PathBuf>,
+}
+
+impl Default for Signer {
```
lucasfernog: we can use `#[derive(Default)]` on the struct

examples/api/src/components/Updater.svelte
```diff
@@ -0,0 +1,46 @@
...
+      onMessage(e);
+    }
+  }
+
+  async function install() {
+    try {
+      document.getElementById("start_update").classList.add("hidden");
+
+      await installUpdate();
+      onMessage("Installation complete, restart required.");
```
lucasfernog, line 33: can we offer an API to restart the app too?

cli/tauri-bundler/src/sign.rs
```diff
@@ -0,0 +1,156 @@
+extern crate minisign;
```
lucasfernog: I think the sign step could be moved to the rust CLI instead of the bundler

tauri-updater/README.md
```diff
@@ -0,0 +1,320 @@
+# Tauri Updater
```
lucasfernog, line 1: the README looks good but we'll need to revisit it before release and also add docs, cc @Laegel @nklayman

tauri-utils/src/config.rs
```diff
@@ -155,6 +155,39 @@ impl Default for WindowConfig {
...
+  /// Display built-in dialog or use event system if disabled.
+  #[serde(default = "default_updater_dialog")]
+  pub dialog: bool,
+}
+
+fn default_updater_dialog() -> bool {
+  true
+}
+
+impl Default for UpdaterConfig {
```
lucasfernog, line 180: same as the config on the cli (use derive)

cli/tauri-bundler/src/error.rs
```diff
@@ -54,6 +53,14 @@ pub enum Error {
   ShellScriptError(String),
   #[error("`{0}`")]
   GenericError(String),
+  #[error("Unable to create updater package")]
+  UpdateBundler,
+  #[error("Decode Error: `{0}`")]
```
wusyong:
```suggestion
  #[error(transparent)]
```
You could use `#[error(transparent)]` to forward the message from other errors.

tauri-utils/src/config.rs
```diff
@@ -155,6 +155,39 @@ impl Default for WindowConfig {
   }
 }
 
+/// The Updater configuration object.
+#[derive(PartialEq, Deserialize, Debug, Clone)]
+#[serde(tag = "updaterConfig", rename_all = "camelCase")]
+pub struct UpdaterConfig {
+  /// Whether the updater is active or not.
+  #[serde(default)]
```
wusyong, line 163: No need to add this if the field name stay the same iirc.

cli/tauri-bundler/src/error.rs
```diff
@@ -54,6 +53,14 @@ pub enum Error {
   ShellScriptError(String),
   #[error("`{0}`")]
```
tensor-programming, line 52: the description for these errors should be more descriptive.  As it is right now, the panics would just be whatever the defaults would be. Even just adding `generic error: {0}` would be better.

tauri-api/src/error.rs
```diff
@@ -43,6 +46,10 @@ pub enum Error {
   /// IO error.
   #[error("{0}")]
   Io(#[from] std::io::Error),
+  /// Ignore error.
+  #[error("{0}")]
```
tensor-programming: again, maybe give more description for these errors.

tauri-updater/src/error.rs
```diff
@@ -0,0 +1,39 @@
+use thiserror::Error as DeriveError;
+
+#[derive(Debug, DeriveError)]
+pub enum Error {
```
tensor-programming, line 4: This entire error block should have more descriptive descriptions.

tauri-updater/src/lib.rs
```diff
@@ -1,19 +1,1034 @@
...
+  pub signature: Option<String>,
+  pub should_update: bool,
+}
+
+impl RemoteRelease {
+  // Read JSON and confirm this is a valid Schema
+  fn from_release(release: &serde_json::Value, target: &str) -> Result<RemoteRelease> {
+    // did we have a platforms field ?
+    // if we did, that mean it's a static JSON
+    let is_dynamic_update = release["platforms"].is_null();
```
tensor-programming: careful with these `is_null()` calls.  If you can avoid them try to not use them.  They can add undefined behavior to the code.

tauri-updater/src/lib.rs
```diff
@@ -1,19 +1,1034 @@
...
+    let extract_path = self.extract_path.clone();
+
+    // make sure we NEED to install it ...
+
+    // make sure we can install the update on linux
+    // We fail here because later we can add more linux support
+    // actually if we use APPIMAGE, our extract path should already
+    // be set with our APPIMAGE env variable, we don't need to do
+    // anythin with it yet
+    if target == "linux" && env::var_os("APPIMAGE").is_none() {
```
tensor-programming: coiuld use a `if cfg!(linux)` instead. Might be cleaner too since you won't need to bring in the target and clone it.  Could also rewrite this as an `if let` block.

tauri-updater/src/lib.rs
```diff
@@ -1,19 +1,1034 @@
...
+
+    // make sure we have a valid agent
+    if !headers.contains_key(header::USER_AGENT) {
+      headers.insert(
+        header::USER_AGENT,
+        "tauri/updater".parse().expect("invalid user-agent"),
+      );
+    }
+
+    // Set SSL for linux
```
tensor-programming: If this is for linux only maybe add a `cfg` block.

tauri-updater/src/lib.rs
```diff
@@ -1,19 +1,1034 @@
...
+      )
+    }
+
+    tmp_archive.write_all(&resp.bytes().await?)?;
+
+    // Validate signature ONLY if pubkey is available in tauri.conf.json
+    if let Some(pub_key) = pub_key {
+      // we need an announced signature by the server
+      // if there is no signature, bail out.
+      if self.signature.is_none() {
```
tensor-programming: could be rewritten to `if let None = self.signature { ... }`  Actually surprised clippy didn't complain about this.

tauri-updater/src/lib.rs
```diff
@@ -1,19 +1,1034 @@
...
+    }
+  }
+
+  Ok(())
+}
+
+/// Returns a target os -- If none,
+/// that mean the updater didnt support
+/// the platform
+pub fn get_target() -> Option<String> {
```
tensor-programming: Don't we have this logic somewhere else in the code base?  Maybe should import it from utils or something.

tauri-updater/src/lib.rs
```diff
@@ -1,19 +1,1034 @@
...
+    .map(PathBuf::from)
+    .expect("Can't determine extract path");
+
+  let extract_path_as_string = extract_path.display().to_string();
+
+  // MacOS example binary is in /Applications/TestApp.app/Contents/MacOS/myApp
+  // We need to get /Applications/TestApp.app
+  // todo(lemarier): Need a better way here
+  // Maybe we could search for <*.app> to get the right path
+  if target == "darwin" && extract_path_as_string.contains("Contents/MacOS") {
```
tensor-programming: again could use a `cfg!(target = darwin)` block.

tauri-updater/src/lib.rs
```diff
@@ -1,19 +1,1034 @@
...
+// by our tests in the bundler
+pub fn verify_signature(
+  archive_path: &PathBuf,
+  release_signature: String,
+  pub_key: &str,
+) -> crate::Result<bool> {
+  // we need to convert the pub key
+  let pub_key_decoded = &base64_to_string(pub_key)?;
+  let public_key = PublicKey::decode(pub_key_decoded);
+  if public_key.is_err() {
```
tensor-programming: could be written as `if let Err(_) = public_key { ... }`  Might be more useful that way because you can embed the error into the `bail!` call.

tauri-updater/src/lib.rs
```diff
@@ -1,19 +1,1034 @@
...
+    bail!(
+      crate::Error::Updater,
+      "Something went wrong with pubkey decode"
+    )
+  }
+
+  let public_key_ready = public_key.expect("Something wrong with the public key");
```
tensor-programming: are these `expect` calls necessary? Would be careful to make sure you don't want the code to panic if it doesn't have to.

tauri-updater/src/lib.rs
```diff
@@ -1,19 +1,1034 @@
...
+mod test {
+  use super::*;
+  #[cfg(target_os = "macos")]
+  use std::env::current_exe;
+  #[cfg(target_os = "macos")]
+  use std::fs::File;
+  #[cfg(target_os = "macos")]
+  use std::path::Path;
+
+  macro_rules! aw {
```
tensor-programming: maybe make this a little more descriptive than just `aw!`.  Something like `block!`

tauri-updater/src/macros.rs
```diff
@@ -24,3 +12,23 @@ macro_rules! set_ssl_vars {
...
+    ($e_type:expr, $literal:expr) => {
+        $e_type(format!($literal))
+    };
+    ($e_type:expr, $literal:expr, $($arg:expr),*) => {
+        $e_type(format!($literal, $($arg),*))
+    };
+}
+
+/// Helper for formatting `errors::Error`s and returning early
+macro_rules! bail {
```
tensor-programming: Could use anyhow's `bail!` instead to get the same logic but with more context.

tauri/src/app/event.rs
tensor-programming: `remove_listener` might be a better name for this. Or at least `stop_listening`, `unbind_listener` etc.

tauri/src/error.rs
```diff
@@ -49,6 +49,9 @@ pub enum Error {
   /// Invalid args when running a command.
   #[error("invalid args for command `{0}`: {1}")]
   InvalidArgs(&'static str, serde_json::Error),
+  /// Tauri updater error.
+  #[error("{0}")]
```
tensor-programming: again could use a bit more description.

tauri/src/updater.rs
```diff
@@ -0,0 +1,285 @@
...
+    Some(StatusEvent {
+      error,
+      status: String::from(status),
+    }),
+  );
+}
+
+// Prompt a dialog asking if the user want to install the new version
+// Maybe we should add an option to customize it in future versions.
+async fn ask_if_should_install(
```
tensor-programming: maybe call it `prompt_for_uninstall`.  Just a little more clear.

tensor-programming (commented):
Everything looks pretty solid logic wise.  Couple of names could be improved and the errors need to be more descriptive.

cli/tauri-bundler/src/error.rs
lemarier: Thanks!

lucasfernog:
@lemarier I tried the example from https://github.com/lemarier/tauri-test/releases/download/v0.1.0/updater-example_0.1.0_amd64.AppImage.tar.gz and it has some weird behavior, after downloading the updated version it opens it, but it keeps the current one opened and ask me to restart it.

lemarier (commented):
> @lemarier I tried the example from https://github.com/lemarier/tauri-test/releases/download/v0.1.0/updater-example_0.1.0_amd64.AppImage.tar.gz and it has some weird behavior, after downloading the updated version it opens it, but it keeps the current one opened and ask me to restart it.



> @lemarier I tried the example from https://github.com/lemarier/tauri-test/releases/download/v0.1.0/updater-example_0.1.0_amd64.AppImage.tar.gz and it has some weird behavior, after downloading the updated version it opens it, but it keeps the current one opened and ask me to restart it.

Hi Lucas! 
Could you give a try with this : (I updated the Linux build)
https://github.com/lemarier/tauri-test/releases/download/v0.1.0/updater-example_0.1.0_x64.app.tar.gz

Thanks

lucasfernog (commented):
> > @lemarier I tried the example from [lemarier/tauri-test@`v0.1.0` updater-example_0.1.0_amd64.AppImage.tar.gz (download)](https://github.com/lemarier/tauri-test/releases/download/v0.1.0/updater-example_0.1.0_amd64.AppImage.tar.gz) and it has some weird behavior, after downloading the updated version it opens it, but it keeps the current one opened and ask me to restart it.
> 
> Hi Lucas!
> Could you give a try with this : (I updated the Linux build)
> [lemarier/tauri-test@`v0.1.0` updater-example_0.1.0_x64.app.tar.gz (download)](https://github.com/lemarier/tauri-test/releases/download/v0.1.0/updater-example_0.1.0_x64.app.tar.gz)
> 
> Thanks

Hey, it works! Nice!

cli/core/config_definition.rs
lemarier: Yes, you are right! The main reason I was using a custom Impl, it's because by default, `dialog = true` 
is it possible to overwrite the field value to true with Derive Default without a custom impl?
or we could rename the field `dialog` for `disable_dialog` or something like that? and default to false will be all good.

lucasfernog: yeah serde has a default attribute, there's some examples in this file.

lemarier: yeah but the test is failing because the `UpdaterConfig::Default` didn't return the dialog set to true, JSON will work ok with serde deserialization, but not the UpdaterConfig::Default called from tests 
when I add a `serde(default = "default_dialog")`  it's for deserialization

lucasfernog: yeah you're right :P

cli/tauri-bundler/src/sign.rs
lemarier: Sure! Before I proceed I need to validate few things.

Because right now how it works;

Bundler generates custom updater artifacts as you can see here, and then sign the `zip` or `tar.gz`
https://github.com/tauri-apps/tauri/pull/643/files#diff-4f1d416019fc14a2e07e8f0d6c99717ebe1ec898337945bdad8b13535acb5c42R45

If we want to have the sign in the core, we need to be able to know what's the updater artifact path, right now, we run the build without reading paths:
https://github.com/tauri-apps/tauri/blob/feature/new_updater/cli/core/src/build.rs#L162

I can read them and sign all `zip` and `tar.gz` as it _should be_ the updater artifacts.

As you probably already noticed, the signer is required in the core as well;
https://github.com/tauri-apps/tauri/blob/feature/new_updater/cli/core/src/sign.rs#L64

Should I modify the return of `bundle_project` to be able to pass a type with the PathBuf so I know at 100% its an updater artifact?

examples/api/src/components/Updater.svelte
lemarier: Good catch! I'm not sure if this should be provided by the updater? What do you think? 

Maybe another API, where we can include the manifest (current version, app name, etc..) from the newly added  `package_info` in the tauri context?

We could add some specific functions like relaunch and quit similar to what electron is doing.

cli/tauri-bundler/src/sign.rs
lucasfernog: That's a good idea :)

examples/api/src/components/Updater.svelte
lucasfernog: we can do on a separate PR, I just thought about it since you're already doing the functionality, we just need to reorganize.

cli/tauri-bundler/src/sign.rs
lemarier: here we go!

Screen Shot 2021-03-24 at 12 26 15 PM

examples/api/src/components/Updater.svelte
lemarier: I'll open an Issue to be sure and mark this as resolved.

lemarier:
I think I'm done. Need to update the latest build to test

Edit: Binary updated.
You can find them here: https://github.com/lemarier/tauri-test/releases/tag/v0.1.0 or in the PR description.
OSX and Windows are NOT signed so you can have some weird behavior. I suggest disabling all firewalls for the tests. We made tests with signed binary and everything worked as expected. Also, on OSX you'll need to cmd + click on the app to open it. And move it inside `/Applications` is a good suggestion to replicate the real environment.

Windows we are relaunching the MSI installation. I think it's OK for now. Maybe we could launch the MSI in silent and let it go. We could also generate specific MSI for the updater if needed. We could also only replace binary like we are doing in OSX/Linux but we should get stability in the updater first I think. Because we all know windows is specific and can drive weird errors. If we can isolate them it's better.

nothingismagick (approved):

HuakunShen:
Hello! @lemarier @nothingismagick 

May I know where I can find this tauri-update-server and how to deploy it?

HuakunShen:
OK, I found it https://github.com/lemarier/tauri-update-server

The documentation doesn't seem. to be up to date, some links are expired.

lemarier:
> OK, I found it https://github.com/lemarier/tauri-update-server
> 
> The documentation doesn't seem. to be up to date, some links are expired.

Hi @HuakunShen,

These POCs may be a bit outdated, instead of `darwin` as platform, tauri now expect `darwin-x86_64` by example.

You can use a static JSON as well and host it on amazon S3 by example or a gist and use the RAW 

doc can be found here;
https://tauri.app/v1/guides/distribution/updater/#update-file-json-format

example can be found here;
https://gist.githubusercontent.com/lemarier/6bc28d06b94958d83eaa2790eef03f32/raw/d920848ad037af5c3781cef04bdb7fff6c23a9c3/static_update.json

If you want to implement you own update server, you can refer to this for the complete spec;
https://tauri.app/v1/guides/distribution/updater/#server-support

Hope this help.

PR #655, opened 2020-06-11
use cargo lockfile for tauri dependency during `tauri info`. close #610

### LINKED ISSUE #610: More accurate tauri version in `tauri info`
This is a nice-to-have that I thought of while adding #609. Currently, when using `tauri info` it will read the version for tauri directly from the `Cargo.toml` where it's listed as a dependency. This makes it inconsistent in the following ways.

1. All of the other versions listed in the info include the patch version. This won't happen for tauri if you declare the version for the dependency as `"*"` or `"0.5"` instead of `"0.5.2"`.
2. The version actually used might be a different/newer version than what's in the `Cargo.toml` due to updating the lockfile.

To be clear, I'm talking about the `tauri` entry in the `App` section from `tauri info`.

Ex:

```sh
App
  tauri - 0.5
```

compared to:

```sh
Rust environment
  rustc - 1.43.1
  cargo - 1.43.0
  tauri-bundler - 0.6.0
```

Pull the version for the `tauri` crate during `tauri info` from `Cargo.lock` instead. Inside the lockfile, packages are listed as an array of tables because there can be multiples of the same package with different versions. After filtering the packages to find `tauri`, we might find that there are actually multiple versions in use. Due to this new possibility that there are multiple versions, here is how I propose that case should be handled.

---

While searching the lockfile we find both `0.5.2` and `0.4.3`. We then check `Cargo.toml` to check the version the user declared to make the output clearer.

`Cargo.toml`
```toml
[dependencies]
tauri = "0.5"
```

`tauri info`
```sh
App
  tauri - 0.5 (0.5.2, 0.4.3)
```

---

The other display behaviors should be the same, as it already handles not finding the version as `unknown`. When only a single package is found, reading the `Cargo.toml` file to find the declared version should not happen as we already know the only version in use.

I'm willing to write the PR for this, but I figured I should get feedback first since this would be a change of behavior from the current implementation.

chippers:
- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #610
- [x] No

**Other information:**
As mentioned in the issue, the new possibility is multiple versions found in the lockfile. This looks like:
new_format

I marked this as non-breaking as it only changes the possible output of the `tauri info` command, which I didn't see as a non-breaking guarantee. This could break something that parses the output and strictly expects only a semver string or `"unknown"`.

### DIFF
--- a/cli/tauri.js/src/api/info.ts
+++ b/cli/tauri.js/src/api/info.ts
@@ -6,7 +6,7 @@ import os from 'os'
 import path from 'path'
 import { appDir, tauriDir } from '../helpers/app-paths'
 import { TauriConfig } from './../types/config'
-import { CargoToml } from './../types/cargo'
+import { CargoLock, CargoManifest } from '../types/cargo'
 import nonWebpackRequire from '../helpers/non-webpack-require'
 import { version } from '../../package.json'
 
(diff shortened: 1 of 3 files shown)


lucasfernog:
Nice work!

PR #675, opened 2020-06-14
fix(cli): tell user of missing lockfile or manifest during info

chippers:
- [x] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
Expands on the fix in #674.  It will handle any error (like it did before the original change for #610) except it won't swallow them.  The original block was wrapped in an empty `try { ... } catch(_) {}` which I didn't replicate in the code replacing it.  This new fix will handle any read error for toml files, and communicate when the expected files are missing when attempted to be read.

no lockfile

no manifest no lockfile

It will also now read the manifest along with the lockfile so that it can communicate a missing manifest even if there is a lockfile.

image

### DIFF
--- a/.changes/cli_info_toml_error_handling.md
+++ b/.changes/cli_info_toml_error_handling.md
@@ -1,0 +1,5 @@
+---
+"tauri.js": patch
+---
+
+Handle and communicate missing lockfile and/or manifest while reading versions during `tauri info`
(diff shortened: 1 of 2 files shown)


lucasfernog:
Awesome PR as always, thank you!

PR #702, opened 2020-06-19
fix(tauri) httpRequest API JSON/Text response reader, header value as string, closes #691

### LINKED ISSUE #691: String value in HttpRequestOptions header field gets escaped incorrectly
**Describe the bug**
when you pass `String` value to  `params` or `header` field in `HttpRequestOptions` it will get escaped and adds `"` to both start and end, and this added `"` will get treated as a part of value by the receiver.

<details>
<summary>==== description with example ====</summary>

```rust
let mut header: HashMap<String, Value> = HashMap::new();
let mut params: HashMap<String, Value> = HashMap::new();

header.insert(String::from("Authorization"), String::from("Bearer AUTHTOKEN").into());
params.insert(String::from("foo"), String::from("bar").into());

let option = HttpRequestOptions {
    method: String::from("get"),
    url: String::from("https://postman-echo.com/get"),
    headers: Some(header),
    params: Some(params),
    ...
    response_type: Some(ResponseType::Text),
};

println!("Result: {}", make_request(option).unwrap());
```

This print produces following result (some detail has been omitted and formatted)
```
{
  "args": { 
    "part": "\"id\"",
    "mine": "true"
  },
  "headers": {
    "host": "postman-echo.com",
    "x-forwarded-proto":"https",
    "x-forwarded-port":"443",
    "x-amzn-trace-id":"Root=1-5eeaf475-955a64323b68e37d4180becd",
    "authorization":"\"Bearer ACCESS_TOKEN.aeab-eearwer\"", 
    ... 
```
value of `authorization` field in `header` object should not have escaped `"`.

</details>

**To Reproduce**
1. Create new Tauri project and open `src-tauri/main.rs`
2. in `main` function, create `HttpRequestOptions` and set `header` to have at least one or more `String` value, and set `url` to any HTTP url that echoes HTTP request (I tested with Postman Echo)
3. call `make_request` with given option and observe the result.

**Expected behavior**
Postman echoes correctly formed header, expected like
```
{
  "args": { "foo": "bar" },
  "header": { .., "authorization": "Bearer  AUTHTOKEN", .. },
  ...
}
```

**Screenshots**

OS: `Linux(5.6.13-arch1-1) - linux/x64`
Node: `14.4.0`
NPM: `6.14.5`
Yarn: `1.22.4`
Rustc: `1.44.0`

**Additional context**
If I understand correctly the `HttpRequestOptions` is shared between JS and Rust which forces any HashMap within this struct to have `HashMap<String, Value>` type.

I also tried to test from js side using `get(url: string, options?: HttpOptions): Promise<any>` and had a bit different behavior, and I'm not quite sure if this is a bug related to this issue, or an entirely different thing, or just simply I'm using it wrong.

English is not my native language, I'm sorry if there is any grammar error that makes it hard to understand.
**Stack Trace**

lucasfernog:
- [x] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

Re: #691

### DIFF
--- a/tauri/src/endpoints/http.rs
+++ b/tauri/src/endpoints/http.rs
@@ -14,7 +14,10 @@ pub fn make_request<T: 'static>(
       let response_type = options.response_type.clone();
       request(options).map(
         |response| match response_type.unwrap_or(ResponseType::Json) {
-          ResponseType::Text => format!(r#""{}""#, response),
+          ResponseType::Text => format!(
+            r#""{}""#,
+            response.replace(r#"""#, r#"\""#).replace(r#"\\""#, r#"\""#)
+          ),
           _ => response,
         },
       )
--- a/tauri-api/src/http.rs
+++ b/tauri-api/src/http.rs
@@ -43,9 +43,9 @@ pub struct HttpRequestOptions {
   /// The request URL
   pub url: String,
   /// The request query params
-  pub params: Option<HashMap<String, Value>>,
+  pub params: Option<HashMap<String, String>>,
   /// The request headers
-  pub headers: Option<HashMap<String, Value>>,
+  pub headers: Option<HashMap<String, String>>,
   /// The request body
   pub body: Option<Value>,
   /// Whether to follow redirects or not
@@ -90,9 +90,9 @@ pub struct HttpRequestBuilder {
   /// The request URL
   pub url: String,
   /// The request query params
-  pub params: Option<HashMap<String, Value>>,
+  pub params: Option<HashMap<String, String>>,
   /// The request headers
-  pub headers: Option<HashMap<String, Value>>,
+  pub headers: Option<HashMap<String, String>>,
   /// The request body
   pub body: Option<Value>,
   /// Whether to follow redirects or not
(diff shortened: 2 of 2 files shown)


tauri/src/endpoints/http.rs
```diff
@@ -14,7 +14,10 @@ pub fn make_request<T: 'static>(
       let response_type = options.response_type.clone();
       request(options).map(
         |response| match response_type.unwrap_or(ResponseType::Json) {
-          ResponseType::Text => format!(r#""{}""#, response),
+          ResponseType::Text => format!(
+            r#""{}""#,
+            response.replace(r#"""#, r#"\""#).replace(r#"\\""#, r#"\""#)
```
nothingismagick, line 19: there's the magick. nice one

lucasfernog: we may need more of those on the codebase because the way we eval() error messages, will think about a clever way to do it.

jbolda:
Think we need a change file on this?

Issue #778, opened 2020-07-09
ERROR: tauri.conf.json > tauri > embeddedServer has unknown property port +0ms


thienpow:
try configure
```json
"embeddedServer": {
  "active": true,
  "port": "3000"
}
```
yarn tauri dev or yarn tauri build would throw the error:
ERROR: tauri.conf.json > tauri > embeddedServer has unknown property port +0ms

lucasfernog:
Closed by #782 and #784

PR #782, opened 2020-07-09
to enable custom port

thienpow:
current validator doesn't enable port config, with this patch would have allow the config, however this doesn't include the runner changes... runner might need changes too.

#778

### DIFF
--- a/cli/tauri.js/src/types/config.schema.json
+++ b/cli/tauri.js/src/types/config.schema.json
@@ -401,6 +401,7 @@
               "type": "boolean"
             },
             "port": {
+              "description": "enable custom port, instead of using random port",
               "type": "string"
             }
           },
--- a/cli/tauri.js/src/types/config.schema.json
+++ b/cli/tauri.js/src/types/config.schema.json
@@ -399,6 +399,9 @@
             "active": {
               "description": "whether we should use the embedded-server or the no-server mode",
               "type": "boolean"
+            },
+            "port": {
+              "type": "string"
             }
           },
           "type": "object"
--- a/cli/tauri.js/src/types/config.ts
+++ b/cli/tauri.js/src/types/config.ts
@@ -223,6 +223,7 @@ export interface TauriConfig {
        * whether we should use the embedded-server or the no-server mode
        */
       active?: boolean
+      port?: string
     }
     /**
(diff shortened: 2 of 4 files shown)


cli/tauri.js/src/types/config.schema.json
```diff
@@ -399,6 +399,9 @@
             "active": {
               "description": "whether we should use the embedded-server or the no-server mode",
               "type": "boolean"
+            },
+            "port": {
+              "type": "string"
```
nothingismagick, line 405: I think we should add a description, because it should be clear that this is a fixed port.

cli/tauri.js/src/types/config.ts
```diff
@@ -223,6 +223,7 @@ export interface TauriConfig {
        * whether we should use the embedded-server or the no-server mode
        */
       active?: boolean
+      port?: string
```
nothingismagick, line 226: Do we really want to use a `string` here, or would a number be more appropriate? @lucasfernog

nothingismagick (commented):
Thanks for the detailed work! Just a couple questions to think about.

cli/tauri.js/src/types/config.schema.json
thienpow: at the rust runner code, it do compare if port == "random", unless changing that too..

cli/tauri.js/src/types/config.ts
thienpow: at the rust runner code, it do compare if port == "random", unless changing that too..

cli/tauri.js/src/types/config.schema.json
lucasfernog: yeah just add a JSDoc to the type definition like some of the other types in there

cli/tauri.js/src/types/config.ts
lucasfernog: to be a number the rust code needs changes

nothingismagick: its not a big deal at all

nothingismagick (approved):

nothingismagick:
@thienpow - I have manually reviewed the code and its ok, but in the future please consider signing your commits.

PR #809, opened 2020-07-11
feat(tauri.js) move exported api types into api modules (fix #807)

### LINKED ISSUE #807: Move tauri.js type exports to modules they are used in
I would like to have
```ts
import { emit, listen, Event } from "tauri/api/event";
```

instead of
```ts
import { emit, listen } from "tauri/api/event";
import { Event } from "tauri/api/types/event";
```

Move the types declared in the `api-src/types` dir into the modules they are used in at `api-src`

**Alternative solutions**
The types can also be re-exported by the modules that use them so that they can be imported as a single import. This would have the same type be importable from 2 different paths and may lead IDEs to prompt which path is desired to be added instead of automatically importing.

**Additional context**
I'm not sure if the current separation of types and non-type exports is purposeful and what that purpose may be.  Otherwise, this should be a relatively minor change, but would ultimately end up effecting the public api.

I'll write up a PR implementing the solution suggested in a few minutes.

chippers:
- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes. Issue #807
- [ ] No

**Other information:**
Allows importing types along with code that would be used with them.
```ts
import { emit, listen, Event } from "tauri/api/event";
```
instead of
```ts
import { emit, listen } from "tauri/api/event";
import { Event } from "tauri/api/types/event";
```

This change purposefully has minimal changes, and only directly moves previous type definitions into the modules that use them.


chippers:
I've force-pushed the change file.  I accidentally only added the newly created file, but not the actual content of the change file

nothingismagick (approved):
Thanks @chippers - LGTM!

lucasfernog:
@nothingismagick should we wait for next patch release (#808) before merging this one? We have a lot of new features coming, maybe a `minor` just for this one is too much :D

jbolda:
would it be worth a new branch for "next minor" that we merge this and other minor things into until we are ready?

lucasfernog:
That makes sense but it's an unusual situation, we won't do that often.

PR #830, opened 2020-07-15
(extremely) experimental changes to remove build env vars

chippers:
Hi, this work isn't really meant to be merged.  It's an experimental change to try and remove the requirements on `TAURI_DIST_DIR` and `TAURI_DIR` environmental variables when building tauri apps.  It does this by removing those requirements on the `tauri` and `tauri-api` compile-time and runtime assumptions.

It's currently only been tested with `no-server` with a blank project generated by `yarn tauri init`.  It's also been built once with `yarn tauri build` to generate the `index.tauri.html` file.  It can then be run with a simple `cargo run` (with `no-server` feature).
image

Things that haven't been worked on yet:
1. Tests
  * Some changes need to be made, such as examples run with `test = true` so that we can find the correct `tauri.conf.json` for that example.  Many tests still need to change from the old `config::get()` function and either generate a fake config file for testing or a simple default config.  Some tests need an actual fake file as it tests functions that read the filesystem
2. embedded-server
  * I haven't even looked at `embedded-server` yet.  `no-server` is what I'm used to using anyways + seems less complicated for a proof of concept.
3. proc-macro features
  * Previously, features were used to pass down `no-server` and `embedded-server` to the crates that used them at build + run time.  Something similar can be done with the new proc macro
4. cli
  * I haven't even touched the bundler or node cli, and those would be required to change if something like this eventually comes through.  I've used `yarn tauri build` to generate the `index.tauri.html` but didnt' use it for building.
5. everything else
  * I've changed a minimal amount of things as possible while still building with `no-server` without the env vars.  There are a bunch more considerations that I have not yet looked at and need to be tested more.

I've also had to change `tauri_includedir` and `tauri_includedir_codegen` slightly.  You can [see those changes here](https://github.com/tauri-apps/tauri-includedir/compare/61a7526c5d30998e7c96a68e230353848a195703...chippers:tauri_no_build_env).

Posting this draft PR for some feedback towards building without environmental variables


chippers:
Here's some more notes, I'll try and write them down when I think of things.

Here's what a new `main.rs` looks like:

```rs
use tauri::{AppBuilder, FromTauriConfig};

#[derive(FromTauriConfig)]
struct Config;

fn main() {
  AppBuilder::new::<Config>() /* ... */ .build().run();
}
```

There's two more crates created that are required at a minimum to make this possible.  These are required by `tauri` and not the application.

---

#### `tauri-config`:
This contains only the definitions of the `Config` as rust sees it. It also contains some defaults for the configuration, and a helper function to deserialize it from a path.  This was split out to prevent a circular dependency between `tauri/tauri-api` and the new `tauri-config-derive`.  This way, the derive can use and read the config at compile time, while still letting `tauri` use the same struct at runtime.

---

#### `tauri-config-derive`:
Created to read the configuration file relative to the application using tauri.  Using a proc macro from the application allows us to grab the actual cargo manifest directory of the application instead of the `tauri/tauri-api` dependencies.

---

While there are two new dependencies, there's a good opportunity to remove both the `tauri_includedir` + codegen packages and replace them with proc-macro.  Currently the proc-macro just calls the codegen itself but in a production quality crate it would just be a part of the proc-macro.

chippers:
I've rebased the webview PR because there was a decent number of conflicts and that PR is preparing to move to dev.

I've also started signing my commits

PR #845, opened 2020-07-17
chore: implement covector publish

jbolda:

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [x] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
This implements publishing through covector and removes the bulk of the Github Actions handling publishing. With this, we can also remove the `latest` branch.


jbolda:
Adding a package.json at the root partly for devDeps and partly for better support of update bots such as renovate: https://github.com/renovatebot/renovate/issues/4724

jbolda:
Outputs from actions can only be strings, hence that `jq` jazz. I think if we can move the tangle release to a package or a cli command, we may be able to remove all of that. (I think it may also end up running the second job on every merge, but not a great way to avoid it other than shifting tangle I believe.) @rajivshah3 do you have any thoughts on that?

rajivshah3:
Hmm, yeah the `jq` stuff isn't ideal. @obany do you think we can make a standalone NPM package for the Tangle Release and then just have the GH action be a wrapper around that?

.changes/config.json
```diff
@@ -3,18 +3,152 @@
...
+          "pipe": true
+        },
+        {
+          "command": "echo \"\\`\\`\\`\"",
+          "dryRunCommand": true,
+          "pipe": true
+        },
+        {
+          "command": "cargo publish --no-verify",
+          "dryRunCommand": "cargo publish --no-verify --dry-run",
```
rajivshah3, line 56: Should this be a boolean?

jbolda: It can either be a boolean (where `true` would run the command both in production mode and dry run mode) or a string and it will run that command instead in dry run mode.

rajivshah3: Ah ok

rajivshah3 (approved):

lucasfernog (approved):

nothingismagick (approved):

obany:
> Hmm, yeah the `jq` stuff isn't ideal. @obany do you think we can make a standalone NPM package for the Tangle Release and then just have the GH action be a wrapper around that?

@rajivshah3 So a lib that can be referenced by the current GH Action and with a CLI as well ?

jbolda:
@obany yea, something like that. That is how covector is setup. Mostly because it is rather complicated/fragile to get these actions setup and it helps being able to run something locally. Also the only way to "dynamically loop" through something is to spin up multiple machines which only increases the complexity (and run time).

obany:
Just released v0.7.0 of the action (which actually behaves the same as the old one from a GHA perspective)

But the core is completely rewritten in TypeScript and refactored so that it can also be executed as a CLI from it's npm package.

https://www.npmjs.com/package/@iota/gh-tangle-release (CLI instructions at bottom of README)

PR #850, opened 2020-07-18
refactor(tauri): support for building without environmental variables

chippers:
Both `embedded-server` and `no-server` fully work. Heres a checklist ill keep updated i guess:
* [x] Compressed embedded assets (~~brotli~~ gzip)
* [x] `no-server`
* [x] `embedded-server`
* [x] Pass test suite
* [x] Create additional tests for changed/new code (not really sure what to add for config testing)
* [x] `TAURI_INLINED_ASSETS` support
* [x] `TAURI_CONFIG` support
* [x] Refactor `AppConfig` -> `AppContext` to prevent repetition.  (i.e. `&application.config.config` -> `&application.ctx.config`)
* [x] handle public paths as in webpack instead of iteratively removing parents from assets

---

This is the replacement for #830.  That PR was more about seeing if it was possible, this PR is about actually removing core environmental build variables from tauri.

This branch/pull request should be tracking changes from dev to prevent having a massive conflict at the end.  This touches a few "core" files like `runner.rs` and has some tentacles in otherwise not related code due to moving `tauri-api::config` into `tauri-config` so conflicts are definitely going to happen.

As of the first commit, it can build a `no-server` application that has been run with `tauri dev` or `tauri build` to generate the `index.tauri.html`.
image

As a high level overview of additional changes planned in this branch:
1. All tests need to be updated
2. Support for `embedded-server`. (and every existing way of using tauri)
3. Handle additional env vars like `TAURI_CONFIG` and `TAURI_INLINED_ASSETS` which are currently ignored

See comments on #830 for a more specific overview if needed


chippers:
the last commit added support for `embedded-server`,
image

Because of some necessary changes to codegen, I lifted the code from the `tauri_includedir_codegen` repo into `tauri-macros/src/include_dir.rs` along with the rework.  Since the frontend to that dependency also didn't make much sense, I moved `tauri_includedir::Files` into `tauri-api/src/assets.rs` as `Assets`.  This commit doesn't add support for gzip since it was not used anyways.  It keeps the structure support for it however, and just needs to be actually implemented in the match in `include_dir.rs`.  I left a few `TODO` comments that should be handled by the time this PR would be ready

chippers:
Off the top of my head, some todos:
* replacement for the `get()` testing in `tauri-api/src/config.rs`
* finish gzip support in `tauri-macros/src/include_dir.rs`
* support `TAURI_CONFIG` environmental variable to override any actual config with the content sent in that env var
* make sure `TAURI_INLINED_ASSETS` actually works, since the code was ported
* probably add some tests around the `AppConfig` and some more tests wherever the config or environmental variable was previously used

tauri/src/endpoints/asset.rs
```diff
@@ -19,7 +20,7 @@ pub fn load(
       });
       let mut read_asset;
       loop {
-        read_asset = crate::assets::ASSETS.get(&format!(
+        read_asset = assets.get(&format!(
           "{}/{}",
           option_env!("TAURI_DIST_DIR")
             .expect("tauri apps should be built with the TAURI_DIST_DIR environment variable"),
```
chippers: *notes for myself:*
This needs to be adapted.  See `runner.rs` + `server.rs` for how they handle it now for embedded server assets.  It might be as simple as changing it to `format!("/{}", path.to_string_lossy())` since the keys are already normalized to have a `/` to start them and to use unix/url paths.  If the path passed to this can possibly have windows endings, we may need to move our `to_key` function from `tauri-marcos` to `tauri-api::assets` so that it can be used from proc-macros and this code.

chippers:
I've left the bundler and cli completely untouched, only touching `tauri`, `tauri-api`, and (new) `tauri-macros`.  There's still a few places env vars are used in the latter, such as endpoint assets handling and runner splashscreens.

*note for myself:*
https://github.com/tauri-apps/tauri/blob/d0c41e477e692cd80b95d218f30f4b1d1c4e17d3/tauri/src/app/runner.rs#L292

chippers:
Another relative large change.

Asset compression has been re-implemented in the proc-macro from the code that moved over from `tauri_includedir_codegen`.  This implementation happens to use Brotli because of higher compression ratios + faster/equal decompression speed and for the nice `Read` interfaces it implements allowing us to stream compressed brotli assets directly to the browser with `embedded-server`.

chippers:
exchanged out brotli for gzip with flate2.  removes the extra new brotli dependency and more importantly keeps the scope of this PR smaller.  I think using brotli for asset embedding compression is an interesting conversation to come back to.  replacing gzip with brotli is rather simple, as you can see in the 3693df9395204175a5e71914cfb671e2ae065a3d commit

nothingismagick:
So as this comes closer to something we can merge in, it would be good to know how we can validate this approach. Also, is this going to introduce any kind of breaking changes for consumers or devs?

chippers:
I agree that validation that this doesn't break anything is important.  This would be a breaking change for consumers, as the public API would change in a few ways.  The largest change by far is how entry into the `App` changes to include the struct derived with the new proc-macro.

Previously:
```rs
use tauri::AppBuilder;

fn main() {
  AppBuilder::new()
    .setup(|w, s| todo!())
    .invoke_handler(|w, arg| todo!())
    .build()
    .run();
}
```

With this change:
```rs
use tauri::{AppBuilder, FromTauriConfig};

#[derive(FromTauriConfig)]
struct Config;

fn main() {
  AppBuilder::<Config>::new()
    .setup(|w,s| todo!())
    .invoke_handler(|w, arg| todo!())
    .build()
    .unwrap() // new
    .run();
}
```

That `FromTauriConfig` derive also provides an attribute `#[tauri_config_path = "my/custom/path.json"]` to point to a config file relative to the app's cargo manifest.  An example of that attribute in use can be seen in the `tauri/examples/communication/src-tauri/src/main.rs` diff.  The `build()` now returning a `Result<App, Error>` is due to serde_json parsing the config file during the build.  This might be changed because technically any failures would have happened during the proc-macro and not given a chance to be built by the `AppBuilder`.

In short, the main needed change is the derive + passing the config as the generic to `AppBuilder`. Where that generic is passed in is up for change, but somewhere during the `AppBuilder`.

The other changes to the public api happen in a few public functions exposed in `tauri-api` and `tauri`.
1. `#[doc(hidden)] AsTauriConfig` trait in `tauri-api` which is what the proc-macro implements and the `AppBuilder` requires.
2. crate `tauri_includedir` is changed to `tauri-api::assets::Assets`.  The `tauri_includedir_codegen` counterpart is now part of `tauri-macros` under `include_dir.rs`.
3. Public API changes where `config::get()` was previously used.  These now require a `Config` argument or similar.  These functions previously took no arguments.
    * `tauri-api::cli::get_matches(config: &Config)`
    * `tauri-api::notification::Notification::new(identifier: impl Into<String>)`
4. Various non-public apis take a new argument similar to the previous changes
    * `tauri::endpoint::asset::load(/* ... */, ctx: &AppContext)`
    * `tauri::endpoint::notification::send(/* ... */, config: &Config)`
    * just about all of the setup functions in `tauri/src/app/runner.rs`
5. A new internally used struct to help `App` handle needing to pass the configs during runtime.  `AppContext` which holds a `Config`, `tauri-api::assets::Assets` (embedded files), and a static string of the `index.tauri.html`.

Like before, the exact api changes and ways of passing the data through are up for change.  The main important piece is passing _something_ so that those function have the information.  The `AppContext` is this helper in this implementation.

---

Otherwise, I've left the bundler and cli alone (besides the init template) to help prevent too much churn at once.  Since some of the parameters passed to the changes are just ignored (like `TAURI_DIR` and `TAURI_DIST_DIR`) there shouldn't be much conflict there.  Some of the parameters passed from them are still used like `TAURI_CONFIG` / `TAURI_INLINED_ASSETS`.

I still need to test more examples, so I'll be going through the examples repo one by one later this week and comparing `dev` to this branch.

chippers:
I've started to test examples 1 by 1 in tauri-apps/examples#398.  Before the next testing, I need to expand on the last commit (32a007fd996e796331c8fe4c2677fe8a6ec5f3ad) to fix paths on windows builds.  In short, we need to use the extended path version of windows paths to support paths >260 which came up during testing.

chippers:
All the examples I was able to run from tauri-apps/examples#398 either passed with the same behavior, or were broken in the same way.

This PR is ready for a preliminary review of the core changes, but the interfaces that are exposed should probably be discussed more and decided upon.

Here's what I would consider core changes:
* No more global config (previously a static value fetched from `tauri-api::config::get()`)
* An application context to put previously global application info into (`tauri::app::AppContext`)
* Functions that need some sort of configuration should take the `AppContext` (or a static value kept on it like assets)
* `tauri_includedir::Files` -> `tauri-api::assets::Assets`

Here are some of my thoughts on how to change the interface for better ergonomics:
1. Accept a parameter in `AppBuilder::new()` for the config instead of a generic parameter
    * This allows us to also expose a function-like macro so someone can do `AppBuilder::new(config!())` in addition to the derive.  Using the derive would be the same except you would need to create an instance of that struct, like `AppBuilder::new(Config{})`.  Similar to how the derive has a `#[tauri_config_path = "../my/config/path.json"]` option, the function-like one could allow a parameter for the custom path.
2. Expose a build.rs option along with the proc-macro for more options for users
    * Similar to crates like [phf](https://github.com/sfackler/rust-phf) we could have a `tauri-build` crate for using from `build.rs` and a `macro` feature on `tauri` to expose the proc-macro.  This would mean 2 more crates than the current code.  `tauri-build` for the build.rs dependency, and `tauri-shared` (or w/e name) for the code used by both `tauri-build` and `tauri-macros`.  There's a number of people against disk access happening during proc-macros and this would be a way to allow a choice in it.
    * We already have the template generating a `build.rs` with a dependency for windows exectutables so the same should be able to be done for `tauri-build`.  I haven't tested it yet but we should also be able to handle the current code to generate that windows executable metadata during the runtime of `tauri-build`
    * This would require change 1 so that we can pass the config to the `AppBuilder::new()`.  See [phf_codegen](https://github.com/sfackler/rust-phf) for examples on how they expose the name of the item exposed and the file it's found in so that the user could then do `include!(concat!("OUT_DIR"), "/tauri-config.rs")` and use the item generated (let's say TAURI_CONFIG) as `AppBuilder::new(TAURI_CONFIG)`

Let me know any thoughts on those 2 changes and sub-items.  As of right now I am leaning towards implementing both for the ergonomics so I would like to know any thoughts/changes/criticisms of those.

The review might be kind of messy since the interfaces might not be final, but if you read the core changes before the diff, you should be able to see the places where those items apply.

lucasfernog:
Hey @chippers i'm sorry it took so long but we're back. I'll make some time in the next few days to review this amazing work! This is something that we need :)

chippers:
Cool, I'll take a look at some of the updates since 8f06882fd100e73004d1d13c34d9a2d1536d8b66 and see if it needs any changes before updating from `dev` again. Feel free to mention me on the discord `@chip` if you want to discuss parts of it

edit: looks like it might need some minor changes beside a merge due to the new `__tauri.js` file that it needs.  I'll look into how this is generated and if it should be generated during the build/macro

chippers:
I've updated this branch to the `tauri-v0.11.1` release (14 days ago), but some of the more recent changes have many more conflicts and I will need to look into them in the next few days

lucasfernog:
I was going to do that for you :) I think I can help

lucasfernog:
We should wait for WRY to land on dev before fixing the conflict issues here.

lucasfernog:
We'll need to archive [tauri-includedir](https://github.com/tauri-apps/tauri-includedir) since this change kills it :P

lucasfernog (approved):

.changes/config-refactor.md
```diff
@@ -0,0 +1,21 @@
+---
+"tauri-utils": minor
+"tauri-api": minor
+"tauri": minor
+---
+
+The Tauri files are now read on the app space instead of the `tauri` create.
+Also, the `AppBuilder` `build` function now returns a Result.
+
```
nothingismagick: Please don't put usage in the changelog. This should be put in a README / in the inline documentation / on the site.

api/src/fs.ts
```diff
@@ -1,4 +1,4 @@
-import { promisified } from './tauri'
+import { promisified } from "./tauri";
```
nothingismagick: You are changing our style. Please revert `"` to `'` and `no-semis` as well as `no-trailing-commas`

nothingismagick (changes requested):
This is flipping awesome, but please don't change the TS style and watch out for overly expressive changelog stuff.

api/src/fs.ts
nothingismagick: That is why eslint is failing.
https://github.com/tauri-apps/tauri/pull/850/checks?check_run_id=1860478292

lucasfernog: i've seen this change happening by accident several times, maybe it's something with vscode/prettier

.changes/config-refactor.md
lucasfernog: oops :D

api/src/fs.ts
lucasfernog: ah, it was a missing prettier config on the api module :)

lucasfernog:
My bad :joy:

PR #874, opened 2020-07-21
Helper scripts to run cargo clippy + test for all features

chippers:
- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
Runs clippy + tests on all the crates in the workspace.  Crates with mutually exclusive features (`tauri`) are run separately, and for each mutually exclusive feature.  This was inspired by having #850 pass CI checks, but actually fail some of the tests due to features not being enabled (`tauri/cli` in this case).  I didn't even notice this until I typo'd `--all-targets` as `--all-features`.

We also may want to look at doing something similar in CI because some other features besides `cli` might be slipping through the cracks

### DIFF
--- a/.scripts/cargo-check.ps1
+++ b/.scripts/cargo-check.ps1
@@ -1,0 +1,15 @@
+$ErrorActionPreference = "Stop"
+Write-Output "Checking tauri crates"
+
+
+$commands=@("clippy","test")
+$features=@("no-server","embedded-server")
+foreach ($command in $commands) {
+  foreach ($feature in $features) {
+    Write-Output "[$command][$feature] checking tauri"
+    cargo $command --manifest-path tauri/Cargo.toml --all-targets --features "$feature,cli,all-api"
+  }
+
+  Write-Output "[$command] checking other crates"
+  cargo $command --workspace --exclude tauri --all-targets --all-features
+}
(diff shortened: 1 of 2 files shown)


lucasfernog:
Nice work. To use it on CI, it needs an arg to choose which command to run, because the test and lint workflows are separated.

lucasfernog (approved):

chippers:
The scripts are less useful for CI since the powershell one doesn't actually exit on the first error. Something to do with the executable run return code being different than the status code that powershell looks at.

We already do something similar in CI right now with matrix:
https://github.com/tauri-apps/tauri/blob/61fbf6837d33a38c2bf914cbd56d3284d18457d0/.github/workflows/core-lint-fmt.yml#L45
I think we can just change the matrix to the mutually exclusive features, and add all the other features into the command running it.  Maybe split the command into `tauri` + other crates like these commands do. I'll open an issue about it and take a look sometime later

chippers:
I removed the early return line from the ps1 script since it doesn't work to prevent it looking like that line is actually used.

lucasfernog:
I think we should use this same script on CI.

PR #930, opened 2020-08-07
fix #920: Create recipes 

zakpatterson:
**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [X] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [x] Refactor
- [X] Build-related changes
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes. Issue #___
- [X] No

**The PR fulfills these requirements:**

- [X] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #xxx[,#xxx]`, where "xxx" is the issue number)
- [X] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md).

If adding a **new feature**, the PR's description includes:
- [ ] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)

**Other information:**
fixes #920 

A recipe consists of:
	* Name
	* List of NPM dependencies required (dev and production)
	* Sub-commands needed
		For example, the react recipes run create-react-app to up the UI. Future recipes can do anything, they could pull down a sample UI, clone a repository, download a static HTML file, initialize another environment that compiles to JS/wasm etc.

The devPath and dist folders nomally configured by the init command are overridden by the recipe definition.

Example process for initializing a tauri project with the React + Typescript recipe:

2020-08-07-103915_606x175_scrot

Complete output (with some yarn output removed)
```
/home/projects/recipe-example
⟩ yarn tauri init
yarn run v1.22.4
$ /home/projects/recipe-example/node_modules/.bin/tauri init
[tauri]: running init
? What is your app name? Recipe Example
? What should the window title be? Recipe Example
? Would you like to add a UI recipe? React with Typescript
 dependency:manager Installing missing dependencies... +0ms
 dependency:cargo-commands "tauri-bundler" is already installed +14ms
 app:spawn [sync] Running "cargo generate-lockfile" +2ms

    Updating crates.io index
 dependency:crates "tauri" is already installed +889ms
 dependency:npm-packages "tauri" is already installed +4s
 recipe:install Installing dependencies for React with Typescript +0ms
 dependency:npm-packages Installing typescript... +4s
 app:spawn [sync] Running "yarn add typescript" +0ms

[ yarn install elided]

 dependency:npm-packages Installing @types/node... +10s
[...]

PR #937, opened 2020-08-12
chore: add prettier for js formatting

jbolda:

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [x] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**


jbolda:
So we had lint-staged configured incorrectly, and it wasn't running eslint. I fixed that. It seems we have linting config different in CI and/or prettier is fighting other lints. Will have to clean that up.

nklayman:
@jbolda shouldn't eslint run after prettier so that things like semis and quotes are proper?

jbolda:
I think we need to adjust the prettier settings so it straight up doesn't conflict, otherwise they will fight each other. If the settings are the same, it shouldn't really matter, but I think eslint makes sense second purely to catch something that prettier changed that we don't want (eg wrong setting likely being the culprit). I saw notes of tslint in the comments though; is eslint still what we want to go with?

nklayman:
Tslint is deprecated and eslint is the official replacement, so we should stick with that.

I've found it difficult to make the prettier config match eslint, I've always just left prettier as default and run eslint afterwards which works pretty perfectly for me. Since eslint will overwrite any problems there isn't really a point trying to fix the prettier config (although it may improve performance if that ever becomes an issue).

jbolda:
RE: tslint, thanks @nklayman 

I think I just about have the config fixed so prettier won't step on eslint, but a couple settings I am not quite certain on. Our files aren't consistent (which is good that we are doing this).

settings to clear up:
- are we using semicolons?
- seems like no trailing commas?
- using single quotes for strings?

Laegel:
@jbolda

- are we using semicolons? **never seen any semi in Tauri JS code, so I guess not**
- seems like no trailing commas? **not sure about this one, likely not**
- using single quotes for strings? **single quotes**

jbolda:
A couple examples:
- this uses semicolons: https://github.com/tauri-apps/tauri/blob/b9aebfaf14ca98bfdad428b0f155099545954477/cli/tauri.js/api-src/bundle.ts
- the only case I saw with a trailing comma (which is why I turned it off): https://github.com/tauri-apps/tauri/blob/b9aebfaf14ca98bfdad428b0f155099545954477/cli/tauri.js/api-src/fs.ts

jbolda:
Removed the semicolons, that does seem to give a diff with less changes. I think I happened to only be looking at the couple files that had semicolons.

**EDIT** and @Laegel pointed out that it seems I was looking at an old diff, and that might have happened from an earlier commit. I think errors from lint steps made some of those diffs fun.

jbolda:
It is interesting that I am getting some eslint errors locally but not in the eslint workflow:

```shell
❯ yarn lint-fix
yarn run v1.22.4
$ eslint --fix --ext ts ./src/**/*.ts ./api-src/**/*.ts

C:\Users\Jacob\Documents\dev\tauri\tauri\cli\tauri.js\src\api\dependency-manager\cargo
-commands.ts
  27:15  error  Unsafe assignment of an any value             @typescript-eslint/no-un
safe-assignment
  35:13  error  Unsafe member access .answer on an any value  @typescript-eslint/no-un
safe-member-access

C:\Users\Jacob\Documents\dev\tauri\tauri\cli\tauri.js\src\api\dependency-manager\cargo
-crates.ts
  72:15  error  Unsafe assignment of an any value             @typescript-eslint/no-un
safe-assignment
  80:13  error  Unsafe member access .answer on an any value  @typescript-eslint/no-un
safe-member-access

C:\Users\Jacob\Documents\dev\tauri\tauri\cli\tauri.js\src\api\dependency-manager\npm-p
ackages.ts
  34:17  error  Unsafe assignment of an any value             @typescript-eslint/no-un
safe-assignment
  42:15  error  Unsafe member access .answer on an any value  @typescript-eslint/no-un
safe-member-access

C:\Users\Jacob\Documents\dev\tauri\tauri\cli\tauri.js\src\api\dependency-manager\rust.
ts
  12:16  error  'download' is defined but never used  @typescript-eslint/no-unused-var
s

C:\Users\Jacob\Documents\dev\tauri\tauri\cli\tauri.js\src\runner.ts
  361:67  error  Prefer using nullish coalescing operator (`??`) instead of a logical
or (`||`), as it is a safer operator  @typescript-eslint/prefer-nullish-coalescing
  432:59  error  Prefer using nullish coalescing operator (`??`) instead of a logical
or (`||`), as it is a safer operator  @typescript-eslint/prefer-nullish-coalescing

✖ 9 problems (9 errors, 0 warnings)
```

jbolda:
Merged in #930 to format that as well. I still can't get eslint to fail a build though.

jbolda:
😍

rajivshah3 (approved):

PR #1006, opened 2020-09-01
feat: Port path api to js

### LINKED ISSUE #929: [Feature][API][JS] Provide a Way to Get the Actual Paths to the Data/Config/etc. Dirs
It's been an issue of mine that while the Tauri JavaScript API allows you to read file with paths relative to the `LocalData` dir for instance, there is no direct way for me to get the actual path represented by `LocalDir`. The only indirect way that I have noticed to do that is to do something like `readDir('/', {dir: Dir.LocalData})` and then grab the path for one of the files as a reference, because it contains the full path to the contained files.

I would like way to get the full paths associated to the directories in the `Dir` enum.

I suppose the reason this may not have been done was because of the nuance of having to join the paths?

**Additional context**
I wanted to use this because I have an app that stores data in `[LocalDataDir]/appname/` and the app I'm making is a way to conveniently edit files in the dir, so I wanted to make a file dialog that has the default path set to `[localDataDir]/appname`, but because the `defaultPath` option doesn't support paths relative to the `Dir` enum, I couldn't do it. Even if the default dir option could be relative to the `Dir` enum, I would still want to be able to get the full path for the different dirs for other reasons.

amrbashir:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
Closes #929


jbolda:
Would you please give some context behind the purpose of this PR and if it's related to prior discussion or an issue?

amrbashir:
The purpose it to make accessing these directories available from js side. I think it can ease the development for tauri users a bit, since they won't have to write this by themselves. 

No, this is not related to prior discussion but I think there is an issue open #929 .

tauri/src/endpoints.rs
```diff
@@ -138,6 +139,125 @@ pub(crate) fn handle(webview: &mut Webview<'_>, arg: &str) -> crate::Result<()>
           #[cfg(not(rename_file))]
           allowlist_error(webview, error, "renameFile");
         }
+        AppDir { callback, error } => {
```
lucasfernog: Maybe it would be easier to change all these command types to a single one with a "type" argument or something

tauri/Cargo.toml
```diff
@@ -63,6 +63,25 @@ create-dir = [ ]
 remove-dir = [ ]
 remove-file = [ ]
 rename-file = [ ]
+app-dir = [ ]
```
lucasfernog: I don't think we need a feature flag for each method; probably just one for all of them since it's the same dependency related.

lucasfernog (changes requested):

tauri/Cargo.toml
amrbashir: Yeah that makes more sense.

tauri/src/endpoints.rs
AmrBinBashir: Okay, I will make it into a single endpoint and inside `path.rs` I will make all the methods private and only expose one method to handle matching the argument to the related private method.

And I will make another endpoint for `ResolvePath` method since it will receive different arguments.

amrbashir: Okay, I will make it into a single endpoint.

```rust
pub fn get_dir ( dir: BaseDirectory ) {
   match dir {
      BaseDirectory::Cache => cache_dir()
   }
}
```

And I will make another endpoint for `ResolvePath` method since it will receive different arguments.

amrbashir:
I pushed new changes but I am not sure if I named the endpoints and methods correctly. If you have some better names or other comments, let me know.

lucasfernog (approved):

cli/tauri.js/api-src/path.ts
```diff
@@ -8,7 +8,8 @@ import { BaseDirectory } from './fs'
  */
 async function appDir(): Promise<string> {
   return await promisified<string>({
-    cmd: 'getDirectory',
+    cmd: 'resolvePath',
+    path: '.',
```
amrbashir, line 12: @lucasfernog this will add a "dot" to the end of the path, I believe an empty string is better.

amrbashir: One more thing, if a base directory doesn't exist, the user will get an error like this `Path Error:unable to determine base dir path`. In my opinion this will confuse the user as he doesn't know the underlying api is just resolving a path using a base dir, instead it would be better to just return null like tauri's Path api in rust.

PR #1059, opened 2020-10-18
fix(readme): update electron comparison table, fix #1048

### LINKED ISSUE #1048: [docs] Changes to Tauri-Electron comparison list
Hey guys, I just discovered Tauri, and I already love it, great job!
However I think that the feature comparison list needs some work. It needs more of useful information and less of weird information on it.

Now, let's break down some of the metrics on the comparison list:

## 1. Can Render PDF (Tauri: yes; Electron: no)
What does that even mean? Opening PDFs? Displaying PDFs? Actually rendering PDFs using Nvidia RTX ray-tracing? 

As far as I know neither Tauri nor Electron were created specifically for making "PDF reader" apps. So why is it a main feature?
If by "rendering" you mean "displaying PDFs", then the comparison is false, it's actually very easy to do in Electron:

### Electron 10.1.3: `mainProcess.js`
```
function openPDF () {
  let pdfWindow = new BrowserWindow({
    width: 1280,
    height: 720,
    webPreferences: {
      plugins: true
    }
  })

  pdfWindow.loadURL('file://C:/test/test.pdf')
}
```

## 2. Multithreading (Tauri: yes; Electron: no)
Multithreading is an ability to run multiple blocking tasks concurrently. And we all know that Electron has multiple ways of achieving this:

- Web worker
- Node worker_threads
- Spawning separate node processes
- Executing binary files via system terminal

Which, again, means that the comparison is factually false. I would argue it's even easier to utilize all the threads of your CPU with Electron because creating a worker.js and a message event listener is much easier than creating a proper multithreading system on Rust.

## 3. Splashscreen (Tauri: yes; Electron: yes)
This comparison doesn't even make any sense. The Nokia 3310 had a splashscreen animation. Splashscreen is just a loading screen, which can be done on anything that runs code. You can even do it with pure HTML without any scripting.

Am I missing something or does this comparison convey absolutely no meaningful information? It's like saying that both Tauri and Electron require a computer to run and can both perform computations. Don't you say! Incredible, that! 😄

## 4. Speed

I was quite impressed by the ~0.5 sec launch speed of my test Tauri app.

Guys, why would you compare the ability to "Render PDF" and "To have a Splashscreen" but not speed - one of the most important metrics to devs and users? The app launching speed, the computational speed of different tasks (Rust vs V8), the compilation speed, etc.

## 5.  Interface Service Provider (Tauri: Varies; Electron: Chromium)

This comparison, unfortunately, doesn't convey any useful information either. 

How does it all work? How is it possible that we can use Node's npm packages on Rust? Can I just install `node` bin module from npm and run ALL Node's modules on Rust? What kind of black magic is this?

We know that Electron supports image, audio, video decoding, it gives us the browser `window` and `document` objects that have different listeners (click, scroll, drag, etc), it supports the majority of HTML / CSS / JS features, and so on. But it's not clear how Tauri handles all of this. Does everything just works the same? Or do we have to create our own low-level listeners in Rust, or what?

What HTML / CSS / JS features do we lose with Tauri? 

## 6. Supported file types

I can see that Tauri can decode and display some types of images, but why can't it decode and display videos at all? What file types will be supported in the future? 

Why is "supported file types" metric not on the comparison list? This sort of information would actually be useful to have on the comparison list.

--- 

Please, don't take this criticism in a wrong way, I'm not spending my time writing all this just to hate on Tauri, on the contrary, I'm doing this to help and make it a tiny bit better. I'm really excited there's a potential proper competitor to Electron. Thank you for making it, guys!

nklayman:
Changes suggested in #1048

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: Docs

- [ ] Yes. Issue #___
- [x] No

**Other information:**


nothingismagick (approved):

nothingismagick:
@nklayman - care to finish this off and mark as ready?

nklayman:
Since wry support is moving along so quickly, I'm gonna wait for that to add the performance benchmarks and finish this.

nklayman:
@nothingismagick I updated the benchmarks with the latest versions of tauri/electron. I used `mprof` to track memory usage, which also tracks child processes, which is why I think the numbers are so different from the original ones.

README.md
```diff
@@ -95,32 +95,29 @@ If you are interested in making a tauri-app, please visit the [documentation web
...
 | Detail | Tauri | Electron |
 |--|--------|----------|
-| Binary Size MacOS | 0.6 MB (1) | 47.7 MB |
-| Memory Consumption MacOS | 13 MB | 34.1 MB |
+| Installer Size Linux | 3.1 MB | 52.1 MB |
+| Memory Consumption Linux | 180 MB | 462 MB |
+| Launch Time Linux | 0.39s | .80s |
 | Interface Service Provider | Varies | Chromium |
 | Backend Binding | Rust | Node.js (ECMAScript) |
 | Underlying Engine | C/C++ | V8 (C/C++) |
```
nothingismagick, line 103:
```suggestion
| Underlying Engine | Now: C/C++ / Soon: Rust | V8 (C/C++) |
```

README.md
```diff
@@ -95,32 +95,29 @@ If you are interested in making a tauri-app, please visit the [documentation web
 
 | Detail | Tauri | Electron |
 |--|--------|----------|
-| Binary Size MacOS | 0.6 MB (1) | 47.7 MB |
-| Memory Consumption MacOS | 13 MB | 34.1 MB |
+| Installer Size Linux | 3.1 MB | 52.1 MB |
```
nothingismagick, line 98: Did you use this when making your final binary?
```
[profile.release]
panic = "abort"
codegen-units = 1
lto = true
incremental = false
opt-level = "z"
```

nothingismagick: I used upx when I calculated the MacOS binary size as well...
https://tauri.studio/en/docs/guides/bundler/anti-bloat#upx

nothingismagick (approved):
See my comments.

README.md
nklayman: I used the cargo config but not upx as that slows down the launch time (right?) and it's not fair to use benchmarks from different apps.

README.md
nklayman: We're going to have to change a lot of things in the readme when beta lands so I vote we just wait for that.

README.md
nklayman: It also can't be used on a `.deb` installer.

nklayman:
@nothingismagick this is long overdue so I'm going to merge this now, if you want to run the benchmarks on another platform to to get the lower installer size without hurting the startup time then that can be a separate PR.

PR #1106, opened 2020-12-05
feat(cli): add create-tauri-app

sanket143:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

Discussions:
- [Discord thread](https://discord.com/channels/616186924390023171/731495028677148753/777634687287820289)
- [GitHub Discussions](https://github.com/tauri-apps/tauri/discussions/1032#discussioncomment-99534)

Comments with `--NOTES--` header is just for references. All comments with `--NOTES--` will get cleared once it is ready.


sanket143:
@nklayman Though it's still a little rough and hardcoded, can you tell me whether this is the right way to do it?

sanket143:
How I have thought of doing this is to have a folder `recipes` for init scripts of all recipes which will setup required files and directories basically the project structure. Once that's successful, it can install dependencies and stuff depending on whether it's on Deno, Node or Rust. For now, I have simply called `installDependencies()` in `recipes/index.js` because currently Tauri only supports rust for the backend but it can also be modularized.

lucasfernog:
Wouldn't that bloat the codebase? what do you think about having a separate github repo for each recipe?

nklayman:
I'm super late but I think the recipes should all be in the same repo. There isn't much code here anyways, and we don't want to have our repo list full of `tauri-template-x`. It would also make releasing and issue handling more difficult.

jbolda (commented):
Thanks for putting this together. We did want to shift over the `tauri create` command to this as well, but I think I can handle that. We will have a bit of overlap between the things, but I think we can pretty cleanly merge them. I will get that setup and pushed to a feature branch, and then I can get this merged into that (so leaving this open / untouched for the moment).

jbolda:
This did diverge a bit from the `tauri create` command, but I think it will actually still work out fine. We were also talking about pulling in degit (as that is what svelte expects as well).

I am imagining two main logical paths, and one of those splits into two as well. The first split would either be a shell command such as running `create-react-app` followed by a `tauri init` and a bit of wiring. The other path would be scaffolding / templating (which works well for vanilla js here or svelte for example). We can wire up degit to help us pull in community created templates, and then we would choose to either run it through this scaffolding that you have set up or just a copy/paste.

nklayman:
I think both need to be supported because there isn't a cli for vanilla js apps, but you can't use a template for vue-cli. Some frameworks require a template and some a cli command.

jbolda:
Ok, so it sounds like we can use the "shell out" approach for vue then?

nklayman:
Yeah, just running `vue create my-app` should work fine (as long as you can pipe through it's prompts, which should be solved by setting `stdio` to `pipe`).

jbolda:
I rebased this onto the feature branch. I undid a bit of a wiring with the expectation to merge the code paths and wire it all back up in a follow up PR.

Thanks again @sanket143 !

PR #1116, opened 2020-12-13
chore(audit): ignore tiny-http request smuggling attack vuln

lucasfernog:
I believe it doesn't affect us. Can you confirm @nothingismagick ?

https://github.com/tiny-http/tiny-http/issues/173


lucasfernog:
We should keep an eye on https://github.com/tiny-http/tiny-http/pull/190

PR #1166, opened 2021-01-30
fix(tauri/asset): escape octal sequences in css

nklayman:

- [x] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

Octal escape sequences, sometimes used for special unicode characters, are not allowed in template literals. Since Tauri's asset loader for no-server injects css into a template literal, it will throw an error if the css contains an octal sequence. This PR will replace `\` with `\\` in css assets which will allow the octal sequence to pass through to the css without triggering a JS error.

### DIFF
--- a/tauri/src/endpoints/asset.rs
+++ b/tauri/src/endpoints/asset.rs
@@ -76,7 +76,8 @@ pub fn load(
                   document.getElementsByTagName("head")[0].appendChild(css);
                 }})(`{css}`)
               "#,
-              css = asset_str
+              // Escape octal sequences, which aren't allowed in template literals
+              css = asset_str.replace("\\", "\\\\").as_str()
             ));
           } else {
             webview_ref.eval(asset_str);
--- a/.changes/css-inliner-octal.md
+++ b/.changes/css-inliner-octal.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch
+---
+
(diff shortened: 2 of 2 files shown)


tauri/src/endpoints/asset.rs
```diff
@@ -76,7 +76,8 @@ pub fn load(
                   document.getElementsByTagName("head")[0].appendChild(css);
                 }})(`{css}`)
               "#,
-              css = asset_str
+              // Escape octal sequences, which aren't allowed in template literals
+              css = asset_str.replace("\\", "\\\\").as_str()
```
nothingismagick, line 80: what if there are already e.g. `\\\\`? Wouldn't that make `\\\\\\\\`? or is this a non-concern?

nothingismagick (changes requested):
Please don't forget to add a .changes file to this PR.

tauri/src/endpoints/asset.rs
nklayman: That is desired behavior. Since the content is being injected into a template literal, the extra slashes will be reverted back to original when the content is actually injected into the dom.

nklayman:
Okay I added it :+1:.

lucasfernog (approved):

PR #1173, opened 2021-01-31
fix(tauri/app): make splashscreen work, fixes #991

### LINKED ISSUE #991: [docs] Splashscreen example gives compilation errors
I've tried to get the [splashscreen example](https://tauri.studio/en/docs/usage/recipes/visual/splashscreen) to work, but when I run `tauri dev`, I get:

```
[tauri]: running dev
 app:spawn Running "cargo run" +0ms

   Compiling pw-all v0.1.0 (/Users/emile/github/pwall/src-tauri)
error[E0599]: no method named `handle` found for mutable reference `&mut webview_official::webview::Webview<'_>` in the current scope
  --> src/main.rs:12:28
   |
12 |       let handle = webview.handle();
   |                            ^^^^^^ method not found in `&mut webview_official::webview::Webview<'_>`

error[E0308]: mismatched types
  --> src/main.rs:14:33
   |
14 |       tauri::close_splashscreen(&handle);
   |                                 ^^^^^^^ types differ in mutability
   |
   = note: expected mutable reference `&mut webview_official::webview::Webview<'_>`
                      found reference `&_`

error: aborting due to 2 previous errors

Some errors have detailed explanations: E0308, E0599.
For more information about an error, try `rustc --explain E0308`.
error: could not compile `pw-all`.
```

Is there a simple ready-to-run example app somewhere?

nklayman:

- [x] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
Docs still need to be updated to match.


tauri/src/app/runner.rs
```diff
@@ -293,14 +287,6 @@ fn build_webview(
   // TODO waiting for webview window API
   // webview.set_fullscreen(fullscreen);
 
-  if has_splashscreen {
```
lucasfernog, line 303: if I recall correctly, you can't remove this code, it's needed to use the tauri API on the splashscreen HTML page

nklayman: I don't think a splash screen would ever need access to the Tauri api. However, that would be necessary if we wanted it to be available just in case. The old code expected a file that no longer is generated, so we'd have to refactor it somehow.

lucasfernog:
I'll refactor the tauri entry point so that it uses the `webview.init` API, which guarantees that the code is run before page load. That would also fix that problem we had where a dev writing an app that's not SPA would lose the tauri API on page change.

tauri/src/app/runner.rs
lucasfernog: I'm on it: https://github.com/tauri-apps/tauri/pull/1186

lucasfernog (approved):

lucasfernog:
@nklayman can you confirm my merge here didn't break your logic? :D

lucasfernog:
I'll close this PR because this API will be removed when we add the multi window feature - coming soon :)

PR #1217, opened 2021-02-11
feat(core): add mult-window support

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

This PR adds multi-window support to Tauri. Check out the new example:

```bash
$ cd tauri
$ cargo run --example multiwindow
```

For now the windows can be configured on `tauri.conf.json` but a Rust API will be provided soon on the AppBuilder struct.

The window API will be improved on a subsequent PR, with the event system refactored to simplify multi-window usage and APIs to manipulate the windows.


tauri-utils/src/config.rs
```diff
@@ -6,16 +6,75 @@ use serde_json::Value as JsonValue;
...
+#[derive(PartialEq, Deserialize, Debug, Clone)]
 pub struct WindowConfig {
+  #[serde(default = "default_window_label")]
+  /// The window identifier.
+  pub label: String,
+  /// The window webview URL.
+  #[serde(default)]
+  pub url: WindowUrl,
+  /// The horizontal position of the window's top left corner
+  pub x: Option<f64>,
```
nklayman, line 61: I think we should accept `usize` for these and convert them to i64 on the wry side. I can't think of a scenario where you'd actually pass a non integer so it makes more sense to accept an integer imo, it will save people from having to add the `.0`, which could be confusing for those unfamiliar with Rust and statically typed languages.

tauri/examples/communication/src-tauri/src/main.rs
```diff
@@ -18,19 +18,21 @@ struct Context;
 
 fn main() {
   tauri::AppBuilder::<tauri::flavors::Wry, Context>::new()
-    .setup(|dispatcher, _source| async move {
-      let mut dispatcher = dispatcher.clone();
-      tauri::event::listen(String::from("js-event"), move |msg| {
+    .setup(|webview_manager| async move {
+      let current_webview = webview_manager.current_webview().unwrap().clone();
+      let current_webview_ = current_webview.clone();
```
nklayman, line 23: Is there any way to get around this quite ugly clone mess? If a function like `listen` always requires a second clone, could it clone automatically?

lucasfernog: No way around it, Rust is verbose :/

tauri-utils/src/config.rs
lucasfernog: I was just following the `wry` interface. We should follow the same convention.
This is the `tauri.conf.json` stuff and the deserialization would also work with integers. The only real issue regarding this would be the Rust API to create windows, but I can also change that to accept both `usize` and `f64`.

nklayman: I think that Wry itself should take `usize`s, but if Tauri accepts either that's ok too.

PR #1218, opened 2021-02-11
fix(tauri/examples): fix build errors

nklayman:

- [x] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

### DIFF
--- a/tauri/examples/communication/src-tauri/src/main.rs
+++ b/tauri/examples/communication/src-tauri/src/main.rs
@@ -13,6 +13,7 @@ struct Reply {
 }
 
 #[derive(tauri::FromTauriContext)]
+#[config_path = "examples/communication/src-tauri/tauri.conf.json"]
 struct Context;
 
 fn main() {
--- a/tauri/examples/communication/src-tauri/src/main.rs
+++ b/tauri/examples/communication/src-tauri/src/main.rs
@@ -13,7 +13,6 @@ struct Reply {
 }
 
 #[derive(tauri::FromTauriContext)]
-#[config_path = "examples/communication/src-tauri/tauri.conf.json"]
 struct Context;
 
 fn main() {
--- a/tauri/examples/api/public/index.html
+++ b/tauri/examples/api/public/index.html
@@ -9,7 +9,7 @@
 
   <link rel='icon' type='image/png' href='favicon.png'>
   <link rel='stylesheet' href='global.css'>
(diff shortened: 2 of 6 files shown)


tauri/examples/communication/src-tauri/src/main.rs
```diff
@@ -13,7 +13,6 @@ struct Reply {
 }
 
 #[derive(tauri::FromTauriContext)]
-#[config_path = "examples/communication/src-tauri/tauri.conf.json"]
```
lucasfernog: can you keep this line of code? this example is a `Cargo` example, so it should be accessible with `$ cargo run --example communication` on the `tauri` folder.
The API example doesn't use this syntax because it needs to be compiled first, and we're not commiting all the outputs.

nklayman: Ohh. Seems like there should be a way for the FromTauriContext macro to detect the proper `src-tauri` directory somehow.

lucasfernog: maybe it should have a fallback if the config_path isn't found, idk, we can discuss about it later

@chippers what do you think? (we're talking about allowing this example to be run both as a Cargo example and as a standalone project).

lucasfernog (approved):

tauri/examples/communication/src-tauri/src/main.rs
chippers: The problem is that what the proc-macro is able to see is crate manifest path of the calling crate. During an example this will be set to the manifest of the crate it is an example of, and standalone as its own manifest.  The path in `config_path` is relative to that manifest. I'm not sure how it would be able to detect the correct config file to use when used from both cargo projects.

This probably should be handled by the application code and not from the macro.  In this case, there might be a way to jerry rig where the current path is by using something like:
```rust
concat!(
    env!("CARGO_MANIFEST_DIR"),
    "/",
    file!(),
    "/../tauri.config.json"
);
```
as the value to the `config_path`. The magic being that `file!()` is a path to the source file relative to the manifest.

This might need to come with a change that absolute paths should not try to resolve relative to the manifest (if not already the case, I need to check).

PR #1221, opened 2021-02-12
fix(bundler): remove unused code

nklayman:

- [x] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

### DIFF
--- a/cli/tauri-bundler/src/bundle.rs
+++ b/cli/tauri-bundler/src/bundle.rs
@@ -65,7 +65,7 @@ pub fn bundle_project(settings: Settings) -> crate::Result<Vec<PathBuf>> {
 
   #[cfg(windows)]
   {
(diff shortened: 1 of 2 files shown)


lucasfernog:
The bundler should still work on non-Tauri projects, so I think we should close this. What do you think @nklayman ?

nklayman:
Well it's giving us check annotations that I'd like to get rid of. Plus, the code doesn't actually use the value now. We could add support for using it, but I don't think we should leave it there and ignore it. Maybe prefix it with an underscore so it doesn't give check issues and add a TODO to use it?

lucasfernog:
yeah I missed it, but it should be using `.is_some()` instead. I'll add this change.

PR #1225, opened 2021-02-13
feat(core): window API

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

This PR extends the window manipulation API (currently we only have `setTitle`), now every option configurable on `tauri.conf.json > tauri > windows` can be changed at runtime (position, size, min size, max size, fullscreen, maximized state, minimized state, transparent, always on top, decorations, resizable flag, visibility flag, icon).


parker-codes:
So exciting!

nklayman:
On the example (and probably on any app with these params) the `maxHeight` will do nothing if you do not set `maxWidth` too (same for `minHeight` as well), and vice versa. Also, putting a debounce on the `maxHeight/Width` might be a good idea, because now the window will shrink super small as you type in the number since it will set a single-digit value first.

PR #1232, opened 2021-02-14
feat(core): globalShortcut API

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

This PR introduces a global shortcut API. Run the API example to test it out :)

```javascript
import { registerShortcut, unregisterShortcut } from '@tauri-apps/api/globalShortcut'

registerShortcut('ctrl + shift + a', () => alert('tauri!')).catch(console.error)
registerShortcut('super + T', () => alert('it works'))

unregisterShortcut('super + T').catch(console.error)
```

* Tested on Linux, macOS, Windows
* This only works because of the [wry](https://github.com/tauri-apps/wry) migration :)


ken0x0a:
\[Question (macOS)]

Are shortcuts, registered using this PR code, appears as MenuItem like this?

lucasfernog:
We don't have menu support yet but I think we'd need to do it manually.

ken0x0a:
It seems this is something related to do that for macOS.
https://github.com/ryanmcgrath/cacao/blob/master/src/macos/menu/item.rs

lucasfernog:
Cool, we'll handle that when WRY supports menus.

.changes/shortcut-api.md
```diff
@@ -0,0 +1,7 @@
+---
+"api": patch
+"tauri-api": patch
+"tauri": patch
```
nklayman: Wouldn't this be a minor release since it adds a feature?

api/src/globalShortcut.ts
```diff
@@ -0,0 +1,39 @@
+import { promisified, transformCallback } from './tauri'
+
+/**
+ * register a global shortcut
+ * @param shortcut shortcut definition, modifiers and key separated by "+" e.g. Control+Q
```
nklayman: With the example you provide here, I think it would be better to use CommandOrControl so that it works the same on Mac. See [this](https://www.electronjs.org/docs/api/accelerator#platform-notice) for more details.

api/src/globalShortcut.ts
```diff
@@ -0,0 +1,39 @@
...
+      cmd: 'register',
+      shortcut,
+      handler: transformCallback(handler)
+    }
+  })
+}
+
+/**
+ * unregister a global shortcut
+ * @param shortcut shortcut definition, modifiers and key separated by "+" e.g. Control+Q
```
nklayman: Same as above

tauri-api/Cargo.toml
```diff
@@ -35,6 +35,7 @@ tauri-utils = { version = "0.5", path = "../tauri-utils" }
 clap = { version = "=3.0.0-beta.2", optional = true }
 notify-rust = { version = "4.2.2", optional = true }
 once_cell = "1.5.2"
+tauri-hotkey = { git = "https://github.com/tauri-apps/tauri-hotkey-rs", branch = "dev", optional = true }
```
nklayman, line 38: Shouldn't this get published to crates.io before we release? Do we want to do it before merge so we don't forget?

tauri/src/endpoints/global_shortcut.rs
```diff
@@ -0,0 +1,86 @@
...
+  /// Unregister a global shortcut.
+  Unregister {
+    shortcut: String,
+    callback: String,
+    error: String,
+  },
+}
+
+impl Cmd {
+  #[allow(unused_variables)]
```
nklayman: Why is this here?

nklayman (changes requested):
Awesome feature to have! Just a few questions/suggestions.

tauri-api/Cargo.toml
lucasfernog: we will publish it, but it's still in development so i'll just leave it as a git dep here. (yeah let's hope we don't forget it :joy: )

tauri/src/endpoints/global_shortcut.rs
lucasfernog: hmm I rewrote the method and it isn't needed anymore; thanks for finding that

api/src/globalShortcut.ts
lucasfernog: CommandOrControl isn't supported yet, will be added later

tauri-api/Cargo.toml
nklayman: We might want to consider making a list or adding it to that hackmd at least.

api/src/globalShortcut.ts
nklayman: Oh ok. Maybe we could use `alt` instead since that is cross-platform and then we don't have to remember to change this later when we do add CommandOrControl.

lucasfernog: Ctrl works just fine on macOS though

tauri-api/Cargo.toml
lucasfernog: yeah we should manage our TODOs better

api/src/globalShortcut.ts
nklayman: It does work, but according to macworld.com:
> On a Mac you use the Command key where on a PC you would use Control (or Ctrl)

So event though it is a key, it's not what a user would expect it to be.

lucasfernog: there you go

nklayman: Sorry for nitpicking, but I could totally see someone copying that example and then opening an issue complaining that "Tauri keyboard shortcuts are broken reeeeee" :joy:.

PR #1237, opened 2021-02-15
refactor(core): improve HTTP API, closes #1098

### LINKED ISSUE #1098: Improved http client
I'm using v0.12.0 for reference.

The current http client only returns the body from the response. It needs to return *all* of the response meta data as well. Status, headers etc.

I would just return a response object from the rust side and allow people to handle the response however they need in javascript. It would probably make sense to try to emulate the api of popular http clients in the js world, like axios. Perhaps tauri could even supply a backend for one of those clients.

browser's fetch api et al.

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**


lucasfernog:
@nklayman I'd love to hear your inputs on the API interface (file [http.ts](https://github.com/tauri-apps/tauri/pull/1237/files#diff-099376f00b23847520b0ebb8d25dc0d4e5e7c2294999e617e3ff6d32ffca8da0))

nklayman:
I think the `createClient`/`dropClient` adds some unnecessary complexity and could scare people into just using the webview fetch. Why is that necessary? If that is required, I think we should at least provide a `request` function that automatically creates and then drops the client.

lucasfernog:
> I think the `createClient`/`dropClient` adds some unnecessary complexity and could scare people into just using the webview fetch. Why is that necessary? If that is required, I think we should at least provide a `request` function that automatically creates and then drops the client.

The Client adds some optimizations but yeah we should add a global default client.

PR #1239, opened 2021-02-16
refactor(core): all API are now promise based

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

The Tauri API is now Promise-based everywhere.

`promisified` was removed (use `invoke` exclusively now)
`tauri::execute_promise` helper was removed (the return value is used to resolve/reject the promise instead)
`tauri::execute_promise_sync` helper was removed (the command message has a `mainThread: bool` field that allows your command to run on the main thread instead of a green thread.


api/src/event.ts
```diff
@@ -13,13 +13,13 @@ export type EventCallback<T> = (event: Event<T>) => void
...
-function listen<T>(
+async function listen<T>(
   event: string,
   handler: EventCallback<T>,
   once = false
-): void {
-  invoke({
-    module: 'Event',
+): Promise<void> {
+  await await invoke({
```
nklayman, line 21: Why are there two `await`s here? You could just return the `invoke()` call instead, that way it will return any value that `invoke()` returns, and I think that will help with error handling too.

api/src/event.ts
```diff
@@ -35,9 +35,9 @@ function listen<T>(
  * @param event the event name
  * @param [payload] the event payload
  */
-function emit(event: string, payload?: string): void {
-  invoke({
-    module: 'Event',
+async function emit(event: string, payload?: string): Promise<void> {
+  await invoke({
```
nklayman, line 39: Again, you can just return the `invoke()` call.

api/src/fs.ts
```diff
@@ -61,8 +61,8 @@ async function readTextFile(
   filePath: string,
   options: FsOptions = {}
 ): Promise<string> {
-  return await promisified<string>({
-    module: 'Fs',
+  return await invoke<string>({
```
nklayman, line 64: You don't need to await it if you are returning it.

lucasfernog: the await is added by the stupid linter

api/src/event.ts
lucasfernog: linter bug probably lol

api/src/fs.ts
nklayman: Jeez is that eslint that's doing all that stuff?

lucasfernog: yeah, can you see what's going on? I hope that's not the default config, I hope we messed up :joy:

nklayman: Okay so it seems like there is a trade-off if you just `return` instead of `return await`. Just `return` has better performance because it creates less promises, but `return await` will show that function in the stack trace. However, due to some rollup stuff (I assume), the stack trace is the same no matter what. I opened #1245 to fix this.

PR #1241, opened 2021-02-16
feat(core): add window event system

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No

**Other information:**

In addition to the global event channel, Tauri now provides events to specific windows.


api/src/event.ts
```diff
@@ -1,49 +1,8 @@
-import { invoke, transformCallback } from './tauri'
+import { emit as emitEvent } from './helpers/event'
```
chippers, line 1: Since users can call this export from the helper anyways, what about moving the one that accepts a window label to:
 ```typescript
/**
 * Emits an event for a specific window to the backend. This should probably not be called manually,
 * use `emit` from `TauriWindow` (link if possible from JSDoc) to emit an event for a specific window.
 */
async function emitWindow(event: string, windowLabel: string, payload: string): Promise<void> {
  /* ... */
}
```
with a doc comment mentioning how it should probably not be called explicitly, and reference the docs to `TauriWindow`'s `emit`.  Avoids an extra directory + file for something that doesn't really prevent it anyways.

Then we can keep emit as
```typescript
async function emit(event: string, payload?: string): Promise<void> {
  return await emitWindow(event, undefined, payload)
}
```

lucasfernog: IMO it looks weird that way, at least to import the helper someone really need to dig into the source code. Exposing on the /event api would show up with code completion and would confuse regular users.

PR #1296, opened 2021-02-25
feat(core): Use Wry custom protocol instead of embedded server

nklayman:

- [ ] Bugfix
- [x] Feature
- [x] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes. Issue #___
- [ ] No

**Other information:**
YAY! It's finally here!!!!


lucasfernog (approved):

lucasfernog:
This still needs a fix on macOS.

tauri/src/app/webview/wry.rs
```diff
@@ -218,8 +218,7 @@ impl ApplicationDispatcherExt for WryDispatcher {
         Some(wry_callbacks),
         custom_protocol.map(|p| wry::CustomProtocol {
           name: p.name.clone(),
-          // TODO: Map error type instead of unwraping
-          handler: Box::new(move |a| Ok((*p.handler)(a).unwrap())),
+          handler: Box::new(move |a| (*p.handler)(a).map_err(|_| wry::Error::InitScriptError)),
```
nklayman, line 221: Shouldn't we add a TODO for including the path of the asset so the error message is more helpful?

tauri/src/app/webview/wry.rs
```diff
@@ -477,8 +476,7 @@ impl ApplicationExt for WryApplication {
...
```
nklayman, line 470: same as above

tauri/src/app/utils.rs
```diff
@@ -256,15 +231,32 @@ pub(super) fn build_webview<A: ApplicationExt + 'static>(
     let custom_protocol = CustomProtocol {
       name: "tauri".into(),
       handler: Box::new(move |path| {
-        assets
-          .get(&Assets::format_key(path), AssetFetch::Decompress)
-          .ok_or_else(|| crate::Error::AssetNotFound(path.to_string()))
+        let path = path.to_string();
+        #[cfg(target_os = "macos")]
+        let path = path.replace("tauri://", "");
+        #[cfg(any(target_os = "macos", target_os = "windows"))]
```
nklayman: We need to change Wry to not handle this for linux and then handle it here.

tauri/src/app/webview/wry.rs
lucasfernog: wry doesn't have any kind of error handling on custom protocol right now, so the error doesn't matter; it should probably eval a `console.error` on the webview. Can you look into that or maybe open an issue?

nklayman: On linux at least it gives you a pretty nice error on the Webview, I assume other platforms will do the same as it will be just like a failed network request, so this can be resolved.
image

PR #1301, opened 2021-02-26
feat(core): improved command matching with macros, fixes #1157

### LINKED ISSUE #1157: Simplify Command Matching
## The Problem

The main.rs file gets pretty bulky. Yes, things can be split apart into simpler chunks, and that's what I'm proposing - I just think Tauri can assist with some "sugar".
This probably isn't a priority. I just think of how much simpler it *could* be, especially for beginners to Rust.

Specifically I'd like to point out:

- deserializing the command from a string
- matching commands inline
- `move ||`
- destructuring and passing along Tauri-specific values like the callback and error fields

image
[Link to code](https://github.com/parker-codes/todos-app/blob/master/src-tauri/src/main.rs#L18-L80)

## Preferred Solution

I see a few options:

1. Create a macro that converts all variant matches into methods of the same name, like `GetAllTodos` into `get_all_todos()`. The problem then becomes passing the value into the method. If there are multiple fields in the variant, would they be passed in order? That doesn't seem right. It would be cleanest to pass the command itself, but enum variants aren't real types, so the value can't be destructured and passed in. [Here's a playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=76147320e0edfa8c13dd93767dd1fad1) of how the command can be passed, but it still needs to be matched on the other end, which detracts from the desired cleanliness. 
2. Deserialize the command into a tuple variant. I think the syntax is quite ugly and confusing, but all of the RFCs have been closed or slow moving. See https://github.com/rust-lang/rfcs/pull/1450 and https://github.com/rust-lang/rfcs/pull/2593. Also see this [Stack Overflow post](https://stackoverflow.com/questions/49705007/is-there-a-way-to-use-existing-structs-as-enum-variants) for clarification on why this is ugly. Serde has some great features that may be able to deserialize this, like `flatten`. This way would probably be pretty simple.

Both of these together could look like:

```rs
#[derive(Deserialize)]
struct GetAllTodos {
    callback: String,
    error: String,
}

#[derive(Deserialize)]
struct CreateTodo {
    title: String,
    callback: String,
    error: String,
}

// some type of flatten?
#[derive(Deserialize)]
#[serde(tag = "cmd", rename_all = "camelCase", flatten)]
pub enum Cmd {
    GetAllTodos(GetAllTodos),

    CreateTodo(CreateTodo),

    // .. more
}
```

And then potentially an attribute or macro to generate this, if desired: 
```rs
match command {
    Cmd::GetAllTodos(cmd) => get_all_todos(cmd),
    Cmd::CreateTodo(cmd) => create_todo(cmd),
}
```

And the usage could look like:
```rs
fn get_all_todos(cmd: GetAllTodos) {
    // stuff here
}
```

Another thing to think about here would be handling responses and return types like via `tauri::execute_promise()`. Is there any way for these methods and params to be abstracted or hidden as well (like success and error callbacks)? Perhaps the commands or functions could have an attribute like `#[tauri(type = "execute_promise")]` which acts like a decorator?

## Conclusion

This is all to hopefully help simplify how people use it - not necessarily code cleanliness. If there is a "cleaner" structure that can be done without any API changes, perhaps it's worth showcasing that in the examples instead of the very straightforward examples that are there currently.

nklayman:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No
- [x] Kinda

**Other information:**

This PR adds two macros, `#[command]` and `generate_handler![]`. Combined, these allow command handling to look like this:

```rust
// Definition in main.rs

#[command]
fn reply_string(s: String) -> String {
  s
}
// Or, if the function needs the WebviewManager
#[command(with_manager)]
fn reply_string(manager: tauri::WebviewManager, s: String) -> String {
  println!("{}", manager.current_window_label());
  s
}

fn main() {
  tauri::AppBuilder::<Context>::new()
    .invoke_handler(generate_handler![reply_string])
    .build()
    .unwrap()
    .run();
}
```

```js
// Calling from js
window.__TAURI__
  .invoke({
    cmd: 'reply_string',
    s: 'Hi'
  })
  .then((res) => console.log('Response:', res))
```

`#[command]` adds a second function, `NAME_wrapper` to function with name `NAME`. This wrapper function accepts the `WebviewManager` and `serde_json::Value` as args, and returns a `tauri::InvokeResponse`. The `serde_json::Value` should match the names and types of the original function args, ie the JSON for `fn original(name: String)` would be `{"name": "some string"}`. The wrapper function casts the value to a typed struct that contains all the function args, and passes them to the original function. It also passes the `WebviewManager` as the first arg if it is a `with_webview` command. It then converts the response from the original function to a `tauri::InvokeResponse`, and then returns it.

The `generate_handler![]` macro takes an array of paths to `#[command]` functions and returns a closure compatible with `invoke_handler`. This closure parses the arg string into a `tauri::DispatchInstructions`, a simple struct with two params: `cmd`, a string and `args`, a `serde_json::Value`. The `cmd` param is used in a generated match statement that matches it to one of the functions passed to this macro. It then calls that function's wrapper function with the `WebviewManager` and `args`, the JSON value that should hold all of that function's arguments. It returns the output from that function back to the webview.

TODO:
- [x] Clean up macro and use less .unwrap()
- [x] Have commands return a `Result` for better error handling
- [x] Improve error message from arg parse failure
- [x] Add support for functions that take the webview as an arg
- [x] ~~Support for inferring webview arg type~~ We decided to not include this
- [x] Support async functions
- [x] Update JS API to have cleaner syntax
- [x] Update init template to use macros
- [x] Update examples to use macros
- [ ] (after merge) Type check the webview arg for `with_webview` commands
- [ ] (after merge) Add support for state/context


tauri/src/app.rs
```diff
@@ -62,6 +62,14 @@ impl<T: Serialize> From<T> for InvokeResponse {
   }
 }
 
+#[derive(Deserialize)]
+#[allow(missing_docs)]
+#[serde(tag = "cmd", rename_all = "camelCase")]
+pub struct DispatchInstructions {
```
lucasfernog, line 69: this forces the user to call the API with `{ type: 'myType', args: JSON.stringify({ my: 'args', like: 'this' })` right?

We can skip the JSON.stringify need by declaring the struct like this:

```rust
#[derive(Deserialize)]
pub struct DispatchInstructions {
  pub cmd: String,
  #[serde(flatten)]
  pub args: serde_json::Value,
}
```

lucasfernog: also, this struct could just be defined on the macro instead right? probably better

nklayman: Yeah it does require that. Would that move all other fields in the object into the `args` param then?
I also think we should change the `invoke` js function to work like `invoke('command_name', args)` instead of grouping the params and command name together. Especially if someone sorts their objects alphabetically, it could be confusing to have the command name stuck in with the args.

lucasfernog: yeah every field that's not named `cmd` would be part of the `args` value.

lucasfernog: yeah i agree with the invoke change, we can do it on the next PR.

lucasfernog: btw should we change cases here? `invoke('commandName')` to run the `fn command_name`? since the default on JS is camelCase, and on Rust it's snake_case

nklayman: I just added the flatten stuff :+1:

tauri-macros/src/command.rs
```diff
@@ -0,0 +1,96 @@
...
+        application_ext_generic = quote!();
+        types.drain(0..1);
+        names.drain(0..1);
+        quote!(_webview,)
+      } else {
+        quote!()
+      }
+    }
+    None => quote!(),
+  };
```
chippers: Because macros don't have type information, this is only checking that the type ident only matches a `WebviewManager` directly.  Since that seems like an easy enough name collision that will cause unexpected (for the user) name conflicts, i would suggest detecting if there is a webview being passed into the arguments with an item.  maybe something like
```rust
#[command(webview = "webview")]
fn reply_string(webview: tauri::WebviewManager<tauri::flavors::Wry>, s: String) -> String {
  println!("{}", webview.current_window_label());
  s
}
```

It's slightly repetitive if they are both called webview, but is very explicit and can prevent conflicts.  If you know the properties of all webview managers, you can even compile-time test inside that property to spit out a more specific compile error if the `webview` item parameter isnt actually a webview type.

nklayman: Suggestion applied, thanks for the tip! However, I implemented it as a boolean that would set the first arg to be the webview.

tauri-macros/src/command.rs
```diff
@@ -0,0 +1,116 @@
...
+        arg_name.clone().unwrap(),
+        arg_type.unwrap_or_else(|| panic!("Invalid type for arg \"{}\"", arg_name.unwrap())),
+      )
+    })
+    .unzip();
+
+  // If function doesn't take the webview, wrapper just takes webview generically and ignores it
+  // Otherwise the wrapper uses the specific type from the original function declaration
+  let mut webview_arg_type = quote!(tauri::WebviewManager<A>);
+  let mut application_ext_generic = quote!(<A: tauri::ApplicationExt>);
```
chippers: we can use the root crate path to prevent any collisions with any in-scope `tauri` items that aren't actually the crate

```suggestion
  let mut webview_arg_type = quote!(::tauri::WebviewManager<A>);
  let mut application_ext_generic = quote!(<A: ::tauri::ApplicationExt>);
```

tauri-macros/src/command.rs
```diff
@@ -0,0 +1,116 @@
...
+      // Tell wrapper to pass webview to original function
+      quote!(_webview,)
+    }
+    // Tell wrapper not to pass webview to original function
+    _ => quote!(),
+  };
+
+  quote! {
+    #function
+    pub fn #fn_wrapper #application_ext_generic(_webview: #webview_arg_type, arg: serde_json::Value) -> Option<tauri::InvokeResponse> {
```
chippers:
```suggestion
    pub fn #fn_wrapper #application_ext_generic(_webview: #webview_arg_type, arg: ::serde_json::Value) -> Option<::tauri::InvokeResponse> {
```

tauri-macros/src/command.rs
```diff
@@ -0,0 +1,116 @@
...
+      #[derive(serde::Deserialize)]
+      #[serde(rename_all = "camelCase")]
+      struct ParsedArgs {
+        #(#names: #types),*
+      }
+      let parsed_args: ParsedArgs = serde_json::from_value(arg).unwrap();
```
chippers:
```suggestion
      let parsed_args: ParsedArgs = ::serde_json::from_value(arg).unwrap();
```

tauri-macros/src/command.rs
```diff
@@ -0,0 +1,116 @@
...
+      Some(#fn_name(#webview_arg_maybe #(parsed_args.#names),*).into())
```
chippers:
```suggestion
      ::std::option::Option::Some(#fn_name(#webview_arg_maybe #(parsed_args.#names),*).into())
```

tauri-macros/src/command.rs
```diff
@@ -0,0 +1,116 @@
...
+      quote!(_webview,)
+    }
+    // Tell wrapper not to pass webview to original function
+    _ => quote!(),
+  };
+
+  quote! {
+    #function
+    pub fn #fn_wrapper #application_ext_generic(_webview: #webview_arg_type, arg: serde_json::Value) -> Option<tauri::InvokeResponse> {
+      #[derive(serde::Deserialize)]
```
chippers:
```suggestion
      #[derive(::serde::Deserialize)]
```
This one might require more thought because idk how this will handle `serde` not having the `"derive"` feature enabled - or what the solution might be

tauri-macros/src/command.rs
```diff
@@ -0,0 +1,116 @@
...
+    let mut func = func.clone();
+    let mut last_segment = func.segments.last_mut().unwrap();
+    last_segment.ident = format_ident!("{}_wrapper", last_segment.ident);
+    func
+  });
+
+  quote! {
+    |webview, arg| async move {
+      let dispatch: Result<tauri::DispatchInstructions, serde_json::Error> =
+      serde_json::from_str(&arg);
```
chippers:
```suggestion
      let dispatch: ::std::result::Result<::tauri::DispatchInstructions, ::serde_json::Error> =
      ::serde_json::from_str(&arg);
```

chippers (commented):
we can use the root crate path to prevent any collisions with any in-scope items that clash (more likely for `Result`, but also possible for the crate names)

lucasfernog (approved):

PR #1303, opened 2021-02-28
use Wry as default ApplicationExt

chippers:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes. Context now needs to be first during `AppBuilder` (no releases with Context yet and it will change later anyways)
- [ ] No

**Other information:**
When using `::tauri::WebviewManager` without generic arguments, it now defaults to using `WryApplication`

### DIFF
--- a/tauri/src/app.rs
+++ b/tauri/src/app.rs
@@ -13,6 +13,7 @@ pub(crate) mod webview;
 mod webview_manager;
 
 pub use crate::api::config::WindowUrl;
+use crate::flavors::Wry;
 pub use webview::{
   wry::WryApplication, ApplicationDispatcherExt, ApplicationExt, Callback, Icon, Message,
   WebviewBuilderExt,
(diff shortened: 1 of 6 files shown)


lucasfernog (approved):

tauri/src/app.rs
```diff
@@ -199,7 +203,7 @@ pub struct AppBuilder<A: ApplicationExt, C: AsTauriContext> {
   webviews: Vec<Webview<A>>,
 }
 
-impl<A: ApplicationExt + 'static, C: AsTauriContext> AppBuilder<A, C> {
+impl<A: ApplicationExt + 'static, C: AsTauriContext> AppBuilder<C, A> {
```
nklayman, line 206: It's not really that important but it seems like it'd be best to switch the type defs too so it isn't backwards.

nklayman (approved):

tauri/src/app.rs
chippers: i didn't really focus on it because it's going away during #1226

chippers:
ah the formatter is using nightly

tauri/src/app.rs
nklayman: Oh cool!

PR #1362, opened 2021-03-17
fix(cli): use correct arg in `_blanks` links polyfill

amrbashir:

- [X] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [X] No

**Other information:**

### DIFF
--- a/.changes/fix-blank-hyperlinks-polyfiil.md
+++ b/.changes/fix-blank-hyperlinks-polyfiil.md
@@ -1,0 +1,5 @@
+---
+"tauri-cli": patch
+---
+
+Fixes `<a target="_blank">` polyfill.
--- a/cli/core/src/templates/tauri.js
+++ b/cli/core/src/templates/tauri.js
@@ -171,7 +171,7 @@ if (!String.prototype.startsWith) {
                 __tauriModule: "Shell",
                 message: {
                   cmd: "open",
-                  uri: target.href,
+                  path: target.href,
                 },
               });
               e.preventDefault();
--- a/examples/api/src/App.svelte
+++ b/examples/api/src/App.svelte
@@ -1,6 +1,6 @@
 <script>
   import { onMount } from "svelte";
-  import { open } from "@tauri-apps/api/window";
+  import { open } from "@tauri-apps/api/shell";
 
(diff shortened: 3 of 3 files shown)


lucasfernog (approved):

lucasfernog:
Thanks for catching this :)

amrbashir:
No problem, I was hesitant to fix the `exmaples/api` (kinda related but not really😂) in the same PR, let me know and I can revert it.

lucasfernog:
It's a small regression and it's related so we can keep it here too.

lucasfernog:
By the way it would be super cool if you could setup commit signing when you have some free time :) it's super nice

amrbashir:
Will do.

nklayman:
Should we make _blank open a new Tauri window? That seems like better behavior and is what electron does.

lucasfernog:
I would love to see what @nothingismagick thinks about that.

nothingismagick:
I left a comment in the discord, but will post it here:

Should absolutely be configurable IMHO. Either globally or as an attr... I was actually thinking about that the other day. And defaulting to open the default system browser tab could also be a theoretical security risk.

Going off of what Parker said, to be specific, something like a:
- config option to specify block-all or allow-all, or run it through a function

It could also be a data-action attribute that the dev could specify, however in case they ingest untrusted JS it would probably be good to have a policy about what is allowed...

nothingismagick:
`<a href="/some_path" target="_blank" data-open="subwindow">LINK</a>`
`<a href="/some_path" target="_blank" data-open="browser">LINK</a>`
`<a href="/some_path" target="_blank" data-open="self">LINK</a>`

amrbashir:
I like what Parker proposed but also the data-attribute is good to have too so maybe add a field in `tauri.conf.json`:
```jsonc
"links": {
	"mode": "allow-all", // or "block-all"
	"allow-strategy": "data-attribute" // or "function"
}
```

nklayman:
Also, if we default to opening the user's browser, it could be considered a security risk because a malicious link could become RCE.

PR #1370, opened 2021-03-18
Use JSON.parse instead of literal JS for callback formatting

WilliamVenner:
https://github.com/GoogleChromeLabs/json-parse-benchmark

- [x] Other, please describe:

Optimisation

- [x] No

**Other information:**

Believe it or not, `JSON.parse` is [actually faster](https://github.com/GoogleChromeLabs/json-parse-benchmark) than directly evaluating literal JavaScript for JSON objects.

It does actually make sense - when evaluating actual JavaScript the interpreter is looking for variables to interpolate, and is using a parsing pipeline made for interpreting actual code. Whereas with `JSON.parse`, it's just looking at raw JSON, which is of course much faster than interpreting actual JavaScript.

This should introduce up to x2 speed improvements for formatting callback JS.

Unfortunately I don't know how to set up the dev environment for Tauri right now, it's very complex and I failed the last time I tried. Just want to get this PR up here so it will at least eventually be implemented. If I manage to get a dev environment set up or a maintainer can contribute (edits are on) then the following stuff needs doing for this PR to be complete:

- [x] Changes file
- [x] Object syntax checking - only JSON.parse if the JSON string begins with `{` or `[` (the speed improvement only applies for JSON objects/arrays, not literals such as strings, booleans, numbers, etc.)
- [x] String length check + use literal if larger than the maximum string length
> ECMAScript 2016 (ed. 7) established a maximum length of 2^53 - 1 elements. Previously, no maximum length was specified. In Firefox, strings have a maximum length of 2**30 - 2 (~1GB).


tauri-api/src/rpc.rs
```diff
@@ -25,16 +30,37 @@ use serde_json::Value as JsonValue;
 /// assert!(cb.contains(r#"window["callback-function-name"]({"value":"some value"})"#));
 /// ```
 pub fn format_callback<T: Into<JsonValue>, S: AsRef<str>>(function_name: S, arg: T) -> String {
+  let as_str = {
+    let json_value = arg.into();
+    let as_str = json_value.to_string();
+    if as_str.len() < MAX_JSON_STR_LEN {
+      if let &JsonValue::Array(_) | &JsonValue::Object(_) = &json_value {
```
lucasfernog:
```suggestion
      if matches!(json_value, JsonValue::Array(_) | JsonValue::Object(_)) {
```

tauri-api/src/rpc.rs
```diff
@@ -25,16 +30,37 @@ use serde_json::Value as JsonValue;
...
+      }
+    }
+    as_str
+  };
+
   format!(
     r#"
       if (window["{fn}"]) {{
-        window["{fn}"]({arg})
+        window["{fn}"]({arg}))
```
lucasfernog: why the extra `)`?

WilliamVenner:
Tested the following conditions:

- [X] Responding with a HashMap
- [X] Responding with a >1 GB HashMap (took a long time, but did send as a js literal)
- [X] Responding with a primitive `&'static str` and `String`

WilliamVenner:
https://github.com/tauri-apps/tauri/blob/f57b53314418008eee7ecdfd614740ab3f92b107/tauri-api/src/rpc.rs#L9-L55

tauri-api/src/rpc.rs
```diff
@@ -1,8 +1,65 @@
 use serde::Serialize;
 use serde_json::Value as JsonValue;
 
+/// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length#description
+//  ECMAScript 2016 (ed. 7) established a maximum length of 2^53 - 1 elements. Previously, no maximum length was specified.
+//  In Firefox, strings have a maximum length of 2**30 - 2 (~1GB). In versions prior to Firefox 65, the maximum length was 2**28 - 1 (~256MB).
+pub const MAX_JSON_STR_LEN: usize = usize::pow(2, 30) - 2;
```
nklayman, line 11: Tauri doesn't use Firefox on any platform, so you should be good to use the larger constraint defined by ES2016.

tauri-api/src/rpc.rs
```diff
@@ -1,8 +1,65 @@
...
+
+/// Safely transforms & escapes a JSON String -> JSON.parse('{json}')
+//  Single quotes are the fastest string for the JavaScript engine to build.
+//  Directly transforming the string byte-by-byte is faster than a double String::replace()
+pub fn escape_json_parse(mut json: String) -> String {
+  const BACKSLASH_BYTE: u8 = '\\' as u8;
```
nklayman: Do you need to declare the type twice here?

tauri-api/src/rpc.rs
```diff
@@ -22,9 +79,30 @@ use serde_json::Value as JsonValue;
 /// let cb = format_callback("callback-function-name", serde_json::to_value(&MyResponse {
 ///   value: "some value".to_string()
 /// }).expect("failed to serialize"));
-/// assert!(cb.contains(r#"window["callback-function-name"]({"value":"some value"})"#));
+/// assert!(cb.contains(r#"window["callback-function-name"](JSON.parse('{"value":"some value"}'))"#));
 /// ```
 pub fn format_callback<T: Into<JsonValue>, S: AsRef<str>>(function_name: S, arg: T) -> String {
+  let as_str = {
```
nklayman: Why have this? Since this just returns `as_str` at the end anyways, you could get rid of it and save a few lines.

tauri-api/src/rpc.rs
```diff
@@ -22,9 +79,30 @@ use serde_json::Value as JsonValue;
...
+              console.warn("[TAURI] Couldn't find callback id {fn} in window. This happens when the app is reloaded while Rust is running an asynchronous operation.")
+            }}
+          "#,
+          fn = function_name.as_ref(),
+          arg = escape_json_parse(as_str)
+        )
+      }
+    }
+    as_str
+  };
```
nklayman: See above, I think these could be removed.

nklayman (changes requested):
Looks good overall, just a few suggestions. Also, the formatting appears to be off, make sure to reformat the `rpc.rs` file when you're done.

tauri-api/src/rpc.rs
WilliamVenner: Yes - it casts `char` to its `u8` byte equivalent.

tauri-api/src/rpc.rs
WilliamVenner: It implicitly drops `json_value` as it's not needed outside of that block.

WilliamVenner: `json_value` actually stores the value of the JSON inside the enum, so if it's not dropped it could end up being a horrible waste of memory. Upon review, I am actually about to improve the memory usage a bit in the next commit.

tauri-api/src/rpc.rs
WilliamVenner: Ok, I actually tested on Chrome and Edge and it turns out their maximum string size is the same as Firefox's.

WilliamVenner: https://stackoverflow.com/a/34958490

tauri-api/src/rpc.rs
nklayman: Oh yeah it's a const so you can't let it infer the type, my bad :facepalm:.

tauri-api/src/rpc.rs
nklayman: It would get dropped at the end of the function anyways though. Wouldn't that be fine?

WilliamVenner: Nope, if you want to push a massive string, like a few GB, it'd end up duplicating in memory for no reason.

Demonstration that even on release mode, it doesn't aggressively drop stuff:

https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=b8b5df6933a792c03f40e221c29084ed

I would imagine the compiler assumes the value is most likely not going to be big enough to aggressively drop. Uncomment the explicit drop to see the difference.

It'd be much better if you could explicitly consume the value into a string. Maybe I'll look into it a bit more.

WilliamVenner: Yeah there doesn't seem to be a way to consume it into a string yet, unfortunately. This is the best we can do.

nklayman: Okay, sounds good then :+1:.

nklayman (approved):
Looks good now, great job!

tauri-api/src/rpc.rs
```diff
@@ -1,8 +1,75 @@
...
+///
+/// In Firefox, strings have a maximum length of 2\*\*30 - 2 (~1GB). In versions prior to Firefox 65, the maximum length was 2\*\*28 - 1 (~256MB).
+pub const MAX_JSON_STR_LEN: usize = usize::pow(2, 30) - 2;
+
+/// Safely transforms & escapes a JSON String -> JSON.parse('{json}')
+//  Single quotes are the fastest string for the JavaScript engine to build.
+//  Directly transforming the string byte-by-byte is faster than a double String::replace()
+pub fn escape_json_parse(mut json: String) -> String {
+  const BACKSLASH_BYTE: u8 = '\\' as u8;
+  const SINGLE_QUOTE_BYTE: u8 = '\'' as u8;
```
nklayman: Fix the clippy error here with:
```rust
  const BACKSLASH_BYTE: u8 = b'\\';
  const SINGLE_QUOTE_BYTE: u8 = b'\'';
```

WilliamVenner: Nice one!

tauri-api/src/rpc.rs
lucasfernog: can you update the comment here to reflect that?

lucasfernog (approved):
I love this, such a clever PR!

nklayman:
Delete line 56 to fix the check btw.

tauri-api/src/rpc.rs
lucasfernog: @WilliamVenner pls :pleading_face:

WilliamVenner: @lucasfernog Sorry, didn't see your comment here. The comment has already been updated to reflect that information.

lucasfernog: oh great, sorry i missed that too (this was a resolved conversation anyway :D )

nklayman:
Looks like you need to rerun formatting on it :/

PR #1371, opened 2021-03-18
fix(config): serde_json::Value::Object takes a Map instead of HashMap

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

The ToTokens impl for the PluginConfig was failing to compile since serde_json::Value::Object takes a serde_json::Map and the hashmap_lit was giving it a std::collections::HashMap.

### DIFF
--- a/tauri-utils/src/config.rs
+++ b/tauri-utils/src/config.rs
@@ -448,10 +448,11 @@ mod build {
     quote! { vec![#(#items),*] }
   }
 
-  /// Create a `HashMap` constructor, mapping keys and values with other `TokenStream`s.
+  /// Create a map constructor, mapping keys and values with other `TokenStream`s.
   ///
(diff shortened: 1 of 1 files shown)


wusyong (approved):

chippers (approved):
Looks fine to me, I must have not tested any plugin configs that used a map when i was creating it so that enum type value must have slipped through the cracks. this should work cause `Map` and `HashMap` both have basically the same parameters for `new()` and `insert()`, with the `Map` requiring `String`.  Since the transformer for the string value uses `str_lit()` it should be fine

PR #1379, opened 2021-03-23
fix(tauri-api): remove `.exe` from `app_name` on windows

amrbashir:

- [X] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [X] No

**Other information:**
Previously, On windows , the app_dir would be something like this `C:\Users\username\AppData\Roaming\app_name.exe\`

### DIFF
--- a/tauri-api/src/path.rs
+++ b/tauri-api/src/path.rs
@@ -193,7 +193,11 @@ fn app_name() -> crate::Result<String> {
     .expect("failed to get exe filename")
     .to_string_lossy();
 
-  Ok(app_name.to_string())
+  if cfg!(target_os = "windows") {
+    Ok(app_name.replace(".exe", ""))
+  } else {
+    Ok(app_name.to_string())
+  }
 }
 
 /// Returns the path to the suggested directory for your app config files.
--- a/tauri-api/src/path.rs
+++ b/tauri-api/src/path.rs
@@ -189,15 +189,11 @@ pub fn resource_dir() -> Option<PathBuf> {
 fn app_name() -> crate::Result<String> {
   let exe = std::env::current_exe()?;
   let app_name = exe
-    .file_name()
+    .file_stem()
     .expect("failed to get exe filename")
     .to_string_lossy();
 
-  if cfg!(target_os = "windows") {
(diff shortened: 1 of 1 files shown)


tauri-api/src/path.rs
```diff
@@ -193,7 +193,11 @@ fn app_name() -> crate::Result<String> {
     .expect("failed to get exe filename")
     .to_string_lossy();
 
-  Ok(app_name.to_string())
+  if cfg!(target_os = "windows") {
+    Ok(app_name.replace(".exe", ""))
```
lucasfernog: We should probably just use file_stem instead of file_name, see https://doc.rust-lang.org/std/path/struct.Path.html#method.file_stem

tauri-api/src/path.rs
```diff
@@ -193,7 +193,11 @@ fn app_name() -> crate::Result<String> {
     .expect("failed to get exe filename")
     .to_string_lossy();
 
-  Ok(app_name.to_string())
+  if cfg!(target_os = "windows") {
```
wusyong:
```suggestion
#[cfg(target_os = "windows")]
```
How about this and `#[cfg(not(target_os = "windows"))]` below.
Conditional compilation can include the only code required. This could reduce the branch condition in actual code base.

lucasfernog (changes requested):

tauri-api/src/path.rs
amrbashir: I tried that but the compiler kept yelling about missing semicolon. Anyway Lucas suggested we use `file_stem` so we won't need any conditions.

tauri-api/src/path.rs
amrbashir: yup that is much better.

lucasfernog (approved):

lucasfernog:
I see your commit signature, @amrbashir , nice work :D

amrbashir:
I've seen the `Verified` before but thought it was reserved to repo owners or something LOL. Thanks for telling me.

PR #1392, opened 2021-03-26
refactor(core): remove async from app hooks, add InvokeMessage type

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No

**Other information:**

Simplifying the API a little bit, the `Sync` requirement was too much :P command macros are still the same.

### DIFF
--- a/tauri/src/app.rs
+++ b/tauri/src/app.rs
@@ -1,7 +1,10 @@
-use futures::future::BoxFuture;
 use serde::{Deserialize, Serialize};
 use serde_json::Value as JsonValue;
-use tauri_api::{config::Config, private::AsTauriContext};
+use tauri_api::{
+  config::Config,
+  private::AsTauriContext,
+  rpc::{format_callback, format_callback_result},
+};
 
 use std::{
   collections::HashMap,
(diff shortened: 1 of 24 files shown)


tauri/src/plugin.rs
```diff
@@ -24,67 +20,53 @@ pub trait Plugin<A: ApplicationExt + 'static>: Send + Sync {
...
-    command: String,
-    payload: &JsonValue,
-  ) -> crate::Result<JsonValue> {
-    Err(crate::Error::UnknownApi(None))
-  }
+  fn extend_api(&mut self, webview_manager: WebviewManager<A>, message: InvokeMessage<A>) {}
 }
 
 /// Plugin collection type.
 pub type PluginStore<A> = Arc<Mutex<Vec<Box<dyn Plugin<A> + Sync + Send>>>>;
```
wusyong: I assume the Sync trait here can be removed too.

tauri/src/app.rs
```diff
@@ -21,12 +24,171 @@ pub use webview::{
...
+  callback: String,
+  error: String,
+  #[serde(rename = "mainThread", default)]
+  pub(crate) main_thread: bool,
+  #[serde(flatten)]
+  inner: JsonValue,
+}
+
+/// An invoke message.
+pub struct InvokeMessage<A: ApplicationExt> {
```
wusyong, line 45: Is the reason for this type being public because of macro? If so, maybe we can add a doc(hidden) to it.

lucasfernog: this is a wrapper around the wry rpc and also provides APIs to resolve/reject the promise. I think I'll let it be documented since we do not force users to use the macros.

Issue #1394, opened 2021-03-27
cargo check, cargo build extremely slow even on development profile


WilliamVenner:
**Describe the bug**
Building for development profile is extremely slow, sometimes taking up to 5 minutes long.

**To Reproduce**
1. Add some assets into the configured dist folder
2. Run cargo check
3. Wait forever

**Expected behavior**
cargo check and cargo build (_without_ --release) should be fast and snappy to aid the development process.

**Screenshots**
![](https://cdn.discordapp.com/attachments/731495047962558564/825453010766921728/unknown.png)

`cargo rustc --profile=check -- -Zunstable-options --pretty=expanded`

![](https://cdn.discordapp.com/attachments/731495047962558564/825456686382186506/unknown.png)

**Platform and Versions (please complete the following information):**

OS: Windows 10 64-bit
Node: 14.15.0
NPM: 6.14.8
Yarn: 2.0.0-rc.27
Rustc: 1.15.0

[cargo-timing-20210327T191032Z.zip](https://github.com/tauri-apps/tauri/files/6216390/cargo-timing-20210327T191032Z.zip)

WilliamVenner:
It's back @chippers ?

chippers:
> 
> 
> It's back @chippers ?

I think it got reverted because we weren't able to serve files in dev. the refactor needed to fix it properly just landed this weekend so i can get a similar fix in the next day or so

PR #1395, opened 2021-03-27
Fix: #1394 Extremely slow dev builds caused by embedding + compressing assets at compile time

WilliamVenner:

- [X] Bugfix #1394

- [X] No

**Other information:**

On development builds, Tauri will now directly serve files from the configured dist folder instead of embedding, recompiling and compressing every time.

### DIFF
--- a/core/tauri-codegen/src/context.rs
+++ b/core/tauri-codegen/src/context.rs
@@ -24,12 +24,17 @@ pub fn context_codegen(data: ContextData) -> Result<TokenStream, EmbeddedAssetsE
   let assets = EmbeddedAssets::new(&dist_dir)?;
 
   // handle default window icons for Windows targets
+  #[cfg(not(debug_assertions))]
   let default_window_icon = if cfg!(windows) {
     let icon_path = config_parent.join("icons/icon.ico").display().to_string();
     quote!(Some(include_bytes!(#icon_path)))
   } else {
     quote!(None)
   };
+
+  // in development builds, don't use an icon as it slows down cargo check
+  #[cfg(debug_assertions)]
+  let default_window_icon = quote!(None);
 
   // double braces are purposeful to force the code into a block expression
   Ok(quote! {{
--- a/core/tauri-codegen/src/embedded_assets.rs
+++ b/core/tauri-codegen/src/embedded_assets.rs
@@ -51,6 +51,7 @@ pub enum EmbeddedAssetsError {
 pub struct EmbeddedAssets(HashMap<AssetKey, (String, Vec<u8>)>);
 
 impl EmbeddedAssets {
+  #[cfg(not(debug_assertions))]
   /// Compress a directory of assets, ready to be generated into a [`tauri_api::assets::Assets`].
   pub fn new(path: &Path) -> Result<Self, EmbeddedAssetsError> {
     WalkDir::new(&path)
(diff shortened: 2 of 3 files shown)


core/tauri-codegen/src/context.rs
```diff
@@ -24,13 +24,18 @@ pub fn context_codegen(data: ContextData) -> Result<TokenStream, EmbeddedAssetsE
...
     let icon_path = config_parent.join("icons/icon.ico").display().to_string();
     quote!(Some(include_bytes!(#icon_path)))
   } else {
     quote!(None)
   };
 
+  // in development builds, don't use an icon as it slows down cargo check
+  #[cfg(debug_assertions)]
+  let default_window_icon = quote!(None);
+
```
chippers, line 38:
```suggestion
  // in development builds, don't use an icon as it slows down cargo check
  let default_window_icon = if cfg!(windows) && cfg!(not(debug_asssertions)) {
    let icon_path = config_parent.join("icons/icon.ico").display().to_string();
    quote!(Some(include_bytes!(#icon_path)))
  } else {
    quote!(None)
  };
```

tauri/src/app/utils.rs
```diff
@@ -215,17 +218,42 @@ pub(super) fn build_webview<A: ApplicationExt + 'static>(
...
+            })
+        }
+
+        // In release builds, fetch + serve decompressed embedded assets.
+        #[cfg(not(debug_assertions))]
+        {
+          assets
+            .get(&path)
+            .ok_or(crate::Error::AssetNotFound(path))
+            .map(Cow::into_owned)
```
chippers, line 256: should move this functionality to a struct in `tauri-utils/src/assets.rs` that implements `Assets`.  We can then leave the debug logic to the macro when it's generated and this file could be left alone.  the struct probably needs a constructor like
```rust
pub const fn from_dist_dir(dist_dir: impl Into<PathBuf>) -> Self {
  Self(dist_dir))
}
```
and then we can construct it from the macro.

core/tauri-codegen/src/embedded_assets.rs
```diff
@@ -73,6 +74,14 @@ impl EmbeddedAssets {
...
   }
 
+  #[cfg(debug_assertions)]
+  /// A dummy EmbeddedAssets for use during development builds.
+  /// Compressing + including the bytes of assets during development takes a long time.
+  /// On development builds, assets will simply be resolved & fetched from the configured dist folder.
+  pub fn new(_: &Path) -> Result<Self, EmbeddedAssetsError> {
+    Ok(EmbeddedAssets(HashMap::new()))
+  }
+
```
chippers, line 84: we can change this logic to a `if cfg!` in the `context.rs` file to create the file system assets struct if debug_assertions are enabled

chippers (changes requested):
Thanks for contributing!

I tried to alleviate some pain by tuning the compression level to release mode or not, but the bulk of the time is including them anyways.

chippers:
ill go ahead and implement the changes

lucasfernog:
Note that the tauri crate can detect dev better than checking debug assertions: https://github.com/tauri-apps/tauri/blob/dev/tauri/src/app/utils.rs#L19

So a flag can be passed around instead.

lucasfernog (approved):

lucasfernog:
@chippers will take care of the requested changes.

PR #1400, opened 2021-03-29
wip: big refactor, removing application cycle + allowing generic parameters to the application

chippers:
Compiles, but doesn't start up yet. Many things probably broken. I focused on bringing `application` out from under `WebviewManager`. There was some accidental refactoring of `PluginStore` and `Listeners`.

big features:
* can have fully typed events and window labels (unless i messed something up)
* can accept a implemented trait for assets, allowing more types of asset stores to be created

the above is subject to change because everything is broken.

Webview stuff is now called window.  `ApplicationExt` -> `Runtime`, `ApplicationDispatcherExt` -> `Dispatch`.

`Window` (prev. `WebviewDispatcher`) should now be the thing to pass to any hooks, it should be able to access the manager through an `Arc`, which houses all windows.  it's kinda convoluted because `Wry` only really gives you a dispatcher/proxy during rpc calls, limiting what can be in our own hooks. windows are also able to create other windows, which is why the link to the manager is needed.

good luck lol


examples/helloworld/src-tauri/src/main.rs
```diff
@@ -9,8 +9,11 @@ fn my_custom_command(argument: String) {
 }
 
 fn main() {
+  let ctx = tauri::generate_context!();
```
lucasfernog: we should get back to the old .build(tauri::generate_context!()) before merging

chippers:
i don't think the items and their position where they are exposed are final, but this should be ready to merge.  most of the changes i'm thinking about following this can be done separately.

let me know your thoughts on the organization. most of it focuses on putting things under a new "runtime" module, and having items mostly using/used by the runtime be under there.  Items that are likely to be used directly by users are closer to the root module, e.g. the new `mod hooks`.  Items likely to be used should probably continue to be exposed on the root module, but I didn't pay too much attention to that for right now

tauri/src/error.rs
```diff
@@ -46,6 +46,12 @@ pub enum Error {
   /// Invalid args when running a command.
   #[error("invalid args for command `{0}`: {1}")]
   InvalidArgs(&'static str, serde_json::Error),
+  /// Encountered an event string that didn't parse into a correct event
+  #[error("raw event string doesn't parse to an event type: {0}")]
+  InvalidEvent(String),
```
lucasfernog: this isn't used right now; are we going to use this event kind instead of a panic on event parse?

lucasfernog (approved):

tauri/src/error.rs
chippers: Probably not, we can just panic with an error. If someone has a custom event it will be their own error type that is panicked

chippers: For the docs about having your own event type, we should make it clear that they will want to have a variant for "unknown" ones because of the events both js and tauri itself can generate.  Even better is if we can expose all our events as an enum that meets the requirements of `Tag` and let them have a `Tauri(TauriEvent)` (or w/e name) variant for easy official tauri typings

PR #1429, opened 2021-04-05
refactor(tauri): remove app runner, use builder

chippers:

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No

**Other information:**
The "runner" was never really used.  Really the runner is the underlying runtime, wry by default. Instead of `AppBuilder::build(context).run().unwrap()` it's now just `tauri::Builder::run(context).unwrap()`.  all the builder setup/hooks/plugin methods are the same.

other projects use the same naming typically. e.g `hyper` which uses `Builder` struct as the builder to a `Server`, which is created with `Builder::run`. Similarly, we don't have any steps that can take place between `build()` and `run()`, so it's a bit redundant.

### DIFF
--- a/cli/core/Cargo.lock
+++ b/cli/core/Cargo.lock
@@ -50,9 +50,9 @@ checksum = "eab1c04a571841102f5345a8fc0f6bb3d31c315dec879b5c6e42e40ce7ffa34e"
 
 [[package]]
 name = "attohttpc"
-version = "0.16.3"
+version = "0.17.0"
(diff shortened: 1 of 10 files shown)


chippers:
this PR includes a fix for plugins not being passed to the `WindowManager` constructor

PR #1430, opened 2021-04-06
speed up asset inclusion on debug mode (fixes #1394)

### LINKED ISSUE #1394: cargo check, cargo build extremely slow even on development profile

chippers:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
---
This draft is going to be a workspace to keep track of different changes to speed up the build. The goal being as fast as possible asset inclusion for debug builds.

This first commit is mostly for release builds, but I'm curious how it handles on debug. Doing more work to prevent later work is pretty iffy for proc-macros because they are also compiled in debug mode for debug builds - which can make any fast (--release) computation work rather slow. The steps will go from trying to prevent excess work in the macro being done, to including uncompressed files, to finally reading straight from disk. If this last step is avoidable, I wish to avoid it.

Any compile times I post below is just the general wall clock time cargo tells us after running (only last step is the time):
```sh
cargo b --bin api
cargo clean -p api
cargo b --bin api
```

When this issue is ready to merge, it will fix #1394 

baseline from dev:
```
~/c/c/t/tauri ❯❯❯ cargo b --bin api
   Compiling api v0.1.0 (/home/chip/code/chippers/tauri-apps/tauri/examples/api/src-tauri)
    Finished dev [unoptimized + debuginfo] target(s) in 38.52s
```


chippers:
```
~/c/c/t/tauri ❯❯❯ cargo b --bin api
   Compiling api v0.1.0 (/home/chip/code/chippers/tauri-apps/tauri/examples/api/src-tauri)
    Finished dev [unoptimized + debuginfo] target(s) in 27.31s
```
~30% speedup for the latest commit, which added hashing.

It also injected `include_bytes!` for the compressed assets instead of generating them inline. This is where the extra performance came from. Downside: a build script is now always required to use the proc-macro. Slight upside: We recommend always having one anyways for other build generated stuff in `tauri-build`.

Next, I'm going to remove the hashing and see how much it affects performance.

lucasfernog:
> ```
> ~/c/c/t/tauri ❯❯❯ cargo b --bin api
>    Compiling api v0.1.0 (/home/chip/code/chippers/tauri-apps/tauri/examples/api/src-tauri)
>     Finished dev [unoptimized + debuginfo] target(s) in 27.31s
> ```
> 
> ~30% speedup for the latest commit, which added hashing.
> 
> It also injected `include_bytes!` for the compressed assets instead of generating them inline. This is where the extra performance came from. Downside: a build script is now always required to use the proc-macro. Slight upside: We recommend always having one anyways for other build generated stuff in `tauri-build`.
> 
> Next, I'm going to remove the hashing and see how much it affects performance.

Forcing a build script is indeed just fine.

chippers:
minimal speed changes
```
~/c/c/t/tauri ❯❯❯ cargo b --bin api
   Compiling api v0.1.0 (/home/chip/code/chippers/tauri-apps/tauri/examples/api/src-tauri)
    Finished dev [unoptimized + debuginfo] target(s) in 27.09s
```
looks like the blake3 hashing algorithm is still pretty fast in debug mode even for a moderate amount of assets. (api public dir is `308KiB`) I would probably like to keep that functionality around because it can cause pretty big wins for files that haven't changed content - especially larger ones. I'll revert the hash removal commit and experiment with removing compression in debug mode

chippers:
looks like we've probably hit the bottom as far as lowering build time from the assets. This new commit adds `DiskAssets` which just takes a base path, and then serves assets directly from the filesystem at runtime. the only code generated for it from the macro is `::tauri::api::assets::DiskAssets::from_dist_dir("/home/chip/code/chippers/tauri-apps/tauri/examples/api/public")` which should be pretty insignificant. I will probably revert 9c9248a because it could lead to assets getting out of sync between rust and frontend assets and it doesn't provide any speedups.

```
~/c/c/t/tauri ❯❯❯ cargo b --bin api
   Compiling api v0.1.0 (/home/chip/code/chippers/tauri-apps/tauri/examples/api/src-tauri)
    Finished dev [unoptimized + debuginfo] target(s) in 26.92s
```



next thing i will check is if spitting out the config to tokens has any effect on speed.  I'll replace it with a serde_json::from_str call with the json file.

chippers:
here's a measurement for a commit e601dc2a0c7f16670cb377b164a0661d4c1e298f im about to upload after i rebase. it replaces generating the config file into token literals and swaps it out with using serde_json to serialize it to a file, and then `serde_json::from_slice(include_bytes!("config.json"))`.

Seems overall slower, i will revert.
```
~/c/c/t/tauri ❯❯❯ cargo b --bin api
   Compiling api v0.1.0 (/home/chip/code/chippers/tauri-apps/tauri/examples/api/src-tauri)
    Finished dev [unoptimized + debuginfo] target(s) in 28.24s
```

chippers:
6724a52
```
~/c/c/t/tauri ❯❯❯ cargo b --bin api
   Compiling api v0.1.0 (/home/chip/code/chippers/tauri-apps/tauri/examples/api/src-tauri)
    Finished dev [unoptimized + debuginfo] target(s) in 26.08s
```

chippers:
looks like a vast majority of the remaining time is from linking. This is relatively expected because we a have a lot of transitive dependencies, and some of them are pretty large (e.g. `gtk` and its dependencies).  Here's a comparison with building with the pretty fast lld linker:
```
~/c/c/t/tauri ❯❯❯ cargo b --bin api
   Compiling api v0.1.0 (/home/chip/code/chippers/tauri-apps/tauri/examples/api/src-tauri)
    Finished dev [unoptimized + debuginfo] target(s) in 8.09s
```

Almost 70% speedup from our 30% speedup, or ~79% speedup from the start of the PR. I am assuming this is pretty close to the limit, but I'll need to profile with `-Zself-profile` to make sure it's still mostly linker time.

I don't know if it's appropriate for our repo to have the `.cargo/config` for it, but it should definitely be recommended to all users due to the sheer amount of linking that happens to the end binaries.

With this linker, changing a string in the api application and recompiling (not using `cargo clean -p api` in between), a rebuild is about 3.7s on my machine.

chippers:
changing the string without the linker is about ~14.8s (compared to lld's ~3.7s).

```
/c/c/t/tauri ❯❯❯ cargo b --bin api
   Compiling api v0.1.0 (/home/chip/code/chippers/tauri-apps/tauri/examples/api/src-tauri)
    Finished dev [unoptimized + debuginfo] target(s) in 14.79s
```

chippers:
| Code | Clean | String | Clean(lld) | String(lld) |
| ------ | ------ | ------ | ----------- | ---------- |
| `dev`| ~38.5s | ~24.7 | ~18.6s | ~13.8s |
| 1bdd789 | ~24.1s | ~14.8s | ~8s | ~3.7s |
| now | ~24.1s| ~14.8s | ~8s | ~3.7s |

seems like the change to how the assets are generated (inline -> `include_bytes!`) made the only realistic change. That change was pretty important though, because it was essentially additive to the time the linker needed.

The speedups with lld are the biggest wins. Dirty builds taking only ~30% of the time to build compared to `dev`. Clean 43% of the time to build. Without the linker, its still only 60% of the time to clean or dirty build compared to `dev`.

I'm going to add the cargo config script for the lld linker to the 3 example directories, because we want those to reflect how users should have their projects to setup. We should also make sure to add the lld linker information to the docs, as it can greatly increase debug iteration time.

After I add the configs, this PR is ready to merge, although a issue in the future might be created to track any potential wins from losing specific dependencies to help minimize linking time.

WilliamVenner:
Am I missing something here? I am very confused as to why compressing + embedding assets is necessary for debug builds.

lucasfernog:
> Am I missing something here? I am very confused as to why compressing + embedding assets is necessary for debug builds.

You may still need to share your debug builds with someone, so embedding is helpful.

Note that `tauri dev` will skip that if you're using a dev server.

lucasfernog (approved):

chippers:
> 
> 
> Am I missing something here? I am very confused as to why compressing + embedding assets is necessary for debug builds.

It doesn't actually take much of the build time (for api example at least). With how embedding was previously done, it could take some time cause it would spit the files out from the macro itself. Now that behavior is reverted to how it was before, where it writes out the compressed file to `OUT_DIR` and then uses `std::include_bytes!` on the filename. This is why having a build script is necessary again.

Because it's not taking much time the follow benefits are true:
1. Debug and Release builds are a lot closer in logic and functionality
2. Debug builds cannot serve files that have changed since the application was built, which could have the effect of the front-end updating but the back-end not being updated to accommodate it.
3. Being able to share debug builds.

Basically there were not any compilation wins I saw besides the original improvement, so there's more value in having less behavior changes between profiles.

See https://github.com/tauri-apps/tauri/pull/1430#issuecomment-813761039 for the differences between embedding and not embedding as far as rough compile time goes

PR #1443, opened 2021-04-07
refactor(tauri-api): remove unsafe usage in rpc callback generator

chippers:

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No (no release)

**Other information:**
Hi, sorry I didn't see #1370 before it was merged otherwise I would have brought the following points up (@WilliamVenner I'm tagging you so you can discuss this too).

1. Unsafe
     * #1370 introduces the first unsafe code to be directly used in `tauri`. Of course `wry` uses a lot of unsafe under the hood, but it's mostly self contained. We should probably be avoiding unsafe in `tauri` unless there is a good, measurable reason to use it.
    * IMO the introduced unsafe code seems relatively useless because the unsafely manipulated String (as a Vec) is then later used to concatenate into a new String allocation. I think effort is better spent to just build it normally and avoid `format!`.
    * I'm not against all unsafe code, but there should probably be a measurable benchmark against a safe alternative so that we can understand the benefits.
 
2. Minimum size to escape at
    * The gains of `JSON.parse` are only seen after a certain size, otherwise it's slower for small items. `10KiB` was just set arbitrarily from the [advice given in the blog post that introduced the practice.](https://v8.dev/blog/cost-of-javascript-2019#json).
    * Not only is it slower for small items on the JS side, but it introduces at a minimum 1 extra allocation if we decide to escape an item on the rust side. In the unsafe code it's two extra allocations minimum.
    * I think it could useful to make a benchmark to measure what this minimum threshold is if we wish to absolutely maximize performance of this function. Since we build for different targets, we might even be able to set different sizes for different `target_os`'s.
 
3. API exposure
    * I moved the constants and the escaping function out of the `pub` scope. I'm not sure if this was planned to expose, but I'm not sure of the use-case of calling either the constants or the escaping function outside of `fn format_callback`. If possible we should avoid putting it in pub scope so that we don't need to worry about backwards compatibility with them in the future.

Here's some notes of the specifics of this implementation change:
* `fn format_callback` now returns a `crate::Result<String>` instead of `String`. This is because constructing a `RawValue` is a bit of a pain, so the next easiest parameter to use is a `Serialize`.  We can just accept a `RawValue` if that's desired and leave it up to the caller. Also, it takes a reference of the serializable item now.
* `debug_assert!` for passing string literal too large to be represented as a JavaScript string literal
* I left 2 `todo` comments that I am interested to know your thoughts on

I tried to keep the spirit of minimizing allocations to allow this callback serialization to stay as fast as possible. It would be helpful to have benchmarks for this because I have no idea how much this specific serialization matters in the larger picture, and some data would be great.

### DIFF
--- a/tauri-api/src/rpc.rs
+++ b/tauri-api/src/rpc.rs
@@ -11,7 +11,7 @@ use serde_json::value::RawValue;
 const MAX_JSON_STR_LEN: usize = usize::pow(2, 30) - 2;
 
 /// Minimum size JSON needs to be in order to convert it to JSON.parse with [`escape_json_parse`].
-// todo: this number should be benchmarked and checked for optimal range, I set 10KiB conservatively
+// todo: this number should be benchmarked and checked for optimal range, I set 10 KiB arbitrarily
 // we don't want to lose the gained object parsing time to extra allocations preparing it
 const MIN_JSON_PARSE_LEN: usize = 10_240;
 
@@ -53,7 +53,8 @@ fn escape_json_parse(json: &RawValue) -> String {
 /// Formats a function name and argument to be evaluated as callback.
 ///
 /// This will serialize primitive JSON types (e.g. booleans, strings, numbers, etc.) as JavaScript literals,
-/// but will serialize arrays and objects whose serialized JSON string is smaller than 1 GB as `JSON.parse('...')`
+/// but will serialize arrays and objects whose serialized JSON string is smaller than 1 GB and larger
+/// than 10 KiB with `JSON.parse('...')`.
 /// https://github.com/GoogleChromeLabs/json-parse-benchmark
 ///
 /// # Examples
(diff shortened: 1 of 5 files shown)


tauri-api/src/rpc.rs
```diff
@@ -105,24 +100,36 @@ pub fn format_callback<T: Into<JsonValue>, S: AsRef<str>>(function_name: S, arg:
...
+  // from here we know json.len() > 1 because an empty string is not a valid json value.
+  let json = raw.get();
+  let first = json.as_bytes()[0];
 
-    // Explicitly drop json_value to avoid storing both the Rust "JSON" and serialized String JSON in memory twice, as <T: Display>.tostring() takes a reference.
-    drop(json_value);
+  #[cfg(debug_assertions)]
+  if first == b'"' {
+    debug_assert!(
+      json.len() < MAX_JSON_STR_LEN,
```
chippers, line 114:
```suggestion
      json.len() >= MAX_JSON_STR_LEN,
```

tauri-api/src/rpc.rs
```diff
@@ -111,7 +111,7 @@ pub fn format_callback<T: Serialize, S: AsRef<str>>(
   #[cfg(debug_assertions)]
   if first == b'"' {
     debug_assert!(
-      json.len() < MAX_JSON_STR_LEN,
+      json.len() >= MAX_JSON_STR_LEN,
```
chippers:
```suggestion
      json.len() < MAX_JSON_STR_LEN,
```

lucasfernog (approved):

PR #1452, opened 2021-04-11
feat(cli/tauri.js): download prebuilt cli

nklayman:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

Need to move binary-releases repo to Tauri Org.


cli/tauri.js/package.json
```diff
@@ -54,6 +54,7 @@
     "chalk": "4.1.0",
     "cross-spawn": "7.0.3",
     "fs-extra": "9.1.0",
+    "got": "^11.8.2",
```
lucasfernog: pin this

cli/tauri.js/src/helpers/download-cli.ts
```diff
@@ -0,0 +1,39 @@
...
+    platform = 'windows'
+  } else if (platform === 'linux') {
+    platform = 'linux'
+  } else if (platform === 'darwin') {
+    platform = 'macos'
+  } else {
+    throw Error('Unsupported platform')
+  }
+  const exe = platform === 'windows' ? '.exe' : ''
+  const url = `https://github.com/nklayman/tauri-binary-releases/releases/download/tauri-cli-v${version}/tauri-cli_${platform}${exe}`
```
lucasfernog: update to tauri-apps url

cli/tauri.js/src/helpers/download-cli.ts
```diff
@@ -0,0 +1,39 @@
+import stream from 'stream'
+import { promisify } from 'util'
+import fs from 'fs'
+import got from 'got'
+import { CargoManifest } from '../types/cargo'
+import path from 'path'
+const pipeline = promisify(stream.pipeline)
+const currentTauriCliVersion = (): string => {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-var-requires, @typescript-eslint/no-unsafe-member-access
+  const tauriCliManifest = require('../../../core/Cargo.toml') as CargoManifest
```
jbolda: I am pretty certain with how we bundle this that this comes in as a static value? Might be worth a comment in the code for future us.

cli/tauri.js/src/helpers/download-cli.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+const downloadCli = async (): Promise<void> => {
+  const version = currentTauriCliVersion()
+  let platform: string = process.platform
+  if (platform === 'win32') {
+    platform = 'windows'
+  } else if (platform === 'linux') {
+    platform = 'linux'
+  } else if (platform === 'darwin') {
```
jbolda: Do we need to do anything special for apple silicon?

cli/tauri.js/src/helpers/rust-cli.ts
```diff
@@ -42,6 +46,14 @@ export function runOnRustCli(
       process.cwd(),
       onClose
     )
+  } else if (process.env.NODE_ENV === 'production') {
```
jbolda: Since this gets bundled at publish, we need to confirm that we are running the build before publish with this flag turned on. I wonder if we make this more of the "default" and the other options can get logic that makes them dead code elims on publish.

cli/tauri.js/src/helpers/rust-cli.ts
```diff
@@ -42,6 +46,14 @@ export function runOnRustCli(
...
+    await downloadCli()
+    pid = spawn(
```
jbolda: Should we make the move to execa at this point? Feels like a good thing to take care of now.

cli/tauri.js/src/helpers/download-cli.ts
```diff
@@ -0,0 +1,39 @@
...
+    throw Error('Unsupported platform')
+  }
+  const exe = platform === 'windows' ? '.exe' : ''
+  const url = `https://github.com/nklayman/tauri-binary-releases/releases/download/tauri-cli-v${version}/tauri-cli_${platform}${exe}`
+  const outPath = path.join(__dirname, `../../bin/tauri-cli${exe}`)
+
+  console.log('Downloading Tauri CLI')
+  // TODO: Check hash of download
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access, security/detect-non-literal-fs-filename
+  await pipeline(got.stream(url), fs.createWriteStream(outPath))
```
jbolda: Can we no-op if it exists already?

jbolda (commented):
Looks good. Couple thoughts here. Also did you try running this on windows by chance? I know bins with `.exe` need some special considerations, but since we are running this with child_process, maybe we don't need to worry about that.

cli/tauri.js/src/helpers/download-cli.ts
nklayman: The above case will just run it without downloading if it exists already.

cli/tauri.js/src/helpers/download-cli.ts
nklayman: I'm not sure. I think it will automatically use Rosetta 2?

cli/tauri.js/src/helpers/download-cli.ts
nklayman: Yeah I need to move the repo first. I'm leaving it this way for now so it can be tested.

cli/tauri.js/package.json
nklayman: Done :+1:

cli/tauri.js/src/helpers/rust-cli.ts
nklayman: Oh yeah, the env isn't set properly. What is the best way to set it? Should I just use cross-env?

cli/tauri.js/src/helpers/download-cli.ts
nklayman: Yeah, there's really no reason to wrap this in a function either.

cli/tauri.js/src/helpers/rust-cli.ts
nklayman: Updated to use cross-env as I don't think there is a better solution.

cli/tauri.js/src/helpers/download-cli.ts
nothingismagick: yeah, merely using rosetta is kind of cheap. will work, but in the long term its not a proper solution.

cli/tauri.js/src/helpers/rust-cli.ts
nklayman: If it aint broke, don't fix it? I don't see much of a reason to switch to execa since it all seems to work fine now, and it would probably best to have it as a separate PR as it isn't really related to this PR.

jbolda: 👍

lucasfernog:
So, is this still WIP?

cli/tauri.js/src/helpers/download-cli.ts
nklayman: Updated to use the new repo now :+1:

lucasfernog (approved):

PR #1454, opened 2021-04-11
refactor(tauri): expose all items meant to be public

chippers:

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No

**Other information:**
Some small breaking changes because some item exports have moved.

If the item is meant to be used by the user, it is now exposed under the root.

Items meant to be implemented by a webview runtime are staying in `mod runtime`.

Note that some of the code for items in the first category are still under the runtime module because they are closely related code-wise. They are not exported from there however.

Specifically for the `mod sealed` now under `lib.rs` - not the final resting place for the private implementations. I have some ideas in the future to make this nicer, but for now it's okay


core/tauri/src/lib.rs
```diff
@@ -14,7 +14,7 @@
 pub use error::Error;
 pub use tauri_api as api;
 pub(crate) use tauri_api::private::async_runtime;
-pub use tauri_macros::*;
+pub use tauri_macros::{command, generate_handler};
```
lucasfernog, line 17: the lib.rs file is pretty big now - but i guess it's ok like this

lucasfernog (approved):

core/tauri/src/lib.rs
chippers: yeah, im hoping in the future to refactor the sealed part somewhere else which would make it look a lot more manageable, but for now this works fine

PR #1463, opened 2021-04-12
refactor: more consistent js-api docs

amrbashir:

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [X] Code style update
- [X] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No

**Other information:**

I moved out the event logic from the helper into the main event file because it only re-exported `emit` with `undefined` windowLabel which I don't understand why? I think it is ok to let users send events from one window to another as they please.


nothingismagick (commented):
I can't speak to how this will be rendered (will let @Laegel approve or not), but the work looks great. One question though - why did you drop `@description` everywhere - I thought that was important... no?

amrbashir:
Because it looked better without `@description`
here is a function without

nothingismagick:
> I moved out the event logic from the helper into the main event file because it only re-exported emit with undefined windowLabel which I don't understand why? I think it is ok to let users send events from one window to another as they please.

@lucasfernog is this ok?

amrbashir:
Because it looked better without `@description` 

Here is a function without `@description`

image


And here is a function with `@description`
image

nothingismagick:
> Because it looked better without `@description`
> 
> Here is a function without `@description`
> 
> image
> 
> And here is a function with `@description`
> image

agreed

lucasfernog:
The helper event.ts file is like a private internal API wrapper. Users should either use event.ts for global events, or webview.ts to send events to any window they like. We do it this way because the high level APIs are much easier to handle than passing labels to a function.

amrbashir:
I see, I will revert it back quickly.

amrbashir:
btw @Laegel using `@ignore` or `@hidden` or `@internal` will still include an entry in the sidebar, to avoid that we need to pass `--exclude path/to/helpers/`

Laegel:
Alright, thanks @amrbashir ! Going to review right now.

Laegel:
So, JFYI, we could have avoided writing types in the annotations as Typedoc is clever enough to fetch them from the code.

E.g.
```ts
/**
 * Exits immediately with the given `exitCode`
 *
 * @param {number} [exitCode=0] The exit code to use
 * @returns {Promise<void>}
 */
async function exit(exitCode: number = 0): Promise<void> {
  return invokeTauriCommand({
    __tauriModule: 'App',
    mainThread: true,
    message: {
      cmd: 'exit',
      exitCode
    }
  })
}
 ```
could have been written:
```ts
/**
 * Exits immediately with the given `exitCode`
 *
 * @param The exit code to use
 * @returns
 */
async function exit(exitCode: number = 0): Promise<void> {
  return invokeTauriCommand({
    __tauriModule: 'App',
    mainThread: true,
    message: {
      cmd: 'exit',
      exitCode
    }
  })
}
```

I need to patch the Typedoc action since the folder has been moved to the "tooling" folder.

amrbashir:
Yeah I was going that direction but mid-refactoring I saw default values and thought It needs types so I added types to all for consistency. 

I will remove them.

Laegel (approved):

tooling/api/src/fs.ts
```diff
@@ -167,12 +163,9 @@ function arrayBufferToBase64(buffer: ArrayBuffer): string {
...
  * Writes a binary file
  *
- * @param {FsBinaryFileOption} file
- * @param {string} file.path Path of the file
- * @param {ArrayBuffer} file.contents Contents of the file
- * @param {FsOptions} [options] Configuration object
- * @param {BaseDirectory} [options.dir] Base directory
- * @returns {Promise<void>}
+ * @param file File configuration object
+ * @param [options]
```
Laegel, line 167: What's the `[options]` for?

amrbashir: in the `fs` module, there is a lot of [options] which are general configuraion object. I think I skipped those because typedoc doesn't seem to generate description for optional args. 

let me know and I will add a description to them.

Issue #1512, opened 2021-04-15
Cannot import @tauri-apps/api/tauri


Pierstoval:
**Describe the bug**
Cannot import `@tauri-apps/api/tauri` from a `<script lang="ts">` tag when frontend is build with Svelte kit and latest Tauri version.

The project was created **today**, with a totally fresh Tauri install (using the new beta version), and the frontend is a plain Svelte Kit application with one single change: instead of using the `node` bundle adapter, I'm using the `static` one, in order to always build an HTML/CSS app instead of a Node.js one, because Tauri needs a static app.

What I intend to do is simply use the `invoke` function from the Tauri API in order to communicate with the Rust backend (and I just want to return a string).

This is sort of a "test app", so nothing complex at all in here, hence why I create an issue, because I think a freshly bootstrapped app like that shouldn't have bugs, and I don't know if the issue comes from me or Tauri or Svelte.

**To Reproduce**
1. Clone the `rewrite` branch of this repo: https://github.com/Orbitale/Compotes/tree/rewrite (NOTE in 2025: it was years ago, this is no longer relevant, but I just keep the link for history reasons)
2. Install everything with `yarn install`, then run `yarn dev` to start the frontend server
3. Start the Tauri app with `yarn tauri dev` in another terminal instance
4. The app should automatically rebuild, and you should see this **frontend** error:
  <details><summary>(click to display stack trace 👇)</summary>

  ```
  Cannot find module '@tauri-apps/api/tauri' from '/var/www/Compotes2/src/routes'

  Error: Cannot find module '@tauri-apps/api/tauri' from '/var/www/Compotes2/src/routes'
      at Function.resolveSync [as sync] (/var/www/Compotes2/node_modules/resolve/lib/sync.js:102:15)
      at resolveFrom$2 (/var/www/Compotes2/node_modules/vite/dist/node/chunks/dep-66eb515d.js:3884:29)
      at resolve (/var/www/Compotes2/node_modules/vite/dist/node/chunks/dep-66eb515d.js:69065:22)
      at nodeRequire (/var/www/Compotes2/node_modules/vite/dist/node/chunks/dep-66eb515d.js:69044:25)
      at ssrImport (/var/www/Compotes2/node_modules/vite/dist/node/chunks/dep-66eb515d.js:69002:20)
      at eval (/var/www/Compotes2/src/routes/index.svelte:7:31)
[...]

PR #1517, opened 2021-04-16
fix: #1512

Laegel:

- [x] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes. Issue #1512 
- [ ] No

**Other information:**
I've made my tests by running and publishing to a [local Verdaccio](https://verdaccio.org/docs/en/installation#docker-image).

How the package looks like in node_modules:
node_modules

An extract of my code (made with an old Vue CLI and Vue 2):
tauri-client

Hovering the `invoke` function displays the types as expected:
typings

Result browser-side:
vue

With the Compotes project (see #1512):
compotes


lucasfernog (approved):

nothingismagick (approved):

nothingismagick:
Question: is there ever a reason for us to ship the TS?

Laegel:
> Question: is there ever a reason for us to ship the TS?

Not really, that's why we only have type definitions. :)

tooling/api/package.json
```diff
@@ -2,33 +2,30 @@
...
+    "./globalShortcut": "./globalShortcut.js"
   },
   "funding": {
     "type": "opencollective",
     "url": "https://opencollective.com/tauri"
   },
   "scripts": {
     "build": "rimraf ./dist && rollup -c --silent",
-    "prepublishOnly": "yarn build",
+    "preparePublish": "yarn build && cp package.json dist/ && cd dist && npm publish",
```
amrbashir: I am not quite familiar with covector workflow but shouldn't covector be the one to `cd dist && npm publish`? also I think covector config needs to be updated with the new `dist` path.

Laegel: I think you're right, this was a POC and should probably be adapted to our custom process.

chippers: We should probably replace the `&&` steps with something like `run-s` from dep `npm-run-all` because not all versions of powershell understand double ampersands to mean "if okay then". (thank you windows)

Additionally, to support windows, we should probably make the steps to copy the file over to dist into a `publishDistBuild.js` file and use the node way of copying over the file. I think this could solve:
1. the `exports` key doesn't need to be as expressive as it is currently.  I think we mostly need a `"main"` key.
2. since we are exporting ESM, we should be adding `"type": "module",` to the package.json keys

chippers: we also really need to copy over the license file too so that the final package contains our licensing

nothingismagick:
But lets merge and release

tooling/api/package.json
```diff
@@ -2,33 +2,31 @@
...
+    "./globalShortcut": "./globalShortcut.js"
   },
   "funding": {
     "type": "opencollective",
     "url": "https://opencollective.com/tauri"
   },
   "scripts": {
     "build": "rimraf ./dist && rollup -c --silent",
-    "prepublishOnly": "yarn build",
+    "preparePublish": "yarn build && copyfiles package.json LICENSE* CHNAGELOG.md dist",
```
Laegel:
```suggestion
    "preparePublish": "yarn build && copyfiles package.json LICENSE* CHANGELOG.md dist",
```

nothingismagick (approved):

.changes/config.json
```diff
@@ -144,9 +144,66 @@
       "manager": "javascript",
       "assets": [
         {
-          "path": "./tooling/api/tauri-apps-api-${ pkgFile.version }.tgz",
+          "path": "./tooling/api/dist/tauri-apps-api-${ pkgFile.version }.tgz",
```
jbolda, line 147: `npm pack` is going to place the tar here, but I think when we `npm publish`, that this will be included. We probably want to move it or have npm ignore it.

.changes/config.json
```diff
@@ -144,9 +144,66 @@
...
+          "command": "echo \"\\`\\`\\`\n\n</details>\n\"",
+          "dryRunCommand": true,
+          "pipe": true
+        },
+        {
+          "command": "yarn build",
+          "dryRunCommand": true
+        },
+        {
+          "command": "cd ./dist && npm pack",
```
jbolda: I just confirmed that this won't work. There was an issue upstream where the args aren't processed correctly so that is run as one single command. We need to `shell: true`, and I fixed it temporarily by searching for `||`, but I didn't consider `&&`. Can we move all of this into the `package.json` script?

amrbashir: can I just add another command before it that changes the directory instead of combining them?

jbolda (approved):

PR #1551, opened 2021-04-20
Make sure custom protocol is handled as secure context on macOS

### LINKED ISSUE #1550: Treat macOS custom protocol as https for WebRTC
**Describe the bug**
macOS custom protocol should be treated as HTTPS.

**To Reproduce**
N/A

**Expected behavior**
WebRTC should be enabled on custom protocol when permissions are granted.

**Additional context**
Linked with https://github.com/tauri-apps/wry/issues/85#issuecomment-823499749

`localhost` should be used as a custom protocol host in macOS. Instead of using `tauri://` we should use `tauri://localhost/` as a base.

lemarier:

- [x] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

Do not seem to affect other platforms need to test on Windows and Linux to be sure it didn’t break anything.

Fix #1550
Related to https://github.com/tauri-apps/wry/issues/85

### Without the fix:
Screen Shot 2021-04-20 at 3 11 15 PM

***

### With the fix
Screen Shot 2021-04-20 at 3 10 14 PM

### DIFF
--- a/.changes/macos-custom-protocol-https.md
+++ b/.changes/macos-custom-protocol-https.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch
+---
+
+Make sure custom protocol is treated as secure content on macOS.
--- a/core/tauri/src/runtime/manager.rs
+++ b/core/tauri/src/runtime/manager.rs
@@ -369,7 +369,7 @@ mod test {
     );
 
     #[cfg(custom_protocol)]
-    assert_eq!(manager.get_url(), "tauri://studio.tauri.example");
+    assert_eq!(manager.get_url(), "tauri://localhost");
 
     #[cfg(dev)]
     assert_eq!(manager.get_url(), manager.config().build.dev_path);
--- a/core/tauri/src/runtime/manager.rs
+++ b/core/tauri/src/runtime/manager.rs
@@ -129,13 +129,13 @@ impl<P: Params> WindowManager<P> {
     if self.inner.config.build.dev_path.starts_with("http") {
       self.inner.config.build.dev_path.clone()
     } else {
-      format!("tauri://{}", self.inner.config.tauri.bundle.identifier)
+      "tauri://localhost".into()
     }
   }
 
   #[cfg(custom_protocol)]
   fn get_url(&self) -> String {
-    format!("tauri://{}", self.inner.config.tauri.bundle.identifier)
+    "tauri://localhost".into()
   }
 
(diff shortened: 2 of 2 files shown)


nothingismagick (approved):
If there were a flair possible for this kind of PR, then I would grant it legendary status that is literally a game changer. Thanks man. I owe you a caipi. On the Seychelles

lucasfernog:
I'm checking how local storage behaves with this solution. I think the webview caches it based on the url :/

lemarier:
> I'm checking how local storage behaves with this solution. I think the webview caches it based on the url :/

maybe we could make sure local storage use `app` and not `tauri-shared` directory for storage.

lucasfernog:
> > I'm checking how local storage behaves with this solution. I think the webview caches it based on the url :/
> 
> maybe we could make sure local storage use `app` and not `tauri-shared` directory for storage.

That would be great

lemarier:
> That would be great

### macOS

It's stored inside `~/Library/WebKit` and uses the `package.id`

Screen Shot 2021-04-20 at 4 33 59 PM

### Windows

We already use [with_user_data_folder](https://github.com/tauri-apps/wry/pull/120) so not a problem

### Linux

Need to check this out

lucasfernog:
Linux might need this: https://webkitgtk.org/reference/webkit2gtk/stable/WebKitWebsiteDataManager.html#WebKitWebsiteDataManager--local-storage-directory

lucasfernog:
webkit stores localstorage under `$HOME/.local/share/webkitgtk/localstorage`, so we do need to change it. I'll PR wry.

lucasfernog:
Unfortunately `navigator.mediaDevices` is still undefined on Linux :( we probably need to tweak something on the wry end.

lemarier:
> Unfortunately `navigator.mediaDevices` is still undefined on Linux :( we probably need to tweak something on the wry end.

ok I'll try to make some tests to see what we can do

lucasfernog:
Since this is a macOS PR i'll merge it.

wusyong:
I think we haven't enabled it yet but the method/signal should be here:
https://webkitgtk.org/reference/webkit2gtk/stable/WebKitWebView.html#WebKitWebView-permission-request

PR #1582, opened 2021-04-22
fix(api): distribution

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No

**Other information:**


nothingismagick:
Don't we have to ship the typings? And can we please ship the .ts as well?

lucasfernog:
> Don't we have to ship the typings? And can we please ship the .ts as well?

We are shipping the typings on the dist folder. I'll copy the ts files.

tooling/api/package.json
```diff
@@ -2,30 +2,71 @@
...
+      "require": "./tauri.cjs"
+    }
   },
   "funding": {
     "type": "opencollective",
     "url": "https://opencollective.com/tauri"
   },
   "scripts": {
-    "build": "rimraf ./dist && rollup -c --silent && copyfiles package.json LICENSE* CHANGELOG.md dist",
+    "build": "rimraf ./dist && rollup -c --silent && copyfiles package.json LICENSE* CHANGELOG.md src/*.ts src/**/*.ts dist",
```
amrbashir:
```suggestion
    "build": "rimraf ./dist && rollup -c --silent && copyfiles -f package.json LICENSE* CHANGELOG.md src/*.ts dist && copyfiles -f src/helpers/*.ts dist/helpers",
```

amrbashir (approved):

tooling/api/package.json
lucasfernog: why?

amrbashir: passing the `-f` flag to flatten the `src` directoy into `dist`, otherwise it will create a `src` directory inside `dist`.
and copying the files of `helpers` separately because we don't want them to be in `dist` but in `dist/helpers`

lucasfernog: I guess we should stick with the /dist/src folder, but i'd like to know if this is what @nothingismagick wanted

amrbashir: this how it is built with your changes
image

lucasfernog: I think having 6 files for each api, one with a different extension (.js, .d.ts, .ts, .mjs) would confuse consumers :joy: if you're crazy enough to use the typescript source directly, just use the src folder

amrbashir: if we ship `dist/src` then any bundler will just use `*.d.ts` because it won't find any `*.ts` in the dist directory and it has no idea there is a `src` folder

lucasfernog: I wonder what kind of bundler config would pick up the typescript files over the .js file

PR #1615, opened 2021-04-25
feat: setup testing for CTA

jbolda:
This begins to implement a testing setup for CTA. It may also serve as an e2e test.

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

### DIFF
--- a/tooling/create-tauri-app/test/index.spec.ts
+++ b/tooling/create-tauri-app/test/index.spec.ts
@@ -6,86 +6,134 @@ import execa from 'execa'
 import fixtures from 'fixturez'
 const f = fixtures(__dirname)
 import path from 'path'
+import fs from 'fs'
 
 const ctaBinary = path.resolve('./bin/create-tauri-app.js')
 const clijs = path.resolve('../cli.js/')
 const api = path.resolve('../api/')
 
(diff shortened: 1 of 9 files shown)


tooling/create-tauri-app/test/index.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+            },
+            reactjs: () => {
+              expect(packageFileOutput['scripts']).toMatchObject({
+                start: 'react-scripts start',
+                build: 'react-scripts build',
+                test: 'react-scripts test',
+                eject: 'react-scripts eject',
+                tauri: 'tauri'
+              })
+            },
```
amrbashir: why not just check for `tauri` since that's the only thing we are adding and it makes sure the test won't fail if `create-react-app` decides to change its scripts

jbolda: They haven't changed the scripts for years so I wasn't worried. Now that refactored the code a bit and made this easier to do, it probably makes sense to do now just in case.

PR #1622, opened 2021-04-26
feat(core): support generics (especially Param) in #[command]

chippers:

- [x] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
In `#[command(with_window)]` commands, you could not specify a type for `Event` or `Label`, making any function that uses those associated types unusable from within that command function. Without those associated types specified, the developer will run into errors like `"Expected type Params::Event but received struct String"` when using things like `window.emit(...)`. Previously, the following code could not work:
```rust
#[command(with_window)]
pub fn import_csv(
  window: tauri::Window<impl tauri::Params<Event = String>>,
  event: String,
  payload: Option<String>,
) -> Result<(), String> {
  window
    .emit(&"output".into(), Some("test".to_owned()))
    .expect("failed to emit");

  return Ok(());
}
```

This is because the `import_csv_wrapper` function generated by `#[command]` would always use `<P: ::tauri::Params>` as the generic.

With this first commit in this draft, the above code works because the following is now true:
*  The generic parameter is taken from the generic parameter used inside the first argument's type.
    * It expects the first parameter to be a `Type::Path` (like a struct). (unlikely to change)
    * That `Type::Path` must have exactly 1 generic parameter. (unlikely to change)
    * That 1 generic parameter **MUST** be an `impl Trait` of `impl tauri::Params`. **This will change before the PR is merged**. This is a side effect of testing the core changes before I flesh out support for all ways of specifying generics. The following will be supported:
        * `fn my_command<P: Params<Event = String>>(window: Window<P>)`
        * `fn my_command<P>(window: Window<P>) where P: Params<Event = String>`
        * `fn my_command(window: Window<impl Params<Event = String>>)`
        * any missing?

**Note:** The code is pretty shaky and only accepts a happy path. This was done to test the proof of concept and to make sure it actually solves what it set out to do. At this point, I am relatively confident that this approach will work and the code will be refactored

---

Tasks to do before the PR is merged:
1. [x] Refactor the error handling of the `tauri-macros` crate to allow better readability with full error handling
2. [x] Accept all ways of specifying a trait and its associated types (or at least the 3 above).
3. [ ] use `static_assertions` and `impls` crate for compile-time detection of types/traits, so that we can spit out more specific errors if the wrong struct or trait is input into `window: Window<P>`


chippers:
This is still marked as a draft because there is still [1 example command that doesn't work with type inference without changes](https://github.com/tauri-apps/tauri/pull/1622/files#diff-c2960c42ff4a9fc5ae7158aa2f622bfcc5ef13494d15e5fb4e48dd3d4b89dabc). Ideally I want to support it, so I will look into ways to help inference it which may or may not involve small changes to `CommandArg`.

command that doesn't work: https://github.com/tauri-apps/tauri/blob/a6baec55a261d520c09cefbe031cd230af9fb801/examples/splashscreen/src-tauri/src/main.rs#L54-L63
```
error[E0282]: type annotations needed
  --> examples/splashscreen/src-tauri/src/main.rs:54:3
   |
54 |   #[tauri::command]
   |   ^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `P` declared on the function `close_splashscreen`
...
77 |       .invoke_handler(tauri::generate_handler![close_splashscreen])
   |                       -------------------------------------------- in this macro invocation
   |
   = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to previous error
```

It's not able to tell that the `P` in `State` is the same as the `P` in `CommandItem`.  Ideally this is something I can fix on the `impl CommandArg for State` but might need to tweak the trait slightly if not.

**Note:** that if you add a type that can successfully inference, then the error goes away e.g.  `_window: tauri::Window<P>`. 

I'm also curious if we could implement `CommandArg` for `Params` which just returns a `()` to make it an easy hack to add inference. `_: P`

chippers:
I solved the type inference by making a `StateP` which adds a `P: Params` generic parameter. Not sure if this should replace `State` instead and always require the `P: Params` so I've kept this marked as a draft until a decision is made.

edit: It probably shouldn't be replaced, so the draft question is if we should keep the `StateP` helper

chippers:
`StateP` has been removed.

The following command function arguments now work:
* `Pat::Wild` (e.g. `_: Window<P>`)
* `Pat::Struct` (e.g. `Person { name, age }: Person` where the key is the last path segment `"Person"`)
* `Pat::TupleStruct`  (e.g. `InlinePerson(name, age): InlinePerson` where the key is the last path segment `"InlinePerson"`)

lucasfernog (commented):
I wanna kiss this code; can you just add the change file?

lucasfernog:
Also, we'll need to work on documentation (e.g. the `_: Window<P>` trick).

lucasfernog (approved):

PR #1623, opened 2021-04-26
refactor(core): simplify usage of app event and window label types

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No

**Other information:**

### DIFF
--- a/core/tauri/src/endpoints/window.rs
+++ b/core/tauri/src/endpoints/window.rs
@@ -7,9 +7,9 @@ use crate::Manager;
 use crate::{api::config::WindowConfig, endpoints::InvokeResponse, Params, Window};
 use serde::Deserialize;
 
+use crate::runtime::manager::tauri_event;
 use crate::Icon;
 use std::path::PathBuf;
-use crate::runtime::manager::tauri_event;
 
 #[derive(Deserialize)]
 #[serde(untagged)]
--- a/core/tauri/src/endpoints/window.rs
+++ b/core/tauri/src/endpoints/window.rs
@@ -9,6 +9,7 @@ use serde::Deserialize;
 
 use crate::Icon;
 use std::path::PathBuf;
+use crate::runtime::manager::tauri_event;
 
 #[derive(Deserialize)]
 #[serde(untagged)]
@@ -125,16 +126,8 @@ impl Cmd {
           let pending =
             crate::runtime::window::PendingWindow::with_config(options, label.clone(), url);
 
-          let raw_event = "tauri://window-created";
-          let event: M::Event = raw_event.parse().unwrap_or_else(|_| {
-            panic!(
-              "Window module failed to parse tauri event into Params::Event: {}",
-              raw_event,
-            )
-          });
-
           window.create_window(pending)?.emit_others(
-            &event,
+            &tauri_event("tauri://window-created"),
             Some(WindowCreatedEvent {
               label: label.to_string(),
             }),
--- a/core/tauri/src/endpoints/window.rs
+++ b/core/tauri/src/endpoints/window.rs
@@ -100,7 +100,7 @@ struct WindowCreatedEvent {
 
 impl Cmd {
   #[allow(dead_code)]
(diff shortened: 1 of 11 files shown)


core/tauri/src/endpoints/window.rs
```diff
@@ -124,13 +124,23 @@ impl Cmd {
           let url = options.url.clone();
           let pending =
             crate::runtime::window::PendingWindow::with_config(options, label.clone(), url);
-          window.create_window(pending)?.emit_others_internal(
-            "tauri://window-created".to_string(),
+
+          let raw_event = "tauri://window-created";
+          let event: M::Event = raw_event.parse().unwrap_or_else(|_| {
```
lucasfernog: should this use `tauri_event` fn?

chippers: ah yeah, i made that before i did that helper fn

chippers:
0bfd525 changes all the interfaces that **might** not need an owned tag into a new trait, `TagRef`. Some additional bounds are also required when wanting to use it as a key into an internal hashmap e.g. `P::Event: Borrow<E>`.

If an item is going to be stored (like an Event or Label tag in an event Listener), we accept a `Into<Event>`.

I think that commit covers all the event stuff, I'll double check to make sure there's no window labels lying around in the API that need the same treatment.

chippers:
Looks like `Labels` are already mostly handled internally and I already updated the once place where it's passed into a `Manager`.

This PR is probably good once the tests pass.  I don't think this is actually a breaking change, because it previously accepted `&String` because `P::Event` was `String` and it required `&P::Event`. The new bounds still accept a `&String`.

edit: nope, `.emit(&"event".into(), None)` is breaking, solution: `.emit("event", None)`

lucasfernog:
@chippers any chance we can also accept owned values where `&Event` or `&Label` is being used (e.g. get_window or emit)? looks just a little weird writing `.emit(&MyEvent::Kind)` :(

I tried and it seems like it's not easy or even possible :/

chippers:
> 
> 
> @chippers any chance we can also accept owned values where `&Event` or `&Label` is being used (e.g. get_window or emit)? looks just a little weird writing `.emit(&MyEvent::Kind)` :(
> 
> I tried and it seems like it's not easy or even possible :/

I'll look into it tonight or tomorrow.  It would look less weird if the event or label was already stored in a variable.  I chose to take the reference because those items don't necessarily clone, so this is the only way to absolutely prevent an allocation by the end user.

imo immediately referencing a newly created value isn't too weird anyways. I'll still take a look at possibilities tonight/tomorrow

lucasfernog:
> > @chippers any chance we can also accept owned values where `&Event` or `&Label` is being used (e.g. get_window or emit)? looks just a little weird writing `.emit(&MyEvent::Kind)` :(
> > I tried and it seems like it's not easy or even possible :/
> 
> I'll look into it tonight or tomorrow. It would look less weird if the event or label was already stored in a variable. I chose to take the reference because those items don't necessarily clone, so this is the only way to absolutely prevent an allocation by the end user.
> 
> imo immediately referencing a newly created value isn't too weird anyways. I'll still take a look at possibilities tonight/tomorrow

alright, don't waste too much time on it :P

chippers:
In short, we can allow `MyEvent::Foo` and `&MyEvent::Foo` with the downsides of:
1. A `Cow` is created for every call
2. `String`, `&String` can no longer be used without turbofishing

Since allowing `String`, `&String`, and `&str` to be used directly is more important, no additional changes will be made. See `HashMap::get` for a similar api mechanics to the one we are exposing now.
```rust
let map: HashMap<MyEvent, String> = HashMap::new();
map.get(&MyEvent::Foo);
```
is similar to
```rust
window.emit(&MyEvent::Foo, None);
```

This PR is ready if there are no more reviews

PR #1652, opened 2021-04-29
Update App.svelte

jswhisperer:
make it work

- [x] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

### DIFF
--- a/examples/api/src/App.svelte
+++ b/examples/api/src/App.svelte
@@ -1,6 +1,6 @@
 <script>
   import { onMount } from "svelte";
-  import { open } from "@tauri-apps/api/shell";
+  import { open } from "@tauri-apps/open";
 
   import Welcome from "./components/Welcome.svelte";
   import Cli from "./components/Cli.svelte";


examples/api/src/App.svelte
```diff
@@ -1,6 +1,6 @@
 <script>
   import { onMount } from "svelte";
-  import { open } from "@tauri-apps/api/shell";
+  import { open } from "@tauri-apps/open";
```
lucasfernog, line 3: how did this work? that's not a package name

lucasfernog (changes requested):

lucasfernog:
Maybe it didn't work because you didn't build the `tooling/api` package? Try following the contribution guide or just running `yarn build` on that folder. On my end it works just fine.

examples/api/src/App.svelte
jswhisperer: Apologies, any kind of instructions to run the /examples would be very advantageous. So far I've figured out you need to use yarn over npm because of the "link" property. It seems you are saying build should be run before dev, that isnt typical in javascript projects would be great to add that also. Thanks for your valuable time sir.

lucasfernog: Build should be run on the `api` npm package, not the api example. but yeah i agree it needs some instructions.

nothingismagick:
> Apologies, any kind of instructions to run the /examples would be very advantageous. So far I've figured out you need to use yarn over npm because of the "link" property. It seems you are saying build should be run before dev, that isnt typical in javascript projects would be great to add that also. Thanks for your valuable time sir.

@gregpalaci - thanks for your concern! Please remember, that while a lot of the bits and pieces of tauri work right now, we are still ironing out a great deal and are focussed on the code, not the documentation. We will do a documentation push once we have placed the beta code in the hands of the external auditors.

PR #1665, opened 2021-04-30
feat(core): add state management, closes #1655

### LINKED ISSUE #1655: Implement state for command handlers
The current API for commands forces users to use global states because we only pass the `Window` and the invoke arguments (from JS) to the command function. We should manage state for commands.

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No

**Other information:**


core/tauri-macros/src/command.rs
```diff
@@ -40,40 +59,55 @@ pub fn generate_command(attrs: Vec<NestedMeta>, function: ItemFn) -> TokenStream
...
+    let arg_type =
+      arg_type.unwrap_or_else(|| panic!("Invalid type for arg \"{}\"", arg_name.unwrap()));
+
+    let mut path_as_string = String::new();
+    for segment in &arg_type.segments {
+      path_as_string.push_str(&segment.ident.to_string());
+      path_as_string.push_str("::");
+    }
+
+    if ["State::", "tauri::State::"].contains(&path_as_string.as_str()) {
```
chippers: This is problematic because any renamed import will fail. This is also the reason why we went with `with_window` instead of trying to detect the window parameter. Since we can't type detect during the macro, there isn't a name-based way to do this. I think the following choices might be viable:
1. add `with_state` and operate the `with`s in a left to right order
2. create a `FromCommand` trait and have parameters implement it `from_command(msg: InvokeMessage<_>) -> Result<Self>`  (this is similar to what rocket does).  We then use type inference to call that trait method on all the parameters being passed to the function
    * because the "data" parameters aren't name anywhere else, this means we would only realistically be able to accept struct that implment it also, and not use something like `name: String`
3. (maybe, i haven't tested this) accept a `#[window]` and `#[state]` attributes in the parameters, something like the following:
    * ```rust
      #[command]
      fn my_command(
        #[window] window: Window<impl Params>,
        #[state] state: State<'_, String>,
        name: String,
        body: String
      ) {
        todo!()
      }
      ```

ill probably discuss it on discord

chippers (dismissed):
It looks alright, except for how we pass the state into the command parameters.  I have some ideas that we can discuss on discord

chippers:
I marked this as a draft because I haven't tested it personally (compiles + local tests pass) and it still needs a bit more work. I'll get around to it sometime tomorrow unless someone gets to it first

1. Docs about the `FromCommand` trait
2. remove `with_window` attribute from everywhere since it's completely unread
3. test it actually works with real applications lol
4. changefile
5. am i missing anything?

chippers:
Look good to me :)

PR #1683, opened 2021-05-03
refactor(core): merge invoke items into single struct, allow ?

chippers:

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No (no release since the breaking change that this refactors)

**Other information:**
Not sure if `InvokeError` and `InvokeResponse` is the final abstraction to use - but we can finalize that after we get command returns using a trait. I undid most of the `InvokeResponse` interface changes so that we can use `?` inside of the closures. I added `InvokeError` to have an easy to use error type for it and `CommandArg`.  We still convert to an `InvokeResponse` before formatting the callback so that we can parse the value into a json error instead of failing parsing altogether

### DIFF
--- a/core/tauri/src/plugin.rs
+++ b/core/tauri/src/plugin.rs
@@ -4,11 +4,7 @@
 
 //! Extend Tauri functionality.
 
-use crate::{
-  api::config::PluginConfig,
-  hooks::{InvokeMessage, InvokeResolver, PageLoadPayload},
-  App, Invoke, Params, Window,
-};
+use crate::{api::config::PluginConfig, App, Invoke, PageLoadPayload, Params, Window};
 use serde_json::Value as JsonValue;
 use std::collections::HashMap;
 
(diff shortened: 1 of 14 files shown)


core/tauri/src/plugin.rs
```diff
@@ -37,40 +37,40 @@ pub trait Plugin<M: Params>: Send {
...
 
   /// Callback invoked when the webview performs a navigation.
   #[allow(unused_variables)]
-  fn on_page_load(&mut self, window: Window<M>, payload: PageLoadPayload) {}
+  fn on_page_load(&mut self, window: Window<P>, payload: PageLoadPayload) {}
 
   /// Add invoke_handler API extension commands.
   #[allow(unused_variables)]
-  fn extend_api(&mut self, message: InvokeMessage<M>, resolver: InvokeResolver<M>) {}
+  fn extend_api(&mut self, message: InvokeMessage<P>, resolver: InvokeResolver<P>) {}
```
lucasfernog: should this take `Invoke` instead?

chippers: Yes, i missed that

lucasfernog (approved):

PR #1734, opened 2021-05-06
feat(core): #[command] return with autoref specialization workaround fix #1672 

### LINKED ISSUE #1672: Accept a trait in the return of a `#[command]`
Accept a trait in the return of a `#[command]` function. Default implementations will include `()` and `Result`.

We should implement this at the same time as #1665  and #1622  to change all aspects of `#[command]` codegen in a single release. It's likely I will implement this functionality this weekend.

`/* ... */`

**Additional context**
All 3 PRs will have the end result of a small change of `with_window` not doing anything as all things needed for knowing how to handle window will be moved to a trait. It's not a breaking change, but we will not be reading it and it should be removed from all examples and documentation

chippers:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

**Other information:**
So we can't really return a trait while also having a default implementation for `Serialize` because `Result` is also `Serialize` if the contents are.  Since the contents need to be `Serialize` to be useful as a command result, the implementations would conflict if we tried to have a specific implementation for `Result`. If we don't have a specific implementation for result, the value would be sent as `Result<Result<T, E>, InvokeError>`.

This is not the only solution, but is the one with the least user-land code needed. An alternative would be to manually `impl` every type that `Serialize` does for a `CommandResult` trait, and also require the user to implement it for any types they have that are serializable that they wish to return from commands. Not ideal, but possible.

### DIFF
--- a/.changes/async-commands.md
+++ b/.changes/async-commands.md
@@ -1,0 +1,6 @@
+---
+"tauri": patch
+"tauri-macros": patch
+---
+
+Only commands with a `async fn` are executed on a separate task. `#[command] fn command_name` runs on the main thread.
--- a/.changes/command-return.md
+++ b/.changes/command-return.md
@@ -1,0 +1,6 @@
+---
+"tauri": patch
+"tauri-macros": patch
+---
+
+Improves support for commands returning `Result`.
--- a/core/tauri-macros/src/command/mod.rs
+++ b/core/tauri-macros/src/command/mod.rs
@@ -5,7 +5,7 @@
 use proc_macro2::Ident;
 use syn::{Path, PathSegment};
 
-pub use self::{handler::Handler, wrapper::Wrapper};
+pub use self::{handler::Handler, wrapper::wrapper};
 
 mod handler;
 mod wrapper;
--- a/core/tauri-macros/src/command/mod.rs
+++ b/core/tauri-macros/src/command/mod.rs
@@ -5,10 +5,7 @@
 use proc_macro2::Ident;
 use syn::{Path, PathSegment};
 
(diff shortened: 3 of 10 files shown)


chippers:
marked as draft until we decide this is the solution we want to use

chippers:
**Note:** The simple async state command is disabled for now until I look at it again tomorrow. Because `async` functions add `+ '_` to the resulting future, this command expects `State<'_, MyState>` to live until `'static' because that is the lifetime of the resulting future.

Either a way to specify that lifetime better is needed in the underlying code, or a documented workaround to specify that type is needed.

chippers:
`intellij-rust` throws a false positive when using `async` in `#[command(async)]`.  Since `async` is preferable, it will be kept for now.

If https://github.com/intellij-rust/intellij-rust/issues/7214 is closed as wont-fix, then we should add another option to be used. e.g. `future`

chippers:
This PR is ready to review - all commands that existed before it work.

We should probably make sure that the commands example is fully fleshed out, and that every command in rust has a button on the frontend - but that can wait for another PR.

The autoref specialization workaround is what lets us "unwrap" the result of the function if it happens to be `Result<impl Serialize, impl Into<InvokeError>>`.  Without it, the results themselves would be serialized and the variants would be exposed to the sent Ok and Err values. You can tell if this happens with the commands example because the result will be `[object Object]`.  None of the current commands in that example should result in that.

lucasfernog (approved):

PR #1754, opened 2021-05-09
refactor(core): clearer case conversion + docs

chippers:

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
the docs were a lacking for the quick case conversion i just wrote.

empty command keys (like if used with a wildcard `_: String`) for an `impl Serialize` type that are non-optional will now have a specific error message related to that.

if a new string was going to be allocated was a bit hidden, so I moved the underscore check to outside the conversion function.

### DIFF
--- a/core/tauri-macros/src/command/wrapper.rs
+++ b/core/tauri-macros/src/command/wrapper.rs
@@ -154,10 +154,10 @@ fn parse_arg(command: &Ident, arg: &FnArg) -> syn::Result<TokenStream2> {
     }
   };
 
-  // we only support patterns that allow us to extract some sort of keyed identifier.
+  // we only support patterns that allow us to extract some sort of keyed identifier
   let mut key = match &mut arg {
     Pat::Ident(arg) => arg.ident.to_string(),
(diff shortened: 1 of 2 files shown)


core/tauri-macros/src/command/wrapper.rs
```diff
@@ -187,23 +190,18 @@ fn parse_arg(command: &Ident, arg: &FnArg) -> syn::Result<TokenStream2> {
...
+/// Convert a snake_case string into camelCase, no underscores will be left.
+fn snake_case_to_camel_case(key: &str) -> String {
+  let mut camel = String::with_capacity(key.len());
+  let mut to_upper = false;
 
-    camel
-  } else {
-    s
+  for char in key.chars() {
+    match char {
```
chippers:
```suggestion
  for c in key.chars() {
    match c {
```

using char as a binding makes syntax highlighting look a bit weird

lucasfernog (approved):

PR #1821, opened 2021-05-13
update-docs workflow was using an old rustdocusaurus version

Laegel:

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

### DIFF
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -37,7 +37,7 @@ jobs:
         working-directory: ./tauri/core/tauri
         run: cargo doc --no-deps
       - name: run rustdocusaurus
-        uses: tauri-apps/rustdocusaurus/github-action@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/update-docs.yml
```diff
@@ -37,7 +37,7 @@ jobs:
         working-directory: ./tauri/core/tauri
         run: cargo doc --no-deps
       - name: run rustdocusaurus
-        uses: tauri-apps/rustdocusaurus/github-action@v1
+        uses: tauri-apps/rustdocusaurus/github-action@v1.0.3
```
lucasfernog, line 40: you should probably do the same as covector and update the v1 tag every time a new patch is released, but let's do it this way for now.

lucasfernog (approved):

PR #1885, opened 2021-05-21
feat(api): export `BaseDirectory` in `path` module

amrbashir:

- [ ] Bugfix
- [X] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [X] No

- [X] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #xxx[,#xxx]`, where "xxx" is the issue number)

**Other information:**

### DIFF
--- a/.changes/api-export-BaseDirectory.md
+++ b/.changes/api-export-BaseDirectory.md
@@ -1,0 +1,6 @@
+---
+"api": patch
+---
+
+Export `BaseDirectory` in `path` module
+
--- a/tooling/api/src/path.ts
+++ b/tooling/api/src/path.ts
@@ -469,5 +469,6 @@ export {
   templateDir,
   videoDir,
   currentDir,
-  resolvePath
+  resolvePath,
+  BaseDirectory
 }


amrbashir:
People seemed to always import `BaseDirectory` rather than `Dir` and `BaseDirectory` is more descriptive for its job so I removed the `Dir` alias

tooling/api/src/fs.ts
```diff
@@ -367,7 +367,7 @@ export type {
 }
 
 export {
-  BaseDirectory as Dir,
```
lucasfernog: we need to keep this because it's a breaking change :/

lucasfernog (changes requested):

tooling/api/src/fs.ts
amrbashir: aaah you're right, will revert it back

PR #1926, opened 2021-05-31
feat(core): allow `dev_path`, `dist_dir` as array of paths, fixes #1897

### LINKED ISSUE #1897: Exclude src-tauri from relative devPath
**Describe the bug**
Not tested by me - came from Discord help-triage.

The developer said:
> hi. The web side of my app is just hand-written html/css/js. I do not need a local development server. I would like the Tauri/Rust backend to be the server. What should I do when npm run tauri init asks me for the URL of my development server?

I replied that they can put a relative path instead of a dev server location. My example was to put `../`, thinking that devs may keep their vanilla assets at the top level.

Noah highlighted that this may cause issues because it would then include `/src-tauri`.  Obviously this can be bypassed if these are placed in a sub level, like `/src`.

**Expected behavior**
We should exclude the `/src-tauri` directory if pointed at the root level.

We could also add a hint in the docs that devs can use a relative path. This question has come up a few times recently.

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No

**Other information:**


lucasfernog:
The downside of this is that public paths get messy when specifying a file path on the array (currently I just assume it's a root file path). Ideally users should put all their front end assets on a specific folder and point devPath/distDir to that folder, but I agree it's easier to use an array when the frontend is very simple (like most of our examples).

PR #1957, opened 2021-06-06
feat(examples): improve api example, closes #1535

### LINKED ISSUE #1535: API Example Revision
There's a few of things here that aren't really working and we need to iron out before we release the beta and put this example up on the docs site:

- generally: I think that the console display should have newest result at the top instead of the end.
- I get linker errors on a macOS M1, so need to comment out the .cargo/config setting to get it to even build
- CLI example makes no sense / needs information
- new Window panics
- http example should be populated with something
- shell example dies too soon
- should be resizeable / transparent 
- would like to have the sidebar links have the proper mouse pointer (the one that looks like a hand) on hover
- updater should never allow versions older than the "minted timestamp" of the current version to be installed
```
{"body":"This is the latest version! Once updated you shouldn't see this prompt."
,"date":"2020-06-25T14:14:19Z","version":"2.0.0"}
```

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No


lucasfernog:
- [x] transparency
- [x] message history with date, shows latest first
- [x] CLI example docs
- [x] HTTP input URL default value
- [ ] frameless (needs the tauri plugin for titlebars?)

PR #1978, opened 2021-06-11
Fix workflows

rhysd:

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

I found some mistakes in GitHub Actions workflow of this repository. This PR fixes them.

### DIFF
--- a/.github/workflows/core-lint-fmt.yml
+++ b/.github/workflows/core-lint-fmt.yml
@@ -162,11 +162,6 @@ jobs:
 
       - name: Get current date
         run: echo "CURRENT_DATE=$(date +'%Y-%m-%d')" >> $GITHUB_ENV
-        if: matrix.platform == 'macos-latest' || matrix.platform == 'ubuntu-latest'
-
-      - name: Get current date
-        if: matrix.platform == 'windows-latest'
-        run: echo "CURRENT_DATE=$(Get-Date -Format "yyyy-MM-dd")" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
 
       - name: Cache cargo registry
         uses: actions/cache@v2.1.4
--- a/.github/workflows/test-bundler.yml
+++ b/.github/workflows/test-bundler.yml
@@ -96,11 +96,6 @@ jobs:
             override: true
             components: rustfmt, clippy
 
-      - name: install webkit2gtk (ubuntu only)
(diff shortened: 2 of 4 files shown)


.github/workflows/test-bundler.yml
```diff
@@ -96,11 +96,6 @@ jobs:
             override: true
             components: rustfmt, clippy
 
-      - name: install webkit2gtk (ubuntu only)
```
nothingismagick, line 99: We need webkit.

.github/workflows/core-lint-fmt.yml
```diff
@@ -162,11 +162,6 @@ jobs:
 
       - name: Get current date
         run: echo "CURRENT_DATE=$(date +'%Y-%m-%d')" >> $GITHUB_ENV
-        if: matrix.platform == 'macos-latest' || matrix.platform == 'ubuntu-latest'
-
```
nothingismagick, line 166: Time is needed for the cache.

nothingismagick (changes requested):
I don't understand what you think this is really fixing. Please file an issue and explain everything you want this PR to resolve.

nothingismagick:
BUT! Thanks for looking out for us :)

rhysd:
> Please file an issue and explain everything you want this PR to resolve.

I'm currently developing static checker for GA workflows and it detects some mistakes on workflows of this repository. So I just fixed them. You seem to think this PR is not useful so I'm closing this.

.github/workflows/core-lint-fmt.yml
rhysd: This `if` condition is always false. Check Actions result. `matrix.platform` is always `null` since no such matrix variation exists.

.github/workflows/test-bundler.yml
rhysd: This step is never run since `if` condition is always false. See Actions result.

nothingismagick:
I see. The problem is that a heavy handed approach that just deletes things doesn't solve the actual problem, its like amputation instead of removing the sliver.

Nevertheless, thanks for pointing out the issues.

Issue #1977, opened 2021-06-11
Option to use external executor


DBTaylor:
It would be nice if Builder had an run_async method that returned a future to be run on an executor and Tauri used that executor instead of its internal one. Like:
```rust
#[tokio::main]
async fn main() {
	tauri::Builder::default()
		.run_async(tauri::generate_context!())
		.await
		.expect("error while running tauri application");
}
```
I have some managed state that needs to be initialized in an async context. Right now I block_on in setup and use App::manage, but it would be nice if I had the option to initialize pre-Tauri, await, and use Builder::manage.

ravern:
Same here. Trying to `manage` an `sqlx` connection pool but async is required to create the pool and start the connections.

I saw that in the past, `setup` actually took in an `async` closure? What changed?

ravern:
I found this when looking through the examples. It's not exactly using an external executor, but at least it allows you to share the runtime with `tauri`. And I think it solves your problem for initialising async state as well.

https://github.com/tauri-apps/tauri/blob/c1f8e11342941b846479d73bf50cb0a328ac3574/examples/splashscreen/src-tauri/src/main.rs#L21-L39

Perhaps there could be a `tauri::async_runtime::handle` function that we could use to get a `tokio::runtime::Handler` out to use.

lucasfernog:
Unfortunately this is not possible with the current tauri architecture. If you can find a way to implement this, we'd love to see it as a PR! For now you can use the `async_runtime` module.

PR #1983, opened 2021-06-13
feat(tauri): #1977 add `async_runtime::handle` for access to async runtime

ravern:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

### DIFF
--- a/.changes/api-async-runtime-handle.md
+++ b/.changes/api-async-runtime-handle.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch
+---
+
+Adds `async_runtime::handle` to allow obtaining a handle for external usage of Tauri's async runtime.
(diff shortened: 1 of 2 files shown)


chippers:
if this is added, it should be behind a **non-default** `async-runtime-tokio` feature flag to prevent having a dependency re-export which would prevent switching runtimes/apis in the future

lucasfernog:
We should create a wrapper Handle too, so we do not return a tokio type directly.

ravern:
> We should create a wrapper Handle too, so we do not return a tokio type directly.

If this was to make it easier to rework the underlying runtime in the future, perhaps a trait would be better?

```rust
trait RuntimeHandle {
  fn spawn(...);
  fn block_on(...);
}

fn handle() -> impl RuntimeHandle { ... }
```

lucasfernog:
Yeah a trait works too.

lucasfernog:
See #2399.

PR #2047, opened 2021-06-22
feat(core): improve RPC security, closes #814

### LINKED ISSUE #814: Security model of message passing
At this [hackernews](https://news.ycombinator.com/item?id=23771779) comment, a very good point was raised about the null origin.

The OP wrote:

> I recommend using pre-existing interfaces for launching apps like custom scheme URIs, or if really necessary writing individual handlers for the heavy lifting. I think the postMessage approach is great, too but it's vital that the caller `origin` is checked. The web app shouldn't need to run arbitrary commands on the computer.

Instead of allowing the webview to initiate an event, requiring the rust side to do so because we create a randomized handle for each event and can permit it to only be used once. In practice, today, we use the webview to create the event listener, however at build time we can anticipate exactly what listeners are needed - so at runtime we can prepare the proper handle

**Additional context**
http://archive.is/UXD8j

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No

**Other information:**


lucasfernog:
Also see https://github.com/tauri-apps/wry/issues/79#issuecomment-786150571

nothingismagick (approved):
This looks really great. I like the low latency approach!

nklayman (approved):
Looks great to me! This is exactly what I had in mind when I came up with this idea, thanks for implementing it.

PR #2072, opened 2021-06-25
Don't run the update-docs workflow for unimportant changes

thadguidry:
- Added ignored (!) paths that we don't want the update-docs workflow to even start and instead skip it entirely.  See https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-positive-and-negative-patterns-1
- Probably might be other folders or files that could be ignored to skip the update-docs workflow (I don't know what they might be however)

- [ ] Bugfix
- [ ] Feature
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [X] Other, please describe: CI/CD

- [ ] Yes. Issue #___
- [X] No

**Other information:**

### DIFF
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -7,6 +7,11 @@ name: update-docs
 on:
   release:
     types: [published]
+    paths:
+      - '!README.md'
+      - '!LICENSE'
+      - '!.gitignore'
+
   workflow_dispatch:
     inputs:
       gitName:


Laegel:
Thanks for the PR @thadguidry 
I'm not sure we would make a release that doesn't imply any modification on the docs but I guess in this case it's "better safe than sorry", right?
What do you think @lucasfernog ?

thadguidry:
Ah then we probably want context at 2 event levels. A push or a release? And we could deal with each uniquely with their own criteria.  On this PR it touches the docs ci/cd workflow and not a step within it.

There are options at different levels.

What do you think?

Laegel:
Why would we trigger the workflow on a push event?

thadguidry:
@Laegel Thanks for explaining more about the Docs updating on Discord!  So now that I understand more, we don't need this and can close this.

PR #2104, opened 2021-06-28
feat: asset custom protocol to access local file system

### LINKED ISSUE #2103: Support for accessing files from local filesystem with in html
As of now, I am not able to access files from the local file system within HTML.
Reading files as binary then converting to base64 cant be considered as a solution to this issue as this is super slow especially for video files.
I believe local file system access within the renderer is a very basic requirement.

wusyong:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
Fix #2103 

Which feature flag should this feature put behind btw?

### DIFF
--- a/core/tauri/src/manager.rs
+++ b/core/tauri/src/manager.rs
@@ -343,6 +343,15 @@ impl<P: Params> WindowManager<P> {
         current_window_label = label.to_js_string()?,
       ));
 
+    webview_attributes.uri_scheme_protocols.insert(
+      "asset".into(),
+      Box::new(move |url| {
+        let path = url.replace("asset://", "");
+        let data = crate::async_runtime::block_on(async move { tokio::fs::read(path).await })?;
+        Ok(data)
+      }),
+    );
+
     #[cfg(dev)]
     {
       webview_attributes = webview_attributes.initialization_script(&format!(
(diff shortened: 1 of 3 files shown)


lucasfernog (approved):

lucasfernog:
We need to decide where to document this.

wusyong:
Where do we put the documents about loading assets. Perhaps it should put there too.

lucasfernog:
We need to create one lol. Good thing it's documentation month.

core/tauri/src/manager.rs
```diff
@@ -343,6 +343,15 @@ impl<P: Params> WindowManager<P> {
         current_window_label = label.to_js_string()?,
       ));
 
+    webview_attributes.uri_scheme_protocols.insert(
```
lemarier, line 346: Could we have a setter or something to update the path? 
Let say a user update change the path via a UI, asking a restart is a bit overkill I think?

lucasfernog: which path? the protocol just loads `asset:///path/to/file`.

lemarier: ho sorry, just saw that it's the dev responsibility to map the right path!
alls good

PR #2179, opened 2021-07-08
feat: simplfy `Args`

wusyong:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes. Issue #___
- [ ] No

**Other information:**
Remove all generics from `Args` and implement `Params` to it directly.
It also makes signatures of all types and functions based on it
simpler. The compiler time could be improved without the need for macro too.

This means `Arg` is also the default arg type and it requires `wry`
feature flag. If there's a new runtime in the future, all it needs to do
is add another `Arg` under its feature flag.

Custom command and params should still work, see `params` example to
learn how to define it now. Not much needs to change.

### DIFF
--- a/.changes/simpler-args.md
+++ b/.changes/simpler-args.md
@@ -1,0 +1,5 @@
+---
+"create-tauri-app": minor
+---
+
+Simplify `Args` and many other types that based on it.
--- a/core/tauri-runtime/src/lib.rs
+++ b/core/tauri-runtime/src/lib.rs
@@ -132,7 +132,7 @@ pub mod private {
 }
 
 /// Types associated with the running Tauri application.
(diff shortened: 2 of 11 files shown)


chippers:
So the big reason why we have both `Params` and `Args` as they are right now is so that we can expand the associated types in the future as long as they have a default type because:
1. `Args` is not exposed to the user at all
2. `Params` is sealed, so no user can implement it

Unsealing it means we can no longer expand types on it in the future without a breaking change because users can now `impl` it for their own types and it will break their codebase when we add another type

PR #2191, opened 2021-07-09
refactor(core): remove `Params` and replace with strings

chippers:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] **YES**

**Other information:**
See #2179 for prior work that sparked this PR. I've added most of my TODO notes to the change log instead of throughout the code because the diff is large enough to lose track of them easily.

The conversation in discord spanned a couple hours and a few more smaller discussion after that, so I will try to sum up the motivation behind the breaking change in bullet points:
  * `Params` was added during a large refactor multiple months ago, and allowed us to put user-defined types throughout the codebase (bonus: statically dispatched).
  * Features grew and the list of associated types grew.  We used `Args` as a private marker`Params`  type to simplify the generic parameters in most of the internal code but it needed a lot of generics on it. This was of course particularly painful where we had to use `Args` instead of `Params` such as the builder and internal window manager.
  * Not many people were taking much advantage of the strong types possible through `Params` (ofc it was just released during beta) and instead struggled to comprehend the now mentally complex Rust API (particularly to people learning or new to Rust) in places. The not many people measurement is vague, but based on hardly any questions in the discord were about using it and instead about trying to use strings in the API where `Params` made things more complex type-wise. Also in a cursory GitHub search of projects using Tauri beta the first 4-5 pages had no projects where non-default (`String`) types were used.
  * Because of the nature of the traits used in the associated types on `Params`, the types effectively always had to represent a string in some form (`Display` or `Serialize`). Storing those types directly with static dispatch in internal implementation is not really necessary as they had to be turned into a string at some point anyways.
  * Due to the previous point, a strongly typed API for `tauri` doesn't require being injected into the internals for all current cases. It can instead live in another crate (TBD) with a less stable API as we figure out the most ergonomic way to allow complex compile time type checking without ruining the ease of development without it.

So this PR is the result of those bullet points. Because another crate can effectively wrap `tauri` with a stronger API, the complexity introduced in `tauri` itself for it is unnecessary and should be shed before  `v1`. The _more_ strongly typed API can be created separately at a later time.

**This is a breaking change.**  It completely removes `Params` and replaces all the associated types previously on it with either strings (`Event`, `Label`, `MenuId`, `SystemTrayMenuId`) or trait objects (`Assets`). Items that previously took `Params` now take only `Runtime`. It also removes traits that served solely to help manage those string types, such as `Tag`, `TagRef`, and `MenuId`.

I was (hopefully) consistent about putting `impl Into<String>` in places where we are accepting an owned string in a public api, let me know if I missed any. We might also want to change all the public API that accepts `&str` to `impl AsRef<str>` but I didn't do that as I went.

We should really flesh out the changelog on this change since this is a decently sized breaking change.

Marked as draft until we finish off the todos and the change file.


core/tauri-macros/src/lib.rs
```diff
@@ -60,3 +60,44 @@ pub fn generate_context(items: TokenStream) -> TokenStream {
   let path = parse_macro_input!(items as ContextItems);
   context::generate_context(path).into()
 }
+
+/// Adds a default value `crate::Wry` to the last generic item of a struct/enum
+#[doc(hidden)]
+#[proc_macro_attribute]
+pub fn default_runtime_wry(_attrs: TokenStream, input: TokenStream) -> TokenStream {
```
wusyong: Maybe just name it `default_runtime` since it might be another runtime and we'll set it depends on the feature flag in the future.

wusyong:
Thanks! I think this will really improve a lot on code readability and still keep the runtime generic configurable.

core/tauri/Cargo.toml
```diff
@@ -86,7 +86,7 @@ tokio-test = "0.4.2"
 mockito = "0.30"
 
 [features]
-default = [ "wry" ]
+default = [ "wry", "menu" ]
```
wusyong: I suppose this is not intended?

.changes/weak-typing.md
```diff
@@ -0,0 +1,38 @@
...
+`Params` has been removed, along with all the associated types on it. Functions that previously accepted those
+associated types now accept strings instead. Type that used a generic parameter `Params` now use `Runtime` instead. If
+you use the `wry` feature, then types with a `Runtime` generic parameter should default to `Wry`, letting you omit the
+explicit type and let the compiler infer it instead.
+
+`tauri`:
+
+* See `Params` note
+* **TODO** should we change all the `&str` to `AsRef<str>`?
+* **TODO** (probably list all the methods/functions that changed)
```
wusyong: There are too many functions and methods. We could just state the generics of many functions are updated but shouldn't affect majority of users.

core/tauri/Cargo.toml
chippers: I forgot I set that before i set the `--all-features --all-targets` on my IDE clippy args, will be reverted

core/tauri-macros/src/lib.rs
chippers: Yeah, and then accept the default runtime in the attributes. will look like `#[default_runtime(Wry)]` or `#[default_runtime(crate::Wry)]` depending on if it was imported (which is probably won't since it's behind a feature flag).

wusyong (approved):

PR #2199, opened 2021-07-12
fix: center and focus not being allowed in config

FabianLars:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

### DIFF
--- a/core/tauri-runtime-wry/src/lib.rs
+++ b/core/tauri-runtime-wry/src/lib.rs
@@ -442,6 +442,10 @@ impl WindowBuilder for WindowBuilderWrapper {
       window = window.position(x, y);
     }
 
+    if config.center {
+      window = window.center();
+    }
+
     if config.focus {
       window = window.focus();
     }
--- a/tooling/cli.rs/config_definition.rs
+++ b/tooling/cli.rs/config_definition.rs
@@ -268,6 +268,9 @@ pub struct WindowConfig {
   /// Disabling it is required to use drag and drop on the frontend on Windows.
   #[serde(default = "default_file_drop_enabled")]
   pub file_drop_enabled: bool,
+  /// Whether or not the window starts centered or not.
+  #[serde(default)]
+  pub center: bool,
   /// The horizontal position of the window's top left corner
   pub x: Option<f64>,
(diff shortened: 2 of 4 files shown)


FabianLars:
The only problem i have is that changing focus via the config doesn't make a difference at all (at least on Windows).

core/tauri-runtime-wry/src/lib.rs
```diff
@@ -442,6 +442,10 @@ impl WindowBuilder for WindowBuilderWrapper {
       window = window.position(x, y);
     }
 
+    if config.center {
```
FabianLars, line 445: This way center overwrites x and y. I don't know if it should do that tho.

FabianLars, line 445: This way center overwrites x and y. I don't know if it should do that tho.

tooling/cli.rs/config_definition.rs
```diff
@@ -292,6 +295,9 @@ pub struct WindowConfig {
   /// Whether the window starts as fullscreen or not.
   #[serde(default)]
   pub fullscreen: bool,
+  /// Whether the window will be initially hidden or focused.
+  #[serde(default = "default_focus")]
+  pub focus: bool,
```
FabianLars, line 300: Changing this in tauri.conf.json doesn't seem to make a difference (at least on Windows).

amrbashir: We decided to remove `focus` from `WindowBuilder` and `WindowAttributes` at Tao, so if tauri wants to support this, they would have to explicitly call `window.set_focus()`.

FabianLars: Should i remove the focus stuff again? -> Make this pr only about the center thingy?

amrbashir: It is @lucasfernog decision.

lucasfernog (approved):

PR #2202, opened 2021-07-12
Revert "fix(wry): resize webview on window move, closes #1911"

### LINKED ISSUE #1911: <select> opens outside window on first try
`<select>` elements open outside the application window. When opened a second time, they open in the correct position. The problem occurs again when moving the window to a new location.

OS: Windows, version 10.0.19043 X64
Node: 15.14.0
Rustc: 1.52.1

lucasfernog:
Reverts tauri-apps/tauri#2135

### DIFF
--- a/.changes/fix-webview-resize-on-move.md
+++ b/.changes/fix-webview-resize-on-move.md
@@ -1,5 +1,0 @@
----
-"tauri-runtime-wry": patch
----
-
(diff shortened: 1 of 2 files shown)


lucasfernog:
The fix caused more issues (both performance and resize event issues) so we need a different solution, probably letting the webview2 folks fixing it.

Issue #2281, opened 2021-07-22
URL in `WindowManager.prepare_window()` has double slashes


woubuc:
### **Describe the bug**
When a page URL is joined with the base URL, too many slashes are added, causing some dev servers (in my case, the Nuxt dev server) to return a 404 instead of showing the page.

With config..
```json
"devPath": "http://localhost:3000",
```
..[`WindowManager.get_url()`](https://github.com/tauri-apps/tauri/blob/74a278fcfc60f51c879347476c3f76e286a3fd69/core/tauri/src/manager.rs#L184-L198) returns `http://localhost:3000/` *with* a trailing slash, even though in the config file it doesn't have a trailing slash.

However, [`WindowManager.prepare_window()`](https://github.com/tauri-apps/tauri/blob/74a278fcfc60f51c879347476c3f76e286a3fd69/core/tauri/src/manager.rs#L548-L572) combines the base url with the path like this..
```rs
format!("{}/{}", url, path.to_string_lossy())
```

..which means that this code..
```rs
app.create_window(
    "login".into(),
    WindowUrl::App("login".into()),
    |builder, attrs| { [...] },
).unwrap();
```
..results in the webview trying to load `http://localhost:3000//login` because the format call adds an additional slash inbetween the base url and the path. This is a problem if the dev server doesn't normalise the path and just tries to match `//login` - resulting in a 404 instead of the expected page.

If I replace that format line in `manager.rs` in the Tauri project with the following, to prevent adding a slash inbetween the parts if the base url already has a trailing slash..
```rs
let separator = match url.ends_with("/") {
    true => "",
    false => "/",
};

format!("{}{}{}", url, separator, path.to_string_lossy())
```
..then I get `http://localhost:3000/login` and everything works as expected.

I don't know what platform and compatibility considerations the URLs in Tauri have, so I don't know if this fix would be applicable for all users and build targets. But if the above fix is acceptable I'll gladly create a pull request.

### **Platform and Versions (required):**
```
Operating System - Windows, version 10.0.19042 X64
Webview2 - 91.0.864.71

Node.js environment
  Node.js - 14.16.0
  @tauri-apps/cli - 1.0.0-beta.6
  @tauri-apps/api - 1.0.0-beta.5

Global packages
  npm - 6.14.11
  yarn - 1.22.4

Rust environment
  rustc - 1.53.0
  cargo - 1.53.0

App directory structure
/.idea
/.nuxt
/assets
/components
/dist
/i18n
[...]

PR #2282, opened 2021-07-22
fix(#2281): Prevent double slashes when joining URLs

woubuc:
- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**


core/tauri/src/manager.rs
```diff
@@ -561,7 +561,14 @@ impl<R: Runtime> WindowManager<R> {
           true,
           // ignore "index.html" just to simplify the url
           if path.to_str() != Some("index.html") {
-            format!("{}/{}", url, path.to_string_lossy())
+            // If the base URL has a trailing slash, we don't need to put an additional slash
+            // between the base URL and the path.
+            let separator = match url.ends_with("/") {
```
amrbashir: What if `path` doesn't have a leading slash? I think it is better to check for leading slash in `path` instead.
```suggestion
            let separator = match path.starts_with("/") {
```

amrbashir (changes requested):
Thanks for finding and fixing the bug. 
Don't forget to add a change file in `.changes` folder

core/tauri/src/manager.rs
woubuc: In that case I reckon we should check both? Because if either one has a slash, no additional slash should be added in between. Should we also remove one of the slashes if the domain has a trailing slash *and* the path has a leading slash?

Checking the path as well makes this a little more complicated and at that point it might be more advisable to use a proper "url join & normalise" function that handles all sorts of edge cases, rather than rolling our own thing with `format!()` in here.

amrbashir: yeah that's a valid concern but for the past couple of months we have been adding a slash to `url` no matter what and it worked fine so I don't think there will ever be a trailing slash in `url` but `path` can have a leading slash though. what do you think @lucasfernog ?

amrbashir: also if you want to make a separate function to handle both cases, go ahead it will be better.

woubuc: > we have been adding a slash to `url` no matter what and it worked fine so I don't think there will ever be a trailing slash in `url`

I gotta say I'm a little confused by this statement. This PR is a fix for #2281 which is specifically about a trailing slash in `url` that causes it to *not* work fine.

In any case, I'll look into proper URL concatenation and normalisation here because that seems like the better solution anyways. I saw that Tauri already uses the [url](https://crates.io/crates/url) crate internally for the base URLs so I'm gonna explore the codebase a little more to see if we can't just use the [`Url.join()`](https://docs.rs/url/2.2.2/url/struct.Url.html#method.join) function which [does exactly what we need](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=95f4444200e4bfb8e514f931c759c698).

Gonna mark the PR as a draft for now.

amrbashir: > > we have been adding a slash to `url` no matter what and it worked fine so I don't think there will ever be a trailing slash in `url`
> 
> I gotta say I'm a little confused by this statement. This PR is a fix for #2281 which is specifically about a trailing slash in `url` that causes it to *not* work fine.

Sorry if I wasn't clear, the trailing slash is manually added by us for quite some time now, so you can assume that `url` will never have a trailing slash.

But yeah it is better to check for trailing slash in the base `url` and leading slash in `path` provided by the user

woubuc:
* The custom `format!()` call has been replaced with the [`Url.join()`](https://docs.rs/url/2.2.2/url/struct.Url.html#method.join) function which handles concatenating and normalising URLs a lot better.
  * `WindowManager.get_url()` now returns a `Cow<Url>` to make it possible for other functions to use the `Url` struct directly. I've opted to use `Cow` to prevent cloning the URL since we just need to read it.
  * The relevant tests in `manager.rs` have been updated as well to reflect this change.
  * I've also changed the check for external URLs to compare against `Url.scheme()` instead of converting the entire `Url` to a string first.
* A new `Error` variant `InvalidUrl` is now used to return parse errors that arise from the `url` functions (e.g. when trying to load a path with invalid characters).
* I've also reduced some of the duplicate code in the `WindowManager.get_url()` functions, the cfg-dependant lines are now a separate function and there is only one `match`. I think this makes it a bit clearer what happens here. Do let me know if I've misunderstood what the cfg differences are though.

core/tauri/src/manager.rs
```diff
@@ -180,20 +181,27 @@ impl<R: Runtime> WindowManager<R> {
...
-    match &self.inner.config.build.dist_dir {
-      AppUrl::Url(WindowUrl::External(url)) => url.to_string(),
-      _ => "tauri://localhost".into(),
+  /// Get the base URL to use for webview requests.
+  ///
+  /// In dev mode, this will be based on the `devPath` configuration value.
+  fn get_url(&self) -> Cow<'_, Url> {
+    match self.base_path() {
+      AppUrl::Url(WindowUrl::External(url)) => Cow::Borrowed(url),
+      _ => Cow::Owned(Url::parse("tauri://localhost").unwrap()),
```
amrbashir, line 204: just a small nit-pick, the comments are not needed, the code is clear enough

amrbashir (approved):
LGTM, thanks!

core/tauri/src/manager.rs
woubuc: IDE hinting shows function signature and doc comments, not the internal code, so I like to write doc comments for every function no matter how simple it is to make it easier when using them elsewhere in the code. So it's mostly a force of habit. I agree that the code is probably clear enough, but I don't think it's a problem to have these doc comments here?

woubuc:
Looks like I forgot to run Clippy locally before committing.. Should be fixed now.

core/tauri/src/manager.rs
amrbashir: It is not a problem, it was just a preference of mine, but I am not the main maintainer anyway so you'd better wait for his review.

lemarier:
Thanks @woubuc !

We do appreciate a lot your contribution!
Have a great day.

PR #2291, opened 2021-07-24
`App.create_window()` to accept any `Into<String>` type (fix #2290)

### LINKED ISSUE #2290: [Feature request] Make `App.create_window()` accept any `Into<String>` type for the label
The `App.create_window()` function takes a label as its first argument. This currently needs to be a `String` so I need to use `.into()` on every call to this function.

I'd like to be able to pass a `&str` label to the function without having to add `.into()`.

**Additional context**
I've [implemented this change on my fork](https://github.com/woubuc/tauri/commit/00aad69c8515496b031e829f423848a1fb8d0d41) to see how complicated it would be. All internal functions seem to already use `impl Into<String>` for labels, it's only the main public-facing function that is fixed on a `String`.

Changing the type to `impl Into<String>` will cause `.into()` to stop working, so since this is the public-facing API this would constitute a breaking change and developers would need to update their uses of `app.create_window()` when updating their Tauri version. I reckon now is the time to make a usability change like this, since Tauri is still in beta.

chippers:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes. Issue #2290 
- [ ] No

**Other information:**
See #2290.

### DIFF
--- a/.changes/fix-create-window-str.md
+++ b/.changes/fix-create-window-str.md
@@ -2,4 +2,4 @@
 "tauri": major
 ---
 
-Fix `App.create_window()` to accept any `Into<String>` type instead of only `String` itself.
+Change `App.create_window()` and `AppHandle.create_window()` to accept an `Into<String>` type instead of `String`.
--- a/.changes/fix-create-window-str.md
+++ b/.changes/fix-create-window-str.md
@@ -1,5 +1,5 @@
 ---
-"tauri": major
+"tauri": patch
 ---
 
 Change `App.create_window()` and `AppHandle.create_window()` to accept an `Into<String>` type instead of `String`.
--- a/.changes/fix-create-window-str.md
+++ b/.changes/fix-create-window-str.md
@@ -1,0 +1,5 @@
+---
+"tauri": major
+---
+
+Fix `App.create_window()` to accept any `Into<String>` type instead of only `String` itself.
--- a/core/tauri/src/app.rs
+++ b/core/tauri/src/app.rs
@@ -225,7 +225,12 @@ macro_rules! shared_app_impl {
   ($app: ty) => {
     impl<R: Runtime> $app {
       /// Creates a new webview window.
-      pub fn create_window<F>(&self, label: String, url: WindowUrl, setup: F) -> crate::Result<()>
+      pub fn create_window<F>(
+        &self,
+        label: impl Into<String>,
+        url: WindowUrl,
+        setup: F,
+      ) -> crate::Result<()>
       where
         F: FnOnce(
           <R::Dispatcher as Dispatch>::WindowBuilder,
--- a/examples/api/src-tauri/src/main.rs
+++ b/examples/api/src-tauri/src/main.rs
@@ -82,7 +82,7 @@ fn main() {
           }
           "new" => app
             .create_window(
-              "new".into(),
+              "new",
               WindowUrl::App("index.html".into()),
               |window_builder, webview_attributes| {
                 (window_builder.title("Tauri"), webview_attributes)


lemarier (approved):

woubuc:
I've updated the changefile to reflect both changed functions, like @chippers suggested in #2290.

I do want to note that this may be considered a breaking change (which is why I marked it as 'major' in the changefile). Especially because @chippers said this isn't a major change and I don't entirely agree. (I was going to comment in the issue but I feel it's more appropriate to continue this conversation here in the PR).

> One thing to be changed is the changefile level, this should be a patch not major change

Because while the type is technically backwards-compatible (since `String` is also `Into<String>`), in practice most developers who use this function will need to update their code.

The current `create_window()` function takes a `String` for the label. So when a developer uses that function to create a window, they'll do something like this:
```rs
app.create_window("my_label".into(), [...]);
```

Note the `.into()`. This is also how it's done in the example.

With this PR, `create_window()` accepts any `Into<String>` type for the label, so developers will be able to just do this:
```rs
app.create_window("my_label", [...]);
```

However, when trying to use the original code with `.into()` after applying this change, you get the following error:
```
error[E0283]: type annotations needed
  --> examples\api\src-tauri\src\main.rs:84:14
   |
84 |             .create_window(
   |              ^^^^^^^^^^^^^ cannot infer type for type parameter `impl Into<String>` declared on the associated function `create_window`
85 |               "new".into(),
   |               ------------ this method call resolves to `T`
   |
   = note: cannot satisfy `_: Into<std::string::String>`
```

So because the `impl Into<String>` type doesn't point to a concrete type, `.into()` can't determine what type to convert the value into. This means that most (if not all) developers who use this function will be presented with these errors when they upgrade their Tauri version and they'll have to fix them to get their app to compile. Theoretically the function still accepts the same type as before and this PR just expands the accepted types a bit, but I would in fact consider this a major change because it *will* break code.

If you don't consider this a breaking change, let me know and I'll bring the changefile down to a patch level.

chippers:
> I've updated the changefile to reflect both changed functions, like @chippers suggested in #2290.
> 
> I do want to note that this may be considered a breaking change (which is why I marked it as 'major' in the changefile). Especially because @chippers said this isn't a major change and I don't entirely agree. (I was going to comment in the issue but I feel it's more appropriate to continue this conversation here in the PR).
> 
> > One thing to be changed is the changefile level, this should be a patch not major change
> 
> Because while the type is technically backwards-compatible (since `String` is also `Into<String>`), in practice most developers who use this function will need to update their code.
> 
> The current `create_window()` function takes a `String` for the label. So when a developer uses that function to create a window, they'll do something like this:
> ```rs
> app.create_window("my_label".into(), [...]);
> ```
> 
> Note the `.into()`. This is also how it's done in the example.
> 
> With this PR, `create_window()` accepts any `Into<String>` type for the label, so developers will be able to just do this:
> ```rs
> app.create_window("my_label", [...]);
> ```
> 
> However, when trying to use the original code with `.into()` after applying this change, you get the following error:
> ```
> error[E0283]: type annotations needed
>   --> examples\api\src-tauri\src\main.rs:84:14
>    |
> 84 |             .create_window(
>    |              ^^^^^^^^^^^^^ cannot infer type for type parameter `impl Into<String>` declared on the associated function `create_window`
> 85 |               "new".into(),
>    |               ------------ this method call resolves to `T`
>    |
>    = note: cannot satisfy `_: Into<std::string::String>`
> ```
> 
> So because the `impl Into<String>` type doesn't point to a concrete type, `.into()` can't determine what type to convert the value into. This means that most (if not all) developers who use this function will be presented with these errors when they upgrade their Tauri version and they'll have to fix them to get their app to compile. Theoretically the function still accepts the same type as before and this PR just expands the accepted types a bit, but I would in fact consider this a major change because it *will* break code.
> 
> If you don't consider this a breaking change, let me know and I'll bring the changefile down to a patch level.

This is a fair point, but we are kind of abusing the patch level while we are in beta. A recent change that also did this is the `Params` removal which had a similar resolution of just removing unnecessary code.  We are still pre-audit for a short amount of time, so necessary breaking changes that align with a stable future API are still currently marked `patch`.

Good point about the breaking change however, and I'll update the PR original comment to specify it

woubuc:
I've updated the changefile to patch level.

Issue #2287, opened 2021-07-23
[Feature request] Keep app running after all windows are closed


woubuc:
**Is your feature request related to a problem? Please describe.**
I am building an app that should "live" in the system tray. The app window can be closed and later reopened by clicking on the tray icon. However, when the last window of the app closes, Tauri exits.

**Describe the solution you'd like**
I would like an option to keep the app running after all windows are closed, so that I can create a new window when the user clicks on the tray icon.

**Describe alternatives you've considered**
Hiding the window instead of closing it will keep the app active, so this is a possible workaround. But this means that the window still exists in memory and still uses resources while that isn't necessary.

woubuc:
I'd be interested in figuring out how to implement this and creating a PR.

On first look, there seems to be an easy way to add a parameter to `App.run()` to prevent killing the app [here in `tauri/app.rs`](https://github.com/tauri-apps/tauri/blob/70a1941468f55f0dc09ac2e13802945891d766f4/core/tauri/src/app.rs#L305-L315), but I think it's probably more appropriate to have this config passed down from a builder setter through the runtime to the event loop handler where the `RunEvent::Exit` is created [here in `tauri-runtime-wry/lib.rs`](https://github.com/tauri-apps/tauri/blob/70a1941468f55f0dc09ac2e13802945891d766f4/core/tauri-runtime-wry/src/lib.rs#L1948-L1951). Is my understanding correct or would this logic go somewhere else?

woubuc:
I've gone ahead and hacked something together based on my (so far fairly limited) understanding of the codebase.

1. A `RunMode` enum can be set on the builder
2. This run mode is then passed through to the app and then to the runtime
3. The `Runtime.run()` and `Runtime.run_iteration()` functions have a new parameter `run_mode` which adds the run mode to the `EventLoopIterationContext`.
4. From there the event loop handler can use it where needed.

I don't want to create a PR just yet because there hasn't been enough discussion to determine if this is a desired approach, but you can check out the [feat/run-mode branch](https://github.com/woubuc/tauri/tree/feat/run-mode) in my fork, or [the diff with the `dev` branch](https://github.com/woubuc/tauri/compare/dev...feat/run-mode) to see just what I changed to make it work.

If this approach sounds good then I'll go through it, clean it up where necessary, and create a proper PR for review. If y'all would prefer this feature to be approached in a different way, no problem, I'll just delete my branch and we can discuss how to do this properly. 😄

[...]

PR #2293, opened 2021-07-24
feat(#2287): Add `ExitRequested` event to let users prevent app from exiting 

woubuc:
- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #2287 
- [x] No

**Other information:**
See issue #2287


amrbashir (commented):
LGTM, @lucasfernog @chippers would it be better to combine `CloseRequestApi` and `ExitRequestApi` into one type with `prevent_default()` method that can be used with all events.
With the current implementation, we will have to keep adding new types for every event that user want to take control of.

woubuc:
> combine `CloseRequestApi` and `ExitRequestApi` into one type with `prevent_default()` method that can be used with all events

My 2c: I've personally never been a big fan of Javascript's generic-sounding `preventDefault`. I think the API is clearer if it explicitly says `prevent_close()` and `prevent_exit()` so you know what specific behaviour you're preventing.

amrbashir:
> > combine `CloseRequestApi` and `ExitRequestApi` into one type with `prevent_default()` method that can be used with all events
> 
> My 2c: I've personally never been a big fan of Javascript's generic-sounding `preventDefault`. I think the API is clearer if it explicitly says `prevent_close()` and `prevent_exit()` so you know what specific behaviour you're preventing.

Yeah that's a good point, I just don't want to have a lot of types that will be harder to maintain as we add similar types for more events. I want to somehow scope this functionality into one type or an enum of some sort and if we can't then your PR is good to go.

woubuc:
> I just don't want to have a lot of types that will be harder to maintain as we add similar types for more events. I want to somehow scope this functionality into one type or an enum of some sort and if we can't then your PR is good to go.

We could probably achieve both goals, at least somewhat. What if we implement the logic for sending and checking messages in one internal struct that contains all the needed functionality, but use differently typed wrappers for the public API? 

Something like:
```rs
// Shared internal struct, events just use this if they want to get a response
struct OneTimeResponse<T>(Sender<T>);

impl<T> OneTimeResponse<T> {
    fn new() {
        // Shared channel creation logic (no longer in each individual event)
    }
    fn send(self, response: T) {
        // One-time response send logic
        // Each API wrapper just calls this without needing to know internal implementation
    }
}

// The public API wrappers only contain calls to the shared struct, no own logic
struct ExitRequestedApi(OneTimeResponse<ExitRequestedAction>);

impl ExitRequestedApi {
    fn prevent_exit(self) {
        self.0.send(ExitRequestedAction::Prevent);
    }
}
```

This way, all the actual logic that we need to maintain for sending & receiving these one-time responses is located in 1 struct. The internal events and APIs all use this main struct, but the public API types remain user-friendly, specific and flexible.

amrbashir:
> We could probably achieve both goals, at least somewhat. What if we implement the logic for sending and checking messages in one internal struct that contains all the needed functionality, but use differently typed wrappers for the public API?
> 
> Something like:
> 
> ```rust
> // Shared internal struct, events just use this if they want to get a response
> struct OneTimeResponse<T>(Sender<T>);
> 
> impl<T> OneTimeResponse<T> {
>     fn new() {
>         // Shared channel creation logic (no longer in each individual event)
>     }
>     fn send(self, response: T) {
>         // One-time response send logic
>         // Each API wrapper just calls this without needing to know internal implementation
>     }
> }
> 
> // The public API wrappers only contain calls to the shared struct, no own logic
> struct ExitRequestedApi(OneTimeResponse<ExitRequestedAction>);
> 
> impl ExitRequestedApi {
>     fn prevent_exit(self) {
>         self.0.send(ExitRequestedAction::Prevent);
>     }
> }
> ```
> 
> This way, all the actual logic that we need to maintain for sending & receiving these one-time responses is located in 1 struct. The internal events and APIs all use this main struct, but the public API types remain user-friendly, specific and flexible.

tbh, I don't think this helps a lot, and since I don't have any good ideas about this, I won't hold this PR because of it.

amrbashir (approved):

woubuc:
> tbh, I don't think this helps a lot

I think it would help keep maintenance down by having the actual logic implementation in 1 place instead of spread out & duplicated across multiple structs/functions in various crates.

But I agree that it's not a big issue at this point so it can be revisited later if (when) we get more events using the same pattern, if y'all prefer that.

amrbashir:
> > combine `CloseRequestApi` and `ExitRequestApi` into one type with `prevent_default()` method that can be used with all events
> 
> My 2c: I've personally never been a big fan of Javascript's generic-sounding `preventDefault`. I think the API is clearer if it explicitly says `prevent_close()` and `prevent_exit()` so you know what specific behaviour you're preventing.
> 

I have thought about this again and I think `prevent_default()` should be fine, users will know exactly what they are preventing because it will be passed through the Event itself.

```rs
match event {
    Event::WindowCloseRequested { api, .. } => {
        api.prevent_default(); // here it is obvious we are preventing Window closing
    }
    Event::ExitRequested { api, .. } => {
       api.prevent_default(); // and here prevent app exit
    }
}
```

woubuc:
> I have thought about this again and I think `prevent_default()` should be fine

For the 2 events that use it at this point, I agree. But I'm looking to the future:

- What if an event has more options than just "prevent"?
- What if we add events where a non-specific "prevent" isn't entirely clear? 

Since we're establishing a pattern here for how to handle this type of event (like you said in #2287, we should strive for consistency in this data flow), I think it's worth taking a moment to figure out how to handle this pattern properly in a more general sense.

I also follow [your earlier suggestion](https://github.com/tauri-apps/tauri/pull/2293#issuecomment-886737666) regarding maintainability:

> I want to somehow scope this functionality into one type or an enum of some sort

With the only difference being that I think it would make more sense to abstract the actual functionality into a generic implementation, rather than reducing the public API for it down to a single rigid type. The end result in both cases is that each individual event doesn't add to the maintenance load since it reuses the same implementation that's already there.

Perhaps it might be useful to get some more feedback from a code architecture standpoint on this issue, and determine whether we want to establish patterns for things like this in the codebase, or we just look at every piece of the puzzle separately - personally I tend to look at the bigger picture, especially for larger projects like Tauri.

Of course, I don't want this discussion to block the current PR, if you'd prefer to just merge this in and figure out the bigger picture stuff later then go ahead. Although in my experience "later" can often turn into "never" because, hey, it works doesn't it 😄

amrbashir:
> > I have thought about this again and I think `prevent_default()` should be fine
> 
> For the 2 events that use it at this point, I agree. But I'm looking to the future:
> 
> * What if an event has more options than just "prevent"?
> * What if we add events where a non-specific "prevent" isn't entirely clear?

I hardly think that would ever be the case. we just want users to have the ability to prevent our internal handling of the event, nothing more. 
gtk has `Inhibit(true or false)` and javascript has `preventDefault()` for all kinds of events so I don't think it is bad idea in general


> Perhaps it might be useful to get some more feedback from a code architecture standpoint on this issue, and determine whether we want to establish patterns for things like this in the codebase, or we just look at every piece of the puzzle separately - personally I tend to look at the bigger picture, especially for larger projects like Tauri.

yeah this needs to be discussed further, probably in discord. and you as a tauri user/contributor your feedback is extremely important.

> Of course, I don't want this discussion to block the current PR, if you'd prefer to just merge this in and figure out the bigger picture stuff later then go ahead. Although in my experience "later" can often turn into "never" because, hey, it works doesn't it 😄

yeah the PR is fine to me but I can't merge it. @lucasfernog @chippers are the main maintainers.

lucasfernog (approved):

lucasfernog:
Nice work @woubuc

PR #2324, opened 2021-07-29
Add back WebDriver support

chippers:
This reverts commit b4426eda9e64fcdd25a2d72e548b8b0fbfa09619.

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No

**Other information:**
Hopefully this time with less multi-window bugs.

I'm marking this as a draft for right now because we currently use `TAURI_AUTOMATION=true` environmental variable to detect if we should enable automation on linux. I was thinking that some people might set that true in their shell or profile on their development machine, and may unknowning affect their other Tauri projects. Should we expose some sort of config setting to allow them to set their own env var name to watch? Maybe something on the config like:
```json
{
  /* ... snip ... */
  "webdriver": {
    "flag": {
      "name": "MY_APP_AUTOMATION",
      "value": "true",
    }
  }
  /* ... snip ... */
}
```
and maybe we default value the name of the application from cargo env + `_AUTOMATION` with `true` as the value. Thoughts?


chippers:
during conversation of the env flag, I saw a way to perhaps be able to disable it completely on webview2 which would bring it up to feature parity of webkit2gtk.  If that works, then I will flesh out the api in Wry to be able to disable it and then expose it somewhere for Tauri to be able to set it or not.  This would allow developers to completely disable it at compile time, and only enable it when building for automation specifically

chippers:
> during conversation of the env flag, I saw a way to perhaps be able to disable it completely on webview2 which would bring it up to feature parity of webkit2gtk. If that works, then I will flesh out the api in Wry to be able to disable it and then expose it somewhere for Tauri to be able to set it or not. This would allow developers to completely disable it at compile time, and only enable it when building for automation specifically

Looks like the blink renderer flag only disables setting the `window.webdriver` to `true` and not disable the WebDriver completely.

chippers:
I think that the `TAURI_AUTOMATION=true` env var is probably fine as is.  If it becomes a problem in the future, we can offer some way to set it. We can also not consider `TAURI_AUTOMATION=true` part of the stable Tauri interface so that we are free to change it in the future until we settle on a stable interface.

I'm going to un-draft this, although I suggest reviewers to run the examples a number of times each on their platform - especially the multiwindow example. I myself will be testing it on various Linux and Window variations today

lucasfernog (approved):

PR #2362, opened 2021-08-07
Tooling: [API] Changed target in tsconfig to ecma2021

### LINKED ISSUE #2360: the js api should probably be compiled for a target higher than ES3
Despite my app targeting a recent ES version, I noticed that my bundle was full of shim and that a copyright notice from Microsoft was attached to it because of said shim.
The shims and copyright notices came from importing @tauri-apps/api because `tooling/api/tsconfig.json` do not have a `target` set and the default value is `"ES3"`

Since this js will be run inside WRY, it seems safe to assume that ES6 is supported and `"target" : "ES6"` should be added to `tooling/api/tsconfig.json`

Patil2099:
Fixes #2360

### DIFF
--- a/.changes/api-target-es2021.md
+++ b/.changes/api-target-es2021.md
@@ -1,0 +1,5 @@
+---
+"api": patch
+---
+
+Change target to ES2021.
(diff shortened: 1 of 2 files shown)


nothingismagick:
Can't we shoot for ecma2018?

lucasfernog:
I just ran wry's detect_js_ecma example on all platforms and it shows we support ecma2021.

PR #2396, opened 2021-08-11
fix(core): unique WebContext for each window when TAURI_AUTOMATION unset

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No

**Other information:**


chippers:
I think we should share the webcontext regardless of the env var value so that running and testing an application have the same behavior.  Until we start seeing bugs stemming from that, I think there's more value in having them the same.

Also we should be applying our "core" custom protocols only once, and leave the rest up to the user. The current interface in Wry expects something like
```rust
let mut webview = /* WebviewBuilder stuff */;
let mut set_protocol = false;
if !set_protocol {
  webview = webview.with_custom_protocol("wry".into(), |_window, _uri| todo!());
  set_protocol = true;
}
```

Ideally we also remove the user-set custom protocols from the attributes builder and move it somewhere on the application itself because different windows handling the same URI differently is a break from the concept of what a URI is

lucasfernog:
Reopening since we couldn't find the solution yet. The temporary solution was also suggested by @wusyong @chippers

PR #2420, opened 2021-08-13
feat(core): expose `set_activation_policy`, closes #2258

### LINKED ISSUE #2258: feat(macos): expose `set_activation_policy`
### Is your feature request related to a problem? Please describe.
I would like to make a launcher / spotlight / alfred type application. Currently tauri apps have a menu and a dock icon, which isn't really compatible with this use case.

Ideally, we would have the ability to create NSPanel type windows, as described [here](https://github.com/tauri-apps/tao/issues/136). This would allow the apps to hide and show and not steal focus when they come to the foreground. 

In conjuction with the hide window functionality, this would also solve https://github.com/tauri-apps/tauri/issues/2063

However, after doing some exploration I couldn't find any reference to NSPanel in the [cocoa bindings](https://github.com/servo/core-foundation-rs) that [wry](https://github.com/tauri-apps/wry) depends on, so this might not be possible without multiple PR's and potentially a lot of work.

### Describe alternatives you've considered
[tao](https://github.com/tauri-apps/tao) has the ability to create applications without a dock menu item and menu bar, which can be seen in this example [here](https://github.com/tauri-apps/tao/blob/91a4d26fb4aa5ae13f384f7104ec5fb8a7be859e/examples/system_tray_no_menu.rs#L40). 

As wry exposes taos functionality, this should be possible in tauri, but it is not exposed in the public api. 

I think if this were to be expose by the public tauri api we could create similar apps to an NSPanel window, when used in combination with a sidecar application to manage window focus. 

### Additional context

Some more info on NSPanel:
https://developer.apple.com/documentation/appkit/nspanel
https://www.markusbodner.com/til/2021/02/08/create-a-spotlight/alfred-like-window-on-macos-with-swiftui/
https://stackoverflow.com/questions/54054473/create-spotlight-like-window-in-swift-4

What an NSPanel window would look like: 

image

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No

**Other information:**

### DIFF
--- a/.changes/runtime-set-activation-policy.md
+++ b/.changes/runtime-set-activation-policy.md
@@ -1,6 +1,6 @@
 ---
 "tauri-runtime": patch
-"tauri-runtime-wry" patch
+"tauri-runtime-wry": patch
 ---
 
 Adds `set_activation_policy` API to the `Runtime` trait.
--- a/.changes/runtime-set-activation-policy.md
+++ b/.changes/runtime-set-activation-policy.md
@@ -1,0 +1,6 @@
+---
+"tauri-runtime": patch
+"tauri-runtime-wry" patch
+---
+
+Adds `set_activation_policy` API to the `Runtime` trait.
--- a/.changes/runtime-set-activation-policy.md
+++ b/.changes/runtime-set-activation-policy.md
@@ -3,4 +3,4 @@
 "tauri-runtime-wry": patch
 ---
 
-Adds `set_activation_policy` API to the `Runtime` trait.
+Adds `set_activation_policy` API to the `Runtime` trait (macOS only).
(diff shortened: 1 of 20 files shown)


.changes/runtime-set-activation-policy.md
```diff
@@ -0,0 +1,6 @@
+---
+"tauri-runtime": patch
+"tauri-runtime-wry": patch
+---
+
+Adds `set_activation_policy` API to the `Runtime` trait.
```
amrbashir:
```suggestion
Adds `set_activation_policy` API to the `Runtime` trait (macOS only).
```

PR #2433, opened 2021-08-15
feat: add `Event::Ready`

lemarier:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

### DIFF
--- a/.changes/tauri-ready-event.md
+++ b/.changes/tauri-ready-event.md
@@ -1,0 +1,7 @@
+---
+"tauri": minor
+"tauri-runtime": minor
+"tauri-runtime-wry": minor
+---
+
+Add `Event::Ready` on the `run()` callback. Triggered once when the event loop is ready.
--- a/core/tauri-runtime-wry/src/lib.rs
+++ b/core/tauri-runtime-wry/src/lib.rs
@@ -44,7 +44,7 @@ use wry::{
       PhysicalPosition as WryPhysicalPosition, PhysicalSize as WryPhysicalSize,
       Position as WryPosition, Size as WrySize,
     },
-    event::{Event, WindowEvent as WryWindowEvent},
+    event::{Event, StartCause, WindowEvent as WryWindowEvent},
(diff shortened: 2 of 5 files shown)


amrbashir:
Seems like we are bit by bit adding new events and making the `run()` callback similar to tao's `EventLoop::run()`, should we just move all the events handling there instead of hooks on the builder ? I think it will make it easier for users since they won't have to keep cloning/moving their structs into multiple closures.

ofc removing the hooks are breaking changes so we could keep hooks and also expose all events inside `run()` and later in tauri 2.0 we could remove one of them based on what users like more or just let them coexist?

lemarier:
> Seems like we are bit by bit adding new events and making the `run()` callback similar to tao's `EventLoop::run()`, should we just move all the events handling there instead of hooks on the builder ? I think it will make it easier for users since they won't have to keep cloning/moving their structs into multiple closures.

Yeah ofc this would be great but not sure whats the main plan, but as we may support others runtime, not sure exposing tao event loop directly is safe. So it'll require anyhow some kind of wrapper.

> 
> ofc removing the hooks are breaking changes so we could keep hooks and also expose all events inside `run()` and later in tauri 2.0 we could remove one of them based on what users like more or just let them coexist?

Yes I think you are 100% right here

amrbashir:
> Yeah ofc this would be great but not sure whats the main plan, but as we may support others runtime, not sure exposing tao event loop directly is safe. So it'll require anyhow some kind of wrapper.

No, I didn't mean tao's events directly, I meant exposing tauri's events, instead of splitting some of them on the `Builder` and some of them inside `run()`

lucasfernog:
I think we're ok right now, the other events are not related to the event loop.

lucasfernog (approved):

PR #2484, opened 2021-08-18
fix: missing asset protocol path

wusyong:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**


lucasfernog (approved):

wusyong:
Let me do some final check to see if it's not required on mac/linux

core/tauri/src/manager.rs
```diff
@@ -284,6 +284,9 @@ impl<R: Runtime> WindowManager<R> {
     }
     if !webview_attributes.has_uri_scheme_protocol("asset") {
       webview_attributes = webview_attributes.register_uri_scheme_protocol("asset", move |url| {
+        #[cfg(target_os = "windows")]
+        let path = url.replace("asset://localhost/", "");
```
amrbashir, line 288: shouldn't it be 
```suggestion
        let path = url.replace("https://asset.localhost/", "");
```

wusyong: No this is in wry's context. Wry already handle that. The URL we get here will be `asset://...`

amrbashir: oh ok. didn't know that.

wusyong: But I suppose I should update the change file since it's windows only.

ArturKovacs:
Is it true that the application code will have to produce `asset://localhost/path/to/file` on Windows but it must produce `asset://path/to/file` on every other system? If this is true what's the reasoning behind this distinction?

It seems simpler to just have to specify `asset://path/to/file` on every platform.

lucasfernog:
Windows doesn't support custom protocols yet (such as `asset://`). That's why we have a `convertFileSrc` helper API.

ArturKovacs:
Sorry I may have been confused about how this should be used. So if I understand it correctly my javascript should look something like this:
```js
let video = new HTMLVideoElement();
video.src = convertFileSrc("C:/path/to/my video.mp4");
```

wusyong:
@ArturKovacs yes.
Btw we are looking into ways for streaming data. So you won't have to get entire binary for the video.

ArturKovacs:
Sounds great, thank you!

lucasfernog (approved):

PR #2546, opened 2021-08-27
refactor: rename predefined paths in `path`&`os` modules

amrbashir:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [X] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] Yes. Issue #___
- [ ] No

- [X] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix: #xxx[,#xxx]`, where "xxx" is the issue number)

**Other information:**


lucasfernog:
TBH I think the current API is just fine, we don't need to be 100% aligned with the Node.js interfaces.

amrbashir:
lol I forgot about this PR but I agree.

.changes/api-os-path-dir-naming.md
```diff
@@ -0,0 +1,8 @@
+---
+"api": patch
+---
+
+* **Breaking Change** Renamed `tempdir` to `tmpdir` (it was a typo).
+* **Breaking Change** Removed all predefined-paths functions in `path` module to be lowercase for example: `appDir` is now `appdir` (better allignment with NodeJS naming convention and os module).
```
amrbashir, line 6: what do you think about keeping all changes in this PR except this one?

amrbashir: I mean `tempDir` to `tmpDir` is logical because of `/tmp` directory on linux.

The other two are re-exports which I think are okay but we can just move `tmpDir` to path module and remove any paths exports from os module.

lucasfernog: `tempDir` is ok too, we don't need to associate with the Linux default temporary directory. I think it's ok to keep the non breaking changes.

amrbashir: moving `tempDir` to path is a breaking change  and re-exporting from path module will not follow the allowList so I will just go ahead and close the PR

PR #2555, opened 2021-08-30
Fix recommended run dev command in create-tauri-app

CYBAI:
`--dev` is not a valid option for `tauri` cli; instead, we should use
`-- dev` or just `dev`. Also, we can just use `tauri dev` for all
package managers so we don't need to do the npm checking.

| Before | After |
| --- | --- |
| image | image |

- [x] Bugfix

- [x] No

**Other information:**


CYBAI:
@wusyong could you please help me check this PR? 🙏 I've confirmed it locally and the screenshot is in the PR description. thank you!

amrbashir (changes requested):
Thanks for the PR, but please keep the double-dash when npm is used, because in some npm versions (I think npm v6), it will fail without it.

The fix should be to just add a space after the double-dash, so `'--'` should be `'-- '`  

Edit: I was wrong `--` is only need when additional arguments like `--ci` so your PR is fine

amrbashir (approved):

wusyong (approved):

CYBAI:
@amrbashir thank you so much for your confirmation! 🙏

and thanks for quick review from both of you!!

PR #2763, opened 2021-10-18
Setting the control flow to Poll by default

jelmansouri-legion:
When running the tauri loop with run_iteration, the control flow is set to exit when the even loop are cleared. But since the control flow which is persistent across frames never changes, subsequent events are not handled by Tauri. Causing the webview not to resize for example.

- [X] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [X] No

**Other information:**

### DIFF
--- a/core/tauri-runtime-wry/src/lib.rs
+++ b/core/tauri-runtime-wry/src/lib.rs
@@ -1815,6 +1815,7 @@ impl Runtime for Wry {
     self
       .event_loop
       .run_return(|event, event_loop, control_flow| {
+        *control_flow = ControlFlow::Poll;
         if let Event::MainEventsCleared = &event {
           *control_flow = ControlFlow::Exit;
         }


lucasfernog:
Thanks for catching this!

jelmansouri-legion:
Thanks for the merge, and the fix afterwards, I actually missed that Wait was used instead of Poll.
Would it make sense if the default behavior configurable, we have a game engine loop running and Poll would be better in our case:
```
// ControlFlow::Poll continuously runs the event loop, even if the OS hasn't
// dispatched any events. This is ideal for games and similar applications.

// ControlFlow::Wait pauses the event loop if no events are available to process.
// This is ideal for non-game applications that only update in response to user
// input, and uses significantly less power/CPU time than ControlFlow::Poll.
```
I can do a PR if you agree that it makes sense.
Thanks,

lucasfernog:
We've had issues with ControlFlow::Poll (100% cpu usage) so that's why we focus on Wait.

Issue #878, opened 2020-07-21
Make it more independent of project structure


ales-tsurko:
**Is your feature request related to a problem? Please describe.**

As far as I understand, Tauri can't be used without `src-tauri` directory. It's supposed to be an extension of a JS project. As for Rust-based projects, you can't omit `src-tauri` directory and should put everything related to Rust there.

For example, you already have a Cargo workspace with a structure like:
```
gui/          # JS-based GUI
some_crate_A/ # a crate, which is a part of the workspace
some_crate_B/ # another crate, which is a part of the workspace
src/          # Rust main crate, application's logic
Cargo.toml
```

But If you want to integrate Tauri into this project, you should change your project structure to something like:
```
src/              # JS-based GUI
src-tauri/                   
    some_crate_A/ # a crate, which is a part of the workspace
    some_crate_B/ # another crate, which is a part of the workspace
    src/          # Rust main crate, application's logic
    Cargo.toml
    tauri.conf.json
package.json
```

This makes the structure inconvenient for build/CI/CD scripts because you need to either change your project build tool to a JS one or add additional script at the root, or document everywhere that `src-tauri` should be considered the root and scripts should be run from there.

**Describe the solution you'd like**

Add an option to Tauri CLI to initialize a project in an existing directory. Also, it's worth providing instructions how to make it manually because the Cargo.toml and other files with same names as in the Tauri template could already exist in the project.

So you would, for example, just add `tauri.conf.json` to your project to make it possible to run `tauri dev` from its root and consider it a Tauri project. The described previously project structure would look like this after the change:

```
gui/          # JS-based GUI
some_crate_A/ # a crate, which is a part of the workspace
some_crate_B/ # another crate, which is a part of the workspace
src/          # Tauri Rust code, application's logic
Cargo.toml
tauri.conf.json
```

**Describe alternatives you've considered**
None.

**Additional context**
None.

chippers:
> Describe alternatives you've considered
None.

I think the default of belonging to the crate the `tauri.conf.json` is in is a good default.  I think an additional field in the config file to set override the app directory could be useful too so that more project layouts are possible.  Something like `tauri.conf.json > build > appDir` or w/e you wish to name the key.

nothingismagick:
We had a conversation about this yesterday at discord, and I think that this project layout is good alternative that will rise in popularity once we also make the rust-based CLI available...

boehs:
[...]

PR #2765, opened 2021-10-18
custom app and `src-tauri` folders, fix #2752

### LINKED ISSUE #2752: Custom folder for Rust source code
I want to split my codebase into two separate folders: the ui/app and a rust workspace where one of the crates is the tauri app and the rest are libraries that are used by the app.

My idea was to add options to the `package.json` and `tauri.conf.json` to specify the relative path to the respectively other config.

folder structure:

```
.
├── crates
│   ├── lib1
│   │   ├── Cargo.toml
│   │   └── src
│   ├── lib2
│   │   ├── Cargo.toml
│   │   └── src
│   └── tauri_app
│       ├── Cargo.toml
│       ├── src
│       └── tauri.conf.json
├── LICENSE
├── README.md
└── ui
    ├── package.json
    └── src
```

package.json:

```json
{
  "tauri": {
    "rustConfigPath": "../crates/tauri_app/"
  }
}
```

tari.conf.json

```json
{
  "build": {
    "appPath": "../../ui/",
  }
}
```

I found [this](https://github.com/tauri-apps/tauri/commit/a8cff6b3bc3288a53d7cdc5b3cb95d371309d2d6) commit but it requires the app and tauri app to be just one folder level apart.

crapStone:

- [ ] Bugfix
- [x] Feature (closes #2752)
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**

### DIFF
--- a/tooling/cli.rs/config_definition.rs
+++ b/tooling/cli.rs/config_definition.rs
@@ -818,6 +818,10 @@ pub struct BuildConfig {
   /// The path to the app's dist dir. This path must contain your index.html file.
   #[serde(default = "default_dist_dir")]
   pub dist_dir: AppUrl,
+  /// Directory path to app code. Can be omitted if the `package.json` is relative to the parent directory of this file.
+  /// Defaults to `None`.
+  #[serde(skip_serializing_if = "Option::is_none")]
+  pub app_dir: Option<PathBuf>,
   /// A shell command to run before `tauri dev` kicks in.
   ///
   /// The PLATFORM, ARCH, FAMILY, PLATFORM_TYPE and TAURI_DEBUG environment variables are set if you perform conditional compilation.
@@ -867,6 +871,7 @@ fn default_build() -> BuildConfig {
     runner: None,
     dev_path: default_dev_path(),
     dist_dir: default_dist_dir(),
+    app_dir: None,
     before_dev_command: None,
     before_build_command: None,
     features: None,
--- a/tooling/cli.rs/schema.json
+++ b/tooling/cli.rs/schema.json
@@ -260,6 +260,13 @@
       "description": "The Build configuration object.",
       "type": "object",
       "properties": {
+        "appDir": {
+          "description": "Directory path to app code. Can be omitted if the `package.json` is relative to the parent directory of this file. Defaults to `None`.",
+          "type": [
+            "string",
+            "null"
+          ]
+        },
         "beforeBuildCommand": {
           "description": "A shell command to run before `tauri build` kicks in.\n\nThe PLATFORM, ARCH, FAMILY, PLATFORM_TYPE and TAURI_DEBUG environment variables are set if you perform conditional compilation.",
           "type": [
--- a/tooling/cli.rs/src/build.rs
+++ b/tooling/cli.rs/src/build.rs
@@ -8,7 +8,7 @@ use heck::KebabCase;
 use tauri_bundler::bundle::{bundle_project, PackageType};
 
 use crate::helpers::{
-  app_paths::{app_dir, tauri_dir},
+  app_paths::{app_dir as get_app_dir, tauri_dir},
   command_env,
   config::{get as get_config, AppUrl},
   execute_with_output,
@@ -85,12 +85,18 @@ impl Build {
     if let Some(before_build) = &config_.build.before_build_command {
       if !before_build.is_empty() {
         logger.log(format!("Running `{}`", before_build));
+
+        let app_dir = match config_.build.app_dir.clone() {
+          Some(app_dir) => app_dir,
+          None => get_app_dir().to_owned(),
+        };
+
         #[cfg(target_os = "windows")]
         execute_with_output(
           Command::new("cmd")
             .arg("/C")
             .arg(before_build)
-            .current_dir(app_dir())
+            .current_dir(app_dir)
             .envs(command_env(self.debug)),
         )
         .with_context(|| format!("failed to run `{}` with `cmd /C`", before_build))?;
@@ -99,7 +105,7 @@ impl Build {
           Command::new("sh")
             .arg("-c")
             .arg(before_build)
-            .current_dir(app_dir())
+            .current_dir(app_dir)
             .envs(command_env(self.debug)),
         )
         .with_context(|| format!("failed to run `{}` with `sh -c`", before_build))?;
--- a/tooling/cli.rs/src/dev.rs
+++ b/tooling/cli.rs/src/dev.rs
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: MIT
 
 use crate::helpers::{
-  app_paths::{app_dir, tauri_dir},
+  app_paths::{app_dir as get_app_dir, tauri_dir},
   command_env,
   config::{get as get_config, reload as reload_config},
   manifest::{get_workspace_members, rewrite_manifest},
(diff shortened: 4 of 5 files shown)


amrbashir:
I know this is still a draft and we are thankful for your contribution but I will leave a comment anyways.

It seems like you are just adding an `appDir` field to `tauri.conf.json` but that hardly change anything. I mean `tauri dev/build/info` would still need to read the `tauri.conf.json` file to know where is the app directory and it will fail to find the file if you have it 2-levels deep in your project tree and we will be back to the same problem.

IMO, current behavior should stay the same but we add a cli option that users can use to point to the app directory.
- `tauri dev/build/info` behaves the same as it is now.
- `tauri dev/build/info --tauri-dir path/to/your/tauri/app` to use with unusual project structure.
the only downside I see for this is that `tauri info` won't be able to pickup where `package.json` is and won't show any info related to it.

crapStone:
I'm currently working on the same thing in the JS cli but there I want to add an optional config option in the `package.json` for the rust directory.

I could add both: the cli argument for the rust directory and the app directory in the `tauri.conf.json` and don't mess with the `package.json`.
My current code shouldn't break anything, unless you specifically need it it isn't even visible in the config. And if I understood it correctly the JS cli is just a thin wrapper around the Rust cli.

amrbashir:
> I'm currently working on the same thing in the JS cli but there I want to add an optional config option in the `package.json` for the rust directory.

`tauri dev/build` in the JS cli is the same as the rust cli, in fact it just passes it to a pre-compiled version of the cli.

> I could add both: the cli argument for the rust directory and the app directory in the `tauri.conf.json` and don't mess with the `package.json`.

I think the cli argument would suffice, given how much work it is to reading `appDir` from `tauri.conf.json` or from `package.json` what do you think @lucasfernog ? maybe have two cli arguments, `--tauri-config-json` and `--package-json`

> My current code shouldn't break anything, unless you specifically need it it isn't even visible in the config.

Apologies, It is not breaking anything, I didn't read your issue and thought `appDir` was referring to the tauri dir and not `package.json` so I was confused that `tauri config.json` will contain a field pointing to itself.

> And if I understood it correctly the JS cli is just a thin wrapper around the Rust cli.

yeah it is.

lucasfernog:
I think reading package.json/tauri.conf.json to check the path complicates the code a little bit, and it's not generic enough (there are some users that don't even have a package.json, so you can't rely on that to find the right tauri.conf.json path). Also a CLI argument is boring for users, having to write that all the time.

amrbashir:
The cli argument should be for people living on the edge with a really different project structure, ofc it is a tedious task to type in the path every time that's why I think most people will use it in a package.json script or a shell script at least.

kimlimjustin:
As a Tauri user who's looking for this feature, I'd like to say that this is a great feature to have, and IMO, the CLI option would be a better approach to handle this.

> Also a CLI argument is boring for users, having to write that all the time.

As a user, I don't think so. With the help of the `scripts` section of `package.json`, what you need to run is only `yarn dev/build/info`(or `npm build/build/info`). 

What I meant by the `scripts` section is something like this:
```json
"name": "tauri-app",
"scripts": {
    "dev": "tauri dev --tauri-dir path/to/app",
    "build": "tauri build --tauri-dir path/to/app",
    "info": "tauri info --tauri-dir path/to/app"
}
```

OJFord:
In my case, I have a normally-structured Tauri project (UI/app/package.json contents and all) that is a Cargo workspace member alongside a couple of others:
```
.
+-Cargo.toml
+-tauri-app
|  +-Cargo.toml
|  +-package.json
|  +-src-tauri
|  + [...]
|
+-some-lib
|  +-Cargo.toml
|  + [...]
[...]
```

i.e. if I `cd tauri-app`, everything is 'normal' and any `tauri` command will work as expected. But it's usually more convenient for me to be at the `.` level here, where the `./Cargo.toml` has contents like:
```toml
[workspace]
members = [
    "tauri-app",
    "some-lib",
]
```

so if I run `cargo tauri dev`, it of course complains it doesn't look like a Tauri project.

> Also a CLI argument is boring for users, having to write that all the time.

> ofc it is a tedious task to type in the path every time that's why I think most people will use it in a package.json script or a shell script at least.

What would be _really_ convenient would be if tauri-cli could detect that I'm in a cargo workspace, and look at the members to see which look(s) like a tauri project(s). (And error only if none are.)

However, that still wouldn't solve the layout in #2752 without `cargo` offering a change-directory arg (I'm pretty sure it doesn't).

I came across this in trying to switch from a `package.json` script which I could run as `yarn --cwd=tauri-app dev` to using `cargo tauri dev`/`cargo build` as the entrypoint. As far as I'm aware, the implementation of that is nothing more than you get from `(cd tauri-app && yarn dev)` in bash, just an added convenience.

If tauri had a similarly behaving argument - `cd` and execute as normal - then issues of varying project structure could be resolved by having to be in the appropriate directory for the command run, so using the suggested layout this would be the same `--dir` option for any command, but in #2752 might be different for `info` vs. `dev`/`build` since `src-tauri` and `package.json` are not in the same directory.

crapStone:
I like your ideas. Unfortunately I didn't have much time in the last few weeks to work on this, but during the winter holidays there should be enough time.

OJFord:
> What would be really convenient would be if tauri-cli could detect that I'm in a cargo workspace, and look at the members to see which look(s) like a tauri project(s). (And error only if none are.) [...] [or if] tauri had a similarly behaving argument - cd and execute as normal - then issues of varying project structure could be resolved by having to be in the appropriate directory for the command run, so using the suggested layout this would be the same --dir option for any command, but in #2752 might be different for info vs. dev/build since src-tauri and package.json are not in the same directory.

This actually looks to be done in a third alternative way in #2654: it's not yet in a release and I haven't tried it, but it seems `cargo tauri` will now look for a `tauri.conf.json` (recursively) in any subdirectories. This should solve the workspace member case, my layout, without caring that there's a cargo workspace or not, just the presence of that file. I suspect this would also work for your layout described in #2752 modulo the gotcha I mentioned above, as long as package.json et al. not looked for. i.e. I suspect you can get dev/build working fine given appropriate pre-commands/build.rs.

Edit: I have now tried it, and indeed it works nicely for me. It would trip over if multiple workspace members had `tauri.conf.json`s though I suppose, but not an issue for me.

CjS77:
Without this feature, you literally cannot run the examples in the tauri repo itself using `cargo tauri dev`, which really hurts when you're trying to climb the learning curve as a newcomer to the project. Even #2654 doesn't quite close the gap because there are multiple Tauri apps in the workspace.

FabianLars:
@CjS77 You won't be able to run run them with that command afterwards either, because they are configured to be executed via `cargo run --example` (with exceptions). This is mutually exclusive.

CjS77:
I'm overly generalising, apologies.

The specific issue is that the `resources` example doesn't work properly using `cargo run ...` because the resources are not bundled.

But it doesn't run using `cargo tauri dev` either because you run into either the "project thinks its in a workspace" issue or the "this isn't a Tauri project" issue depending on the folder you try run the project from.

boehs:
~~Closes #878?~~ Closed by #2654

PR #2825, opened 2021-10-30
feat(macOS): Add `AppHandle::show()` and `AppHandle::hide()`

probablykasper:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
Exposes macOS-only `show_application` and `hide_application` from tao.


.changes/mac-app-hide.md
```diff
@@ -0,0 +1,7 @@
+---
+"tauri-runtime-wry": minor
+"tauri-runtime": minor
+"tauri": minor
+---
+
+Add `AppHandle::hide()` for hiding the entire application on macOS
```
amrbashir: what about `AppHandle::show()` and `AppHandle::hide_others` ? I believe you add them to tao, right ?

probablykasper: Wasn't planning on adding `show` in this PR, but I'll go for it. They're all in tao already, yeah.

I don't really think it's worth adding `hide_others` though

amrbashir: I think you should add it, you never know if someone will need it.

probablykasper: I could, but not sure I want the method to be there if it might never even be used, not even Electron has that one. Maybe that's a weird concern though?

amrbashir: we don't need to be electron and we already expose it in TAO and this is just simply forwarding it so I don't see a reason why not.

probablykasper: Just so there's not a bunch of unused methods to sift through to find what you want. I'm not mentioning Electron because Tauri needs to be like it. Electron has hundreds of thousands of users, so if Electron is missing a simple feature, it's a good indication that nobody wants it. Either way, it's not a big deal, can add it if you still think it should be.

amrbashir: I don't have a strong opinion about it, do what you feel is right.

probablykasper: Then I'll leave it out for now. Also, added `show`.

probablykasper:
@amrbashir Any idea why the `build-artifacts` action is failing, or how to go about fixing it? The `EventLoopWindowTargetExtMacOS` trait is imported, and `app_handle.show()` does work when I run `tauri dev`.
```
error[E0599]: no method named `show_application` found for reference `&EventLoopWindowTarget<Message>` in the current scope
    --> /Users/runner/work/tauri/tauri/core/tauri-runtime-wry/src/lib.rs:1952:20
     |
1952 |         event_loop.show_application();
     |                    ^^^^^^^^^^^^^^^^ method not found in `&EventLoopWindowTarget<Message>`
```

amrbashir:
don't know really, but I think the updater-test-artifacts workflow has been broken for a while. @lucasfernog might have a better idea.

probablykasper:
@amrbashir @lucasfernog either of you wanna give this one a look so it can be merged?

lucasfernog:
> @amrbashir @lucasfernog either of you wanna give this one a look so it can be merged?

I think this one will need to wait for v2.

probablykasper:
Opened PR https://github.com/tauri-apps/tauri/pull/3689 instead, from a different branch

PR #2883, opened 2021-11-13
fix(cli.rs): terminate the beforeDevCommand, closes #2794

### LINKED ISSUE #2794: `beforeDevCommand` is not killed on app exit
### **Describe the bug**
Closing the `tauri dev` app window doesn't close the dev server started with `beforeDevCommand` (confirmed using the `create-react-app` template for `create-tauri-app`)

### **To Reproduce**
1. Create an app using `create-react-app` template
2. Run `tauri dev`
3. Close the app window using the `x` button or alt + f4
4. Run `tauri dev` again. 

### **Expected behavior**
The dev server should be gracefully shut down when the app is closed.

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No

**Other information:**


amrbashir:
steps I took to produce the issue or at least one of them: 
1. create a svelte project using `create-tauri-app`
2. run `tauri dev`
3. wait only until the svelte devServer is up and before the tauri window is shown then use `Ctrl+C` to terminate the process. 

results: 
- before this PR, the `tauri dev` process exists but it doesn't terminate the `devServer` of svelte and you can't use this terminal session anymore.
- after this PR, using `Ctrl+C` doesn't terminate the process, it instead hangs, sometimes indefinitely and sometimes until the tauri window appears but when that happens the devServer of svelte is killed.

probablykasper:
@amrbashir Could be this:
- https://github.com/sveltejs/template/issues/243

amrbashir:
yeah I think it is related, I talked about it in [discord](https://discord.com/channels/616186924390023171/731495105034321962/901043111332032512) before.

amrbashir:
This issue is driving me crazy, I managed to reproduce it with `vite` in #2998. Same steps, wait for `vite` devServer to run, press `Ctrl+C`, `tauri dev`process is terminated but not `vite` process and the terminal session is unusable.

amrbashir:
A bit of an update, in svelte official template, I found out that they are using `ignore` instead of `inherit` https://github.com/sveltejs/template/blob/4fd09c2a191b017ca9eb9cb9921fa873e95f44b4/rollup.config.js#L21

Switching it to `inherit` doesn't fix the whole issue but at least `Ctrl+C` terminates it properly.

lucasfernog:
@probablykasper can you still reproduce this issue? I couldn't after a lot of attempts.

FabianLars (dismissed):
Seems to work wonderfully on windows 🥳

probablykasper:
@lucasfernog Yes. Tried running `create-react-app` with the Svelte template, then ran `npm run tauri dev` and closed the window it opened. The `npm run tauri dev` command exits, but the `node` process is left running

lucasfernog:
I tried the same steps but I couldn't reproduce it.

probablykasper:
Unable to reproduce it both with this PR and without?

probablykasper:
Screen recording:
https://user-images.githubusercontent.com/11315492/159995391-4316158f-be50-4f8b-b986-fe4cead80886.mp4

lucasfernog:
Looks like we need a better script to kill child processes on Unix (currently `kill -TERM -P {pid}`).

probablykasper:
I don't think there's a better way to do it than that. As long as `SIGTERM` is actually called on the child process, then it should be a bug in the child process if it's not doing what it's told

Best bet might be to instead use a different Svelte template in `create-tauri-app`

lucasfernog:
Yeah but I think we can fix this on our end, I'm testing atm.

lucasfernog:
@probablykasper try again, it works now!

tooling/cli/scripts/kill-children.sh
```diff
@@ -0,0 +1,10 @@
+function getcpid() {
+    cpids=`pgrep -P $1|xargs`
+    for cpid in $cpids;
+    do
+        echo "$cpid"
+        getcpid $cpid
+    done
+}
+
+kill -9 $(getcpid $1)
```
probablykasper: The change is you're calling 9 (SIGKILL) instead of SIGTERM, right? I'm not sure that's a great idea :/

lucasfernog: Let me try with SIGTERM again

lucasfernog: hmm I think there's no other way, either we leave Svelte users with this super annoying bug or we use SIGKILL

probablykasper: I wouldn't want to risk using SIGKILL, could have some rough consequences since it takes away processes' ability to shut down properly :/

As Amr pointed out, the Svelte template might just be spawning it's child process incorrectly

lucasfernog: i'll just revert it, maybe there's a place for us to document it

PR #2916, opened 2021-11-17
Update integration.md

rdwnsjjd:
I think it is better


FabianLars:
_This was written for the first PR, but i wasn't fast enough_
1) That's the wrong branch. Should be `next`. The one you corrected isn't deployed on tauri.studio.
2) https://github.com/tauri-apps/tauri-docs/issues/291 -> It's still "broken" on the `next` branch, but as the steps are interchangeable, they should probably be called something like `1a` and `1b` or whatever.

rdwnsjjd:
I came here from website, `Edit this page`
what is wrong?

FabianLars:
I'm gonna close this because as mentioned before it's not the version of the docs that's deployed. (Just compare the headlines in this PR with the ones on the website)
> what is wrong?

The links in the docs (still) link to the `dev` branch, although it's the `next` branch that's deployed.

Issue #920, opened 2020-08-03
Init Templates / Recipes / UI Framework Choice


jbolda:
We want to do our best to remove any barriers for someone trying out Tauri. We have seen new users look to the examples repo which we have maintained thus far as a core testing playground. It has been discussed primarily on discord that we provide a command that makes testing and creating toy apps a breeze.

We have been talking about "recipes" for a bit now. We are thinking we "curry" a command like `create-react-app` and then follow up with the necessary config to set up Tauri to work with it. Like an init command for a greenfield tauri project build with X framework.

An ideal implementation would pull from a list of "templates" which list the command necessary to run and the required config in Tauri that needs to be set. In this manner, we should be able to offload most of the heavy lifting. I think it is open for discussion if this should be part of the init command or it's own separate command. I am imagining something like `tauri init --template` and this enables one of the questions to prompt which framework you would like to use.

Prompts are important so a user isn't expected to type in a framework correctly. It seems pulling from a list of templates is also important as it will enable more easily expanding that if the only requirement is a `.json` file (or equivalent).

I would love input on what others imagine this should be and how we might avoid pitfalls. My expectation is this is scoped to tauri.js and js only libs, and we can add similar as other CLI and bindings are added.

zakpatterson:
My goal for a first PR would be to allow a user to init a new project with a react dev server generated by `create-react-app`, so that it's clear how to add support for other dev frontend choices.

So far, the help and init procedure are:

```
⟩ yarn tauri init -h
yarn run v1.22.4
$ /home/projects/example/node_modules/.bin/tauri init -h
[tauri]: running init

  Description
    Inits the Tauri template. If Tauri cannot find the tauri.conf.json
    it will create one.
  Usage
    $ tauri init
  Options
    [......]
    --plugin, -p         Add UI framework. Supported: react-js
    
Done in 0.26s.

```

* [X] Set up `yarn tauri init` CLI prompt changes
* [X] Recipes can specify NPM dev and prod dependencies
* [x] Recipes can specify post-installation instructions, such as `create-react-app`
* [ ] Identify and update documentation via PR:
  * https://tauri.studio/en/docs/development/
  * https://tauri.studio/en/docs/development/integration#2-initialize-tauri

```
⟩ yarn tauri init
yarn run v1.22.4
$ /home/projects/example/node_modules/.bin/tauri init
[tauri]: running init
? What is your app name? 
? What should the window title be? Tauri App
[...]

PR #2940, opened 2021-11-23
fix(bundler): bundle additional webkit files. patch absolute paths in libwebkit*.so files. fixes #2805,#2689

### LINKED ISSUE #2689: `symbol lookup error` with the `.AppImage` on Ubuntu 19.04
### **Describe the bug**

When executing the `.AppImage` compiled with Tauri on an Ubuntu 19.40, I get this error:

```
guillem@guillem:~/Baixades$ ./holochain-launcher_0.3.2_amd64.AppImage 
Gtk-Message: 00:03:55.561: Failed to load module "canberra-gtk-module"
Gtk-Message: 00:03:55.567: Failed to load module "canberra-gtk-module"
  /usr/lib/x86_64-linux-gnu/webkit2gtk-4.0/WebKitNetworkProcess: symbol lookup error: /usr/lib/x86_64-linux-gnu/webkit2gtk-4.0/WebKitNetworkProcess: undefined symbol: NetworkProcessMainUnix
/usr/lib/x86_64-linux-gnu/webkit2gtk-4.0/WebKitWebProcess: symbol lookup error: /usr/lib/x86_64-linux-gnu/webkit2gtk-4.0/WebKitWebProcess: undefined symbol: WebProcessMainUnix
```

And I get a blank window.

Error doesn't show up if I use a `.deb` package, so I assume it's related to the bundling process.

### **To Reproduce**
1. Create a Tauri app.
2. Compile it inside an Ubuntu 18.04.
3. Try to execute the `.AppImage` in an Ubuntu 19.04.

### **Expected behavior**

The error doesn't happen.

### **Screenshots**
Captura de pantalla de 2021-09-30 00-11-01

### Tauri info
```


Operating System - Ubuntu, version 18.04 X64

Node.js environment
  Node.js - 12.22.6

Global packages
  npm - 6.14.15
  yarn - 1.22.11

Rust environment
  rustc - 1.54.0
  cargo - 1.54.0

App directory structure
/public
/tests
/.git
/.github
/node_modules
/src-tauri
/src
/scripts

App
  tauri.rs - 1.0.0-beta.8
  build-type - bundle
  CSP - default-src blob: data: filesystem: ws: http: https: tauri: 'unsafe-eval' 'unsafe-inline' 'self' img-src: 'self'
  distDir - target/webpack_dist
  devPath - target/webpack_dist
  framework - Vue.js (Vue CLI)
  bundler - Webpack
```

### LINKED ISSUE #2805: Github Action AppImage build only works on Ubuntu
### **Description**
The current AppImage build (using the GitHub Action) only works on Ubuntu. It seems to be looking for WebKit-related binaries in `/usr/lib/x86_64-linux-gnu/webkit2gtk-4.0/`, which only exists on Ubuntu.

This defeats the point of an AppImage, and in the nominal case it shouldn't be looking for any dependencies at all.

Attempting to launch it on Arch Linux fails with:
```
Gtk-Message: 20:10:51.793: Failed to load module "xapp-gtk3-module"
Gtk-Message: 20:10:51.793: Failed to load module "appmenu-gtk-module"
Gtk-Message: 20:10:51.850: Failed to load module "canberra-gtk-module"
Gtk-Message: 20:10:51.851: Failed to load module "canberra-gtk-module"

** (rite:400327): ERROR **: 20:10:52.013: Unable to fork a new child process: Failed to execute child process “/usr/lib/x86_64-linux-gnu/webkit2gtk-4.0/WebKitNetworkProcess” (No such file or directory)
```

### **Steps to reproduce**
1. Build an AppImage with the Tauri GitHub Action. [Sample AppImage](https://github.com/xyzshantaram/rite/releases/download/rite-v0.2.11/rite_0.2.11_amd64.AppImage).
2. Try to run it on a non-Ubuntu distro. I've tested on Arch and Fedora, it doesn't work on either.

### **Expected behavior**
App launches and works normally.

### **Output of `cargo tauri info`**
```
Operating System - Arch Linux, version Unknown X64

Node.js environment
  Node.js - 14.18.1

Global packages
  npm - 8.1.0
  yarn - 1.22.17

Rust environment
  rustc - 1.55.0
  cargo - 1.55.0
```

Others on the Discord (`KaKi87#2368`, `Karar#0413`) have reported this too.

Possibly related to #2689.

FabianLars:
- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

### ~~This needs to be tested on different distros before being merged!~~ :white_check_mark: 
~~Speaking of testing, fedora still won't work because of those `permission denied` errors (unrelated to this PR).~~
^Fixed this in the linuxdeploy plugin linked below.

I'm not sure if it actually fixes #2689, but it _should_ because it looks like a version mismatch between the bundled webkit and the system's one.

This is a really ugly fix imo, but i don't really think much can be improved concept wise, but who knows 🤷 

Anyway, here goes the explanation and stuff.
Basically we had 2 problems:
1) we didn't include files that webkit needs at runtime
2) the paths of those files are hardcoded into the libwebkit*.so file (hardcoded paths are a big no-no in AppImages).

The first issue's fix is pretty easy but also somewhat risky as we have to _search_ for those files as each distro stores them somewhere else. That's the main reason why we need more testing, who knows what absurd paths are out there in the wild...
- for example:
  - ubuntu: `WebKitNetworkProcess`, `WebKitWebProcess` and `injected-bundle/libwebkit2gtkinjectedbundle.so` in `/usr/lib/x86_64-linux-gnu/webkit2gtk-4.0/` 
  - fedora: `WebKitNetworkProcess` and `WebKitWebProcess` in `usr/libexec/webkit2gtk-4.0/` and `libwebkit2gtkinjectedbundle.so` in `/usr/lib64/webkit2gtk-4.0/injected-bundle/`
The injectedbundle file doesn't seem to be needed as far as i can see, but it will log an error at launch (without crashing) so probably better to just include it.

The fork of linuxdeploy-plugin-gtk is needed because this plugin will overwrite the libwebkit*.so files and therefore we can't patch those files in `template/appimage.sh`. -> This one took me too long to notice...

Link to the change: https://github.com/FabianLars/linuxdeploy-plugin-gtk/blob/master/linuxdeploy-plugin-gtk.sh#L330
And yes we have to glob pattern match the file and folder names again because they seem to differ between distros too 😒 

We can also move the fork to the tauri org, i don't really care, but i don't expect such a change to make it into upstream at all.
Furthermore i won't even try to get something like this into webkit itself, that's no bueno for my mental health lol.

## Update 1
1. We now use mkdir to create the folders before copying to make them writable. Otherwise only the first file will be copied.
2. The `find` command can throw "permission denied" errors on protected folders. Normally, this would be fine as it just continues after throwing them. In this case we need to add `|| true` to not trigger the `set -euxo pipefail` guard. Unfortunately this will skip the exit on other errors from these commands too, so we can still end up with broken AppImages after a "successful" script execution, but at least they will be logged when invoked with `--verbose`.

## Update 2
Fixed the `os error 13` in the gtk plugin which crashed the appimage script on fedora (and probably on other distros too).
This also fixes the appimage folder being undeletable on the following runs which also crashed the bundler but even earlier.

## Update 3
Test results are coming in, check comment down below.

## How to test
0. (Optional) Clone the fork and run the examples instead of step 1&2
1. Install the rust based cli from the fork: `cargo install tauri-cli --force --git https://github.com/FabianLars/tauri --branch fix/appimage` -> remember to revert/remove this afterwards :)
2. Use `cargo tauri build` to build your project, the npm/yarn command won't work. (You may need to use tauri's next branch)
3. Copy the AppImage to another system (needs to be another distro or at least another version). Alternatively you could temporarily rename the above mentioned files if you find them.

### DIFF
--- a/.changes/bundler-fix-appimage.md
+++ b/.changes/bundler-fix-appimage.md
@@ -1,0 +1,5 @@
+---
+'tauri-bundler': patch
+---
+
+Fixes AppImage crashes caused by missing WebKit runtime files.
--- a/tooling/bundler/src/bundle/linux/templates/appimage
+++ b/tooling/bundler/src/bundle/linux/templates/appimage
@@ -12,10 +12,10 @@ cp -r ../appimage_deb/data/usr "{{app_name}}.AppDir"
 
 cd "{{app_name}}.AppDir"
 
-# Copy WebKit files
-find /usr/lib* -name WebKitNetworkProcess -exec cp --parents '{}' "." \;
-find /usr/lib* -name WebKitWebProcess -exec cp --parents '{}' "." \;
-find /usr/lib* -name libwebkit2gtkinjectedbundle.so -exec cp --parents '{}' "." \;
+# Copy WebKit files. We need to pipe errors to dev/null, so that the find
+find /usr/lib* -name WebKitNetworkProcess -exec mkdir -p $(dirname '{}') \; -exec cp --parents '{}' "." \; || true
+find /usr/lib* -name WebKitWebProcess -exec mkdir -p $(dirname '{}') \; -exec cp --parents '{}' "." \; || true
+find /usr/lib* -name libwebkit2gtkinjectedbundle -exec mkdir -p $(dirname '{}') \; -exec cp --parents '{}' "." \; || true
 
 wget -q -4 -O AppRun https://github.com/AppImage/AppImageKit/releases/download/continuous/AppRun-x86_64 || wget -q -4 -O AppRun https://github.com/AppImage/AppImageKit/releases/download/12/AppRun-aarch64
 chmod +x AppRun
--- a/tooling/bundler/src/bundle/linux/templates/appimage
+++ b/tooling/bundler/src/bundle/linux/templates/appimage
@@ -12,7 +12,7 @@ cp -r ../appimage_deb/data/usr "{{app_name}}.AppDir"
 
 cd "{{app_name}}.AppDir"
 
-# Copy WebKit files. We need to pipe errors to dev/null, so that the find
+# Copy WebKit files.
 find /usr/lib* -name WebKitNetworkProcess -exec mkdir -p $(dirname '{}') \; -exec cp --parents '{}' "." \; || true
 find /usr/lib* -name WebKitWebProcess -exec mkdir -p $(dirname '{}') \; -exec cp --parents '{}' "." \; || true
 find /usr/lib* -name libwebkit2gtkinjectedbundle.so -exec mkdir -p $(dirname '{}') \; -exec cp --parents '{}' "." \; || true
--- a/tooling/bundler/src/bundle/linux/templates/appimage
+++ b/tooling/bundler/src/bundle/linux/templates/appimage
@@ -12,6 +12,11 @@ cp -r ../appimage_deb/data/usr "{{app_name}}.AppDir"
 
 cd "{{app_name}}.AppDir"
 
+# Copy WebKit files
+find /usr/lib* -name WebKitNetworkProcess -exec cp --parents '{}' "." \;
+find /usr/lib* -name WebKitWebProcess -exec cp --parents '{}' "." \;
+find /usr/lib* -name libwebkit2gtkinjectedbundle.so -exec cp --parents '{}' "." \;
+
 wget -q -4 -O AppRun https://github.com/AppImage/AppImageKit/releases/download/continuous/AppRun-x86_64 || wget -q -4 -O AppRun https://github.com/AppImage/AppImageKit/releases/download/12/AppRun-aarch64
 chmod +x AppRun
 
@@ -22,7 +27,7 @@ ln -s "usr/share/applications/{{app_name}}.desktop" "{{app_name}}.desktop"
 
 cd ..
 
-wget -q -4 -O linuxdeploy-plugin-gtk.sh "https://raw.githubusercontent.com/linuxdeploy/linuxdeploy-plugin-gtk/master/linuxdeploy-plugin-gtk.sh"
+wget -q -4 -O linuxdeploy-plugin-gtk.sh "https://raw.githubusercontent.com/FabianLars/linuxdeploy-plugin-gtk/master/linuxdeploy-plugin-gtk.sh"
 wget -q -4 -O linuxdeploy-x86_64.AppImage https://github.com/linuxdeploy/linuxdeploy/releases/download/continuous/linuxdeploy-x86_64.AppImage
 
 chmod +x linuxdeploy-plugin-gtk.sh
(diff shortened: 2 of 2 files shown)


FabianLars:
So i'm getting pretty confident so far, so here are test result i have so far.
## Testresults
### what i personally tested
- built on ubuntu 20.04:
  - works on manjaro and fedora 35
- built on manjaro:
  - works on fedora 35 (can't work on ubuntu 20.04 because of glibc)
- built on fedora 35
  - works on another (fresh) fedora system lol (glibc is even newer than manjaro's, so i can't test that)

### what others tested
- built on debian
  - works on arch https://discord.com/channels/616186924390023171/731495084834684928/913080942954963004
- building on centos 7.3 works

lucasfernog:
Looking good!

FabianLars:
Sooo, i'd like to prepare this PR to be **ready** for merging, because it looks like it's working pretty reliable and doesn't introduce any regressions and therefore shouldn't need any substantial changes anymore. There are 2 things to talk about first tho.

<x>1)</x> Moving the plugin fork to the tauri org:
Should i transfer it (== 6 commits diff) or create a new fork directly in the org (== 1 commit diff)? I'm fine with both.

<x>2)</x> Possible code changes:
Basically 3 possibilities:
- <x>1.</x> We leave it all as it is.
- <x>2.</x> We move the template changes into the gtk plugin instead.
  - This _can_ (but shouldn't) reintroduce folder permission issues again, because linuxdeploy and/or the plugin like to f up the folder creation. So maybe we need to add another chmod call 🤷 
  - This _can_ (but shouldn't) introduce other issues if the gtk plugin did some funny stuff with the binaries or something (because they were already there when the plugin ran). Sounds really unlikely after i read it out loud...
  - Probably better in a "separation of concerns" way or whatever it's called.
- ~~3. We create another plugin just for that webkit stuff~~
  - Actually, scratch that, i tried it a little bit and it's not worth the hassle:
    - a) it's too risky as in "it's actually less reliable", at least in some simple testing i did after the discord talk.
    - b) we need to duplicate the permission code fixes.
    - c) we still need to fork the gtk plugin which is reason enough imo.
    - also webkit**gtk** is gtk related enough for 2. to make more sense than 3. i guess

So yeah either 1. or 2., I'm fine with both, but still leaning towards 1. cause it looks like the safest bet 🤔

lucasfernog:
I'm fine with a direct transfer of the plugin.

xyzshantaram:
@FabianLars absolute legend

TexZeTech:
So I know this is not the same project but I would really like some help trying to figure it out. I'm using Orca slicer for my klipper printer and I cannot get the Device tab to display the mainsail interface anymore. I believe it's related to one of the issues in this thread, at this point, any and all help would be appreciated.

[OrcaSlicer #920](https://github.com/SoftFever/OrcaSlicer/issues/920)

Thank you for your time.

PR #2997, opened 2021-12-01
fix: extract an entire source archive into a specified path is not working if the archive includes dirs.

xuchaoqian:
- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**
We use zip as our primary archive format for all platforms. On macOS, the abc.app is a directory, run zip command will output abc.app.zip, then extract it, we should get the abc.app directory again, but Tauri does not work as expected now. The patch fixed this bug.


amrbashir (approved):

core/tauri/src/api/file/extract.rs
```diff
@@ -119,9 +119,31 @@ impl<'a> Extract<'a> {
...
-          let path = into_dir.join(file.name());
-          let mut output = fs::File::create(path)?;
-          io::copy(&mut file, &mut output)?;
+          // Decode the file name from raw bytes instead of using file.name() directly.
+          // file.name() uses String::from_utf8_lossy() which may return messy characters
+          // such as: τê▒Σ║ñµÿô.app/, that does not work as expected.
+          // Here we require the file name must be a valid UTF-8.
+          let file_name = String::from_utf8(file.name_raw().to_vec())?;
+          let out_path = into_dir.join(&file_name);
+          if file_name.ends_with('/') || file_name.ends_with('\\') {
```
amrbashir: can't you just check `file.is_dir()` instead ?

core/tauri/src/api/file/extract.rs
```diff
@@ -119,9 +119,31 @@ impl<'a> Extract<'a> {
...
+            fs::create_dir_all(&out_path)?;
+          } else {
+            if let Some(out_path_parent) = out_path.parent() {
+              if !out_path_parent.exists() {
```
amrbashir: No need to check if parent exists or not, just use `fs::create_dir_all()` directly, it will handle it.

amrbashir (changes requested):

core/tauri/src/api/file/extract.rs
xuchaoqian: They are basically the same in this case, because both / and \ are ASCII characters. file.is_dir() uses file.name() internally, here we check it ourself just wants the code logic be consistent.

>

core/tauri/src/api/file/extract.rs
xuchaoqian: It's necessary.  Because some zip archives may not contain directory entries.

core/tauri/src/api/file/extract.rs
amrbashir: If that is the case then `file.is_dir()` is better to use here, it is a lot more concise.

core/tauri/src/api/file/extract.rs
amrbashir: that's not what I meant. 
```rs
// at this point we are dealing with a file
if let Some(out_path_parent) = out_path.parent() {
	if !out_path_parent.exists() {
	}
}
```

amrbashir: that's not what I meant. 
```rs
// at this point we are dealing with a file
if let Some(out_path_parent) = out_path.parent() {
	// at this point the out path has a parent directory that needs to be created first
	// here you call `fs::create_dir_all()` only if this parent doesn't exits
	// what I am saying is you don't need to check if it exists or not because `fs::create_dir_all()` will handle that. 
	if !out_path_parent.exists() {
		fs::create_dir_all(&out_path_parent)?;
	}
}
```
you can just do this 
```rs
if let Some(out_path_parent) = out_path.parent() {
	fs::create_dir_all(&out_path_parent)?;
}
```

core/tauri/src/api/file/extract.rs
xuchaoqian: Done.

core/tauri/src/api/file/extract.rs
xuchaoqian: Done.

core/tauri/src/async_runtime.rs
```diff
@@ -123,7 +124,11 @@ mod tests {
   #[tokio::test]
   async fn handle_abort() {
     let handle = handle();
-    let join = handle.spawn(async { 5 });
+    let join = handle.spawn(async {
+      // Here we sleep 1 second to ensure this task to be uncompleted when abort() invoked.
+      tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+      5
+    });
```
amrbashir, line 333: It is irrelevant to this PR but I am gonna approve it anyway and wait for @lucasfernog to approve it too.

amrbashir (approved):

core/tauri/src/async_runtime.rs
xuchaoqian: Thank you for your review.

lucasfernog: yeah i just saw this test fail on windows, thanks for fixing it!

lucasfernog (approved):

PR #3031, opened 2021-12-06
feat: detect `data-tauri-drag-region` on parentElement, closes #3030

### LINKED ISSUE #3030: feat: detect `data-tauri-drag-reigon` on parent also
It seems that vuetify (a vue component framework) `v-app-bar` renders two HTML elements, the outer one will have the attribute but the inner one, won't. When you click on a `v-app-bar`, the element that receives the click event is the inner one, and Tauri doesn't find that it has `data-tauri-drag-region` so it won't start dragging. you need to find a way to add it to the inner one.

_Originally posted by @amrbashir in https://github.com/tauri-apps/tauri/discussions/3026#discussioncomment-1756684_

amrbashir:

- [x] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

**Other information:**


amrbashir:
I also updated the formatting of some files in the repo, just a `yarn format` in repo root

FabianLars:
Hmm i didn't test it yet, but this does sound "scary", what if the actual target is a button or link or whatever clickable (or user-made clickable) thing? This seems to introduce many possible side effects 🤔

amrbashir:
@FabianLars well spotted, thanks. If we also check for parent, then the buttons in a custom titlebar for example would stop working. I don't think there is a way around this issue.

PR #3035, opened 2021-12-06
feat(cli.js): allow configuring tauri app path on package.json #2752

lucasfernog:
`package.json > tauri > appPath` defines the path to the tauri crate and cli.js sets the proper CWD when running cli.rs

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [ ] No

**Other information:**

### DIFF
--- a/tooling/cli.js/bin/tauri.js
+++ b/tooling/cli.js/bin/tauri.js
@@ -5,6 +5,7 @@
 
 import chalk from 'chalk'
 import updateNotifier from 'update-notifier'
+import { existsSync, readFileSync } from 'fs'
 import { createRequire } from 'module'
 const require = createRequire(import.meta.url)
 const pkg = require('../package.json')
@@ -74,10 +75,21 @@ ${chalk.yellow('Options')}
     if (process.argv && process.env.NODE_ENV !== 'test') {
       process.argv.splice(0, 3)
     }
+    let cwd = null
+    if (existsSync('package.json')) {
+      const packageJson = JSON.parse(readFileSync('package.json').toString())
+      if ('tauri' in packageJson) {
+        const { tauri: tauriConfig } = packageJson
+        if (tauriConfig.appPath) {
+          cwd = tauriConfig.appPath
+        }
+      }
+    }
     ;(
       await runOnRustCli(
         command,
-        (process.argv || []).filter((v) => v !== '--no-update-notifier')
+        (process.argv || []).filter((v) => v !== '--no-update-notifier'),
+        { cwd }
       )
     ).promise
       .then(() => {
--- a/tooling/cli.js/bin/tauri.js
+++ b/tooling/cli.js/bin/tauri.js
@@ -87,7 +87,7 @@ ${chalk.yellow('Options')}
           cwd = resolvePath(dirname(pkgJsonPath), tauriConfig.appPath)
           console.log(cwd)
           if (!existsSync(cwd)) {
-            console.error(`Configured appPath '${cwd}' does not exist.`)
+            console.error(`Configured appPath in package.json '${cwd}' does not exist.`)
             process.exit(1)
           }
         }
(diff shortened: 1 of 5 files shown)


amrbashir:
if we are gonna add this, then we should probably add the other way around too.
So `package.json` has a key to link to `tauri.conf.json` (the file itself is better than the directory) and `tauri.conf.json` has link to `package.json`, it probably better to bake that in the rust cli, so rust users can have its benefits too.

lucasfernog:
> if we are gonna add this, then we should probably add the other way around too. So `package.json` has a key to link to `tauri.conf.json` (the file itself is better than the directory) and the other way around.

I think the Rust CLI won't even know about Node.js or the frontend application. Currently it only needs this information for `tauri info` and `beforeDev/beforeBuildCommand`, and my idea was to allow defining the `cwd` for the scripts and letting `cli.js` pass information to `cli.rs`. This keeps the `cli.rs` code clean, avoiding this back and forth path information, and makes it easier to have new bindings in the future.

amrbashir:
> Currently it only needs this information for `tauri info` and `beforeDev/beforeBuildCommand`, and my idea was to allow defining the `cwd` for the scripts and letting `cli.js` pass information to `cli.rs`. 

yeah you're right, this makes more sense.

We also had some reports about some people not able to use the rust cli with a cargo workspace but that's for another PR I guess.

lucasfernog:
Can you link that issue? Last time I checked it worked fine, maybe it's some special setup.

amrbashir:
don't remember exactly, maybe it was this https://github.com/tauri-apps/tauri/pull/2765#issuecomment-962494040 .

tooling/cli.js/bin/tauri.js
```diff
@@ -74,10 +75,21 @@ ${chalk.yellow('Options')}
     if (process.argv && process.env.NODE_ENV !== 'test') {
       process.argv.splice(0, 3)
     }
+    let cwd = null
+    if (existsSync('package.json')) {
```
amrbashir: we could enhance this a bit using https://github.com/sindresorhus/pkg-up , I know I sometimes run npm clis in a nested directory. this would be a cool addition IMO.

amrbashir (changes requested):

tooling/cli.js/bin/tauri.js
lucasfernog: good idea, gonna do it tomorrow

probablykasper: pkg-up is literally just calling `findUp('package.json', {cwd})` from https://github.com/sindresorhus/find-up, so I'd say use `find-up` directly

lucasfernog: done

tooling/cli.js/bin/tauri.js
```diff
@@ -74,10 +77,27 @@ ${chalk.yellow('Options')}
...
+    const pkgJsonPath = findUpSync('package.json')
+    if (pkgJsonPath) {
+      const packageJson = JSON.parse(readFileSync(pkgJsonPath).toString())
+      if ('tauri' in packageJson) {
+        const { tauri: tauriConfig } = packageJson
+        if (tauriConfig.appPath) {
+          cwd = resolvePath(dirname(pkgJsonPath), tauriConfig.appPath)
+          console.log(cwd)
+          if (!existsSync(cwd)) {
+            console.error(`Configured appPath '${cwd}' does not exist.`)
```
amrbashir:
```suggestion
            console.error(`Configured appPath in package.json '${cwd}' does not exist.`)
```

amrbashir (approved):

PR #3076, opened 2021-12-15
fix(bundler): detect per-user webview2 installations

### LINKED ISSUE #3113: [bug] cargo tauri info always shows "Webview2 & Visual Studio Build Tools" not installed
### Describe the bug

cargo tauri info always shows "Webview2 & Visual Studio Build Tools" not installed. 

But actually I have installed them already:
image

image

### Reproduction

_No response_

### Expected behavior

_No response_

### Platform and versions

```shell
Operating System - Windows, version 10.0.19043 X64
Webview2 - Not installed
Visual Studio Build Tools - Not installed

Node.js environment
  Node.js - 14.18.2

Global packages
  npm - 6.14.15
  pnpm - Not installed
  yarn - 1.22.17

Rust environment
  rustc - 1.57.0
  cargo - 1.57.0

Rust environment
  rustup - 1.24.3
  rustc - 1.57.0
  cargo - 1.57.0
  toolchain - stable-x86_64-pc-windows-msvc (environment override by RUSTUP_TOOLCHAIN)

App directory structure
/build
/node_modules
/public
/src
/src-tauri

App
  tauri.rs - 1.0.0-beta.8
  build-type - bundle
  distDir - ../build
  devPath - http://localhost:3000/
  framework - React
```

### Stack trace

_No response_

### Additional context
1) I use the next tauri-cli: cargo install tauri-cli --git https://github.com/tauri-apps/tauri.git --branch next
2) And the installer will download Webview2 runtime every time, even if I have run the installer many times:
image

FabianLars:
Edit: This partially resolves #3113

I have no idea about wix, so this may be not the right way and i wasn't able to test this yet, sorry.
I removed the `win64="no"` thing because (at least on 64bit) the per-user key is not in `WOW6432Node`.

For reference:
- https://github.com/MicrosoftEdge/WebView2Announcements/issues/37
- https://github.com/MicrosoftEdge/WebView2Feedback/issues/2019#issuecomment-994089021

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding Issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes. Issue #___
- [x] No

### DIFF
--- a/tooling/bundler/src/bundle/windows/templates/main.wxs
+++ b/tooling/bundler/src/bundle/windows/templates/main.wxs
@@ -219,7 +219,7 @@
         {{#if install_webview}}
         <!-- WebView2 -->
         <Property Id="WVRTINSTALLED">
-            <RegistrySearch Id="WVRTInstalledSystem" Root="HKLM" Key="SOFTWARE\Microsoft\EdgeUpdate\Clients\{F3017226-FE2A-4295-8BDF-00C3A9A7E4C5}" Name="pv" Type="raw" />
+            <RegistrySearch Id="WVRTInstalledSystem" Root="HKLM" Key="SOFTWARE\Microsoft\EdgeUpdate\Clients\{F3017226-FE2A-4295-8BDF-00C3A9A7E4C5}" Name="pv" Type="raw" Win64="no" />
             <RegistrySearch Id="WVRTInstalledUser" Root="HKCU" Key="SOFTWARE\Microsoft\EdgeUpdate\Clients\{F3017226-FE2A-4295-8BDF-00C3A9A7E4C5}" Name="pv" Type="raw"/>
         </Property>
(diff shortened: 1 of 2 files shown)


FabianLars:
Weird, it didn't seem to fail locally 🤔 brb

Edit: I really really have to stop using the github editor for small commits if i'm too stupid to copy-paste my local changes...

lucasfernog:
I think we should use the default value for Win64 instead of `no`. From the [docs](https://wixtoolset.org/documentation/manual/v3/xsd/wix/registrysearch.html):

```
Instructs the search to look in the 64-bit registry when the value is 'yes'. When the value is 'no', the search looks in the 32-bit registry. The default value is based on the platform set by the -arch switch to candle.exe or the InstallerPlatform property in a .wixproj MSBuild project: For x86 and ARM, the default value is 'no'. For x64, ARM64, and IA64, the default value is 'yes'.
```

lucasfernog (approved):

FabianLars:
gimme a second, i'll push the same change for the tauri info command

FabianLars:
> I think we should use the default value for Win64 instead of no. From the docs:

yeah i read the docs on that too, but i didn't get it at all lol, but because the key is actually in WOW6432Node i thought i leave it how it was for the machine-wide key and remove it for the user-wide one (because that one isn't in WOW6432Node.
This probably needs testing 🤔

FabianLars:
yeah, had to revert it. It wasn't able to detect machine-wide installations (on 64bit systems) without `Win64="no"`.

The user-wide installation stuff seems to work for wix and `tauri info`. Although i can't test it on 32-bit, but i didn't find any special info on that so i guess it's the same key.

Issue #3137, opened 2021-12-30
[bug] loading local assets which include url query strings via `tauri.convertFileSrc` fail to load


christian-schulze:
### Describe the bug

The `asset` custom protocol implementation appears to treat URL query strings as part of the literal file path, so fails to load the asset.

I've tracked this down to [`core/tauri/src/manager.rs#323`](https://github.com/tauri-apps/tauri/blob/next/core/tauri/src/manager.rs#L322):
```rust
pending.register_uri_scheme_protocol("asset", move |request| {
  #[cfg(target_os = "windows")]
  let path = request.uri().replace("asset://localhost/", "");
  #[cfg(not(target_os = "windows"))]
  let path = request.uri().replace("asset://", "");
  let path = percent_encoding::percent_decode(path.as_bytes())
    .decode_utf8_lossy()
    .to_string();
  let path_for_data = path.clone();
 ...
```
I've been able to hack a temporary solution by using a regex to filter out the query string, the new code looks like this:
```rust
pending.register_uri_scheme_protocol("asset", move |request| {
  #[cfg(target_os = "windows")]
  let path = request.uri().replace("asset://localhost/", "");
  #[cfg(not(target_os = "windows"))]
  let path = request.uri().replace("asset://", "");
  let re = Regex::new(r"\?.*").unwrap();
  let path = re.replace_all(&path, "");
  let path = percent_encoding::percent_decode(path.as_bytes())
    .decode_utf8_lossy()
    .to_string();
  let path_for_data = path.clone();
```

### Reproduction

1. Use `yarn create tauri-app` to create a React app using Vite.
2. Create a local `index.html` file and `code.js` file.
3. Modify the `index.html` file to load the `code.js` file via a `script` tag, but append `?asdf` to the `src` attribute.
4. Modify the `code.js` file to log a message using `console.log`.
5. Modify the `App` react component to render an iframe which points to the `index.html` file, passing the path through `tauri.convertFileSrc`.

You *should* see the console.log message you added, but instead there is an error.

### Expected behavior

Files loaded into an iframe via `tauri.convertFileSrc` should be able to load their own assets which include url query strings.

### Platform and versions

```shell
Operating System - Arch Linux, version Unknown X64

Node.js environment
  Node.js - 16.11.0
  @tauri-apps/cli - 1.0.0-beta.10
  @tauri-apps/api - 1.0.0-beta.8

Global packages
  npm - 8.1.3
  yarn - 1.22.17

Rust environment
[...]

PR #3141, opened 2021-12-30
fix: #3137 - remove querystring when using asset protocol

christian-schulze:

issue: #3137

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist
- [x] When resolving issues, they are referenced in the PR's title (e.g `fix #___, #___`)

### Other information

This resolves an issue where the asset protocol loader would treat url querystrings as part of the literal file path. A scenario where this is an issue is when loading a local html file into an iframe, and that iframe contains script tags which include query strings in the src attributes.

attention @amrbashir

### DIFF
--- a/core/tauri/Cargo.toml
+++ b/core/tauri/Cargo.toml
@@ -73,6 +73,7 @@ raw-window-handle = "0.4.2"
 minisign-verify = { version = "0.2", optional = true }
 os_info = { version = "3.0.9", optional = true }
 futures-lite = "1.12"
+regex = "1"
 
 [target."cfg(any(target_os = \"linux\", target_os = \"dragonfly\", target_os = \"freebsd\", target_os = \"openbsd\", target_os = \"netbsd\"))".dependencies]
 gtk = { version = "0.14", features = [ "v3_20" ] }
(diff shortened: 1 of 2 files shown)


FabianLars:
hmm, i honestly don't like to pull in the whole regex crate just for splitting at a single symbol :thinking: 
Also, can this break legit use cases? for example on linux, the question mark is **allowed** in file and folder names

lucasfernog:
Yeah I tried to use split instead of regex but I just thought about the path problem; Windows can also use the `\\?\` prefix IIRC.

lucasfernog:
I guess we should use `url::Url::parse` instead and pull the path.

christian-schulze:
@FabianLars @lucasfernog I've pushed a change to use `url::Url`, appears to work as desired. Is there someone who can verify this on Windows?

lucasfernog:
I'll try this on Windows.

lucasfernog:
Works well on Windows, nice catch with `[..Position::AfterPath]`. Won't work with canonicalized paths, but I think that won't be a problem. We can rethink this if someone requests it, the querystring/fragment parsing is more important right now IMO.

christian-schulze:
@lucasfernog Thanks for your attention reviewing this. Please let me know if there are any other changes needed to get this merged.

Happy new year :tada:

lucasfernog:
Happy new year @christian-schulze thanks for your contribution!

PR #3157, opened 2022-01-04
Added example for CPP Dll useage

Hecatron:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [X] Other, please describe:

This is basically an example of using tauri within a dll library (windows).
So that the webview is generated / run from within the dll.

The idea being you can run / call the dll from an existing C / C++ project for interop with the project.

Nope


nothingismagick:
Really, really interesting!

Hecatron:
I was hoping to use this method to write a freecad extension which normally uses python / dll's normally written in c++
the idea being to write a rust / tarui dll library that could be loaded instead.

nothingismagick:
Amazing

examples/cpp-dll/src-app1/src/main.rs
```diff
@@ -0,0 +1,22 @@
...
+
+use libloading::{Library, Symbol};
+type LibFunctionType1 = fn();
+
+fn main() {
+  let library_path = "../src-dll1/target/debug/dll1.dll";
+  println!("Loading lib_test1() from {}", library_path);
+
+  unsafe {
+    let lib = Library::new(library_path).unwrap();
```
amrbashir: use `library_filname()` function for cross-platform support
```suggestion
use libloading::{Library, library_filename, Symbol};
type LibFunctionType1 = fn();

fn main() {
  let library_path = library_filename("../src-dll1/target/debug/dll1");
  println!("Loading lib_test1() from {}", library_path);

  unsafe {
    let lib = Library::new(library_path).unwrap();
```

examples/cpp-dll/src-app1/src/main.rs
```diff
@@ -0,0 +1,22 @@
...
+    let _func: Symbol<LibFunctionType1> = lib.get(b"lib_test1").unwrap();
```
amrbashir: Could use a better name
```suggestion
    let run_tauri: Symbol<LibFunctionType1> = lib.get(b"run_tauri").unwrap();
```

examples/cpp-dll/src-dll1/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "dll1"
```
amrbashir: Also could use a better name, and the directory too should be called `src-tauri`
```suggestion
name = "tauri-app"
```

examples/cpp-dll/src-dll1/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+// This is an example of a taui app built into a dll
+// Calling lib_test1 within the dll will launch the webview
+
+#![cfg_attr(
+  all(not(debug_assertions), target_os = "windows"),
+  windows_subsystem = "windows"
+)]
+
+#[no_mangle]
+pub extern fn lib_test1() {
```
amrbashir:
```suggestion
pub extern fn run_tauri() {
```

amrbashir (changes requested):
Thank you for this cool example

examples/cpp-dll/src-dll1/Cargo.toml
Hecatron: I've used tauri_app just to follow the snake_case as that's what the build outputs

examples/cpp-dll/src-dll1/src/lib.rs
Hecatron: Now renamed

examples/cpp-dll/src-app1/src/main.rs
Hecatron: Now renamed

examples/cpp-dll/src-app1/src/main.rs
Hecatron: Thanks for this one, I wanted to do it this way but didn't know about library_filename

Hecatron:
Let me know if anything else needs changing

amrbashir (changes requested):
A couple more things to change: 
- the example directory name from `cpp-dll` to `tauri-dynamic-lib`
- the read me in this example should have the name changes you just made, and it should not be focused on Windows dll or C/C++. The wording should be more general, it doesn't matter which language consumes this dynamic lib nor what OS it is on.

Hecatron:
Ok made some more changes, let me know if you spot anything else

examples/tauri-dynamic-lib/Readme.md
```diff
@@ -0,0 +1,36 @@
+# Readme
+
+This is an example of running tauri within a dynamic shared library.
```
amrbashir, line 3:
```suggestion
This is an example of compiling tauri as a dynamic shared library and running it from another app.
```

amrbashir:
```suggestion
This is an example of compiling tauri as a dynamic shared library and running it from another app.
```

examples/tauri-dynamic-lib/Readme.md
```diff
@@ -0,0 +1,36 @@
...
+
+  * src-tauri is an example of a library containing code to launch a tauri webview window.
+  * src-app1 is a small example of calling the function within the library.
```
amrbashir:
```suggestion
  * src-app1 is a small example of calling tauri from a dynamic shared library through FFI.
```

amrbashir (changes requested):

examples/tauri-dynamic-lib/Readme.md
Hecatron: now updated

examples/tauri-dynamic-lib/Readme.md
Hecatron: now updated

lucasfernog:
Maybe the example could also show how to interop with C code?

Hecatron:
> Maybe the example could also show how to interop with C code?

It may be different based on which platform you're using which might complicate things.
(which is why I did it in rust as that has a lot of polish / abstracts away the implementation)
Typically in real life you'd probably use something like https://github.com/eqrion/cbindgen
or https://github.com/rust-lang/rust-bindgen or https://cxx.rs/ for the interop to generate the needed headers

But basically all you're doing from the C side is calling a function in a dll or shared library.
But with C there's lots of different ways to do the same thing based on the platform and libs in use.

What I was more interested in to be honest was if this sort of implementation would mess up the bundling or the rust build side of things since tauri uses custom build scripts, fortunately it seems to all just work which is nice.

amrbashir (approved):
LGTM! but I will wait for Lucas to approve before merge.

PR #3158, opened 2022-01-05
feat(bundler): initialize msi install path with previous location

FabianLars:
With this PR we now save the installation directory to the registry and use that value to initialize the default install dir on following installations (upgrades).
This was requested a few times so that tauri's buit-in updater doesn't default to `C:\Program Files\app_name` (ofc useful for manual update-installations too).

The default path if the reg key doesn't exist is the same as before.

I intentionally added the reg value stuff in a separate `Component` (for clarity) and inside a `RegistryKey` object (for extensibility).

With this change i also changed `manufacturer` to be the middle part of the bundle identifier instead of the whole thing. This was imho just plain wrong.
The whole bundle identifier is still used as the `AppId`, mainly because notifications don't work without it.

Furthermore I changed the dummy registry values used as keyPaths for the shortcuts to be all in the same registry "folder".
Instead of having an `installed` key in 3 locations, we now have 3 different (dummy) keys in the same path, much cleaner imo.

Just to be extra clear, this is not a breaking change. New installers are still compatible with previous ones. Upgrading old installations with the new installer works normally (old reg keys will be removed by the old uninstaller).

- [x] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/.changes/bundler-msi-init-installdir.md
+++ b/.changes/bundler-msi-init-installdir.md
@@ -1,0 +1,5 @@
+---
+"tauri-bundler": patch
+---
+
+Initialize the preselected installation path with the location of the previous installation.
--- a/tooling/bundler/src/bundle/windows/msi/wix.rs
+++ b/tooling/bundler/src/bundle/windows/msi/wix.rs
@@ -465,8 +465,10 @@ pub fn build_wix_app_installer(
 
   data.insert("product_name", to_json(settings.product_name()));
   data.insert("version", to_json(settings.version_string()));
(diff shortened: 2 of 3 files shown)


lucasfernog (commented):
could use a change file here

FabianLars:
should it mention the registry key too or is this enough?

lucasfernog (approved):
LGTM

PR #3165, opened 2022-01-05
Initial support of egui native window

wusyong:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/.changes/egui.md
+++ b/.changes/egui.md
@@ -1,0 +1,6 @@
+---
+"tauri": patch
+"tauri-runtime-wry": patch
+---
+
+Add method to create egui window. Users could enable this via `egui` feature.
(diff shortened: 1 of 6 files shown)


amrbashir:
Shouldn't this be a runtime of its own rather than adding it inside `tauri-runtime-wry` ? I mean that's the whole point of the runtime layer .

wusyong:
This isn't a web runtime really. It still uses tao as backend. The purpose of this method is same as `create_tao_window` for the experiment (but for OpenGL).

amrbashir:
> This isn't a web runtime really.

Yeah, I am saying it should have its own runtime -> `tauri-runtime-egui` and depending on the feature flag, tauri will choose the runtime

wusyong:
This is for creating additional windows, not entire application. With `tauri-runtime-egui`, you won't be able to create an egui and a tauri window at the same time.
I do hope we can add this just as a plugin but tauri doesn't expose the same event from tao. And doing so won't be a trivial task.
We are lacking ways to interact with OpenGL so this is the minimum approach to try.

amrbashir:
I see, I guess we can revisit this in tauri v2, once we refactor `tauri::Builder::run()` and expose all events

Issue #3050, opened 2021-12-09
Error while installing tauri cli via cargo


mtvbrianking:
### **Describe the bug**

I faced with the following error(s) when installing `tauri/cli` via Cargo.

- unresolved import `clap::crate_version`
- no `crate_version` in the root
- expected struct `App`, found `&yaml_rust::yaml::Yaml`

### **To Reproduce**

Steps to reproduce the behaviour:

```
cargo install tauri-cli --version ^1.0.0-beta
```

### **Expected behavior**

It should install successfully...

### **Screenshots**

![image_2021-12-09_12-01-05](https://user-images.githubusercontent.com/5412360/145367831-1204cf41-0027-494e-b3fb-45bce48da60a.png)

### **Platform and Versions (required):**

```
Operating System: Ubuntu 20.04

Node.js environment
Node.js: v14.18.2
@tauri-apps/cli: ___________
@tauri-apps/api: ___________

Global packages
npm: 6.14.15 
yarn: 1.22.17

Rust environment
rustc: 1.59.0-nightly (e6b883c74 2021-12-08)
cargo: 1.58.0-nightly (40dc28175 2021-12-06)
```

### **Additional context**

I'm trying to build [Graviton App](https://github.com/Graviton-Code-Editor/Graviton-App)

### **Stack Trace**

```
bmatovu@home-pc:graviton$ cargo install tauri-cli --version ^1.0.0-beta
    Updating crates.io index
  Installing tauri-cli v1.0.0-beta.7
  .
  .
  .
   Compiling tauri-cli v1.0.0-beta.7
error[E0432]: unresolved import `clap::crate_version`
 --> /home/bmatovu/.cargo/registry/src/github.com-1ecc6299db9ec823/tauri-cli-1.0.0-beta.7/src/main.rs:6:12
  |
6 | use clap::{crate_version, load_yaml, App, AppSettings, ArgMatches};
  |            ^^^^^^^^^^^^^ no `crate_version` in the root

error: cannot determine resolution for the macro `crate_version`
   --> /home/bmatovu/.cargo/registry/src/github.com-1ecc6299db9ec823/tauri-cli-1.0.0-beta.7/src/main.rs:265:14
    |
265 |     .version(crate_version!())
    |              ^^^^^^^^^^^^^
    |
    = note: import resolution is stuck, try simplifying macro imports

error[E0308]: mismatched types
[...]

PR #3186, opened 2022-01-09
Update docs for proper installation of cli

tayloredwebsites:
got 3 errors installing cli
according to https://github.com/tauri-apps/tauri/issues/3050
the install line must be changed to pull from the next branch from github

Note: I did the cargo subcommand option when I had issues with yarn.  I got yarn running, but wanted to get docs updated for others. 
I hope this makes sense.  I am just getting into tauri today.

- [ ] Bugfix
- [ ] Feature
- [x] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist
- [ ] N/A - When resolving issues, they are referenced in the PR's title (e.g `fix #___, #___`)
- [ ] N/A - A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md).
- [ ] N/A - I have added a convincing reason for adding this feature, if necessary

### DIFF
--- a/docs/development/integration.md
+++ b/docs/development/integration.md
@@ -65,7 +65,7 @@ If you decide to use Tauri as a local package with npm (not yarn), you will have
 This will install `tauri-cli` as a Cargo subcommand on the cargo binary folder (by default on `$HOME/.cargo/bin`):
 
 ```bash
-cargo install tauri-cli --version ^1.0.0-beta
+cargo install tauri-cli --git https://github.com/tauri-apps/tauri --branch next
 ```
 
 For more installation options, see [`cargo install`](https://doc.rust-lang.org/cargo/commands/cargo-install.html#description)


FabianLars:
_ah damn, i didn't save my comment..._

Thanks for being considering contributing!
Unfortunately, this won't get merged for 2 reasons:
1) The next release is not that far off anymore, so it might just not be worth it, especially if we take into account that the rust cli is pretty unpopular, which is kinda sad if you ask me, but it is what it is.
2) The next branch of this repo will be copied into the docs repo only on new releases, so the doc change would be released together with the fix.

In theory we could do that change in the docs repo directly, but imo that's not worth because of 1)

You already turned this into a draft, but i'm gonna close it anyway. Thanks again.

PR #3249, opened 2022-01-19
tauri-studio JS doc generation 

lorenzolewis:

- [ ] Bugfix
- [ ] Feature
- [x] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: Adds yarn script for `tooling/api` to generate documentation

- [x] No

### Checklist

### Other information
This is to fix and simplify the JavaScript documentation build pipeline required due to changes https://github.com/tauri-apps/tauri-docs/pull/337 (i18n updates)

#### Changes:

Generated website can be viewed here: https://deploy-preview-346--tauri.netlify.app/

- Add `generate-docs` to yarn script in tooling/api
- Add "docs" to `.gitignore` in tooling/api so that derived documentation isn't committed
- Modify "update-docs" GitHub workflow:
  - No longer build Rust documentation (website is using docs.rs for now)
  - No longer add dependencies required to build rust documentation
  - Use `yarn generate-docs` to build js documentation instead of https://github.com/tauri-apps/typedocusaurus
  - Remove `sidebar.js` move command (sidebar now dynamically generated by website)
  - Update JS API documentation copy path


lucasfernog:
`Remove architecture.md move command (now hosted in tauri-studio repo)` can't we keep it as a copy from this repo? Otherwise it can be boring having to update both.

.github/workflows/update-docs.yml
```diff
@@ -40,43 +40,33 @@ jobs:
...
-          targetPath: ${{ github.workspace }}/tauri-docs/docs/en/api/js
-          docusaurusPath: ./tauri-docs/
+        run: yarn && yarn generate-docs
 
-      # Moving docs for Indexation
-      - name: copy docs
-        working-directory: ./tauri
-        run: |
-          mv docs/sidebar.json ${{ github.workspace }}/tauri-docs/sidebars/core.json
-          cp -r docs/!(.templates) ${{ github.workspace }}/tauri-docs/docs/en
```
lucasfernog, line 78: I think you need to keep this script, otherwise the `docs` folder won't be copied to `tauri-docs`.

.github/workflows/update-docs.yml
```diff
@@ -40,43 +40,33 @@ jobs:
...
-          mv ARCHITECTURE.md ${{ github.workspace }}/tauri-docs/docs/en/about/architecture.md
```
lucasfernog, line 79: I'd love to keep this too

lorenzolewis:
> `Remove architecture.md move command (now hosted in tauri-studio repo)` can't we keep it as a copy from this repo? Otherwise it can be boring having to update both.

We can, I'd like to have a discussion with the team on keeping all of these types of things (guides, getting started, etc.) in the Tauri-docs repo instead of here. @ksnyde and I had this discussion and right now from a new-joiner's perspective it's confusing to have them in both places (even if one is copied over as part of build). If not convinced then we can keep it for now and circle back with a larger discussion overall.

.github/workflows/update-docs.yml
lorenzolewis: This is what [line 69](https://github.com/tauri-apps/tauri/pull/3249/files#diff-6cb924cb587613a1d83384b44ca0c806ccf512202b3f29f529fc6825cef21de2R69) is doing, just collapsed it down to one line

.github/workflows/update-docs.yml
lorenzolewis: https://github.com/tauri-apps/tauri/pull/3249#issuecomment-1016522707 Commented on this here, we can do whatever makes the most sense for this specific piece, though

lorenzolewis:
Also, wasn't 100% sure that this was the right branch to put this one (given we'd want this action attached to whatever branch 1.0 is going to be deployed from)

FabianLars:
For guides and stuff this discussion makes sense, but not for the architecture.md file, which is more important in tauri than in it is in tauri-docs imo (or at least of the same importance).
I don't really care about the guides and stuff tho, i'm fine with both approaches, although with dev vs next the current approach kinda makes sense i guess (the docs here are not quite the same as in the tauri-docs repo content wise).

.github/workflows/update-docs.yml
lucasfernog: oops :D

lucasfernog:
The reason for having the docs in this repo is to simplify the process of updating the documentation when a new feature/fix PR is merged. I agree that it simplifies the lifes of the core team but complicates things for new contributors. Maybe we could extend the tauri-bot to allow creating a clone PR from tauri-docs to tauri? Currently we do this for issues, to create a copy of a tauri issue to tao or wry. Repo: https://github.com/tauri-apps/tauri-apps-bot

.github/workflows/update-docs.yml
FabianLars: well, it's not exactly the same right? the old version copies all docs including guides and line 69 is only for api docs if i got that right.

lorenzolewis: You're right @FabianLars , that's what I get for not being as familiar with this repo (yet)

lucasfernog (approved):

.github/workflows/update-docs.yml
```diff
@@ -40,43 +40,39 @@ jobs:
...
+      - name: Copy JS docs
+        run: cp -r tauri/tooling/api/docs/* tauri-docs/docs/api/js/
+        
+      - name: Copy Architecture doc
+        run: cp -r tauri/ARCHITECTURE.md tauri-docs/docs/
+
+      - name: Copy Guides docs
+        run: cp -r docs/!(.templates) tauri-docs/docs/en
```
FabianLars: that should be docs instead of docs/en, right? not sure with all those recent changes haha

lorenzolewis:
Switching this over to a draft until we can have a talk as a group on overall repo divisions

lorenzolewis:
Alright, this one should be good to go now

FabianLars:
Looking at <https://github.com/tauri-apps/tauri-docs/pull/346/files>
Are the changes to `docs/sidebar.json` and `version.txt` correct? (the files are all the way at the bottom in that link above)

Also, that's how the sidebar looks in that PR:
Screenshot 2022-01-20 132852

Besides that LGTM

If we merge this then we need the same changes here btw: <https://github.com/tauri-apps/tauri/blob/next/.github/workflows/covector-version-or-publish.yml#L76>. If you want we can do that in a seperate PR after we did some deployed test runs with that one here.

lorenzolewis:
I'm not going to merge that into the docs site since it's the docs from `next`. I'm going to run a local copy of this against the beta 8 tag.

`sidebar.json` is a bit more tricky (along with the other guides docs) because that goes into the conversation of where those files belong (tauri or tauri-docs repo), so my view is keep it as-is for this PR.

Sidebar will render like this (have a PR I'm working on for tauri-docs that will include the beta 8 updates as well as this sidebar update):
Screenshot 2022-01-20 at 12 41 52

For convector, let's look at that in a separate PR. Having a quick glance, there's a lot of duplicate logic there so think we can just call out to the `generate-docs` action maybe instead of having the duplication.

FabianLars:
Alrighty 👍 

Edit:
> I'm not going to merge that into the docs site since it's the docs from next. I'm going to run a local copy of this against the beta 8 tag.

Yep i remembered that :P And i think you said this because of my remark about version.txt? If so, then that's now what i wanted to hint at the change in that file is `v1.0.0-beta.8` -> `1.0.0-beta.8`, just wanted to know if we need that **v**.
Also, something similar for sidebar.json because i thought that's not supposed to be copied over (autogenerated)? Just wanted you to double check that 👍

FabianLars (approved):

lorenzolewis:
@FabianLars for the versions, I think that's just because when you run the `generate-docs` workflow manually it asks for a version number https://github.com/tauri-apps/tauri/blob/1eec7282c76e0a2bef030a764efece8d0c145ce3/.github/workflows/update-docs.yml#L21

On release (which assuming that's how the previous run was done that was pushed), it's pulled from the event release tag name
https://github.com/tauri-apps/tauri/blob/1eec7282c76e0a2bef030a764efece8d0c145ce3/.github/workflows/update-docs.yml#L94

For the sidebar copy, that goes under the topic of which repo owns the guides (I made some changes in tauri-docs on the sidebar without realising that it's actually generated here). So think we'll want to circle back on that

lorenzolewis:
Confirmed that version.txt is just because I didn't type in the v on the manual workflow ;)

lorenzolewis:
@FabianLars @lucasfernog  this one is now ready to merge, ran a successful PR against tauri-docs https://github.com/tauri-apps/tauri-docs/pull/351

FabianLars (approved):

PR #3263, opened 2022-01-22
Docs removal

lorenzolewis:

- [ ] Bugfix
- [ ] Feature
- [x] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

- Covector templates are moved into .scripts/covector/templates
- Markdown header image is moved to .github/splash.png
- All guide type docs have been removed (now maintained in tauri-docs)


.github/workflows/update-docs.yml
```diff
@@ -68,12 +68,6 @@ jobs:
       - name: Copy JS docs
         run: cp -r tauri/tooling/api/docs/* tauri-docs/docs/api/js/
         
-      - name: Copy Architecture doc
-        run: cp -r tauri/ARCHITECTURE.md tauri-docs/docs/about/
```
lucasfernog, line 72: shouldn't we keep this?

lorenzolewis: No, we're going to do a link to this page, and the version on the site is going to be adapted to be more high-level. So for now I've copied it over so we have a starting point, but the goal of this and the one on the site will be different.

lucasfernog: ok

PR #3274, opened 2022-01-24
fix(cli.rs): prefix the "before script" env vars with `TAURI_`

JonasKruckenberg:
Before we exported env vars to the "before scripts" (`beforeDevCommand` and `beforeBuildCommand`) like this `ARCH`, `PLATFORM` and `VERSION` which is problematic for two reasons:
1. It can clash with env variables defined by the OS (`ARCH` is very generic for example)
2. It wasn't clear that the env vars above came from the tauri bundling process.
3. Tools like vite, have a whitelist by prefix for security reasons. So having everything prefixed with `TAURI_` will make life easier and more secure for our tauri+vite users.

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes, and the changes were approved in issue #___
- [ ] No

### Checklist

### DIFF
--- a/.changes/before-script-envs.md
+++ b/.changes/before-script-envs.md
@@ -2,4 +2,4 @@
 "cli.rs": patch
 ---
 
-Define `PLATFORM`, `ARCH`, `FAMILY`, `PLATFORM_TYPE` and `TAURI_DEBUG` environment variables for the `beforeDevCommand` and `beforeBuildCommand` scripts.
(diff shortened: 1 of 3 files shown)


FabianLars (approved):
Good idea imo

JonasKruckenberg:
> Good idea imo

I caused the problem, I fix the problem lol 😂

lucasfernog (approved):

PR #3286, opened 2022-01-26
fix(bundler): Use `arch` instead of `llvm_target`. fix #3285

### LINKED ISSUE #3285: [bug] Incorrect llvm target
### Describe the bug

https://github.com/tauri-apps/tauri/blob/9bb68973dd10f3cb98d2a95e5432bfc765d77064/tooling/bundler/src/bundle/platform.rs#L27-L30

This line uses the rust compiler to identify the current target arch, which is problematic for two reasons:
1. It uses nightly rust through an undocumented semi-private environment variable *even* if the users chose a stable rust toolchain
2. It identifies the arch incorrectly. It identifies my m1 macbook pro as `arm64-apple-macosx11.0.0` (it even identifies my colleagues macbook as `x86_64-apple-macosx10.12.0`)

### Reproduction

1. Install or update rust
2. run `RUSTC_BOOTSTRAP=1 rustc -Z unstable-options --print target-spec-json`

### Expected behavior

It should correctly identify the current target as `aarch64` (the `arch` property of the same json object even correctly states `aarch64`).

### Platform and versions

```shell
Operating System - Mac OS, version 12.2.0 X64

Node.js environment
  Node.js - 16.5.0
  @tauri-apps/api - Not installed

Global packages
  npm - 8.3.1
  pnpm - 6.27.2
  yarn - 1.22.17

Rust environment
  rustc - 1.58.1
  cargo - 1.58.0

Rust environment
  rustup - 1.24.3
  rustc - 1.58.1
  cargo - 1.58.0
  toolchain - stable-aarch64-apple-darwin (environment override by RUSTUP_TOOLCHAIN)

App directory structure
/node_modules
/scripts
/src-tauri
/src

App
  tauri.rs - path:"../../../core/tauri" [1.0.0-beta.8] (no lockfile)
  build-type - bundle
  CSP - default-src blob: data: filesystem: ws: wss: http: https: tauri: 'unsafe-eval' 'unsafe-inline' 'self'
  distDir - ["../index.html"]
  devPath - ["../index.html"]
```

### Stack trace

_No response_

### Additional context

We should use `rustc --print cfg` instead of `RUSTC_BOOTSTRAP=1 rustc -Z unstable-options --print target-spec-json` or even remove the whole command and just use the `cfg!` macros instead

JonasKruckenberg:
The llvm target misidentified the arch on apple silicon, causing sidecar bundling to fail.
This fixes #3285.

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


lucasfernog:
I see you didn't change the command to `rustc --print cfg` as stated on the issue. Is that intentional?

JonasKruckenberg:
> I see you didn't change the command to `rustc --print cfg` as stated on the issue. Is that intentional?

nope, just didn't get to it yet bc the parsing is a bit more involved

PR #3305, opened 2022-01-29
fix(api.js) Replace `number[]`with `Uint8Array`. fixes #3306

### LINKED ISSUE #3306: Replace `number[]` with `Uint8Array`
Our api functions that deal with binary data currently have the type `number[]` which serializes to the correct JSON, but has sub-par DX when used with other JavaScript tools and libraries. 
The attached PR replaces all occurrences of `number[]` with `Uint8Array` the correct `Vec<u8>` equivalent.

This is a breaking change though, as functions no longer accept plain number arrays. Typescript users will get correct errors, but non-ts users will receive `invalid type: map, expected a sequence`

This change will only affect the user facing interface, as Uint8Arrays will be converted into `number[]` under the hood to get  preserve the correct serialization

JonasKruckenberg:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

attacked issue #3306

- [ ] No

### Checklist


FabianLars:
if we merge this we need a changefile too :P
ngl, i'm glad i'm not the only one who always forgets them 😂

JonasKruckenberg:
> ngl, i'm glad i'm not the only one who always forgets them 😂

Throwback to the time I merged a PR **and** deleted the branch without a changefile and had to open a super embarrassing PR with only a changefile 😂 

Edit:
> if we merge this we need a changefile too :P

After your comment on the issue, I feel like making this a patch thing is good enough 🤔 

Edit2: Not that it matters much, this will result in a major release anyway lol

FabianLars:
yeah we did `patch` all the time, for pretty much every change. Really doesn't seem to matter for pre-releases, or at least from 0.x to 1.0-pre, whatever.

tooling/api/src/http.ts
```diff
@@ -37,7 +37,7 @@ enum ResponseType {
   Binary = 3
 }
 
-type Part = string | number[]
+type Part = string | Uint8Array
```
FabianLars, line 40: I think this causes issues at the form() function on line 60 (can't add the comment there)

FabianLars (changes requested):

tooling/api/src/http.ts
JonasKruckenberg: Yikes, yeah good catch. I think I'll add an internal `serializePayload` function or something.

tooling/api/src/fs.ts
```diff
@@ -119,15 +119,17 @@ async function readTextFile(
 async function readBinaryFile(
   filePath: string,
   options: FsOptions = {}
-): Promise<number[]> {
-  return invokeTauriCommand<number[]>({
+): Promise<Uint8Array> {
+  const arr = invokeTauriCommand<number[]>({
```
amrbashir:
```suggestion
  const arr = await invokeTauriCommand<number[]>({
```

PR #3318, opened 2022-02-01
feat: Add `universal-darwin-macos` build target, closes #3317

### LINKED ISSUE #3317: [feat] Build universal binaries for MacOS
### Describe the problem

By default, binaries are built for a single CPU architecture, either x86_64 or aarch64. MacOS supports [Universal Binaries](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary). that allow a single binary (and hence package & dmg) which contains both. This is recommended by Apple because it simplifies application distribution.

I'd like to be able to build a universal dmg that can be distributed to any MacOS user, directly from `tauri build`.

I suggest adding a pseudo-target `universal-darwin-macos` for this purpose, which invokes the [`lipo`](https://ss64.com/osx/lipo.html) command to combine two arch-specific binaries into a universal one. I've tried it out and it works for me.

Given such a feature, we also need a way to deal with external binaries:

- Either Tauri requires a universal binary be present at build-time
- Or we find the arch-specific binaries and combine them at bundle-time (see below).

#### Use an external build script

A custom build script would work, but it would require quite a bit of redundant and custom tooling. This is because the `lipo` invocation needs to happen _between_ `cargo build` and dmg packaging.

#### Use cargo [blocked]

Cargo [does not support](https://github.com/rust-lang/cargo/issues/8875) universal binaries at this time.

#### Use cargo-lipo

[`cargo-lipo`](https://github.com/TimNN/cargo-lipo) is a tool that uses `lipo` under the hood. Not investigated.

#### Use a Rust-lib instead of lipo

`lipo` is included in MacOS (perhaps with XCode?) but it may be harder to access it from Windows or Linux for cross-platform building. If that's hard to get lipo working, it may be worth looking into using a rust lib that does the same thing, to avoid external dependencies.

### Additional context

It's an open question what exactly universal binaries mean for build systems. Should it be part of the arch triple (`universal-darwin-macos`) or should it be considered separate platform-specific config? I haven't seen best practices on this yet, feedback welcome.

betamos:
See #3317.

This is my first PR so I may have missed something (despite the checklist).

Note that external binaries are not addressed yet, see issue for discussion. Also, docs are missing - I'll add those once the approach is approved.

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


lucasfernog:
Nice work, I'll fix the merge conflict and test it now.

lucasfernog (approved):
You only missed the change file and would be good if you had your commits signed too, but this was an amazing PR! Thank you ❤️

betamos:
Thanks! Note that sidecar still doesn't work with universal but I'll send out another one for that.

PR #3321, opened 2022-02-02
Expose event interface. fixes #2733

### LINKED ISSUE #2733: The event module is private despite being exposed in the public API
### **Describe the bug**
The event module is private, making it impossible to reference `Event` or `EventHandler`.

### **To Reproduce**
1. Try to import `tauri::event::{Event, EventHandler}`

### **Expected behavior**
The module is not private.

### **Platform and Versions (required):**

```
Operating System - Windows, version 10.0.19042 X64
Webview2 - 94.0.992.38

Node.js environment
  Node.js - 16.11.0

Global packages
  npm - 8.0.0
  yarn - 1.22.15

Rust environment
  rustc - 1.57.0-nightly
  cargo - 1.57.0-nightly

App directory structure
/.git
/.idea
/node_modules
/public
/src
/src-tauri

App
  tauri.rs - 1.0.0-beta.8
  build-type - bundle
  distDir - Set automatically by Vue CLI plugin
  devPath - Set automatically by Vue CLI plugin
  framework - Vue.js (Vue CLI)
  bundler - Webpack
```

### **Additional context**
I would like to do this:
```rs
fn some_func(&self) {
    self.handle.listen_global("send-packet", listener);
}

fn listener(event: Event) {
    // ...
}
```

Unfortunately, there is no way for me to `use` the necessary type, so this is impossible.

In addition, it is impossible to store the result of `listen_global` to be `unlisten`ed later, since there's no way to reference `EventHandler` in a struct.

I wonder if it's a Rust bug that the API is even allowed to do this. It's essentially exposing a private type in the public API, which you can't do normally.

Chaostheorie:
Expose `tauri::event` and rename `tauri::event::Event` to `EmittedEvent`. Temporary fix for #2733. This is supposed to be a POC for my internal use until #2733 is resolved in v1.x.

- [x] Bugfix
- [x] Refactor

- [x] Yes, and the changes aren't approved 
- [ ] No

> Note: Tests were passing on GNU/Linux 5.15.18-1-lts with rustc 1.58.0

### DIFF
--- a/.changes/api-change-events.md
+++ b/.changes/api-change-events.md
@@ -1,0 +1,6 @@
+---
+"tauri": patch
+---
+
+- `tauri::event::Event` and `tauri::event::EventHandler` are now exported as `tauri::Event` and `tauri::EventHandler`
+- **breaking**: The old `tauri::app::Event` has been renamed to `tauri::app::RunEvent` and is available as `tauri::RunEvent`
--- a/.changes/api-change-events.md
+++ b/.changes/api-change-events.md
@@ -2,5 +2,5 @@
 "tauri": patch
 ---
 
-- `tauri::event::Event` and `tauri::event::EventHandler` are now exported as `tauri::Event` and `tauri::EventHandler`
-- **breaking**: The old `tauri::app::Event` has been renamed to `tauri::app::RunEvent` and is available as `tauri::RunEvent`
+* **Breaking change**: Renamed `tauri::Event`  to `tauri::RunEvent`
+* Exported `tauri::Event` and `tauri::EventHandler` so you can define a function and pass it to `Window::listen`
--- a/core/tauri/src/app.rs
+++ b/core/tauri/src/app.rs
@@ -16,7 +16,7 @@ use crate::{
     http::{Request as HttpRequest, Response as HttpResponse},
     webview::{WebviewAttributes, WindowBuilder},
     window::{PendingWindow, WindowEvent},
-    Dispatch, ExitRequestedEventAction, RunEvent, Runtime,
+    Dispatch, ExitRequestedEventAction, RunEvent as RunTimeEvent, Runtime,
   },
   sealed::{ManagerBase, RuntimeOrDispatch},
   utils::assets::Assets,
@@ -75,7 +75,7 @@ impl CloseRequestApi {
 /// An application event, triggered from the event loop.
 #[derive(Debug)]
 #[non_exhaustive]
-pub enum Event {
+pub enum RunEvent {
   /// Event loop is exiting.
   Exit,
   /// The app is about to exit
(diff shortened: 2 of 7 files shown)


core/tauri/src/lib.rs
```diff
@@ -35,7 +35,7 @@ pub mod command;
 /// The Tauri API endpoints.
 mod endpoints;
 mod error;
-mod event;
+pub mod event;
```
amrbashir: don't export this whole mod. just do `pub use` for the types you need exported only like how it is done a few lines below.

core/tauri/src/event.rs
```diff
@@ -23,12 +23,12 @@ impl fmt::Display for EventHandler {
 
 /// An event that was triggered.
 #[derive(Debug, Clone)]
-pub struct Event {
+pub struct EmittedEvent {
```
amrbashir: This should stay `Event`

amrbashir (changes requested):
Thanks for taking the time to fix this. 

You need to:
1. rename `EmittedEvent` to `Event` and remove any mention of `EmittedEvent`
2. rename `Event` to `RunEvent`, obviously this will clash with the `runtime::RunEvent` import in `core/tauri/src/app.rs` so you're just gonna need to rename the import and not the original struct, something like this `use runtime::RunEvent as RuntimeRunEvent`

core/tauri/src/lib.rs
Chaostheorie: It seems that only Event and Eventhandler are needed directly.

core/tauri/src/app.rs
```diff
@@ -16,7 +16,7 @@ use crate::{
     http::{Request as HttpRequest, Response as HttpResponse},
     webview::{WebviewAttributes, WindowBuilder},
     window::{PendingWindow, WindowEvent},
-    Dispatch, ExitRequestedEventAction, RunEvent, Runtime,
+    Dispatch, ExitRequestedEventAction, RunEvent as RunTimeEvent, Runtime,
```
amrbashir:
```suggestion
    Dispatch, ExitRequestedEventAction, RunEvent as RuntimeRunEvent, Runtime,
```

core/tauri/src/lib.rs
```diff
@@ -84,12 +81,14 @@ pub use {
 };
 pub use {
   self::app::WindowMenuEvent,
+  self::event::{Event, EventHandler},
```
amrbashir, line 84: I think only `Event` needs to be exported

amrbashir (changes requested):
You also need to add a changefil in `.changes` folder and don't forget to mention the breaking change `Event` -> `RunEvent`

core/tauri/src/lib.rs
FabianLars: No we need EventHandler so the user can unlisten to events.
Edit: To clarify: "so the user can unlisten to events **later**", currently they can only immediately after listening (cause they can't store the EventHandler anywhere)

amrbashir: Ah okay

Chaostheorie:
What would be the appropriate level for this change? Semver would dictate to use major since the API is changed, but semver doesn't seem to be enforced for `next`.

FabianLars:
It's not enforced because of the prelease stuff, so patch is fine, as this needs to land in v1 anyway (mainly because of the EventHandler fix)

Chaostheorie:
There seems to be a change request pending, but I can't figure out, what it is referring to. 

@amrbashir your name appears in the UI, have I overlooked something?

FabianLars:
@Chaostheorie Don't worry, that's just the UI not updating until he approves the PR. As far as i can see you got everything he requested so far 👍

.changes/api-change-events.md
```diff
@@ -0,0 +1,6 @@
+---
+"tauri": patch
+---
+
+- `tauri::event::Event` and `tauri::event::EventHandler` are now exported as `tauri::Event` and `tauri::EventHandler`
+- **breaking**: The old `tauri::app::Event` has been renamed to `tauri::app::RunEvent` and is available as `tauri::RunEvent`
```
amrbashir:
```suggestion
* **Breaking change**: Renamed `tauri::Event`  to `tauri::RunEvent`
* Exported `tauri::Event` and `tauri::EventHandler` so you can define a function and pass it to `Window::listen`
```

amrbashir (approved):

FabianLars (approved):

Chaostheorie:
@amrbashir and @FabianLars thank you for the great help and explanations.

PR #3344, opened 2022-02-06
fix(core): window-specific event delivery, closes #3302

### LINKED ISSUE #3302: [bug] Adding a listener to a child window fires events from the parent window
### Describe the bug

If I create a new window from a WebView and attach a listener to it, the events that the listener receives are actually from the parent window.
e.g. if I listen to `tauri://move` on the child window, the listener fires when the parent window moves.

### Reproduction

1. Run this code in the WebView
``` javascript
  function createWindow() {
    const label = Math.random().toString();
    const webview = new WebviewWindow(label);
    windowMap[label] = webview;
    webview.once('tauri://error', function () {
      onMessage("Error creating new webview")
    });
    webview.listen('tauri://move', event => console.log(event));
  }
```
2. Move the child window
3. Nothing appears in the console logs
4. Move the parent window
5. Move events are logged in the console.

### Expected behavior

I would expect to see the move events of the child window being logged to the console.  Not those of the parent.

### Platform and versions

```shell
tauri-apps/tauri@next 2b9e6ccc38836b27a3850faa697e845325a3866d

Operating System - Mac OS, version 11.6.2 X64

Node.js environment
  Node.js - 14.18.2

Global packages
  npm - 6.14.15
  pnpm - Not installed
  yarn - 1.22.17

Rust environment
  rustc - 1.57.0
  cargo - 1.57.0

Rust environment
  rustup - 1.24.3
  rustc - 1.57.0
  cargo - 1.57.0
  toolchain - stable-x86_64-apple-darwin (default)

App directory structure
/dist
/node_modules
/public
/src-tauri
/src

App
  tauri.rs - 1.0.0-beta.8
  build-type - bundle
  CSP - default-src blob: data: filesystem: ws: wss: http: https: tauri: asset: customprotocol: 'unsafe-eval' 'unsafe-inline' 'self' img-src: 'self'
  distDir - ../dist
  devPath - http://localhost:5000/
  framework - Svelte
```

### Stack trace

_No response_

### Additional context

_No response_

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


core/tauri-runtime/src/window.rs
```diff
@@ -105,14 +105,20 @@ pub struct PendingWindow<R: Runtime> {
...
   pub menu_ids: Arc<Mutex<HashMap<MenuHash, MenuId>>>,
 
-  /// A HashMap mapping JS event names with listener ids associated.
-  pub js_event_listeners: Arc<Mutex<HashMap<String, HashSet<u64>>>>,
+  /// A HashMap mapping JS event names with associated listener ids.
+  pub js_event_listeners: Arc<Mutex<HashMap<JsEventListenerKey, HashSet<u64>>>>,
 }
 
-fn validate_label(label: &str) {
+pub fn is_label_valid(label: &str) -> bool {
```
amrbashir, line 112: nit: It should be `is_valid_label()` to match `assert_valid_label()`

lucasfernog:
Any help testing this would be very welcome. I did some tests using the API example and the multiwindow example and it looks correct, but there might be other use cases that I didn't cover. I'll merge this so it can get more visibility.
Note that you need the `next` branch `@tauri-apps/api` package, so you'd need to clone the repo, `cd tooling/api && yarn && yarn build && yarn link` and yarn link it on your project.

FabianLars:
I did some tests earlier and it seemed to work well in regard to the linked issue.
I also tried to test for unwanted side effects and didn't find any, but that's something you can't really be a 100% sure about, so take that with a grain of salt.

Edit: So yes, more people doing their own testing is indeed a good idea 🙏

lucasfernog:
Thanks Fabian. Good thing we still have time to make tests and get feedback from the community (specially until the stable release) to get this working perfectly :P

PR #3356, opened 2022-02-08
Remove target triple from sidecar bin paths, closes #3355

### LINKED ISSUE #3355: [bug] Sidecar binary fails to be located if target is `universal-apple-darwin`
### Describe the bug

The recently added `--target universal-apple-darwin` pseudo-target produces a binary which contains the target triple _in its file name_. This build info isn't accessible for an application at runtime, resulting in a failure to locate the binary at runtime.

### Reproduction

1. Try running or building the sidecar example with `--target universal-apple-darwin`.

#### 1. Omit target triple from bundled binary file names

Simply remove the target triple while copying the binary. This means that the user-facing contract remains consistent, so users would simply create a universal binary prior to build and name it `my-sidecar-universal-apple-darwin`. At dev/build time, it would be renamed to `my-sidecar`.

#### 2. Include the target triple with the main executable

We could keep naming the same, and instead include the target triple as a constant - instead of inferring it at runtime.

#### My thoughts

I am in favor of 1, because it's consistent with the main executable and simplifies runtime logic.

However, note that there is one restriction: using the universal target requires the user to create a universal binary also for sidecar, prior to the build. It is **not** possible to mix and match, and it would be hard to do so in the future without exposing more complicated knobs/flags etc to the user. I think this is ok, simple, and expected.

However, we need to instruct users on how to build universal sidecar binaries themselves, which is very simple on MacOS (and in the future, it's likely `lipo` or equivalent will be available across platforms).

### Platform and versions

```shell
MacOS
```

### Stack trace

_No response_

### Caveat: possibly breaking change

The proposed change would affect both the CLI and the runtime, so users would have to upgrade both in lockstep. This could constitute a breaking change if users upgrade these separately. Please let me know if we need considerations for that case.

### PR coming

I already fixed this in my fork and will send out a PR with the changes.

betamos:
Consistent with the main executable, bundled sidecar binaries
now get the original name - `my-sidecar[.exe]`.

The previous target triple resulted in an path that could not be
reconstructed at runtime if the target
was a universal binary (MacOS)

Clarified documentation on external binaries.

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/.changes/sidecar-runtime-rename.md
+++ b/.changes/sidecar-runtime-rename.md
@@ -1,0 +1,6 @@
+---
+"tauri": patch
+"tauri-bundler": patch
+---
+
+**Breaking change**: The sidecar's target triple suffix is now removed at build time.
--- a/.changes/universal-apple-target-sidecar.md
+++ b/.changes/universal-apple-target-sidecar.md
@@ -1,0 +1,9 @@
+---
+"tauri-bundler": patch
+"api": patch
+---
+
+When building Universal macOS Binaries through the virtual target `universal-apple-darwin`:
+
+- Expect a universal binary to be created by the user
+- Ensure that binary is bundled and accessed correctly at runtime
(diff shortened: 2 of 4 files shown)


lucasfernog (approved):
Nice work!

lucasfernog:
It's not actually a breaking change, but I pushed a changelog entry for it just in case someone is not using our API to resolve the sidecar path.

betamos:
Awesome, thanks!

Issue #2813, opened 2021-10-27
More devserver options


rondonjon:
Please add the following development config options:

- option to pause the Tauri build (or at least its console output) while the dev server is spinning up; both are writing to the console while building, and I am often missing the dev server's error in the Tauri log.

- option to terminate the Tauri build if a devserver command was configured and if that command fails with a non-zero exit code; current behavior is that Tauri keeps building and displays a blank page

- option to force-terminate the devserver when Tauri is terminated (related: #2794)

amrbashir:
1. we can't know when the devServer is finished, in fact we don't know if it is a server or not, we are only executing the `beforeDevCommamd` you specified in `tauri.conf.json`. The solution is to run your devServer in a separate terminal.

2. Good suggestion, we will definitely consider that.

3. This issue has been brought up alot but there is not much that we can do tbh, we are only spawning a process for `beforeDevCommand` and that command can other processes that we can't control nor can we even know about them and it is the job of the process who spawned them to clean them as it exists. We will definitely look into improving it but no guarantees.

amrbashir:
I am gonna go ahead and close this but feel free to open another issue for No.2

PR #3358, opened 2022-02-08
feat(cli.rs): wait for dev URL to be reachable, exit if command fails

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### Other information

Related to #2813

### DIFF
--- a/tooling/cli.rs/src/dev.rs
+++ b/tooling/cli.rs/src/dev.rs
@@ -189,57 +189,59 @@ pub fn command(options: Options) -> Result<()> {
   let (child_wait_tx, child_wait_rx) = channel();
   let child_wait_rx = Arc::new(Mutex::new(child_wait_rx));
 
-  if let AppUrl::Url(WindowUrl::External(dev_server_url)) = config
-    .lock()
-    .unwrap()
-    .as_ref()
-    .unwrap()
-    .build
-    .dev_path
-    .clone()
-  {
-    let host = dev_server_url
-      .host()
-      .unwrap_or_else(|| panic!("No host name in the URL"));
-    let port = dev_server_url
-      .port_or_known_default()
-      .unwrap_or_else(|| panic!("No port number in the URL"));
-    let addrs;
-    let addr;
-    let addrs = match host {
-      url::Host::Domain(domain) => {
-        use std::net::ToSocketAddrs;
-        addrs = (domain, port).to_socket_addrs()?;
-        addrs.as_slice()
-      }
-      url::Host::Ipv4(ip) => {
-        addr = (ip, port).into();
-        std::slice::from_ref(&addr)
-      }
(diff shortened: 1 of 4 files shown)


tooling/cli.rs/src/dev.rs
```diff
@@ -107,22 +107,40 @@ pub fn command(options: Options) -> Result<()> {
...
+
+      let child = SharedChild::spawn(&mut command)
+        .unwrap_or_else(|_| panic!("failed to run `{}`", before_dev));
+      let child = Arc::new(child);
+      let child_ = child.clone();
+      let logger_ = logger.clone();
+      std::thread::spawn(move || {
+        let status = child_.wait().expect("failed to wait on before dev command");
+        if !status.success() {
+          logger_.error("The before dev command terminated with a non-zero status code.");
```
amrbashir:
```suggestion
          logger_.error("The \"beforeDevCommand\" terminated with a non-zero status code.");
```

tooling/cli.rs/src/dev.rs
```diff
@@ -169,6 +187,59 @@ pub fn command(options: Options) -> Result<()> {
...
+        addr = (ip, port).into();
+        std::slice::from_ref(&addr)
+      }
+      url::Host::Ipv6(ip) => {
+        addr = (ip, port).into();
+        std::slice::from_ref(&addr)
+      }
+    };
+    let mut i = 0;
+    let sleep_interval = std::time::Duration::from_secs(5);
```
amrbashir: 5 seconds is too long, don't you think?

amrbashir (approved):

tooling/cli.rs/src/dev.rs
FabianLars: Hmm tricky. With vite you don't even get into a second iteration, because it's too fast.
With webpack taking like 30 seconds it would get quite spammy with timeouts < 5secs.
Maybe reduce the timeout, but only log the message every x tries?

lucasfernog: yeah I was mainly trying to reduce the spam, since some frameworks like Quasar can take a minute to build. i'll reduce the logging instead.

lucasfernog: pushed, let me know what you think

FabianLars (approved):

amrbashir (approved):
LGTM! I can't test the ghost process issue though (currently away) but I posted the steps to try and reproduce it in the other PR.

lucasfernog:
Yeah the ghost process is part of a separate investigation and PR.

amrbashir:
Okay, This PR is fine as is then.

PR #3381, opened 2022-02-10
feat(cli.rs): add tests to node binding, delete cli.js

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


amrbashir:
might be the time to rename `cli.rs` to just `cli` since `cli.js` is gone now?

lucasfernog:
> might be the time to rename `cli.rs` to just `cli` since `cli.js` is gone now?

@amrbashir the folder name yeah I agree, but we still need a convention to name those two so I think the current one is ok.

amrbashir:
yeah I was referring mainly to the folder but we can also refer to them as `cli` and `cli/node` or `cli/npm` as it is now a really thin wrapper. Current convention is fine too.

PR #3379, opened 2022-02-10
feat(core): expose `setup_with_config` on the plugin builder

lucasfernog:
### What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

### Does this PR introduce a breaking change?

- [ ] Yes, and the changes were approved in issue #___
- [ ] No

### Checklist
- [ ] When resolving issues, they are referenced in the PR's title (e.g `fix: remove a typo, closes #___, #___`)
- [ ] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md).
- [ ] I have added a convincing reason for adding this feature, if necessary

PR #3384, opened 2022-02-10
Implement default runtime for plugin structs

JonasKruckenberg:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

I overlooked this in the original PR, both structs should implement the default runtime.

### DIFF
--- a/.changes/fix-plugin-default-runtime.md
+++ b/.changes/fix-plugin-default-runtime.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch
+---
+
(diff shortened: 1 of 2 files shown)


lucasfernog:
I think we should keep the Runtime as required. By default it forces the plugin to be runtime agnostic, and also the default_runtime impl crashes with #3379, because the config generic has a default type.

JonasKruckenberg:
Yeah makes sense

PR #3422, opened 2022-02-12
doc: Add plugin::Builder examples and expand doc comments

JonasKruckenberg:

- [ ] Bugfix
- [ ] Feature
- [x] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/core/tauri/src/plugin.rs
+++ b/core/tauri/src/plugin.rs
@@ -92,9 +92,9 @@ type OnEvent<R> = dyn Fn(&AppHandle<R>, &RunEvent) + Send + Sync;
 /// impl Default for Builder {
 ///   fn default() -> Self {
 ///     Self {
-///       optionA: "foo".to_string(),
-///       optionB: "bar".to_string(),
-///       optionC: false
+///       option_a: "foo".to_string(),
+///       option_b: "bar".to_string(),
+///       option_c: false
 ///     }
 ///   }
 /// }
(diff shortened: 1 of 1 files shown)


core/tauri/src/plugin.rs
```diff
@@ -60,6 +60,76 @@ type OnWebviewReady<R> = dyn Fn(Window<R>) + Send + Sync;
...
+/// }
+/// ```
+///
+/// When plugins expose more complex configuration options, it can be helpful to provide a Builder instead:
+///
+/// ```rust
+/// use tauri::{plugin::{Builder as PluginBuilder, TauriPlugin}, runtime::Runtime};
+///
+/// pub struct Builder {
+///   optionA: String,
```
lucasfernog: use the rust convention here, option_a instead of optionA

core/tauri/src/plugin.rs
```diff
@@ -60,6 +60,76 @@ type OnWebviewReady<R> = dyn Fn(Window<R>) + Send + Sync;
...
+///
+///   pub fn option_c(mut self, option_c: bool) -> Self {
+///     self.optionC = option_c;
+///     self
+///   }
+///
+///   pub fn build<R: Runtime>(self) -> TauriPlugin<R> {
+///     PluginBuilder::new("example")
+///       .setup(|app_handle| {
+///         // use the options here to do stuff
```
lucasfernog, line 125: to use the options here you need to `move` the value, you should document that (maybe just print some option just to show it works)

core/tauri/src/plugin.rs
```diff
@@ -180,6 +340,28 @@ impl<R: Runtime, C: DeserializeOwned> Builder<R, C> {
...
+  /// ```rust
+  /// use tauri::{plugin::{Builder, TauriPlugin}, RunEvent, runtime::Runtime};
+  ///
+  /// fn init<R: Runtime>() -> TauriPlugin<R> {
+  ///   Builder::new("example")
+  ///     .on_event(|app_handle, event| {
+  ///       match event {
+  ///         RunEvent::ExitRequested { api, .. } => {
+  ///           // Prevents the app from exiting.
+  ///           // This will case the core thread to continue running in the background even without any open windows.
```
lucasfernog:
```suggestion
  ///           // This will cause the core thread to continue running in the background even without any open windows.
```

core/tauri/src/plugin.rs
JonasKruckenberg: Thanks clippy 😉

core/tauri/src/plugin.rs
JonasKruckenberg: Yeah good idea 👍🏻

core/tauri/src/plugin.rs
lucasfernog: you forgot this part LOL i'm really clippy

JonasKruckenberg: Ahh jeez, coding on your phone never works, I have no idea how amr does it

JonasKruckenberg: fixed!

lucasfernog (approved):

PR #3431, opened 2022-02-13
feat(cli.rs): allow passing arguments to the build runner, closes #3398

### LINKED ISSUE #3398: [feat] allow passing arguments to cargo build
### Describe the problem

I am trying to debug an issue with my project setup that causes tauri's `cargo build` invocation to fully rebuild the entire project, plus dependencies after rust-analyzer has finished checking the project source. According to comments on rust-lang/cargo#4961 I need the output of invoking `cargo build` with the flag `-vvvv` to do this, but the problem only appears when `cargo` is invoked through `tauri dev` and doesn't show up on a manual `cargo build` invocation.

I would like a way to pass arguments directly to the underlying `cargo build` calls performed by `tauri build` and `tauri dev`. This would allow me to pass the flag `-vvvv` directly and get fully verbose output from the build process.

### Alternatives considered

An alternative solution would be a separate flag under `tauri build`/`tauri dev` to directly set the verbosity level that will be requested from the build process, but if more flags are required in the future then supporting each one separately can be harder to maintain than simply allowing users to pass them directly to the toolchain.

### Additional context

I have asked on the Tauri Discord server about existing ways to get this output, and have been told that it is currently not yet possible to pass the flag I require. I have been asked to open this feature request, as it sounds like a useful feature to have. According to rust-lang/cargo#4961 directly passing the verbose flag to `cargo build` is needed to get the information, and therefore an alternative solution cannot be found.

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/.changes/refactor-cli-dev-args.md
+++ b/.changes/refactor-cli-dev-args.md
@@ -2,4 +2,4 @@
 "cli.rs": patch
 ---
 
-**Breaking change:** The extra arguments passed to `tauri dev` are now propagated to the runner. To pass arguments to your binary using Cargo, now you need to run `tauri dev -- -- --arg --to --your --binary`.
+**Breaking change:** The extra arguments passed to `tauri dev` using `-- <ARGS>...` are now propagated to the runner (defaults to cargo). To pass arguments to your binary using Cargo, you now need to run `tauri dev -- -- <ARGS-TO-YOUR-BINARY>...` (notice the double `--`).
--- a/.changes/refactor-cli-dev-args.md
+++ b/.changes/refactor-cli-dev-args.md
@@ -1,0 +1,5 @@
+---
+"cli.rs": patch
+---
+
+**Breaking change:** The extra arguments passed to `tauri dev` are now propagated to the runner. To pass arguments to your binary using Cargo, now you need to run `tauri dev -- -- --arg --to --your --binary`.
--- a/.changes/cli-runner-args.md
+++ b/.changes/cli-runner-args.md
@@ -1,0 +1,5 @@
+---
+"cli.rs": patch
+---
+
+Allow passing arguments to the `build` runner (`tauri build -- <ARGS>...`).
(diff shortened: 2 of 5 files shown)


amrbashir:
What about `tauri dev`? I think the author of the issue wanted that.

lucasfernog:
> What about `tauri dev`? I think the author of the issue wanted that.

Tauri dev is tricky because we're already using the `-- <ARGS>` approach to pass commands to the binary. So maybe we change dev to work the same way as build, which already comes with support to passing args to the binary (`tauri dev -- --frozen -- --arg-to-binary`. Ugly but it works.

amrbashir:
yeah, I hardly think anyone uses it but so we don't introduce any breaking change we could do something like this
`tauri build/dev --cargo-args="-vvv -Z unstable-options"`

the name could be better

amrbashir:
> So maybe we change dev to work the same way as build, which already comes with support to passing args to the binary (`tauri dev -- --frozen -- --arg-to-binary`. Ugly but it works.

or that

lucasfernog:
I went with a small breaking change, now is the time anyway. Keeping the syntax the same is more important IMO.

.changes/refactor-cli-dev-args.md
```diff
@@ -0,0 +1,5 @@
+---
+"cli.rs": patch
+---
+
+**Breaking change:** The extra arguments passed to `tauri dev` are now propagated to the runner. To pass arguments to your binary using Cargo, now you need to run `tauri dev -- -- --arg --to --your --binary`.
```
amrbashir: I am not sure if this simplifies it or overcomplicates it but I am afraid users might miss the double double-dash.

```suggestion
**Breaking change:** The extra arguments passed to `tauri dev` using `-- <ARGS>...` are now propagated to the runner (defaults to cargo). To pass arguments to your binary using Cargo, you now need to run `tauri dev -- -- <ARGS-TO-YOUR-BINARY>...` (notice the double `--`).
```

amrbashir (dismissed):

PR #3448, opened 2022-02-13
feat(ci): simplify cache

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: CI

- [ ] No

### Checklist

### DIFF
--- a/.github/workflows/artifacts-updater.yml
+++ b/.github/workflows/artifacts-updater.yml
@@ -16,6 +16,11 @@ on:
       - 'tooling/bundler/**'
       - 'examples/updater/**'
 
+env:
+  RUST_BACKTRACE: 1
+  CARGO_INCREMENTAL: 0 # This is set to 0 by the https://github.com/Swatinem/rust-cache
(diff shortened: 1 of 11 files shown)


FabianLars:
Honestly i feel like <https://github.com/Swatinem/rust-cache> would be a good idea for most of the caches here, especially since it does some extra stuff to make the caches more efficient.
Also we may want to consider adding this: <https://github.com/tauri-apps/tauri-docs/blob/dev/.github/workflows/update-docs.yml#L26-L27> to some of the workflows.

lucasfernog:
I'd prefer being consistent, and I see rust-cache would be better if we were only caching one target folder per job, but sometimes we're caching multiple.

lucasfernog:
This PR also simplifies a migration, so hopefully someone better on GH actions can give it a try later.

PR #3523, opened 2022-02-20
fix: read Command output ending with a carriage return, closes #3508

### LINKED ISSUE #3508: [bug] Receiver coming from Command::spawn() never gets stdout if child process uses output clearing
### Describe the bug

When using the sidecar to execute long running sub commands, I try to receive new stdout events.

For that I wrote this code:

```rust
let (mut rx, mut child) = cmd.spawn().expect("Failed to spawn sidecar");

    // read events such as stdout
    while let Some(event) = rx.recv().await {
        if let CommandEvent::Stdout(line) = event {
            debug!("{}", line);
            window.emit("upload-progress", Payload {
                message: line,
            });
        }
    }
    debug!("upload complete");
    window.emit("upload-complete", {});
});
```

But the future of `rx.recv()` is never resolved when spawning a command that clears its output before writing to it again (i.e. human readable progress in a terminal). It only resolves after the child process exited and then it dumps out all the lines at once.

Using a command that doesn't clear out before updating works fine and it _seems_ like, that it also works fine when the command clears out and writes multiple lines.

I encountered that using the `rclone sync source:path target:path --progress --progress-terminal-title --stats-one-line` command.

### Reproduction

1. Create an app
2. Using sidecar, spawn a long running command that always clears stdout before writing to it
3. Try to receive events from command
4. `rx.recv().await` doesn't resolve
5. Wait until command finsihed
6. `rx.recv().await` resolved with the complete output

### Expected behavior

`rx.recv().await` should resolve for every event during running command

### Platform and versions

```shell
Operating System - Mac OS, version 11.6.0 X64

Node.js environment
  Node.js - 16.13.0
  @tauri-apps/cli - 1.0.0-rc.4
  @tauri-apps/api - 1.0.0-rc.1

Global packages
  npm - 8.1.0
  pnpm - Not installed
  yarn - 1.22.10

Rust environment
  rustup - 1.24.3
  rustc - 1.58.1
  cargo - 1.58.0
  toolchain - stable-x86_64-apple-darwin 

App directory structure
/test
/dist
/node_modules
/src-tauri
/.git
/.idea
/src

App
  tauri - 1.0.0-rc.2
  tauri-build - 1.0.0-rc.2
  tao - 0.6.2
  wry - 0.13.1
  build-type - bundle
  distDir - ../dist
  devPath - http://local-desktop.syra.live:3000/
  framework - React
✨  Done in 9.42s.
```

### Stack trace

_No response_

### Additional context

Discussion with @FabianLars about that here in discord: https://discord.com/channels/616186924390023171/731495047962558564/944267064175132732

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### Other information

The `process::Command` API will now read lines until a `\n` or a `\r` is found.

### DIFF
--- a/core/tauri/src/api/process/command.rs
+++ b/core/tauri/src/api/process/command.rs
@@ -55,11 +55,11 @@ pub struct TerminatedPayload {
 #[serde(tag = "event", content = "payload")]
 #[non_exhaustive]
 pub enum CommandEvent {
-  /// Stderr line.
+  /// Stderr bytes until a newline (\n) or carriage return (\r) is found.
   Stderr(String),
-  /// Stdout line.
+  /// Stdout bytes until a newline (\n) or carriage return (\r) is found.
   Stdout(String),
-  /// An error happened.
+  /// An error happened waiting for the command to finish or converting the stdout/stderr bytes to an UTF-8 string.
   Error(String),
   /// Command process terminated.
   Terminated(TerminatedPayload),
@@ -261,9 +261,17 @@ impl Command {
     let guard_ = guard.clone();
     spawn(move || {
       let _lock = guard_.read().unwrap();
-      let reader = BufReader::new(stdout_reader);
-      for line in reader.lines() {
+      let mut reader = BufReader::new(stdout_reader);
+
+      let mut buf = Vec::new();
+      loop {
+        buf.clear();
+        let n = read_command_output(&mut reader, &mut buf).unwrap();
+        if n == 0 {
+          break;
+        }
         let tx_ = tx_.clone();
+        let line = String::from_utf8(buf.clone());
         block_on_task(async move {
           let _ = match line {
             Ok(line) => tx_.send(CommandEvent::Stdout(line)).await,
@@ -277,9 +285,17 @@ impl Command {
     let guard_ = guard.clone();
     spawn(move || {
       let _lock = guard_.read().unwrap();
-      let reader = BufReader::new(stderr_reader);
-      for line in reader.lines() {
+      let mut reader = BufReader::new(stderr_reader);
+
+      let mut buf = Vec::new();
+      loop {
+        buf.clear();
+        let n = read_command_output(&mut reader, &mut buf).unwrap();
+        if n == 0 {
+          break;
+        }
         let tx_ = tx_.clone();
+        let line = String::from_utf8(buf.clone());
         block_on_task(async move {
           let _ = match line {
             Ok(line) => tx_.send(CommandEvent::Stderr(line)).await,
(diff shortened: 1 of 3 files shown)


core/tauri/src/api/process/command.rs
```diff
@@ -261,9 +261,17 @@ impl Command {
...
     spawn(move || {
       let _lock = guard_.read().unwrap();
-      let reader = BufReader::new(stdout_reader);
-      for line in reader.lines() {
+      let mut reader = BufReader::new(stdout_reader);
+
+      let mut buf = Vec::new();
+      loop {
+        buf.clear();
+        let n = read_command_output(&mut reader, &mut buf).unwrap();
```
chippers: is there a reason to crash on error here instead of just using `?` to bubble it up?

core/tauri/src/api/process/command.rs
```diff
@@ -277,9 +285,17 @@ impl Command {
     let guard_ = guard.clone();
     spawn(move || {
```
chippers: we should probably refactor this copied logic into a simple function and call it once for stdout and stderr to prevent any future mistakes. from reading it it looks like the only differences are the reader and the output.

something like:
```rust
fn spawn_pipe_reader(reader: PipeReader, wrapper: impl Fn(String) -> CommandEvent) -> std::io::Result<()> {
  // -snip-
}
```

and then you should be able to do
```rust
spawn_pipe_reader(stdout_reader, CommandEvent::Stdout)?;
spawn_pipe_reader(stderr_reader, CommandEvent::Stderr)?;
```

core/tauri/src/api/process/command.rs
```diff
@@ -390,6 +406,50 @@ impl Command {
...
+    let (done, used) = {
+      let available = match r.fill_buf() {
+        Ok(n) => n,
+        Err(ref e) if e.kind() == std::io::ErrorKind::Interrupted => continue,
+        Err(e) => return Err(e),
+      };
+      match memchr::memchr(b'\n', available) {
+        Some(i) => {
+          buf.extend_from_slice(&available[..=i]);
+          (true, i + 1)
```
chippers: this suggestion is a micro optimization because exclusive ranges still seem be faster (not that inclusive isn't also fast), but we may as well do it since we calculate it anyways.
```suggestion
          let end = i + 1;
          buf.extend_from_slice(&available[..end]);
          (true, end)
```

core/tauri/src/api/process/command.rs
```diff
@@ -390,6 +406,50 @@ impl Command {
...
+        }
+        None => match memchr::memchr(b'\r', available) {
+          Some(i) => {
+            buf.extend_from_slice(&available[..=i]);
+            (true, i + 1)
```
chippers: this suggestion is a micro optimization because exclusive ranges still seem be faster (not that inclusive isn't also fast), but we may as well do it since we calculate it anyways.
```suggestion
            let end = i + 1;
            buf.extend_from_slice(&available[..end]);
            (true, end)
```

chippers (changes requested):
This looks ok to me in general. Ideally we wrap the `PipeReader` stuff inside a function so we can reuse the logic exactly for stdout/stderr.

The only required change that is requested is the `unwrap()` one, or an explanation of it added to the code in a comment.

As far as logic goes, I don't see anything wrong. There's more room for optimization with memchr (I think), but I really don't see the point if this isn't proven to be a bottleneck. It can always be done later since this approach doesn't really lock anything into public api

core/tauri/src/api/process/command.rs
lucasfernog: hey you should probably PR this to the rust api too :joy:

core/tauri/src/api/process/command.rs
lucasfernog: fixed it

core/tauri/src/api/process/command.rs
lucasfernog: good idea, done

PR #3566, opened 2022-02-27
fix(core): docs.rs on Windows and macOS

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


core/tauri-utils/src/pattern/isolation.rs
```diff
@@ -10,10 +10,42 @@ use std::string::FromUtf8Error;
...
+  impl Random {
+    pub fn expose(self) -> [u8; 32] {
+      [0; 32]
+    }
+  }
+
+  pub fn rand_generate(_rng: &SystemRandom) -> Result<Random, super::Error> {
+    Ok(Random {})
+  }
+}
```
chippers:
```suggestion
mod ring_impl {
  #[cfg(not(feature = "__isolation-docs"))]
  compile_error!("Isolation random number generator was used without enabling the `isolation` feature.");

  pub struct Unspecified;

  pub struct SystemRandom;

  impl SystemRandom {
    pub fn new() -> Self {
      unimplemented!()
    }
  }

  pub struct Random;

  impl Random {
    pub fn expose(self) -> [u8; 32] {
      unimplemented!()
    }
  }

  pub fn rand_generate(_rng: &SystemRandom) -> Result<Random, super::Error> {
    unimplemented!()
  }
}
```

chippers (changes requested):
I don't mind the `__isolation-docs` workaround if `ring` itself if causing `cargo doc` to fail. We just will want to make sure that *is* what causes it.  The suggested change makes it a single compile-time error if it is somehow used without enabling the isolation feature or the isolation docs workaround, along with making using it with only the docs workaround a runtime error since it's not secure.

lucasfernog:
I tested the doc generation directly using the docs.rs repo and ring was one of the issues, so i had to disable it.

PR #3572, opened 2022-02-28
fix(api): do not throw an exception if __TAURI_METADATA__ is not set, fixes #3554

### LINKED ISSUE #3554: [bug] TypeError: Cannot read properties of undefined (reading '__currentWindow')
### Describe the bug

Trying to update my app to latest Tauri I get this error `TypeError: Cannot read properties of undefined (reading '__currentWindow')`.

This is all the `@tauri-apps/api` imports I have in my app: 
import { once } from '@tauri-apps/api/event';
import { exit } from '@tauri-apps/api/process';
import type { Child } from '@tauri-apps/api/shell';
import { Command } from '@tauri-apps/api/shell';
import { writeFile, readTextFile, removeFile, readDir, createDir } from '@tauri-apps/api/fs';

I did guard all calls with `if ('__TAURI__' in window) { }` but it's not enough.

### Reproduction

_No response_

### Expected behavior

_No response_

### Platform and versions

```shell
▶ cargo tauri info

Operating System - Mac OS, version 12.2.1 X64

Node.js environment
  Node.js - 16.14.0
  @tauri-apps/cli - 1.0.0-rc.5
  @tauri-apps/api - 1.0.0-rc.1

Global packages
  npm - 8.1.4
  pnpm - Not installed
  yarn - 1.22.17

Rust environment
  rustup - 1.24.3
  rustc - 1.58.1
  cargo - 1.58.0
  toolchain - stable-aarch64-apple-darwin 

App directory structure
/node_modules
/public
/scripts
/.github
/src-tauri
/.git
/.idea
/src

App
  tauri - 1.0.0-rc.3
  tauri-build - 1.0.0-rc.3
  tao - 0.6.2
  wry - 0.13.2
  build-type - bundle
  distDir - ../build
  devPath - http://localhost:3003/
  framework - React
  bundler - Webpack
```

### Stack trace

_No response_

### Additional context

_No response_

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


tooling/api/src/window.ts
```diff
@@ -1157,13 +1157,24 @@ class WebviewWindow extends WindowManager {
...
+    window.__TAURI_METADATA__.__currentWindow.label,
+    {
+      // @ts-expect-error
+      skip: true
+    }
+  )
+} else {
+  console.error(
+    'Could not find `window.__TAURI_METADATA__`. The `appWindow` value will reference the `main` window label.'
+  )
```
kevin-legion, line 1172: Sorry, is this really an error? It seems it could be a warning.

Also, it could be nice to be able to disable this kind of warnings with an option, an [`import.meta`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import.meta) attribute or in the `window`, like `window.__TAURI_WARNING__ = false` or something.
Running an web app in a web browser seems like a rather legit use case.

JonasKruckenberg: I agree that an error is a bit much. We should change it to a `console.warn`.
Let's not add a special window property to disable errors, this constant will be deprecated and gone as soon as possible anyway.
But @lucasfernog maybe we can make the message more explicit and mention that it's not necessarily cause for concern and the warning will be removed in a future version

lucasfernog: done

PR #3618, opened 2022-03-05
feat(core): add `WindowBuilder::on_request`, closes #3533

### LINKED ISSUE #3533: [feat] allow changing the CSP before it gets injected into the webview

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/core/tauri/src/manager.rs
+++ b/core/tauri/src/manager.rs
@@ -819,28 +819,24 @@ impl<R: Runtime> WindowManager<R> {
       if let Some(csp) = &asset.csp_header {
         builder = builder.header("Content-Security-Policy", csp);
       }
-      match builder.body(asset.bytes) {
-        Ok(mut response) => {
-          if let Some(handler) = &web_resource_request_handler {
-            handler(request, &mut response);
+      let mut response = builder.body(asset.bytes)?;
+      if let Some(handler) = &web_resource_request_handler {
+        handler(request, &mut response);
 
-            // if it's an HTML file, we need to set the CSP meta tag on Linux
-            #[cfg(target_os = "linux")]
-            if let (Some(original_csp), Some(response_csp)) = (
-              asset.csp_header,
-              response.headers().get("Content-Security_Policy"),
-            ) {
-              let response_csp = String::from_utf8_lossy(response_csp.as_bytes());
-              if response_csp != original_csp {
-                let body = set_html_csp(&String::from_utf8_lossy(response.body()), &response_csp);
-                *response.body_mut() = body.as_bytes().to_vec();
-              }
-            }
+        // if it's an HTML file, we need to set the CSP meta tag on Linux
+        #[cfg(target_os = "linux")]
+        if let (Some(original_csp), Some(response_csp)) = (
+          asset.csp_header,
+          response.headers().get("Content-Security_Policy"),
+        ) {
+          let response_csp = String::from_utf8_lossy(response_csp.as_bytes());
+          if response_csp != original_csp {
+            let body = set_html_csp(&String::from_utf8_lossy(response.body()), &response_csp);
+            *response.body_mut() = body.as_bytes().to_vec();
           }
-          Ok(response)
         }
-        Err(e) => Err(e),
       }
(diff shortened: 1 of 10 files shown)


core/tauri/src/window.rs
```diff
@@ -141,16 +161,92 @@ impl<R: Runtime> WindowBuilder<R> {
...
+  ///             *csp = HeaderValue::from_str(&csp_string).unwrap();
+  ///           }
+  ///         }
+  ///       })
+  ///       .build()
+  ///       .unwrap();
+  ///     Ok(())
+  ///   });
+  /// ```
+  pub fn on_request<F: Fn(&HttpRequest, &mut HttpResponse) + Send + Sync + 'static>(
```
JonasKruckenberg: Wdyt about renaming this method to something more specific? This might clash with IPC requests callbacks we implement in the future 🤔
Maybe `on_http_request` ?

lucasfernog: Yeah I thought about that too, though `http` might confuse users too..

JonasKruckenberg: Yeah that's true, even though custom protocols are *technically* still http right?
Alternatively `on_network_request`?

lucasfernog: Yeah I think we'll have to go with http_request as it's more specific

core/tauri/src/window.rs
```diff
@@ -141,16 +161,92 @@ impl<R: Runtime> WindowBuilder<R> {
...
+  ///   window::WindowBuilder,
+  /// };
+  /// use std::collections::HashMap;
+  /// tauri::Builder::default()
+  ///   .setup(|app| {
+  ///     WindowBuilder::new(app, "core", WindowUrl::App("index.html".into()))
+  ///       .on_request(|request, response| {
+  ///         if request.uri().starts_with("tauri://") {
+  ///           // if we have a CSP header, Tauri is loading an HTML file
+  ///           //  on this exampl, let's dynamically change the CSP
```
JonasKruckenberg:
```suggestion
  ///           //  for this example, let's dynamically change the CSP
```

JonasKruckenberg (changes requested):

core/tauri/src/window.rs
JonasKruckenberg: Another option that'd be more familiar to electron converts:
`on_protocol_request`?

lucasfernog: webview2 calls it `WebResourceRequested`, I like that one, so I changed to `on_web_resource_request`.

core/tauri/src/manager.rs
```diff
@@ -801,11 +815,32 @@ impl<R: Runtime> WindowManager<R> {
...
         .replace("tauri://localhost", "");
       let asset = manager.get_asset(path)?;
-      let mut response = HttpResponseBuilder::new().mimetype(&asset.mime_type);
-      if let Some(csp) = asset.csp_header {
-        response = response.header("Content-Security-Policy", csp);
+      let mut builder = HttpResponseBuilder::new().mimetype(&asset.mime_type);
+      if let Some(csp) = &asset.csp_header {
+        builder = builder.header("Content-Security-Policy", csp);
+      }
+      match builder.body(asset.bytes) {
```
JonasKruckenberg: Why match here?

JonasKruckenberg (dismissed):

core/tauri/src/manager.rs
lucasfernog: I was going to do something with an error but it doesn't make sense, gonna revert

JonasKruckenberg (approved):

PR #3631, opened 2022-03-07
feat(core): reimplement `readTextFile` for performance

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


tooling/api/src/fs.ts
```diff
@@ -142,7 +142,7 @@ async function readTextFile(
   return invokeTauriCommand<number[]>({
```
FabianLars: this can/should be string and with that the `TextDecoder` stuff in L149 can be removed too.

lucasfernog: nice catch, totally missed that one.

PR #3686, opened 2022-03-12
feat(core): add `WindowEvent::FileDrop`, closes #3664

### LINKED ISSUE #3664: [feat] file-drop payload in rust
### Describe the problem

Currently you can only listen to file-drop events via the `listen` api which provides only a serialized string payload which is obviously not optimal.

Since i don't see a way to extend `listen` in a way that would allow non-serialized/string payloads my first idea was to extend `on_window_event` for this, but idk if that's possible since it's basically a proxy for wry events.

### Alternatives considered

_No response_

### Additional context

Can't believe it took me _months_ and multiple reminders to create this issue...

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/.changes/refactor-file-drop.md
+++ b/.changes/refactor-file-drop.md
@@ -1,0 +1,6 @@
+---
+"tauri-runtime": minor
+"tauri-runtime-wry": minor
+---
+
+The file drop event is now part of the `WindowEvent` enum instead of a having a dedicated handler.
--- a/.changes/runtime-file-drop-event.md
+++ b/.changes/runtime-file-drop-event.md
@@ -1,0 +1,5 @@
+---
+"tauri-runtime": patch
+---
+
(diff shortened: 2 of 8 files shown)


FabianLars:
With that change we could theoretically remove the `tauri://` events from rust again right? They are not reallyyy usable anyway (and afaik pretty much nobody uses them either) 🤷

Of course we still need the updater events.

lucasfernog:
Well we can also offer the updater events as part of the App::run callback. I think we'll rely on it more for v2.

lucasfernog:
@FabianLars #3687 crazy change but at least no one has to deal with the magic string event payloads on the Rust side.

Issue #3726, opened 2022-03-17
[bug] Trying to create a window with a `.` in the label will not create the window neither throw an error.


firstdorsal:
### Describe the bug

Trying to create a window with a `.` in the label will not create the window neither throw an error.

```ts
const webview = new WebviewWindow("hello.test", {
                            url: "/"
});
```
This is okay but at least not stated in the docs:
https://tauri.studio/docs/api/js/classes/window.WebviewWindow#constructor

### Platform and versions

```shell
Operating System - Arch Linux, version Rolling Release X64

Node.js environment
  Node.js - 17.7.1
  @tauri-apps/cli - 1.0.0-rc.6
  @tauri-apps/api - 1.0.0-rc.2

Global packages
  npm - 8.5.4
  pnpm - Not installed
  yarn - 1.22.18

Rust environment
  rustup - 1.24.3
  rustc - 1.59.0
  cargo - 1.59.0
  toolchain - stable-x86_64-unknown-linux-gnu 

App directory structure
/dist
/node_modules
/src-tauri
/src
/.git
/public

App
  tauri - 1.0.0-rc.4
  tauri-build - 1.0.0-rc.4
  tao - 0.6.4
  wry - 0.13.3
  build-type - build
  CSP - default-src tauri: asset: https://asset.localhost
  distDir - ../dist
  devPath - http://localhost:3000/
  framework - React
```

FabianLars:
> This is okay but at least not stated in the docs

It does mention that it requires the label to be alphanumeric. (Tbf alphanumeric can _optionally_ include additional characters like punctuation in some definitions). This still needs to be adjusted since it's alphanumeric plus `-`, `/`, `:` and `_`.

But you're right about the error, it's supposed to throw one, just like it does in rust. #3544

firstdorsal:
Oh yes you are right it is stated below but when trying to get it working everyone will read the example (like me) and try it out not reading what is stated half a page down beneath the fold.
![image](https://user-images.githubusercontent.com/10563452/158996819-bdf3e4a8-b1f2-41a7-ae9b-ba277ce03987.png)

So maybe something like this?
```ts
// the label may only contain alphanumeric characters: a-zA-Z-/:_
const webview = new WebviewWindow('theUniqueAlphaNumericLabel', {
  url: 'path/to/page.html'
})
```

Thank you!

FabianLars:
@JonasKruckenberg What do you think?
[...]

PR #3730, opened 2022-03-18
docs: update WebviewWindow doc comment

JonasKruckenberg:

- [ ] Bugfix
- [ ] Feature
- [x] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist
- [x] I have added a convincing reason for adding this feature, if necessary: See #3726

### DIFF
--- a/tooling/api/src/window.ts
+++ b/tooling/api/src/window.ts
@@ -1128,7 +1128,7 @@ class WindowManager extends WebviewWindowHandle {
 class WebviewWindow extends WindowManager {
   /**
    * Creates a new WebviewWindow.
-   * * @param label The webview window label. Must be alphanumeric: `a-zA-Z-/:_`.
+   * * @param label The unique webview window label. Must be alphanumeric: `a-zA-Z-/:_`.
    * @returns The WebviewWindow instance to communicate with the webview.
    */
   constructor(label: WindowLabel, options: WindowOptions = {}) {
--- a/tooling/api/src/window.ts
+++ b/tooling/api/src/window.ts
@@ -1095,10 +1095,10 @@ class WindowManager extends WebviewWindowHandle {
 }
 
 /**
- * Create new webview windows and get a handle to existing ones. 
+ * Create new webview windows and get a handle to existing ones.
  *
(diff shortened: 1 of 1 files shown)


JonasKruckenberg:
Does this need a changefile too? I think so right?

FabianLars:
afaik no, because there are no user facing changes 🤔

tooling/api/src/window.ts
```diff
@@ -1096,6 +1096,10 @@ class WindowManager extends WebviewWindowHandle {
 
 /**
  * Create new webview windows and get a handle to existing ones.
+ *
+ * Windows are identified by a *label*  a unique identifier that can be used to reference it later.
+ * It may only contain alphanumeric characters: `a-zA-Z-/:_`.
```
FabianLars: i'm not 100% happy with that phrasing since `-/:_` is not part of the alphanumeric subset, it's more like alphanumeric _plus_ these symbols.
But it's not that bad, so i could live with it 🤷 

Same for the other comment btw

FabianLars (changes requested):

JonasKruckenberg:
> afaik no, because there are no user facing changes 🤔

alright. I thought maybe bc these comments show up in users IDEs too but that doesn't make a whole lot of sense

tooling/api/src/window.ts
JonasKruckenberg: Check latest version, I adapted your comment on the original issue to make the main doc comment more helpful. I want to defer the meat of documentation to the main comment but I could update the second comment too

FabianLars: i think it's fine to keep the other comment short. Maybe a little hint that it needs to be unique tho? Something like `The unique webview window label`? If that's too long we can probably cut the "webview" here 🤔

JonasKruckenberg: good point, IMO we ~~drop the webview~~ and add `unique`

Edit: It's fine, we're not thaaat restricted in size

lucasfernog (approved):

PR #3739, opened 2022-03-20
feat(updater): separate intel and apple silicon targets, closes #3359

### LINKED ISSUE #3359: [feat] Update request support include CPU info
### Describe the problem

My app is built on the arm and x86 architectures of macOS, and I would like the application updates to return different download links based on CPU information.

The CPU information can be passed through the request path or the headers.

It might be supported in https://github.com/tauri-apps/tauri/blob/next/core/tauri/src/updater/core.rs#L287

### Alternatives considered

_No response_

### Additional context

_No response_

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


JonasKruckenberg:
I think this solution doesn't scale well, think about windows on ARM for example.

If it has to be s breaking change (which I also think it has to) then let's make it a proper 3rd parameter to the URL.

FabianLars:
As said on discord i agree with jonas.
Furthermore, what about darwin universal? Ofc this can be handled server side by just providing the universal bundle no matter the arch, just wanted to make sure we think about that and (officially) set on an approach here

JonasKruckenberg:
I propose we create two new variables that can optionally be present in the URL: `target_arch` `target_os`.

Those variables are replaced by their equivalent `cfg!(<variable>)` values for predictability.

While we're at it we can also consider exposing `target_feature`, `target_endian`, `target_pointer_width` and `debug_assertions` for good measure. The former can be helpful when sidecar binaries are at play and the last for opting users into a "debug build" channel

JonasKruckenberg:
Looking good 👍🏻

core/tauri/src/app.rs
```diff
@@ -1087,6 +1104,38 @@ impl<R: Runtime> Builder<R> {
     self
   }
 
+  /// Sets the current platform's target name for the updater.
+  ///
+  /// By default Tauri looks for one of `linux`, `win32`, `win64`, `darwin-silicon` or `darwin-intel`
```
JonasKruckenberg: I think this is not up to date anymore correct?

lucasfernog: oops didn't see this one, thanks

PR #3752, opened 2022-03-22
Fix importing `@tauri-apps/api` in Node.js

probablykasper:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
This makes it possible to import `@tauri-apps/api` in Node.js. If you have a dev server with SSR for whatever reason, you'll be able to import as normal. If you try to use the API in Node.js, the error will come from where you actually use it instead of where you imported it


probablykasper:
I know it's a weird one, but have actually ran into this several times with SvelteKit and had a hard time debugging

lucasfernog (changes requested):
Thanks for this fix, can you add a change file?

probablykasper:
Sure, added

lucasfernog (dismissed):

tooling/api/src/helpers/env-check.ts
```diff
@@ -0,0 +1,35 @@
...
+  return typeof window !== 'undefined'
+}
+
+function isLinux(): boolean {
+  if (isBrowser()) {
+    return navigator.appVersion.includes('Linux')
+  } else {
+    return process.platform === 'linux'
+  }
+}
```
amrbashir, line 17: This doesn't make sense at all. `isLinux()` and siblings are checks that is meant to be used in runtime only and not evaluated at compile/bundle time and shouldn't  care about Node.js environment at all. Also `@tauri-apps/api` should only work in browser environment.

amrbashir (changes requested):

tooling/api/src/helpers/env-check.ts
probablykasper: It's not for compile time, it's to make it actually be possible to import the package in SSR and as a result make errors easier to debug

amrbashir: If your SSR server is running on Linux, `isLinux()` will fallback to Node.js impl and always return true while the app actually runs on Windows or macOS.

probablykasper: I don't imagine people doing that, the only use case I really know is for dev servers (SvelteKit for example does not make it obvious how you turn off SSR). The point is not for it to actually be supported in Node.js, just to be possible to import.

amrbashir: I understand what you want to achieve but this isn't the way. Instead we need to wrap all our apis under some guards so imports won't break.

probablykasper: Then `sep`, `delimiter` and `EOL` would have to become functions

amrbashir: I don't like that right now, it is a breaking change. And as I always say, we can only do so little about SSR. Tauris apis can only be used at runtime inside Tauri only so people shouldn't use SSR with tauri, if they do then it is their job to fix any quirks about it.

probablykasper: I agree, and I'm not sure they should ever become functions either

amrbashir: Yeah, they shouldn't. SSR users should deal with this on their own and just use async imports.

probablykasper: Why, what's the issue with having it be possible to import it in this way?

amrbashir: it obviously doesn't work and the proposed solution is not a viable solution and instead introduces an unexpected behavior as I mentioned earlier where the server runs on a  Linux while the app runs on a Windows or macOS. I opened a PR to revert this. Further discussions should be there.

PR #3754, opened 2022-03-22
Support window parenting on macOS, closes #3751

### LINKED ISSUE #3751: [feat] Allow creation of child Window for macOS
### Describe the problem

It is possible to "bind" two windows as parent & child in most operating systems. By doing so, they share additional behavior which is currently not possible to replicate with Tauri. 

Right now, the solution is partially implemented in Tauri and Tao. When looking through the Git history, it seems like maybe it was "forgotten" to extend support to at least include macOS. 

In May 2021, the feature to model a parent/child relationship between windows was added for Windows OS. 
* [Link to Tauri API](https://github.com/tauri-apps/tauri/blob/5d538ec27c246274df4ff5b8057ff78b6364a43f/core/tauri-runtime/src/webview.rs#L166)
* [Link to feature request ticket](https://github.com/tauri-apps/tauri/issues/1872)

In August 2021, the same feature for macOS was included in tao. The PR owner was so kind even to include a working example that showcases the desired behavior.
* [Link to Tao PR](https://github.com/tauri-apps/tao/pull/160)
* [Link to example implementation](https://github.com/tauri-apps/tao/blob/dev/examples/parentwindow.rs)

**Desired solution:** Include tao functionality for parent/child relation into Tauri the same way as it already is for Windows.

probablykasper:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
Closes #3751
Usage:
```rust
#[command]
pub fn create_child_window(id: String, app: AppHandle) {
  let main = app.get_window("main").unwrap();

  let child = window::WindowBuilder::new(&app, id, WindowUrl::default())
    .title("Child")
    .inner_size(400.0, 300.0);

  #[cfg(target_os = "macos")]
  let child = child.parent_window(main.ns_window().unwrap());
  #[cfg(target_os = "windows")]
  let child = child.parent_window(main.hwnd().unwrap());

  child.build();
}
```

Implemented it in the same way as for Windows, but it would be nice if usage was the same for all platforms. Not sure what the way to implement that would be, maybe `RawWindowHandle`? Would also be a breaking change, unless a different name is used.


examples/parent-window/index.html
```diff
@@ -0,0 +1,62 @@
...
+    })
+    container.appendChild(createWindowButton)
+
+    var allWindows = window.__TAURI__.window.getAll()
+    for (var index in allWindows) {
+      var label = allWindows[index].label
+      if (label === windowLabel) {
+        continue
+      }
+      createWindowMessageBtn(label)
```
vdvman1: Where is this method defined? Also this wouldn't update as windows are added and removed
Not entirely sure what this section is for

probablykasper: It's copy pasted from a different example. Went and removed it :)

probablykasper:
The example is failing because the Windows `parent_window` expects `HWND` but `main.hwnd().unwrap()` returns `*mut std::ffi::c_void`. Is that a bug with the Windows implementation?

lucasfernog:
@probablykasper I think the best solution is to change `parent_window` to return a HWND so I pushed that, but now the child window is not visible (it works if I remove the `child.parent_window()` call.

probablykasper:
Hm, no clue why that would be. Any idea if it works in tao atm?

.changes/parent-window-hwnd.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri": patch
+---
+
+**Breaking change:** The `Window::parent_window` method now returns *HWND* instead of *c_void* on Windows.
```
amrbashir:
```suggestion
**Breaking change:** The `Window::hwnd` method now returns *HWND* from `windows-rs` crate instead of *c_void* on Windows.
```

amrbashir (changes requested):

lucasfernog:
Parent window directly on tao is weird but at least it shows the child window. I think wry is touching the window and messing with it too. Do you know if that's the case @amrbashir

lucasfernog (approved):

PR #3777, opened 2022-03-25
refactor(bundler): allow downgrades, add option to disallow on Windows

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/core/tauri-utils/src/config.rs
+++ b/core/tauri-utils/src/config.rs
@@ -264,6 +264,11 @@ pub struct WindowsConfig {
   /// The fixed version can be downloaded [on the official website](https://developer.microsoft.com/en-us/microsoft-edge/webview2/#download-section).
   /// The `.cab` file must be extracted to a folder and this folder path must be defined on this field.
   pub webview_fixed_runtime_path: Option<PathBuf>,
+  /// Validates a second app installation, blocking the user from installing an older version.
+  /// 
+  /// For instance, if `1.2.1` is installed, the user won't be able to install app version `1.2.0` or `1.1.5`.
+  #[serde(default)]
+  pub disallow_downgrades: bool,
   /// Configuration for the MSI generated with WiX.
   pub wix: Option<WixConfig>,
 }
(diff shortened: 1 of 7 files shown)


JonasKruckenberg:
Sorry if I'm not up to speed on this one, but why do we need to change the default behavior? Making downgrades optional would also simplify the naming :) `disallow_downgrades` -> `allow_downgrades` and make it much easier to understand

lucasfernog:
I feel most users would want to allow downgrades. And it's also the behavior on Linux and macOS too (AFAIK).

FabianLars:
as far as i am concerned it's because a) we allow downgrades on macos and linux and b) it's like the main "selling point" of the server variant of the updater.

Edit: argh, too slow :))

JonasKruckenberg:
hmmmm yeah that makes sense. What about the name though? `disallow_downgrades` is a nice alliteration but it's also a mouthful 😅

amrbashir:
I think "allow_downgrades" with default value of true is better naming.

lucasfernog:
Changed it @amrbashir @JonasKruckenberg

amrbashir (dismissed):

core/tauri-utils/src/config.rs
```diff
@@ -264,6 +264,13 @@ pub struct WindowsConfig {
   /// The fixed version can be downloaded [on the official website](https://developer.microsoft.com/en-us/microsoft-edge/webview2/#download-section).
   /// The `.cab` file must be extracted to a folder and this folder path must be defined on this field.
   pub webview_fixed_runtime_path: Option<PathBuf>,
+  /// Validates a second app installation, blocking the user from installing an older version if set to `false`.
+  ///
+  /// For instance, if `1.2.1` is installed, the user won't be able to install app version `1.2.0` or `1.1.5`.
+  ///
+  /// The default value of this flag is `true`.
+  #[serde(default)]
```
JonasKruckenberg: But isn't `serde[default]` for bools `false`?

lucasfernog: nice catch, thanks

lucasfernog: that's a problem with configuration flags that defaults to true :( way easier if it defaults to false

PR #3792, opened 2022-03-28
feat(updater): expose builder, allow setting a custom version checker

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/.changes/updater-check-api.md
+++ b/.changes/updater-check-api.md
@@ -2,4 +2,4 @@
 "tauri": patch
 ---
 
-Added `check_for_updates` method to `App` and `AppHandle`.
+Added `updater` method to `App` and `AppHandle`, a builder to check for app updates.
--- a/.changes/updater-custom-version-checker.md
+++ b/.changes/updater-custom-version-checker.md
@@ -2,4 +2,4 @@
 "tauri": patch
 ---
 
-Allow using a custom updater version checker via `App::updater().version_comparator()`.
+Allow using a custom updater version checker via `App::updater().version_checker()`.
--- a/.changes/updater-custom-version-checker.md
+++ b/.changes/updater-custom-version-checker.md
@@ -2,4 +2,4 @@
 "tauri": patch
 ---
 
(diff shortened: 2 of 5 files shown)


JonasKruckenberg:
Here is goes 😉: I felt `version_comparator` is a custom `Ord` implementation so I was confused that it didn't return `std::cmp::Ordering`.
`updater_checker` is better, but feels less descriptive and more like it's a callback that is responsible for checking the actual http endpoint.

I would suggest `should_update` instead. It's simple and returning a bool makes a lot of sense from the name alone. 

Edit: We also already use `should_update` as the value name, so why not just make that the function name too?

lucasfernog:
`should_update` feels more like a builder method to set a bool than a closure though.

lucasfernog:
I changed it to `version_checker` recently btw.

JonasKruckenberg:
> I changed it to `version_checker` recently btw.

Ah yeah sorry meant that 😅

JonasKruckenberg:
> `should_update` feels more like a builder method to set a bool than a closure though.

Good point...

How about `should_install_update`?

That way it's clearer that it means the noun update, not the verb.

lucasfernog:
> > `should_update` feels more like a builder method to set a bool than a closure though.
> 
> Good point...
> 
> How about `should_install_update`?

Better than `should_update`. If no one can come up with a better name, let's go with this one.

JonasKruckenberg:
We can also shorten it to `should_install` maybe 🤔 though `should_install_update` is clearer

Edit: I think I prefer `should_install`, since it's already clear that it's about updates from the context (the getter is called `updater()` after all)
Plus `App.updater().should_install_update()` feel unnecessarily redundant. `App.updater().should_install()` just feels nicer to me

lucasfernog:
Good point @JonasKruckenberg

JonasKruckenberg (approved):

PR #3816, opened 2022-03-30
fix(driver): expose native-host option and set default to 127.0.0.1, closes #3815

### LINKED ISSUE #3815: [bug] tauri-driver does not work in docker
### Describe the bug

Currently tauri-driver does not work in a Linux based docker container. This is because it calls the Linux WebDriver with `WebKitWebDriver --port=4445` and without the `--host` argument. The default `host` for `WebKitWebDriver` is `::1` (IPv6 equivalent of 127.0.0.1), however IPv6 cannot be used inside a docker container without explicitly [enabling it](https://docs.docker.com/config/daemon/ipv6/) in the docker daemon (which might not be feasible in all environments).

I would suggest to call the WebDriver with `WebKitWebDriver --port=4445 --host=127.0.0.1` by default. I'm preparing a PR for this.

### Reproduction

(This is the minimal reproduction I could think of. Actually installing tauri-driver and running it in docker causes the same error).

```bash
docker run --rm -it ubuntu:20.04
```

Inside docker:
```bash
apt update
apt install webkit2gtk-driver

WebKitWebDriver --port=4445
```
results in error:
> Failed to start HTTP server at port 4445: Could not listen on address ::1, port 4445: Error binding to address [::1]:4445: Cannot assign requested address

### Expected behavior

Inside docker:
```bash
WebKitWebDriver --port=4445 --host=127.0.0.1
```

-> no Error

### Platform and versions

```shell
Docker container based on `ubuntu:20.04`.

`rustc 1.59.0 (9d1b2106e 2022-02-23)`
```

### Stack trace

_No response_

### Additional context

_No response_

pwespi:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

Closes #3815.

Adding the `--host` option has no effect in Windows, but it does not harm either.

### DIFF
--- a/.changes/webdriver-native-host.md
+++ b/.changes/webdriver-native-host.md
@@ -1,0 +1,5 @@
+---
+"tauri-driver": patch
+---
+
+Expose `native-host` option in tauri-driver and set default to `127.0.0.1`.
--- a/tooling/webdriver/src/cli.rs
+++ b/tooling/webdriver/src/cli.rs
@@ -13,6 +13,7 @@ FLAGS:
 OPTIONS:
   --port NUMBER           Sets the tauri-driver intermediary port
   --native-port NUMBER    Sets the port of the underlying WebDriver
+  --native-host HOST      Sets the host of the underlying WebDriver (Linux only)
   --native-driver PATH    Sets the path to the native WebDriver binary
 ";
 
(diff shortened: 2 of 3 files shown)


chippers (approved):

pwespi:
@chippers Thanks a lot for the quick merge and release. And thank you for your work on tauri-driver. It is very valuable. :-)

Issue #2358, opened 2021-08-05
feat(macos): dock management api 


OceanBelongsToMe:
### **Platform and Versions (required):**

```
Operating System

Node.js environment
Node.js 
@tauri-apps/cli 
@tauri-apps/api 

Global packages
npm 
yarn 

Rust environment
rustc 
cargo 

App directory structure


App
tauri.rs 
build-type 
CSP 
```

probablykasper:
Hard to say what you mean because you didn't say much. Windows don't show up in the macOS Dock, so `window.set_skip_taskbar` isn't something that works on macOS. If you're talking about hiding the app from the Dock, that's mentioned in https://github.com/tauri-apps/tauri/issues/2258

amrbashir:
It is not possible for macOS so I will rename this issue to dock management.

amrbashir:
related to tauri-apps/tao#105

betamos:
My use case, for whoever is looking at this in the future:

The user can control whether the app runs in (a) both foreground and background or (b) foreground only, using a checkbox:

- Tray icon is shown if and only if background mode is enabled.
- Dock icon is always shown when main window is open.
- Dock icon is hidden if main window is closed, under mode (a).

`set_activation_policy` almost solves my problem, but:

1. I can't seem to change it when the app is running, since it's exposed as `&mut self` on the `App` struct. (I'm not an expert on MacOS, but it appears this should be allowed, perhaps through the AppHandle?)
2. When set to `Accessory`, the dock icon is not shown, even if main window is open. (This appears to be MacOS default, so not expecting it to be overridden by Tauri).

betamos:
Check out these commits here:

https://github.com/betamos/tauri/commits/betamos-head

(Relies on an unpublished addition to Tao)

I got everything to work except for one issue:

- User persists app in dock (either through "Show recent" or by adding it manually to the dock)
- Set activation policy to accessory/prohibited (disables icon, but doesn't remove it)
- Click icon (this doesn't *show* the app automatically, and there's no way to intercept the event atm, see https://github.com/tauri-apps/tao/issues/218)

It works well otherwise, and everything I added is platform-gated with `cfg(target_os="macos")`. It's basically a thin layer on top of MacOS APIs. I'd be happy to contribute but someone should probably vet/approve/feedback the approach first.

## Alternative approach

[...]

PR #3840, opened 2022-04-02
macOS: Expose new Tao APIs

betamos:
These changes exposes Tao APIs that are only affecting MacOS:

- `Window.show_application()`
- `Window.hide_application()`
- `Window.set_application_policy_at_runtime(..)`

Following precedent set by `set_activation_policy`, these APIs are all target-gated to MacOS only.

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

Related to #2358.

### DIFF
--- a/.changes/set-activation-policy-at-runtime.md
+++ b/.changes/set-activation-policy-at-runtime.md
@@ -1,0 +1,7 @@
+---
+"tauri": patch
+---
+
+MacOS: Added `Window.set_activation_policy_at_runtime` method, which allows
+dynamically changing the activation policy (previously only possible at
+application-build time).
--- a/.changes/set-activation-policy-at-runtime.md
+++ b/.changes/set-activation-policy-at-runtime.md
@@ -2,6 +2,6 @@
 "tauri": patch
 ---
 
-MacOS: Added `Window.set_activation_policy_at_runtime` method, which allows
+macOS: Added `Window.set_activation_policy_at_runtime` method, which allows
 dynamically changing the activation policy (previously only possible at
 application-build time).
--- a/.changes/show-hide-application.md
+++ b/.changes/show-hide-application.md
@@ -1,0 +1,7 @@
+---
+"tauri": patch
+---
+
+MacOS: Added `Window.show_application` and `Window.hide_application` method,
+which shows or hides the entire app. The `Window.show` and `Window.hide`
+methods are unaffected by this change.
--- a/.changes/show-hide-application.md
+++ b/.changes/show-hide-application.md
@@ -2,6 +2,6 @@
 "tauri": patch
 ---
 
-MacOS: Added `Window.show_application` and `Window.hide_application` method,
+macOS: Added `Window.show_application` and `Window.hide_application` method,
 which shows or hides the entire app. The `Window.show` and `Window.hide`
 methods are unaffected by this change.
--- a/core/tauri-runtime/src/lib.rs
+++ b/core/tauri-runtime/src/lib.rs
@@ -528,6 +528,14 @@ pub trait Dispatch<T: UserEvent>: Debug + Clone + Send + Sync + Sized + 'static 
   /// Hides the window.
   fn hide(&self) -> Result<()>;
 
+  /// Shows the application on MacOS (independent of current window).
+  #[cfg(target_os = "macos")]
+  fn show_application(&self) -> crate::Result<()>;
+
+  /// Hides the application on MacOS (independent of current window).
+  #[cfg(target_os = "macos")]
+  fn hide_application(&self) -> crate::Result<()>;
+
   /// Closes the window.
   fn close(&self) -> Result<()>;
 
--- a/core/tauri-runtime/src/lib.rs
+++ b/core/tauri-runtime/src/lib.rs
@@ -261,6 +261,7 @@ pub struct RunIteration {
 /// Application's activation policy. Corresponds to NSApplicationActivationPolicy.
 #[cfg(target_os = "macos")]
 #[cfg_attr(doc_cfg, doc(cfg(target_os = "macos")))]
+#[derive(Debug, Clone)]
 #[non_exhaustive]
 pub enum ActivationPolicy {
   /// Corresponds to NSApplicationActivationPolicyRegular.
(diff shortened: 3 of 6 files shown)


core/tauri-runtime/src/lib.rs
```diff
@@ -528,6 +529,21 @@ pub trait Dispatch<T: UserEvent>: Debug + Clone + Send + Sync + Sized + 'static
   /// Hides the window.
   fn hide(&self) -> Result<()>;
 
+  /// Shows the application on MacOS (independent of current window).
```
JonasKruckenberg: I think some more explanations would be helpful here, what does hiding the app mean in detail? Does it affect the dock icon? Does it affect Mission Control? Or the app quick switcher?

JonasKruckenberg (changes requested):

amrbashir:
This is the same as #3689, we won't merge this anytime soon. These api will have to wait for v2.

core/tauri-runtime/src/lib.rs
betamos: I'd love to, but the only way to know would be experimental. Apple is deliberately vague about this, probably because things like dock and mission control are subject to changes across MacOS versions. Being more detailed than Apple's own docs won't be much help. Here's an example of what we call `show_application`:

https://developer.apple.com/documentation/appkit/nsapplication/1428761-unhide?language=objc

betamos:
@amrbashir noted, and sorry for creating yet another duplicate PR. Same story, I didn't know this was already implemented. For context, these features are important for MacOS apps to behave in line with user expectations, that's likely why we both arrived at the same features.

Are we blocked due to feature freeze for 1.0 or because we haven't yet approved the API design? If the latter, we could parallelize the design discussions without putting load on the team.

Note the difference between this PR and #3689:

- This PR also contains the `set_activation_policy_at_runtime` method.
- This PR exposes methods on `Window` whereas the other exposes them at `AppHandle` (which is arguably a better design)

amrbashir:
It is blocked because we can't introduce new APIs that isn't audited so it will have to wait for v2 so it can be audited.

.changes/set-activation-policy-at-runtime.md
```diff
@@ -0,0 +1,7 @@
+---
+"tauri": patch
+---
+
+MacOS: Added `Window.set_activation_policy_at_runtime` method, which allows
```
lorenzolewis:
```suggestion
macOS: Added `Window.set_activation_policy_at_runtime` method, which allows
```

.changes/show-hide-application.md
```diff
@@ -0,0 +1,7 @@
+---
+"tauri": patch
+---
+
+MacOS: Added `Window.show_application` and `Window.hide_application` method,
```
lorenzolewis:
```suggestion
macOS: Added `Window.show_application` and `Window.hide_application` method,
```

core/tauri-runtime/src/lib.rs
lorenzolewis:
```suggestion
  /// Shows the application on macOS (independent of current window).
```

core/tauri-runtime/src/lib.rs
```diff
@@ -528,6 +529,21 @@ pub trait Dispatch<T: UserEvent>: Debug + Clone + Send + Sync + Sized + 'static
...
+  #[cfg(target_os = "macos")]
+  fn show_application(&self) -> crate::Result<()>;
+
+  /// Hides the application on MacOS (independent of current window).
```
lorenzolewis:
```suggestion
  /// Hides the application on macOS (independent of current window).
```

core/tauri-runtime/src/lib.rs
```diff
@@ -528,6 +529,21 @@ pub trait Dispatch<T: UserEvent>: Debug + Clone + Send + Sync + Sized + 'static
...
+  #[cfg(target_os = "macos")]
+  fn hide_application(&self) -> crate::Result<()>;
+
+  /// Sets the activation policy at runtime on MacOS (independent of current window).
```
lorenzolewis:
```suggestion
  /// Sets the activation policy at runtime on macOS (independent of current window).
```

core/tauri/src/window.rs
```diff
@@ -1058,6 +1061,39 @@ impl<R: Runtime> Window<R> {
     self.window.dispatcher.hide().map_err(Into::into)
   }
 
+  /// Shows the application on MacOS (independent of current window).
```
lorenzolewis:
```suggestion
  /// Shows the application on macOS (independent of current window).
```

core/tauri/src/window.rs
```diff
@@ -1058,6 +1061,39 @@ impl<R: Runtime> Window<R> {
...
+  #[cfg(target_os = "macos")]
+  pub fn show_application(&self) -> crate::Result<()> {
+    self
+      .window
+      .dispatcher
+      .show_application()
+      .map_err(Into::into)
+  }
+
+  /// Hides the application on MacOS (independent of current window).
```
lorenzolewis:
```suggestion
  /// Hides the application on macOS (independent of current window).
```

core/tauri/src/window.rs
```diff
@@ -1058,6 +1061,39 @@ impl<R: Runtime> Window<R> {
...
+  #[cfg(target_os = "macos")]
+  pub fn hide_application(&self) -> crate::Result<()> {
+    self
+      .window
+      .dispatcher
+      .hide_application()
+      .map_err(Into::into)
+  }
+
+  /// Sets the activation policy at runtime on MacOS (independent of current window).
```
lorenzolewis:
```suggestion
  /// Sets the activation policy at runtime on macOS (independent of current window).
```

lucasfernog:
@betamos we've merged the show/hide APIs from #3689, can you open a new PR for the `set_application_policy_at_runtime` API, following the same strategy (adding the APIs to App and AppHandle instead of Window)?

PR #3865, opened 2022-04-06
fix: add priority field in debian/control

volgar1x:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

Tauri's bundler currently do not add `Priority:` field in `debian/control` file which is [documented here](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-priority). It is a simple addition that may benefit companies which validate content of `debian/control` and require that a `Priority:` field is effectively set.

### DIFF
--- a/.changes/debian-priority.md
+++ b/.changes/debian-priority.md
@@ -1,0 +1,5 @@
+---
+"tauri-bundler": patch
+---
+
+Set the Debian control file `Priority` field to `optional`.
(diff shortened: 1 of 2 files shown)


lucasfernog:
Thanks!

Issue #3699, opened 2022-03-14
Do not expose tauri-runtime on tauri


lucasfernog:
### Describe the problem

The tauri-runtime crate will stay on the alpha releases while Tauri stable is released. For this reason, we cannot expose tauri-runtime (`pub use tauri_runtime as runtime`).

### Describe the solution you'd like

We need a dedicated module that only exposes what is stable for users to import).

FabianLars:
#3749

PR #3886, opened 2022-04-11
feat(core): Expose wry and tao

dceddia:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

The goal of this change is to be able to call tao APIs (like platform-native ones on WindowBuilder) from Tauri without having to separately depend on tao.

This change depends on this one in wry: https://github.com/tauri-apps/wry/pull/549

### DIFF
--- a/core/tauri-runtime-wry/src/lib.rs
+++ b/core/tauri-runtime-wry/src/lib.rs
@@ -83,7 +83,7 @@ pub use wry::application::platform::macos::{
 };
 
 pub use wry;
-pub use wry::tao;
+pub use wry::application as tao;
 
 use std::{
   collections::{
--- a/core/tauri-runtime-wry/src/lib.rs
+++ b/core/tauri-runtime-wry/src/lib.rs
@@ -82,6 +82,9 @@ pub use wry::application::platform::macos::{
   NativeImage as WryNativeImage, WindowExtMacOS,
 };
 
+pub use wry;
+pub use wry::tao;
+
 use std::{
   collections::{
     hash_map::Entry::{Occupied, Vacant},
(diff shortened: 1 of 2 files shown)


core/tauri-runtime-wry/src/lib.rs
```diff
@@ -82,6 +82,9 @@ pub use wry::application::platform::macos::{
   NativeImage as WryNativeImage, WindowExtMacOS,
 };
 
+pub use wry;
+pub use wry::tao;
```
amrbashir:
```suggestion
pub use wry::application as tao;
```

amrbashir (changes requested):

amrbashir (approved):

FabianLars:
Doesn't this change would re-introduce the same problem as #3699?

dceddia:
Oh, bummer, yeah this looks like it's reintroducing that.

Maybe my issue can be solved another way. I'm using some APIs from `tao`, and right now tauri doesn't expose tao, so I'm depending on tao in my Cargo.toml directly. The problem then is if tauri updates to a newer version of tao, like what happened between rc.4 and rc.5 (or 6, I forget now) where tao went from 0.6.4 to 0.7.0, my app won't compile until I update that dependency.

[This forum post](https://users.rust-lang.org/t/add-a-crate-with-same-version-as-a-dependency-is-using/17389/12) sounds like what I want cargo to do, basically, "I don't care which `tao` you use, just make it the same as the one `tauri` uses". (Except transitively, because I guess tauri doesn't directly depend on tao) Maybe specifying `tao = ">=0"` will work for the time being?

lucasfernog:
Thanks for catching that Fabian. Too bad we can't expose tao and wry as it's not stable.

I face the same problem at work, we need to update everything all the time. It's ok but it can be annoying sometimes :/

PR #3934, opened 2022-04-21
fix(cli): handle symlinks in updater bundler, closes #3933

### LINKED ISSUE #3933: [bug] Error while creating updater tar.gz
### Describe the bug

I'm bundling an app for distribution that includes some macOS Frameworks. These Frameworks have symlinks inside them, which is pretty common. The bundle process fails with this error:

```
Error: failed to bundle project

Caused by:
    0: Failed to tar.gz update directory
    1: Failed to tar.gz update directory
    2: `Is a directory (os error 21)`
    3: Is a directory (os error 21)
```

Digging into the code for the bundler, [this section](https://github.com/tauri-apps/tauri/blob/dev/tooling/bundler/src/bundle/updater_bundle.rs#L219-L224) that creates the tar.gz file is expecting each path to be a directory or a file. This fails to process the Frameworks because they have symlinks that point to directories – it's taking the `else` branch when the path is not actually a file.

Here's a PR that seems to fix the issue: https://github.com/tauri-apps/tauri/pull/3934

### Reproduction

In `tauri.conf.json`, under `macOS.frameworks`, specify a framework that include symlinks. Any old framework on the system will work.

Here's [a minimal example repo](https://github.com/dceddia/tauri-build-fails-with-framework/tree/main) where I'm pointing to a framework inside Google Chrome.app

### Expected behavior

It should build and create the .tar.gz file for the updater.

### Platform and versions

```shell
Environment
  › OS: Mac OS 12.0.0 X64
  › Node.js: 16.13.0
  › npm: 8.1.0
  › pnpm: Not installed!
  › yarn: 1.22.18
  › rustup: 1.24.3
  › rustc: 1.59.0
  › cargo: 1.59.0

Packages
  › @tauri-apps/cli [NPM]: 1.0.0-rc.8
  › @tauri-apps/api [NPM]: 1.0.0-rc.3
  › tauri [RUST]: 1.0.0-rc.4,
  › tauri-build [RUST]: 1.0.0-rc.4,
  › tao [RUST]: 0.6.4,
  › wry [RUST]: 0.13.3,

App
  › build-type: bundle
  › CSP: unset
  › distDir: ../build
  › framework: Svelte
  › bundler: Rollup

App directory structure
  ├─ node_modules
  ├─ public
  ├─ scripts
  ├─ src-tauri
  ├─ build
  ├─ .vscode
  └─ src
```

### Stack trace

_No response_

### Additional context

_No response_

dceddia:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


lucasfernog:
Nice catch! Though linking frameworks shows how slow the extract process is. I'll address that in another PR.

lucasfernog:
Do we also need to handle symlinks [here](https://github.com/tauri-apps/tauri/blob/c54ddfe9338e7eb90b4d5b02dfde687d432d5bc1/core/tauri/src/api/file/extract.rs#L220)? I'm not sure how macOS links frameworks in the app bundle.

dceddia:
Ooh yeah, probably need to handle symlinks at extraction time too. I imagine otherwise it’ll end up creating hard directories and files and/or falling over.

lucasfernog:
The #3963 change is essential for frameworks in the updater bundle since the extract process was super slow (and a framework folder has a lot of files).

dceddia:
I added a test to check that symlinks inside frameworks extract correctly, and updated the `extract` function to handle symlinks. I noticed the built-in `entry.unpack()` function seems to handle all the different types of files, so I went with that. (it didn't handle directories unfortunately, so I left the branch that deals with those).

dceddia:
Even though the test is passing and the symlinks look good after a test run, this isn't working in my actual app. Not sure why yet...

Update: oof, nevermind me. I had built and installed a local copy of `tauri-cli` but my app was still building against the rc10 from crates.io instead of this version with the fixes 🤦‍♂️ Once I point my Cargo.toml at this copy it works fine.

dceddia:
Heads up that I think I found an issue with this where it's making symlinks to absolute paths, maybe only sometimes, maybe only if the app name has a space in it. I'm looking into it, but just wanted to leave a note here not to merge this yet.

dceddia:
I think this is good now. I've been running with the last fix ("avoid broken symlinks in updater bundle") for a few releases of my app, haven't seen any problems myself, and haven't had any complaints from users about Mac updates being broken.

HeavenVolkoff:
Hello,

Just stumbled upon the issue linked to this PR while working on https://github.com/spacedriveapp/spacedrive/pull/768. Our app requires a `framework` to be bundled, but after linking it in `tauri.macOS.frameworks` the bundle step for the updater started to fail with the error message `Failed to tar.gz update directory: 'Is a directory (os error 21)'`.

Is there anything else that needs to be addressed for this PR to move forward?

lucasfernog:
Let's try to release it on 1.4, I'll let the auditors know about this.

wusyong:
@lucasfernog what's breaking changes in this PR?

lucasfernog:
There's no breaking changes, it could be a patch too.

HeavenVolkoff:
> There's no breaking changes, it could be a patch too.

It would be great if this could be included in a patch release.

Also, the failed test seems to be just a typo: `bundle_path` should be `bundle_paths` here https://github.com/tauri-apps/tauri/pull/3934/files#diff-e7c0c60e9bef4a36a5940e86d4b58739b653f64ce5b116a84a6bd3a64f7268aaR317

core/tauri/src/api/file/extract.rs
```diff
@@ -100,13 +101,8 @@ impl<'a, R: Read> Entry<'a, R> {
...
         } else {
-          let mut out_file = fs::File::create(into_path)?;
-          io::copy(&mut entry, &mut out_file)?;
-
-          // make sure we set permissions
-          if let Ok(mode) = entry.header().mode() {
-            set_perms(into_path, Some(&mut out_file), mode, true)?;
-          }
+          // handle files, symlinks, hard links, etc. and set permissions
+          entry.unpack(into_path)?;
```
chippers, line 116: Using [`Entry::unpack_in`](https://docs.rs/tar/latest/tar/struct.Entry.html#method.unpack_in) will help prevent unpacking a file outside the destination.
```suggestion
          entry.unpack_in(into_path)?;
```

chippers (dismissed):

core/tauri/src/api/file/extract.rs
lucasfernog: We actually need to use unpack() here. The implementation is completely messed up, but it basically needs to extract `YourApp-0.1.1.app/Contents/MacOS/file` into `/tmp/tauri-update-blabla/ContentsMacOS/file` and later we rename `/tmp/tauri-update-blabla` to `/path/to/YourApp.app`

chippers: oof

lucasfernog: But what we can do it check if the path has any parent component and error out... I don't think it should ever happen.

chippers (approved):

morajabi:
can we have this in v2 as well?

lucasfernog:
Yeah @morajabi we'll merge 1.x into dev for the last time when 1.5 lands.

PR #3960, opened 2022-04-25
feat(cli): check and notify about updates on `tauri dev`, closes #3789

### LINKED ISSUE #3789: [feat] Show a message if a newer version of `tauri-cli` is available
### Describe the problem

`tauri-cli` can be installed via `npm` or `cargo`.
At the moment there is no opportunity to know that the new version of `tauri-cli` came out.

My suggestion is to show a message if a new version of `tauri-cli` is available.
This will be an indicator for a user that it's time to update.

### Alternatives considered

I found this useful feature in the [GitHub CLI application](https://github.com/cli/cli/blob/trunk/internal/update/update.go) and [npm](https://github.com/npm/cli/blob/latest/lib/utils/update-notifier.js).

### Additional context

Ready-made crates to implement that:
* https://github.com/mgrachev/update-informer
* https://github.com/tarikeshaq/update-notifier

codekidX:
this feature helps the users of tauri know that there is a new version of CLI avaliable. it uses `metadata.json` from the tooling/cli directory (raw asset from github) and checks with the bundled version of metadata.json.

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist
- [ x] I have added a convincing reason for adding this feature, if necessary

### DIFF
--- a/tooling/cli/src/dev.rs
+++ b/tooling/cli/src/dev.rs
@@ -78,6 +78,7 @@ pub fn command(options: Options) -> Result<()> {
 fn command_internal(options: Options) -> Result<()> {
   let logger = Logger::new("tauri:dev");
 
+  #[cfg(not(debug_assertions))]
   match check_for_updates() {
     Ok((msg, sleep)) => {
       if sleep {
@@ -295,6 +296,7 @@ fn command_internal(options: Options) -> Result<()> {
   }
 }
 
+#[cfg(not(debug_assertions))]
 fn check_for_updates() -> Result<(String, bool)> {
   let current_version = crate::info::cli_current_version()?;
   let current = semver::Version::parse(&current_version)?;
--- a/tooling/cli/src/dev.rs
+++ b/tooling/cli/src/dev.rs
@@ -78,6 +78,20 @@ pub fn command(options: Options) -> Result<()> {
 fn command_internal(options: Options) -> Result<()> {
   let logger = Logger::new("tauri:dev");
 
+  match check_for_updates() {
+    Ok((msg, sleep)) => {
+      if sleep {
+        logger.log(msg);
+        std::thread::sleep(std::time::Duration::from_secs(3));
+      } else {
+        logger.log(msg);
+      }
+    }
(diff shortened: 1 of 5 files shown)


lucasfernog (approved):
Nice PR :) thank you!

tooling/cli/src/dev.rs
```diff
@@ -78,6 +78,21 @@ pub fn command(options: Options) -> Result<()> {
 fn command_internal(options: Options) -> Result<()> {
   let logger = Logger::new("tauri:dev");
 
+  #[cfg(not(debug_assertions))]
+  match check_for_updates() {
+    Ok((msg, sleep)) => {
+      if sleep {
+        logger.log(msg);
+        std::thread::sleep(std::time::Duration::from_secs(3));
```
amrbashir, line 86: Why 3 secs sleep when there is a new version available?

codekidX: @amrbashir it is to make user aware of the update, if we don't sleep at that time next command might clear the terminal buffer and user might miss it completely or not care about it.

amrbashir: but that delay is bad for UX IMO. Like I'd want the dev process to start as soon as possible.

codekidX: what can be an alternative, may be ask for auto update (Y/n) instead?

amrbashir: No, that also is a bad UX. I think we should just remove this delay. And if users don't see the notification, it is not that important. 

Also note that after stable 2.0 release, users might deliberately choose to stay on 1.x for stability and we probably shouldn't annoy them because of that.

I honestly think we should remove this notification after 1.0 stable release is out because the CLI and other Tauri components will be stable by then and there will be no breaking changes moving forwards so you wouldn't need to update the CLI as often as the "beta" or "rc" phase.

Issue #1656, opened 2021-04-29
Change `drag-region` to data attribute


lucasfernog:
Currently the draggable region is marked with a class (`drag-region`) but we've been discussing on Discord about changing to be a data attribute instead.

jecsham:
@lucasfernog 
This sometimes doesn't work properly due to overlaped layers

### Working
```html
        <div  data-tauri-drag-region className="title-panel">
          
        </div>
```

### Not working
```html
        <div  data-tauri-drag-region className="title-panel">
          <div className="uk-text-small uk-text-muted p-n">
            <p id="text-title">
                App Title
            </p>
          </div>
        </div>
```

lucasfernog:
Hey @jecsham can you file an issue so we don't forget about this?

jecsham:
> Hey @jecsham can you file an issue so we don't forget about this?

Sure, I will prepare one

FabianLars:
No need for a new issue. We already tried fixing it, but it's actually not possible without breaking stuff. So instead you always need to add the attribute to the actual click target(s) (most of the time it's the innermost element(s).

ref: #3031

betamos:
I tried using this attribute, but gave up after playing whack-a-mole with adding `data-tauri-drag-region` to every non-interactive element and it's children. If you have a fixed draggable region, this is... managable. But if you (like me) want everything non-interactive to be draggable, it's not maintainable.

Tried a few options until I found something that works fantastic for me:

```js
import { window as tauriWindow } from "@tauri-apps/api";
// ...
const noDragSelector = "input, a, button"; // CSS selector
document.addEventListener("mousedown", async e => {
    if (e.target.closest(noDragSelector)) return; // a non-draggable element either in target or its ancestors
    await tauriWindow.appWindow.startDragging();
});
```

This solution technically has the opposite issue - playing whack-a-mole with _interactive_ elements, but, this is a much easier task in practice. I can't speak for all apps, but it may be worth considering for Tauri. For instance, we could make use of the (currently unused) attribute value to provide the above mentioned behavior:

```html
<div data-tauri-drag-region> works exactly like it does today </div>
<div data-tauri-drag-region="input, a, textarea"> custom no-drag selector </div>
<div data-tauri-drag-region="default"> some sensible default no-drag-selector </div>
```

mokeyish:
> @lucasfernog This sometimes doesn't work properly due to overlaped layers
> ### Working
> 
> ```
[...]

PR #3965, opened 2022-04-25
feat(macos): `title_bar_style` and `hidden_title` window options, closes #2663

### LINKED ISSUE #2663: Feature request: Add setting for titlebar style with native window controls support

When creating frameless windows you're required to provide your own UI for window controls. This has multiple problems with consistency and accessibility and requires more work to get close to the UX that is already provided by the native window system.

For example on macOS there are different actions based on how you interact with the window controls ("traffic lights"):

- Unfocused windows makes traffic lights appear gray
    - Unfocused windows still respond to mouse events for the traffic lights such as hovering. 
- Hovering the traffic lights will display the icons, based on system preference.
- Holding `Alt`/`Option` key while clicking on "resize" will maximize the window as opposed to the default fullscreen behavior.
- Traffic lights can still be used even if the underlying application becomes unresponsive.

And many more edge-cases that needs to be handled in order to provide native-like experience.

These cases only describes macOS-specific parts, but could easily apply to Windows and Linux as well, which would require more work to provide native-like UI.

Frameless windows also has several problems as outlined in https://github.com/tauri-apps/tauri/issues/2549 and described in [Electron's documentation](https://www.electronjs.org/docs/api/frameless-window#custombuttonsonhover)

Support the ability to set the titlebar style [similarly to Electron's `titleBarStyle` setting](https://www.electronjs.org/docs/latest/tutorial/window-customization#apply-custom-title-bar-styles-macos-windows) which allows for custom title bars while preserving native window controls in different ways.

I don't believe there are any alternatives and needs to be solved within Tauri.

probablykasper:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
I assume this needs to wait for v2


core/tauri-runtime-wry/src/lib.rs
```diff
@@ -878,6 +878,14 @@ impl WindowBuilder for WindowBuilderWrapper {
     self
   }
 
+  #[cfg(target_os = "macos")]
+  fn transparent_titlebar(mut self, transparent: bool) -> Self {
+    self.inner = self.inner.with_title_hidden(true);
+    self.inner = self.inner.with_fullsize_content_view(true);
+    self.inner = self.inner.with_titlebar_transparent(transparent);
```
amrbashir: I think it would be better to expose each method alone, so users can mix and match between them as they see fit. Or provide a similar interface to electron's `titlebarStyle` so it can be used in `tauri.conf.json`

amrbashir (changes requested):
> I assume this needs to wait for v2

These APIs existed when tao was being audited so if it is really a high demand feature, we can maybe include it?

core/tauri-runtime-wry/src/lib.rs
probablykasper: I tried that originally, but the title was hidden and content view fullsize anyway, not sure why. Electron's `titleBarStyle` only really lets you show or hide the whole titlebar, the other options it has (`hiddenInset` and `customButtonsOnHover`) seem to just be for controlling the traffic light buttons

probablykasper:
> These APIs existed when tao was being audited so if it is really a high demand feature, we can maybe include it?

It is the most upvoted issue, so it sure does have demand

core/tauri-runtime-wry/src/lib.rs
amrbashir: > I tried that originally, but the title was hidden and content view fullsize anyway, not sure why. 

Then we need to look into why this happens. I still would like each function to be exposed on its own.

probablykasper: Enabling `with_title_hidden` on it's own does work

xuchaoqian:
> These APIs existed when tao was being audited so if it is really a high demand feature, we can maybe include it?

Yes,  it is really a high demand feature. 
Lots of apps need to customize title bar. This is a very useful and basic feature for them.

core/tauri-runtime-wry/src/lib.rs
probablykasper: So the issue is only that `transparent_titlebar` results in `fullsize_content_view` - `hidden_title` works fine, it was just white on a white background for me.

`fullsize_content_view` is being set in the `create_webview` function [here](https://github.com/tauri-apps/tauri/blob/a67005d96bbea36a50532d76e78d3d843b89b16a/core/tauri-runtime-wry/src/lib.rs#L2771-L2774). Any idea why that's being set there, or who would know?

probablykasper: Alright so that line is there due to https://github.com/tauri-apps/tauri/pull/3915

amrbashir: tbh I'd prefer reverting that PR since it only happens in devtool, and leave the option for users to enable/disable full content size. Alternatively, since that bug only happened in dev, maybe apply the full content view only in dev mode. cc @lucasfernog

lucasfernog: exposing all functions separately might be too low level, but i like an interface like `titlebarStyle` from Electron.

probablykasper: I'd agree that `fullsize_content_view` is too low level, but I do like having `transparent_titlebar` and `hidden_title` separate

lucasfernog: we can probably offer both ways :D

probablykasper: Something like this?
```rs
enum TitleBarStyle {
  Visible,
  Transparent,
  Hidden,
}
window.title_bar_style(TitleBarStyle::Transparent);
```

betamos:
I use this in my Tauri fork. I love it.

Screen Shot 2022-06-24 at 12 31 34 PM

Some thoughts on the feature itself:

- It's much preferable to use real window chroming than faking it with browser markup, especially for MacOS:
  - Alt-button works and changes the meaning of buttons.
  - Hovering gives me menus of moving window to the side/different screen, etc.
  - Traffic lights tone out to gray when window is blurred.
  - Whenever the OS vendor changes their UI, the app won't look outdated.
- Details matter. Just like how we want the Tauri website to look professional and modern, many designers, devs and users care about those same things. TL;DR: I think it's worth prioritizing.
- Dragging the app can become less intuitive without title bar and it's important to consider from a UX perspective. I solved it by making all non-interactive elements draggable ([here's how](https://github.com/tauri-apps/tauri/issues/1656#issuecomment-1161495124)). This improves UX also on other platforms.

Some technical thoughts:

- You are likely to want to want minor MacOS CSS tweaks due to the visual differences of the title bar:
  - Platform CSS media-queries don't work that well, so...
  - You have to check platform through `tauri.os.platform()` and change the CSS that way
  - [untested] Use [small/large viewport units](https://css-tricks.com/the-large-small-and-dynamic-viewports/) in your CSS.
- The `fullsize_content_view` param does not appear to have any effect. I have both `transparent_titlebar` and `hidden_title` turned on.
- I think the enum API would be better than booleans. It's easier to document and takes up shorter space. It also prevents invalid/ineffective combinations of the underlying booleans.

probablykasper:
I'm not sure if going the `TitleBarStyle` enum route is ideal. There are a few different situations to cover:
- Showing the title bar
- Transparent titlebar with fullsize content view. This allows you to put html under the titlebar.
- Transparent titlebar without fullsize content view. This gives you a different look without needing to worry about different OS versions having different title bar heights and you don't run into https://github.com/tauri-apps/tauri/issues/4316

None of the above "hide" the titlebar, because the title and traffic lights can still be visible. In fact, in tao (and Xcode storyboards), hiding the titlebar
- hides the traffic lights
- disables the window border
- disables rounded window corners

probablykasper:
Went with this:
```rs
enum TitleBarStyle {
  Visible,
  Transparent,
  Hidden,
}
```

JonasKruckenberg:
> I solved it by making all non-interactive elements draggable

You mean all elements within the titlebar?

> Went with this:
> ```rs
> enum TitleBarStyle {
>  Visible,
>  Transparent,
>  Hidden,
> }
> ```

I like this 👍

betamos:
> You mean all elements within the titlebar?

No, everywhere in the window. I've seen it on some apps in the wild and it's great. Works for people that assume you have to drag from top too, so no learning curve, and no other downsides afaik.

probablykasper:
> I like this 👍

@JonasKruckenberg Sorry, this is actually what I went with:
```rust
enum TitleBarStyle {
Visible,
Transparent,
Overlay,
}
```

The reasoning is what I explained above

xuchaoqian:
Can we merge this PR? Many people are waiting for this feature...

graddotdev:
Waiting this feature too...

p0ryae:
Looking forward to the merge. This feature is highly in demand with people.

lucasfernog:
Looks good to me, but I'm not seeing any differences with the Transparent mode (though Overlay looks amazing!) @probablykasper

lucasfernog (approved):

PR #3974, opened 2022-04-26
fix(tauri): docs.rs build error

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/.changes/fix-tauri-docsrs.md
+++ b/.changes/fix-tauri-docsrs.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch
+---
+
+Fixes `docs.rs` documentation build.
--- a/core/tauri/Cargo.toml
+++ b/core/tauri/Cargo.toml
@@ -67,7 +67,6 @@ tar = "0.4.36"
 tempfile = "3"
 zip = { version = "0.6", default-features = false, optional = true }
 ignore = "0.4"
-either = "1.6"
 flate2 = "1.0"
 http = "0.2"
 bincode = "1.3"
--- a/core/tauri/Cargo.toml
+++ b/core/tauri/Cargo.toml
@@ -26,9 +26,12 @@ features = [
   "api-all",
   "cli",
   "__updater-docs",
+  "__fs-extract-api-docs",
   "system-tray",
   "devtools",
   "http-multipart",
+  "gtk-tray",
+  "icon-png",
   "dox"
 ]
 rustdoc-args = [ "--cfg", "doc_cfg" ]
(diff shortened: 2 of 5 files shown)


FabianLars:
Can we somehow remove ayatana-tray from the default features? disabling it is kinda ugly rn

lucasfernog:
And force everyone to pick one?

FabianLars:
yeah i don't know, it's tricky right now. For me on fedora i have to disable the default features and then add 4 feature flags (`"wry", "compression", "objc-exception", "gtk-tray"`) to make it work again.

As a side note we need to update the docs for that asap too (if we leave it as-is).

lucasfernog:
Disabling it is not that bad:

```toml
[dependencies]
tauri = { path = "../../../core/tauri", default-features = false, features = ["wry", "compression", "objc-exception"] }

[features]
default = [ "custom-protocol", "tauri/ayatana-tray" ]
custom-protocol = [ "tauri/custom-protocol" ]
gtk-tray = [ "tauri/gtk-tray" ]
```

lucasfernog:
I thought we agreed on having ayatana-tray as a default feature since libappindicator isn't supported on newer distro releases (debian I think).

FabianLars:
Well maybe my actual problem here is that the tray is generally enabled by default because it never was before (it was only enabled if you had a systray config in tauri.conf.json).

lucasfernog:
The tray isn't enabled by default. Can you share what's happening in your end?

FabianLars:
i get a cc linker error because ayatana is a default feature even if i don't have a systemtray configured (ayatana is also listed in the lockfile unless i disable the default features).

you ofc still can't use the actual systray apis because they are behind the system-tray feature flag.

lucasfernog:
Ahh you're right :/ I don't think we can fix it.

lucasfernog:
Unless we force the user to pick the feature.

FabianLars:
Yeah i only see 3 shitty alternatives rn:
- like you said, force the user
- somehow detect the correct one in the cli (either by checking for fedora or for the lib if that's possible)
- completely drop libappindicator 🤷 

So i guess we leave it as-is for v1 and maybe consider dropping it for v2 since we need to look into changing our build tooling for new webkitgtk versions anyway where we probably end up with custom docker containers or something...

linux really likes to give headaches

lucasfernog:
Maybe we should leave the choice to the user. Currently everyone will need to have one of the libs installed in their system, even if not using the system tray feature :/ (and we also get easier toggling between gtk/ayatana since no one will need to disable default-features).

PR #4007, opened 2022-04-29
fix(cli): powershell crashing on SIGINT, closes #3997

### LINKED ISSUE #3997: [bug] exiting `yarn tauri dev` with ctrl+c crashes the powershell terminal
### Describe the bug

This is different then the old ctrl+c bug where it rendered the terminal basically useless. This time it actually exits the terminal 😅 

It happens only with yarn if your project uses beforeDevCommand (doesn't seem to matter which manager you use inside the command tho)

It seems to be a yarn@v1 specific problem, pnpm, npm, cargo all work fine (didn't test yarn2+)
image
Note that if you use powershell directly _instead of_ inside of MS Terminal it closes the whole window 🤷 

----
cmd.exe has a similar problem, but it doesn't crash or exit. Instead it just asks if i really want to close the dev process (twice?). Again, only with yarn. All good with pnpm etc.
image
If you manage to insert a J/Y the terminal is completely fine again.

### Reproduction

1. run `yarn tauri dev`
2. close the app via ctrl+c (doing that before the app opens is enough)

### Expected behavior

_No response_

### Platform and versions

```shell
Environment
  › OS: Windows 10.0.19044 X64
  › Webview2: 100.0.1185.50
  › MSVC:
  › Node.js: 16.15.0
  › npm: 8.5.5
  › pnpm: 6.32.10
  › yarn: 1.22.18
  › rustup: 1.24.3
  › rustc: 1.60.0
  › cargo: 1.60.0

Packages
  › @tauri-apps/cli [NPM]: 1.0.0-rc.9
  › @tauri-apps/api [NPM]: 1.0.0-rc.4
  › tauri [RUST]: 1.0.0-rc.8 (no lockfile),
  › tauri-build [RUST]: no manifest (no lockfile),
  › tao [RUST]: no manifest (no lockfile),
  › wry [RUST]: no manifest (no lockfile),

App
  › build-type: bundle
  › CSP: default-src 'self'
  › distDir: ../dist
  › framework: React

App directory structure
  ├─ dist
  ├─ node_modules
  ├─ src
  └─ src-tauri
```

### Stack trace

_No response_

### Additional context

_No response_

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### Other information

### DIFF
--- a/.changes/fix-cli-powershell.md
+++ b/.changes/fix-cli-powershell.md
@@ -1,0 +1,6 @@
+---
+"cli.rs": patch
+"cli.js": patch
+---
+
+Fixes a Powershell crash when sending SIGINT to the dev command.
--- a/tooling/cli/src/dev.rs
+++ b/tooling/cli/src/dev.rs
@@ -25,7 +25,7 @@ use std::{
   fs::FileType,
   io::{BufReader, Write},
   path::{Path, PathBuf},
-  process::{exit, Command},
+  process::{exit, Command, Stdio},
   sync::{
     atomic::{AtomicBool, Ordering},
     mpsc::channel,
@@ -126,6 +126,7 @@ fn command_internal(options: Options) -> Result<()> {
           .pipe()?; // development build always includes debug information
         command
       };
+      command.stdin(Stdio::piped());
 
       let child = SharedChild::spawn(&mut command)
         .unwrap_or_else(|_| panic!("failed to run `{}`", before_dev));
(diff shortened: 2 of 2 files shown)


FabianLars (approved):
That fixed it 🥳 

while testing it i discovered that npm still asks for confirmation, but without actually breaking anything (you can press ctrl-c or Y+enter to confirm like expected) and since the confirmation is standard on windows i'd leave it as-is (i wish there would be a setting to turn that off os-wide...)

lucasfernog:
Yeah I wish we could turn it off, but we can't. We could automatically send a `Y` to stdin, but we'd also need to detect the presence of the message and the language (yikes).

PR #4032, opened 2022-05-02
fix(build): append .exe binary based on target triple instead of running OS, closes #3870

### LINKED ISSUE #3870: [feat] Use configured target triple for filename
### Describe the problem

For now, when compiling a project, the final filename for the binary depends on the OS, as stated in the `build.rs` file here for instance:

https://github.com/tauri-apps/tauri/blob/82c7855bf83326eafcad861642edd117684fa506/tooling/cli/src/build.rs#L211-L225

I could successfully manage to cross-compile the project by using [cross-rs/cross](https://github.com/cross-rs/cross) as runner instead of Cargo, with this command:

```
$ cargo-tauri build --target=x86_64-pc-windows-gnu --runner=$HOME/.cargo/bin/cross
```

(executed in Ubuntu 20.04 in WSL 2 on a Windows 10 machine)

The binary is built, I can even run it (but only if I copy `WebView2Loader.dll` next to the `.exe` file, which is another subject).

However, the build command still fails like this:

```
❯ cargo-tauri build --target=x86_64-pc-windows-gnu --runner=$HOME/.cargo/bin/cross                                                                                                                                                       16:26:15
   Compiling my-project-name v0.1.0 (/project)
    Finished release [optimized] target(s) in 46.73s
Error: failed to rename `/var/www/my-project/src-tauri/target/x86_64-pc-windows-gnu/release/my-project-name` to `/var/www/my-project/src-tauri/target/x86_64-pc-windows-gnu/release/my-project-name`

Caused by:
    No such file or directory (os error 2)
```

The reason is that the `my-project-name` file doesn't exist: it was built for Windows, so it is instead named `my-project-name.exe`.

The solution to make it work seems (according to the code I read in the tauri build process) that the `build.rs` code linked above should change behavior:

Instead of using something like this to determine the file name:
```
    #[cfg(windows)]
    let product_path = out_dir.join(format!("{}.exe", product_name));
```

Tauri should instead rely on determining whether it has to add `.exe` (or if it has to convert it to "kebab case" as of the linux-specific code we can see later in the file) depending on the **target triple** rather than what OS is building the project.

I don't really know if this is feasible, especially considering the amount of different target triples (there's a [list of target triples](https://github.com/cross-rs/cross#supported-targets) in the Cross repository), but I guess it might benefit future cross-compilation-based features.

### Alternatives considered

For now, the only alternative is to "just ignore" the error with shell features, which is not optimal since one has to add other shell commands to see if the project has _actually_ been successfully built or not.

### Additional context

_No response_

Pierstoval:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

Yes, a minor one I hope:

`cli::interface::rust::get_binaries()` gets a new `target: Option<String>` argument, which is the target triple you want to build your application for.

### Checklist

### Other information

The side-effect of this is that it enables using [cross](https://github.com/cross-rs/cross) as build runner to attempt cross-compiling projects.

**Request for suggestions:**

As the target is currently optional because it should fall back to the current target triple, I need some help knowing how to get the current target triple (in a cleaner way than [what I currently found](https://codeutility.org/how-can-i-find-the-current-rust-compilers-default-llvm-target-triple-stack-overflow/)) so that we can no longer depend on an optional value.

I thought we could use these build-in constants ARCH, OS and FAMILY (from there: https://doc.rust-lang.org/std/env/consts/index.html ), but not sure of their reliability regarding Tauri's ecosystem.

### DIFF
--- a/.changes/cli-bin-ext-from-target-triple.md
+++ b/.changes/cli-bin-ext-from-target-triple.md
@@ -1,0 +1,6 @@
+---
+"cli.rs": patch
+"cli.js": patch
+---
+
+Resolve binary file extension from target triple instead of compile-time checks to allow cross compilation.
--- a/core/tauri-utils/src/resources.rs
+++ b/core/tauri-utils/src/resources.rs
@@ -29,7 +29,11 @@ pub fn external_binaries(external_binaries: &[String], target_triple: &str) -> V
       "{}-{}{}",
       curr_path,
       target_triple,
-      if target_triple.contains("windows") { ".exe" } else { "" }
+      if target_triple.contains("windows") {
+        ".exe"
+      } else {
+        ""
+      }
     ));
   }
   paths
--- a/core/tauri-utils/src/resources.rs
+++ b/core/tauri-utils/src/resources.rs
@@ -29,7 +29,7 @@ pub fn external_binaries(external_binaries: &[String], target_triple: &str) -> V
       "{}-{}{}",
       curr_path,
       target_triple,
-      if cfg!(windows) { ".exe" } else { "" }
+      if target_triple.contains("windows") { ".exe" } else { "" }
     ));
   }
   paths
--- a/tooling/bundler/src/bundle/settings.rs
+++ b/tooling/bundler/src/bundle/settings.rs
@@ -322,11 +322,7 @@ impl BundleBinary {
   /// Creates a new bundle binary.
   pub fn new(name: String, main: bool) -> Self {
     Self {
-      name: if cfg!(windows) {
-        format!("{}.exe", name)
-      } else {
-        name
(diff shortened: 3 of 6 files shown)


lucasfernog:
We already have a function to get the current target triple :) awesome PR!

Pierstoval:
Thanks a lot for all the fixes @lucasfernog !

lucasfernog (approved):
Thanks :heart:

Pierstoval:
Thanks for merging 🚀

PR #4055, opened 2022-05-04
Fix binary extension duplicate "." char

Pierstoval:
Followup of #4032 which introduced a slight bug as stated in https://github.com/tauri-apps/tauri/issues/3870#issuecomment-1117075228


lucasfernog (approved):
Nice catch!

PR #4060, opened 2022-05-04
feat(cli): Improve CLI logging

JonasKruckenberg:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

This PR overhauls the CLI's output, completely moving it from a custom logging solution to the standard `log` and `env_logger` crates. Towards this goal I removed **all** printing functionality from the bundler, cleaning up concerns and bringing the bundler one step closer to being a reusable library.

### Major new feature

- The CLI now offers *way* more debug output, logging parameters and outputs of each executed command and shell script plus more. 
- The produced output now look much closer to cargo's output (and IMO better than before)
      For this we leverage a custom KV attribute for log macros: `action` that will take the log levels position in the output. `info!(action = "Bundling"; "{}", app_product_name);` produces `Bundling <App Name>`
- The log level can be controlled via the `--verbose` flag and the `RUST_LOG` env var. (The flag can appear multiple times, where `--verbose` means `Debug` and `--verbose --verbose` means `Trace`)


JonasKruckenberg:
I think this doesn't compile on linux or windows 🤔 but my rust-analyzer seems to be broken rn

tooling/bundler/src/bundle/settings.rs
```diff
@@ -386,7 +384,6 @@ pub struct Settings {
 #[derive(Default)]
 pub struct SettingsBuilder {
   project_out_directory: Option<PathBuf>,
-  verbose: bool,
```
lucasfernog, line 389: we should mark this change as a breaking change in the change file, and show the solution (i know there's some people using the tauri bundler as a library)

lucasfernog (changes requested):
top5 best Tauri PRs easily

amrbashir:
Bug alert, on Windows, if you have `before(Build|Dev)Command`, it prints the `before(Build|Dev)Command` output then hangs indefinitely and if `before(Build|Dev)Command` clears the terminal like how vite devServer does by default, it will only print `Running BeforeDevCommand (`pnpm dev`)`.

tooling/bundler/src/bundle/settings.rs
JonasKruckenberg: True! Good catch

JonasKruckenberg:
> Bug alert, on Windows, if you have `before(Build|Dev)Command`, it prints the `before(Build|Dev)Command` output then hangs indefinitely and if `before(Build|Dev)Command` clears the terminal like how vite devServer does by default, it will only print `Running BeforeDevCommand (`pnpm dev`)`. 

Is that using the latest commit from @lucasfernog? Maybe it has to do with the piping? I had a bunch of issues while developing with disappearing output too.

Maybe not though, since it's the same implementation as before basically.

@lucasfernog do we need `os_pipe` in the first place? We could just use the standard `Stdio::inherit()` if don't need the output.

lucasfernog:
We do need the os_pipe @JonasKruckenberg otherwise the binding just won't show the output.

lucasfernog:
@amrbashir that was fixed in my commits.

amrbashir:
Oh then I might've checked with an older commit. I tested the branch twice and I missed a git pull in between.

Edit: just tested latest commits and it works.

lucasfernog (approved):

JonasKruckenberg:
Niiice 🥳

PR #4061, opened 2022-05-04
feat(core): configure msiexec display options, closes #3951

### LINKED ISSUE #3951: [feat] Silent (or quieter) Windows updater
### Describe the problem

The Windows updater currently just re-runs the installer at update time, which gets the job done, but the UX could be better if it was more hands-off.

The MSI installers support a few flags to make the installation less interactive. There's a [fully silent mode](https://stackoverflow.com/questions/8560166/silent-installation-of-a-msi-package) where it shows nothing onscreen, and a [passive mode](https://stackoverflow.com/questions/33484750/msiexec-difference-between-quiet-and-passive) where it only shows a progress bar. Chatting with Fabian in Discord it seems like the "passive" mode might be a good middle ground - an automatic installation, but with some feedback so the user knows what's going on.

### Alternatives considered

_No response_

### Additional context

I'd be happy to take a stab at implementing it - just grepping around for `msiexec` I see it's referenced in `core/tauri/src/updater/core.rs` and also `tauri/tooling/bundler/src/bundle/windows/templates/update-task.xml`. I think this would only require changing the `/qb+` to `/passive` in those two spots.

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


FabianLars:
1) i'm pretty sure /qn and /quiet is the same thing (with /quiet being "newer" iirc)
2) i tend to vote for using passive as the default option so users see some feedback and know that the update is actually executing.

core/tauri-utils/src/config.rs
```diff
@@ -1872,6 +1872,94 @@ impl<'de> Deserialize<'de> for UpdaterEndpoint {
...
+  Passive,
+}
+
+impl WindowsUpdateInstallMode {
+  /// Returns the associated `msiexec.exe` arguments.
+  pub fn msiexec_args(&self) -> &'static [&'static str] {
+    match self {
+      Self::BasicUi => &["/qb+"],
+      Self::NoUi => &["/qn"],
+      Self::Quiet => &["/quiet"],
```
FabianLars: I'm like 95% sure that /qn and /quiet is the same. i think /quiet is just "newer".

core/tauri-utils/src/config.rs
```diff
@@ -1872,6 +1872,94 @@ impl<'de> Deserialize<'de> for UpdaterEndpoint {
...
+  }
+}
+
+/// The updater configuration for Windows.
+#[skip_serializing_none]
+#[derive(Debug, Default, PartialEq, Clone, Serialize, Deserialize)]
+#[cfg_attr(feature = "schema", derive(JsonSchema))]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct UpdaterWindowsConfig {
+  /// The installation mode for the update on Windows. Defaults to `quiet`.
```
FabianLars:
```suggestion
  /// The installation mode for the update on Windows. Defaults to `passive`.
```

core/tauri/src/updater/core.rs
```diff
@@ -757,8 +771,8 @@ fn copy_files_and_run<R: Read + Seek>(
       Command::new("msiexec.exe")
         .arg("/i")
         .arg(found_path)
-        // quiet basic UI with prompt at the end
-        .arg("/qb+")
+        .args(msiexec_args)
+        .arg("/promptrestart")
```
FabianLars, line 775: Why do we want to prompt for a system restart here? Can't think of a reason to do this in the context of installing/updating _tauri_ apps.

tooling/bundler/src/bundle/windows/templates/update-task.xml
```diff
@@ -37,7 +37,7 @@
   <Actions Context="Author">
     <Exec>
       <Command>cmd.exe</Command>
-      <Arguments>/c "msiexec.exe /i %TEMP%\\{{{product_name}}}.msi /qb+"</Arguments>
+      <Arguments>/c "msiexec.exe /i %TEMP%\\{{{product_name}}}.msi {{{msiexec_args}}} /promptrestart"</Arguments>
```
FabianLars, line 40: Same as above.

tooling/cli/schema.json
```diff
@@ -1792,6 +1809,22 @@
       "type": "string",
       "format": "uri"
     },
+    "UpdaterWindowsConfig": {
+      "description": "The updater configuration for Windows.",
+      "type": "object",
+      "properties": {
+        "installMode": {
+          "description": "The installation mode for the update on Windows. Defaults to `quiet`.",
```
FabianLars:
```suggestion
          "description": "The installation mode for the update on Windows. Defaults to `passive`.",
```

FabianLars (changes requested):

core/tauri/src/updater/core.rs
lucasfernog: when using /passive the msiexec was sometimes rebooting the system automatically. `/promptrestart` sets msiexec to prompt the user for a restart if it is needed.

core/tauri-utils/src/config.rs
lucasfernog: you're right, gonna remove it

core/tauri/src/updater/core.rs
FabianLars: ah alrighty, the parameter sounded like it would always prompt for restart even if the app wouldn't need it 👍

FabianLars (approved):

PR #4063, opened 2022-05-05
refactor(example): simplify streaming example by using asset protocol

keiya01:

- [ ] Bugfix
- [ ] Feature
- [x] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/examples/streaming/README.md
+++ b/examples/streaming/README.md
@@ -2,4 +2,4 @@
 
 A simple Tauri Appplication showcase the streaming functionality.
 
-To execute run the following on the root directory of the repository: `cargo run --example streaming`.
+To execute run the following on the root directory of the repository: `cargo run --example streaming --features protocol-asset`.
--- a/examples/streaming/index.html
+++ b/examples/streaming/index.html
@@ -22,7 +22,7 @@
     const video = document.getElementById('video_source')
     const source = document.createElement('source')
     source.type = 'video/mp4'
-    source.src = convertFileSrc('example/test_video.mp4', 'stream')
+    source.src = convertFileSrc('test_video.mp4')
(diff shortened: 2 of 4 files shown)


amrbashir:
The streaming example is an example of sending only the bytes requested in the range header while the `asset` protocol will send the whole file, so you can't simply replace streaming with `asset` protocol.

keiya01:
I think asset protocol has feature to use range header too :eyes:
https://github.com/tauri-apps/tauri/blob/c82b4761e1660592472dc55308ad69d9efc5855b/core/tauri/src/manager.rs#L545

amrbashir:
Oh, good to know that, thanks.

Still, since it is a still an example about streaming, it should show a manual implementation that can be a reference to others who want to implement their own.

keiya01:
Maybe we can also export function to process the range header.
What do you think about it?

amrbashir:
you should ask @lucasfernog

keiya01:
@lucasfernog 
I think we can export a function to process the range header.
What do you think about it?
This is because currently we have some large boilerplate to process the range header.

lucasfernog:
We can't really expose a range parser because we didn't even implement multiple ranges yet.

FabianLars:
I actually interpreted this example as an example for the `register_uri_scheme_protocol` function with streaming just being a good way to showcase it.
I somewhat regularly forward people to the example (for the custom protocol), so i'd rather not remove it like this.

keiya01:
I understood.
Thank you for your reply.

PR #4094, opened 2022-05-10
feat(tauri-runtime-wry): add plugin API

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/.changes/tauri-runtime-wry-plugin.md
+++ b/.changes/tauri-runtime-wry-plugin.md
@@ -1,0 +1,5 @@
+---
+"tauri-runtime-wry": patch
+---
+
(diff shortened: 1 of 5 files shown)


JonasKruckenberg:
I think it'd make more sense to make this a callback on the `Plugin` trait instead of a completely separate thing wdyt?

You could call it `on_wry_event` or something 🤔

Edit: Or rather, we can have the `TauriPlugin` struct that is returned by the builder implement `tauri_runtime_wry::Plugin` too, that way we can just pass the whole plugin through to wry

PR #4153, opened 2022-05-17
chore(config) Fix config links

lorenzolewis:

- [ ] Bugfix
- [ ] Feature
- [x] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

- Fixes links in the config object for the generated json schema
- Rearranges the order of the description to match the Config object
- Remove comment in example JSON (fixes it not being valid JSON)
- Add `preserve_order` feature to `schemars` crate to generate schema in reference order instead of alphabetical order


lorenzolewis:
@FabianLars this should be the fix for the links

core/tauri-utils/src/config.rs
```diff
@@ -2248,10 +2248,10 @@ impl PackageConfig {
 ///
 /// `tauri.conf.json` is composed of the following objects:
 ///
-/// - [`build`](BuildConfig): The build configuration
-/// - [`package`](PackageConfig): Package settings
-/// - [`plugins`](PluginConfig): The plugins config
-/// - [`tauri`](TauriConfig): The Tauri config
```
FabianLars, line 2254: The problem with this is that it breaks the links on docs.rs. Idk if we care about this in this scenario tho

lorenzolewis: Hmm the option on the docusaurus side is to just log but not error on broken markdown links, but I wasn't a fan of that option. 

Do you think we even need the links since the object (with generated links) will be just below?

FabianLars: > Do you think we even need the links since the object (with generated links) will be just below?

If being just below it is enough, we may leave the pr as-is, because the link on docs.rs does nothing (just a fragment) instead of navigation to a 404 for example. And i'd much rather have a "inactive" link in tauri-utils' docs.rs page than on tauri.studio :D

jeremycorbeil9 (dismissed):

lucasfernog (approved):

Issue #4161, opened 2022-05-18
[bug] `UpdateBuilder::target` weird behavior when specifying target


JonasKruckenberg:
### Describe the bug

These are the lines in question:
https://github.com/tauri-apps/tauri/blob/393c774dfd0d523bf10ceef2a03393640bf0f44d/core/tauri/src/updater/core.rs#L330-L340

Aside from being weird and hard to understand the section has a logic issue, that would potentially prevent the updater from finding the correct platform entry.

### Reproduction

Let's make an example to illustrate this:

I'm setting `UpdateBuilder::target` to `darwin-x86_64` even though my machines arch is `aarch64`.
This will cause the updater to fill out the url `https://example.com/{{target}}/{{arch}}` like so `https://example.com/darwin-x86_64/aarch64` which is absolutely not what I intended. 

It gets weirder though:
Let's also say I'm using the url parameters only for statistics and always return the `static JSON format`. 
In this situation the updater would download the `x86_64` artifact (like intended) **even though** it clearly announced to the update server that it's arch is `aarch64`.

### Expected behavior

given the target `darwin-x86_64`
`{{target}}` would get replaced with `darwin` and `{{arch}}` would be replaced by `x86_64`.

### Platform and versions

```shell
N/A
```

### Stack trace

```shell
N/A
```

lucasfernog:
Tbh I think if you use the custom target option, you don't need the `{{arch}}`. Also I couldn't reproduce the issue where the updater looks for the wrong value in the JSON, since if a custom target is set we always use it:
```rust
let json_target = if has_custom_target { 
   target.clone() 
} ...
```

PR #4162, opened 2022-05-18
refactor(updater): replace manual parsing with struct definitions

JonasKruckenberg:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

There a number of chances to further tighten down the interface, but they would introduce a breaking change (something @nothingismagick explicitly wanted to avoid)

- **Strongly type `target`**. Currently `target` is a String. By creating a custom type `UpdateTarget { os, arch }` we could constrain the target to valid target combinations only and outright reject unsupported/malformed targets. This would also help to solve #4161.
- **Require `version` field to be strict semver.** Currently we accept both `v1.0.0` and `1.0.0` but only the latter is valid semver (the `semver` crate will reject the former). Requiring strict semver would allow us to get rid of the custom deserialize fn. This would also have the advantage that the `should_update` callback would receive proper `Version` structs instead of `&str` and could perform comparisons much easier (and reliably).
- **Remove name alias**. Currently the version field can be called `version` or `name`. We never have `name` in any docs and it's rather obscure. Getting rid of it would simplify the interface.
- **Add chrono type for `pub_date`**. Currently `pub_date` is a String, and we pass it through as-is. [chrono](https://docs.rs/chrono/latest/chrono/index.html)::[DateTime](https://github.com/tauri-apps/tauri/compare/refactor/updater-json?expand=1) is ISO 8601 and implements `Deserialize`. This would make updates more secure as malformed dates would be rejected during deseralization as soon as possible.

### DIFF
--- a/core/tauri/Cargo.toml
+++ b/core/tauri/Cargo.toml
@@ -60,7 +60,7 @@ tauri-macros = { version = "1.0.0-rc.6", path = "../tauri-macros" }
 tauri-utils = { version = "1.0.0-rc.6", features = [ "resources" ], path = "../tauri-utils" }
 tauri-runtime-wry = { version = "0.5.1", path = "../tauri-runtime-wry", optional = true }
 rand = "0.8"
-semver = { version = "1.0", features = ["serde"]}
+semver = { version = "1.0", features = [ "serde" ] }
 serde_repr = "0.1"
 state = "0.5"
 tar = "0.4.36"
--- a/core/tauri/Cargo.toml
+++ b/core/tauri/Cargo.toml
@@ -60,7 +60,7 @@ tauri-macros = { version = "1.0.0-rc.6", path = "../tauri-macros" }
 tauri-utils = { version = "1.0.0-rc.6", features = [ "resources" ], path = "../tauri-utils" }
 tauri-runtime-wry = { version = "0.5.1", path = "../tauri-runtime-wry", optional = true }
 rand = "0.8"
-semver = "1.0"
+semver = { version = "1.0", features = ["serde"]}
 serde_repr = "0.1"
 state = "0.5"
 tar = "0.4.36"
--- a/core/tauri/src/updater/core.rs
+++ b/core/tauri/src/updater/core.rs
@@ -175,13 +175,13 @@ impl RemoteRelease {
 
   #[cfg(target_os = "windows")]
   /// Optional: Windows only try to use elevated task
-  pub fn with_elevated_task(&self) -> bool {
-    match self.inner {
-      RemoteReleaseInner::Dynamic(ref platform) => Ok(&platform.with_elevated_task),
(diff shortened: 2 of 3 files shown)


lucasfernog:
@JonasKruckenberg I know you'll think the deserialize impl is too ugly, but it's needed to have better error messages. `#[serde(flatten)]` just gives us useless errors (`data did not match any variant of untagged enum RemoteReleaseInner`).

JonasKruckenberg:
> @JonasKruckenberg I know you'll think the deserialize impl is too ugly, but it's needed to have better error messages. `#[serde(flatten)]` just gives us useless errors (`data did not match any variant of untagged enum RemoteReleaseInner`).

Hmmm, you're right, that is very ugly... 😂 Can't we use something like https://github.com/dtolnay/path-to-error and simply convert the serde_json errors into something nicer, without weird whacky struct workarounds?

lucasfernog:
> Hmmm, you're right, that is very ugly... joy Can't we use something like [dtolnay/path-to-error](https://github.com/dtolnay/path-to-error) and simply convert the serde_json errors into something nicer, without weird whacky struct workarounds?

We can't, serde doesn't give us any information with the flatten attribute, so the path_to_error crate just returns `.` for the error path.

JonasKruckenberg:
Stupid flatten 🙄 another idea, we just duplicate `version`, `pub_date` and `notes` fields and have `RemoteRelease` be an enum?

lucasfernog:
We also need to duplicate the ReleaseManifestPlatform fields right? url, signature and with_elevated_task since that's where the flatten is used.

JonasKruckenberg:
Hmmm the _Dynamic_ variant would either have to duplicate or use use `flatten` yeah 🤔

But would flatten in that context still be a problem?

lucasfernog:
I think flatten is always a problem for the error message.

tweidinger (dismissed):
Looks good to me. I don't see breaking or security impact changes. The PR mentions additionally (breaking) changes which sound like good improvements. Especially the chrono type for the date.

lucasfernog:
@JonasKruckenberg I tried the enum change, but this is what serde_json gives me: `SerdeJson(Error("data did not match any variant of untagged enum RemoteRelease", line: 0, column: 0))`. Seems like the custom deserializer is the only choice :D

JonasKruckenberg:
> @JonasKruckenberg I tried the enum change, but this is what serde_json gives me: `SerdeJson(Error("data did not match any variant of untagged enum RemoteRelease", line: 0, column: 0))`. Seems like the custom deserializer is the only choice :D

It very much looks like it yeah :/

lucasfernog (dismissed):

PR #4192, opened 2022-05-22
refactor: cleanup logic to get path from protocol req url, closes #4132

### LINKED ISSUE #4132: [bug] `asset` protocol will replace `asset://` too many times
### Describe the bug

I noticed that this code:
https://github.com/tauri-apps/tauri/blob/3e702cf8b15762cdca43c8d7ff6f6e8ee9670244/core/tauri/src/manager.rs#L520-L523=

(EDIT: Don't know why it's not showing embedded code for that URL, nevertheless here it is:)#
```rust
        #[cfg(target_os = "windows")]
        let path = filtered_path.replace("asset://localhost/", "");
        #[cfg(not(target_os = "windows"))]
        let path = filtered_path.replace("asset://", "");
```

will consume extra occurrences of `asset://` in the URL. (See https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=52f3c5bdb3655430b0493537fe762ec4)

```rust
use url::Url;

fn main() {
    let parsed_path = Url::parse("asset://blahblah/asset://123").unwrap();
    println!("{:?}", &parsed_path[..].replace("asset://", ""));
}
```

I don't think this is major but it may be worth being a little bit more paranoid in case this could interact badly with some application code and lead to an exploit.

Sorry if this issue is invalid because it's somehow being escaped by this point, but I thought it would be better safe than sorry.

### Reproduction

_No response_

### Expected behavior

_No response_

### Platform and versions

```shell
This is based on an examination of the source code only.
```

### Stack trace

_No response_

### Additional context

_No response_

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/core/tauri/src/manager.rs
+++ b/core/tauri/src/manager.rs
@@ -504,9 +504,9 @@ impl<R: Runtime> WindowManager<R> {
         let parsed_path = Url::parse(request.uri())?;
         let filtered_path = &parsed_path[..Position::AfterPath];
         #[cfg(target_os = "windows")]
-        let path = filtered_path.replace("asset://localhost/", "");
+        let path = filtered_path.trim_start_matches("asset://localhost/");
         #[cfg(not(target_os = "windows"))]
-        let path = filtered_path.replace("asset://", "");
+        let path = filtered_path.trim_start_matches("asset://");
         let path = percent_encoding::percent_decode(path.as_bytes())
           .decode_utf8_lossy()
           .to_string();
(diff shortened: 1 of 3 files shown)


core/tauri/src/manager.rs
```diff
@@ -504,9 +504,9 @@ impl<R: Runtime> WindowManager<R> {
         let parsed_path = Url::parse(request.uri())?;
         let filtered_path = &parsed_path[..Position::AfterPath];
         #[cfg(target_os = "windows")]
-        let path = filtered_path.replace("asset://localhost/", "");
+        let path = filtered_path.trim_start_matches("asset://localhost/");
```
reivilibre, line 507: `trim_start_matches` will remove the match multiple times; is that really desirable?
I think [`strip_prefix`](https://doc.rust-lang.org/std/primitive.str.html#method.strip_prefix) is a little more appropriate here.

Is there any reason why the path wouldn't start with that prefix at this point?

lucasfernog: You're right lol

PR #4246, opened 2022-05-31
fix: benchmark build error

Beanow:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [x] Other, please describe:

A bugfix, not in the Tauri codebase but benchmark tooling.

- [x] No

### Checklist

### Other information

The specifics are a bit over my head here, but my understanding is:

We're using compiler optimizations for the benchmarks here
https://github.com/tauri-apps/tauri/blob/0c635959162dd9e2ef70897f7792034db7e66435/tooling/bench/tests/Cargo.toml#L6-L11

The `panic=abort` one prompts us to compile the std as well https://github.com/tauri-apps/tauri/blob/0c635959162dd9e2ef70897f7792034db7e66435/.github/workflows/bench.yml#L106

And it seems there's a known issue with `-Z build-std` and the LTO we've enabled:
https://github.com/rust-lang/rust/issues/96486

The workaround for which seems to be copying a profile from `rust-lang/rust` to our own `Cargo.toml`:
https://github.com/rust-lang/cargo/issues/10118#issuecomment-1006000210

So that's what I've done here, and it seems to work in a local Docker container test as well as GH actions on my fork:
https://github.com/Beanow/tauri/runs/6671511937?check_suite_focus=true

### DIFF
--- a/tooling/bench/tests/Cargo.toml
+++ b/tooling/bench/tests/Cargo.toml
@@ -9,3 +9,27 @@ codegen-units = 1
 lto = true
 incremental = false
 opt-level = "s"
+
+# See https://github.com/rust-lang/cargo/issues/10118#issuecomment-1006000210
+# And https://github.com/rust-lang/rust/issues/96486
+[profile.release.package.compiler_builtins]
+# The compiler-builtins crate cannot reference libcore, and it's own CI will
+# verify that this is the case. This requires, however, that the crate is built
+# without overflow checks and debug assertions. Forcefully disable debug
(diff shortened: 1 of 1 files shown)


lucasfernog (approved):
Wow I was trying to fix it but couldn't find this patch, thank you!

lemarier:
Bravo!

Issue #3859, opened 2022-04-05
[feat] Auto-detect appindicator library during runtime


wusyong:
### Describe the problem

Right now there's two libraries being used for system tray on Linux: libappindicator and libayatana-appindicator. ayatana is former's fork since the old one is unmaintained. This becomes an issue because different distros might use different one. I would say the ratio is around 50:50.
Decide one during compile time means the other half won't work.

### Describe the solution you'd like

Add a feature or default to auto detect the library. We can use pkg_config crate to probe if the library is exist on the end user device.

peterroe:
Hello, We need more information about your system, Maybe you can run `npm run tauri info ` or `cargo tauri info` to get it.

PR #4247, opened 2022-05-31
refactor(core): drop `ayatana-tray` and `gtk-tray` Cargo features

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


FabianLars:
- Since we're talking about .deb here (and therefore mainly about debian & ubuntu) we should probably use ayatana as the default.
- iirc, we now depend on `pkg-config` _at runtime_, so we need to add that to the Depends field too right?

lucasfernog:
Yeah you're right, I'll make that change.

lucasfernog:
Ref #3859

PR #4254, opened 2022-06-02
Fix: sidecar cleanup when only `shell-sidecar` is enabled

busyboredom:

- [ x] Bugfix

- [ x] No

### Checklist

### Other information

Sidecar processes are not cleaned up properly unless `shell-execute` is enabled. This is problematic for apps in which only `shell-sidecar` is needed.

This fix calls `crate::api::process::kill_children()` when either `shell-execute` or `shell-sidecar` are enabled.

### DIFF
--- a/.changes/fix-sidecar-cleanup.md
+++ b/.changes/fix-sidecar-cleanup.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch
+---
+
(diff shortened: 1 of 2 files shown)


busyboredom:
I didn't make an issue for this 'cause it's such a simple problem/solution, but if you need me to open an issue before this can be merged I will :)

lucasfernog (approved):
Nice catch, thanks!

PR #4277, opened 2022-06-05
refactor: simplify api.js bundling

amrbashir:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


lucasfernog:
We need to test this on Webpack and CRA. https://github.com/tauri-apps/tauri/issues/2360#issuecomment-903894967

amrbashir:
The `target` bump in `tsconfig` is not really important and can be reverted, will test CRA anyways.

amrbashir:
CRA template created with `create-react-app` works fine.

tooling/api/tsup.config.ts
```diff
@@ -0,0 +1,30 @@
...
+    globalName: '__TAURI_IIFE__',
+    clean: false,
+    minify: true,
+    platform: 'browser',
+    dts: false,
+    // esbuild `globalName` option generates `var __TAURI_IIFE__ = (() => {})()`
+    // but we want to assign it to the `window` object, this footer takes care of that
+    footer: {
+      js: 'window.__TAURI__ = __TAURI_IIFE__'
+    }
```
imranbarbhuiya:
```suggestion
    globalName: '__TAURI__',
    clean: false,
    minify: true,
    platform: 'browser',
    dts: false,
```

global functions automatically gets assigned to `window` and no need to do it again.

**Edit:** Attaching image of a test
image

imranbarbhuiya (changes requested):

tooling/api/tsup.config.ts
amrbashir: from MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var
> Note that in both NodeJS [CommonJS](https://www.commonjs.org/) modules and native [ECMAScript modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules), top-level variable declarations are scoped to the module, and are not, therefore added as properties to the global object.

So there is cases where `var` doesn't assign to the global `window` object and on Windows we are using  webview2 [`AddScriptToExecuteOnDocumentCreated`](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2?view=webview2-1.0.1264.42#addscripttoexecuteondocumentcreated) and it seems it doesn't assign `var` to `window` so I make sure to assign it manually.

amrbashir: I will update the comment though

mrquantumoff (dismissed):
Looks nice!

tooling/api/src/index.ts
```diff
@@ -13,4 +13,42 @@
...
+import * as globalShortcut from './globalShortcut'
+import * as http from './http'
+import * as notification from './notification'
+import * as path from './path'
+import * as process from './process'
+import * as shell from './shell'
+import * as tauri from './tauri'
+import * as updater from './updater'
+import * as window from './window'
+import * as os from './os'
```
lorenzolewis, line 31: This bothers me because it's the only non-alphabetised one on the list ;)

amrbashir: I will fix that once we are ready to merge this PR

tooling/api/package.json
```diff
@@ -36,19 +36,11 @@
     "access": "public"
   },
   "engines": {
-    "node": ">= 12.22.0",
+    "node": ">= 14.6.0",
```
lucasfernog, line 39: do we really need to bump this? if so we need to add it to the changelog

amrbashir: node 12 has reached end of life 5 months ago

lucasfernog (approved):
Let's keep an eye to see if this doen't break any app, I've tested it on Quasar and CRA

PR #4283, opened 2022-06-07
fix: add a default icon to notifications on linux

betamos:
Currently, the notification api allows an optional icon. This has no effect on Windows and MacOS, where the icon is always taken from the application. This change has no effect on those platforms, but sets a default icon based on the binary path, which ensures notifications, by default, uses the app's icon.

This icon is also shown in dev mode if the app is installed. Otherwise, it just displays no icon.

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/.changes/linux-notification-icon.md
+++ b/.changes/linux-notification-icon.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch
+---
+
+Set notification icon to app icon on Linux.
--- a/core/tauri/src/api/notification.rs
+++ b/core/tauri/src/api/notification.rs
@@ -83,6 +83,8 @@ impl Notification {
     }
     if let Some(icon) = self.icon {
       notification.icon(&icon);
+    } else {
+      notification.auto_icon();
     }
     #[cfg(windows)]
     {


core/tauri/src/endpoints/notification.rs
```diff
@@ -61,7 +61,14 @@ impl Cmd {
     if let Some(body) = options.body {
       notification = notification.body(body);
     }
-    if let Some(icon) = options.icon {
+
+    // Icon is only effective on linux. Here we set the fallback icon to the binary name, which is
+    // the name used by the linux bundler.
+    let icon = options
```
lucasfernog: we should probably do this only on Linux with a `#[cfg]` guard.

lucasfernog (changes requested):
Please add a change file, check out the `.changes` folder to see how to do it. Basically:

```markdown
---
"tauri": patch
---

Set notification icon to app icon on Linux.
```

core/tauri/src/endpoints/notification.rs
betamos: Resolved by using `auto_icon` instead.

lucasfernog (approved):

PR #4285, opened 2022-06-07
fix(test): replace test_binary with a real binary file

meowtec:
File `test_binary` is a text file, not really a binary.
When clone a Git repository' to a windows machine, `LF`s are NOT always transformed to `CRLF`. That depends on the git config. Thus `cargo test` will fail on some developer's windows machine.

So should replace it with a real binary file (1px x 1px GIF file).

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe:

- [x] No

### Checklist


lucasfernog (approved):
Nice catch, thank you!

PR #4297, opened 2022-06-08
Append binary extension to filename based on target in Settings

Pierstoval:
This is a followup of #4032 and #4055, for the issue #3870.

I'm still attempting to cross-compile using `cross`, and I don't know if this specific change will break compat, so I'm creating this PR so that CI tests it on several other tools :)

### DIFF
--- a/core/tauri-runtime-wry/src/lib.rs
+++ b/core/tauri-runtime-wry/src/lib.rs
@@ -2434,9 +2434,9 @@ fn handle_user_message<T: UserEvent>(
           .get(&id)
           .and_then(|w| w.inner.as_ref())
         {
-          if let Err(e) = webview.evaluate_script(&script) {
+          if let Err(_e) = webview.evaluate_script(&script) {
             #[cfg(debug_assertions)]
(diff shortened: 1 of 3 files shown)


lucasfernog:
@Pierstoval the binary extension is already part of the `name`: https://github.com/tauri-apps/tauri/blob/55f89d5f9d429252ad3fd557b1d6233b256495e0/tooling/cli/src/interface/rust.rs#L195

Pierstoval:
> @Pierstoval the binary extension is already part of the `name`:

What's really strange is that it doesn't seem to be propagated 🤔

I will look into this to check what's going on.

Edit: I rebased, re-tested locally, and my build still works! May I have found a bug somewhere? 🤔

lucasfernog:
@Pierstoval can you share a repo of the app that has build issues without this fix?

lucasfernog:
If there's something wrong we need to find a different solution since this breaks the build for normal apps (including the updater CI test here).

lucasfernog:
Please file an issue with reproduction steps for the issue you're facing :)

PR #4301, opened 2022-06-08
feat(bundler): Add Pacman package support, closes #3728

### LINKED ISSUE #3728: [feat] Arch support for tauri bundler
### Describe the problem

Currently, tauri bundler can't make Arch packages

I would like to add support for Arch packages

### Alternatives considered

_No response_

### Additional context

I'd love to add this myself, but the contributing guide says to open an issue first

oowl:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


tooling/bundler/src/bundle/linux/pacman.rs
```diff
@@ -0,0 +1,241 @@
...
+  writeln!(file, "Icon={}", bin_name)?;
+  writeln!(file, "Name={}", settings.product_name())?;
+  writeln!(file, "Terminal=false")?;
+  writeln!(file, "Type=Application")?;
+  Ok(())
+}
+
+
+/// Generate the icon files and store them under the `data_dir`.
+fn generate_icon_files(settings: &Settings, package_dir: &Path) -> crate::Result<BTreeSet<LinuxIcon>> {
```
betamos: This looks identical to the debian version (perhaps other functions too?). Could you break out any shared code, perhaps into a lib file in the parent (linux) module?

(Note https://github.com/tauri-apps/tauri/pull/4282 which changes this logic)

FabianLars: generally i have to agree here, but we can worry about that later too if you're not comfortable with making these changes (once i get to the rpm implementation i'll probably end up in a similar situation).
On that note this PR won't be merged until after v1 dropped, just so that we're on the same place ✌️

lucasfernog:
Awesome PR as an Arch BTW user! Can't wait to take a look at it after v1 lands.

oowl:
Hello, Anything I need to do for this pr?

lucasfernog:
I didn't check it yet @attenuation we need to sort out how to include this - might be merged to a `next` branch.

RubenKelevra:
Just a archlinux package maintainer chiming in here:

Yes, ArchLinux does have binary packages, but the amount of users who contributes them is very small. There are much more maintainer who are not "trusted" and thus only providing building instructions to the AUR.

This PR covers only using an local tauri app and create a binary for it and ship it as packaged binary.

It would actually be much better to have a tool which can generate a PKGBUILD from a tauri app, which fetches the tauri app's source via an url, fetches tauri as dependency and builds everything locally. This is something every AUR user can use.

Going this route does not block the generation of binaries for the repositories either, as repository packages are always meant to be built inside a chroot and be fully self-contained.

nothingismagick:
> It would actually be much better to have a tool which can generate a PKGBUILD from a tauri app, which fetches the tauri app's source via an url, fetches tauri as dependency and builds everything locally. This is something every AUR user can use.


Yes! This is so fundamentally important for proper package distribution. Thankyou for sharing your perspective!

tooling/bundler/src/bundle/linux/util.rs
```diff
@@ -0,0 +1,91 @@
+// Copyright 2019-2021 Tauri Programme within The Commons Conservancy
```
nothingismagick: Just a small note. We are now in 2022, and this should  be read by our regex properly:
```
// Copyright 2019-2022 Tauri Programme within The Commons Conservancy
```
Right @lucasfernog ?

nothingismagick (commented):
I can't test this on linux, but I think the idea is sound. Especially since its is proactive.

RubenKelevra:
@nothingismagick wrote:

> > It would actually be much better to have a tool which can generate a PKGBUILD from a tauri app, which fetches the tauri app's source via an url, fetches tauri as dependency and builds everything locally. This is something every AUR user can use.
>
> Yes! This is so fundamentally important for proper package distribution. Thankyou for sharing your perspective!

I'm happy to help, just mention me if there's anything you need to know to make sure I'm seeing this in time.

There's a [PKGBUILD-proto-file](https://gitlab.archlinux.org/pacman/pacman/blob/master/proto/PKGBUILD.proto) for creating packages under arch.


I think the best practice would be that you can download any Tauri archive. Run a command and get a PKGBUILD generated for it.

A package maintainer would compare this to his current PKGBUILD make the necessary changes and build the package for the repository in a clean chroot.

_I'll write a longer answer below how a generation could look like._

RubenKelevra:
The build instructions, named PKGBUILD, in Arch are pretty straight forward. Tauri can simply generate the build script and let the arch tooling do the job of compiling the tauri app by reading the PKGBUILD files.

The reason for this is, the whole [Arch User Repository](https://en.wikipedia.org/wiki/Arch_Linux#Arch_User_Repository_.28AUR.29) is not in binary files, but in PKGBUILD files. So the pregenerated files would only need minor adjustments to be uploaded there by the maintainers. The packages for the binary repositories of Arch are also published by the maintainers as PKGBUILD files, and then build and distributed as binaries, too.

If there's a tauri-app stored for example in a .zip-archive, it the url to it would be stored in the source-array like this: `source=("https://example.com/$pkgname-$pkgver.zip")`.

This will be extracted in the $srcdir, so as path it would end up in: `$srcdir/$pkgname-$pkgver/`.

And there the tauri app could be built.

It would be nice if the build-process could accept a general "package-root" variable where the files are placed, instead of putting them into the source folder under something like `src-tauri/target/release/bundle/linux/packageroot`

If that's not possible the package() { } section would look like that:

```bash
    cd "$srcdir/$pkgname-$pkgver"
    package_src='src-tauri/target/release/bundle/linux/packageroot'
    install -Dm 755 "$package_src/usr/bin/$pkgname" "$pkgdir/usr/bin/$pkgname"
    install -Dm 644 "$package_src/usr/share/applications/$pkgname.desktop" "$pkgdir/usr/share/applications/$pkgname.desktop"
    install -Dm 644 "$package_src/usr/share/icons/hicolor/256x256@2/apps/$pkgname.png" /usr/share/icons/hicolor/256x256@2/apps/$pkgname.png
    install -Dm 644 "$package_src/usr/share/icons/hicolor/32x32/apps/$pkgname.png" /usr/share/icons/hicolor/256x256@2/apps/$pkgname.png
    install -Dm 644 "$package_src/usr/share/icons/hicolor/128x128/apps/$pkgname.png" /usr/share/icons/hicolor/256x256@2/apps/$pkgname.png
```


## In detail:

| Field        | Necessary | Detail |
| :---         |   :---:   |          ---: |
| Maintainer   | ✔️        | Name of the person who's responsible for maintaining this package (*not* the builder of the package per se).  <br>So probably necessary to ask the user for this.  |
| `pkgname=`  | ✔️        | Could be filled by `jq '.package.productName' tauri.conf.json`. Should *never* contain parts of the version as a number.  |
| `pkgver=`   | ✔️        | Version set in the tauri config file: `jq --ascii-output --raw-output '.package.version' tauri.conf.json`. Field can contain letters, numbers, periods and underscores, but not a hyphen (-).  |
| `pkgrel=` | ✔️ | Is a arch internal version number for multiple package releases for the same version. Dependency fixes, build errors, etc. |
| `epoch=` | ❌ | Usually not needed. |
| `pkgdesc=` | ✔️ | Could be filled by `jq '.tauri.bundle.longDescription' tauri.conf.json` |
| `arch=()` | ✔️ | List of architectures. Should be something like `arch=('i686' 'x86_64' 'armv7h' 'aarch64')` |
| `url=` | ✔️ | e.g. https link to github project. Could maybe added as field to the tauri conf to have a source for that. |
| `license=()` | ✔️ | Just an array of the licenses of the package. Some Licenses need to be bundled into the the package, but for most it's enough to mention them there to fulfill the attribution. |
| `groups=()` | ❌ | Usually not needed. |
| `depends=()` | ✔️ | See docs: https://tauri.app/v1/guides/getting-started/prerequisites/#setting-up-linux |
| `makedepends=()` | ✔️ | Compilers and toolsets necessary only at compile time, like 'rust' |
| `checkdepends=()` | ❌ | Only if there are depencencies for an automatic testing function of the binary |
| `optdepends=()` | ❌ | Only necessary for optional depencencies for special functions |
| `provides=()` | ❌ |  |
| `conflicts=()` | ❌ |  |
| `replaces=()` | ❌ |  |
| `backup=()` | ❌ |  |
| `options=()` | ❌ |  |
| `install=` | ❌ | Should be filled by the package maintainer |
| `changelog=` | ❌ | Should be filled by the package maintainer |
| `source=()` | ✔️ | Should be filled by the package maintainer |
| `noextract=()` | ❌ |  |
| `b2sums=()` | ✔️ | Should be filled by the package maintainer |
| `validpgpkeys=()` | ❌ | Should be filled by the package maintainer |
| `prepare() { }` | ❌ | Prepare the build enviroment e.g. download stuff and apply patches. Docs: https://wiki.archlinux.org/title/Creating_packages#prepare() |
| `build() { }` | ❌ | Build the binary offline. Docs: https://wiki.archlinux.org/title/Creating_packages#build() |
| `package() { }` | ❌ | Move the files from the build enviroment to the package root to include them into the package. Docs: https://wiki.archlinux.org/title/Creating_packages#package() |

RubenKelevra:
@nothingismagick 

So 99% can currently be extracted from the tauri.conf and written to a PKGBUILD-file to read it with standard arch linux tools.

This would require a conversion tool of some kind ... should I write one? I'm currently gravitating towards shellscript for this.

If you guys think that's fine I would submit a patch for that. :)

oguzkaganeren:
Hi, it would be good if it is merged. Is there any reason to wait to merge the pull request?

tooling/bundler/src/bundle/linux/pacman.rs
```diff
@@ -0,0 +1,118 @@
...
+use std::process::Command;
+
+use std::{
+  fs,
+  io::Write,
+  path::{Path, PathBuf},
+};
+
+pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
+  // unimplemented!();
```
rmburg, line 36:
```suggestion
```

D3vil0p3r:
@oowl @lucasfernog @FabianLars  could you please merge this PR? Currently on Arch Linux the !strip option in PKGBUILD seems to not work, making impossible to create an AppImage, and consequently creating a Tauri app package in AL.

FabianLars:
Just reacting because i was pinged: I stopped working on Linux specific things some time ago so i keep somewhat ignoring this, sorry. Also, there was another approach/change proposed [here](https://github.com/tauri-apps/tauri/pull/8656) and there was not enough discussion for me as a non-arch user to quickly decide which is better. But again, i didn't look at that PR much either, so maybe it is obvious...

Lastly, from what i can see NO_STRIP=true still works absolutely fine, though building appimages on arch is of course not _that_ useful anyway since you can then only share your app with users of (updated) rolling release distros. -> That's probably why we didn't publish a hotfix that forces the NO_STRIP env var

PR #4302, opened 2022-06-08
fix: remove notification permission prompt

### LINKED ISSUE #1252: Notification's permission request should have i18n
We ask for permissions before sending notifications, and the message is always in English. Should we implement i18n or remove that instead? 

https://github.com/tauri-apps/tauri/blob/4d56e5e6919f04e2889900957be02fef1dd8f59f/tauri/src/endpoints/notification.rs#L84

### LINKED ISSUE #4244: [bug] Duplicate permission prompt for notifications on MacOS
### Describe the bug

Using the `notification.sendNotification` API, on first use I'm observing on MacOS Monterey 12.3.1 I'm getting duplicate permission prompts.

First the "soft prompt" from Tauri and then the native one "AppX want's to send you notifications" in the upper right corner.

### Reproduction

1. Use the notification API as per the above instructions.
2. Bundle the app and move it to the applications directory (i.e. don't run in dev mode).
3. Make sure to clear any pre-existing `.tauri-settings` files
4. Approve notifications in the first (Tauri's) dialog
5. Observe the second, native, permission dialog

### Expected behavior

On MacOS, the native dialog and permission system should probably be used, since it allows a lot of fine-grained control and users are expecting it. I don't think one can/should bypass this from Tauri's end.

### Platform and versions

```shell
> tauri
> ./bin/cargo-tauri "info"


Environment
  › OS: Mac OS 12.3.1 X64
  › Node.js: 17.3.0
  › npm: 8.3.0
  › pnpm: Not installed!
  › yarn: 1.22.10
  › rustup: 1.24.3
  › rustc: 1.59.0
  › cargo: 1.59.0

Packages
  › @tauri-apps/cli [NPM]: 1.0.0-rc.7(outdated, latest: 1.0.0-rc.13)
  › @tauri-apps/api [NPM]: 1.0.0-rc.6
  › tauri [RUST]: git+https://github.com/betamos/tauri?branch=betamos-head#99f5919ff2008e82d273b99200b60fbc9ff6d6d3 (1.0.0-rc.8),
  › tauri-build [RUST]: git+https://github.com/betamos/tauri?branch=betamos-head#99f5919ff2008e82d273b99200b60fbc9ff6d6d3 (1.0.0-rc.7),
  › tao [RUST]: 0.8.4,
  › wry [RUST]: 0.16.0,

App
  › build-type: bundle
  › CSP: default-src blob: data: filesystem: ws: wss: http: https: tauri: 'unsafe-eval' 'unsafe-inline' 'self' img-src: 'self'
  › distDir: ../ui/public
  › devPath: ../ui/public

App directory structure
  ├─ ui
  ├─ app
  ├─ bin
  ├─ node_modules
  ├─ agent
  ├─ scripts
  ├─ .git
  └─ .vscode
```

### Stack trace

_No response_

### Additional context

See also https://github.com/tauri-apps/tauri/issues/2317

betamos:
NOTE: This is a draft, so it doesn't yet have a change file and updated docs.

Highlighted a point below which could be important for 1.0 API compatibility.

Unresolved:

- Approval of this whole approach :)
- Whether this is ok w.r.t Apple guidelines. (See older MacOS versions below)
- Whether to remove or deprecate the `requestPermission` and `isPermissionGranted` APIs.

Platform specifics:

- Windows and Linux: permission prompts are not required nor common.
- MacOS 10.15+: A first-use prompt is provided by the OS.
- MacOS older versions: Notifications can be turned off in system settings.

Details:

- removed the permission dialog entirely
- removed the settings module, which only stored the notification preference:
- **the settings module used bincode, which makes no compatibility guarantees
  across rust versions, struct changes etc**
- removed the bincode dep
- the public API is unchanged, but permission related endpoints return
  static responses instead.

closes #4244
closes #1252

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


lorenzolewis:
Don't remove or deprecate the `requestPermission()` and `isPermissionGranted()`, those are important for a developer to know the appropriate action to take. I think what is returned from those needs to be adapted (same with `sendNotification()`), but that's for another discussion.

betamos:
Deprecation is probably not even a thing pre-1.0, I don't know why I even said that :)

Removing OTOH may, counter-intuitively, be a really good idea. It would leave a blank slate for precisely those API changes you're proposing. They would be easy to add later, but hard to change if they're cemented in 1.0. I doubt we have time to get those changes in pre-1.0. Heck, I'm even doubting we can get this PR in, even if people agree would agree with this approach.

In either case, leaving this PR as-is for now to solicit more feedback.

lorenzolewis:
To sum up what you currently have in this PR if I'm understanding correctly:
1. Remove `tauri::settings`, this was only used as part of notifications to date and the implementation wasn't solid
2. Remove logic that reads/writes from `tauri::settings` for notifications
3. Remove dialog/request to grant notification permission
4. Remove `is_permission_granted()`

For point 4, this is the only one we shouldn't remove. Right now it directly reads from `settings`, but I would temporarily just hard code this to return `True` until we can implement a proper OS-level check.

Keeping in mind that `tauri::settings` has only been used in this area so far, @lucasfernog what do you think of the code change contents? Do you think this is something we could merge prior to 1.0 mainly to resolve point 3 above?

lucasfernog:
I agree with you @lorenzolewis

lorenzolewis:
@betamos Do you think that gives you enough info to modify this and get it to a reviewable state with those changes? Only code change is NOT removing `is_permission_granted()` (and then anything else required for a review and merge)

betamos:
Thanks for the feedback. Point 4 refers to an internal method that is no longer needed. The public method remains, there are no API changes in this PR.

Feel free to triple check me on this. There's a little bit of magic in the `#[command_enum]` macro.

Undrafting then :)

core/tauri/src/endpoints/notification.rs
```diff
@@ -68,80 +60,23 @@ impl Cmd {
...
   fn request_notification_permission<R: Runtime>(
     _context: InvokeContext<R>,
   ) -> super::Result<&'static str> {
-    Ok(PERMISSION_DENIED)
-  }
-
-  #[cfg(notification_all)]
-  fn is_notification_permission_granted<R: Runtime>(
-    context: InvokeContext<R>,
-  ) -> super::Result<Option<bool>> {
```
lorenzolewis, line 92: This may not be the appropriate time to address it, but is there a reason this isn't just `Result<bool>`?

lucasfernog: changed it

lucasfernog: it previously returned `None` when the permission was never asked, so like an unknown state

lorenzolewis: Maybe I'm just looking at the wrong place in this file. But if it's changed in rust then would be good to change over in TS as well

lucasfernog (approved):

lucasfernog:
Btw @betamos with that many amazing contributions you should look into setting up [commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) :)

betamos:
> you should look into setting up [commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) :)

- "I have it setup already, what are you talking about?"
- Wait, no, why?
- Ok, I only set it up on my laptop, and I made these commits on my desktop.

Thanks for the reminder. Done now :)

And thanks for attending to this PR. Glad to have this in before 1.0.

PR #4305, opened 2022-06-09
feat(bundler): cache bundling tools in a common dir for all projects

amrbashir:

- [ ] Bugfix
- [X] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] No

### Checklist


amrbashir:
On Linux, I only changed the download path of `linux-deploy` and `AppRun`, I don't even know if the script was already caching them or downloads this files each `tauri build`

FabianLars:
Does the removal of -nc mean that it will always overwrite existing files? Not familiar with wget, sry.

I'm asking because this was the main concern I had with this PR because the appimage tooling is unversioned and it would be kinda bad if users would end up not receiving bug fixes etc because of this.

lucasfernog:
Ahh I forgot we're shipping updates like that, I'll remove the `-nc` flag so it always downloads it.

FabianLars:
Ah lol you _added_ the flag, gosh I can't read diffs in the GitHub mobile app 😂

amrbashir:
maybe the `-N` flag is better to only download if newer ?

lucasfernog:
Ahh `-N` makes sense `-N,  --timestamping              don't re-retrieve files unless newer than local`

PR #4457, opened 2022-06-24
Fix appimage creation in container when host has appimagelauncher enabled

dswd:
I ran into a problem that building the very same project in the very same container worked on one machine but not on the other (same host OS, roughly same version). The error was `./linuxdeploy-x86_64.AppImage: No such file or directory`. As I looked into the problem I found this:

If Appimagelauncher is enabled on the host, it modifies binfmt to inject itself as a launcher for appimages. This also applies to appimages in containers although the appimagelauncher binary is most likely not available there. So the "No such file" does not mean the appimage itself but the launcher it tries to execute.

The easiest way to fix this is to modify the appimage ELF header so that binfmt no longer identifies it as an appimage and lets the binary run itself. This is my PR.

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/.changes/appimage-fix.md
+++ b/.changes/appimage-fix.md
@@ -1,0 +1,5 @@
+---
+"tauri-bundler": patch
+---
+
+Fix AppImage bundling when appimagelauncher is enabled.
--- a/tooling/bundler/src/bundle/linux/templates/appimage
+++ b/tooling/bundler/src/bundle/linux/templates/appimage
@@ -68,6 +68,8 @@ wget -q -4 -N -O linuxdeploy-${ARCH}.AppImage https://github.com/linuxdeploy/lin
 chmod +x linuxdeploy-plugin-gtk.sh
 chmod +x linuxdeploy-${ARCH}.AppImage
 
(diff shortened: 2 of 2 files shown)


lucasfernog:
I tried running appimagelauncherd and the logs shows `ERROR: not an AppImage, skipping` on the dev branch. In your branch, the launcher exits with `Segmentation fault (core dumped)`. Can you double check @FabianLars ?

FabianLars:
@lucasfernog For me it behaves as described by the OP. Without the change appimagelauncher thinks it's an appimage, and with their branch it doesn't, without any side-effects for me (= the build succeeds).
That said i only tested it on my host machine (fedora36) without any container stuff. The latter needs to wait until tomorrow, but as i see no change for "normal" users i'd tend to give my okay for this change :thinking:

lucasfernog (approved):

lucasfernog:
A weird side effect is that I had to reboot after uninstalling appimagelauncher, it was giving me the same `./linuxdeploy-x86_64.AppImage: No such file or directory` error somehow. Just posting here in case someone sees it.

PR #4491, opened 2022-06-27
fix(core): add windows 7 notification support

lucasfernog:

- [x] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/.changes/is-windows-7.md
+++ b/.changes/is-windows-7.md
@@ -1,0 +1,5 @@
+---
+"tauri-utils": patch
+---
+
+Added `platform::is_windows_7`.
--- a/.changes/win7-notifications.md
+++ b/.changes/win7-notifications.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch
+---
+
+Added `Notification::notify` API which includes Windows 7 support.
--- a/.changes/win7-notifications.md
+++ b/.changes/win7-notifications.md
@@ -2,4 +2,4 @@
 "tauri": patch
 ---
 
-Added `Notification::notify` API which includes Windows 7 support.
+Added `Notification::notify` API behind the `windows7-compat` Cargo feature, which includes Windows 7 support.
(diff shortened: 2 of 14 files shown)


JonasKruckenberg:
I don't like that it's not an opt-in thing :/ because many people will not make use of this method. But I think we have too many features as-is, so I'm content.

~~Edit: Without knowing too much about the notifications feature, why can't this be a check inside `show`? That calls out to the `notify_win7` method?~~

EditEdit: I'm stupid and should pay attention to what's happening on discord

lucasfernog:
Basically the problem is that we need the event loop to be running and the notification's show method to be called in the main thread. The current show API cannot do that.

lucasfernog:
I could also introduce a `windows7` or `compatibility` Cargo feature that changes the `show` method signature.

JonasKruckenberg:
> I could also introduce a `windows7` or `compatibility` Cargo feature that changes the `show` method signature.

Hmmm, I mean on paper that is the most elegant solution IMO. But we already have so many features and this would just be one more thing to document 🤔 So I'm fine with whatever you decide.

If you go the feature route, name it `windows7` (or `windows7_compat`) though, we might have more specialized bug fixes for win7 in the future

amrbashir:
I am in favor of adding this behind a feature flag. Most tauri apps won't care about win7 and those who want to support, will need to pay the penalty of having to include another crate.

lucasfernog:
I've added the feature flag but kept the two different names for better documentation. I also put a deprecation warning on `show()` in case `windows7-compat` is enabled so you don't accidentally use it.

lucasfernog:
Done. Tomorrow I'll add the Windows 7 section in the distribution guide (webview install + notifications).

JonasKruckenberg (approved):

PR #4503, opened 2022-06-28
fix(config): remove `deny_unknown_fields`

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/.changes/remove-deny-unknown-fields.md
+++ b/.changes/remove-deny-unknown-fields.md
@@ -1,0 +1,5 @@
+---
+"tauri-utils": patch
+---
+
+Removed `deny_unknown_fields` from the configuration objects. This causes issues when the Tauri CLI is updated but the core crates aren't, and it isn't harmful since the CLI validates the config schema.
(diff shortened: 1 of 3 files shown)


FabianLars:
what do you think about logging unknown fields in the future (only on values _actually specified in the config_ if possible)?

lucasfernog:
The CLI validates the schema so we're not left in the dark here. Idk if it can be improved.

FabianLars:
Yeah, i was thinking about what if users see new features/config options and only update the cli and wonder why the config doesn't have any effect

lucasfernog:
I don't think there's an easy way to do that using serde. Maybe we should just improve the error message instead.

lucasfernog:
I think I'll change the CLI to pass only the merge config instead to avoid conflicts.

lucasfernog:
Closing in favor of #4598

PR #4504, opened 2022-06-28
fix(updater): fallback if tmp is on different mount point, closes #4500

### LINKED ISSUE #4500: [bug] Updater - Tauri API error: Invalid cross-device link (os error 18)
### Describe the bug

During the installUpdate() call, I receive the error "Tauri API error: Invalid cross-device link (os error 18)". I'm using a statically hosted update file following the [Update File JSON Format](https://tauri.app/v1/guides/distribution/updater/#update-file-json-format). I have the updater dialog set to false. I'm trying to update an AppImage.

I found another "Invalid cross-device link" [error](https://github.com/core-api/python-client/issues/176) in an unrelated python project.

### Reproduction

_No response_

### Expected behavior

_No response_

### Platform and versions

```shell
Environment
  › OS: Manjaro 21.3.1 X64
  › Node.js: 16.15.1
  › npm: 8.5.5
  › pnpm: Not installed!
  › yarn: Not installed!
  › rustup: 1.24.3
  › rustc: 1.61.0
  › cargo: 1.61.0
  › Rust toolchain: stable-x86_64-unknown-linux-gnu 

Packages
  › @tauri-apps/cli [NPM]: 1.0.0
  › @tauri-apps/api [NPM]: 1.0.1
  › tauri [RUST]: 1.0.0,
  › tauri-build [RUST]: 1.0.0,
  › tao [RUST]: 0.11.2,
  › wry [RUST]: 0.18.3,

App
  › build-type: bundle
  › CSP: unset
  › distDir: ../public
  › devPath: http://localhost:8080/
  › framework: Svelte
  › bundler: Rollup

App directory structure
  ├─ .git
  ├─ src-tauri
  ├─ public
  ├─ scripts
  ├─ .github
  ├─ .aws-sam
  ├─ src
  └─ node_modules
```

### Stack trace

_No response_

### Additional context

_No response_

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/core/tauri/src/updater/core.rs
+++ b/core/tauri/src/updater/core.rs
@@ -641,50 +641,54 @@ fn copy_files_and_run<R: Read + Seek>(archive_buffer: R, extract_path: &Path) ->
 
   let extract_path_metadata = extract_path.metadata()?;
 
-  let mut tmp_dir = tempfile::Builder::new()
-    .prefix("tauri_current_app")
-    .tempdir()?;
-  let mut tmp_dir_metadata = tmp_dir.path().metadata()?;
+  let tmp_dir_locations = vec![
+    Box::new(|| Some(env::temp_dir())) as Box<dyn FnOnce() -> Option<PathBuf>>,
+    Box::new(dirs_next::cache_dir),
+    Box::new(|| Some(extract_path.parent().unwrap().to_path_buf())),
+  ];
 
-  // fallback to cache_dir or CWD if the tmp directory is not in the same mount point
-  if extract_path_metadata.dev() != tmp_dir_metadata.dev() {
-    let fallback_dir =
-      dirs_next::cache_dir().unwrap_or_else(|| extract_path.parent().unwrap().to_path_buf());
-    tmp_dir = tempfile::Builder::new()
-      .prefix("tauri_current_app")
-      .tempdir_in(fallback_dir)?;
-    tmp_dir_metadata = tmp_dir.path().metadata()?
+  for tmp_dir_location in tmp_dir_locations {
+    if let Some(tmp_dir_location) = tmp_dir_location() {
+      let tmp_dir = tempfile::Builder::new()
+        .prefix("tauri_current_app")
+        .tempdir_in(tmp_dir_location)?;
+      let tmp_dir_metadata = tmp_dir.path().metadata()?;
+
+      if extract_path_metadata.dev() == tmp_dir_metadata.dev() {
+        let mut perms = tmp_dir_metadata.permissions();
+        perms.set_mode(0o700);
+        std::fs::set_permissions(tmp_dir.path(), perms)?;
+
+        let tmp_app_image = &tmp_dir.path().join("current_app.AppImage");
+
+        // create a backup of our current app image
+        Move::from_source(extract_path).to_dest(tmp_app_image)?;
+
+        // extract the buffer to the tmp_dir
+        // we extract our signed archive into our final directory without any temp file
+        let mut extractor =
+          Extract::from_cursor(archive_buffer, ArchiveFormat::Tar(Some(Compression::Gz)));
+
+        return extractor
+          .with_files(|entry| {
+            let path = entry.path()?;
+            if path.extension() == Some(OsStr::new("AppImage")) {
+              // if something went wrong during the extraction, we should restore previous app
+              if let Err(err) = entry.extract(extract_path) {
+                Move::from_source(tmp_app_image).to_dest(extract_path)?;
+                return Err(crate::api::Error::Extract(err.to_string()));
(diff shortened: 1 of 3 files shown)


core/tauri/src/updater/core.rs
```diff
@@ -640,14 +635,28 @@ impl<R: Runtime> Update<R> {
...
+  let mut tmp_dir_metadata = tmp_dir.path().metadata()?;
+
+  // fallback to cache_dir or CWD if the tmp directory is not in the same mount point
+  if extract_path_metadata.dev() != tmp_dir_metadata.dev() {
+    let fallback_dir =
+      dirs_next::cache_dir().unwrap_or_else(|| extract_path.parent().unwrap().to_path_buf());
+    tmp_dir = tempfile::Builder::new()
+      .prefix("tauri_current_app")
+      .tempdir_in(fallback_dir)?;
+    tmp_dir_metadata = tmp_dir.path().metadata()?
```
Beanow: It doesn't look like we're checking the metadata again to be sure those fallbacks are on the same mountpoint?
Especially if `~/.cache` is found and preferred over the current `.appimage` path, this seems quite possible.
I could be running the `appimage` from a network share with `~/.cache` on a local btrfs volume.

core/tauri/src/updater/core.rs
```diff
@@ -661,21 +670,21 @@ fn copy_files_and_run<R: Read + Seek>(archive_buffer: R, extract_path: &Path) ->
...
-        return Err(crate::api::Error::Extract(err.to_string()));
+  extractor
+    .with_files(|entry| {
+      let path = entry.path()?;
+      if path.extension() == Some(OsStr::new("AppImage")) {
+        // if something went wrong during the extraction, we should restore previous app
+        if let Err(err) = entry.extract(extract_path) {
+          Move::from_source(tmp_app_image).to_dest(extract_path)?;
+          return Err(crate::api::Error::Extract(err.to_string()));
+        }
```
Beanow: Since you found a nice option to compare the mountpoints, the ultimate fallback imo should be to avoid move, and copy across the filesystem boundaries instead.

lucasfernog: well I didn't want to change the logic that much specially since it was audited already. I think if we fallback to the current appimage directory we're safe.

core/tauri/src/updater/core.rs
lucasfernog: You're right. I'll see how I can change that.

core/tauri/src/updater/core.rs
Beanow: That is, if none of the 3 dir options appear to be on the same mount point.

Beanow: > well I didn't want to change the logic that much specially since it was audited already. I think if we fallback to the current appimage directory we're safe.

Got it, sure that makes sense. Because I would _expect_ a sibling to the appimage to be the same file system.
Though we should still check that (previous comment).

core/tauri/src/updater/core.rs
Beanow: A loop of candidates maybe :] And it should solve a second issue, namely right now the CWD won't be considered if `dirs_next::cache_dir()` is `Some`.

lucasfernog:
I also removed the `#[cfg(feature = "updater")]`in the core.rs file since we do not have an specific feature for updater documentation generation anymore (we previously had one because the updater feature enabled the zip crate, which didn't cross compile for docs.rs - now fixed by removing its default features).

core/tauri/src/updater/core.rs
lucasfernog: done @Beanow code might look weird with the vector of closures, but i think it's the way to remove duplicated code lol

core/tauri/src/updater/core.rs
lucasfernog: Yeah I did that though it sounds unlikely that it would happen.

core/tauri/src/updater/core.rs
Beanow: Yeah that looks like a good approach to me :ok_hand:

core/tauri/src/updater/core.rs
Beanow: Current solution looks like a fair middle ground, trying three candidates to move or a clear error message :+1:

Beanow:
I'll leave the call up to a maintainer, but my thoughts are addressed :ok_hand:

lucasfernog:
Thanks for the review @Beanow

PR #4509, opened 2022-06-28
feat(bundler/msi): add `perMachine` option, disabled by default, closes #2319

### LINKED ISSUE #2319: perUser InstallScope
I am unsure if there are technical reasons for this not being possible, but wish to raise it to see if it is a possibility.

Could the windows installer support the perUser InstallScope (and I think maybe it'd need the limited InstallPrivileges) wix settings so that a Tauri app would not require admin privileges to install?

I appreciate WebView2 will have it's own limitations, but if that were already installed, could the app itself be less restricted in installation options?

**Additional context**
N/A

amrbashir:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes, and the changes were approved in issue #___
- [ ] No

### Checklist

### DIFF
--- a/.changes/wix-per-machine.md
+++ b/.changes/wix-per-machine.md
@@ -1,0 +1,9 @@
+---
+"tauri-bundler": "patch"
+"cli.rs": "patch"
+"cli.js": "patch"
+---
+
+Added `perMachine` option to `tauri.conf.json > tauri > bundle > windows > wix`, disabled by default.
+**Breaking change** `msi` installer will install your app per-user by default now.
+
--- a/tooling/bundler/src/bundle/windows/msi/wix.rs
+++ b/tooling/bundler/src/bundle/windows/msi/wix.rs
@@ -637,6 +637,14 @@ pub fn build_wix_app_installer(
     data.insert("feature_group_refs", to_json(&wix.feature_group_refs));
     data.insert("feature_refs", to_json(&wix.feature_refs));
     data.insert("merge_refs", to_json(&wix.merge_refs));
+    if wix.per_machine {
+      data.insert("per_machine", to_json(true));
+      data.insert("install_scope", to_json("perMachine"));
+      data.insert("registry_root", to_json("HKLM"));
+    } else {
+      data.insert("install_scope", to_json("perUser"));
+      data.insert("registry_root", to_json("HKCU"));
+    }
     fragment_paths = wix.fragment_paths.clone();
     enable_elevated_update_task = wix.enable_elevated_update_task;
 
--- a/core/tauri-utils/src/config.rs
+++ b/core/tauri-utils/src/config.rs
@@ -402,9 +402,12 @@ pub struct WixConfig {
   pub banner_path: Option<PathBuf>,
   /// Path to a bitmap file to use on the installation user interface dialogs.
   /// It is used on the welcome and completion dialogs.
-
+  ///
   /// The required dimensions are 493px × 312px.
   pub dialog_image_path: Option<PathBuf>,
+  /// Install for all users on the machine.
+  #[serde(default)]
+  pub per_machine: bool,
 }
 
 /// Install modes for the Webview2 runtime.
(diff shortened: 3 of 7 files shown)


.changes/wix-per-machine.md
```diff
@@ -0,0 +1,9 @@
+---
+"tauri-bundler": "patch"
+"cli.rs": "patch"
+"cli.js": "patch"
+---
+
+Added `perMachine` option to `tauri.conf.json > tauri > bundle > windows > wix`, disabled by default.
+**Breaking change** `msi` installer will install your app per-user by default now.
```
lucasfernog, line 8: not really a breaking change right?

amrbashir: It is a slight breaking change, that's why I kept the bump as `patch` (still considering to change it to `minor`). 

Apps that relied heavily on being installed `perMachine` won't work unless they specify `perMachine`. I don't have an example at the top of my head but I thought I'd be cautious and treat it as breaking.

amrbashir:
Only tested with `examples/api`, will test `examples/resources` and `examples/sidecar` in a bit.

.changes/wix-per-machine.md
FabianLars: doesn't it make more sense to keep it enabled by default then?

Edit: Actually i think i prefer perUser being the default, nvm me

lucasfernog: I prefer perUser as default but if it can break any kind of app, we need to keep the per machine as default.

amrbashir: it has a slight chance to break some apps but I think `perUser` should be the default, even if it means this PR shouldn't be merged in v1

FabianLars: even tho i can't think of anything really where changing it to perUser would break anything, but what about merging this pr with perMachine as default and changing the default with v2?

Edit: Okay it can break if the user selects a folder like Program Files, because it doesn't request permissions on demand (is that even possible?)

tooling/bundler/src/bundle/windows/msi/wix.rs
```diff
@@ -637,6 +637,14 @@ pub fn build_wix_app_installer(
...
     data.insert("feature_refs", to_json(&wix.feature_refs));
     data.insert("merge_refs", to_json(&wix.merge_refs));
+    if wix.per_machine {
+      data.insert("per_machine", to_json(true));
+      data.insert("install_scope", to_json("perMachine"));
+      data.insert("registry_root", to_json("HKLM"));
+    } else {
+      data.insert("install_scope", to_json("perUser"));
+      data.insert("registry_root", to_json("HKCU"));
+    }
```
FabianLars, line 647: The else branch (or the default for that matter) needs to be moved out of the `if let Some(wix)` check. Otherwise the bundler will crash if there is no wix property in tauri.conf.json

FabianLars (changes requested):

tooling/bundler/src/bundle/windows/msi/wix.rs
amrbashir: thanks, I fixed it locally but hasn't pushed it yet. there is more issues to solve

.changes/wix-per-machine.md
halo-apps: I have to move my glow stick for you guys. 
I think this is a very important function and it will be very helpul if it can be merged and published as soon as possible.

MarcWadai:
Do you know the reasons this PR was closed and not merged @FabianLars ? The default is still perMachine for MSI installer. Is there any work on going to add this flag just like for NSIS ?

FabianLars:
i don't know the exact (technical) reason but it was closed in favor of the nsis installer which was added after this PR. I also know that quite literally every single stackoverflow discussion about perUser .msi installers always say "don't do it" 😂

MarcWadai:
@FabianLars thx for the answer. For public sector or companies that want the advantage of auto updates with users who does not have admin privileges, this features is essentials ! It is more a standard to deploy MSI than EXE programs. So the best way to go with it i guess is to use a [Custom WiX File](https://v2.tauri.app/distribute/windows-installer/#replacing-the-installer-code-with-a-custom-wix-file)

FabianLars:
Let me know how it goes :) Missing real world examples/experience is one of the reasons I never looked much into this again.

PR #4521, opened 2022-06-29
fix(bundler): Build AppImages inside the target folder

Beanow:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

Previously to avoid an issue with AppImage builds we collected all our output files in `~/.cache`. This solves the reason why we did that and moves back to putting our temporary files in the `target` folder again.

The main problem at that time was, #4305 introduced a change where build tools should be stored in `~/.cache/tauri`.
However one particular file that we download, `AppRun`, was supposed to be bundled, instead of executed as a build tool.
The path where it needed to be was in `{{app_name}}.AppDir/AppRun`.

I've fixed it here by still downloading all dependencies to `~/.cache/tauri`, but also making a copy of `AppRun` in the correct place. Allowing builds in `target` to work again.

---

As an aside, building in `target` is preferable because we don't use a conflict-free unique or fully qualified name when building AppImages, so race conditions might happen when building multiple projects.

Also, "cleaning target" is an easy way to remove your project specific caches and intermediate builds, to get a clean slate. Which wouldn't apply for AppImages as they're not always in `target`.

### DIFF
--- a/tooling/bundler/src/bundle/linux/templates/appimage
+++ b/tooling/bundler/src/bundle/linux/templates/appimage
@@ -16,14 +16,8 @@ else
     linuxdeploy_arch="$ARCH"
 fi
 
-OUTDIR="${PWD}"
-
-cd "{{tauri_tools_path}}"
-
-# remove the folder if it exists - the rest of the script fails if it does
-rm -rf "{{app_name}}.AppDir"
 mkdir -p "{{app_name}}.AppDir"
-cp -r "${OUTDIR}/../appimage_deb/data/usr" "{{app_name}}.AppDir"
+cp -r ../appimage_deb/data/usr "{{app_name}}.AppDir"
 
 cd "{{app_name}}.AppDir"
 mkdir -p "usr/bin"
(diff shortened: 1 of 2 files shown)


lucasfernog:
Works like a charm. Why is this a draft @Beanow ?

Beanow:
Still needs a change entry, and I was thinking if more refactors should be included. Like, asserting the file ends up in the right place now to avoid regressions.

As is it should work though.

tooling/bundler/src/bundle/linux/templates/appimage
```diff
@@ -71,14 +69,12 @@ else
...
-wget -q -4 -N -O linuxdeploy-${ARCH}.AppImage https://github.com/linuxdeploy/linuxdeploy/releases/download/continuous/linuxdeploy-${linuxdeploy_arch}.AppImage
+wget -q -4 -N -O "{{tauri_tools_path}}/linuxdeploy-plugin-gtk.sh" https://raw.githubusercontent.com/tauri-apps/linuxdeploy-plugin-gtk/master/linuxdeploy-plugin-gtk.sh
+wget -q -4 -N -O "{{tauri_tools_path}}/linuxdeploy-${ARCH}.AppImage" https://github.com/linuxdeploy/linuxdeploy/releases/download/continuous/linuxdeploy-${linuxdeploy_arch}.AppImage
 
-chmod +x linuxdeploy-plugin-gtk.sh
-chmod +x linuxdeploy-${ARCH}.AppImage
+chmod +x "{{tauri_tools_path}}/linuxdeploy-plugin-gtk.sh"
+chmod +x "{{tauri_tools_path}}/linuxdeploy-${ARCH}.AppImage"
 
 dd if=/dev/zero bs=1 count=3 seek=8 conv=notrunc of=linuxdeploy-${ARCH}.AppImage
```
Beanow: Ah, see this one I missed.

Beanow: Patched this and set it as ready for review

lucasfernog (approved):

PR #4537, opened 2022-06-30
feat(cli.rs) Shell completions

JonasKruckenberg:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

This adds a subcommand that can generate shell completions for the following shells:
- bash
- zsh
- powershell
- fish
- elvish


lucasfernog:
:man_dancing:

JonasKruckenberg:
There's a fix in that PR for a panick the the `signer sign` command, but I cherry picked that to #4538 so we can have the fix even without any new features being merged into v1

lucasfernog:
Ahh but I wanted to merge this :(((((((((((((( it's a DX feature and the developer can check the script before enabling it.

lucasfernog:
I only tested this on Linux so far.

JonasKruckenberg:
I tested on macOS too, maybe we can can signoff from secdude and @nothingismagick regardless. I think this would be pretty uncontroversial. Even though it adds another dependency sooo idk :/

lucasfernog:
I couldn't get it to work on macOS, specially we need it to work when installing with cargo and running with `cargo tauri`

JonasKruckenberg:
True yeah there are many situations where this doesn't work. Let me look into it

simonhyll:
Bit late to the party but for me it works just fine in bash and I'd love to see this merged even if it doesn't work for all shells. I feel it's better to merge a feature that works for some (especially after having sat on the shelf this long) and we can worry about fixing it for non-bash shells later.

simonhyll (dismissed):
Works on Bash just fine and I feel we can fix other shells later.

simonhyll:
To use it just run `tauri completions --shell bash > tauri.sh`, inspect the file for malicious code, move the file to the right place `sudo mv tauri.sh /etc/bash_completion.d/tauri`

Or, if you're brave, `sudo tauri completions --shell bash > /etc/bash_completion.d/tauri`

lucasfernog:
For reference here's what I did to use it on ZSH:

Append the following to .zshrc:
```
fpath=(~/.completions $fpath)
autoload -U compinit
```

Run the following:
```
tauri completions -s zsh > completions.zsh
mv completions.zsh $HOME/.completions/_tauri
```

lucasfernog:
Found a way to fix PowerShell. Note that to use the completions subcommand redirecting output to a file, you must use `yarn -s` (same flag for other package managers) to omit the yarn output.

lucasfernog:
@JonasKruckenberg @simonhyll I've pushed some changes here to make it work on Bash/Zsh/PowerShell. ~~I know you tested Bash already, but I had to push the underscore thing to fix it on my end. Let me know if you spot any issues.~~

Currently the command outputs the completion script for all package managers (cargo, yarn, npm, pnpm), but we could make it only generate the completion for the current manager too (reduces a lot of the script for non-bash shells).

EDIT: actually I don't need all that replace hack, that's better.

lucasfernog (dismissed):
I looooooooooooooooooove this

simonhyll (approved):

simonhyll (approved):

simonhyll:
> I looooooooooooooooooove this

Ikr?? It's amazing :D

Personally I don't think it matters that the output is slightly longer, the default behavior outputting all of them I think is desirable at least. If we want to limit the output to only specific package managers I'd prefer it if we made an option for manually specifying which to output.

So something like `tauri completions --shell bash --managers npm,cargo,yarn,custom-pm` (we dont have to verify the entered managers, just loop them, makes it more future proof when new package managers come out)
Maybe with an extra `tauri completions --shell bash --current-manager`

lucasfernog (approved):

PR #4577, opened 2022-07-03
fix(updater): do not run in dialog mode on .deb, closes #4573

### LINKED ISSUE #4573: [feat] Updater should handle unmanaged install modes better
### Describe the problem

When running the updater with default settings (`dialog: true`), it will prompt to download & install and fail silently if running e.g. as a `.deb`. The biggest issue is that this is confusing:

- It offers to do something it cannot do.
- It doesn't report the failure to the user.

A few options:

- **unmanaged install, unmanaged updates** (msi, appimage, dmg): search & prompt to install (works today)
- **managed install, unmanaged updates** (deb*, rpm*):
  - search for updates & offer to open download link in browser (tauri.conf.json field?) or
  - search for updates, download and launch the installer for the user, and let the user authenticate etc
-  **managed install, managed updates** (mac app store & windows store?): don't search for updates (fully managed)
- **N/A** (dev mode): don't search for updates at all?

*: don't know if possible to tell the difference between "deb repository url" which IS managed or ad-hoc installed deb (most common today I assume). If it's not feasible, it could be best to not search at all.

### Alternatives considered

_No response_

### Additional context

An additional issue is that we only have one url per [target, arch] tuple, which is hard to extend. E.g. linux *infers* appimage. This would make it tricky to add support for e.g. automatic deb downloads.

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/.changes/updater-dialog-unsupported-platform.md
+++ b/.changes/updater-dialog-unsupported-platform.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch
+---
+
+Only run the updater default dialog mode in supported platforms or development mode.
--- a/core/tauri/src/app.rs
+++ b/core/tauri/src/app.rs
@@ -794,21 +794,27 @@ impl<R: Runtime> App<R> {
     // check if updater is active or not
     if updater_config.active {
       if updater_config.dialog {
-        // if updater dialog is enabled spawn a new task
-        self.run_updater_dialog();
-        // When dialog is enabled, if user want to recheck
-        // if an update is available after first start
-        // invoke the Event `tauri://update` from JS or rust side.
-        handle.listen_global(updater::EVENT_CHECK_UPDATE, move |_msg| {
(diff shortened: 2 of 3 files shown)


betamos:
Didn't know you went ahead with this, so did some duplicate work, see https://github.com/betamos/tauri/commit/0569eb77f94c772f23e8ae50080f494e087197fe

Anyway, I also added a dev mode check to error out if installation is attempted, which we might wanna have to prevent touching files. I can remove the appimage check and send a PR for that part only, perhaps?

lucasfernog:
@betamos I think disabling it on dev makes it harder to develop a custom updater flow, even though the installation will likely fail. We could document it instead.

PR #4598, opened 2022-07-05
refactor: do not pass entire config from CLI to core, send patch instead

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


lucasfernog:
This PR changes the TAURI_CONFIG environment variable to represent the config JSON from the CLI's `config` argument instead of the entire JSON. This prevents issues when the CLI is newer than the core crates, and sends a config value that is unknown to the application, resulting in a build error.

FabianLars (approved):

FabianLars:
tbh i expected this to break _something_ but everything i tested (same version,cli newer, cli older) worked lol, except of course if you add conf properties one side doesn't know of. Because of that we mayyy wanna consider improving the tauri-build panic message _in the future_, because this:
```sh
 [...]
  --- stderr
  thread 'main' panicked at 'error found during tauri-build: Error("unknown field `kekw`, expected `bundleMediaFramework`", line: 0, column: 0)', C:\Users\Fabian-Lars\dev\tauri\core\tauri-build\src\lib.rs:172:5
warning: build failed, waiting for other jobs to finish...
       Error failed to build app: failed to build app
```
not only scares non-rust users but nobody really understands/knows that this means that they need to update their crates.
But that's not important for now imo, just wanted to mention it so we have it in mind.

lucasfernog:
I'll adjust the error message in a separate PR.

PR #4674, opened 2022-07-15
feat(bundler): add `nsis`, closes #4450, closes #2319  

### LINKED ISSUE #2319: perUser InstallScope

### LINKED ISSUE #4450: [feat] Add NSIS support
### Describe the problem

Wix toolset, a bit difficult to use, ugly interface, maybe I'm not using it properly. 

Can you provide more packaged installation tools? such as NSIS,Squirrel.Windows

### Alternatives considered

_No response_

### Additional context

_No response_

amrbashir:

- [ ] Bugfix
- [X] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] No

### Checklist

### TODOs:

- [x] resources
- [x] external binaries
- [x] set AppUserModelID for notifications, jumplists, grouping...etc
- [x] downgrades and upgrades
- [x] Webview2 install modes
- [x] Prompt to kill if the app is running
- [x] updater package
- [x] change file
- [ ] deep link


FabianLars:
The last thing i can think of would be support for other version formats than just `major.minor.patch`.

I just disabled the version check to test it, and it somewhat works as-is. Just noticed 2 problems (prob more i didn't find?):
1. the version in `programs & features` is wrapped in quotes `"version"`
2. Checking the exe properties: product version has the correct version (in this case 0.1.0-alpha) but file version doesn't (0.1.0.0 in this case) - not sure if these are system limitations

Do you know nsis (and potentially windows) limitations/specs here? i'm having difficulties reading the nsis docs, i didn't find any mentions about this.

amrbashir:
> 1. the version in `programs & features` is wrapped in quotes `"version"`

That is actually my fault. It seems that I was wrapping the version in quotes in the registry key.

> 2. Checking the exe properties: product version has the correct version (in this case 0.1.0-alpha)  file version doesn't (0.1.0.0 in this case) - not sure if these are system limitations 

I think this is a Windows limitation.

I think we can allow using `major.minor.patch-pre` format but we would have to write a custom `VersionCompare` function because the built-in doesn't handle that format well, which is needed to detect if we should upgrade, downgrade or re-install

core/config-schema/schema.json
```diff
@@ -1618,6 +1639,46 @@
...
+          "type": [
+            "string",
+            "null"
+          ]
+        },
+        "perMachine": {
+          "description": "Whether the installation will be for all users or just the current user.",
+          "default": false,
+          "type": "boolean"
+        }
```
seahindeniz: Ah, just saw this!

If it's possible: what about having a third state, and making the value more understandable without a boolean value?
- `"allUsers"` - All users
- `"currentUser"` - Current user
- Accepts object
  - `"target"` : `"both" | "allUsers" | "currentUser"`
  - `"recommendedOption"` or `"highlightedOption"` : `"allUsers" | "currentUser"`

The third object will make both options visible and recommend an item, instead of forcing to install the app to a pre-selected target.

amrbashir: The thing is if your app logic depends on writing some files next to the exe, you can't use `all_users` so I wanted the system-wide option to be explicitly opted-in, otherwise your users might report unexpected errors because they installed it to a location that doesn't have write access. Also, that the installer will have to be launched as administrator even if the user plan to use `current_user`.

amrbashir: I think I might change this option to be

* `"all_users"` - All users
* `"current_user"` - Current user
* `"both"`

Although this might complicate the logic of the installer. Thanks for the feedback.

seahindeniz: The elevated access may also be needed and can be implemented, and I already stumbled upon an issue and after discussing with Fabian on Discord, this issue got created #5584.

> Also, that the installer will have to be launched as administrator even if the user plan to use current_user

This kind of concerns me. Some users may not have an admin privilege and still want to install the app. I'm assuming there's no way around that as well?

amrbashir: > > Also, that the installer will have to be launched as administrator even if the user plan to use current_user
> 
> This kind of concerns me. Some users may not have an admin privilege and still want to install the app. I'm assuming there's no way around that as well?

This will only be a concern if you use `all_users` or `both` option when building the installer. Unfortunately, I couldn't find a way around that.

amrbashir: The only way I found is to write 3 installers and combine, one that prompts users to choose `current_user` or `all_users` and includes the other two installers inside it and depending on that, it writes the appropriate installer to disk in `%TEMP%` and launches one of the other two installers. Which is messy and will pollute the installer size.

lucasfernog:
You made it look easy @amrbashir let me know if I did something wrong. Let's merge this tomorrow :heart:

amrbashir:
There are two features I was hoping to address before merge:
- [X] Removing the version limitation and allow `1.0.0-alpha.1`
- [ ] Add combined install modes

I will see if I can finish them tomorrow, if not, I can add them later in another PR.

lucasfernog:
We can add it later too no problem.

amrbashir:
Sure go ahead and merge if you want, I will add the missing features laeter.

tooling/bundler/src/bundle/windows/templates/SemverCompare.nsh
```diff
@@ -0,0 +1,150 @@
...
+
+    ${If} $V3 > $VR3
+        Goto higher
+    ${EndIf}
+
+    ${If} $V3 < $VR3
+        Goto lower
+    ${EndIf}
+
+    ${If} $V4 == "beta"
```
lucasfernog: Any chance we can use a generic logic instead of only allowing alpha/beta as prerelease name? So instead of checking for `== "beta"` we check for `!= ""` and check if `$V4` is lower than `$VR4` (with StrCmp) instead of comparing with `"alpha"`.

amrbashir: yeah, I wasn't sure how semver treats other preleases other than `beta` and `alpha` so thought I would hard code them. I will try with your suggestions.

lucasfernog: it's always a string comparison. `alpha < beta < rc` for instance.

amrbashir: alright, this is finally fixed. Feel free to merge as is. I will work on the other feature in another PR.

amrbashir:
took me a lot of trials and errors and workarounds but `both` modes are finally supported. This PR is ready to be merged.

core/tauri-utils/src/config.rs
```diff
@@ -416,6 +420,57 @@ pub struct WixConfig {
...
+  /// Install the app by default in a `Prgoram Files` which require Adminstrator access
+  /// and so the install will also require Adminstrator access.
+  ///
+  /// Installer metadata will also ve saved under `HKLM` registry path.
+  PerMachine,
+  /// Combines both modes and allows the user to choose at install time
+  /// whether to install for current user or per machine. Note that this mode
+  /// will also require Adminstrator access even if the user wants to install it for current user only.
+  ///
+  /// Installer metadata will also ve saved under `HKLM` or `HKCU` registry path based on the user choice.
```
FabianLars:
```suggestion
  /// Default mode for the installer.
  ///
  /// Install the app by default in a directory that doesn't require Administrator access.
  ///
  /// Installer metadata will be saved under the `HKCU` registry path.
  CurrentUser,
  /// Install the app by default in the `Program Files` folder directory requires Administrator
  /// access for the installation.
  ///
  /// Installer metadata will be saved under the `HKLM` registry path.
  PerMachine,
  /// Combines both modes and allows the user to choose at install time
  /// whether to install for the current user or per machine. Note that this mode
  /// will require Administrator access even if the user wants to install it for the current user only.
  ///
  /// Installer metadata will be saved under the `HKLM` or `HKCU` registry path based on the user's choice.
```
Meant to fix only a few typos (`ve` - `be`) but adjusted the wording a little bit, but i can revert that if you want.
Oh and schema.json of course needs the same changes.

amrbashir: yeah I like that.

lucasfernog (approved):

PR #4730, opened 2022-07-21
[Tests] Add a constructor for State that is available for unittests only

JeremyRubin:
This adds a cfg(test) guarded method for directly constructing a State object, which enables writing unittests against tauri commands. Otherwise, tauri commands need an internal helper function to get references and call the method for testing, and can't get 100% coverage.

- [x] Other, please describe: Improvement to testing infrastructure

- [x] No

### Checklist

### DIFF
--- a/core/tauri/src/state.rs
+++ b/core/tauri/src/state.rs
@@ -19,6 +19,13 @@ impl<'r, T: Send + Sync + 'static> State<'r, T> {
   pub fn inner(&self) -> &'r T {
     self.0
   }
+
+  /// Create a State for testing commands in unittests
+  #[cfg(test)]
+  pub fn new_testonly(r: &'r T) -> Self {
+      State(r)
+  }
+
 }
 
 impl<T: Send + Sync + 'static> std::ops::Deref for State<'_, T> {


core/tauri/src/state.rs
```diff
@@ -19,6 +19,13 @@ impl<'r, T: Send + Sync + 'static> State<'r, T> {
   pub fn inner(&self) -> &'r T {
     self.0
   }
+
+  /// Create a State for testing commands in unittests
+  #[cfg(test)]
```
lucasfernog, line 24: This makes the `new_testonly` function only available to the Tauri crate itself. Here's how rocket handles it: https://api.rocket.rs/v0.4/rocket/struct.State.html#testing-with-state
We do have some test helpers to create a dummy App instance for testing, we should expose that instead.

JeremyRubin: State::from is not as good an approach because it exposes a new API that means it is a breaking change if State ever ends up holding other things internally. A dummy App might be an OK solution too, though.

lucasfernog: Let's track that over #4752

PR #4738, opened 2022-07-22
feat(updater): use full path to msiexec

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/.changes/msiexec-full-path.md
+++ b/.changes/msiexec-full-path.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch
+---
+
+Use full path to `msiexec.exe` in the updater.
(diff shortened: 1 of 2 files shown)


FabianLars:
I don't think we can do this, while the default by a huge margin, there still is no guarantee that C is always the systemroot

lucasfernog:
You're right :( cc @amrbashir

amrbashir:
yeah we need to to use `%SystemRoot%\system32\msiexec.exe`

`%SystemRoot%` is impossible to modify because it is set when Windows is installed. 
More info can be found here https://renenyffenegger.ch/notes/Windows/dirs/Windows/index

PR #4786, opened 2022-07-28
feat(bundler): cache appimage bundle tools

brian14708:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/.changes/cache-appimage-tools.md
+++ b/.changes/cache-appimage-tools.md
@@ -1,0 +1,6 @@
+---
+"cli.rs": patch
+---
+
+Avoid re-downloading AppImage build tools on every build.
+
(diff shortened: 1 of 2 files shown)


FabianLars:
This is what the `-N` flag is for, it enables wget's timestamping so that it only downloads it if the remote file is newer.
With this if check it would never re-download the files even if there's a new version which happens relatively regular.

Thank you nonetheless.

brian14708:
> This is what the `-N` flag is for, it enables wget's timestamping so that it only downloads it if the remote file is newer.

After some testing, I've found that `-N` and `-O` can't be used together and the `-q` option hides the warning.

Quote [man](https://man7.org/linux/man-pages/man1/wget.1.html):
>           For this reason, -N (for timestamp-checking) is not supported
>           in combination with -O: since file is always newly created,
>           it will always have a very new timestamp. A warning will be
>           issued if this combination is used.

brian14708:
> With this if check it would never re-download the files even if there's a new version which happens relatively regular.

I've modified the code to keep using `wget -N`, if you are still interested. 351b676b09584a06daab5e6c0a19ab6811a309ea

FabianLars:
> After some testing, I've found that -N and -O can't be used together and the -q option hides the warning.

Ahh interesting!

> I've modified the code to keep using wget -N, if you are still interested.

Yep, i am interested in that! Buuut i can't re-open PRs after there was a force-push (or branch recreation), see [isaacs/github#361](https://togithub.com/isaacs/github/issues/361), sorry for closing it too quickly, didn't know that this is a thing 😅

PR #4884, opened 2022-08-07
refactor(cli/core/bundler): use absolute path for msiexec and powershell

amrbashir:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/core/tauri/src/updater/core.rs
+++ b/core/tauri/src/updater/core.rs
@@ -802,13 +802,18 @@ fn copy_files_and_run<R: Read + Seek>(
       msi_path_arg.push("\"\"\"");
 
       // run the installer and relaunch the application
-      let powershell_install_res = Command::new("powershell.exe")
+      let system_root = std::env::var("SYSTEMROOT");
+      let powershell_path = system_root.as_ref().map_or_else(
+        |_| "powershell.exe".to_string(),
+        |p| format!("{p}\\System32\\WindowsPowerShell\\v1.0\\powershell.exe"),
+      );
+      let powershell_install_res = Command::new(powershell_path)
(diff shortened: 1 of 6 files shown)


core/tauri/src/updater/core.rs
```diff
@@ -802,13 +802,18 @@ fn copy_files_and_run<R: Read + Seek>(
...
+        |p| format!("{p}\\System32\\WindowsPowerShell\\v1.0\\powershell.exe"),
+      );
+      let powershell_install_res = Command::new(powershell_path)
         .args(["-NoProfile", "-windowstyle", "hidden"])
         .args([
           "Start-Process",
           "-Wait",
           "-FilePath",
-          "msiexec",
+          "$env:SYSTEMROOT\\System32\\msiexec.exe",
```
lucasfernog, line 816: shouldn't we fallback to `msiexec` here too in case SYSTEMROOT is empty? or does powershell always fill it?

amrbashir: - Outside of shells (cmd or powershell), `SYSTEMROOT` is impossible to change but we still need to fallback in case the process is started without env vars or it has been cleared in tauri users code.
- Inside a shell, powershell for instance, you can do `$env:SYSTEMROOT=""` to clear it but the reason we don't fallback here, is because we started the powershell session and we started it without a profile, so `$env:SYSTEMROOT` should be set correctly.

lucasfernog (approved):

PR #4889, opened 2022-08-08
fix(codegen): write output file when contents change

brian14708:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### Other information

Currently an app build is needed even when no code has changed. After some investigation, it was trigger by the writing of the temporary icon file used for embedding.

Even with this change, two builds are required to avoid further rebuilds. Maybe it's better to preprocess the icons in `build.rs`?

### DIFF
--- a/.changes/write-only-changed.md
+++ b/.changes/write-only-changed.md
@@ -1,0 +1,6 @@
+---
+"tauri-codegen": patch
+"tauri-build": patch
+---
+
+Only rewrite temporary icon files when the content change, avoid needless rebuilds.
--- a/core/tauri-build/src/lib.rs
+++ b/core/tauri-build/src/lib.rs
@@ -293,6 +293,10 @@ pub fn try_build(attributes: Attributes) -> Result<()> {
     )?;
   }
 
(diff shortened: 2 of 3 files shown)


lucasfernog (approved):
Clever one, thanks!

lucasfernog:
> Even with this change, two builds are required to avoid further rebuilds. Maybe it's better to preprocess the icons in build.rs?

We have a way to do this, enabling the `codegen` feature of tauri-build and changing your build.rs to:
```rust
fn main() {
  let mut codegen = tauri_build::CodegenContext::new();
  if !cfg!(feature = "custom-protocol") {
    codegen = codegen.dev();
  }
  codegen.build();
  tauri_build::build()
}
```

then change `tauri::generate_context!()` to `tauri::tauri_build_context!()`.

aaaaaaaactually this has a bug, i'll open a PR to fix it.

PR #4893, opened 2022-08-08
feat(cli): add `beforeBundleCommand`, closes #4879

### LINKED ISSUE #4879: [feat] add afterBuildCommand to execute command(s) before creating installer/package/bundle
### Describe the problem

The generated executable is quite large (I know about the `Reducing App Size` chapter), which results in a lot of disk space "wasted".

Out of the box on Windows the generated EXE-file results in nearly 7 MB. But when runtime-compressing the file with someting like https://github.com/upx/upx the generated EXE-file comes down to 2.2 MB (or 2 MB when running upx with `-9` parameter), without some lost functionality (at least what I can tell of).

In addition to compressing the generated files with external tools, this even would open the ability to do other wonderful things with the generated executable (e.g. generating hashsum-files, executing system-tests or other special build-pipeline-like stuff).

I would like to be able via the `tauri.conf.json` file to execute some single command or multiple commands after cargo did build the executable.

### Alternatives considered

Re-executing only the "creating installer/package/bundle" part via Tauri-CLI command after manually executing UPX. Would be better to have this part of tauri, as it reduces manual steps.

Maybe some additional `afterBundleCommand ` would be helpful which targets the phase after the installer/package/bundle was created. This could be used for auto-uploading the result to some server via custom-tailored scripts.

### Additional context

Maybe I am totally missing some built-in cargo-magic, as I am new to the Rust ecosystem. Previous I used to built some application using React and Electron, so this is quite a fascinating way to get into Rust.

amrbashir:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/tooling/cli/src/build.rs
+++ b/tooling/cli/src/build.rs
@@ -7,7 +7,8 @@ use crate::{
     app_paths::{app_dir, tauri_dir},
     command_env,
     config::{
-      get as get_config, AppUrl, BeforeBuildCommand, WindowUrl, MERGE_CONFIG_EXTENSION_NAME,
+      get as get_config, AppUrl, BeforeBuildCommand, BeforeBundleCommand, WindowUrl,
+      MERGE_CONFIG_EXTENSION_NAME,
     },
     updater_signature::{read_key_from_file, secret_key as updater_secret_key, sign_file},
   },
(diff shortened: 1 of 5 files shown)


tooling/cli/src/build.rs
```diff
@@ -198,6 +199,43 @@ pub fn command(mut options: Options) -> Result<()> {
 
   interface.build(interface_options)?;
 
+  if let Some(before_bundle) = config_.build.before_bundle_command.clone() {
```
lucasfernog: I think we should only run `before_bundle_command` after the `if config_.tauri.bundle.active {` block, and only if `package_types` isn't empty.

amrbashir: maybe `afterCargoBuild` is better?

lucasfernog: The hook is only needed because the bundler uses the output binary without letting the user modify it; an `afterCargoBuild` doesn't make sense IMO because if you need to modify the binary and you're not bundling it, you can just run your script after `tauri build`.

lucasfernog (approved):

PR #4914, opened 2022-08-11
refactor(app): run setup and window creation when event loop is ready

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/.changes/refactor-setup.md
+++ b/.changes/refactor-setup.md
@@ -1,0 +1,5 @@
+---
+"tauri": major
+---
+
+The window creation and setup hook are now called when the event loop is ready.
--- a/.changes/refactor-setup.md
+++ b/.changes/refactor-setup.md
@@ -2,4 +2,4 @@
 "tauri": major
 ---
 
-The window creation and setup hook are now called when the event loop is ready.
+**Breaking change:** The window creation and setup hook are now called when the event loop is ready.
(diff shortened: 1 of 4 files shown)


amrbashir:
Doesn't this make the setup hook obsolete? because I could just do the same logic in `RunEvent::Ready`

.changes/refactor-setup.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri": major
+---
+
+The window creation and setup hook are now called when the event loop is ready.
```
amrbashir:
```suggestion
[**Breaking Change**] The window creation and setup hook are now called when the event loop is ready.
```

lucasfernog:
> Doesn't this make the setup hook obsolete? because I could just do the same logic in `RunEvent::Ready`

It does, and we might remove all of these hooks in favor of the run closure, but for now I'll avoid basic breaking changes since we're actively developing mobile.

amrbashir:
I see, but this is still a breaking change and we should note that in the change file.

lucasfernog:
> I see, but this is still a breaking change and we should note that in the change file.

Yeah. This is the beginning of the `next` branch too :)

PR #4992, opened 2022-08-22
feat(cli): Add `icon` command (tauricon)

FabianLars:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist
- [x] I have added a convincing reason for adding this feature, if necessary -> tauricon.js is breaking apart rn.

### Other Information
I'll add some more info here after i got some sleep.

### DIFF
--- a/tooling/cli/Cargo.toml
+++ b/tooling/cli/Cargo.toml
@@ -64,6 +64,8 @@ ignore = "0.4"
 ctrlc = "3.2"
 log = { version = "0.4.17", features = [ "kv_unstable", "kv_unstable_std" ] }
 env_logger = "0.9.0"
+icns = { git = "https://github.com/FabianLars/rust-icns", rev = "bc2e492328cc5415a23a5eb6ec25632e8326f7bd" }
+image = { version = "0.24", default-features = false, features = [ "ico" ] }
 
 [target."cfg(windows)".dependencies]
 winapi = { version = "0.3", features = [ "handleapi", "processenv", "winbase", "wincon", "winnt" ] }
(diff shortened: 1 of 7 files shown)


tooling/cli/Cargo.toml
```diff
@@ -64,6 +64,8 @@ ignore = "0.4"
 ctrlc = "3.2"
 log = { version = "0.4.17", features = [ "kv_unstable", "kv_unstable_std" ] }
 env_logger = "0.9.0"
+icns = { git = "https://github.com/FabianLars/rust-icns", rev = "bc2e492328cc5415a23a5eb6ec25632e8326f7bd" }
```
lucasfernog: We'll need to either publish this crate or move it to a module inside cli.rs

FabianLars: Yes absolutely, that was mainly to show chip (and you) what I was talking about.
Aaand as discussed on discord I'm not really sure how to proceed with the crate (not happy with it for compressed pngs etc, but current state would be fine for 1.1)

lucasfernog: Let's go with `fine for 1.1` for now :P

FabianLars: I guess you'd prefer to have that crate part of the tauri org? If so i'd just move my fork to the org and let you handle the publishing 😂

lucasfernog: Yeah :D

FabianLars:
Marked as ready for review as i don't have any changes for 1.1 planned anymore.

Only the rust-icns crate release is missing, and some testing on macos would be nice.

_Edit:_ Oh and there's a TODO left about the input size, but we specified 1240 as a requirement for so long, i don't think it's really a problem to keep that for now, and it's not a _hard_ requirement anyway, other sizes will work fine too🤷

lucasfernog (approved):

PR #5030, opened 2022-08-24
feat(cli): use templates from wry

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


tooling/cli/src/mobile/mod.rs
```diff
@@ -147,6 +160,17 @@ fn get_config(config: &TauriConfig) -> (Config, Metadata) {
...
+  let last = s.clone().count() - 1;
+  let mut reverse_domain = String::new();
+  for (i, w) in s.enumerate() {
+    if i != last {
+      reverse_domain.push_str(w);
+      reverse_domain.push('.');
+    }
+  }
+  reverse_domain.pop();
+
```
amrbashir, line 178: had to do this today, so here goes my suggesstion 

```suggestion
  let s = config.tauri.bundle.identifier.split('.');
  let s: Vec<_> = s.split(".").collect();
  s.reverse();
  let reverse_domain = s.join("."); 
```

lucasfernog: in this case it's a little trickier since we need to ignore the last element.

amrbashir: ah didn't realize that, thought the bundle identifier would be just the domain

PR #5101, opened 2022-08-30
feat(cli): expose `TAURI_TARGET_TRIPLE` to before*Commands, closes #5091

### LINKED ISSUE #5091: [feat] beforeBuildCommand/beforeDevCommand needs a target triple environment variable
### Describe the problem

beforeBuildCommand and beforeDevCommand claim to be for conditional compilation, yet offer no way of getting the target triple for building

Expose the target triple as an environment variable

### Alternatives considered

_No response_

### Additional context

_No response_

amrbashir:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


tooling/cli/src/interface/rust.rs
```diff
@@ -229,17 +245,11 @@ impl Interface for Rust {
...
-      "windows" => {
-        env.insert("TAURI_PLATFORM_TYPE", "Windows_NT".into());
-      }
-      "darwin" => {
-        env.insert("TAURI_PLATFORM_TYPE", "Darwin".into());
-      }
-      _ => {}
-    }
+      "linux" => env.insert("TAURI_PLATFORM_TYPE", "Linux".into()),
+      "windows" => env.insert("TAURI_PLATFORM_TYPE", "Windows_NT".into()),
```
amrbashir, line 249: should this return `Windows` instead?

tooling/cli/src/interface/rust.rs
```diff
@@ -217,8 +217,24 @@ impl Interface for Rust {
...
+    env.insert(
+      "TAURI_PLATFORM",
+      match host {
+        // keeps compatibility with old `std::env::consts::OS` implementation
+        "darwin" => "macos".into(),
+        "ios-sim" => "ios".into(),
+        "androideabi" => "android".into(),
+        h => h.into(),
+      },
+    );
```
amrbashir, line 237: but the old implementation was just wrong and I expect `TAURI_ARCH` will be used in build tools so for example `darwin`, `androideabi` will be the correct value.

tooling/cli/src/interface/rust.rs
lucasfernog: the previous code returned Windows_NT so we can't change it

amrbashir: add a todo for v2 then?

tooling/cli/src/interface/rust.rs
lucasfernog: they can use the target triple for that. we can't break this, someone is probably using this (specially the platform one).

tooling/cli/src/interface/rust.rs
lucasfernog: sure

tooling/cli/src/interface/rust.rs
amrbashir: This would make sense for the platform but for the arch, I don't like it tbh and I think users should be aware that they are using the wrong arch.

amrbashir: tbh I see this as a bugfix and not breaking but lets add TODO for v2 then

lucasfernog: what's the difference between i686 and x86? it's the same information, and x86 is the format we were using already. I think you're mixing the arch and the platform.

amrbashir: one could be trying to download a sidecar (built with rust) at build time using the arch, and maybe they ended up converting `x86` to `i686` or `i586` (if someone still using that) but `x86` wasn't our intended result in the first place so this should be considered a fix because we are delivering the intended behavior even if it breaks their build script.

lucasfernog: they can easily map that back into i686 or use the target triple directly (which is the usual way of handling this).

amrbashir: they probably are but they can't map it back to i586 for example and they would need to use the target triple but that's beside the point. I simply don't want to wait until v2 to change this wrong behavior and maybe more people will end up depending on this wrong behavior. Also we are changing the tauri-cli which is just a tool and we are not breaking the framework.

lucasfernog (approved):

PR #5128, opened 2022-09-02
feat(tooling): use conditional type in open function

jianxingxuejian:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
the open function return null | string | string[] type, but just in the multiple option is true it will return string[], so i use conditional types to feat it.


lucasfernog:
Nice change, but it's a breaking one and we'll need to save it for v2 :(

Issue #4133, opened 2022-05-15
[feat] Getting a URL to resources (assets, media, videos, files, etc) without loading the entire file into memory; streaming playback


reivilibre:
### Describe the problem

I'd like to make an application that can play videos from the filesystem. In some cases, the videos might be larger than the available system memory, so we don't want to load them as a `Uint8Array`, `Blob` or anything like that.

My reading of the code first introduced in #2104 AFAICT and now living at https://github.com/tauri-apps/tauri/blob/3e702cf8b15762cdca43c8d7ff6f6e8ee9670244/core/tauri/src/manager.rs#L640-L651= suggests that, when using the `asset://` protocol (e.g. through `convertFileSrc`) the entire file will be read into memory as a `Vec<u8>` upon request (unless there's a `Range` header, in which case it will read part of the file).

### Describe the solution you'd like

Naïvely, I'd like the `asset://` protocol to stream the file from disk rather than buffer it all into memory.

It doesn't have to be the asset protocol; any mechanism where the Rust backend can expose a single file on the filesystem and have it streamed into a video tag on the page would be sufficient for my needs.

### Alternatives considered

A workaround might be to expose the file over HTTP and stream it using a `http://` URI. This kind of entirely bypasses Tauri though?

FabianLars:
Did you see our streaming example? https://github.com/tauri-apps/tauri/tree/dev/examples/streaming
Sounds exactly like what you're looking for. The problem with that one is that it doesn't work on linux because webkitgtk didn't support headers for custom protocols until the latest release, but it's not widespread enough to depend on it yet (also, in first tests it looked like it still doesn't work correctly iirc)

reivilibre:
I must admit I didn't come across the examples before, so thanks for linking that one and for mentioning the current status.

However I still don't think that's right.

When there's no `Range` header, it uses `read_to_end` as far as I can see, buffering the entire file into memory.
```rust
content.read_to_end(&mut buf)?;
```

My experience with streaming HTTP requests and video playback (specifically Firefox to be honest; if other browsers behave differently then I haven't had experience with those) is that the video will be played back in one HTTP connection unless the user seeks to a different position (at which point a new request with a `Range` header is sent).

In my experience, the server-side code usually reads 'small' chunks of the file into a buffer and writes them out to the stream whenever possible. I always imagined that TCP applies backpressure so that the server ends up writing only as much as the client is able to cope with; after all, neither client nor server are holding the entire file in memory.

[...]

PR #5136, opened 2022-09-05
feat: new js apis

### LINKED ISSUE #2233: [meta] Tracking issue for Node.js/Deno standard library inspired functions to include in `@tauri-apps/api`
I will be refactoring our js api and aiming to make the it more familiar to users coming from electron/js-land so it is easier to migrate. 

Some of current functions will be deprecated and removed when new variants are released.

We want to gather some feedback from users and which functions they want to add from the **Nodejs standard library**, so comment below and I will add them to this list of new functions:
1. `os` module
   - [x] `EOL`
   - [x] `platform()`
   - [x] `version()`
   - [x] `tempdir()`
   - [ ] `hostname`

2. `path` module
   - [x] `join()`
   - [x] `normalize()`
   - [x] `resolve()`
   - [x] `sep`
   - [x] `delimiter`
   - [x] `dirname()`
   - [x] `basename()`
   - [x] `extname()`
   - [x] `isAbsolute()`

3. `fs` module
	- [ ] `create()`
	- [ ] `open()`
    - [ ] `close()`
    - [ ] `FsFile` class
	- [ ] `writeFile()`
    - [ ] `writeTextFile()`
    - [ ] `readFile()`
    - [ ] `readTextFile()`
    - [ ] `copyFile()`
    - [ ] `readDir()`
    - [ ] `mkdir()`
    - [ ] `remove()`
    - [ ] `rename()`
    - [ ] `exists()`
 
4. `net` module
    - [ ] `createConnection()`
   
5. `dns` module

### LINKED ISSUE #4350: [bug] $platform-$arch id inconsistency
### Describe the bug

The platform id is inconsistent between the JS api and the launcher update json (and possibly elsewhere?)

For example as stated here: https://tauri.studio/v1/api/js/modules/os/  
the platform for windows, x86-64 should be `win32-x86_64`

But the platform id in the json of the tauri updater becomes `windows-x86_64`  
darwin/x86-64 also becomes an even weirder `darwin-intel`
https://docs.rs/tauri/1.0.0-rc.16/tauri/updater/index.html#update-file-json-format

Can we make this two consistent, to avoid confusion and manual translation between the two?

### Reproduction

_No response_

### Expected behavior

_No response_

### Platform and versions

```shell
1.0 rc16
```

### Stack trace

_No response_

### Additional context

_No response_

### LINKED ISSUE #4490: [feat] Universal Fetch API
### Describe the problem

Hey folks, I'm curious as to why Tauri does not opt-in for a kind of universal fetch polyfill (i.e. one that would call the browser `fetch` on the web, and Tauri's `http` module when running in Tauri). I'm attempting to build a hybrid application that I can deploy both to Tauri as well as a web browser (similar to how Notion's app supports both). I was surprised to learn that Tauri's `fetch` implementation is not compatible with the [function signature](https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts#L17373) of the browser `fetch`, resulting in me having to write a compatibility layer over it:

```typescript
export async function fetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
	const { fetch } = await import('@tauri-apps/api/http');

	const res = await fetch(input.toString(), {
		...init,
		method: init.method as HttpVerb,
		body: {
			type: 'Text',
			payload: init.body
		}
	});

	return new Response(JSON.stringify(res.data) as BodyInit, res);
}
```

Furthermore, I have to wrap calls to this fetch implementation in conditional statements that check if the app is running in Tauri. My app is in SvelteKit, and having to juggle calls to different `fetch` implementations from SvelteKit's `load` function, the browser client, and Tauri is all fairly cumbersome. 

As most of the issues I see on here regarding CORS and whatnot have the simple response of "just use Tauri's HTTP module", I think it could be beneficial to make the `fetch` API fully compatible with that of browsers, and automatically swap out the browser's `fetch` with Tauri's `fetch` when the app is built with the HTTP module enabled. 

I'm not sure if there's been a previous discussion around this, or if there were specific reasons the HTTP module has gone in the direction it did.

### Alternatives considered

At the very least, I think it could be good to make the `fetch` API compatible with that of browsers, even if a polyfill is not added automatically. This would prevent myself and other users from having to write shims to make TypeScript not complain. 

### Additional context

_No response_

### LINKED ISSUE #5432: [feat] http supports request progress and cancellation
### Describe the problem

Provides the way to capture request progress and cancel requests.

Similar to axios [progress capturing](https://github.com/axios/axios#-progress-capturing) and [cancellation](https://github.com/axios/axios#cancellation).

```js
import { http } from '@tauri-apps/api';

const controller = new AbortController();

const response = await http.fetch(downloadUrl, {
    method: 'GET',
    responseType: ResponseType.Binary,
    signal: controller.signal,
    onUploadProgress (e) {
        console.log(e.loaded);
        console.log(e.total);
    },
    onDownloadProgress (e) {
        console.log(e.loaded);
        console.log(e.total);
    }
});

controller.abort();
```

### Alternatives considered

_No response_

### Additional context

_No response_

amrbashir:
closes #2233 
closes #4490
closes #4350 

- [ ] Bugfix
- [X] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [X] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] Yes, and the changes were approved in issue #___
- [ ] No

### Checklist

### Other information

Although some of the APIs intended in this PR could be add in a minor version of v1, The majority will be breaking changes so it is better to just include all in one PR for v2. I will also try to split the PR into a reviewable chunks.

## TODOS:
1. `os` module
   - [X] Remove `type()`
   - [X] Adjust `platform()` to match Rust's implementation.
   - [X] `hostname()`
   - [X] Rename `tempdir()` to `tempDir()`

2. `path` module
	- [X] `tempDir()`

2. `fs` module
	- [x] `create()`
	- [x] `open()`
    - [x] `close()`
    - [x] `FsFile` class
    	- [X] `close()`
    	- [X] `read()`
    	- [X] `write()`
    	- [x] `seek()`
    	- [x] `truncate()`
    	- [x] `stat()`
    - [x] `write()`
	- [X] `writeFile()`
    - [X] `writeTextFile()`
    - [X] Remove `writeBinaryFile()`
    - [x] `read()`
    - [X] `readFile()`
    - [X] `readTextFile()`
    - [X] `readTextFileLines()`
    - [X] Remove `readBinaryFile()`
    - [x] Adjust `copyFile()`
    - [x] Adjust `readDir()`
    - [x] `mkdir()`
    - [x] `remove()`
    - [x] `rename()`
    - [x] `seek()`
    - [x] `truncate()`
    - [x] `ftruncate()`
    - [x] `stat()`
    - [x] `lstat()`
    - [x] `fstat()`
    
3. `http` module
	- [x] Make `fetch` compatible with the web api
	  - [ ] stream `file://` on `reqwest-client` feature
	  - [ ] cancel `fetch` using `AbortController`
	- [x] Remove `Client` and `getClient`

4. `window` module
	- [x] Remove `appWindow`
	- [x] Combine `WindowManager`, `WebviewHandle` and `WebViewWindow` into one `Window` class.
    
- [ ] Merge all change files into one big change file


.changes/api-rs-fs-disk-entry.md
```diff
@@ -0,0 +1,6 @@
+---
+"api": "major"
+"tauri": "major"
+---
+
+**Breaking change** Removed `DiskEtry` struct from `tauri::api::dir`. Use `DirEntry` instead.
```
vdvman1: Typo: `DiskEtry` should be `DiskEntry`

amrbashir: thank you, I will keep this in mind when I merge all changelogs together

tooling/api/src/http.ts
```diff
@@ -46,515 +46,76 @@
...
-  options?: FetchOptions
-): Promise<Response<T>> {
-  if (defaultClient === null) {
-    defaultClient = await getClient()
-  }
-  return defaultClient.request({
-    url,
-    method: options?.method ?? 'GET',
-    ...options
+  const res = new Response(Uint8Array.from(data), {
```
Yesterday17, line 110: It seems not suitable for responses with large body size.

FabianLars: The current implementation works the same, we currently can only send "normal" arrays to the backend, so unless the new IPC can support streaming bodies, this is what we get. :/

jlarmstrongiv:
Are `fs` streaming APIs planned?  See https://github.com/tauri-apps/tauri/issues/2233#issuecomment-1510266483 and https://github.com/tauri-apps/tauri/issues/4133

lucasfernog:
@amrbashir we'll need to find some time to move these to the plugins-workspace repo. I would wait a little longer on the FS APIs, since we'll see if [cap-std](https://github.com/bytecodealliance/cap-std) can help us there (see #6107).

amrbashir:
sure, I will start on porting them to plugins-workspace this week

PR #5139, opened 2022-09-05
Allow external URLs to access IPC commands

### LINKED ISSUE #5088: [feat] Inject window.__TAURI__ in allowed remote URLs
### Describe the problem

Currently it is not really possible to communicate directly with tauri from an external website. There is a [workaround](https://github.com/tauri-apps/tauri-plugin-localhost/commit/4126a4346b07f273b330e44bfe00af434d3f2cb2#diff-07aba1df28891af8108090f8fb0b27dd253b2601454d76e14c00e15307f0e879) if you have a static external URL you want to load, but it does not work if you need to inject in arbitrary URLs.

In our case, we have a first window where the user can manage their connection. They can then connect to a remote server which provides the UI. That UI needs to communicate with tauri to store credentials on the user machine, access notifications, etc.

Basically I would like a setting in the `WindowBuilder` to allow injection of the `window.__TAURI__` in any website loaded in that window. I do understand that it can be security risk if people are not careful so that could be gated with a feature flag or a general configuration option that needs to be enabled (in the context for example) before the setting has any effect.

### Alternatives considered

We are currently investigating our options:
- Create a local server so the remote url can talk to it and in turn execute stuff in the rust context
- Use a static wrapper page with an iframe that loads the remote URL so that the static page can relay the messages from the iframe to tauri

Neither are very elegant solutions and I would prefer to be able to communicate directly with tauri.

### Additional context

_No response_

Sytten:
This is very early and likely won't be merged, but this is a possible way to fix #5088

### DIFF
--- a/core/tauri-utils/Cargo.toml
+++ b/core/tauri-utils/Cargo.toml
@@ -32,7 +32,6 @@ json5 = { version = "0.4", optional = true }
 toml = { version = "0.5", optional = true }
 json-patch = "0.2"
 glob = { version = "0.3.0", optional = true }
-globset = { version = "0.4.9", default-features = false, features = [ "serde1" ]}
 walkdir = { version = "2", optional = true }
 memchr = "2.4"
 semver = "1"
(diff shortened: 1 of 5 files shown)


lucasfernog:
Thanks for your contribution! I've extended this on #5918.

PR #5146, opened 2022-09-06
feat(cli/mobile): allow checking `gen` folder into source control

amrbashir:

- [ ] Bugfix
- [X] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/tooling/cli/src/helpers/template.rs
+++ b/tooling/cli/src/helpers/template.rs
@@ -80,13 +80,14 @@ pub fn render_with_generator<
         file_path.set_extension("toml");
       }
     }
-    let mut output_file = out_file_generator(&file_path)?;
-    if let Some(utf8) = file.contents_utf8() {
-      handlebars
-        .render_template_to_write(utf8, &data, &mut output_file)
-        .expect("Failed to render template");
-    } else {
-      output_file.write_all(file.contents())?;
+    if let Ok(mut output_file) = out_file_generator(&file_path) {
+      if let Some(utf8) = file.contents_utf8() {
+        handlebars
+          .render_template_to_write(utf8, &data, &mut output_file)
+          .expect("Failed to render template");
+      } else {
+        output_file.write_all(file.contents())?;
+      }
     }
   }
   for dir in dir.dirs() {
--- a/tooling/cli/src/helpers/template.rs
+++ b/tooling/cli/src/helpers/template.rs
@@ -54,7 +54,7 @@ pub fn render<P: AsRef<Path>, D: Serialize>(
         create_dir_all(&parent)?;
         created_dirs.push(parent);
       }
-      File::create(path)
+      File::create(path).map(Some)
     },
   )
 }
@@ -62,7 +62,7 @@ pub fn render<P: AsRef<Path>, D: Serialize>(
 pub fn render_with_generator<
   P: AsRef<Path>,
   D: Serialize,
-  F: FnMut(&PathBuf) -> std::io::Result<File>,
+  F: FnMut(&PathBuf) -> std::io::Result<Option<File>>,
 >(
   handlebars: &Handlebars<'_>,
   data: &D,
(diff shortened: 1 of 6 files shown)


tooling/cli/src/helpers/template.rs
```diff
@@ -80,13 +80,14 @@ pub fn render_with_generator<
         file_path.set_extension("toml");
       }
     }
-    let mut output_file = out_file_generator(&file_path)?;
```
lucasfernog, line 83: cargo-mobile handles this with the Filter enum instead of ignoring errors here, I think that would be a better fit

for instance this would ignore permission errors and the user might not know what's wrong in their template

amrbashir: do you want the same mechanism? as in read the gitignore file and depending on that, overwrite? because that would mean it won't override any files

amrbashir: I could change the `out_file_generator` to return `Result<Option<File>>` and skip if `None`

lucasfernog: that works for now. I think parsing gitignore is a bit too much

tooling/cli/src/mobile/android/project.rs
```diff
@@ -122,14 +122,21 @@ pub fn gen(
...
         options.mode(0o755);
       }
 
-      options.create_new(true).write(true).open(path)
+      if path.file_name().unwrap() == OsStr::new("BuildTask.kt") {
+        options.create(true).open(path).map(Some)
+      } else if path.exists() {
+        Ok(None)
+      } else {
+        options.create(true).open(path).map(Some)
```
lucasfernog: this is the same as the block for the first `if`, can't you just change it to 
```rust
if path.file_name().unwrap() == OsStr::new("BuildTask.kt") || !path.exists() {
  options.create(true).open(path).map(Some)
} else {
  Ok(None)
}
```

amrbashir: yeah that works

amrbashir:
```suggestion
      if path.file_name().unwrap() == OsStr::new("BuildTask.kt") || !path.exists() {
        options.create(true).open(path).map(Some)
      } else {
        Ok(None)
      ```

File: tooling/cli/src/mobile/ios/project.rs [hunk 3]
```diff
@@ -136,7 +136,16 @@ pub fn gen(
...
 
-      File::create(path)
+      let mut options = OpenOptions::new();
+      options.write(true);
+
+      if path.file_name().unwrap() == OsStr::new("project.yml") {
+        options.create(true).open(path).map(Some)
+      } else if path.exists() {
+        Ok(None)
+      } else {
```
amrbashir:
```suggestion
      if path.file_name().unwrap() == OsStr::new("BuildTask.kt") || !path.exists() {
        options.create(true).open(path).map(Some)
      } else {
        Ok(None)
```

tooling/cli/src/mobile/android/project.rs
```diff
@@ -122,14 +122,20 @@ pub fn gen(
...
         use std::os::unix::fs::OpenOptionsExt;
         options.mode(0o755);
       }
 
-      options.create_new(true).write(true).open(path)
+      if path.file_name().unwrap() == OsStr::new("BuildTask.kt") || !path.exists() {
+        options.create(true).open(path).map(Some)
+      } else {
+        Ok(None)
+      ```
```
amrbashir:
```suggestion
```

tooling/cli/src/mobile/ios/project.rs
```diff
@@ -136,7 +136,15 @@ pub fn gen(
...
 
-      File::create(path)
+      let mut options = OpenOptions::new();
+      options.write(true);
+
+      if path.file_name().unwrap() == OsStr::new("BuildTask.kt") || !path.exists() {
+        options.create(true).open(path).map(Some)
+      } else {
+        Ok(None)
+        options.create(true).open(path).map(Some)
```
amrbashir:
```suggestion
```

PR #5169, opened 2022-09-10
chore(api.js): add `@since` tag where appropriate

amrbashir:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/tooling/api/package.json
+++ b/tooling/api/package.json
@@ -67,6 +67,7 @@
     "tslib": "2.4.0",
     "typedoc": "0.23.14",
     "typedoc-plugin-markdown": "3.13.6",
+    "typedoc-plugin-mdn-links": "^2.0.0",
     "typescript": "4.8.2"
   }
 }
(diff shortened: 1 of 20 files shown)


amrbashir:
After this we should follow the convention to add `@since` tag when a new function, enum, interface, class is added.

When we add a class methods, class property, interface fields, an enum variant, it should have `@since` tag for the this field only. Other fields or methods which exist since their parent (class, interface ...etc) shouldn't have `@since` tag.

lucasfernog:
Looks good but why did you remove some @param tags?

amrbashir:
I removed some `@param` and `@returns` since they were either empty or wasn't needed since the function description/signature was good enough. I did that to declutter the generated documentation and source code.

lucasfernog (approved):

PR #5172, opened 2022-09-11
Remove `futures` & `futures-lite` and use `futures-util` directily.

Berrysoft:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: cleanup dependencies

- [x] No

### Checklist

### Other information
`futures` crate brings an executor by default, while we have already used `tokio`. `futures-lite` brings another executor here, but we only need *one* method from it, which could be simply replaced with `Pin::new`. This cleanup will make `tauri` build a little faster.


lucasfernog (approved):
Nice one, thanks!

PR #5201, opened 2022-09-15
feat(api.js): added option to remove line endings on readTextFile

thegame4craft:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/tooling/api/src/fs.ts
+++ b/tooling/api/src/fs.ts
@@ -71,6 +71,7 @@
  */
 
 import { invokeTauriCommand } from './helpers/tauri'
+import { EOL } from './os';
 
 /**
  * @since 1.0.0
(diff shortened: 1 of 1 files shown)


amrbashir (commented):
I don't think I like this tbh, this function should be for reading only, any further manipulation to the content should be done in user-land.

lucasfernog:
> I don't think I like this tbh, this function should be for reading only, any further manipulation to the content should be done in user-land.

I agree.

amrbashir:
thanks @thegame4craft for the PR but I am afraid we can't accept it for the reason I mentioned above.

PR #5217, opened 2022-09-16
fix(ci): add inputs to publish-cli-js

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: CI fix

- [ ] No

### Checklist

### DIFF
--- a/.github/workflows/covector-version-or-publish.yml
+++ b/.github/workflows/covector-version-or-publish.yml
@@ -171,7 +171,7 @@ jobs:
           token: ${{ secrets.TAURI_BOT_PAT }}
           repository: tauri-apps/tauri
           event-type: publish-clijs
-          client-payload: '{"releaseId": "${{ steps.covector.outputs.cli.js-releaseId }}" }'
+          inputs: '{"releaseId": "${{ steps.covector.outputs.cli.js-releaseId }}" }'
 
       - name: Trigger cli.rs publishing workflow
         if: |
--- a/.github/workflows/publish-cli-js.yml
+++ b/.github/workflows/publish-cli-js.yml
@@ -5,6 +5,10 @@ env:
   MACOSX_DEPLOYMENT_TARGET: '10.13'
 on:
   workflow_dispatch:
+    inputs:
+      releaseId:
+        description: 'ID of the cli.js release'
+        required: true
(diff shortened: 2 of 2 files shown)


.github/workflows/covector-version-or-publish.yml
```diff
@@ -171,7 +171,7 @@ jobs:
           token: ${{ secrets.TAURI_BOT_PAT }}
           repository: tauri-apps/tauri
           event-type: publish-clijs
-          client-payload: '{"releaseId": "${{ steps.covector.outputs.cli.js-releaseId }}" }'
+          inputs: '{"releaseId": "${{ steps.covector.outputs.cli.js-releaseId }}" }'
```
amrbashir, line 174: note sure if `inputs` work since the action documentation doesn't mention it but we could skip this and instead, get the version from `metadata.json` like `publish-cli-rs.yml`

amrbashir: I just realized we need the releaseId not the cli version. Anyways, we can just do the following in `publish-cli-js.yml` and keep this as is because I can't find documentation for `inputs` in the repository-dispatch action
```yml
 RELEASE_ID: ${{ github.event.client_payload.releaseId || inputs.releaseId }}
```

lucasfernog: Well I didn't test this one with the repository-dispatch yet, but the original client_payload you wrote doesn't work: https://github.com/tauri-apps/tauri/actions/runs/3069902936/jobs/4959516330

amrbashir: I am not sure tbh why it failed bth but I have the same setup in CTA and it works

amrbashir: I tracked it down to this https://github.com/tauri-apps/tauri/actions/runs/3069615135/jobs/4958749753#step:9:6 the `releaseId` is set 
empty before dispatching

amrbashir: I think the problem is in the `.` of `cli.js`

PR #5246, opened 2022-09-20
feat(cli): add option to make custom icon sizes, closes #5121

### LINKED ISSUE #5121: [Feat] tauricon: Allow for additional custom generated icon files.
I have the use-case where I serve my front-end as a tauri app and also as a web-app with PWA support. Instead of maintaining two sets of icons, I would like to just generate one set that is used by everything. This is probably a pretty common use-case.

So it would be nice to have a way to define alternative icon formats, sizes and file-names, that are also generated by `tauricon`.

Currently, the files to be generated are defined here: https://github.com/tauri-apps/tauricon/blob/dev/src/helpers/tauricon.config.ts

Probably the easiest implementation would be to keep that same structure and provide it via the cli as a json file:
```sh
npx @tauri-apps/tauricon icons/logo.svg --target icons --extra icons/extra.json
```
And then the `icons/extra.json` could be something like this:
```json
[
  {
    "folder": "../frontend/src/assets/icons",
    "prefix": "icon-",
    "infix": true,
    "suffix": ".png",
    "sizes": [72, 96, 128, 144, 152, 192, 384, 512]
  }
]
```

What are your thoughts about that?

khuongduy354:

- [ ] Bugfix
- [X] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] No

### Checklist


khuongduy354:
Feature requested in [this issue](https://github.com/tauri-apps/tauri/issues/5121). Some keynotes:
- Only apply to .ico and .png
- Added --config to find json config file. 
- If no --config provided, use default hardcoded values, as in previous version

FabianLars:
I don't think there is any value in making it possible to generate custom .ico files. Just like the .icns file it's basically "complete" and there shouldn't be an actual use case requiring different layer sizes.

It's a good start btw! :)

i'll wait with the code nitpicking until it's more feature complete, but one thing already, you can use serde_json::from_reader instead of from_slice. It's a little cleaner and probablyyy faster.

FabianLars:
Hello 👋 
Sorry for the super long delay, it thought you weren't done and then forgot about it after some time :/

Do you still have interest in working on it or should we take it from here?

A few open points i see at first glance:
- It still supports ICO files. As said earlier i don't think this offers much value because the default ico is already "complete"
- There's still a todo for the file name, but that's imo not important because of my next point
- I don't really like the json approach anymore tbh. I think we should go with a more direct cli interaction. What i have in mind right now would be a cli argument that takes an array of sizes, for example `tauri icon --extra 123,420,1337`, which just reuses the existing png loop, including the default naming scheme. I think letting the users rename the files if needed is not too much to ask for.

khuongduy354:
I was kinda tight on schedule at moment, but i think i can work on it this
weekend, you can leave it to me.

Vào 22:04, T.5, 8 Th12, 2022 Fabian-Lars ***@***.***> đã viết:

khuongduy354:
I made some changes, can you have a look? @FabianLars . It remove json config, and --extra flags generate custom sizes .png with same naming scheme. 
Should the --extra generate **additional** besides default png sizes, or it should ignore default ones?

FabianLars:
> Should the --extra generate additional besides default png sizes, or it should ignore default ones?

In my opinion it should create _additional_ files, and keep building the default at the same time.

khuongduy354:
How is it? If you decide to take it, I'll cleanup the comments and test icon files.
@FabianLars

FabianLars:
Yeah i like that. Another thing we may wanna look into is de-duplication so that we don't end up generating the same icon twice - that can be added after the cleanup ofc.

tooling/cli/src/icon.rs
```diff
@@ -41,15 +41,20 @@ pub struct Options {
...
+  #[clap(short, long, use_delimiter = true)]
+  extra: Option<Vec<u32>>,
 }
 
 pub fn command(options: Options) -> Result<()> {
   let input = options.input;
   let out_dir = options.output.unwrap_or_else(|| tauri_dir().join("icons"));
+  let extra_icon_sizes = options
+    .extra
+    .unwrap_or_else(|| vec![32, 16, 24, 48, 64, 256]);
```
FabianLars:
```suggestion
    .unwrap_or_default();
```
sending an empty array is fine. at this point it may be worth to just send the option and handle this in `fn png` instead so that one can get rid of a potentially unused vec.

tooling/cli/src/icon.rs
```diff
@@ -124,7 +129,10 @@ fn ico(source: &DynamicImage, out_dir: &Path) -> Result<()> {
   log::info!(action = "ICO"; "Creating icon.ico");
   let mut frames = Vec::new();
 
-  for size in [32, 16, 24, 48, 64, 256] {
+  let sizes = vec![32, 16, 24, 48, 64, 256];
+  let icon_name = "icon.ico".to_string();
+
+  for size in sizes {
```
FabianLars:
```suggestion
  for size in [32, 16, 24, 48, 64, 256] {
```

tooling/cli/src/icon.rs
```diff
@@ -144,7 +152,7 @@ fn ico(source: &DynamicImage, out_dir: &Path) -> Result<()> {
     }
   }
 
-  let mut out_file = BufWriter::new(File::create(out_dir.join("icon.ico"))?);
+  let mut out_file = BufWriter::new(File::create(out_dir.join(icon_name))?);
```
FabianLars:
```suggestion
  let mut out_file = BufWriter::new(File::create(out_dir.join("icon.ico"))?);
```

tooling/cli/src/helpers/config.json
```diff
@@ -0,0 +1,10 @@
+{
+  "ico": {
+    "name": "test.ico",
+    "sizes": [32, 16, 24, 48, 64, 256]
+  },
+  "png": {
+    "name": "icon.png",
+    "sizes": [32, 128, 256, 512]
+  }
+}
```
FabianLars: file is not needed anymore

tooling/cli/src/icon.rs
```diff
@@ -154,8 +162,16 @@ fn ico(source: &DynamicImage, out_dir: &Path) -> Result<()> {
...
-fn png(source: &DynamicImage, out_dir: &Path) -> Result<()> {
-  for size in [32, 128, 256, 512] {
+fn png(source: &DynamicImage, out_dir: &Path, mut extra_icon_sizes: Vec<u32>) -> Result<()> {
+  //remove duplicate extra icons
+  let mut sizes = vec![32, 16, 24, 48, 64, 256];
+  extra_icon_sizes = extra_icon_sizes
+    .into_iter()
+    .filter(|x| !sizes.contains(x))
+    .collect::<Vec<u32>>();
+  sizes.append(extra_icon_sizes.as_mut());
```
FabianLars: it may be nicer to do something like this
```rs
  sizes.append(extra_icon_sizes.as_mut());
  sizes.sort_unstable();
  sizes.dedup();
```
what do you think?

tooling/cli/src/icon.rs
```diff
@@ -154,8 +162,16 @@ fn ico(source: &DynamicImage, out_dir: &Path) -> Result<()> {
 
 // Generate .png files in 32x32, 128x128, 256x256, 512x512 (icon.png)
 // Main target: Linux
-fn png(source: &DynamicImage, out_dir: &Path) -> Result<()> {
-  for size in [32, 128, 256, 512] {
+fn png(source: &DynamicImage, out_dir: &Path, mut extra_icon_sizes: Vec<u32>) -> Result<()> {
+  //remove duplicate extra icons
+  let mut sizes = vec![32, 16, 24, 48, 64, 256];
```
FabianLars:
```suggestion
  let mut sizes = vec![32, 128, 256, 512];
```

FabianLars (changes requested):
Thanks so far! Only some small-ish things left. On top of what's mentioned in the comments we also need a changefile, example: https://github.com/tauri-apps/tauri/blob/dev/.changes/is-minimized.md

tooling/cli/src/icon.rs
khuongduy354: - I think .filter() takes linear time. Idk how much it cost to cast from vec to iterator and back. 
- Sorting may take longer, and extra cost on dedup()

FabianLars: i don't think performance on a 5 entry vec matters, it's only about the esthetics :D

really doesn't matter, it was just a random thought.

khuongduy354: Users may input tons of extra sizes, but that may not happen :D . It's just my habit for best practice.

FabianLars (dismissed):
imo it's good as-is, thanks!

2 potential changes/improvements we may wanna think about anyway:
- make the extra_icon_sizes arg an `Option` so that all the new code is basically skipped for most users
- If we want to make users able to generate 256 and 512 icons with the default name (_on top_ of the 2 hard coded names), we'd have to hard code the creation of the 128x128@2x.png and icon.png icons outside the for loop.
At this point we could get rid of the whole dedup logic by also hard coding the removal of 32px and 128px icons from the extra icons vec 🤷

khuongduy354:
I don't really get making extra_icon_sizes as `Option`.

khuongduy354:
I altered the icon sizes, as you said, it removes the dedup logic, looks better to me. Do u want it or switch back to the old commit ?

FabianLars:
> I don't really get making extra_icon_sizes as Option.

Doesn't really matter anymore. Without the dedup logic it would add more visual clutter than its worth.

One last thing: The 32 and 128 icon sizes are missing now.

khuongduy354:
So it's done after switching back to prev commit ?

FabianLars:
That would introduce the 256 / 512 naming problem from earlier. Honestly just do something like this again
```rs
for size in [32, 128].into_iter().chain(extra_icon_sizes) {}
```
and call it a day. We had so many iterations that i'm not sure anymore what i'd like to see at this point 😅 

Thanks for following through btw!

FabianLars (dismissed):

FabianLars:
Sorry that it was such a pain. I'm quite new to this PR reviewing stuff x)

Thank you again!

FabianLars (dismissed):

khuongduy354:
Thx a bunch, Im a newbie and was busy recently, so i cant focus entirely on this, it such a pain. Anyways, thanks a lot for your guide

lucasfernog:
LGTM, but maybe we should change it to only run the user-specified icon sizes? Like instead of running `tauri icon --extra-png 24 48` to generate all Tauri icons + the extra ones, we could let the user run `tauri icon --png 24 48 --output ./assets/img`.

FabianLars:
That was indeed the initial idea from the others, but i thought it feels a bit weird (idk why) so i asked to start with this approach first and wait for other opinions

lucasfernog:
I've pushed that. We have time to get feedback and revert it if it feels weird :) we still need to audit the next release anyway.

lucasfernog (approved):

PR #5272, opened 2022-09-23
feat(api): add app-specific directory APIs, closes #5263

### LINKED ISSUE #5263: [feat] Introduce a appDataDir path
### Describe the problem

I am building an application using Tauri in which I would like to store some data in a file on the hard drive, so I was looking into the `fs` module. For available paths I have found [this documentation](https://tauri.app/v1/api/js/path/). I feel like the correct path for storing my data on a linux system would be the `dataDir` path. However, this path only contains the path to the folder, without a folder for the application itself as in `appDir`, therefore I have to handle that on my own, which is a tedious task. Additionally I think that this is so common, that it would warrant another path preset.

I would like to be able to use something like `BaseDirectory.AppData`, that stores data in a folder for my own application within `dataDir`.

### Alternatives considered

No alternatives other than leaving it the way it is and handle that in the application itself come to my mind.

### Additional context

_No response_

caesar:
I'd appreciate some feedback on the approach taken here as it's my first time working with the Tauri codebase.

In particular, should there be tests? I can't find tests anywhere for similar API features such as the existing `appDir` method.

Also, are you happy with the way I have refactored `appDir` and `logDir` to internally call `appConfigDir` and `appLogDir` to avoid duplicated code?

I also wondered if there's an equivalent to `@since` for the Rust documentation comments?

- [ ] Bugfix
- [x] Feature
- [x] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
closes #5263


tooling/api/src/path.ts
```diff
@@ -28,23 +28,76 @@ import { BaseDirectory } from './fs'
 import { isWindows } from './helpers/os-check'
 
 /**
- * Returns the path to the suggested directory for your app config files.
+ * Alias of {@link appConfigDir} for backwards-compatibility purposes.
+ *
+ * @since 1.0.0
+ */
```
amrbashir: keep the original doc comment here and add `@deprecation` notice saying this will be removed in v2 and suggest using the new function

tooling/api/src/path.ts
```diff
@@ -529,7 +582,16 @@ async function videoDir(): Promise<string> {
 }
 
 /**
- * Returns the path to the suggested log directory.
+ * Alias of {@link appLogDir} for backwards-compatibility purposes.
+ *
+ * @since 1.0.0
+ */
```
amrbashir: keep the original doc comment here and add `@deprecation` notice saying this will be removed in v2 and suggest using the new function

core/tauri/src/api/path.rs
```diff
@@ -57,16 +57,26 @@ pub enum BaseDirectory {
...
   /// The Resource directory.
   Resource,
-  /// The default App config directory.
-  /// Resolves to [`BaseDirectory::Config`].
+  /// Alias of [`BaseDirectory::AppConfig`]
   App,
-  /// The Log directory.
-  /// Resolves to `BaseDirectory::Home/Library/Logs/{bundle_identifier}` on macOS
-  /// and `BaseDirectory::Config/{bundle_identifier}/logs` on linux and windows.
+  /// Alias of [`BaseDirectory::AppLog`]
```
amrbashir: Keep the doc comments and add a deprecation note https://doc.rust-lang.org/reference/attributes/diagnostics.html?highlight=deprecate#the-deprecated-attribute saying it will be removed in v2 and suggest the new variant

core/tauri/src/api/path.rs
```diff
@@ -491,3 +531,13 @@ pub fn log_dir(config: &Config) -> Option<PathBuf> {
...
+
+/// Alias of [`app_config_dir`] for backwards-compatibility purposes.
+pub fn app_dir(config: &Config) -> Option<PathBuf> {
+  app_config_dir(config)
+}
+
+/// Alias of [`app_log_dir`] for backwards-compatibility purposes.
+pub fn log_dir(config: &Config) -> Option<PathBuf> {
+  app_log_dir(config)
+}
```
amrbashir: Keep the doc comments and add a deprecation note https://doc.rust-lang.org/reference/attributes/diagnostics.html?highlight=deprecate#the-deprecated-attribute saying it will be removed in v2 and suggest the new function

core/tauri/src/app.rs
```diff
@@ -288,14 +288,34 @@ impl PathResolver {
...
-    crate::api::path::app_dir(&self.config)
+    self.app_config_dir()
   }
 
-  /// Returns the path to the suggested log directory.
+  /// Alias of [`PathResolver::app_log_dir`] for backwards-compatibility purposes.
   pub fn log_dir(&self) -> Option<PathBuf> {
-    crate::api::path::log_dir(&self.config)
+    self.app_log_dir()
   }
```
amrbashir: Keep the doc comments and add a deprecation note https://doc.rust-lang.org/reference/attributes/diagnostics.html?highlight=deprecate#the-deprecated-attribute saying it will be removed in v2 and suggest the new function

tooling/api/src/fs.ts
```diff
@@ -41,21 +41,22 @@
...
- *         "scope": ["$APP/databases/*"]
+ *         "scope": ["$APPDATA/databases/*"]
  *       }
  *     }
  *   }
  * }
  * ```
  *
- * Notice the use of the `$APP` variable. The value is injected at runtime, resolving to the {@link path.appDir | app directory}.
+ * Notice the use of the `$APPDATA` variable. The value is injected at runtime, resolving to the {@link path.appDataDir | app data directory}.
```
amrbashir, line 57: you should also do the same in `core\tauri-utils\src\config.rs`

amrbashir (changes requested):
> In particular, should there be tests? I can't find tests anywhere for similar API features such as the existing `appDir` method.

It is fine, these functions don't have tests.
 
> Also, are you happy with the way I have refactored `appDir` and `logDir` to internally call `appConfigDir` and `appLogDir` to avoid duplicated code?

Yeah that's perfect.

> I also wondered if there's an equivalent to `@since` for the Rust documentation comments?

I don't think there is but it is not needed. We are using `@since` in the js docs because we on

FabianLars:
Another thing we need to think about is localdata vs data because of windows :/

amrbashir:
yeah we could also add `appLocalDataDir`

caesar:
Thanks for the review @amrbashir, I'll make those changes later. I was actually going to ask about deprecation notices too but forgot, so thanks for mentioning that.

I'll look at `appLocalDataDir` too.

tooling/api/src/fs.ts
caesar: Ah yes, good catch, I forgot to look for these references on the Rust side

caesar:
I think I've addressed all the issues discussed, and also fixed the lint error. Would you mind taking another look @amrbashir ?

core/tauri/src/api/path.rs
```diff
@@ -13,61 +13,93 @@ use crate::{Config, Env, PackageInfo};
...
-  App,
-  /// The Log directory.
-  /// Resolves to `BaseDirectory::Home/Library/Logs/{bundle_identifier}` on macOS
-  /// and `BaseDirectory::Config/{bundle_identifier}/logs` on linux and windows.
-  Log,
-  /// A temporary directory.
-  /// Resolves to [`temp_dir`].
-  Temp,
+// we have to wrap the BaseDirectory enum in a module for #[allow(deprecated)]
+// to work, because the procedural macros on the enum prevent it from working directly
```
amrbashir, line 17:
```suggestion
// to work, because the procedural macros on the enum prevent it from working directly
// TODO: remove this workaround in v2 along with deprecated variants
```

amrbashir (changes requested):
Also missing change file, should be a `minor` bump

.changes/app-dirs-api.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri": minor
+---
+
+Add new app-specific directory APIs and deprecate existing `app_dir` and `log_dir` APIs.
```
amrbashir:
```suggestion
---
"tauri": minor
"api": minor
---

Add new app-specific `BaseDirectory` enum variants `AppConfig`, `AppData`, `AppLocalData`, `AppCache` and `AppLog` along with equivalent functions in `path` module and deprecated ambiguous variants `Log` and `App` along with their equivalent functions in `path` module.
```

amrbashir (changes requested):

.changes/app-dirs-api.md
caesar: I considered adding all the details but wasn't sure if the message could be that long. Good to know it can be!

amrbashir (dismissed):

lucasfernog (dismissed):

lucasfernog (approved):

PR #5311, opened 2022-09-30
docs: clarify that `focus` doesn't work atm, closes #5120

### LINKED ISSUE #5120: [bug] Window loses focus when creating another window with `focus: false`, possible deadlock when re-focusing main window manually
### Describe the bug

We want to be able to create a second window on the onFocus event in a React input component, while keeping the same input focused. 

Seems like the intended way of doing this is setting the `focus` property to `false`, but this doesn't seem to do the trick.

We also tried using `setFocus()` inside the "tauri://created" listener to re-focus the primary window manually but this seems to cause a deadlock, as the cursor in the input is blinking but the first button pressed does not register.

### Reproduction

1. Go to repo https://github.com/Andrepav1/tauri-multi-window
2. Run `yarn` && `yarn start`
3. Run `tauri dev`
4. Focus input component and wait window creation
5. Press any key

### Expected behavior

Create the window, while keeping the focus on the input component.

### Platform and versions

Environment
  › OS: Windows 10.0.19044 X64
  › Webview2: 104.0.1293.70
  › MSVC: 
  › Node.js: 16.15.0
  › npm: 8.5.5
  › pnpm: 7.1.0
  › yarn: 1.22.15
  › rustup: 1.25.1
  › rustc: 1.62.1
  › cargo: 1.62.1

Packages
  › @tauri-apps/cli [NPM]: 1.0.5
  › @tauri-apps/api [NPM]: 1.0.2
  › tauri [RUST]: 1.0.5,
  › tauri-build [RUST]: 1.0.4,
  › tao [RUST]: 0.12.2,
  › wry [RUST]: 0.19.0,

App
  › build-type: bundle
  › CSP: default-src blob: data: filesystem: ws: http: https: 'unsafe-eval' 'unsafe-inline' 'self' img-src: 'self'
  › distDir: ../build
  › framework: React

App directory structure
  ├─ .git
  ├─ node_modules
  ├─ public
  ├─ src
  └─ src-tauri

### Stack trace

```shell
no stack trace
```

### Additional context

_No response_

amrbashir:

- [ ] Bugfix
- [ ] Feature
- [x] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/core/tauri-utils/src/config.rs
+++ b/core/tauri-utils/src/config.rs
@@ -834,6 +834,8 @@ pub struct WindowConfig {
   #[serde(default)]
   pub fullscreen: bool,
   /// Whether the window will be initially hidden or focused.
+  ///
+  /// **This doesn't work at the moment and the window will always be initially focused**
   #[serde(default = "default_focus")]
   pub focus: bool,
   /// Whether the window is transparent or not.
--- a/core/tauri/src/window.rs
+++ b/core/tauri/src/window.rs
@@ -326,7 +326,9 @@ impl<'a, R: Runtime> WindowBuilder<'a, R> {
     self
   }
 
-  /// Whether the window will be initially hidden or focused.
+  /// Whether the window will be initially focused or not.
+  ///
+  /// **This doesn't work at the moment and the window will always be initially focused**
   #[must_use]
   pub fn focus(mut self) -> Self {
     self.window_builder = self.window_builder.focus();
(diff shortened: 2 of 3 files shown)


lucasfernog:
What do we need to make it work?

amrbashir:
It needs to be implemented on the tao window builder, I know how to do it on Windows but not on Linux and macOS. I will research this tonight.

FabianLars:
Sounds like something that won't work on Linux... (Or at least not on Wayland)
Good luck :D

amrbashir:
I found a way for it but I never give Linux high hopes about this kind of stuff, but I will add it anyways.

PR #5341, opened 2022-10-03
refactor(core): add allowlist for window creation options

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


FabianLars:
Is it just me or is adding a flag for every single option is a little extreme?

This way everyone would probably just opt for `"all": true` again.

lucasfernog:
I was discussing that with @tweidinger

tweidinger:
> Is it just me or is adding a flag for every single option is a little extreme?
> 
> This way everyone would probably just opt for `"all": true` again.

Yep, this is not an ideal solution but better than the current state (where every additional window could possibly be abused for things like global clipboard access without honoring the main window permissions). I also mentioned that it could be nice to add a flag for permission inheritance, so it is possible to configure it but not necessarily needed.

My quote from the chat on this pr:
> I think thats the straightforward way for it currently👍 
Offtopic: It gets more interesting when thinking about window specific allowlists. Something like a toggle for allowlist inheritance `inheritWindowOptions`  or allowlist templates `windowChildrenOptions`  could be options in the future.

PR #5381, opened 2022-10-10
feat: allow setting window  size and position using physical units close #5228

### LINKED ISSUE #5228: [feat] Window builder physical size/position
### Describe the problem

There's no way to set the size and position of the window in the window builder to physical coordinates.

The size and position methods should be able to accept either logical or physical units.

### Alternatives considered

Setting the window size and position AFTER the window has been created, which for some reason accepts logical or physical coordinates despite the fact that the window builder doesn't? 🤔

### Additional context

My use case is positioning a window on a specific monitor. This can only be done with physical units :(

ysinsane:
Close #5228

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


core/tauri-utils/src/lib.rs
```diff
@@ -211,6 +211,108 @@ impl Default for Env {
   }
 }
 
+/// Size unit type. Effective to size related settings
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+pub enum SizeUnit {
```
amrbashir: This should be renamed `PixelUnit` and used for both size and poistion. No need to have two different enums.

core/tauri-utils/src/lib.rs
```diff
@@ -211,6 +211,108 @@ impl Default for Env {
...
+  Physical
+}
+
+impl Default for SizeUnit {
+    fn default() -> Self {
+        SizeUnit::Logical
+    }
+}
+
+impl Serialize for SizeUnit {
```
amrbashir: Just derive `Serialize` and `Deserialize` instead of manual implementation

tooling/api/src/window.ts
```diff
@@ -68,7 +68,8 @@ import { TauriEvent } from './event'
 
 type Theme = 'light' | 'dark'
 type TitleBarStyle = 'visible' | 'transparent' | 'overlay'
-
+type SizeUnit = 'physical' | 'logical'
+type PositionUnit = 'physical' | 'logical'
```
amrbashir:
```suggestion
type PixelUnit = 'physical' | 'logical'
```

core/tauri-utils/src/lib.rs
ysinsane: Hi, thanks for the thoughtful review, but I notice other enum uses manual impl since the macro device approch need the config value to be capitalized. Eg: you have to set "Physical" instead of "physical" in config file.

amrbashir: yeah, you need to also add `#[serde(rename_all = "camelCase"]`

ysinsane:
@amrbashir  May I ask anything more I could do on this PR? I am willing to do more to pass this PR.

amrbashir:
It is just waiting for lucas review. He is a bit busy so we appreciate your patience.

lucasfernog:
I did check it a week ago and I think we should consider a different format for specifying physical units. My first idea was to use a number for logical values e.g. `100` and a string for physical ones such as `100px`. The approach on this PR is kinda restrictive since you can't mix logical and physical values in the configuration.

amrbashir:
I think @ysinsane had that before, where we add a `sizeUnit` and `positionUnit` fields but I asked him to combine them but I am fine with changing them back. 

I think `100px` would hint at there maybe some other css units like `100vw` and I don't like that.

ysinsane:
I actually kind of like lucas's idea, although it takes more work to implement it. As for detail, I'd like to know your idea on mix of logical and physical unit, for example 100 window width and 100px window height?
Cause I found it is not supported by the WRY API. @lucasfernog

amrbashir:
I don't feel like we should support different width and height. It doesn't make sense tbh and since it requires changes upstream I don't think we should do it now, maybe after switch to winit. For now, we can just add `sizeUnit` and `positionUnit`, what do you think @lucasfernog ?

lucasfernog:
Yeah we shouldn't mix different formats for the same property. I think for v1 we could go with the unit parameters, but refactor the entire configuration format for v2.

Miniontoby:
Hi!

I am now using this branch/fork and I don't seem to get it to work yet; as in the sizes still don't work... 
image

When I manually correct it with shortcuts from windows it looks like: 
image

So yeah the code I use:
```ts
const webviewwindow = new WebviewWindow('presentation', {
    url: this.url, fullscreen: true, resizable: false,
    title: 'presentation',
    x: monitor.position.x, y: monitor.position.y,
    width: monitor.size.width, height: monitor.size.height,
    pixelUnit: 'physical'
});
```

What am I doing wrong?


Btw when adding this code:
```ts
webviewwindow.once('tauri://created', async function () {
    console.log(monitor.position, await webviewwindow.innerPosition())
    console.log(monitor.size, await webviewwindow.innerSize())
});
```

This is the output:
```json
{ "type": "Physical", "x": -1680, "y": -908 }, { "type": "Physical", "x": -1680, "y": -908 }

{ "type": "Physical", "width": 1680, "height": 1050 }, { "type": "Physical", "width": 1920, "height": 1080 }
```

PR #5394, opened 2022-10-12
chore(examples): remove the legacy X-UA-Compatible meta

mvasilkov:
We aren't targeting Internet Exploder, so it's safe to bury its legacy.


lucasfernog (approved):
Thanks!

PR #5399, opened 2022-10-13
feat(macos): add `tabbing_identifier` option, closes #2804, #3912

### LINKED ISSUE #2804: Items automatically added to `View` menu on macOS
### **Describe the bug**
When you have a `View` menu, these items are automatically added:

image

I initially thought macOS did this automatically (it does for `Edit > Start Dictation` and `Edit > Emoji & Symbols`), but apps like Discord don't have it.

### **To Reproduce**
Create a menu called `View`:
```rust
.add_submenu(Submenu::new(
    "View",
    Menu::new(), // can be empty or not
  ));
```

### **Expected behavior**
Either don't auto-add them, or provide a way to not auto-add them

### **Platform and Versions (required):**
<details>
<summary>Details</summary>

```
Operating System - Mac OS, version 10.15.7 X64

Node.js environment
  Node.js - 14.18.1

Global packages
  npm - 6.14.15
  yarn - 1.22.17

Rust environment
  rustc - 1.56.0
  cargo - 1.56.0

App directory structure
/node_modules
/.github
/src-tauri
/build
/.git
/assets
/src

App
  tauri.rs - 1.0.0-beta.8
  build-type - bundle
  distDir - ../build
  devPath - http://localhost:8000
  framework - Svelte
```

</details>

### LINKED ISSUE #3912: How to create multiple separate windows, instead of "tabs"?
Windows created by `App::create_window()` are just like tabs, sharing a same window, but I want they each had a separate window, can tauri do this? If so, how?

Sorry if I've missed anything, but i've been googling for a long time, checked the docs and issues of tauri and tao, got nothing at all.

multiple windows sharing a window like tabs

caesar:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
Closes #2804, closes #3912

Uses the new APIs introduced in Tao in https://github.com/tauri-apps/tao/pull/586 and https://github.com/tauri-apps/tao/pull/592.


FabianLars:
Is it just the Windows/Linux user in me or is this something that should be off by default (or always)?

caesar:
I'd love it to be off by default but that's not default behaviour on macOS…
Should we override it by default? I don't know. 🤔

Edit: for reference, default behaviour on macOS (ie, when this is on) is to follow the preference set by the user.
The default system preference is for "automatic tabbing" to be enabled in fullscreen mode only.

caesar:
Just a thought – if it's off by default that will technically be a breaking change, so even if we do want that, maybe best to add it as-is for v1.2 and then change the default in v2.0?

lucasfernog (dismissed):
Nice addition, thanks! This also closes #2804

lucasfernog:
electron actually uses [tabbingIdentifier](https://developer.apple.com/documentation/appkit/nswindow/1644704-tabbingidentifier) and disables tabbing if the window is transparent or if it doesn't have decorations.
```cpp
if (tabbingIdentifier.empty() || transparent() || !has_frame()) {
    [window_ setTabbingMode:NSWindowTabbingModeDisallowed];
  } else {
    [window_ setTabbingIdentifier:base::SysUTF8ToNSString(tabbingIdentifier)];
  }
```

caesar:
> This also closes #2804

Ah, I was sure there was another issue, but I failed to find it, thanks.

> electron actually uses [tabbingIdentifier](https://developer.apple.com/documentation/appkit/nswindow/1644704-tabbingidentifier)

I did try that it it didn't seem to have any effect, but I might have done something wrong. Might be worth playing with it again.



> and disables tabbing if the window is transparent or if it doesn't have decorations.

That seems sensible. Though I must say I'm quite drawn to the idea of disabling it by default, and providing an option to enable. What do you think?

lucasfernog:
I can also play with the tabbingIdentifier later - I expect it to define groups of windows that can be tabbed together. I like the idea of disabling it by default, and it's what electron does.

lucasfernog:
I've opened a [PR](https://github.com/tauri-apps/tao/pull/592) to add tabbing identifier support on tao. I couldn't understand how it works when changing the identifier at runtime, but the builder `with_tabbing_identifier` method works fine. I think we should include it here and work similar to electron, they have that nice logic for it.

caesar:
<del>Wouldn't keeping both options make most sense, since both exist on macOS? I don't think they quite do the same thing.
At the very least, if a tabbing identifier isn't set, we should set `NSWindowTabbingModeDisallowed` like in the Electron code you shared above. Otherwise, unless a unique tabbing identifier is set for each window, windows will still be grouped into tabs.</del>

Edit: Apologies, now I look properly, I see that's effectively what you've done.
Seems good, but I'll go over it properly now 😄

caesar:
Apologies, now I look properly, I see that's effectively what you've done.
Seems good, but I'll go over it properly now 😄

caesar:
Looks good! Thanks @lucasfernog.

lucasfernog (approved):

PR #5406, opened 2022-10-15
feature: added variants to the Error enum to help constructive error messages

yankeeinlondon:
all user defined `#[command]` blocks -- whether in a plugin or in userland -- must return a `Result<T, tauri::error>` but the variants on **tauri::error** do not provide variants which an author in this situation would typically want to use. To aid them toward a sensible set of variants I have created the following:

1. `PluginCommandError` - forces plugin author to identify the plugin and the command and then allows a textual description of the error
2. `CommandError` - allows the userland author to identify the command and a textual description
3. `OtherError` - a means to allow other unexpected/non-covered use cases to have a safety valve where an appropriate variant is not otherwise available.

### DIFF
--- a/core/tauri/src/error.rs
+++ b/core/tauri/src/error.rs
@@ -144,7 +144,7 @@ pub enum Error {
 
   /// a "safety valve" error for authors not finding a more appropriate error variant
   #[error("{0}")]
-  OtherError(String)
+  OtherError(String),
 }
 
 pub(crate) fn into_anyhow<T: std::fmt::Display>(err: T) -> anyhow::Error {
--- a/core/tauri/src/error.rs
+++ b/core/tauri/src/error.rs
@@ -88,6 +88,16 @@ pub enum Error {
   /// Error initializing plugin.
   #[error("failed to initialize plugin `{0}`: {1}")]
   PluginInitialization(String, String),
+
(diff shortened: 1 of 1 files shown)


yankeeinlondon:
Folks, not sure what this error means or how it got there as it seems to be unrelated to the code in this PR directly:

> can't set `normalize_doc_attributes = true`

lucasfernog:
User defined commands can return any kind of [serializable](https://docs.rs/serde/latest/serde/ser/trait.Serialize.html) type, it doesn't need to use the `tauri::Error` type. The recommended way is to define your own Error enum, for instance:

```rust
use serde::{Serialize, Serializer};

#[derive(Debug, thiserror::Error)]
pub enum Error {
  #[error("IO error: {0}")]
  Io(#[from] std::io::Error),
  #[error(transparent)]
  Tauri(#[from] tauri::Error),
  #[error(transparent)]
  TauriApi(#[from] tauri::api::Error),
}

impl Serialize for Error {
  fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
  where
    S: Serializer,
  {
    serializer.serialize_str(self.to_string().as_str())
  }
}

type Result<T> = std::result::Result<T, Error>;

#[tauri::command]
fn my_cmd() -> Result<String> {
  Ok("response".into())
}
```

PR #5460, opened 2022-10-21
fix(cli): apply `.taurignore` rules to workspace members, closes #5355

### LINKED ISSUE #5355: [bug] cargo workspace memebers can NOT be ignored by `.taurignore`
### Describe the bug

Hello, after making lots of changes on my tauri +yew code there is a weird thing about appeared few days ago.
Every time I change the code it close the app and open it again instead of just refresh? 
Now, I have no idea how to debug this because my code became really large.

### Reproduction

1. create monorepo multiple workspaces with cargo
2. change in file in a workspace

### Expected behavior

Just reload or refresh the app instead of closing it.

### Platform and versions

1.1.1

### Stack trace

```shell
.
```

### Additional context

_No response_

amrbashir:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/tooling/cli/src/interface/rust.rs
+++ b/tooling/cli/src/interface/rust.rs
@@ -4,8 +4,8 @@
 
 use std::{
   collections::HashMap,
-  fs::File,
-  io::{BufRead, Read},
+  fs::{File, FileType},
+  io::{BufRead, Read, Write},
   path::{Path, PathBuf},
   process::{Command, ExitStatus},
   str::FromStr,
@@ -20,7 +20,7 @@ use std::{
 use anyhow::Context;
 #[cfg(target_os = "linux")]
 use heck::ToKebabCase;
-use log::info;
+use log::{debug, info};
 use notify::RecursiveMode;
 use notify_debouncer_mini::new_debouncer;
 use serde::Deserialize;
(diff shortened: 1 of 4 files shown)


tooling/cli/src/interface/rust.rs
```diff
@@ -364,27 +398,33 @@ impl Rust {
...
-      });
+      if !ignore_matcher.matched(path, true).is_ignore() {
+        info!("Watching {} for changes...", path.display());
+        lookup(path, |file_type, p| {
+          if p != path {
+            debug!("Watching {} for changes...", p.display());
+            let _ = watcher.watcher().watch(
+              &p,
+              if file_type.is_dir() {
+                RecursiveMode::Recursive
```
amrbashir, line 442: Since we are watching recursively, there maybe some files in this directory that are ignored and shouldn't trigger restart, so we need to use the `ignore_matcher` in the event loop to check if it is not ignored before triggering restart

lucasfernog: nice catch, pushed it

tooling/cli/src/interface/rust.rs
```diff
@@ -255,6 +257,59 @@ impl Interface for Rust {
...
+        let _ = ignore_builder.add_line(None, &line);
+      }
+
+      matchers.push(ignore_builder.build().unwrap());
+    }
+  }
+
+  IgnoreMatcher(matchers)
+}
+
```
amrbashir, line 312: I am not sure why you need to store multiple ignore matcher, we just need to collect all `.taurignore` files in the common parent and add it to a single ignore matcher
```suggestion
fn build_ignore_matcher(dir: &Path) -> GitIgnore {
  let mut ignore_builder = GitignoreBuilder::new(path.parent().unwrap());

  // ignore crate doesn't expose an API to build `ignore::gitignore::GitIgnore`
  // with custom ignore file names so we have to walk the directory and collect
  // our custom ignore files and add it using `ignore::gitignore::GitIgnoreBuilder::add`
    for entry in ignore::WalkBuilder::new(dir)
      .require_git(false)
      .ignore(false)
      .overrides(
        ignore::overrides::OverrideBuilder::new(dir)
          .add(".taurignore")
          .unwrap()
          .build()
          .unwrap(),
      )
      .build()
      .flatten()
  {
    let path = entry.path();
    ignore_builder.add(path);

    if let Ok(ignore_file) = std::env::var("TAURI_DEV_WATCHER_IGNORE_FILE") {
      ignore_builder.add(dir.join(ignore_file));
    }

    for line in crate::dev::TAURI_DEV_WATCHER_GITIGNORE.lines().flatten() {
      let _ = ignore_builder.add_line(None, &line);
    }
  }

	ignore_builder
}

```

lucasfernog: well I tried that, but then relative paths doesn't work. For instance, if you have a workspace with the `lib/` and `src-tauri/` members, and a `lib/.taurignore` file ignoring `src/lib.rs` (which relatively to the common ancestor is `lib/src/lib.rs`), it only ignores properly if you use a dedicated `GitignoreBuilder::new(path)` where `path` points to the `lib/` folder.

lucasfernog: See [GitignoreBuilder::new](https://docs.rs/ignore/latest/ignore/gitignore/struct.GitignoreBuilder.html#method.new) for more info, particularly:
```
The path given should be the path at which the globs for this gitignore file should be matched. Note that paths are always matched relative to the root path given here. Generally, the root path should correspond to the directory containing a .gitignore file.
```

amrbashir: Ah, I think I only tested with `.taurignore` in the root. This is fine then.

PR #5515, opened 2022-10-30
feat: add `title` getter on window, closes #5023

### LINKED ISSUE #5023: [feat] Make Window Title Available in appWindow Instance
### Describe the problem

It's tricky when making custom title bars for a window—or trying to update the window title with something based on the current title—to actually pin down in code what the current title is.

On the `appWindow` level, we have the ability to call `appWindow.setTitle("...")`. It would be good if we could get a reference to the current title here as well. A property of `appWindow.title` would probably be best, but exposing it through `appWindow.getTitle()` would also work fine.

### Alternatives considered

_No response_

### Additional context

_No response_

amrbashir:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


lucasfernog (dismissed):

lucasfernog (approved):

elibroftw:
When will this be in production? I'm working on a custom titlebar and this is the second last piece of perfection.
image
Also how do I disable the native title bar? I have decorations set to false.

elibroftw:
image
```js
appWindow.title().then(title => {
      if (windowTitle !== title) setWindowTitle(title);
    });
```
image

elibroftw:
image

elibroftw:
nvm just have to upgrade cargo most likely,

elibroftw:
I updated cargo.toml, and still same error. Does it work for you guys?

elibroftw:
Seems like the other plugins like tauri-plugin-window-state and tauri-plugin-store and tauri-plugin-single-instance depend on 1.2.5?

Had to set `tauri = { version = "1.3",`

FabianLars:
> Seems like the other plugins like tauri-plugin-window-state and tauri-plugin-store and tauri-plugin-single-instance depend on 1.2.5?

Nope, they only [specify `"1"`](https://github.com/tauri-apps/plugins-workspace/blob/v1/Cargo.toml#L8), so if we ignore the fact that some plugins use apis introduced in 1.1 and 1.2 the plugins don't care about the minor version and a simple `cargo update` is _supposed_ to update it (adding 1.3 in your cargo.toml file is basically the same in this context since it tells cargo to upgrade to at least 1.3)

PR #5563, opened 2022-11-06
fix(core): Invoke event listener in windows safely to avoid causing uncaught errors in windows that have loaded external urls

0rvar:
When opening a tauri window with an External URL, events sent with `emit_all`, including internal events like `tauri://window-created`, are still delivered. 
Since these windows have not loaded the necessary code, the invocation fails and causes an uncaught javascript error in the window's webview.

Functionally speaking, this is fine, but it's not very neat and has some noticeable effects:
* These errors will show up in the target websites error logging, if such is implemented.
* During development of apps with an error view, these errors will show up as a big modal covering the content - see screenshot below.

Screenshot 2022-11-06 at 21 31 15
Screenshot 2022-11-08 at 21 38 50

This PR adds a simple safeguard that checks that the listener function exists before calling it.

- [x] Bugfix

- [x] No

### Checklist

### Other information
Implementation approach briefly discussed here: https://discord.com/channels/616186924390023171/986184094050316358/1038527853236592661

### DIFF
--- a/core/tauri/src/manager.rs
+++ b/core/tauri/src/manager.rs
@@ -1408,7 +1408,7 @@ fn on_window_event<R: Runtime>(
       let windows = windows_map.values();
       for window in windows {
         window.eval(&format!(
-          r#"window.__TAURI_METADATA__.__windows = window.__TAURI_METADATA__.__windows.filter(w => w.label !== "{}");"#,
+          r#"var metadata = window.__TAURI_METADATA__; if (metadata != null) {{ metadata.__windows = window.__TAURI_METADATA__.__windows.filter(w => w.label !== "{}"); }}"#,
           label
         ))?;
       }
--- a/core/tauri/src/manager.rs
+++ b/core/tauri/src/manager.rs
@@ -1408,7 +1408,7 @@ fn on_window_event<R: Runtime>(
       let windows = windows_map.values();
       for window in windows {
         window.eval(&format!(
-          r#"var metadata = window.__TAURI_METADATA__; if (metadata != null) {{ metadata.__windows = window.__TAURI_METADATA__.__windows.filter(w => w.label !== "{}"); }}"#,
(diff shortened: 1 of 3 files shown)


lucasfernog (dismissed):

core/tauri/src/manager.rs
```diff
@@ -1408,7 +1408,7 @@ fn on_window_event<R: Runtime>(
       let windows = windows_map.values();
       for window in windows {
         window.eval(&format!(
-          r#"window.__TAURI_METADATA__.__windows = window.__TAURI_METADATA__.__windows.filter(w => w.label !== "{}");"#,
+          r#"(function () {{ var metadata = window.__TAURI_METADATA__; if (metadata != null) {{ metadata.__windows = window.__TAURI_METADATA__.__windows.filter(w => w.label !== "{}"); }} }})()"#,
```
amrbashir:
```suggestion
          r#"(function () {{ const metadata = window.__TAURI_METADATA__; if (metadata != null) {{ metadata.__windows = window.__TAURI_METADATA__.__windows.filter(w => w.label !== "{}"); }} }})()"#,
```

core/tauri/src/window.rs
```diff
@@ -1448,7 +1448,7 @@ impl<R: Runtime> Window<R> {
     payload: S,
   ) -> crate::Result<()> {
     self.eval(&format!(
-      "window['{}']({{event: {}, windowLabel: {}, payload: {}}})",
+      "(function () {{ var fn = window['{}']; fn && fn({{event: {}, windowLabel: {}, payload: {}}}) }})()",
```
```suggestion
      "(function () {{ const fn = window['{}']; fn && fn({{event: {}, windowLabel: {}, payload: {}}}) }})()",
```

amrbashir (changes requested):
is there a reason to use `var` instead of `let`?

lucasfernog (approved):

PR #5615, opened 2022-11-12
fix(core): sync windows metadata across all windows, closes #5571

### LINKED ISSUE #5571: [bug] Windows do not have access to each other (macOS)
### Describe the bug

I have a program which creates 2 windows `A` and `B`. 
After 2 seconds, `A` creates 2 children `A_child1` and `A_child2` using a tauri command. 
Every window displays the windows it has access to using `window.getAll()`, and refreshes this list every 2 seconds.

The problem is that some windows can't access some others (see screenshot below).

image

The behaviour is really strange, and I think this has to do with timing. Even at the beginning, with the first 2 windows, it seems that `A` can't access `B` (see screenshot)

image

My fellow developper tried this out on Windows, and the effect is less weird (only `A_child1` and `A_child2` have issues) but it's still suspicious

image

### Reproduction

1. Download the reproduction example: https://github.com/armand-colin/tauri-issue-window-label
2. Run it

### Expected behavior

I'd expect that all the windows will have access to each other _eventually_. I'd expect some timing to kick in, but here the windows never appear in the list.

### Platform and versions

Environment
  › OS: Mac OS 13.0.0 X64
  › Node.js: 14.19.1
  › npm: 6.14.16
  › pnpm: 6.11.0
  › yarn: 1.22.15
  › rustup: 1.25.1
  › rustc: 1.65.0
  › cargo: 1.65.0
  › Rust toolchain: stable-x86_64-apple-darwin 

Packages
  › @tauri-apps/cli [NPM]: 1.1.1
  › @tauri-apps/api [NPM]: 1.1.0
  › tauri [RUST]: 1.1.1,
  › tauri-build [RUST]: 1.1.1,
  › tao [RUST]: 0.14.0,
  › wry [RUST]: 0.21.1,

App
  › build-type: bundle
  › CSP: unset
  › distDir: ../dist
  › devPath: http://localhost:1420/

App directory structure
  ├─ node_modules
  ├─ src-tauri
  ├─ .git
  ├─ .vscode
  └─ src

### Stack trace

_No response_

### Additional context

_No response_

amrbashir:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


FabianLars (dismissed):
Tested on Windows.

lucasfernog:
Unfortunately this does not fix the sync issue completely. The main issue is that we use an initialization script to set the windows metadata (the init script is static and the metadata is dynamic). I started working on a fix for this when the issue was raised, I'll check if it doesn't introduce a breaking change (last time I thought it would).

An easy way to test is run the API example, create a second window, close the main window, and then reload the page of the second window. The metadata will contain the `main` label.

lucasfernog:
Ahh I remembered why I didn't finish my own work, it uses the custom protocol to inject the metadata so it needs the dev server proxy to work on dev :(

lucasfernog:
Let's try this solution for now then :(

FabianLars:
Disabling page reloads is a common request anyway :D

Can't we use the normal IPC instead of a custom protocol?

lucasfernog:
We can but if we use it, there'll be a small delay where you'll see a wrong list of labels.

lucasfernog:
Works on Windows, but it doesn't work on Linux. It seems like the first eval() call is lost on Linux, probably because the window isn't loaded yet. (i'll merge this anyway if we can't figure it out, it might be some webkit2gtk race condition).

lucasfernog:
Ok it gives me this error: `Err(Error { domain: WebKitJavascriptError, code: 699, message: "undefined:1:40: TypeError: undefined is not an object (evaluating 'window.__TAURI_METADATA__.__windows = [\"A\",\"B\"].map(function (label) { return { label: label } })')" })`.

It makes sense, but if I change the script to set `window.x = 5`, the variable isn't there when the app loads.

lucasfernog:
I'll push a possible wry fix and let the wry gods decide.

lucasfernog:
Aaand I had to do the same on macOS. Maybe this is more workarounds than we should have? cc @amrbashir @wusyong (see https://github.com/tauri-apps/wry/pull/815).

amrbashir:
I think it is fine to queue the `eval()` scripts until the page loads as most users will expect full access to the `window` object.

lucasfernog (approved):

PR #5639, opened 2022-11-17
fix double serialize on invoke

LucasJavaudin:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

When using the `invoke` function asynchronously to run a Rust function and return a value that can be serialized, the serialized value is serialized a second time, uselessly. This is because of a useless conversion in the function `hooks::respond_async_serialized`: The task returns a `Result<JsonValue, InvokeError>` that is converted into a `InvokeResponse` by the call to `into()`. An `InvokeResponse` is in practise exactly equal to a `Result<JsonValue, InvokeError>` so the conversion should be immediate. However, the call to `into()`  use the impl `From<Result<T, InvokeError>> for InvokeResponse` (where `T: Serialize`), which serialize the `JsonValue` into a `JsonValue` (and that is not instantaneous).

### DIFF
--- a/.changes/fix-invoke-double-serialize.md
+++ b/.changes/fix-invoke-double-serialize.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch
+---
+
+Fixes a double serialization on the IPC.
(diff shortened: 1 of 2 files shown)


lucasfernog (approved):
Nice catch, thanks!

Issue #5411, opened 2022-10-16
[feat] window with wry navigation events


ctaggart:
### Describe the problem

I'm building a Tauri app. For oauth, I want to create a new window where I can listen to the navigation events.  After the user authentications, the code is passed to the return url. This workflow is implemented in this wry example:
https://github.com/tauri-apps/wry/blob/dev/examples/navigation_event.rs

### Describe the solution you'd like

I'd like the same solution, but using the Tauri API. When I tried using that wry example with my Tauri app, I end up closing the entire application when I close the auth window. I'm a little lost at this lower level.

### Alternatives considered

`on_page_load` looks like it applies to just pages with Tauri JavaScript #4037.  I started looking at Tauri Plugins, but not sure what the right extension point is for integrating a custom wry window and getting Tauri to manage it.

FabianLars:
I'd count this as a duplicate of https://github.com/tauri-apps/tauri/issues/4080.

In the meantime you could use a small localhost listener, which would also allow using the default browser for logging in, or if you wanna get your hands dirty you could use the webview bindings yourself via https://docs.rs/tauri/latest/tauri/window/struct.Window.html#method.with_webview

PR #5686, opened 2022-11-24
Expose wry navigation_handler via WindowBuilder closes #4080

### LINKED ISSUE #4080: [feat] Improved native navigation API
Currently the ability to access and control navigation on a Window from the native rust code is extremely limited. The best you can hope for is evaluating JavaScript to change the URL. Since the native WebView APIs provide information regarding current URL with the ability to change it would make nice to expose some form of this from the Window structure.

Ideally there would be hooks to get the current window URL, navigate to a new URL and the ability to register listeners for changes in URL. WRY already provides a mechanism for attaching a navigation controller that's called every time the view URL changes. This could easily be used to provide navigation events as well as the ability to limit navigation to external URLs. Likewise, the native WebView structures allow retrieving and changing the URL, it's just a matter of creating a unified interface for accessing and manipulating that information and then plumbing it through the many abstraction layers.

silvanshade:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

#### Rationale

Currently, it is difficult to use `tauri` for certain situations, for example, those involving OAuth redirects. This has been noted in issues #3020, #5411, and elsewhere.

In some cases it is possible to find workarounds, such as by launching a local `warp` server and handling `localhost` redirects that way, but this is not always possible, particularly when the service in question uses deep linking and app-specific custom URI schemes.

A simple way to address this problem is to allow handling of URL navigation events with a callback configured for the webview. This is already possible now with `wry` using [`WebViewBuilder::with_navigation_handler`](https://docs.rs/wry/latest/wry/webview/struct.WebViewBuilder.html#method.with_navigation_handler). However, as far as I could tell, there was no way to access this functionality from `wry` directly through `tauri` prior to this PR.

This PR resolves that problem by allowing the user to set a navigation handler with `tauri::WindowBuilder` which then gets passed on to `wry` when the window is created via the runtime.


silvanshade:
Fixes #5411.

lucasfernog (dismissed):
Nice PR, thanks!

PR #5758, opened 2022-12-03
feat(cli): detect SolidJS and SolidStart

AR10Dev:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

### DIFF
--- a/tooling/cli/src/helpers/framework.rs
+++ b/tooling/cli/src/helpers/framework.rs
@@ -96,8 +96,8 @@ impl fmt::Display for Bundler {
 
 pub fn infer_from_package_json(package_json: &str) -> (Option<Framework>, Option<Bundler>) {
   let framework_map = [
-    ("solid-js", Framework::SolidJS, Some(Bundler::Vite)),
     ("solid-start", Framework::SolidStart, Some(Bundler::Vite)),
+    ("solid-js", Framework::SolidJS, Some(Bundler::Vite)),
     ("svelte", Framework::Svelte, Some(Bundler::Rollup)),
     ("@sveltejs/kit", Framework::SvelteKit, Some(Bundler::Vite)),
     ("@angular", Framework::Angular, Some(Bundler::Webpack)),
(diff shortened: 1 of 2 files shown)


tooling/cli/src/helpers/framework.rs
```diff
@@ -21,6 +23,8 @@ pub enum Framework {
 impl Framework {
   pub fn dev_path(&self) -> String {
     match self {
+      Self::SolidJS => "http://localhost:5173",
```
lucasfernog: I just tried `npx degit solidjs/templates/js my-app` and it runs on localhost:3000

tooling/cli/src/helpers/framework.rs
```diff
@@ -88,6 +96,8 @@ impl fmt::Display for Bundler {
 
 pub fn infer_from_package_json(package_json: &str) -> (Option<Framework>, Option<Bundler>) {
   let framework_map = [
+    ("solid-js", Framework::SolidJS, Some(Bundler::Vite)),
+    ("solid-start", Framework::SolidStart, Some(Bundler::Vite)),
```
lucasfernog, line 99: order matters here, `SolidStart` projects also includes a `solid-js` dependency. So we need to check for `solid-start` first.

lucasfernog (changes requested):

tooling/cli/src/helpers/framework.rs
AR10Dev: It runs on localhost:3000 because in the [Vite config file](https://github.com/solidjs/templates/blob/master/js/vite.config.js#L7), the port is specified. I have test it using my [own template](https://github.com/AR10Dev/solid-tailwind-ts-vite) and it runs on localhost:5173, the default port of Vite 3.

lucasfernog: Yeah but I assume the official template is more used, if so we should go with that one.

AR10Dev: I think this issue is related even to few other frameworks. I think the best way to solve this problem is to read the bundler configuration file. I will try to solve it in the next few days 😊

lucasfernog (approved):

PR #5788, opened 2022-12-08
feat(bundler): Add support for creating NSIS bundles on unix hosts

FabianLars:
This should not be merged into the nsis branch directly, i just used it as a base here for a better diff view

----

This adds support for building nsis setup bundles (including the updater bundle) on Linux hosts. macos should hopefully work too but i am not sure how to get makensis there.

I explicitly have only the msvc toolchain in mind when talking about windows targets, even when cross compiling so the setup is a little bit more involved:

- Install nsis, for example on ubuntu the command is `sudo apt install nsis`
- the default MS linker doesn't work on Linux so we also need the LLD linker, on ubuntu it's just `sudo apt install lld`
- add the msvc rust target `rustup target add x86_64-pc-windows-msvc`
- Install the windows sdks:
cargo install xwin
xwin --accept-license splat --output ~/xwin
- Add or modify `.cargo/config.toml` with these contents:
```toml
[target.x86_64-pc-windows-msvc]
linker = "lld"
rustflags = [
  "-Lnative=/home/username/xwin/crt/lib/x86_64",
  "-Lnative=/home/username/xwin/sdk/lib/um/x86_64",
  "-Lnative=/home/username/xwin/sdk/lib/ucrt/x86_64"
]
```

Now to the fun part:
- Install this branch from git: `cargo install tauri-cli --git https://github.com/tauri-apps/tauri --branch feat/nsis-linux`
- run `cargo tauri build --target x86_64-pc-windows-msvc`

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No, nopefully not.

### Checklist


tooling/bundler/src/bundle.rs
```diff
@@ -43,26 +42,37 @@ pub fn bundle_project(settings: Settings) -> crate::Result<Vec<Bundle>> {
...
       PackageType::IosBundle => macos::ios::bundle_project(&settings)?,
+      // dmg is dependant of MacOsBundle, we send our bundles to prevent rebuilding
+      #[cfg(target_os = "macos")]
+      PackageType::Dmg => macos::dmg::bundle_project(&settings, &bundles)?,
+
       #[cfg(target_os = "windows")]
-      PackageType::WindowsMsi => windows::msi::bundle_project(&settings, false)?,
-      #[cfg(target_os = "windows")]
-      PackageType::Nsis => windows::nsis::bundle_project(&settings, false)?,
+      PackageType::WindowsMsi if target == "windows" => {
```
amrbashir: isn't this a redundant check since we only support WiX compilation on Windows only

tooling/bundler/src/bundle.rs
```diff
@@ -43,26 +42,37 @@ pub fn bundle_project(settings: Settings) -> crate::Result<Vec<Bundle>> {
...
+      PackageType::Deb if target == "linux" => linux::debian::bundle_project(&settings)?,
       #[cfg(target_os = "linux")]
-      PackageType::Rpm => linux::rpm::bundle_project(&settings)?,
+      PackageType::Rpm if target == "linux" => linux::rpm::bundle_project(&settings)?,
       #[cfg(target_os = "linux")]
-      PackageType::AppImage => linux::appimage::bundle_project(&settings)?,
-      // dmg is dependant of MacOsBundle, we send our bundles to prevent rebuilding
-      #[cfg(target_os = "macos")]
-      PackageType::Dmg => macos::dmg::bundle_project(&settings, &bundles)?,
+      PackageType::AppImage if target == "linux" => linux::appimage::bundle_project(&settings)?,
```
amrbashir: ditto

amrbashir (changes requested):
Other than these two nit-picks, I think this PR is ready

tooling/bundler/src/bundle.rs
FabianLars: correct, it's leftover stuff from when i just tried to make it work in the first place :D good catch, thx

FabianLars:
Should we maybe add a log output saying that cross compilation is experimental/unsupported or something? i have a feeling that saying this in the docs won't be enough x)

FabianLars:
oh and another thing, should nsis' verbose logging be tied to the `--verbose` flag?

amrbashir:
> Should we maybe add a log output saying that cross compilation is experimental/unsupported or something? i have a feeling that saying this in the docs won't be enough x)

Sounds good!

> oh and another thing, should nsis' verbose logging be tied to the --verbose flag?

by default, nsis is executed with maximum logging `/V4`, and will be presented with `--verbose`

FabianLars:
> by default, nsis is executed with maximum logging /V4, and will be presented with --verbose

Ah yeah you're right. i always used the `--verbose` flag in my tests 😂 stupid muscle memory

tooling/bundler/Cargo.toml
```diff
@@ -33,17 +33,21 @@ tempfile = "3.3.0"
 log = { version = "0.4.17", features = [ "kv_unstable" ] }
 dirs-next = "2.0"
 
-[target."cfg(target_os = \"windows\")".dependencies]
+# dependencies for Windows targets
```
lucasfernog, line 38: covector will strip this later :( so sad

FabianLars:
The timing of the warning may be a bit weird. It prints after cargo build, before it starts creating the bundles, so if `tauri build` fails at the actual compilation step already then this warning won't be seen, but at least it's not spammed away from the compiler output, and it's next to the other bundler logs.

Or maybe it needs some rephrasing so that it's more about the bundling step and not just `cross-platform compilation` 🤔

tooling/bundler/src/bundle.rs
```diff
@@ -43,25 +42,38 @@ pub fn bundle_project(settings: Settings) -> crate::Result<Vec<Bundle>> {
...
 
+  let target_os = settings
+    .target()
+    .split('-')
+    .nth(2)
+    .unwrap_or(std::env::consts::OS)
+    .replace("darwin", "macos");
+
+  if target_os != std::env::consts::OS {
+    warn!("Cross-platform compilation is experimental and does not support all features. Please use a matching host system for full compatibility.");
```
iilyak, line 53: A link to documentation which describes what exactly is not supported would be great.

iilyak: Is it only `updates` and `signing`?

lucasfernog:
For macOS:

Install nsis:
`brew install nsis`

Install the lld linker:
`brew install llvm`
(also add `/opt/homebrew/opt/llvm/bin` to your PATH like suggested in the install output

add the msvc rust target:
`rustup target add x86_64-pc-windows-msvc`

Install the windows sdks:
`cargo install xwin`
`xwin --accept-license splat --output ~/xwin --disable-symlinks`

(I had to add the disable-symlinks flag otherwise it failed to run)

Add or modify .cargo/config.toml with these contents:
```
[target.x86_64-pc-windows-msvc]
linker = "lld"
rustflags = [
  "-Lnative=/Users/username/xwin/crt/lib/x86_64",
  "-Lnative=/Users/username/xwin/sdk/lib/um/x86_64",
  "-Lnative=/Users/username/xwin/sdk/lib/ucrt/x86_64"
]
```

FabianLars:
marked it as draft again because of missing cross-compilation support in tauri-build. Going to look how we can add it with winres/winresource.

Edit: I think i found a way to use mingw's windres with the msvc setup from above. going to create tauri-winres tomorrow and update tauri-build accordingly.

FabianLars:
On Linux we now also need to do `sudo apt install binutils-mingw-w64` - this will install the resource compiler for both 32 and 64 bit.
I'm also not 100% sure about the consequences of the winres changes, but that's why we print the warning i guess 😂 

Now the problem is macOS, looks like you have to install the whole mingw-w64 toolchain :/ (`brew install mingw-w64` ?)

tooling/bundler/src/bundle/windows/nsis.rs
```diff
@@ -56,6 +60,13 @@ const NSIS_REQUIRED_FILES: &[&str] = &[
   "Include/FileFunc.nsh",
   "Include/x64.nsh",
 ];
+#[cfg(not(target_os = "windows"))]
+const NSIS_REQUIRED_FILES: &[&str] = &[
+  "Plugins/x86-unicode/NScurl.dll",
+  "Plugins/x86-unicode/ApplicationID.dll",
+  "Plugins/x86-unicode/nsProcess.dll",
+  "Plugins/x86-unicode/nsis_semvercompare.dll",
+];
```
amrbashir, line 69: you should also check for other important plugins, like `Mui2`, `FileFunc` and `x64`

FabianLars: Hmm okay. What should we do if these files are missing? panic and ask the user to re-install nsis? or should we download the windows portable zip and extract them from there?

amrbashir: panic and require a reinstall

FabianLars: we have a classic Linux problem here. On most distros these plugins _should_ be in `/usr/share/nsis/Include` but we probably can't hardcode that since there are always distros that like to divert from the standards 🙄 

i'm gonna think about it a bit over the weekend because the idea of scanning the output of makensis for `!include: could not find: "MUI2.nsh"` etc doesn't sound nice either 😂

amrbashir: I guess it is fine then, this is a rare case anyways and `--verbose` logs will show the error if someone reports it.

tooling/bundler/src/bundle.rs
FabianLars: > A link to documentation which describes what exactly is not supported would be great.

There are no docs yet, but i guess a link to https://tauri.app/v1/guides/building/cross-platform should be a fairly safe bet.
"What exactly is not supported" - That's actually a bit tricky to know upfront. Obvious candidates are code signing and the .msi bundle target.
The idea of that message is more like "This is experimental and you should prefer native hosts, we don't know what works yet either" etc. Also i'd like to have it give us a bit more room to add features that only works if you compile it on the native host without that being a huge surprise to users.

lucasfernog:
Still works on Linux, and also works on macOS without downloading any extra dependency @FabianLars :partying_face:

lucasfernog (approved):

PR #5789, opened 2022-12-08
Support sending raw byte data to the "data" event for child command's stdout and stderr

### LINKED ISSUE #5465: [feat] Support sending byte data to the "data" event for child command's stdout and stderr
### Describe the problem

See related issue: https://github.com/tauri-apps/tauri/issues/5309

Currently we receive type `string` for the the "data" event listener of `child.stdout.on("data" ....)` and `child.stderr.on("data" ....)`. This is suitable for the vast majority of simple communication, but challenging to work-around for the cases in which binary data must be fed directly through for applications that use more complex communication formats.

In my specific case, I'm trying to read `msgpack` messages via the command output of a child process, and currently that is impossible without writing a workaround solution for several reasons:

1. The default operation if no encoding specified is to attempt a conversion to utf_8 using rust's built-in decoder, executed [here](https://github.com/tauri-apps/tauri/blob/dev/core/tauri/src/api/process/command.rs#L411)
2. If there is an encoding, then it uses `encoding_rs` which **correctly** replaces invalid utf-8 with replacement characters.

In scenario 1, we observe the symptoms of https://github.com/tauri-apps/tauri/issues/5309, which is correct behavior. In scenario 2, encoding_rs eliminates unrecognized or invalid utf8 sequences entirely, which is also correct behavior.

I would like a flag in `SpawnOptions` that allows for the output of the process to be sent directly to the front end, either as an array of numbers resembling a `Uint8Array` or an escaped string with control characters like `\x...` or `\u...`.

### Alternatives considered

I'm going to have to work around this for now by implementing a shim executable that can perform this translation beforehand, and then use that as my sidecar. I cannot change the message format of my current sidecar, so a work-around is required to get communication working properly.

### Additional context

If you're interested, the sidecar application is neovim, and I'm trying to communicate with it using RPC directly. I have previously implemented this in Node and Electron. This is currently blocking development while I hack a solution into existence.

cymruu:
Support sending raw byte data to the "data" event for child command's stdout and stderr. Still WIP

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes, and the changes were approved in issue #5465
- [ ] No

### Checklist

### Other information
Closes #5465


core/tauri/src/api/process/command.rs
```diff
@@ -364,14 +373,20 @@ impl Command {
...
-          CommandEvent::Stderr(line) => {
-            stderr.push_str(line.as_str());
-            stderr.push('\n');
-          }
+          CommandEvent::Stdout(line) => match line {
+            Buffer::Text(line) => {
+              stdout.push_str(line.as_str());
+              stdout.push('\n');
+            }
+            Buffer::Raw(_) => {}
```
cymruu: @amrbashir I'd like to implement "raw" data also for `output` method. What do you think would be the best approach to that?

amrbashir: you can make `stdout` and `stderr` of the `Output` struct to be `Vec<u8>` and anyone can call `String::from_utf8` on it to turn it into a string

core/tauri/src/api/process/command.rs
```diff
@@ -406,9 +421,14 @@ fn spawn_pipe_reader<F: Fn(String) -> CommandEvent + Send + Copy + 'static>(
...
-            Some(encoding) => Ok(encoding.decode_with_bom_removal(&buf).0.into()),
-            None => String::from_utf8(buf.clone()),
+          let line = match encoding {
+            EncodingWrapper::Text(character_encoding) => match character_encoding {
+              Some(encoding) => Ok(Buffer::Text(
+                encoding.decode_with_bom_removal(&buf).0.into(),
+              )),
+              None => String::from_utf8(buf.clone()).map(|text| Buffer::Text(text)),
+            },
+            EncodingWrapper::Raw => Ok(Buffer::Raw(buf.clone())),
```
amrbashir: maybe try always sending raw data here and move the decoding logic to `endpoints/shell.rs` right before passing it to Javascript

core/tauri/src/endpoints/shell.rs
```diff
@@ -31,13 +38,61 @@ fn command_child_store() -> &'static ChildStore {
   &STORE
 }
 
-#[derive(Debug, Clone, Deserialize)]
+#[derive(Debug, Clone, Serialize)]
+#[serde(tag = "event", content = "payload")]
+#[non_exhaustive]
+enum JSCommandEvent {
```
cymruu, line 44: I had to introduce this extra type to get serde serialize it into javascript event properly. Couldn't use `CommandEvent` because it now contains `Vec<u8>` which would always serialize to `number[]`. 

Maybe you have some ideas how I can improve this?

core/tauri/src/api/process/command.rs
```diff
@@ -350,27 +340,29 @@ impl Command {
...
     let (mut rx, _child) = self.spawn()?;
 
     let output = crate::async_runtime::safe_block_on(async move {
       let mut code = None;
-      let mut stdout = String::new();
-      let mut stderr = String::new();
+      let mut stdout = Vec::<u8>::new();
+      let mut stderr = Vec::<u8>::new();
+
+      const NEWLINE_BYTE: u8 = 10;
```
amrbashir:
```suggestion
      const NEWLINE_BYTE: u8 = b'\n';
```

core/tauri/src/api/process/command.rs
```diff
@@ -219,19 +219,12 @@ impl Command {
...
-    self
-  }
-
   /// Spawns the command.
   ///
   /// # Examples
   ///
   /// ```rust,no_run
-  /// use tauri::api::process::{Command, CommandEvent};
+  /// use tauri::api::process::{Buffer, Command, CommandEvent};
```
amrbashir: what's the use of this import?

core/tauri/src/endpoints/shell.rs
```diff
@@ -31,13 +38,61 @@ fn command_child_store() -> &'static ChildStore {
...
+    match event {
+      crate::api::process::CommandEvent::Terminated(payload) => JSCommandEvent::Terminated(payload),
+      crate::api::process::CommandEvent::Error(error) => JSCommandEvent::Error(error),
+      crate::api::process::CommandEvent::Stderr(line) => get_event_buffer(line, encoding)
+        .map(JSCommandEvent::Stderr)
+        .unwrap_or_else(|e| JSCommandEvent::Error(e.to_string())),
+      crate::api::process::CommandEvent::Stdout(line) => get_event_buffer(line, encoding)
+        .map(JSCommandEvent::Stdout)
+        .unwrap_or_else(|e| JSCommandEvent::Error(e.to_string())),
+    }
```
amrbashir: just import `CommandEvent` into scope instead of importing it multiple times like this

amrbashir (changes requested):

core/tauri/src/endpoints/shell.rs
amrbashir: that's fine

core/tauri/src/api/process/command.rs
cymruu: leftover from one of previous approaches, thanks for pointing it out

cymruu:
@amrbashir I have addressed your comments. Thank you for reviewing!

core/tauri/src/api/process/command.rs
```diff
@@ -20,6 +20,9 @@ use std::os::windows::process::CommandExt;
 const CREATE_NO_WINDOW: u32 = 0x0800_0000;
 
 use crate::async_runtime::{block_on as block_on_task, channel, Receiver, Sender};
+
+pub use crate::endpoints::shell::Buffer;
```
amrbashir: I believe this is not needed anymore.

core/tauri/src/endpoints.rs
```diff
@@ -34,7 +34,7 @@ mod path;
 #[cfg(process_any)]
 mod process;
 #[cfg(shell_any)]
-mod shell;
+pub(crate) mod shell;
```
amrbashir: not needed

tooling/api/src/shell.ts
```diff
@@ -487,25 +487,39 @@ class Command extends EventEmitter<'close' | 'error'> {
...
         stdout.push(line)
       })
-      this.stderr.on('data', (line: string) => {
+      this.stderr.on('data', (line: EventPayload) => {
         stderr.push(line)
       })
+
       this.on('close', (payload: TerminatedPayload) => {
-        resolve({
+        return resolve({
```
amrbashir: don't think that this return is needed

tooling/api/src/shell.ts
```diff
@@ -487,25 +487,39 @@ class Command extends EventEmitter<'close' | 'error'> {
...
       this.spawn().catch(reject)
     })
   }
+
+
+  private collectEvents(events: EventPayload[]): string | Uint8Array {
+    if (this.options.encoding === 'raw') {
+      return new Uint8Array(events.reduce<Uint8Array>((p, c) => {
+        return (new Uint8Array([...p, ...c as Uint8Array, 10]))
+      }, new Uint8Array()))
```
amrbashir:
```suggestion
      return events.reduce<Uint8Array>((p, c) => {
        return (new Uint8Array([...p, ...c as Uint8Array, 10]))
      }, new Uint8Array())
```

amrbashir (changes requested):
Since this is a breaking change and will be merged into `next`, can you rebase the changes in top of the `next` branch?

amrbashir (dismissed):
LGTM. Thanks! I will leave the final review and merge to lucas.

lucasfernog:
Looks really good. I'll make some tests and see if we can improve the api types.

lucasfernog:
Let me know what you think about my change :P I'm not a TypeScript expert but I tried to let it infer most of the types from the encoding argument.

cymruu:
@lucasfernog thank you for the improvement with result type inferring. I love it! I have merged with `next`branch again.

BTW what is the reason behind `bundle.global.js` being included in the repository?

FabianLars:
`bundle.global.js` powers the `withGlobalTauri` (= the apis in `window.__TAURI__`). Building/Bundling that on the dev's machine is not really an option

cymruu:
@lucasfernog @FabianLars Is the feature still needed? Is there anything else I can do regarding the PR? 
I'm happy to help 🙂

lucasfernog:
I just need some time to test this, my idea was to wait for 1.3 to launch before merging it since you already merged dev into this.

lucasfernog (dismissed):

lucasfernog (approved):

lucasfernog:
Amazing work!

PR #5918, opened 2022-12-26
feat(core): allow configuring remote domains with IPC access, closes #5088

### LINKED ISSUE #5088: [feat] Inject window.__TAURI__ in allowed remote URLs

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


core/tauri-utils/src/config.rs
```diff
@@ -1106,6 +1106,20 @@ impl Default for DisabledCspModificationKind {
   }
 }
 
+/// External command access definition.
+#[derive(Debug, PartialEq, Eq, Clone, Deserialize, Serialize)]
+#[cfg_attr(feature = "schema", derive(JsonSchema))]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct ExternalCommandAccessScope {
+  /// The url to allow, matched against the webview URL using a glob pattern.
+  pub url: Url,
```
Sytten: The reason I used a string here is to allow URL patterns. In our case we need to allow all URLs (`*`) but there is a more "legitimate" use case to allow all sub domains of a particular domain (`*.example.com`). Since it is converted into a glob anyway down the line it is fine.

core/config-schema/schema.json
```diff
@@ -2547,6 +2557,37 @@
         }
       ]
     },
+    "ExternalCommandAccessScope": {
```
Sytten: Unsure if the name should change since it also whitelist plugin

core/tauri-utils/src/config.rs
```diff
@@ -1141,6 +1155,21 @@ pub struct SecurityConfig {
...
   #[serde(default, alias = "dangerous-disable-asset-csp-modification")]
   pub dangerous_disable_asset_csp_modification: DisabledCspModificationKind,
+  /// Allow external urls to send command to Tauri.
+  ///
+  /// By default, external urls do not have access to `window.__TAURI__`, which means they cannot
+  /// communicate with the commands defined in Rust. This prevents attacks where an externally
+  /// loaded malicious or compromised sites could start executing commands on the user's device.
+  ///
+  /// This configuration allows a set of external urls to have access to the Tauri commands.
+  /// Wildcards patterns are accepted and can be used to allow either all paths or
```
Sytten: AFAIK if you use an URL, this is not true anymore. But like I said I prefer the glob.

core/tauri/src/window.rs
```diff
@@ -1320,8 +1337,13 @@ impl<R: Runtime> Window<R> {
...
         let resolver = InvokeResolver::new(self, payload.callback, payload.error);
-
         let invoke = Invoke { message, resolver };
+
+        if !is_local && scope.is_none() {
+          invoke.resolver.reject("Scope not defined");
+          return Ok(());
+        }
+
         if let Some(module) = &payload.tauri_module {
```
Sytten, line 1440: Hummm I don't like that. Currently there is no check for modules like `fs`. I think we should do something similar to plugin where we can whitelist each module.

core/tauri/src/window.rs
```diff
@@ -1330,6 +1352,17 @@ impl<R: Runtime> Window<R> {
...
+              .map(|s| s.plugins.contains(&plugin_name))
+              .unwrap_or(true)
+            {
+              invoke.resolver.reject("Plugin not allowed");
+              return Ok(());
+            }
+          }
           manager.extend_api(invoke);
         } else {
           manager.run_invoke_handler(invoke);
```
Sytten, line 1465: Since we forked Tauri to ship our patched version, I had some ideas. I think whitelisting  which command can be run would be really cool, but not a blocker for this PR.

Sytten (changes requested):
Thanks for taking on this challenge. I left some comments, some are blocking IMO.

Apart from those comments, I remember there was a reason why I didnt reuse the same function to inject the scripts in the window, but I don't remember why...

core/config-schema/schema.json
```diff
@@ -2495,6 +2497,14 @@
               "$ref": "#/definitions/DisabledCspModificationKind"
             }
           ]
+        },
+        "dangerousExternalCommandAccess": {
+          "description": "Allow external urls to send command to Tauri.\n\nBy default, external urls do not have access to `window.__TAURI__`, which means they cannot communicate with the commands defined in Rust. This prevents attacks where an externally loaded malicious or compromised sites could start executing commands on the user's device.\n\nThis configuration allows a set of external urls to have access to the Tauri commands. Wildcards patterns are accepted and can be used to allow either all paths or a subset of paths for a domain.\n\n**WARNING:** Only use this option if you either have internal checks against malicious external sites or you can trust the allowed external sites. You application might be vulnerable to dangerous Tauri command related attacks otherwise.",
```
tweidinger: I would explicitly state that _only_  _fully_ trusted websites are allowed.

core/tauri-utils/src/config.rs
tweidinger: Since in the general use case this is extremely fragile, it should be hard to allow all domains. Non pentest/security apps are the exception with edge cases and deliberately 'insecure' settings.

core/tauri-utils/src/config.rs
lucasfernog: It should still work, we use the same logic for the HTTP allowlist.

core/tauri/src/window.rs
lucasfernog: I thought about adding an allowlist configuration object for each external access scope, but maybe the root allowlist can be used for that instead? Any reason a Tauri app should have multiple allowlists, one for each URL?

Also this can be confusing since you'd need to enable the allowlist on the root object to add it to the binary, and also allow it in the external access scope.

Sytten: Yes this is important since it relates to security. Let me give you our usecase.

We have a first in-app window that manages the connections of the user. That window has access to the filesystem to store the configuration.

The user the clicks on a connection which opens a second window with the external URL. We really do not want this window accessing the filesystem directly. We just want it to access certain carefully crafted commands that are safe to call.

This is why I originally only allowed commands and even then I wanted to build a way to limit which commands could be called 

Though I agree there could be a default to allow it to take the same values as the general allow list for modules.

core/tauri-utils/src/config.rs
Sytten: I will need to test but I am pretty certain the Url parser rejects a single `*`?

FabianLars: yes but you can do `https://*` (or `https://**`?) with effectively the same effect.

Edit: stuff like `https://*.domain.com` also works

Sytten: I guess this is good for official TaurI, wont cut it for us since we also need to support http but I will maintain my fork.

FabianLars: well, you can also do `http://...` of course. Sure, it's a bit annoying if you want to support both protocols at the same time because it only accepts a single url (especially if we take into account the other convo about finer allowlist-like control)

core/tauri/src/window.rs
lucasfernog: See the message above. Also, it would need a lot of code to add this :(

core/tauri-utils/src/config.rs
Sytten: Yup, thus my recommendation to keep it as a free string

core/tauri/src/window.rs
Sytten: This can come in a subsequent PR, happy to pitch in for the implementation

core/tauri-utils/src/config.rs
lucasfernog: Allowing `http` is super dangerous. And we could also allow an array of urls on the same scope.

core/tauri/src/window.rs
lucasfernog: Yeah I think we should add an option to block all Tauri APIs. That's easy and won't bloat the codebase with runtime allowlist checks.

lucasfernog: Pushed a change to address this.

core/tauri-utils/src/config.rs
Sytten: I mean sure but at the same time this feature is an escape hatch and its made abundantly clear with the name that its dangerous. Http is perfectly acceptable for a lot of use cases like local installations or forwarded via an SSH tunnel.

austenc:
This PR opens up some really cool possibilities. Also, I know it's probably still a WIP.

After trying it out I think that the messaging and/or error related to the `windows` part of the config can be improved. More specifically, it wasn't clear what value to use for `windows` -- the app's name (the "title" of the window), or `main`. 

Originally had something like this in my `tauri.conf.json`:
```json
"dangerousExternalCommandAccess": [
    {
      "url": "https://my-localhost-app-url",
      "windows": ["my-app-name"]
    }
]
```
When changed to `"windows": ["main"]` it works. 

Not sure whether it's a matter of improving the `Scope not defined` error message, or whether it's a documentation thing, but that was a tiny papercut I ran into. Thanks again for the work on this!

lucasfernog:
@austenc we do mention it's the window labels, see https://github.com/tauri-apps/tauri/pull/5918/files#diff-ff73997dde3dc4965aa0c59ce0b37c7d208ae5b6b7686bf19e14d93c98cc1bacR1116. I'll improve the error message though.

austenc:
Thanks for helping out a rust noob :) And thanks again for the work on this feature, really great!

core/tauri-utils/src/config.rs
```diff
@@ -1106,6 +1106,23 @@ impl Default for DisabledCspModificationKind {
...
 
+/// External command access definition.
+#[derive(Debug, PartialEq, Eq, Clone, Deserialize, Serialize)]
+#[cfg_attr(feature = "schema", derive(JsonSchema))]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct ExternalCommandAccessScope {
+  /// The url to allow, matched against the webview URL using a glob pattern.
+  pub url: Url,
+  /// The list of window labels this scope applies to.
+  pub windows: Vec<String>,
```
tweidinger, line 1209: As the window label is accessible as a property in the window API endpoint, it is risky to distinguish windows only by their label and not by another identification method.
Since the current code base is not capable to do this, it would be opening up another hole for adversaries.

core/tauri-utils/src/config.rs
tweidinger: How about only allowing `https://` in the normal configuration while you could still manipulate the state (and scope) from rust core if you _really_ _really_ need to and understand the implications?
The same goes for wildcards like `https://*` and only allow `https://example.tld/*` or `https://*.example.tld` this way the already insecure feature is still somewhat reasonable.

1zun4secondary:
How do I exactly use this tauri pull request?
I imported the library into my rust-project via
`tauri = { git = "https://github.com/tauri-apps/tauri.git", branch = "feat/remote-ipc", features = ["api-all"] }`
which worked out.

But when I run `npm run tauri dev`, it will say:
```
Error `tauri.conf.json` error on `tauri > security`: Additional properties are not allowed ('dangerousExternalCommandAccess' was unexpected)
```

Which makes sense, because the NPM tauri-cli library is still the v1.2.2. I don't know much about NPM, so I was wondering how I can use this pull request now?

FabianLars:
@1zun4 see https://tauri.app/v1/guides/faq#how-can-i-use-unpublished-tauri-changes (`Using the Tauri CLI from source`) - i recommend using the cargo cli instead since it's easier to use a git version imo.

Oh and make sure to also use the `tauri-build` crate from git.

1zun4secondary:
> @1zun4 see https://tauri.app/v1/guides/faq#how-can-i-use-unpublished-tauri-changes (`Using the Tauri CLI from source`) - i recommend using the cargo cli instead since it's easier to use a git version imo.
> 
> Oh and make sure to also use the `tauri-build` crate from git.

Thank you very much! It works flawless with my application. Looking forward for it being merged.

> When changed to `"windows": ["main"]` it works.
> 
> Not sure whether it's a matter of improving the `Scope not defined` error message, or whether it's a documentation thing, but that was a tiny papercut I ran into. Thanks again for the work on this!

... took me a while to figure it out... haha

ddx95:
Hello, is there any way to modify/add dynamically a value to the dangerousExternalCommandAccess field from the main.rs ? I would need, for example, to get my remote server address from another configuration file, a environment variable or a register key.

tm1000:
What's the status of this PR?

prestomation:
Looks like there is lots of interest in this feature. What can the community to do help it land?

nothingismagick (commented):
First of all, this is an approach that removes security features that underly the entire premise of Tauri - all of us can agree on this. It is basically permitting people to drive a motorcycle without a helmet.

To that extent I would propose that we empower the individual to KNOW that this is going on and DENY such permissive use, especially in "*" wildcard situations.

One idea to harden this is to use a model similar to the way our Isolation pattern works, where the app sends an authorization to the remote, and the remote has to reply properly or tauri IPC execution is prevented.

I think @tauri-apps/wg-security must absolutely sign off on this before it enters canon, especially since this modifies the isolation.js guardrails as well.

lucasfernog:
The security team has been auditing this PR and the finds will be public soon - along with the proposed solution to improve security.

Sytten:
Can't wait for their input, I still stand with the review I made 3 months ago. This feature is an escape hatch for "dangerous" usecases, so for me it should allow users to whitelist each element as they wish. It should be very specific in its definition but broad in what it allows (like: I want to allow plugin X for all websites). It should definitely not be used 99% of the time. Our usecase for it was to be able to allow loading remote UI that can interact with **some** whitelisted tauri commands. It would even be ok with a compiler warning if this feature is used.

core/tauri/src/manager.rs
```diff
@@ -480,7 +471,7 @@ impl<R: Runtime> WindowManager<R> {
       });
     }
 
-    let window_url = Url::parse(&pending.url).unwrap();
+    let window_url = pending.current_url.lock().unwrap().clone();
     let window_origin =
       if cfg!(windows) && window_url.scheme() != "http" && window_url.scheme() != "https" {
         format!("https://{}.localhost", window_url.scheme())
```
prestomation, line 480: Using this change, I had to add  `tauri.localhost` to `dangerousRemoteUrlIpcAccess` for a windows release build to work with the shipped javascript bundle. This didn't seem to be necessary for OSX. This block looks relevant, but haven't dived into the details here.

lucasfernog: thanks for catching this, I'll take a look.

lucasfernog: Fix pushed. Tag me if you find anything weird :)

lucasfernog:
One thing I did not touch is usage of remote URLs with the isolation protocol enabled. Since we can't inject the isolation protocol scripts (?) it does not work at all right now.

core/tauri/src/window.rs
```diff
@@ -1395,39 +1395,28 @@ impl<R: Runtime> Window<R> {
...
-        if matches_window && matches_url && scope.is_none() {
-          scope.replace(s);
+      match self
+        .ipc_scope()
+        .remote_access_for(&self.window.label, &current_url)
+      {
+        Ok(scope) => Some(scope),
+        Err(e) => {
+          if e.matches_window {
+            scope_not_found_error_message = format!("Scope not defined for URL `{current_url}`");
```
tillmann-crabnebula: @lucasfernog can we point users to the documentation for this new feature, so people know what to search for?

> Scope not defined for URL `{current_url}`. If the domain should be able to access the IPC Layer please refer to  *dangerousRemoteUrlIpcAccess* in the *security* component of the Tauri configuration.

tooling/cli/schema.json
```diff
@@ -2606,12 +2606,12 @@
...
+        "dangerousRemoteUrlIpcAccess": {
+          "description": "Allow external domains to send command to Tauri.\n\nBy default, external domains do not have access to `window.__TAURI__`, which means they cannot communicate with the commands defined in Rust. This prevents attacks where an externally loaded malicious or compromised sites could start executing commands on the user's device.\n\nThis configuration allows a set of external domains to have access to the Tauri commands. Wildcards patterns are accepted and can be used to allow either all paths or a subset of paths for a domain.\n\n**WARNING:** Only use this option if you either have internal checks against malicious external sites or you can trust the allowed external sites. You application might be vulnerable to dangerous Tauri command related attacks otherwise.",
```
tillmann-crabnebula: Wildcards are not longer a thing by default, so we should remove this documentation part.
I would also add that the developer must ensure trust of the allowed domains during the lifetime of the desktop application as subdomain takeover is a huge risk.

tillmann-crabnebula (commented):
Minor documentation changes but implementation looks good to me :+1:

lucasfernog:
@tillmann-crabnebula I've pushed some changes to make this work with the isolation pattern.

I'm planning on merging this and release 1.3 next week if we can get all blog posts ready.

lucasfernog:
We had to change the approach here from "configuring remote URLs with glob patterns" to "configuring domains" for security reasons. If you need to allow ALL URLs, that is super dangerous and you'll need to do so by listening to the navigation event via `WindowBuilder::on_navigation` and manually adding the domain to the scope.

tweidinger (approved):

Sytten:
Thats fair, at least there is a way to do it. I will give it a try this week. When is it planned to be released?

lucasfernog:
We're working to release it this week.

1zun4secondary:
huuray

limitedmage:
Hello, will this be released for v1.x or will this have to wait until v2.0?

FabianLars:
It will be part of v1.3

luucasrb:
Can we have this feature in the `next` branch as a 2.0 feature? I would love to use it in my context, but I cannot downgrade to the 1.x.

FabianLars:
@luucasrb It already is in the next branch. iirc it was also part of alpha.9 but not 100% sure about that.

luucasrb:
Thanks, you are right. My confusion was that the `dangerousRemoteDomainIpcAccess` config is only in the v1 docs but not in the next docs.

JVariance:
> We had to change the approach here from "configuring remote URLs with glob patterns" to "configuring domains" for security reasons. If you need to allow ALL URLs, that is super dangerous and you'll need to do so by listening to the navigation event via `WindowBuilder::on_navigation` and manually adding the domain to the scope.

I want to add domains inside on_navigation, but I just don't succeed. Can you give me an example, how I can do it?

PR #5950, opened 2022-12-31
feat: allow specifying a resource map, closes #5844

### LINKED ISSUE #5844: [bug] cannot run sidecar/cmd from anywhere but src-tauri if it has shared libraries
### Describe the bug

running an executable that depends on other shared libraries won't work unless you place all the files/folders and the binary itself directly in the src-tauri folder and copy them as individual `resources`

### Reproduction

_No response_

### Expected behavior

should be able to run executables (with shared library deps) from a folder inside `src-tauri` otherwise the `src-tauri` directory becomes a huge mess

### Platform and versions

x86_64-pc-windows-msvc
tested in 1.2 and 2-alpha.0 both fail

### Stack trace

_No response_

### Additional context

_No response_

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


tooling/bundler/src/bundle/settings.rs
```diff
@@ -703,9 +709,14 @@ impl Settings {
...
-    match self.bundle_settings.resources {
-      Some(ref paths) => ResourcePaths::new(paths.as_slice(), true),
-      None => ResourcePaths::new(&[], true),
+    match (
+      &self.bundle_settings.resources,
+      &self.bundle_settings.resources_map,
+    ) {
+      (Some(paths), None) => ResourcePaths::new(paths.as_slice(), true),
+      (None, Some(map)) => ResourcePaths::from_map(map, true),
+      (Some(_), Some(_)) => panic!("cannot use both `resources` and `resources_map`"),
```
amrbashir, line 747: why panic here? seems like both can co-exist and merged into one final map

lucasfernog: Well the Tauri configuration does not allow both to be provided... so it's not worth implementing it until someone else is using the bundler.

amrbashir: cool

amrbashir (dismissed):

amrbashir (dismissed):
I updated the NSIS implementation and used `dunce` to simplify the path on WiX and NSIS

Other changes LGTM

amrbashir (dismissed):

amrbashir (approved):

Issue #6015, opened 2023-01-09
[feat] Dynamically control whether Menu, SystemMenu is displayed or not.


lencx:
### Describe the problem

Because the application needs to read from the configuration file whether the menu is displayed or not, but currently tauri does not control the parameters of menu by code, it needs to save them to variables and process them several times.

```rust
fn main() {
	// config file
	let has_menu = true;
	let has_tray_menu = true;
	let mut builder = tauri::Builder::default();
	if has_menu {
		builder = builder.menu(Menu::new());
	}
	if has_tray_menu {
		builder = builder.system_tray(SystemTray::new());
	}
	builder
		.run(tauri::generate_context!())
		.expect("error while running tauri application");
}
```

### Describe the solution you'd like

Is it possible to provide a way to set whether to show them or not without modifying the original api, for example by adding `show_menu` and `show_system_tray`.

```rust
fn main() {
	// config file
	let has_menu = true;
	let has_tray_menu = true;
	let mut builder = tauri::Builder::default()
		.menu(Menu::new())
		.system_tray(SystemTray::new())
		.show_menu(has_menu)
		.show_system_tray(has_tray_menu)
		.run(tauri::generate_context!())
		.expect("error while running tauri application");
}
```

tk103331:
Maybe it would be better to provide a method like this, **menu_if** for example


``` rust
    pub fn menu_if<F: Fn() -> Menu + Send + Sync + 'static>(mut self, show: bool, menu_init: F) -> Self {
        if show {
            let menu = menu_init();
            self.menu.replace(menu);
        }
        self
    }
```

This can control whether to initialize the menu, because if you don't want to display it, you don't need to initialize it.

And then call like this:

``` rust
    let show_menu = true; // condition
    Builder::default()
        .menu_if(show_menu, || Menu::new())
        .build();
```

amrbashir:
@tk103331 sounds good, feel free to open a PR for this

PR #6022, opened 2023-01-09
feat: dynamic settings menu (#6015)

lencx:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/core/tauri/src/app.rs
+++ b/core/tauri/src/app.rs
@@ -1391,7 +1391,7 @@ impl<R: Runtime> Builder<R> {
   ///     )),
   ///   ]));
   /// ```
-  pub fn menu_if<F>(mut self, is_show: bool, menu: F) -> Self
+  pub fn menu_if<F>(mut self, condition: bool, menu: F) -> Self
   where
     F: Fn() -> Menu + Send + Sync + 'static
   {
--- a/core/tauri/src/app.rs
+++ b/core/tauri/src/app.rs
@@ -1369,7 +1369,7 @@ impl<R: Runtime> Builder<R> {
     self
   }
 
-  /// Dynamically the menu to use on all windows.
+  /// Dynamically sets the menu to use on all windows.
   ///
   /// # Examples
   /// ```
--- a/core/tauri/src/app.rs
+++ b/core/tauri/src/app.rs
@@ -1328,7 +1328,7 @@ impl<R: Runtime> Builder<R> {
   ///
   /// let is_show = true;
   /// tauri::Builder::default()
-  ///   .system_tray_if(is_show, SystemTray::new().with_menu(
+  ///   .system_tray_if(is_show, || SystemTray::new().with_menu(
   ///     SystemTrayMenu::new()
   ///       .add_item(CustomMenuItem::new("quit", "Quit"))
   ///       .add_item(CustomMenuItem::new("open", "Open"))
@@ -1380,7 +1380,7 @@ impl<R: Runtime> Builder<R> {
   ///
   /// let is_show = true;
   /// tauri::Builder::default()
-  ///   .menu_if(is_show, Menu::with_items([
+  ///   .menu_if(is_show, || Menu::with_items([
   ///     MenuEntry::Submenu(Submenu::new(
   ///       "File",
   ///       Menu::with_items([
--- a/core/tauri/src/app.rs
+++ b/core/tauri/src/app.rs
@@ -1318,7 +1318,7 @@ impl<R: Runtime> Builder<R> {
     self
   }
 
-  /// Dynamically sets the given system tray to be built before the app runs.
+  /// Sets the given system tray to be built before the app runs if `condition` is `true`.
   ///
   /// Prefer the [`SystemTray#method.build`](crate::SystemTray#method.build) method to create the tray at runtime instead.
   ///
--- a/core/tauri/src/app.rs
+++ b/core/tauri/src/app.rs
@@ -1336,7 +1336,7 @@ impl<R: Runtime> Builder<R> {
   /// ```
   #[cfg(all(desktop, feature = "system-tray"))]
   #[cfg_attr(doc_cfg, doc(cfg(feature = "system-tray")))]
-  pub fn system_tray_if<F>(mut self, is_show: bool, tray_menu: F) -> Self
+  pub fn system_tray_if<F>(mut self, condition: bool, tray_menu: F) -> Self
   where
     F: Fn() -> tray::SystemTray + Send + Sync + 'static
   {
--- a/core/tauri/src/app.rs
+++ b/core/tauri/src/app.rs
@@ -1318,6 +1318,33 @@ impl<R: Runtime> Builder<R> {
     self
   }
 
+  /// Dynamically sets the given system tray to be built before the app runs.
+  ///
+  /// Prefer the [`SystemTray#method.build`](crate::SystemTray#method.build) method to create the tray at runtime instead.
+  ///
+  /// # Examples
+  /// ```
+  /// use tauri::{CustomMenuItem, SystemTray, SystemTrayMenu};
+  ///
+  /// let is_show = true;
+  /// tauri::Builder::default()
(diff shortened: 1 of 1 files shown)


core/tauri/src/app.rs
```diff
@@ -1318,6 +1318,33 @@ impl<R: Runtime> Builder<R> {
...
+  /// tauri::Builder::default()
+  ///   .system_tray_if(is_show, SystemTray::new().with_menu(
+  ///     SystemTrayMenu::new()
+  ///       .add_item(CustomMenuItem::new("quit", "Quit"))
+  ///       .add_item(CustomMenuItem::new("open", "Open"))
+  ///   ));
+  /// ```
+  #[cfg(all(desktop, feature = "system-tray"))]
+  #[cfg_attr(doc_cfg, doc(cfg(feature = "system-tray")))]
+  pub fn system_tray_if(mut self, is_show: bool, tray_menu: tray::SystemTray) -> Self {
```
amrbashir: the second argument should be a closure that returns a system tray

core/tauri/src/app.rs
```diff
@@ -1342,6 +1369,34 @@ impl<R: Runtime> Builder<R> {
...
+  ///       "File",
+  ///       Menu::with_items([
+  ///         MenuItem::CloseWindow.into(),
+  ///         #[cfg(target_os = "macos")]
+  ///         CustomMenuItem::new("hello", "Hello").into(),
+  ///       ]),
+  ///     )),
+  ///   ]));
+  /// ```
+  pub fn menu_if(mut self, is_show: bool, menu: Menu) -> Self {
```
amrbashir: the second argument should be a closure that returns a menu

JonasKruckenberg:
I really strongly dislike the naming scheme of `*_if` maybe we can come up with a better name here before merging this

amrbashir (changes requested):

amrbashir:
> I really strongly dislike the naming scheme of *_if maybe we can come up with a better name here before merging this

To me, it feels more idiomatic, and feels similar to `unwrap_or`. Also `druid` crate has a similar function https://docs.rs/druid/latest/druid/struct.MenuDesc.html#method.append_if

lencx:
> I really strongly dislike the naming scheme of `*_if` maybe we can come up with a better name here before merging this

How about `dyn_menu` and `dyn_system_tray`?

core/tauri/src/app.rs
lencx: @amrbashir Can I ask why closures are recommended here?

amrbashir: because it will be lazily evaluated if `is_show = true`, which means less memory consumption and allocations if `is_show = false`

lencx: Thanks for the answer, I'll modify it right away.

JonasKruckenberg:
> How about dyn_menu and dyn_system_tray?

Hmm to me this feels worse than `_if` maybe amr is indeed right and the original is not that bad 🤔 
`add_system_tray_if` sounds a bit nicer to me, but I guess it's redundant so don't worry

tk103331 (commented):
change the examples to use closure?

lencx:
> change the examples to use closure?

😅 I made a low-level mistake, false will not call the function, this is the most basic programming knowledge, I suddenly forgot.

FabianLars:
To be completely honest i don't see the value in this new api, but i also didn't see the problem in #6015. What's wrong with the standard builder pattern? (genuine question) You can do the same stuff with it already and this new api doesn't really feel rust-y too me 😅

lencx:
@FabianLars This pr is just a suggestion and can be close it if team members see little value in it. Because its existence does reduce some insignificant workload. 😅

amrbashir:
this is a mere quality of life improvement to turn this:
```rs
let mut builder = tauri::Builder::default();
if show_menu {
	builder = builder.menu(Menu::default);
}
builder.build().unwrap();
```
into this:
```rs
tauri::Builder::default()
	.menu_if(show_menu, Menu::default)
	.build()
	.unwrap()
```

core/tauri/src/app.rs
```diff
@@ -1318,6 +1318,36 @@ impl<R: Runtime> Builder<R> {
...
+  /// tauri::Builder::default()
+  ///   .system_tray_if(is_show, || SystemTray::new().with_menu(
+  ///     SystemTrayMenu::new()
+  ///       .add_item(CustomMenuItem::new("quit", "Quit"))
+  ///       .add_item(CustomMenuItem::new("open", "Open"))
+  ///   ));
+  /// ```
+  #[cfg(all(desktop, feature = "system-tray"))]
+  #[cfg_attr(doc_cfg, doc(cfg(feature = "system-tray")))]
+  pub fn system_tray_if<F>(mut self, is_show: bool, tray_menu: F) -> Self
```
amrbashir:
```suggestion
  pub fn system_tray_if<F>(mut self, condition: bool, tray_menu: F) -> Self
```

core/tauri/src/app.rs
```diff
@@ -1342,6 +1372,37 @@ impl<R: Runtime> Builder<R> {
...
+  ///       "File",
+  ///       Menu::with_items([
+  ///         MenuItem::CloseWindow.into(),
+  ///         #[cfg(target_os = "macos")]
+  ///         CustomMenuItem::new("hello", "Hello").into(),
+  ///       ]),
+  ///     )),
+  ///   ]));
+  /// ```
+  pub fn menu_if<F>(mut self, is_show: bool, menu: F) -> Self
```
amrbashir:
```suggestion
  pub fn menu_if<F>(mut self, condition: bool, menu: F) -> Self
```

core/tauri/src/app.rs
```diff
@@ -1318,6 +1318,36 @@ impl<R: Runtime> Builder<R> {
     self
   }
 
+  /// Dynamically sets the given system tray to be built before the app runs.
```
amrbashir:
```suggestion
  /// Sets the given system tray to be built before the app runs if `condition` is `true`.
```

core/tauri/src/app.rs
```diff
@@ -1342,6 +1372,37 @@ impl<R: Runtime> Builder<R> {
     self
   }
 
+  /// Dynamically sets the menu to use on all windows.
```
amrbashir:
```suggestion
  /// Sets the menu to use on all windows if `condition` is `true`.
```

amrbashir (changes requested):

FabianLars:
> this is a mere quality of life improvement to turn this:...

Yeah i got that, and i see the appeal, it's just something i feel like you don't often see in rust world. But if you guys like it then i'm fine with that. Just wanted to make you guys think about it a second time :)

JonasKruckenberg:
> this is a mere quality of life improvement to turn this:
> 
> ```rust
> let mut builder = tauri::Builder::default();
> if show_menu {
> 	builder = builder.menu(Menu::default);
> }
> builder.build().unwrap();
> ```
> 
> into this:
> 
> ```rust
> tauri::Builder::default()
> 	.menu_if(show_menu, Menu::default)
> 	.build()
> 	.unwrap()
> ```

I honestly don't think it's that big of a saving **and** thinking about it some more, this would open us up to adding conditional versions of almost *all* builder methods 🤔 So maybe @FabianLars has a point here and we should pass on this one.
Also removing something that mad it into the codebase is really hard. We should merge wisely

amrbashir:
> thinking about it some more, this would open us up to adding conditional versions of almost all builder methods

yeah right, didn't think about that!

lucasfernog:
I don't think we should do this tbh. As said earlier in the discussion, users can handle this by storing the builder in a variable. What could be nice in the future would be allowing `None` to be given as an argument, so you could do `.system_tray(if should_show { Some(tray) } else { None })`.

lencx:
This PR holds without breaking the existing api, and changing it to `Option` would indeed be a better way to go.

PR #6025, opened 2023-01-10
feature: disable mouse event when building windows on Linux, closes #5913

### LINKED ISSUE #5913: [feat] Disable mouse event when building windows on Linux
### Describe the problem

On linux, there might have burst mouse events periodically and this could block gtk's main eventloop.
For more information, please see https://github.com/tauri-apps/tao/issues/635

With latest tao patch, we can call [with_cursor_moved_event](https://docs.rs/tao/latest/tao/platform/unix/trait.WindowBuilderExtUnix.html#tymethod.with_cursor_moved_event) and set it to false as workaround. Mouse event isn't useful since webview itself can already detect such event.

### Alternatives considered

Other alternatives are described in https://github.com/tauri-apps/tao/issues/635.
Please comment there if you have more ideas.

### Additional context

_No response_

shogohida:
Signed-off-by: Shogo Hida <shogo.hida@gmail.com>

Fixes #5913 

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/core/tauri-runtime-wry/src/lib.rs
+++ b/core/tauri-runtime-wry/src/lib.rs
@@ -722,7 +722,8 @@ impl WindowBuilder for WindowBuilderWrapper {
       .always_on_top(config.always_on_top)
       .content_protected(config.content_protected)
       .skip_taskbar(config.skip_taskbar)
-      .theme(config.theme);
+      .theme(config.theme)
+      .disable_mouse_event(config.disable_mouse_event);
 
     #[cfg(target_os = "macos")]
     {
@@ -949,6 +950,12 @@ impl WindowBuilder for WindowBuilderWrapper {
 
   fn get_menu(&self) -> Option<&Menu> {
     self.menu.as_ref()
+  }
+
+  #[cfg(target_os = "linux"))]
+  fn disable_mouse_event(mut self, disable: bool) -> Self {
+    self.inner = self.inner.disable_mouse_event(disable);
+    self
   }
 }
 
--- a/core/tauri-runtime-wry/src/lib.rs
+++ b/core/tauri-runtime-wry/src/lib.rs
@@ -952,7 +952,7 @@ impl WindowBuilder for WindowBuilderWrapper {
     self.menu.as_ref()
   }
 
-  #[cfg(target_os = "linux"))]
+  #[cfg(target_os = "linux")]
   fn disable_mouse_event(mut self, disable: bool) -> Self {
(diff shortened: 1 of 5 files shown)


core/tauri-runtime-wry/src/lib.rs
```diff
@@ -722,7 +722,8 @@ impl WindowBuilder for WindowBuilderWrapper {
       .always_on_top(config.always_on_top)
       .content_protected(config.content_protected)
       .skip_taskbar(config.skip_taskbar)
-      .theme(config.theme);
+      .theme(config.theme)
+      .disable_mouse_event(config.disable_mouse_event);
```
wusyong: You should create a target specific block like below to call this method.

core/tauri-runtime-wry/src/lib.rs
```diff
@@ -950,6 +951,12 @@ impl WindowBuilder for WindowBuilderWrapper {
   fn get_menu(&self) -> Option<&Menu> {
     self.menu.as_ref()
   }
+
+  #[cfg(target_os = "linux")]
+  fn disable_mouse_event(mut self, disable: bool) -> Self {
```
wusyong: After thinking about it, I think there's no reason to expose a method for it. We should just disable it in `with_config`.

shogohida: Thanks for your comment! I made some changes according to it. Is my understanding correct? If not, please give me some suggestion of the code

https://github.com/tauri-apps/tauri/pull/6025/commits/5d1e22a28cb011771f8cac9eac034549e3169400

wusyong: Sorry this is not what I mean. I should comment in `config.rs` instead.
What you should do is:
- Remove `disable_mouse_event` in config.rs
- Bring this method back
- Add this trait method in `core/tauri-runtime/src/webview.rs`

And the code block you added should work.
Please at least run the build once on your device.
Also you need to add a change file to explain the change.

wusyong (dismissed):

wusyong (dismissed):

wusyong:
@lucasfernog CI failed with Windows builds which seems to be irrelevant to this PR. Could I merge this one?

lucasfernog (approved):

PR #6066, opened 2023-01-15
fix: removed greet from template

simonhyll:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
Since there's no `greet` command added to the template building fails. Either we remove `greet` as per this pull request, or we add a `greet` command, because it's very annoying to init a new project only to have to fix it before it can be built.

### DIFF
--- a/tooling/cli/templates/app/src-tauri/src/lib.rs
+++ b/tooling/cli/templates/app/src-tauri/src/lib.rs
@@ -1,7 +1,6 @@
 #[cfg_attr(mobile, tauri::mobile_entry_point)]
 pub fn run() {
   tauri::Builder::default()
-    .invoke_handler(tauri::generate_handler![])
     .run(tauri::generate_context!())
     .expect("error while running tauri application");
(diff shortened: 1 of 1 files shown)


tooling/cli/templates/app/src-tauri/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 #[cfg_attr(mobile, tauri::mobile_entry_point)]
 pub fn run() {
   tauri::Builder::default()
-    .invoke_handler(tauri::generate_handler![greet])
```
amrbashir, line 4: Go a bit further and remove the whole line

amrbashir (changes requested):

amrbashir (approved):

PR #6079, opened 2023-01-17
fix(api): compatible with ssg mode

yang4515:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
When using the ssg mode to package web applications, the code will run in the node environment without the window and navigator variables, which will cause packaging errors.

### DIFF
--- a/tooling/api/src/helpers/os-check.ts
+++ b/tooling/api/src/helpers/os-check.ts
@@ -5,15 +5,15 @@
 /** @ignore */
 
 function isLinux(): boolean {
-  return navigator.appVersion.includes('Linux')
+  return typeof navigator !== 'undefined' && navigator.appVersion.includes('Linux')
 }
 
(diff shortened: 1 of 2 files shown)


amrbashir:
but this will result in wrong values for variables depending on these methods, currently only `EOL` const in `os` module but I understand that just importing the os module will result in the error, so I think we should define `EOL` as a function and target this PR for v2 as for `appWindow` variable, it will be removed in v2 anyways and replaced with `getCurrent()`.

lucasfernog:
This was moved to the `os` plugin in https://github.com/tauri-apps/plugins-workspace so we need PR the proposed changes there.

Issue #2533, opened 2021-08-25
Compile error for async stateful commands that don't return a `Result`


probablykasper:
### **Describe the bug**
Stateful async commands that don't return a `Result` give a compiler error.

### **To Reproduce**
Example:
```rust
#[command]
pub async fn get_info(app: State<'_, Data>) -> bool {
  true
}
```
There's no error if you remove `async`, remove the `app` argument, or change the return type to a `Result` works.

### Platform and Versions
<details>
<summary>Details</summary>

```
Operating System - Mac OS, version 10.15.7 X64

Node.js environment
  Node.js - 14.16.0
  @tauri-apps/cli - 1.0.0-beta.10
  @tauri-apps/api - 1.0.0-beta.8

Global packages
  npm - 6.14.11
  yarn - 1.22.10

Rust environment
  rustc - 1.54.0
  cargo - 1.54.0

App directory structure
/node_modules
/src-tauri
/build
/.git
/src

App
  tauri.rs - 1.0.0-beta.8
  build-type - bundle
  CSP - default-src blob: data: filesystem: ws: wss: http: https: tauri: 'unsafe-eval' 'unsafe-inline' 'self' img-src: 'self'
  distDir - ../build
  devPath - http://localhost:8000
  framework - Svelte
```

</details>

### **Stack Trace**
```
error[E0597]: `__tauri_message__` does not live long enough
  --> src/cmd.rs:93:1
   |
93 |   #[command]
   |   ^^^^^^^^^-
   |   |        |
   |   |        `__tauri_message__` dropped here while still borrowed
   |   borrowed value does not live long enough
   |   argument requires that `__tauri_message__` is borrowed for `'static`
```

zsmatrix62:
same problem

ricardomatias:
Is there any interim solution?

FabianLars:
For now you just have to return a Result. If you don't want to return anything from the command it can be as simple as `Result<(), ()>` or in case of the OP's example it'd be `Result<bool, ()>`.

This doesn't change what you'll receive on the JavaScript side. Ok will resolve the Promise and Err will reject it.

willcrichton:
If you have an async function that takes a `State<'_, T>` as input and returns a `Result<(), String>`, what are the conditions under which Tauri will return an `Err`? Is it if the frontend tries to call two commands that both request `State<'_, T>` simultaneously?

FabianLars:
[...]

PR #6124, opened 2023-01-22
Improve compile errors for async commands without `Result` return type

### LINKED ISSUE #6733: [bug] Asynchronous #[tauri_command]s fail to compile with a borrowed value in the function constructor
### Describe the bug

When compiling async #[tauri_command]s, those with a borrowed value in the constructor such as &str fail to compile with a strange error that doesn't hint at this at all.

I'm not sure if this is how rust is meant to work, but I'm mainly writing this issue to see if 

- This can be fixed, so you can use &str in the constructor of async #[tauri_command]s
- If that's not possible, maybe a better error message, that addresses the issue can be generated (this would have saved me some time 😢 )
- If that's not possible, the [documentation](https://tauri.app/v1/guides/features/command/#async-commands) on the Tauri Website features a note about borrowed values, like &str being banned.

I'm unsure how to fix the error message, but I'm going to see if I can update the docs myself by submitting a PR?

*Edit: I have submitted the PR, which can be found here [https://github.com/tauri-apps/tauri-docs/pull/1214](https://github.com/tauri-apps/tauri-docs/pull/1214)*

### Reproduction

1. Create a new Tauri project, with any frontend, using the stable tauri and stable cargo/rustc.
2. In main.rs place this code:
``` 
#[tauri::command]
async fn test_function(borrowed_value_i_ruin_everything: &str) -> String {
    return format!("HELLO!")
}

fn main() {
    tauri::Builder::default()
        .invoke_handler(tauri::generate_handler![test_function])
} 
```
3. The program will fail to compile when running `cargo tauri dev`, citing: 
```
error[E0597]: __tauri_message__ does not live long enough 
  --> src\main.rs:1:1
   |
29 | #[tauri::command]
   | ^^^^^^^^^^^^^^^^-
   | |               |
   | |               __tauri_message__ dropped here while still borrowed
   | borrowed value does not live long enough
   | argument requires that __tauri_message__ is borrowed for 'static
   |
  ::: src\main.rs:8:25
   |
9  |         .invoke_handler(tauri::generate_handler![test_function])
   |                         -------------------------------------------------- in this macro invocation
   |
   = note: this error originates in the macro __cmd__http_request which comes from the expansion of the macro tauri::generate_handler`(in Nightly builds, run with -Z macro-backtrace for more info)
```

### Expected behavior

Functions with a borrowed value in the constructor, like &str, compile fine, or maybe a more helpful error message is generated explaining borrowed constructors not being allowed, because the one at current suggests an issue with the Tauri Macro.

### Platform and versions

```text
Environment
  › OS: Windows 10.0.19044 X64
  › Webview2: 112.0.1722.48
  › MSVC: 
  › Node.js: 16.17.0
  › npm: 8.15.0
  › pnpm: Not installed!
  › yarn: Not installed!
  › rustup: 1.25.1
  › rustc: 1.65.0
  › cargo: 1.65.0

Packages
  › @tauri-apps/cli [NPM]: 1.2.0 (outdated, latest: 1.2.3)
  › @tauri-apps/api [NPM]: 1.2.0
  › tauri [RUST]: 1.2.4,
  › tauri-build [RUST]: 1.2.1,
  › tao [RUST]: 0.15.8,
  › wry [RUST]: 0.23.4,

App
  › build-type: bundle
  › CSP: unset
  › distDir: ../dist
  › framework: React

App directory structure
  ├─ dist
  ├─ node_modules
  ├─ public
  ├─ src
  └─ src-tauri
```

### Stack trace

```
error[E0597]: __tauri_message__ does not live long enough 
  --> src\main.rs:1:1
   |
29 | #[tauri::command]
   | ^^^^^^^^^^^^^^^^-
   | |               |
   | |               __tauri_message__ dropped here while still borrowed
   | borrowed value does not live long enough
   | argument requires that __tauri_message__ is borrowed for 'static
   |
  ::: src\main.rs:8:25
   |
9  |         .invoke_handler(tauri::generate_handler![test_function])
   |                         -------------------------------------------------- in this macro invocation
   |
   = note: this error originates in the macro __cmd__http_request which comes from the expansion of the macro tauri::generate_handler`(in Nightly builds, run with -Z macro-backtrace for more info)
```

### Additional context

Thank you for your time in reviewing this issue.

Also note that in the information section, I ran `cargo tauri info`, but it also outputted information about my NPM Tauri CLI installation, which is outdated as I no longer use it (in favour of the Cargo Tauri CLI, which was used to reproduce this bug).

Additionally, the information section states I don't have Visual Studio installed. This is untrue. I don't have a Visual Studio installed that contains the graphical editor, I have *Visual Studio Build Tools 2022*, enough to correctly compile rust code.

Pascal-So:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

Issue #2533 points out that async commands currently have to return a `Result`, but this is not very obvious from the compile error that is shown to the user.

```rust
#[tauri::command]
async fn myAsyncCommand(name: &str) {
    tokio::time::sleep(Duration::from_secs_f32(1.)).await;
}
```

The above code resulted in the following error, which was not comprehensible to me when I encountered this while writing my first tauri app:

```
error[E0597]: `__tauri_message__` does not live long enough
  --> src/main.rs:15:1
   |
15 | #[tauri::command]
   | ^^^^^^^^^^^^^^^^-
   | |               |
   | |               `__tauri_message__` dropped here while still borrowed
   | borrowed value does not live long enough
   | argument requires that `__tauri_message__` is borrowed for `'static`
...
29 |         .invoke_handler(tauri::generate_handler![myAsyncCommand])
   |                         ---------------------------------------- in this macro invocation
   |
   = note: this error originates in the macro `__cmd__myAsyncCommand` which comes from the expansion of the macro `tauri::generate_handler` (in Nightly builds, run with -Z macro-backtrace for more info)
```

With the change introduced in this PR, the error now appears as follows:

```
error: async commands that contain references as inputs must return a `Result`
  --> src/main.rs:16:25
   |
16 | async fn myAsyncCommand(name: &str) {
   |                         ^^^^

error: cannot find macro `__cmd__myAsyncCommand` in this scope
  --> src/main.rs:27:50
   |
27 |         .invoke_handler(tauri::generate_handler![myAsyncCommand]);
   |                                                  ^^^^^^^^^^^^^^
```

This is what the user sees in their editor:

image

The problem with `State` described in the issue is also caught:

image

Errors are also generated on `&'static` references, because they also generate the same `__tauri_message__ does not live long enough` when tested without this change.

This change probably doesn't catch all possible errors, but it should catch the most common ones. It looks for reference types in the arguments, as well as for types that contain lifetime arguments. Can anyone think of a situation where this check would be too strict? The intention here is to rather catch too little than too much, because we don't want this to be a breaking change.

This PR should be reverted as soon as #2533 is resolved.

### DIFF
--- a/.changes/improve-async-cmd-error-message.md
+++ b/.changes/improve-async-cmd-error-message.md
@@ -1,0 +1,5 @@
+---
+"tauri-macros": 'patch:enhance'
+---
+
+Improve compiler error message when generating an async command that has a reference input and don't return a Result.
--- a/core/tauri-macros/src/command/wrapper.rs
+++ b/core/tauri-macros/src/command/wrapper.rs
@@ -5,7 +5,7 @@
 use heck::{ToLowerCamelCase, ToSnakeCase};
 use proc_macro::TokenStream;
 use proc_macro2::TokenStream as TokenStream2;
-use quote::{format_ident, quote};
+use quote::{format_ident, quote, quote_spanned};
 use syn::{
   ext::IdentExt,
   parse::{Parse, ParseStream},
@@ -103,6 +103,63 @@ pub fn wrapper(attributes: TokenStream, item: TokenStream) -> TokenStream {
     resolver: format_ident!("__tauri_resolver__"),
   };
 
+  // Tauri currently doesn't support async commands that take a reference as input and don't return
+  // a result. See: https://github.com/tauri-apps/tauri/issues/2533
+  //
+  // For now, we provide an informative error message to the user in that case. Once #2533 is
+  // resolved, this check can be removed.
+  let mut async_command_check = TokenStream2::new();
+  if function.sig.asyncness.is_some() {
+    // This check won't catch all possible problems but it should catch the most common ones.
+    let mut ref_argument_span = None;
+
+    for arg in &function.sig.inputs {
+      if let syn::FnArg::Typed(pat) = arg {
+        match &*pat.ty {
+          syn::Type::Reference(_) => {
+            ref_argument_span = Some(pat.span());
+          }
+          syn::Type::Path(path) => {
+            // Check if the type contains a lifetime argument
+            let last = path.path.segments.last().unwrap();
+            if let syn::PathArguments::AngleBracketed(args) = &last.arguments {
+              if args
+                .args
+                .iter()
+                .any(|arg| matches!(arg, syn::GenericArgument::Lifetime(_)))
+              {
+                ref_argument_span = Some(pat.span());
+              }
+            }
+          }
+          _ => {}
+        }
+
+        if let Some(span) = ref_argument_span {
+          if let syn::ReturnType::Type(_, return_type) = &function.sig.output {
+            // To check if the return type is `Result` we require it to check a trait that is
+            // only implemented by `Result`. That way we don't exclude renamed result types
+            // which we wouldn't otherwise be able to detect purely from the token stream.
+            // The "error message" displayed to the user is simply the trait name.
+            async_command_check = quote_spanned! {return_type.span() =>
+              #[allow(unreachable_code, clippy::diverging_sub_expression)]
+              const _: () = if false {
+                trait AsyncCommandMustReturnResult {}
+                impl<A, B> AsyncCommandMustReturnResult for Result<A, B> {}
+                let _check: #return_type = unreachable!();
+                let _: &dyn AsyncCommandMustReturnResult = &_check;
+              };
+            };
+          } else {
+            return quote_spanned! {
+              span => compile_error!("async commands that contain references as inputs must return a `Result`");
+            }.into();
+          }
+        }
+      }
+    }
+  }
+
   // body to the command wrapper or a `compile_error!` of an error occurred while parsing it.
   let body = syn::parse::<WrapperAttributes>(attributes)
     .map(|mut attrs| {
@@ -121,6 +178,8 @@ pub fn wrapper(attributes: TokenStream, item: TokenStream) -> TokenStream {
 
   // Rely on rust 2018 edition to allow importing a macro from a path.
   quote!(
+    #async_command_check
+
     #function
 
     #maybe_macro_export


Pascal-So:
ah, looks like I missed something, I'll check it out tomorrow

Pascal-So:
In my original version of the PR I made the mistake of returning an error irrespective of whether or not any of the arguments contained any references, should be fixed now. The (non-ignored)tests and clippy now run without errors on my linux.

Can anyone see a situation that would cause this error to occur when it shouldn't?

Pascal-So:
I just thought of a potential problem: do you think anyone would rename their result type and have a command returning `MyResult`? This currently wouldn't work with this PR

amrbashir:
> do you think anyone would rename their result type and have a command returning MyResult

I think it is somewhat common that projects tend to define their own `Result` type combined with an `Error` type like this:
```rs
enum Error {
}

type Result<T> = std::result::Result<T, Error>;
```

Pascal-So:
> ```rust
> type Result<T> = std::result::Result<T, Error>;
> ```

this wouldn't be a problem since we only check if the type is called `Result`, and that's about all we really can do at the token level. This common pattern wouldn't pose a problem, problems would only occur when the type is renamed to something like `FooResult` which I don't remember seeing in practice anywhere.

amrbashir:
> which I don't remember seeing in practice anywhere.

true but still there is a chance someone uses that.

Pascal-So:
Yeah you're right, the risk of this being a breaking change for someone is too high. I thought about it some more and I think the simplest solution would be to just show the error when no return type is used.

This again restricts the usefulness to the point where we might ask if we should just drop the idea completely, but at least for the case that I encountered while first using tauri, it would have saved me some digging around in the tauri issues to find out what's wrong, because I had an async function taking `&str` that doesn't return anything.

What do you think, should we go ahead and add it only for the trivial case, plus maybe some primitive return types or return types containing references?

amrbashir:
A workaround would be to allow users to specify their result type name, for example: 
```rs
type MyResult<T> = Result<T, MyError>;

#[tauri::command(result = MyResult)]
async fn command(arg: &str) -> MyResult<()> {
	Ok(())
}
```

Pascal-So:
> ```rust
> #[tauri::command(result = MyResult)]
> ```

This would still be a breaking change since users would have to update their attributes when upgrading tauri, otherwise their code might not compile anymore.

In the meantime I've been thinking about this a bit more and questioned one of my above statements:

> [...] and that's about all we really can do at the token level.

While it's true that we can't do more at the token level, I thought of some other fun tricks to detect if the user is returning a `Result` in their command. Given this input code:

```rust
type MyRenamedResult<T> = Result<T, String>;

#[tauri::command]
async fn command_name(a: &str) -> MyRenamedResult<()> {
```

the macro now generates an additional snippet that looks roughly like this:

```rust
const _: () = {
  trait AsyncCommandMustReturnResult {}
  impl<A, B> AsyncCommandMustReturnResult for Result<A, B> {};
  const fn check<T: AsyncCommandMustReturnResult>() {}
  const _: () = {
    check::<MyRenamedResult<()>>();
  };
};
```

This should be 100% reliable in detecting if the command returns a `Result`. This snippet is only generated when  the command is an async functijon and takes a reference or a type with generic lifetime argument as input.

If the user tries to return a different type, they see this compiler error:

```
error[E0277]: the trait bound `Vec<()>: AsyncCommandMustReturnResult` is not satisfied
  --> src/main.rs:48:35
   |
48 | async fn command_name(a: &str) -> Vec<()> {
   |                                   ^^^^^^^ the trait `AsyncCommandMustReturnResult` is not implemented for `Vec<()>`
   |
   = help: the trait `AsyncCommandMustReturnResult` is implemented for `Result<A, B>`
```

This is not perfect but in my opinion much more helpful than some lifetime errors.

For the case where the command doesn't return anything we can still show the complete message I added previously.

```
error: async commands that contain references as inputs must return a `Result`
  --> src/main.rs:48:23
   |
48 | async fn command_name(a: &str) {
   |                       ^
```

With these changes I believe there should be zero false positives, i.e. this is not a breaking change, and I can't think of any way to make the error messages more readable. Unfortunately we can't use the `rustc_on_unimplemented` attribute outside of the compiler, that would be the only way to improve this that I could think of.

chippers (dismissed):

JonasKruckenberg (dismissed):
That trait idea is smart as hell! Though this whole ordeal kinda reinforces why I want to get rid of the command macro entirely 😅
great work though

Pascal-So:
Didn't realize that the trick I used above doesn't match tauri's MSRV 1.59, I now changed it to use a different method to check the trait bound. The generated check now looks like this:

```rust
#[allow(unreachable_code)]
const _: () = if false {
  trait AsyncCommandMustReturnResult {}
  impl<A, B> AsyncCommandMustReturnResult for Result<A, B> {}
  let _check: Vec<()> = unreachable!();
  let _: &dyn AsyncCommandMustReturnResult = &_check;
};

```

The generated error now looks like this:

```
error[E0277]: the trait bound `Vec<()>: AsyncCommandMustReturnResult` is not satisfied
  --> src/main.rs:38:35
   |
38 | async fn command_name(a: &str) -> Vec<()> {
   |                                   ^^^ the trait `AsyncCommandMustReturnResult` is not implemented for `Vec<()>`
   |
   = help: the trait `AsyncCommandMustReturnResult` is implemented for `Result<A, B>`
   = note: required for the cast from `Vec<()>` to the object type `dyn AsyncCommandMustReturnResult`
```

Pascal-So:
@JonasKruckenberg Thanks for the kind words!

> this whole ordeal kinda reinforces why I want to get rid of the command macro entirely

let me know if help is needed on this process to get rid of the macro, I'm looking forward to the moment where my hack in this PR here becomes obsolete :D

Pascal-So:
hmm looks like the new miniz_oxide release is the problem, if I delete my Cargo.lock and add `miniz_oxide = "=0.6.2"` to a Cargo.toml then the tests pass on 1.59 on my machine..

FabianLars:
it's fine to lock it in Cargo.toml and commit it. not the first dependency we have to lock because of the 1.59 requirement :/

Pascal-So:
looks like the issue is already fixed, miniz_oxide 0.6.4 has been yanked. I'll remove my last commit again where I fixed the version to 0.6.2

Pascal-So:
this time around I remembered to enable all the relevant features when testing locally :sweat_smile: I think with this os_info fix the 1.59 builds should now all work, unless it's something that I can't test locally on linux :D

Pascal-So:
:facepalm: didn't realize that not all of the fails on the last run were from 1.59 compatibility.. sorry for the back and forth on this

IntegralPilot:
Yes!! This is so amazing, wish this was in the current version of Tauri as it would have saved lots of time for me.

Pascal-So:
Hi all, I completely forgot that this PR was still open. Since some of the things I had to change in Cargo.toml now also happened on the dev branch, I rebased my work on to the current state of the dev branch.

Everything else is exactly as we left it last time, from my side this is ready to merge (assuming I didn't mess anything up again according to CI :smile:)

Pascal-So:
aaaand of course I forgot to sign the commit before, now it's *actually* ready :smile:

lucasfernog (approved):

lucasfernog:
This is awesome, thanks!

PR #6160, opened 2023-01-28
feat(cli): expose a function to kill dev child

amrbashir:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

Sometimes while using the cli.js API from a JS script, a zombie process is left after killing the JS script using <kbd>ctrl+c</kbd>. I can't reproduce it reliably but having an API to kill the dev app process, the script could execute it upon exit:
```js
process.on('SIGTERM', () => killDevApp());
process.on('exit', () => killDevApp());
```


lucasfernog:
I think we should try to do it on our end instead. Or maybe the JS script could just listen to ctrl+c and kill the CLI process manually?

amrbashir:
> I think we should try to do it on our end instead.

We could do that in cli.js but we'd still need to expose the function through NAPI

> Or maybe the JS script could just listen to ctrl+c and kill the CLI process manually?

Again, we still need to expose a function to kill the process or at least the PID.

lucasfernog:
I still couldn't reproduce the issue :( I actually wanted to listen to the signal on the Rust code and let it handle it instead.

amrbashir:
Yeah I can't really reproduce it reliably but I think it happens only after you triggered the CLI hot-reload at least once or twice.

PR #6179, opened 2023-02-01
fix 1.67 clippy warnings (compatible with MSRV)

chippers:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [x] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
I ran `clippy` on `1.67` and found that non-interpolated `format!` calls are now a warning. I updated every single `format!` call my search&find could gather, behavior _should_ be the same. Not sure what the schema changes are about.

### DIFF
--- a/core/config-schema/schema.json
+++ b/core/config-schema/schema.json
@@ -720,47 +720,19 @@
     },
     "Theme": {
       "description": "System theme.",
-      "oneOf": [
-        {
-          "description": "Light theme.",
-          "type": "string",
-          "enum": [
-            "Light"
-          ]
-        },
-        {
-          "description": "Dark theme.",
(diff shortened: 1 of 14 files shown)


chippers:
looking at the CI report, it looks like a dependency (or transitive dependency) requires 1.60 now, bumping from 1.59.

FabianLars:
which is a bit unfortunate since os_info still seems to compile&work fine on 1.59 😢

lucasfernog:
Time to pin another dependency, oh boy.

lucasfernog (approved):

PR #6236, opened 2023-02-10
fix: safe check window is undefined, closes #6226

### LINKED ISSUE #6226: @tauri-apps/api@1.2.0, "__TAURI_METADATA__" in window cause ReferenceError: window is not defined
https://github.com/tauri-apps/tauri/blob/50576a5e1e83978882d90aca280809da2235a8d5/tooling/api/src/window.ts#L2033

When running headless tests, this might be troublesome.

Should check `typeof window !== 'undefined'` first.

linonetwo:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/tooling/api/src/helpers/os-check.ts
+++ b/tooling/api/src/helpers/os-check.ts
@@ -5,15 +5,15 @@
 /** @ignore */
 
 function isLinux(): boolean {
-  return navigator.appVersion.includes('Linux')
+  return typeof navigator !== 'undefined' && navigator.appVersion.includes('Linux')
 }
 
 function isWindows(): boolean {
-  return navigator.appVersion.includes('Win')
+  return typeof navigator !== 'undefined' && navigator.appVersion.includes('Win')
 }
 
 function isMacOS(): boolean {
-  return navigator.appVersion.includes('Mac')
+  return typeof navigator !== 'undefined' && navigator.appVersion.includes('Mac')
 }
 
 export { isLinux, isWindows, isMacOS }
--- a/tooling/api/src/window.ts
+++ b/tooling/api/src/window.ts
@@ -2030,7 +2030,7 @@ class WebviewWindow extends WindowManager {
 
 /** The WebviewWindow for the current window. */
 let appWindow: WebviewWindow
-if ('__TAURI_METADATA__' in window) {
+if (typeof window !== 'undefined' && '__TAURI_METADATA__' in window) {
   appWindow = new WebviewWindow(
     window.__TAURI_METADATA__.__currentWindow.label,
     {


FabianLars:
Thanks for contributing!

At least the os-check change needs a bit more thinking imho. Frameworks like Nextjs currently fail to build if the items that use os-check are imported in a server-side context. This is actually somewhat of a good thing because it makes sure the dev properly imports them before shipping the app. Now if the error is silenced, it will be interpreted as if all functions return false for variables like `EOL`, `sep`, and `delimeter` which are defined on import, not when you access them.

(Actually not sure if it was really nextjs that would interpret the vars incorrectly or some other ssr-first frameworks, my last test is quite some time ago)

linonetwo:
I'm using pnpm patch in my project https://github.com/toeverything/AFFiNE/blob/master/patches/%40tauri-apps__api%401.2.0.patch , so I'm not hurrying on this, we can discuss about a proper solution on this.

You mean this will warn Nextjs user when they use tauri API package in their server side code? We are also using Nextjs https://github.com/toeverything/AFFiNE/blob/master/packages/app/src/pages/_app.tsx 

And our problem is when running e2e test with playwright `npx playwright test --config=playwright.config.unit.ts`, it somehow requires all the dependencies, including `@tauri-apps/api`. This is similar to the case you said about including tauri API package in their server side code. Maybe it is my usage that is wrong?

linonetwo:
If my usage is wrong, I also hope I can receive a warning. But maybe we can do this in an official way (using actual `console.warn`) instead of the erlang style let-it-crash.

(And maybe I will ignore the warning because I'm running tests and my tests passed...)

FabianLars:
> You mean this will warn Nextjs user when they use tauri API package in their server side code?

Well, it won't just warn them but straight up fail if you do something like this for example

```ts
import { sep } from "@tauri-apps/api/path";
```

the issue with sep is that it's defined then on import, depending on if it detected a windows environment. With your conditional check `isWindows()` will always return false (or undefined or whatever) if imported like this, which _should_ mean that sep is incorrectly defined when running on Windows (using the Unix sep instead).

> If my usage is wrong, I also hope I can receive a warning. But maybe we can do this in an official way (using actual console.warn) instead of the erlang style let-it-crash.

yeah well, in tests a warning, or even silently ignoring it should be fine in many cases i guess, i'm talking about actually deployed apps.

FabianLars:
i think ideally we'd just remove appWindow, sep, delimeter, and EOL and replace them all with a function in tauri 2.0 (at least for appWindow this is actually planned) - that would at least fix the usages in actual apps, for testing environments this may not be enough idk

linonetwo:
Or let tauri pass this information to js from rust side, like how I can easily do this in electron [https://github.com/tiddly-gittly/TidGi-Desktop/blob/91fd128a68e27008abdf7aab3107a05c908de223/src/helpers/system.ts](https://github.com/tiddly-gittly/TidGi-Desktop/blob/91fd128a68e27008abdf7aab3107a05c908de223/src/helpers/system.ts) ?

linonetwo:
> replace them all with a function in tauri 2.0

I haven't use any of `appWindow, sep, delimeter, and EOL` , but I can try this refactor now, is there any reason we have to wait until 2.0? And when will the 2.0 be?

FabianLars:
> Or let tauri pass this information to js from rust side

Wouldn't that have a similar issue? They'd be only defined if running inside a tauri window (where it has access to the js `window` var).

> is there any reason we have to wait until 2.0?

It would be a breaking change for all users on 1.X if we'd remove existing variables. And without actually removing them we don't fix the existing import errors

amrbashir:
closing as this has been fixed in v2 and there is no plans to fix in v1 as it will introduce breaking changes. Thanks for contributions anyway.

PR #6260, opened 2023-02-13
fix(docs): mark internal interfaces as such to deduplicate markdown output

FabianLars:
quick'n'dirty fix to unbloat https://tauri.app/v1/api/js/window/ a bit. windowmanager and webviewwindowhandle aren't used in a public interface anyway so this change should be fine / should not break anything.

I also thought about marking it as `@internal` or something instead of `@ignore`, to make the intend more clear but then i noticed that we use ignore everywhere so i went with that.

We have a similar issue with EventEmitter and Command, but ignoring EventEmitter causes issues as it's referenced by Command.stdout & stderr. I'm open to ideas about that, but would otherwise postpone that to the astro rewrite.

- [ ] Bugfix
- [ ] Feature
- [x] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/tooling/api/src/window.ts
+++ b/tooling/api/src/window.ts
@@ -307,6 +307,7 @@ export type WindowLabel = string
 /**
  * A webview window handle allows emitting and listening to events from the backend that are tied to the window.
  *
+ * @ignore
  * @since 1.0.0
  */
 class WebviewWindowHandle {
@@ -406,6 +407,7 @@ class WebviewWindowHandle {
     return emit(event, this.label, payload)
   }
 
+  /** @ignore */
   _handleTauriEvent<T>(event: string, handler: EventCallback<T>): boolean {
     if (localTauriEvents.includes(event)) {
       if (!(event in this.listeners)) {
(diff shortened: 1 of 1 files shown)


FabianLars:
actually, i'd like to know what @lorenzolewis thinks. don't want to interfere with the astro rewrite. If that's the case i could inject these tags when `tauri-docs` builds the ast.

lorenzolewis:
I'm going to be following the TypeDoc API as closely as possible there when I migrate from the MD-generated method. So if the goal is to not render it then I'd use `@ignore` or `@hidden` like suggested in https://typedoc.org/tags/ignore/

I believe `@internal` is normally rendered in docs by default (there's a flag for it https://typedoc.org/tags/internal/), but ignore or hidden would be better.

In general, I'm going to have as little logic in Astro as possible so that everything is driven directly by the AST

lucasfernog (approved):

PR #6262, opened 2023-02-13
Commit JS AST, update `check-generated-file` checkout action

lorenzolewis:

- [ ] Bugfix
- [ ] Feature
- [x] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

This will start committing the generated JS/TS AST to the repo so that it can be more easily used downstream (tauri-docs). This also updates the checkout action to v3. I tested this to make sure it works on a fork (https://github.com/lorenzolewis/tauri/actions/runs/4166698015), have a few other runs there as well but they were for the below (that didn't make it in)

I did some tests to run a check that the JSON is updated when needed, but the content of the file contains commit hashes for when it was generated which will naturally change from one commit to another. Not sure how to do a basic check here but ideally you'd have CI throw an error if generate-docs wasn't updated and needed to be.


FabianLars:
we should add this to .changes/config.json too so that it's generated together with the changelog.

Also, do we only care about released versions or also dev/next? If we only want it for released versions we could exclude it from the generated-files check and only let covector generate a new one.

lorenzolewis:
> we should add this to .changes/config.json too so that it's generated together with the changelog.

That would be fantastic to make sure it's updated, didn't know that was an option (and don't know how to do it either).


> Also, do we only care about released versions or also dev/next? If we only want it for released versions we could exclude it from the generated-files check and only let covector generate a new one.

I'd like to also do it for whatever the alpha/beta releases get tagged under. I haven't figured out the exact mechanism to drive it on Astro yet but I'm thinking of having a config that you can specify a specific ref for each version the website generates (e.g. `tauri.app/api/core-js/1` is generated from `@tauri-v1.2.4`, `tauri.app/api/core-js/next` is generated from `@tauri-v2.0.0-alpha.3`, or if we maintain `@tauri-v1` and `@tauri-v2` tags it could use those instead).

FabianLars:
> I'd like to also do it for whatever the alpha/beta releases get tagged under

yeah these still count as releases, as-in they're managed by covector and published to npm/cratesio - i meant more like arbitrary git commits sry.

but ig we could also add the ast generator to the `yarn build` command since that needs to be done on each change too.

> That would be fantastic to make sure it's updated, didn't know that was an option (and don't know how to do it either).

i _think_ adding this small snippet into here https://github.com/tauri-apps/tauri/blob/dev/.changes/config.json#L126 should be enough already
```json
"postversion": [ "yarn install", "yarn generate-docs" ]
```
not entirely sure as all that covector stuff is still mostly over my head...

lorenzolewis:
No, I don't think arbitrary git commits would need this done.

Adding to the build command could be good to simplify the process if that's okay with everyone to do.

lorenzolewis:
I added the `generate-docs` to `yarn build` and did a quick run locally to make sure it works. I'm not sure if/how to test it with the covector process so @tauri-apps/wg-devops might need to advise on that. 

Is there a branch to add this to `1.x` so that the same mechanism can be used for v1 and v2?

FabianLars:
the dev branch is currently the 1.x branch. v2 is still on `next`

lorenzolewis:
> the dev branch is currently the 1.x branch. v2 is still on `next`

Cool, so if this gets the thumbs up I can do a similar PR for `next` as well.

lorenzolewis:
bump @tauri-apps/wg-devops @tauri-apps/wg-devops

amrbashir (changes requested):
You need to also ignore the file in .prettierignore in the repo root

lorenzolewis:
> You need to also ignore the file in .prettierignore in the repo root

Should be added in now

amrbashir (dismissed):

lucasfernog (approved):

lorenzolewis:
Should I do the same in `next` or will `dev` get merged into `next`?

amrbashir:
We need to remove `js-api.json` from the generated check list, it will always fail the check since it contains references to the current commit hash, see #6374, which means it will infinitely fail

PR #6283, opened 2023-02-15
feat(cli): add --port, closes #6186

### LINKED ISSUE #6186: Starting multiple apps in dev mode uses same server (127.0.0.1:1430).
### Describe the bug

Hi, I am just starting with Tauri.. I am having an issue with multiple apps, when i am running multiple apps in dev mode, all the apps are sharing the same ui, instead of its own ui.. What i noticed was all the apps frontend communicates to the same server, which is nothing but the first server started while running the first app..  Also the other apps, which are started after first app failing to start app server with following error.. 
 `error binding to 127.0.0.1:1430..` 
 looks like tauri always uses a static server config..

Is there anyway we can set the server host & port while running `cargo tauri dev`
Something like below should help
`cargo tauri dev --host <host-ip/name> --port <port-no>`

### Reproduction

1. Create 2 or more tauri apps using `cargo create-tauri-app`
2. Go ahead & update one of the apps' index.html so that it is now same from the other ones.
3. Now start the apps one by one, in dev mode `cargo tauri dev`. 

We will always see the same ui rendered for all the apps, also check console for the below error for all the apps started after first app. 

  `error binding to 127.0.0.1:1430..` 

Since the port is already bind to first app server, starting from 2nd app the servers failed to start and clients start rendering the ui content from first app.

### Expected behavior

Expect to see the ui specific to the app, also either server should start with different port instead of static one or we should have option of starting the server with different port as below.

`cargo tauri dev --host <host-ip/name> --port <port-no>`

### Platform and versions

Environment
  › OS: Mac OS 13.1.0 X64
  › Node.js: 16.18.0
  › npm: 8.19.2
  › pnpm: Not installed!
  › yarn: Not installed!
  › rustup: 1.25.2
  › rustc: 1.67.0
  › cargo: 1.67.0

Packages
  › @tauri-apps/cli [NPM]: 1.2.3
  › @tauri-apps/api [NPM]: Not installed!
  › tauri [RUST]: 1.2.4,
  › tauri-build [RUST]: 1.2.1,
  › tao [RUST]: 0.15.8,
  › wry [RUST]: 0.23.4,

App
  › build-type: bundle
  › CSP: unset
  › distDir: ../src
  › devPath: ../src
package.json not found

App directory structure
  ├─ src-tauri
  ├─ .vscode
  └─ src

### Stack trace

```shell
thread '<unnamed>' panicked at 'error binding to 127.0.0.1:1430: error creating server listener: Address already in use (os error 48)', /Users/prabhugopal/.cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.23/src/server/server.rs:79:13
        Info Watching /Users/prabhugopal/Learning/tauri/hello-next/src-tauri for changes...
   Compiling hello-next v0.0.0 (/Users/prabhugopal/Learning/tauri/hello-next/src-tauri)
    Finished dev [unoptimized + debuginfo] target(s) in 1.24s
```

### Additional context

_No response_

amrbashir:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/tooling/cli/src/dev.rs
+++ b/tooling/cli/src/dev.rs
@@ -65,6 +65,7 @@ pub struct Options {
   #[clap(long)]
   pub no_dev_server: bool,
   /// Specify port for the dev server for static files. Defaults to 1430
+  /// Can also be set using `TAURI_DEV_SERVER_PORT` env var.
   #[clap(long)]
   pub port: Option<u16>,
 }
@@ -229,7 +230,7 @@ fn command_internal(mut options: Options) -> Result<()> {
       use crate::helpers::web_dev_server::start_dev_server;
       if path.exists() {
         let path = path.canonicalize()?;
-        let server_url = start_dev_server(path, options.dev_server_port);
+        let server_url = start_dev_server(path, options.port);
         let server_url = format!("http://{server_url}");
         dev_path = AppUrl::Url(WindowUrl::External(server_url.parse().unwrap()));
 
--- a/tooling/cli/src/dev.rs
+++ b/tooling/cli/src/dev.rs
@@ -66,7 +66,7 @@ pub struct Options {
   pub no_dev_server: bool,
   /// Specify port for the dev server for static files. Defaults to 1430
   #[clap(long)]
-  pub dev_server_port: Option<u16>,
+  pub port: Option<u16>,
 }
 
 pub fn command(options: Options) -> Result<()> {
--- a/tooling/cli/src/dev.rs
+++ b/tooling/cli/src/dev.rs
@@ -230,7 +230,8 @@ fn command_internal(mut options: Options) -> Result<()> {
       if path.exists() {
         let path = path.canonicalize()?;
         let server_url = start_dev_server(path, options.dev_server_port);
-        dev_path = AppUrl::Url(WindowUrl::External(server_url.to_string().parse().unwrap()));
+        let server_url = format!("http://{server_url}");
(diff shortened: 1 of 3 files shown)


tooling/cli/src/dev.rs
```diff
@@ -64,6 +64,9 @@ pub struct Options {
   /// Disable the dev server for static files.
   #[clap(long)]
   pub no_dev_server: bool,
+  /// Specify port for the dev server for static files. Defaults to 1430
+  #[clap(long)]
+  pub dev_server_port: Option<u16>,
```
lucasfernog: Can we name it `port` instead? Also we should move the default here so the docs is always in sync. `#[clap(long, default_value = 1430]`

amrbashir: `port` seems a bit vague, don't you think?

As for the default, I chose not to use clap `default_value` because I thought that `--dev-server-port` have precedence over the env var and if `None` is passed I would check `TAURI_DEV_SERVER_PORT`, or should it be the other way around?

lucasfernog: You're right about the default value, my bad :D I didn't see the precedence.

About the argument name... well I don't know, it seems kinda redundant that it's the dev server port.

amrbashir: yeah, changed it to `--port`

lucasfernog (approved):

PR #6318, opened 2023-02-19
fix(api): event name types

2nthony:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html

May a better dev experience with this type declaration, we can get code completion instead of `import { TauriEvent } from 'xxx'/event`, and the `TauriEvent["xxx"]` still works.

Tested in source code.

image

### DIFF
--- a/tooling/api/src/event.ts
+++ b/tooling/api/src/event.ts
@@ -12,7 +12,7 @@
 import * as eventApi from './helpers/event'
 import type { EventCallback, UnlistenFn, Event } from './helpers/event'
 
-export type EventName = TauriEvent | string
+export type EventName = `${TauriEvent}`
 
 /**
  * @since 1.1.0
--- a/tooling/api/src/event.ts
+++ b/tooling/api/src/event.ts
@@ -12,7 +12,7 @@
 import * as eventApi from './helpers/event'
 import type { EventCallback, UnlistenFn, Event } from './helpers/event'
 
-export type EventName = `${TauriEvent}`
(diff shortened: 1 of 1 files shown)


tooling/api/src/event.ts
```diff
@@ -12,7 +12,7 @@
 import * as eventApi from './helpers/event'
 import type { EventCallback, UnlistenFn, Event } from './helpers/event'
 
-export type EventName = TauriEvent | string
+export type EventName = `${TauriEvent}`
```
amrbashir: We also need to allow `string` type to be passed, `TauriEvent` is just for auto-completion purposes and not enforce a variant of it. I think we should be using this little trick from type-fest's [`LiteralUnion`](https://github.com/sindresorhus/type-fest/blob/main/source/literal-union.d.ts) instead

```suggestion
export type EventName = `${TauriEvent}` | (string & Record<never, never>)
```

amrbashir (changes requested):

amrbashir (approved):

2nthony:
Will this merge or not?🤔 @amrbashir

amrbashir:
Sorry for the delay but the lead developer is on vacation atm and it is waiting for his approval.

lucasfernog (approved):

Issue #5466, opened 2022-10-23
How to use x86_64-unknown-linux-musl  build bin


cosoc:
### Describe the bug

I want the program to be released without any dynamic dependencies. How should I configure my development environment

```
   Compiling tauri v1.1.1
   Compiling brotli-decompressor v2.3.2
   Compiling httparse v1.8.0
error: failed to run custom build command for `libdbus-sys v0.2.2`

Caused by:
  process didn't exit successfully: `/app_build/src-tauri/target/release/build/libdbus-sys-f5f362d6caae23b6/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-env-changed=DBUS_1_NO_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG_ALLOW_CROSS_x86_64-unknown-linux-musl
  cargo:rerun-if-env-changed=PKG_CONFIG_ALLOW_CROSS_x86_64_unknown_linux_musl
  cargo:rerun-if-env-changed=TARGET_PKG_CONFIG_ALLOW_CROSS
  cargo:rerun-if-env-changed=PKG_CONFIG_ALLOW_CROSS
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64-unknown-linux-musl
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64_unknown_linux_musl
  cargo:rerun-if-env-changed=TARGET_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_x86_64-unknown-linux-musl
  cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_x86_64_unknown_linux_musl
  cargo:rerun-if-env-changed=TARGET_PKG_CONFIG_SYSROOT_DIR
  cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR

  --- stderr
  pkg_config failed: pkg-config has not been configured to support cross-compilation.

  Install a sysroot for the target platform and configure it via
  PKG_CONFIG_SYSROOT_DIR and PKG_CONFIG_PATH, or install a
  cross-compiling wrapper for pkg-config and set it via
  PKG_CONFIG environment variable.
  One possible solution is to check whether packages
  'libdbus-1-dev' and 'pkg-config' are installed:
  On Ubuntu:
[...]

Issue #5007, opened 2022-08-22
[bug] Tauri project won't start


oxneat:
### Describe the bug

Tauri project won't start on : cargo tauri dev

### Reproduction

1) npm create tauri-app
2) cd tauri-app
3) cargo tauri-dev

### Platform and versions

Environment
  › OS: Ubuntu 22.04 X64
  › Node.js: 16.17.0
  › npm: 8.15.0
  › pnpm: Not installed!
  › yarn: 1.22.17
  › rustup: 1.25.1
  › rustc: 1.63.0
  › cargo: 1.63.0
  › Rust toolchain: stable-x86_64-unknown-linux-gnu 

Packages
WARNING: no lock files found, defaulting to npm
  › @tauri-apps/cli [NPM]: 1.0.5
  › @tauri-apps/api [NPM]: Not installed!
  › tauri [RUST]: 1.0.5,
  › tauri-build [RUST]: 1.0.4,

  › tao [RUST]: 0.12.2,
  › wry [RUST]: 0.19.0,

App
  › build-type: bundle
  › CSP: unset
  › distDir: ../src
  › devPath: ../src
package.json not found

App directory structure
  ├─ src-tauri
  ├─ src
  └─ .vscode

### Stack trace

```shell
Info Watching /home/oxneat/Desktop/Applications/tauri-app/src-tauri for changes...
   Compiling libdbus-sys v0.2.2
   Compiling memoffset v0.6.5
   Compiling num-traits v0.2.15
   Compiling num-integer v0.1.45
error: failed to run custom build command for `libdbus-sys v0.2.2`

Caused by:
  process didn't exit successfully: `/home/oxneat/Desktop/Applications/tauri-app/src-tauri/target/debug/build/libdbus-sys-b098d627b06328b8/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-env-changed=DBUS_1_NO_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG
  cargo:rerun-if-env-changed=DBUS_1_STATIC
  cargo:rerun-if-env-changed=DBUS_1_DYNAMIC
[...]

Issue #4373, opened 2022-06-16
[bug] Package dbus-1 was not found in the pkg-config search path


jdu:
### Describe the bug

Just saw the news about 1.0 and wanted to try out tauri after having not looked at it for a wile. Following along to the guide on this page https://tauri.app/v1/guides/getting-started/setup/html-css-js using `cargo tauri` and have hit a bit of a roadblock.

Running `cargo tauri dev` I get the following output:

<details>
<summary>Log Output</summary>

```
cargo tauri dev
   Compiling proc-macro-error v1.0.4
   Compiling lock_api v0.4.7
   Compiling libdbus-sys v0.2.2
   Compiling memoffset v0.6.5
   Compiling num-traits v0.2.15
   Compiling generic-array v0.14.5
   Compiling num-integer v0.1.45
   Compiling num-iter v0.1.43
error: failed to run custom build command for `libdbus-sys v0.2.2`

Caused by:
  process didn't exit successfully: `/home/jdu/Projects/pine/library/src-tauri/target/debug/build/libdbus-sys-6f56d2434582bb2f/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-env-changed=DBUS_1_NO_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG
  cargo:rerun-if-env-changed=DBUS_1_STATIC
  cargo:rerun-if-env-changed=DBUS_1_DYNAMIC
  cargo:rerun-if-env-changed=PKG_CONFIG_ALL_STATIC
  cargo:rerun-if-env-changed=PKG_CONFIG_ALL_DYNAMIC
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG_PATH
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH
  cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64-unknown-linux-gnu
[...]

PR #6332, opened 2023-02-21
feat: Use the zbus-backed of notify-rust

zeenix:
While [zbus](crates.io/crates/zbus) used to have too new MSRV, that is no longer the case for several months now and hence there is no good reason to avoid zbus dependency. The MSRV is tracked in the zbus CI even to avoid any accidental bumps. The MSRV of zbus is 1.60 currently, which is only 1 version above the MSRV of dbus-rs. Neither Rust 1.60 nor 1.59 is available in Ubuntu LTS or Debian unstable so this change doesn't introduce any distro issues AFAICT.

OTOH, using zbus eliminates dependency on a platform lib and hence enhances the portability and ease of building. Issues like these will be a thing of the past after this change: 

https://github.com/tauri-apps/tauri/issues/4373 
https://github.com/tauri-apps/tauri/issues/5007
https://github.com/tauri-apps/tauri/issues/5466

---

Moreover, the latest release of notify-rust enables async APIs based on zbus' async support, in case that is useful for Tauri project in the future.

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


FabianLars:
> The MSRV of zbus is 1.60 currently

That unfortunately is too new. We are stuck with 1.59 on Tauri v1.X - i can't remember what we used to determine that version, maybe what some debian version used to build firefox(?) and even though they raised their version above 1.59 too we can't change our msrv policy in a minor release. or at least didn't want to the last time we talked about it.

That said, this change would be fine on the `next` branch where we will likely change the MSRV policy to something like "last X versions, updated on each minor release"

Edit: Depending on how long we want to keep releasing 1.X version we should probably discuss the msrv _again_, but it looks like we will focus more on 2.0 than minor releases 🤔

zeenix:
> That unfortunately is too new. We are stuck with 1.59 on Tauri v1.X - i can't remember what we used to determine that version,

Well to be fair, that was also quite new at some point so I'm sure there was a very good reason. :)

> maybe what some debian version used to build firefox(?) and even though they raised their version above 1.59 too we can't change our msrv policy in a minor release. or at least didn't want to the last time we talked about it.

IMO that seems a bit too strict. Not bumping MSRV in a patch release (e.g [what `time` crate recently did](https://github.com/time-rs/time/issues/551)) would be not only understandable, it would be a good policy. While strictly speaking MSRV bump is a (potentially) breaking change, as we know, [not all breaking changes are major](https://predr.ag/blog/some-rust-breaking-changes-do-not-require-major-version/).

But it's your project so your rules. :)

> That said, this change would be fine on the `next` branch where we will likely change the MSRV policy to something like "last X versions, updated on each minor release"
> 
> Edit: Depending on how long we want to keep releasing 1.X version we should probably discuss the msrv _again_, but it looks like we will focus more on 2.0 than minor releases thinking

I understand. It all depends on your project's priorities. I would only suggest humbly that you consider an MSRV bump. Perhaps bumping only by 1 version (1.60) for now then switch to the new MSRV policy after 2.0?

amrbashir:
I think it should be fine to bump the MSRV in a minor release and `1.60` is almost 1 year old, that's old enough.

zeenix:
@amrbashir cool. Should I bump the MSRV as part of this PR?

amrbashir:
Fine by me but I'd rather wait for @lucasfernog to approve this first.

lucasfernog:
MSRV 1.60 would also allow us to bump another dependency (serde-with IIRC) so it would be nice. But I'd like to hear from @chippers.

FabianLars:
i mean, i'm all for a msrv bump (my first message was not my _opinion_ but a summary of our past 100 discussions about this), _but_ we should at least clearly communicate this, preferably (long) before actually releasing a new tauri version with it and not just via a small mention in the release notes _on release day_. Just because we were super strict about this so far in 1.X 🤷

We could then also use this chance for a teaser about the MSRV policy changes in v2 to ease users into that.

zeenix:
> _but_ we should at least clearly communicate this, preferably (long) before actually releasing a new tauri version with it and not just via a small mention in the release notes _on release day_. Just because we were super strict about this so far in 1.X shrug

Fair enough. Anything I can help with? Should I bump the MSRV in the PR and then you folks handle the rest?

chippers:
One of the reasons why 1.59 is the magic number we settled on was that Debian packages `1.59`in their stable release for Firefox, and they have a policy that packages must be built by the other packages in that release. Our goal was to hopefully make it easier for Tauri applications to make it into Debian in the future, and this is just the current stable version. Unfortunately, Debian has entered their first soft freeze for the next release and that rustc version is only `1.63`which means we will miss out on some good stuff like workspace dependency inheritance.

Usually when a package maintainer commits to packaging an application, it would be up to them to freeze dependencies and create patches to satisfy the MSRV requirements, but we wanted it to be dead easy for a Tauri application to be able to be accepted.

It's not a permanent solution, just our general goal to allow easier packaging and more likelihood a Tauri app would be packaged.

If we can improve our Linux distribution stuff then this matters a lot less. Once we can have e.g. flatpaks working very well and perhaps an easier way for generating 3rd party repos (like how installing some packages will actually add their own apt repository and pull the packages from there) I think we would be in a much better spot to adopt a `stable - X`policy.

Since no new packages can make it into current or next stable, perhaps we just adopt `testing`'s version for now (and upgrade as it upgrades) until we decide what to do later.

chippers:
@lucasfernog ^

zeenix:
Ping? ☺️

chippers:
> Ping? ☺️

Lucas is taking a break right now, so a decision will likely take a bit of time

lucasfernog:
I feel like we should discuss the MSRV bump with the rest of the community.

zeenix:
> I feel like we should discuss the MSRV bump with the rest of the community.

Tbh, I was hoping you'd just be the last approver to bumping and we can finally get this landed, but sure. Please let me know how I can help with that.

zeenix:
> I feel like we should discuss the MSRV bump with the rest of the community.

How is that going?

zeenix:
thanks @lucasfernog

lucasfernog (approved):

PR #6339, opened 2023-02-22
feat(core): add path plugin, implement mobile APIs

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


lorenzolewis:
@lucasfernog is this one going to be similar to the camera plugin where you initialise it like this:

```rust
tauri::Builder::default()
  .plugin(tauri_plugin_camera::init())
```

Or for these "core" things are they going to just work as imports?

A similar question for things like `clipboard`, `dialog`, `notification`, `updater` and others if they get moved to this sort of a de-coupled system. 

These only have to be initialised on the Rust side, right? And then they can be used with a normal import on the JS side for instance? 

So if notifications move to this in the future, I'd do these steps to send a notification from the JS side:

1. Add the Node dep for `@tauri-apps/notification` (or whatever it will be) and the same on the Rust side
2. Initialise the plugin with something similar to the code block above
3. Then I import the package on the JS side and call the respective function to trigger the notification.

---

For things Tauri directly uses (such as path) those are going to stay in this repo, right? And for those that it doesn't they'll be in the monorepo? Need to figure out how I set up the docs to work.

lucasfernog:
Yeah you're right @lorenzolewis, but the core plugins will be initialized automatically and users will just need to import and use it like a regular module.

lucasfernog:
At least for now anyway, I know @chippers proposed separating the plugin types so we could actually move this plugin to the workspace repo, but I'm not sure if that's possible.

lorenzolewis:
I'm excited to dig in to what the DX looks like once these are in. Especially around package versions and asking devs to upgrade. Could be some enhancements we add to the CLI (maybe an add command as well) that helps with the DX and troubleshooting as we introduce more packages and combinations in.

PR #6356, opened 2023-02-24
feat: add config links to json schema (closes #5479)

### LINKED ISSUE #5479: Documentation link to a config option on Hover
Love the extension so far. Although I find one thing annoying. 
I m not a huge fan of scanning the entire project documentation before hacking it out myslef. 
I start coding from the get-go and gather knowledge about the underlying stack using VS code intellisense.

In the Tauri config file I get nice information about a certain config option but, what I'd love along with it is the ability to go to the documentation section from VS code using hyperlinks
Screenshot 2022-10-24 at 21 09 49

I love tsconfig for this reason
Screenshot 2022-10-24 at 21 08 43

I would be more than happy to help out with a PR here unless I'm missing something

ShaunSHamilton:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

Closes #5479

_I am unsure about the need for any package version bumps - this is really just a docs change_ 🤷‍♂️ 

I considered a few approaches for this, but, after fooling around for too long, I decided to just submit the simple solution.

**Future Considerations**

- The links could be automatically generated (updated) - making maintenance easier
- Certain configs did not get links to look as expected (they are overwritten):
  - The `WindowsConfig` is not updated, because it is defined as a `Vec`:

```rust
  /// The windows configuration.
  #[serde(default)]
  pub windows: Vec<WindowConfig>,
```

As such, the doc comment above the field is taken - contrast other definitions where the struct definition doc comment overwrites the field doc comment.

- It would be nice to have full markdown support in the tooltip using the `markdownDescription` property. However, this is a VSCode field, and is not defined in the JSON Schema. So, the implementation would involve more hacking as the `schemars::schema::Metadata` struct does not include a `markdown_description` field.

### DIFF
--- a/core/config-schema/schema.json
+++ b/core/config-schema/schema.json
@@ -208,7 +208,7 @@
   "additionalProperties": false,
   "definitions": {
     "PackageConfig": {
-      "description": "The package configuration.",
+      "description": "The package configuration.\n\nSee more: https://tauri.app/v1/api/config#packageconfig",
       "type": "object",
       "properties": {
         "productName": {
@@ -231,7 +231,7 @@
       "additionalProperties": false
     },
     "TauriConfig": {
-      "description": "The Tauri configuration object.",
+      "description": "The Tauri configuration object.\n\nSee more: https://tauri.app/v1/api/config#tauriconfig",
       "type": "object",
       "properties": {
         "pattern": {
@@ -508,7 +508,7 @@
       ]
     },
     "WindowConfig": {
-      "description": "The window configuration object.",
+      "description": "The window configuration object.\n\nSee more: https://tauri.app/v1/api/config#windowconfig",
       "type": "object",
       "properties": {
         "label": {
@@ -764,7 +764,7 @@
       ]
     },
     "CliConfig": {
-      "description": "describes a CLI configuration",
+      "description": "describes a CLI configuration\n\nSee more: https://tauri.app/v1/api/config#cliconfig",
       "type": "object",
       "properties": {
         "description": {
@@ -1010,7 +1010,7 @@
       "additionalProperties": false
     },
     "BundleConfig": {
-      "description": "Configuration for tauri-bundler.",
+      "description": "Configuration for tauri-bundler.\n\nSee more: https://tauri.app/v1/api/config#bundleconfig",
       "type": "object",
       "required": [
         "identifier"
@@ -1236,7 +1236,7 @@
       ]
     },
     "AppImageConfig": {
-      "description": "Configuration for AppImage bundles.",
+      "description": "Configuration for AppImage bundles.\n\nSee more: https://tauri.app/v1/api/config#appimageconfig",
       "type": "object",
       "properties": {
         "bundleMediaFramework": {
@@ -1248,7 +1248,7 @@
       "additionalProperties": false
     },
     "DebConfig": {
-      "description": "Configuration for Debian (.deb) bundles.",
+      "description": "Configuration for Debian (.deb) bundles.\n\nSee more: https://tauri.app/v1/api/config#debconfig",
       "type": "object",
       "properties": {
         "depends": {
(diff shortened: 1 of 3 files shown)


amrbashir (changes requested):
Thank you for your contribution. 

While this is a good start, we definitely should also take this opportunity to add links for each field as well, for example this https://tauri.app/v1/api/config/#cliarg.numberofvalues

> The `WindowsConfig` is not updated, because it is defined as a Vec
> 
> As such, the doc comment above the field is taken - contrast other definitions where the struct definition doc comment overwrites the field doc comment.

what do you mean about that? you can use this `https://tauri.app/v1/api/config/#tauriconfig.windows` for the `windows` field and this link `https://tauri.app/v1/api/config/#windowconfig` on the struct and both will show as expected.

ShaunSHamilton:
> what do you mean about that?

```rust
/// Doc comment ignored on `TauriConfig.windows`
struct WindowsConfig {}
/// Doc comment used for `TauriConfig.bundle`
struct BundleConfig {}
struct TauriConfig {
  /// Doc comment used, because type `WindowsConfig` is not base type - `Vec<...>` is
  windows: Vec<WindowsConfig>;
  /// Doc comment overridden, because priority is given to comment above base type - `BundleConfig`
  bundle: BundleConfig;
}
```

---

However, this is now moot, because I will add links to all fields.

amrbashir:
I see what you mean, it is inconsistent, but we have to make the best of it

FabianLars:
@lorenzolewis do you think this is something that should wait for the new docs?

ShaunSHamilton:
One problem with the current docs is it would result in a self reference:

image

FabianLars:
As long as they all use a similar format that should be easy to strip in the doc generator

lorenzolewis:
I'm not familiar with how exactly this is generated (nor all the details of the schema spec) but is there any support for an actual link object? I want to avoid putting information in a generic "description".  

Some things to look at are the spec for hyper-schemas, particularly the `links` object: https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-hyperschema-01#section-5.2

An example of this in practice I found here: https://apisyouwonthate.com/blog/getting-started-with-json-hyper-schema#json-hyper-schema

I'm curious at what TypeScript is using that generates the links in the screenshot below. Are those just hard-coded links in the comment part of the schema? Or are they using a specific field.

Then that leads me to ask if whatever tooling we're using to extract the info from the .rs files into the appropriate schema supports it. 

Screenshot 2023-02-28 at 18 02 32

ShaunSHamilton:
> I'm curious at what TypeScript is using that generates the links in the screenshot below. Are those just hard-coded links in the comment part of the schema? 

https://github.com/SchemaStore/schemastore/blob/b4eccc2fb5ad76fd9c0a70fa67228e5d65e2b562/src/schemas/json/tsconfig.json#L693-L705

> Then that leads me to ask if whatever tooling we're using to extract the info from the .rs files into the appropriate schema supports it.

As said in the PR description:

> It would be nice to have full markdown support in the tooltip using the `markdownDescription` property. However, this is a VSCode field, and is not defined in the JSON Schema. So, the implementation would involve more hacking as the `schemars::schema::Metadata` struct does not include a `markdown_description` field.

lorenzolewis:
I checked and schemars doesn't have support for that hyper protocol which adds the links support.

I think it would be better to instead link out to a single page on all the objects like https://tauri.app/v1/api/config. That would be a lot easier to maintain both in this repo I'd imagine. Plus we're also going to break all of these URLs this summer most likely with some changes coming to the website.

ShaunSHamilton:
Thank you for the input, maintainers.

Could I get clarification on which direction this feature should go?

**Discussed**

a) Discard current feature, pending new docs
b) Add specific property links to every field
c) Add the same link (`/config`) to every field

lorenzolewis:
I'd be a proponent of c) as it would still give you a place in IDEs to get more info without increasing maintenance and gives docs the flexibility to change URL paths. @tauri-apps/wg-tauri are the ones who ultimately maintain this repo so will defer to them for a decision. 

RE: Which path to use if going with option c: https://tauri.app as the path previously stated **will** be changing this spring/summer.

Edit: And thank you @ShaunSHamilton for outlining the discussed points, very much appreciated <3

lucasfernog (approved):

PR #6362, opened 2023-02-26
feat: add attributes for drag region

BillGoldenWater:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
It's useful for set a element that has many children draggable.

eg.

A floating window need to be draggable when drag on any where, and can toggle maximize on double click on title.

Use to need set data-tauri-drag-region everywhere, and need to manually implement it for region that doesn't need maximize.

This PR let developer could achieve this just by set body attributes to `data-tauri-drag-region-container` and title container to `data-tauri-drag-region-container` `data-tauri-drag-region-titlebar`.

### Description:

`data-tauri-drag-region`: mark this element as "drag region"

`data-tauri-drag-region-container`: children are also considered drag regions

`data-tauri-drag-region-titlebar`: the "drag region" maximizes window on double click (default when only `data-tauri-drag-region` is used)

`data-tauri-drag-region-exclude`: this element doesn't trigger `drag`

### Example:

```html
<div class="titlebar"
     data-tauri-drag-region-container="true"
     data-tauri-drag-region-titlebar="true"
>
  <div>Title</div>
  <div>Some decoration</div>
  <button data-tauri-drag-region-exclude>close</button>
</div>
```

### DIFF
--- a/.changes/add-attribute-for-drag-region.md
+++ b/.changes/add-attribute-for-drag-region.md
@@ -17,21 +17,12 @@ Description:
 Example:
 
 ```html
-
-<div>
-  <div class="title"
-       data-tauri-drag-region-container="true"
-       data-tauri-drag-region-titlebar="true"
-  >
-    <div>Title</div>
-    <div>Some decoration</div>
-    <button data-tauri-drag-region-exclude>close</button>
-  </div>
-  <div class="content"
-       data-tauri-drag-region-container="true"
-  >
-    <div>content</div>
-    <div>content</div>
-  </div>
+<div class="titlebar"
+     data-tauri-drag-region-container="true"
+     data-tauri-drag-region-titlebar="true"
+>
+  <div>Title</div>
+  <div>Some decoration</div>
+  <button data-tauri-drag-region-exclude>close</button>
 </div>
 ```
--- a/.changes/add-attribute-for-drag-region.md
+++ b/.changes/add-attribute-for-drag-region.md
@@ -5,3 +5,41 @@
 Added
 attribute `data-tauri-drag-region-container`, `data-tauri-drag-region-title`, `data-tauri-drag-region-interactive`, to
 enable drag on children, control maximize on double click, control is or not interactable on button/input/etc.
+
+Description:
+
+`data-tauri-drag-region`: mark this element as "drag region"
+
+`data-tauri-drag-region-container`: mark the "drag region" effects its children
+
+`data-tauri-drag-region-title`: mark the "drag region" maximize on double click (default add when
+only `data-tauri-drag-region` is set)
+
+`data-tauri-drag-region-interactive`: mark the "drag region" wouldn't prevent event to its children that interact-able,
+like button, input, etc. (check by "is property 'value' exists")
+
+Example:
+
+```html
+
+<div>
+  <div class="title"
+       data-tauri-drag-region-container="true"
+       data-tauri-drag-region-title="true"
+       data-tauri-drag-region-interactive="true"
+  >
+    <div>Title</div>
+    <div>Some decoration</div>
+    <button>close</button>
+    <div data-tauri-drag-region="false" onclick="alert('clicked')">
+      some element not interactive by default
+    </div>
+  </div>
+  <div class="content"
+       data-tauri-drag-region-container="true"
+  >
+    <div>content</div>
+    <div>content</div>
+  </div>
+</div>
+```
--- a/.changes/add-attribute-for-drag-region.md
+++ b/.changes/add-attribute-for-drag-region.md
@@ -1,0 +1,7 @@
+---
+"tauri": "minor"
+---
+
+Added
+attribute `data-tauri-drag-region-container`, `data-tauri-drag-region-title`, `data-tauri-drag-region-interactable`, to
+enable drag on children, control maximize on double click, control is or not interactable on button/input/etc.
--- a/.changes/add-attribute-for-drag-region.md
+++ b/.changes/add-attribute-for-drag-region.md
@@ -12,8 +12,7 @@ Description:
 
 `data-tauri-drag-region-titlebar`: the "drag region" maximizes window on double click (default when only `data-tauri-drag-region` is used)
 
-`data-tauri-drag-region-interactive`: mark the "drag region" wouldn't prevent event to its children that interact-able,
-like button, input, etc. (check by "is property 'value' exists")
+`data-tauri-drag-region-exclude`: this element doesn't trigger `drag`
 
 Example:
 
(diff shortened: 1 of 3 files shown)


core/tauri/scripts/core.js
```diff
@@ -132,28 +132,113 @@
...
+
+    /**
+     * @param {HTMLElement} element
+     * @return {DragInfo|null}
+     */
+    function elementDragInfo(element) {
+      const dragAttr = element.getAttribute("data-tauri-drag-region");
+      const containerAttr = element.getAttribute("data-tauri-drag-region-container");
+      const titleAttr = element.getAttribute("data-tauri-drag-region-title");
+      const interactableAttr = element.getAttribute("data-tauri-drag-region-interact-able");
```
amrbashir: no need for extra dash

```suggestion
      const interactableAttr = element.getAttribute("data-tauri-drag-region-interactable");
```

amrbashir (changes requested):
Could you give some examples on how these might interact with each other? specifically the new ones.

.changes/add-attribute-for-drag-region.md
```diff
@@ -0,0 +1,45 @@
+---
+"tauri": "minor"
+---
+
+Added
+attribute `data-tauri-drag-region-container`, `data-tauri-drag-region-title`, `data-tauri-drag-region-interactive`, to
+enable drag on children, control maximize on double click, control is or not interactable on button/input/etc.
```
amrbashir:
```suggestion
Added new drag attributes,`data-tauri-drag-region-container`, `data-tauri-drag-region-title`, `data-tauri-drag-region-interactive`, to enable drag on children, control maximize on double click, control interactive elements behavior.
```

.changes/add-attribute-for-drag-region.md
```diff
@@ -0,0 +1,45 @@
...
+
+Description:
+
+`data-tauri-drag-region`: mark this element as "drag region"
+
+`data-tauri-drag-region-container`: mark the "drag region" effects its children
+
+`data-tauri-drag-region-title`: mark the "drag region" maximize on double click (default add when
```
amrbashir: Should probably be called `data-tauri-drag-region-titlebar`

.changes/add-attribute-for-drag-region.md
```diff
@@ -0,0 +1,45 @@
...
+
+Added
+attribute `data-tauri-drag-region-container`, `data-tauri-drag-region-title`, `data-tauri-drag-region-interactive`, to
+enable drag on children, control maximize on double click, control is or not interactable on button/input/etc.
+
+Description:
+
+`data-tauri-drag-region`: mark this element as "drag region"
+
+`data-tauri-drag-region-container`: mark the "drag region" effects its children
```
amrbashir:
```suggestion
`data-tauri-drag-region-container`: children are also considered drag regions
```

.changes/add-attribute-for-drag-region.md
```diff
@@ -0,0 +1,45 @@
...
+
+`data-tauri-drag-region-title`: mark the "drag region" maximize on double click (default add when
+only `data-tauri-drag-region` is set)
```
amrbashir:
```suggestion
`data-tauri-drag-region-title`: the "drag region" maximizes window on double click (default when only `data-tauri-drag-region` is used)
```

.changes/add-attribute-for-drag-region.md
```diff
@@ -0,0 +1,45 @@
...
+
+`data-tauri-drag-region-interactive`: mark the "drag region" wouldn't prevent event to its children that interact-able,
+like button, input, etc. (check by "is property 'value' exists")
```
amrbashir: The behavior should be changed to be used on the element that needs to be excluded.
```suggestion
`data-tauri-drag-region-exclude`: this element doesn't trigger `drag`
```

amrbashir (changes requested):

.changes/add-attribute-for-drag-region.md
BillGoldenWater: In my design, exclude single element is done by set `data-tauri-drag-region` to false,
and if `data-tauri-drag-region-container` is also used, then also exclude it's children

the `data-tauri-drag-region-interactive` is for mark a container that shouldn't prevent events to default interact-able element.

amr-crabnebula: yeah I understand but this feels quite unpredictable for me, I'd rather use an explicit attribute to mark an element as not draggable instead of adding to the parent.

amrbashir: yeah I understand but this feels quite unpredictable for me, I'd rather use an explicit attribute to mark an element as not draggable instead of adding to the parent.

amr-crabnebula (changes requested):
~deleted, I keep using my work account~

.changes/add-attribute-for-drag-region.md
```diff
@@ -0,0 +1,37 @@
...
+    <div>Some decoration</div>
+    <button data-tauri-drag-region-exclude>close</button>
+  </div>
+  <div class="content"
+       data-tauri-drag-region-container="true"
+  >
+    <div>content</div>
+    <div>content</div>
+  </div>
+</div>
```
amrbashir:
```suggestion
<div class="titlebar" data-tauri-drag-region-container data-tauri-drag-region-titlebar>
	<span>Title</span>
	<span>Some decoration</span>
    <button  data-tauri-drag-region-exclude>Close</button>
</div>
```

amrbashir (changes requested):
one more question does `data-tauri-drag-region-container` require `data-tauri-drag-region` to be also set or do they work independently?  I think they should be independent from each other and if they are both present `data-tauri-drag-region` would take the precedence and disable all other attributes (i.e. use the old behavior)

lucasfernog:
I think we should target this to v2 so it also has some time to be tested by everyone while it's unstable.

thisislvca:
Hey guys! Any updates on this? Working on an app that needs a custom drag region, and this would be very handy. I find it strange nobody else has asked for it yet lol

nathancovey:
Is this still in the works?

lucasfernog:
probably something for v2.6

BillGoldenWater:
after rethinking it, do we still want to use this design and implementation? 
since it will traverse the DOM tree all the way up ***every time*** a "non-captured" click event is triggered, which would add an O(n) cost, where n is the number of nested levels, even if they are not using the drag feature.


perhaps we could leverage the already existing event bulbbling by giving the user a helper function that will add the needed event handler to the target element for container.
```html
<div id="dragContainer"></div>
<script>
setDragContainer(dragContainer)
</script>
```
(not just a global variable for event handler since looking at my previous implementation, it looks like both click and mousedown are required, and maybe touch?)
```html
<div onclick="tauriDragContainer" onmousedown="tauriDragContainer">
```


or we could just put the container feature behind a flag, so the traversal is opt-in, or opt-out.

PR #6390, opened 2023-03-02
fix(core): rewrite `asset` protocol streaming, closes #6375

### LINKED ISSUE #6375: [bug] Asset protocol crashes app with large video files
### Describe the bug

When seeking into longer/larger videos using Tauri's asset protocol, the video will hang and eventually crash the app entirely.

After quite a bit of testing it seems that video size is what makes Tauri's asset protocol crash the app. My original thinking was that it crashed due to the length of the video, but after generating a really long but small sized countdown video it worked fine. Then through more testing I found that videos larger than `~3.5 GiB` crashed the asset protocol when seeking roughly `80%` into the video, so when about `~3 GiB` needed to be loaded.

Also from my testing it seems that Tauri's asset protocol is seeking the entire video from `start->seek point`, which means that for longer/larger videos it takes a significant amount of time to seek into videos _(from an SSD)_. The reason I think it's seeking the entire video is due to how long the seeking takes on longer videos compared to when using a HTTP server to stream the file _(instant seeking, compared to multiple seconds sometimes with Tauri's asset protocol)_.

I doubt I'm running out of memory since my computer has 32GB of RAM to use. When looking at the task manger it crashes way before all my RAM is utilized. Tauri is using around 4GB of RAM before crashing.

---

I have tried using the [streaming example](https://github.com/tauri-apps/tauri/tree/dev/examples/streaming), it gives the same result as the normal asset protocol.

---

When the app crashes nothing shows up in the Web DevTools or in the terminal running `tauri dev`. When running the app using `RUST_BACKTRACE` the app no longer crashes, instead the video tries to load endlessly but doesn't cause a crash anymore.

---

## Found temporary solution

When I found this issue with the asset protocol I decided to embed [Rocket](https://rocket.rs/) and use it to stream the videos, sadly it had the same issue, forcing me to use a Rocket responder that handled video streaming correctly. After a while I found [rocket_seek_stream](https://github.com/rydz/rocket_seek_stream) made by [rydz](https://github.com/rydz). This fixed my issue and videos could now stream properly, seeking the video now doesn't load from `start->seek point`, instead the video is only loaded from the seek point onwards. Even though I was able to workaround the issue, I would like for Tauri's asset protocol to get fixed since I'd much rather not include Rocket in my project.

When using `rocket_seek_stream` with Rocket the app no longer crashes and seeking inside large videos is instant.

---

## Crash demo

https://user-images.githubusercontent.com/11979966/221927917-4f26e879-2950-49d2-b1f6-993d4882d15d.mp4

### Reproduction

_No response_

### Expected behavior

_No response_

### Platform and versions

**Environment**
  - OS: Windows 10.0.22621 X64
  - Webview2: 110.0.1587.57
  - MSVC:
  - Node.js: 18.14.2
  - npm: 9.3.1
  - pnpm: Not installed!
  - yarn: Not installed!
  - rustup: 1.24.3
  - rustc: 1.66.1
  - cargo: 1.66.1
  - Rust toolchain: stable-x86_64-pc-windows-msvc

**Packages**
  - @tauri-apps/cli [NPM]: 1.2.3
  - @tauri-apps/api [NPM]: Not installed!
  - tauri [RUST]: 1.2.4,
  - tauri-build [RUST]: 1.2.1,
  - tao [RUST]: 0.15.8,
  - wry [RUST]: 0.23.4,

**App**
  - build-type: bundle
  - CSP: default-src 'self'; style-src 'self' 'unsafe-inline'; font-src 'self' data: *; img-src 'self' asset: https://asset.localhost; media-src stream: https://stream.localhost asset: https://asset.localhost
  - distDir: ../src
  - devPath: ../src

**App directory structure**
  ├─ .git
  ├─ .vscode
  ├─ node_modules
  ├─ src
  └─ src-tauri

### Stack trace

_No response_

### Additional context

I created a [small repo](https://github.com/mauritzn/tauri-asset-bug) that allows you to pick a local video to play _(which crashes when seeking close to the end of larger files)_. I was not able to include a video file since the video file needs to be quite large _(multiple GiBs)_, as such the project allows you to pick the video file to play. The video file should be `10GiB+` to reliably crash the app, in my testing `~3.5GiB` is usually enough to crash it, but some larger video files still work _(sometimes)_, as such `10GiB+` has always been a guaranteed crash for me.

The video used should be in a format supported by browsers _(e.g. H264, VP9)_. H265 is not support since browsers do not have codec support for it.

In general the videos I've been trying to play that crash are Twitch VODs, since they are long and large _(e.g. 8+ hours, 10GiB+)_.

amrbashir:

- [X] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue 
- [ ] Code style update
- [X] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] No

### Checklist


ZerNico:
The rewrite fails to get the mime type on files where the file size is below 8192 bytes.

amrbashir:
@ZerNico  Does it panic and fail to load the file or default to `text/html`? Could you provide a minimal repro?

ZerNico:
@amrbashir It errors with `[Error] Origin http://localhost:1420 is not allowed by Access-Control-Allow-Origin. Status code: 404` in the browser. I am pretty sure it's from the code to get the mime_type because of `let mut magic_bytes = [0; 8192];`. It tries to read the file to 8192 bytes but fails as the file is smaller than that.

Here is a minimal reproduction. The 2 txt files to test are in the root of the project.
https://github.com/ZerNico/tauri-asset-bug/

amrbashir:
@ZerNico thanks for testing it out, it should be fixed now, please give it another test.

ZerNico:
@amrbashir awesome, works fine now.

lucasfernog (approved):

Issue #5781, opened 2022-12-08
[bug] failed to bundle project: error running appimage.sh


XuJinNet:
### Describe the bug

Build a  hello word project, run `pnpm tauri build`, get a  error:
```
   Compiling serialize-to-javascript v0.1.1
   Compiling serde_repr v0.1.9
   Compiling state v0.5.3
   Compiling webkit2gtk v0.18.2
    Finished release [optimized] target(s) in 2m 14s
    Bundling tauri-app-js_0.0.0_amd64.deb (/home/xujin/My_Projects/learning/learning-tauri/tauri-app-js/src-tauri/target/release/bundle/deb/tauri-app-js_0.0.0_amd64.deb)
    Bundling tauri-app-js_0.0.0_amd64.AppImage (/home/xujin/My_Projects/learning/learning-tauri/tauri-app-js/src-tauri/target/release/bundle/appimage/tauri-app-js_0.0.0_amd64.AppImage)
       Error failed to bundle project: error running appimage.sh: error running appimage.sh: `failed to run /home/xujin/My_Projects/learning/learning-tauri/tauri-app-js/src-tauri/target/release/bundle/appimage/build_appimage.sh`
 ELIFECYCLE  Command failed with exit code 1.
```

### Reproduction

run `pnpm tauri build`

### Expected behavior

Build success.

### Platform and versions

```
Environment
  › OS: Manjaro 22.0.0 X64
  › Node.js: 16.18.1
  › npm: 8.19.2
  › pnpm: 7.18.1
  › yarn: Not installed!
  › rustup: 1.25.1
  › rustc: 1.65.0
  › cargo: 1.65.0
  › Rust toolchain: stable-x86_64-unknown-linux-gnu 

Packages
  › @tauri-apps/cli [NPM]: 1.2.1
  › @tauri-apps/api [NPM]: 1.2.0
  › tauri [RUST]: 1.2.1,
  › tauri-build [RUST]: 1.2.1,
  › tao [RUST]: 0.15.6,
  › wry [RUST]: 0.22.6,

App
  › build-type: bundle
  › CSP: unset
  › distDir: ../dist
  › devPath: http://localhost:1420/

App directory structure
  ├─ src
  ├─ .vscode
  ├─ src-tauri
  ├─ node_modules
  └─ dist
```

FabianLars:
Please re-run it with the verbose flag; `pnpm tauri build --verbose` Hopefully we get more helpful output then.

[...]

PR #6391, opened 2023-03-03
fix: symlink issue bundling for linux #5781

passivedragon:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

Frankly, this fix is more from @ca0v who suggested it in the issue thread #5781, rather than me just putting it in.

I have test run this on a single project with success, but I haven't tested beyond that.

About the issue:
The bundling script fails if it has run once before, as the links it creates already exist, and it _does not like that_. The added `-f` overwrites the existing links with a new one, which should be the desired behavior.

### DIFF
--- a/.changes/fix-5781.md
+++ b/.changes/fix-5781.md
@@ -1,0 +1,5 @@
+---
+'tauri-bundler': 'patch'
+---
+
+Fixed error during bundling process for the appimage target on subsequent bundling attempts.
--- a/tooling/bundler/src/bundle/linux/templates/appimage
+++ b/tooling/bundler/src/bundle/linux/templates/appimage
@@ -55,9 +55,9 @@ chmod +x "{{tauri_tools_path}}/AppRun-${ARCH}"
 cp "{{tauri_tools_path}}/AppRun-${ARCH}" AppRun
 
 cp "{{icon_path}}" .DirIcon
-ln -s "{{icon_path}}" "{{app_name}}.png"
+ln -sf "{{icon_path}}" "{{app_name}}.png"
 
-ln -s "usr/share/applications/{{app_name}}.desktop" "{{app_name}}.desktop"
+ln -sf "usr/share/applications/{{app_name}}.desktop" "{{app_name}}.desktop"
 
 cd ..
 


FabianLars:
a bit weird that this seems to be required since the AppDir is supposed to be cleaned out before that 🤔

passivedragon:
Hmm, that definitely doesn't happen.

line 19 is the first time the directory is mentioned, and all it does is create it _if it doesn't exist already_
`mkdir -p "{{app_name}}.AppDir"`

it might need `rm -r "{{app_name}}.AppDir/*"` in the line directly after that and before the copy instruction, but from what I've seen from testing, that doesn't seem to be necessary.

FabianLars:
I was talking about the rust code that runs before the bash script is invoked: https://github.com/tauri-apps/tauri/blob/dev/tooling/bundler/src/bundle/linux/appimage.rs#L38

passivedragon:
That's pretty weird, if that failed, it should error and fail then and there to my understanding.
Wouldn't it need to error on bad input? From a quick, shallow read over rust docs seems to me like it might be happier with `output_path.clone()` instead of `&output_path`. I'm a rust noob though.

With even just this change for me though, it manages to build all files for me, all the way to the `.sig` file.

FabianLars:
clone vs `&` won't make a difference but you're right that it should fail if it can't remove the folder, at least the way i understood the docs.

To be clear, i'm fine with merging this PR but i'm irritated that it's needed in the first place (something we probably should look into as i'd imagine that the not deleted folder could cause issues in other places too)

FabianLars:
Oh btw, we'd need a change file similar to https://github.com/tauri-apps/tauri/blob/dev/.changes/bundler-msi-mini-version.md

passivedragon:
That work?

FabianLars:
yep, works for me thanks

FabianLars (approved):

lucasfernog (approved):
I think the fix doesn't make much sense and the problem might be somewhere else, but I'll take it if it works.

PR #6442, opened 2023-03-13
feat: add window effects api

amrbashir:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


core/tauri/src/window.rs
```diff
@@ -1301,6 +1314,11 @@ impl<R: Runtime> Window<R> {
       .map_err(Into::into)
   }
 
+  /// Sets window effects, pass `None` to clear any effects applied if possible.
+  pub fn set_window_effects(&self, effects: Option<WindowEffectsConfig>) -> crate::Result<()> {
```
lucasfernog: I think we should provide a Builder for the window effects, it would be easier to use (with a `From<WindowsEffectsConfig>` impl to simplify the window plugin and if someone wants to use the raw config object, which is unstable btw, if we add new fields it would break users apps unless they use `..Default::default()`)

amrbashir: added in [43a2e71](https://github.com/tauri-apps/tauri/pull/6442/commits/43a2e717ef53199629733ba031abfd242575936d)

lucasfernog:
I'm not sure why but I can't apply any of the effects on the API example (tried both macOS and Windows).

amrbashir:
Seems to be working just fine for me with WindowBuilder and tauri.conf.json 
image

Just make sure to enable transparency (decorations on Windows).

lucasfernog:
Can you share your config too @amrbashir just so I can compare

lucasfernog:
Ahh I forgot we removed transparency by default on the example. Let me check.

core/tauri/src/window.rs
```diff
@@ -1301,6 +1315,37 @@ impl<R: Runtime> Window<R> {
       .map_err(Into::into)
   }
 
+  /// Sets window effects, pass `None` to clear any effects applied if possible.
+  /// Note that the window must be transparent and without decorations.
```
```suggestion
  /// Note that the window must be transparent.
```

amrbashir: should work with decorations as well, we just have that one old bug of Windows https://github.com/tauri-apps/tao/issues/72

lucasfernog (dismissed):

lucasfernog:
Ah I forgot we also need to document this on the config. And maybe it's best to put a reference on the decorations/fullscreen issue until it gets fixed, otherwise people will think the effects aren't working.

lucasfernog (approved):

Issue #6389, opened 2023-03-02
[feat] make tauri icon support SVGs


ImUrX:
### Describe the problem

the ``tauri icon`` cli currently uses rasterized images and downscales them.

### Describe the solution you'd like

Using an SVG would be nice, it can be converted to any resolution!

fetzsav:
I've implemented SVG support for the icon tool in the CLI, here is the PR https://github.com/tauri-apps/tauri/pull/6444

Cioraz:
Hey, I would like to work on this issue.
Correct me if I'm wrong, so basically I've got to change the tauri icon in the cli?

PR #6444, opened 2023-03-14
[feat (issue #6389)] make tauri icon support SVGs

fetzsav:

- [ ] Bugfix
- [ x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x ] No

### Checklist
- [x ] When resolving issues, they are referenced in the PR's title (e.g `fix: remove a typo, closes #___, #___`)
- [x ] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md).
- [ x] I have added a convincing reason for adding this feature, if necessary

### Other information

This allows .SVG files to be passed into the CLI's icon module. This was a request in the issues tab


lorenzolewis:
This is an awesome addition! @amrbashir maybe this is something we could use downstream in CTA later as well to reduce the size even more? 

Edit: Would require a few more changes to the actual core bits of Tauri that pull in the rendered icons I realise, but still cool to move closer to that possibility!

amrbashir:
Cool! If this ends up getting merged, and tauri changes its logic to depend on a single svg icon and generate the others on the fly, it will save us a couple of hunderd kilobytes

fetzsav:
As per discussion in the Discord, there are a couple points worth checking before merge.

1) Resizing of the SVG. Currently, we are not utilizing the SVG, as we are converting it to rasterized and then performing the image resize. I’ll look into this when I get home today.

2) Extra C dependency from nsvg package.  Per Discord discussion, resvg was suggested. Just let me know if you guys would prefer to use a different crate.

In response to the previous comments… are you guys talking about saving extra disk space? So, if an SVG is used, the PNGs arent saved to disk? Just some clarification here would help thanks!

lorenzolewis:
@fetzsav we were just having a slight sidebar for CTA (create-tauri-app) for a few ideas of how this could help get us closer to shrinking the templates down there in the future. You'd still need to generate the raster images to use in icons, this PR just helps us get closer to using SVGs to generate those raster images and then in theory in the future we could use that functionality to optimise CTA.

tl;dr no impact to this, just some enhancements we could get closer to downstream with this PR

lucasfernog:
I think this feature should target the v2 release, v1.3 is pretty much ready and the idea is to focus 100% on v2 when that release is out.

lucasfernog:
I tried to use resvg but couldn't get the resize to work. Can you review @FabianLars ?

FabianLars:
I switched back to resvg because nsvg failed to resize my very first test icon 😂 (it worked with other test files, resvg could handle all i tested). Resizing should work, but i know what issue you were talking about.

Also the ios bg color is now applied after resizing to make sure it's the svg renderer resizing it and not image-rs. Seemed to work on my end but i have no ios setup to test this rn.

FabianLars (dismissed):

lucasfernog (approved):

tooling/cli/src/icon.rs
```diff
@@ -464,14 +464,14 @@ fn resize_and_save_png(
   source: &Source,
   size: u32,
   file_path: &Path,
-  ios_color: Option<Rgba<u8>>,
+  bg_color: Option<Rgba<u8>>,
```
FabianLars, line 467: i swear i named it this way first but reverted it for no reason 😂

PR #6448, opened 2023-03-14
fix(cli): use `npm run tauri -- foo` for correctly passing args to tauri

meowtec:

- [x] Bugfix
- [ ] Feature
- [x] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

If use `npm run tauri android init` for generating Android project, we will get some error like this when `run dev`:

```
error: unexpected argument 'aarch64' found
Usage: npm run tauri android android-studio-script [OPTIONS]
For more information, try '--help'.
```

Executed commandis:

```
npm run  tauri android android-studio-script --target aarch64
```

NPM will consume argument `--target`, then pass `android android-studio-script aarch64` to tauri.

When using npm, we should use `--` for passing args into tauri.

Instead, we should use `npm run  tauri -- android android-studio-script --target aarch64`.

The minimum case is `npm run tauri -v` which will output NPM version.

I have tested with `pnpm` and `yarn`, 

Also see https://github.com/npm/npm/pull/5518

### DIFF
--- a/.changes/npm-pass-args.md
+++ b/.changes/npm-pass-args.md
@@ -1,0 +1,5 @@
+---
+'cli.rs': 'patch'
+---
+
+Correctly pass arguments from `npm run` to `tauri`.
--- a/.changes/npm-pass-args.md
+++ b/.changes/npm-pass-args.md
@@ -1,5 +1,6 @@
 ---
 'cli.rs': 'patch'
+"cli.js": patch
 ---
 
 Correctly pass arguments from `npm run` to `tauri`.
--- a/tooling/cli/src/mobile/init.rs
+++ b/tooling/cli/src/mobile/init.rs
@@ -134,6 +134,7 @@ pub fn exec(
         // remove script path, we'll use `npm_lifecycle_event` instead
         build_args.remove(0);
       }
+      build_args.insert(0, "--".into());
       build_args.insert(0, var("npm_lifecycle_event").unwrap());
       build_args.insert(0, "run".into());
     }
--- a/tooling/cli/src/mobile/ios/xcode_script.rs
+++ b/tooling/cli/src/mobile/ios/xcode_script.rs
@@ -12,7 +12,7 @@ use crate::{
 use clap::Parser;
 use tauri_mobile::{apple::target::Target, opts::Profile, util};
 
-use std::{collections::HashMap, ffi::OsStr, path::PathBuf};
+use std::{collections::HashMap, env::var_os, ffi::OsStr, path::PathBuf};
 
 #[derive(Debug, Parser)]
 pub struct Options {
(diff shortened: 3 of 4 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -46,7 +46,7 @@ body:
     id: info
     attributes:
       label: Platform and versions
-      description: "Output of `npm run tauri info` or `cargo tauri info`"
```
amrbashir: this change is not needed since `npm run tauri info` works just fine

ARCHITECTURE.md
```diff
@@ -113,7 +113,7 @@ Once everything is installed, you can run:
 ```
 yarn tauri dev
 -or-
-npm run tauri dev
```
amrbashir: ditto

examples/api/README.md
```diff
@@ -32,7 +32,7 @@ $ npm install
 # with yarn
 $ yarn tauri dev
 # with npm
-$ npm run tauri dev
```
amrbashir: ditto

examples/resources/README.md
```diff
@@ -33,7 +33,7 @@ $ yarn package
 # with yarn
 $ yarn tauri dev
 # with npm
-$ npm run tauri dev
```
amrbashir: ditto

examples/sidecar/README.md
```diff
@@ -33,7 +33,7 @@ $ yarn package
 # with yarn
 $ yarn tauri dev
 # with npm
-$ npm run tauri dev
```
amrbashir: ditto

amrbashir (changes requested):

amrbashir (dismissed):

FabianLars:
Judging from this single example:
> npm run  tauri android android-studio-script --target aarch64

I don't think this is something that needs changes inside tauri right? Aren't we just talking about this:
```rs
npm run tauri android android-studio-script -- --target aarch64
```
(note the extra `--`)

I mean, this is an issue with all commands (including desktop ones) so wouldn't we just make it inconsistent with this change?

Sorry if i'm missing something obvious here.

meowtec:
@FabianLars Mobile is special: 

> Judging from this single example:
> 
> > npm run  tauri android android-studio-script --target aarch64
> 
> I don't think this is something that needs changes inside tauri right? Aren't we just talking about this:
> 
> ```rust
> npm run tauri android android-studio-script -- --target aarch64
> ```
> 
> (note the extra `--`)
> 
> I mean, this is an issue with all commands (including desktop ones) so wouldn't we just make it inconsistent with this change?
> 
> Sorry if i'm missing something obvious here.

For desktop, it is simple to add `--` before arguments. (And there should be some alert / advice in docs, in the future)

But mobile platform is different: the command `...android-studio-script --target ...` is automatically generated into Android project by `android init` command, and executed by gradle, not explicitly / manually. So it is required to change tauri's code.

lucasfernog (approved):

PR #6519, opened 2023-03-21
fix(tauri-utils): include `\n` in `io::read_line`, closes #6388

### LINKED ISSUE #6388: [bug] Missing newline character for ("on", "data") callback on js api
### Describe the bug

I'm experiencing an issue with the on data callback for the stdout pipeline when spawning a child process.
The child process spawned by Tauri is written in C: it prints a string which might terminate with a newline character or not, and a fflush is called right after this print.
The issue is that Tauri never includes the newline character when the on data listener is called, but I need to know if that newline character was present or not in the stdout.

### Reproduction

I have prepared a Tauri application to showcase this error: https://github.com/TALCo-Team/tauri-issue-newline-stdout

In order to reproduce the error, simply build and run the test application and observe the results reported inside the main page: the stdout outputs of the executed program will be missing the expected newline characters.
You can also check the C source code and try to run the launched executable using the shell instead: you'll see that the newline characters should indeed be present.

### Expected behavior

The Tauri interface should be able to report all newline characters from the stdout of its child processes, but it looks like all trailing newline characters are instead cutoff.

### Platform and versions

yarn tauri dev

The test application has been run on Ubuntu 20.04 and Windows 10 using the latest available version for Tauri, the issue only seems to happen on Ubuntu 20.04.

### Stack trace

_No response_

### Additional context

_No response_

amrbashir:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


lucasfernog:
I see the value of keeping this compatible with the Rust std library, but I worry this is a breaking change for some users. Maybe we could target this for v2 instead.

The use case of the issue is also a little weird (` but I need to know if that newline character was present or not in the stdout.`). If there's no newline or `\r` in the output, Tauri will keep waiting for it to appear and never give the bytes to the user, so he will end up not knowing this anyway.

Currently if the line does not end with `\r`, then you're sure it ends with `\n` and can append it on your end.

amrbashir:
> I see the value of keeping this compatible with the Rust std library, but I worry this is a breaking change for some users. Maybe we could target this for v2 instead.

tbh I see it as a bug fix so not a breaking change

> The use case of the issue is also a little weird (` but I need to know if that newline character was present or not in the stdout.`). If there's no newline or `\r` in the output, Tauri will keep waiting for it to appear and never give the bytes to the user, so he will end up not knowing this anyway.
> 
> Currently if the line does not end with `\r`, then you're sure it ends with `\n` and can append it on your end.

While that maybe true, it is a bad DX (showcased in the example you added) and beside the use case of the issue, unix and windows had different behaviors which needs to align and if you align any of them to the other it will break for some users so I'd rather just fix the buggy behavior and align with rust.

lucasfernog (approved):

PR #6580, opened 2023-03-29
fix(cli): remove config.toml generation for mobile projects, closes #6578

### LINKED ISSUE #6578: [bug] tauri mobile build leaves a `.cargo` folder
### Describe the bug

After I run `tauri android dev` or `tauri android build`, or `tauri ios dev` or `tauri ios build`, it leaves a `.cargo` folder with a config in the `src-tauri` folder. The config confuses cargo. For example, on Windows, the config set the default target to `x86_64-pc-windows-msvc`, and then simply running `cargo build` will compile binaries to `target/x86_64-pc-windows-msvc/debug` but not `target/debug`. This config is also not cross-platform because different OS uses different host target.

### Reproduction

```
cargo tauri android init
cargo tauri android dev
```

### Expected behavior

I don't understand why it generates such config. It shouldn't do that. We don't need to set host target manually at all.

### Platform and versions

```text
Environment
  › OS: Windows 10.0.22621 X64
  › Webview2: 111.0.1661.54
  › MSVC:
      - Visual Studio ???? 2022
  › Node.js: 18.15.0
  › npm: 9.5.0
  › pnpm: Not installed!
  › yarn: Not installed!
  › rustup: 1.25.2
  › rustc: 1.70.0-nightly
  › cargo: 1.70.0-nightly
  › Rust toolchain: nightly-x86_64-pc-windows-msvc

Packages
  › @tauri-apps/cli [NPM]: 2.0.0-alpha.3
  › @tauri-apps/api [NPM]: 2.0.0-alpha.1
  › tauri [RUST]: 2.0.0-alpha.4 (no manifest),
  › tauri-build [RUST]: 2.0.0-alpha.2 (no manifest),
  › tao [RUST]: 0.18.1 (no manifest),
  › wry [RUST]: 0.27.0 (no manifest),

App
  › build-type: bundle
  › CSP: unset
  › distDir: ../dist
  › framework: Vue.js
  › bundler: Vite

App directory structure
  ├─ dist
  ├─ node_modules
  ├─ public
  ├─ src
  └─ src-tauri
```

### Stack trace

_No response_

### Additional context

_No response_

amrbashir:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
I think we can remove this from the CLI and move it to a FAQ on the website so users can opt into it willingly.

### DIFF
--- a/tooling/cli/src/mobile/init.rs
+++ b/tooling/cli/src/mobile/init.rs
@@ -11,7 +11,6 @@ use tauri_mobile::{
     config::Config as AndroidConfig, env::Env as AndroidEnv, target::Target as AndroidTarget,
   },
   config::app::App,
-  dot_cargo,
   target::TargetTrait as _,
   util::{
     self,
--- a/tooling/cli/src/mobile/init.rs
+++ b/tooling/cli/src/mobile/init.rs
@@ -37,7 +37,7 @@ pub fn command(target: Target, ci: bool, reinstall_deps: bool) -> Result<()> {
 }
 
 pub fn configure_cargo(
-  app: &App,
+  _app: &App,
   android: Option<(&mut AndroidEnv, &AndroidConfig)>,
 ) -> Result<()> {
   if let Some((env, config)) = android {
@@ -57,20 +57,7 @@ pub fn configure_cargo(
     }
   }
 
-  let mut dot_cargo = dot_cargo::DotCargo::load(app)?;
-  // Mysteriously, builds that don't specify `--target` seem to fight over
-  // the build cache with builds that use `--target`! This means that
-  // alternating between i.e. `cargo run` and `cargo apple run` would
-  // result in clean builds being made each time you switched... which is
-  // pretty nightmarish. Specifying `build.target` in `.cargo/config`
-  // fortunately has the same effect as specifying `--target`, so now we can
-  // `cargo run` with peace of mind!
-  //
-  // This behavior could be explained here:
-  // https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags
-  dot_cargo.set_default_target(util::host_target_triple()?);
-
-  dot_cargo.write(app).map_err(Into::into)
+  Ok(())
 }
 
 pub fn exec(
--- a/tooling/cli/templates/mobile/android/.gitignore
+++ b/tooling/cli/templates/mobile/android/.gitignore
@@ -1,6 +1,5 @@
 *.iml
 .gradle
-/local.properties
 /.idea/caches
 /.idea/libraries
 /.idea/modules.xml


amrbashir:
This PR needs a new release of tauri-mobile where it has this commit https://github.com/tauri-apps/tauri-mobile/commit/f9d809519ca1efbbde996b3e906fd85983c8609c

tooling/cli/src/mobile/init.rs
```diff
@@ -57,20 +56,7 @@ pub fn configure_cargo(
     }
   }
 
-  let mut dot_cargo = dot_cargo::DotCargo::load(app)?;
-  // Mysteriously, builds that don't specify `--target` seem to fight over
```
lucasfernog, line 61: I think we can't remove the .cargo/config stuff because of the issue described here, Rust cache does not work without it.

amrbashir: Yeah but users would need to ignore `.cargo/config.toml` otherwise, it will be modified with different OSes and will make it hard for users with git. That's why I want to remove it from the CLI and move this piece to the the FAQ on the website.

lucasfernog: What is the change across different OSes? we should try to avoid overwriting it if it already has what we need.

amrbashir: what this workaround does, is that is specifies a default build target, for example:
```toml
[build]
target = 'x86_64-pc-windows-msvc'
```
so if someone else working on the project, uses linux it will change to:
```toml
[build]
target = 'x86_64-unknown-linux-gnu'
```

lucasfernog: Ah right. Still I think it's better us managing it than users not knowing about this, or having to manage it anyway.

amrbashir: but how? the only way is to write it or pass `--target` explicitly for host which both are not desirable IMO. I think this is a fair trade-off, like you only experience cache invalidation only when switching from desktop development to mobile development and vice versa which doesn't happen very often and won't affect CI anyways.

lucasfernog: > which doesn't happen very often

well, there's discussions about running desktop & mobile dev at the same time 😂

amrbashir: oh god, please no

lucasfernog: `build.target` can also be a path to a "relative path to a .json target spec file". Maybe we could use that instead, and change the default template to gitignore that JSON file.

amrbashir: IDK about that, seems like more workarounds and I'd rather keep the current one than go that way tbh.

PR #6610, opened 2023-04-01
feat(cli): improve logic to copy Rust library to iOS project Externals folder

DaAitch:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
Currently with `@tauri-apps/cli@next` we generate the static libs into project root `<project>/gen/..` instead of `<project>/src-tauri/gen/apple/Externals/debug/lib<name>.a`, so `npm run tauri ios dev` fails.

This PR fixes it.


FabianLars:
I don't know much about Tauri's mobile implemention but to be honest i can't imagine that this is the correct solution. The `src-tauri` directory is not required to have this name, or be on that folder level, it's only the standard we use for the templates.

Instead i think we should try to detect the correct "root" dir (the tauri dir) like we do everywhere else.

Either way, thanks for looking into it and for taking the time to contribute :)

DaAitch:
@FabianLars if you can send me a ref where we search the "src-tauri", I can update the PR for this.

FabianLars:
There should be a `tauri_dir()` function in helpers/app_paths.rs i think

DaAitch:
@FabianLars I found it, too :)

PR open again.
`cargo fmt` doesn't work on my machine :/

lucasfernog:
I didn't see your PR earlier and fixed the issue with the same approach on #6699. Checking your PR still made me realize there's a better way to implement it, so let's do it :)

lucasfernog (approved):

PR #6719, opened 2023-04-16
refactor(core): read tray icon only on desktop, refactor Context

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/core/tauri-codegen/src/context.rs
+++ b/core/tauri-codegen/src/context.rs
@@ -328,10 +328,10 @@ pub fn context_codegen(data: ContextData) -> Result<TokenStream, EmbeddedAssetsE
       let ext = system_tray_icon_path.extension();
       if ext.map_or(false, |e| e == "ico") {
         ico_icon(&root, &out_dir, system_tray_icon_path)
-          .map(|i| quote!(context.with_system_tray_icon(#i);))?
+          .map(|i| quote!(context.set_system_tray_icon(#i);))?
       } else if ext.map_or(false, |e| e == "png") {
         png_icon(&root, &out_dir, system_tray_icon_path)
-          .map(|i| quote!(context.with_system_tray_icon(#i);))?
+          .map(|i| quote!(context.set_system_tray_icon(#i);))?
       } else {
         quote!(compile_error!(
           "The tray icon extension must be either `.ico` or `.png`."
@@ -449,7 +449,7 @@ pub fn context_codegen(data: ContextData) -> Result<TokenStream, EmbeddedAssetsE
       _ => panic!("unknown shell open format, unable to prepare"),
     };
 
-    quote!(context.with_shell_scope(#root::ShellScopeConfig {
+    quote!(context.set_shell_scope(#root::ShellScopeConfig {
       open: #shell_scope_open,
       scopes: #shell_scopes
     });)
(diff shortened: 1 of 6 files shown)


core/tauri-codegen/src/context.rs
```diff
@@ -125,6 +125,16 @@ enum Target {
...
 }
 
+impl Target {
+  fn mobile(&self) -> bool {
+    matches!(self, Target::Android | Target::Ios)
+  }
+
+  fn desktop(&self) -> bool {
+    !self.mobile()
+  }
```
chippers: Not 100% sure we've been following this convention throughout the codebase, but since these aren't returning a "field" on the item but a property of it we may want to use the `is_` prefix.

```suggestion
  fn is_mobile(&self) -> bool {
    matches!(self, Target::Android | Target::Ios)
  }

  fn is_desktop(&self) -> bool {
    !self.mobile()
  }
```

core/tauri/src/scope/shell.rs
```diff
@@ -58,7 +58,7 @@ impl From<Vec<String>> for ExecuteArgs {
 }
 
 /// Shell scope configuration.
-#[derive(Debug, Clone)]
+#[derive(Debug, Default, Clone)]
```
chippers: This derive will have the `open` field set to `None` by default, which means that it is _**not**_ matched through the default regex. It should probably match the default of `^((mailto:\w+)|(tel:\w+)|(https?://\w+)).+`. We may want to adjust how we create it from the config so that we don't need to have the "default" filtering regex in more than 1 place, to reduce future duplication mistakes.

core/tauri/src/lib.rs
```diff
@@ -657,25 +663,38 @@ impl<A: Assets> Context<A> {
...
+  #[cfg(desktop)]
+  #[inline(always)]
+  pub fn with_system_tray_icon(&mut self, icon: Icon) {
+    self.system_tray_icon.replace(icon);
+  }
+
+  /// Sets the app shell scope.
+  #[cfg(shell_scope)]
+  #[inline(always)]
+  pub fn with_shell_scope(&mut self, scope: scope::ShellScopeConfig) {
```
chippers: `with_` prefixed methods are typically used for alternative constructors, not replacements of a single field. If there is a `self` being passed to a `with_` function, then it's probably named wrong. I think `set_` in this case works best since it's completely replacing a single item.

```suggestion
  pub fn set_system_tray_icon(&mut self, icon: Icon) {
    self.system_tray_icon.replace(icon);
  }

  /// Sets the app shell scope.
  #[cfg(shell_scope)]
  #[inline(always)]
  pub fn set_shell_scope(&mut self, scope: scope::ShellScopeConfig) {
```

chippers (changes requested):
There's 2 naming convention requests and 1 security-borderline issue. The `is_` naming convention is a take it or leave it, but I think the `with_` one is confusing.

Definitely fix the `Default` scope open regex however.

On a more of a "comment" note, we should probably decide on an architectural way to prevent so many `cfg`s mixed in with application logic. My first item to reach to would be to refactor concepts into Structs/Enums and use `cfg` inside those items to make them noops on cfg'd out platforms. Then the flow of stuff isn't broken up by `cfg` items all the time. This can apply to both the codebase and the codegen codebase.

chippers (approved):

PR #6727, opened 2023-04-17
feat(core): expose `runtime_handle` on AppHandle

lucasfernog:
This function will be needed by the clipboard and global shortcut plugins.

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/.changes/app-runtime-handle.md
+++ b/.changes/app-runtime-handle.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch
+---
+
+Added the `runtime_handle` method on `AppHandle`.
--- a/.changes/expose-runtime.md
+++ b/.changes/expose-runtime.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch
+---
+
+Re-export the `tauri_runtime` crate as `tauri::runtime`.
--- a/core/tauri/src/app.rs
+++ b/core/tauri/src/app.rs
@@ -277,6 +277,11 @@ impl<R: Runtime> AppHandle<R> {
   pub(crate) fn create_proxy(&self) -> R::EventLoopProxy {
     self.runtime_handle.create_proxy()
   }
+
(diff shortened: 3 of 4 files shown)


FabianLars:
Is it planned to stabilize the runtime crate? Otherwise this will likely cause semver issues with rust/cargo

lucasfernog:
I think it's time to stabilize it (I didn't want to put a v1 tag on it because of tao/wry, but so far it's been alright).
Otherwise we could just mark this as unstable (and hide from docs) since it's mostly for internal use anyway.

FabianLars:
> Otherwise we could just mark this as unstable (and hide from docs) since it's mostly for internal use anyway.

Yeah, generally i'd be completely fine with that, i don't care _that much_ about semver strictness, but this would require the plugin and the user to always use the same tauri minor version so that it can even compile, right? I'd imagine that some unlucky users would hit this issue once in a while 😅 

I mean, we can also just try it out and see what happens and stabilize the runtime if needed later 🤷

lucasfernog:
I think Cargo prevents that kind of thing doesn't it? It only compiles a dependency twice if they are semver incompatible. And if we need a new runtime API, we just bump the minor in the plugin, and Cargo will be forced to use that new minor.

FabianLars:
If the crate is stable, yes. If it's not (v0) then cargo is sometimes a bit sensible. But even then it should still only cause issues if the user locks dep versions in incompatible ways (or maybe if they use git deps which cargo doesn't auto-update?) - so it should be mostly fine hopefully.

Maybe i'm mixing up different semver related issues here though idk

Edit: yeah, pretty sure I'm mixing things up a bit. nevermind me I guess.

Issue #6743, opened 2023-04-19
[bug] Failed to bundle an NSIS installer under linux in tauri 2.0.0-alpha.8


karandit:
### Describe the bug

While trying to cross-compile a hello-world tauri-app for Windows from Linux (docker), I got at the very end an error at the bundling phase complaining about some missing folder. 

### Reproduction

0. `apt-get install --no-install-recommends -y nsis libwebkit2gtk-4.1-dev build-essential curl wget libssl-dev libgtk-3-dev libayatana-appindicator3-dev librsvg2-dev`
1. `cargo create-tauri-app`
2. choose Rust, Vanilla
3. change the bundle identifier in `tauri.conf.json > tauri > bundle > identifier` to something else than its current value
4. bump  version in `Cargo.toml > dependencies > tauri` to `2.0.0-alpha.8` to get rid off the `error: failed to select a version for tauri.` problem
5. `cargo tauri build --target x86_64-pc-windows-msvc`

### Expected behavior

the command executes successfully and an NSIS bundler is created

### Platform and versions

```text
WARNING: no lock files found, defaulting to npm

[✔] Environment
    - OS: Debian Unknown X64
    ✔ webkit2gtk-4.1: 2.40.0
    ✔ rsvg2: 2.54.5
    ✔ rustc: 1.68.2 (9eb3afe9e 2023-03-27)
    ✔ Cargo: 1.68.2 (6feb7c9cf 2023-03-26)
    ✔ rustup: 1.25.2 (17db695f1 2023-02-01)
    ✔ Rust toolchain: 1.68.2-x86_64-unknown-linux-gnu (environment override by RUSTUP_TOOLCHAIN)
    - node: 18.15.0
    - npm: 9.5.0

[-] Packages
    - tauri [RUST]: 2.0.0-alpha.8
    - tauri-build [RUST]: no manifest (1.2.1, 2.0.0-alpha.4)
    - wry [RUST]: 0.27.3
    - tao [RUST]: 0.18.3
    - @tauri-apps/api [NPM]: not installed!
    - @tauri-apps/cli [NPM]: 2.0.0-alpha.8

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../src
    - devPath: ../src
```

### Stack trace

```text
Compiling reproduce-bug v0.0.0 (/usr/src/examples/reproduce-bug/src-tauri)
    Finished release [optimized] target(s) in 18.47s
[...]

PR #6744, opened 2023-04-19
fix: copy failure in NSIS bundler #6743

karandit:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/tooling/bundler/src/bundle/windows/nsis.rs
+++ b/tooling/bundler/src/bundle/windows/nsis.rs
@@ -100,6 +100,9 @@ fn get_and_extract_nsis(nsis_toolset_path: &Path, _tauri_tools_path: &Path) -> c
   let data = download(NSIS_APPLICATIONID_URL)?;
   info!("extracting NSIS ApplicationID plugin");
   extract_zip(&data, &nsis_plugins)?;
+
+  create_dir_all(nsis_plugins.join("x86-unicode"))?;
+
   copy(
     nsis_plugins
       .join("ReleaseUnicode")


tooling/bundler/src/bundle/windows/nsis.rs
```diff
@@ -100,6 +100,13 @@ fn get_and_extract_nsis(nsis_toolset_path: &Path, _tauri_tools_path: &Path) -> c
   let data = download(NSIS_APPLICATIONID_URL)?;
   info!("extracting NSIS ApplicationID plugin");
   extract_zip(&data, &nsis_plugins)?;
+  info!("extracted NSIS ApplicationID plugin");
+
+  let nsis_plugins_x86 = nsis_plugins.join("x86-unicode");
+  if !nsis_plugins_x86.exists() {
+    create_dir_all(nsis_plugins_x86)?;
+  }
```
FabianLars:
```suggestion
  create_dir_all(nsis_plugins.join("x86-unicode"))?;
```
No need to check if it exists first imo. create_dir_all won't delete anything if a dir already exists.

FabianLars (changes requested):
Thank you! I could swear i tested this when i first added Linux support but glad you cought it before it went into v1.3.

I added just one small code change request below. Also i'm not reallyyy a fan of the new log just because we usually don't log the completion 😅 idk, just something to think about if we might wanna remove that again for now.

tooling/bundler/src/bundle/windows/nsis.rs
karandit: thank you, I'm not a Rust developer .... yet :)

karandit:
It's my  honor to contribute.
Yeah, it's tricky. I've also observed, during my experiments, that it worked if it was cached once

karandit:
let me remove that extra log line and squash it again

FabianLars:
Don't worry about squashing, we'll do that when that PR gets merged anyway :) (feel free to do it though)

karandit:
btw, I wanted to remove the duplication of this: `nsis_plugins.join("x86-unicode")` which appears in 3 places now, but I bumped into the borrow checker and I thought that maybe I leave it to the pros

FabianLars (approved):

FabianLars:
i tend to say that i prefer the duplicates because i can identify the paths quicker this way and the bundler doesn't need to be overly optimized either. So i'll leave that to the others as well :P

PR #6767, opened 2023-04-22
feat(core): expose Wry's `with_incognito` to Tauri on the `WindowBuilder::incognito` function.

Hyphrio:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
While `WebViewBuilder::data_directory` could simulate incognito mode by setting this to a temporary directory, that is annoying to use and error-prone as users could forget deleting the directory, thus creating large unnecessary directories, filling up the user's storage.

This PR merely exposes `with_incognito` from https://github.com/tauri-apps/wry/pull/916 to Tauri though the new `enable_incognito` function at the `WindowBuilder`, which would allow for a real incognito mode that would be useful for things that need incognito mode, like webview sign-ins.

NOTE: Requires Wry 0.27.3 or higher, thus this should be merged for 2.0.

### DIFF
--- a/core/tauri-runtime/src/webview.rs
+++ b/core/tauri-runtime/src/webview.rs
@@ -30,11 +30,13 @@ pub struct WebviewAttributes {
   pub accept_first_mouse: bool,
   pub additional_browser_args: Option<String>,
   pub window_effects: Option<WindowEffectsConfig>,
+  pub incognito: bool,
 }
 
 impl From<&WindowConfig> for WebviewAttributes {
   fn from(config: &WindowConfig) -> Self {
     let mut builder = Self::new(config.url.clone());
+    builder = builder.incognito(config.incognito);
     builder = builder.accept_first_mouse(config.accept_first_mouse);
     if !config.file_drop_enabled {
       builder = builder.disable_file_drop_handler();
@@ -65,6 +67,7 @@ impl WebviewAttributes {
       accept_first_mouse: false,
       additional_browser_args: None,
       window_effects: None,
+      incognito: false,
     }
   }
 
@@ -124,6 +127,13 @@ impl WebviewAttributes {
   #[must_use]
   pub fn window_effects(mut self, effects: WindowEffectsConfig) -> Self {
     self.window_effects = Some(effects);
+    self
+  }
+
+  /// Enable or disable incognito mode for the WebView.
+  #[must_use]
+  pub fn incognito(mut self, incognito: bool) -> Self {
+    self.incognito = incognito;
     self
   }
 }
--- a/core/tauri-utils/src/config.rs
+++ b/core/tauri-utils/src/config.rs
@@ -938,6 +938,13 @@ pub struct WindowConfig {
   /// - **Linux**: Unsupported
   #[serde(default, alias = "window-effects")]
   pub window_effects: Option<WindowEffectsConfig>,
+  /// Whether or not the webview should be launched in incognito  mode.
+  ///
+  ///  ## Platform-specific:
+  ///
+  ///  - **Android**: Unsupported.
+  #[serde(default)]
+  pub incognito: bool,
 }
 
 impl Default for WindowConfig {
@@ -978,6 +985,7 @@ impl Default for WindowConfig {
       additional_browser_args: None,
       shadow: true,
       window_effects: None,
+      incognito: false,
     }
   }
 }
@@ -2161,6 +2169,7 @@ mod build {
       let additional_browser_args = opt_str_lit(self.additional_browser_args.as_ref());
       let shadow = self.shadow;
       let window_effects = opt_lit(self.window_effects.as_ref());
+      let incognito = self.incognito;
 
       literal_struct!(
         tokens,
@@ -2199,7 +2208,8 @@ mod build {
         tabbing_identifier,
         additional_browser_args,
         shadow,
-        window_effects
+        window_effects,
+        incognito
       );
     }
   }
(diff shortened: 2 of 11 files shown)


core/tauri-runtime/src/webview.rs
```diff
@@ -97,6 +99,13 @@ impl WebviewAttributes {
     self.additional_browser_args = Some(additional_args.to_string());
     self
   }
+
+  /// Enables Incognito/InPrivate mode for the WebView.
+  #[must_use]
+  pub fn enable_incognito(mut self) -> Self {
```
amrbashir:
```suggestion
  pub fn incognito(mut self) -> Self {
```

core/tauri/src/window.rs
```diff
@@ -693,6 +693,13 @@ impl<'a, R: Runtime> WindowBuilder<'a, R> {
     self.webview_attributes.clipboard = true;
     self
   }
+
+  /// Enables Incognito/InPrivate mode for the WebView.
+  #[must_use]
+  pub fn enable_incognito(mut self) -> Self {
```
```suggestion
  pub fn incognito(mut self) -> Self {
```

amrbashir (changes requested):
Thanks for your contribution.

This option needs to also be added to the tauri config and js API

core/tauri-runtime/src/webview.rs
Hyphrio: Not exactly sure where it should be exposed in config but if I did it right it should be exposed in JS

Hyphrio:
I'm not sure if the JS api works so that'll need some testing.

core/tauri-runtime/src/webview.rs
amrbashir: For it to be exposed in config, you need to add it here https://github.com/tauri-apps/tauri/blob/942249060ed12a5d21a2b21c30e0638c1d2b9df0/core/tauri-utils/src/config.rs#L695 and for JS you need to add it to https://github.com/tauri-apps/tauri/blob/942249060ed12a5d21a2b21c30e0638c1d2b9df0/tooling/api/src/window.ts#L2097 
and for both to work you need to modify https://github.com/tauri-apps/tauri/blob/942249060ed12a5d21a2b21c30e0638c1d2b9df0/core/tauri-runtime-wry/src/lib.rs#L699

Hyphrio:
I think the JS api is broken on all versions of Tauri it seems.

core/tauri-runtime/src/webview.rs
Hyphrio: Yeah I figured that out (if im right)

Hyphrio:
I'll put this back to Draft until I finish rebasing.

Hyphrio:
Alright, already rebased.

core/tauri-runtime/src/webview.rs
```diff
@@ -126,6 +131,13 @@ impl WebviewAttributes {
     self.window_effects = Some(effects);
     self
   }
+
+  /// Enables Incognito mode for the WebView.
+  #[must_use]
+  pub fn enable_incognito(mut self) -> Self {
```
amrbashir:
```suggestion
  /// Enable or disable incognito mode for the WebView.
  #[must_use]
  pub fn incognito(mut self, incognito: bool) -> Self {
```

core/tauri-utils/src/config.rs
```diff
@@ -938,6 +938,13 @@ pub struct WindowConfig {
   /// - **Linux**: Unsupported
   #[serde(default, alias = "window-effects")]
   pub window_effects: Option<WindowEffectsConfig>,
+  /// Incognito mode.
```
```suggestion
  /// Whether or not the webview should be launched in incognito mode.
```

core/tauri/src/window.rs
```diff
@@ -741,6 +741,17 @@ impl<'a, R: Runtime> WindowBuilder<'a, R> {
...
     self
   }
+
+  /// Enables Incognito mode.
+  /// 
+  ///  ## Platform-specific:
+  /// 
+  ///  **Android**: Unsupported.
+  #[must_use]
+  pub fn incognito(mut self) -> Self {
```
```suggestion
  /// Enable or disable incognito mode for the WebView.
  /// 
  ///  ## Platform-specific:
  /// 
  ///  **Android**: Unsupported.
  #[must_use]
  pub fn incognito(mut self, incognito: bool) -> Self {
```

amrbashir (changes requested):

core/tauri-runtime/src/webview.rs
```diff
@@ -48,6 +49,9 @@ impl From<&WindowConfig> for WebviewAttributes {
     if let Some(effects) = &config.window_effects {
       builder = builder.window_effects(effects.clone());
     }
+    if config.incognito {
+      builder = builder.incognito(true);
+    }
```
amrbashir: move this up to line 39 and change it to
```suggestion
    builder = builder.incognito(config.incognito);
```

amrbashir (changes requested):
One last small nit-pick

core/tauri-runtime/src/webview.rs
Hyphrio: Should be done!

Hyphrio:
I think after this I'll probably pulling a small PR for the JS api, but if you can do it that would be nice

amrbashir (dismissed):
Thanks

amrbashir (dismissed):

amrbashir (approved):

PR #6813, opened 2023-04-27
feat(core): add IPC channel

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


wusyong (commented):
Is this mobile only feature?

lucasfernog:
> Is this mobile only feature?

I've also implemented it for desktop (it can easily be backported to v1) though the main idea here is to use it to send events from the Kotlin and Swift sides to the webview (eventually we'll also need a communication channel between mobile and Rust, but that's a little complicated since it involves static variables).

JonasKruckenberg:
Is this part of the public API?

lucasfernog:
Yeah it is part of the public API @JonasKruckenberg

chippers (changes requested):
I'm not fully up to date with mobile, but I see what you are trying to do with the channel communication here.

Only suggestion would be to cut the channel prefix once, instead of splitting on the string and grabbing the second.

chippers (dismissed):

chippers (dismissed):

JonasKruckenberg:
Can we hold off on merging this until I had the chance to review it tomorrow?

lucasfernog:
Sure @JonasKruckenberg

core/tauri/src/api/ipc.rs
```diff
@@ -10,6 +10,66 @@ use serde::{Deserialize, Serialize};
...
+    let window = command.message.window();
+    let value: String =
+      Deserialize::deserialize(command).map_err(|e| crate::Error::InvalidArgs(name, arg, e))?;
+    if value.starts_with(CHANNEL_PREFIX) {
+      let callback_id: Option<usize> = value
+        .chars()
+        .skip(CHANNEL_PREFIX.len())
+        .collect::<String>()
+        .parse()
+        .ok();
```
JonasKruckenberg: Maybe use `value.split_once(CHANNEL_PREFIX)`

tooling/api/src/tauri.ts
```diff
@@ -55,6 +55,29 @@ function transformCallback(
   return identifier
 }
 
+class Channel {
```
JonasKruckenberg: I would refactor this to be more similar to the `Websocket` or `BroadcastChannel` api i.e. the constructor takes no arguments and you can listen to events by overriding the `on message` property or calling `.addEventListener('message', () => {})`

JonasKruckenberg (changes requested):

core/tauri/src/api/ipc.rs
lucasfernog: Pushed.

tooling/api/src/tauri.ts
lucasfernog: Good idea, pushed.

wusyong (approved):

JonasKruckenberg (approved):

PR #6827, opened 2023-05-01
fix: default log path for linux and windows updated (fix: #6809)

### LINKED ISSUE #6809: [log] Default log path for linux and windows seems unconventional
> This issue has been upstreamed from https://github.com/tauri-apps/plugins-workspace/issues/1 

~~I believe the convention for storing the log files on Linux is, using the path `/var/log`. ~~ See edit below
Is there a specific reason for choosing `.config` instead? (https://github.com/tauri-apps/tauri-plugin-log/blob/dev/src/lib.rs#L106)

Edit: In the rest of the original issue we said `var/log/` isn't the solution and instead it should be `~/.local/share/` and `%LOCALAPPDATA%` respectively.

unknovvn:
- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes, and the changes were approved in issue #6809
- [ ] No

### Checklist


unknovvn:
Hey @FabianLars. There are multiple issues with rustc version in core test checks. Is there anything I should do to solve them?

FabianLars:
Nah don't worry about it, it's not caused by your change or anything. That's something we need to take a look at separately before the 1.3 release.

FabianLars:
Hmm, we may have to target the next branch (tauri v2) since this is technically a breaking change 🤔 I kinda tend towards making an exception since i expect this api to be fairly rarely used but let's wait for Lucas' opinion.

(We can change it in the Plugin, where this issue was raised either way for v1)

unknovvn:
@FabianLars are any additional changes required for this PR or are we just waiting for it to be approved?

FabianLars:
Just my last comment. We still need to figure out if we want to target v1 or v2 since it's technically a breaking change.

amrbashir:
I'd say this should be for v2 but maybe we could release it as a bug fix? cc @lucasfernog

lucasfernog:
This should land on v2 IMO

unknovvn:
What actions should be made from my side at this point?

amrbashir (dismissed):

amrbashir:
@unknovvn could you just rebase your branch on `next` branch to fix the conflict and trigger CI

amrbashir (dismissed):

amrbashir:
Looks like you need to run `cargo fmt` and you also need to add a change file, see https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md

amrbashir (approved):

amrbashir:
Oh looks like you didn't sign your commits and I can't merge the PR unless they are signed. You need to setup [commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits), then you can sign past commit like [this](https://superuser.com/questions/397149/can-you-gpg-sign-old-commits/1123928#1123928) for example.

unknovvn:
@amrbashir I have totally forgot about signing commits, fixed it now

unknovvn:
@amrbashir looks like merging is still blocked after I have signed all my commits. From commits tab I can see that the only commit which is not indicated as verified is the one where I have rebased my changes onto next branch. I have tried to use same approach as you have suggested earlier but it looks like for this commit it doesn't work

amrbashir:
@unknovvn I took care of it, thanks for your contribution

PR #6850, opened 2023-05-03
chore: update reproduction wording

amrbashir:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [X] Other, please describe: Github Bug Reports

- [ ] No

### Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -28,12 +28,14 @@ body:
 
   - type: textarea
     id: reproduction
+    validations:
+      required: true
     attributes:
       label: Reproduction
       description: A link to a reproduction repo.
       placeholder: |
-        Please provide a [minimal reproduction](https://stackoverflow.com/help/minimal-reproducible-example)
-        > [Why reproduction is required](https://antfu.me/posts/why-reproductions-are-required)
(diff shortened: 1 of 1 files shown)


simonhyll:
Mixed feelings. Demanding a repro is understandable but you dont want to risk not being made aware of an issue just because a user doesnt know how to create a repro. If we demand it I think we might wanna make a special "repro" template that people can use similar to the one we have but more catered to just debugging, along with some basic guides on how to make a good repro

Even if the OP fails to provide sufficient repro info the fact they created an issue still has some value

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -28,13 +28,14 @@ body:
 
   - type: textarea
     id: reproduction
+    validations:
+      required: true
     attributes:
       label: Reproduction
-      description: Steps to reproduce the behavior.
+      description: A link to a reproduction repo.
```
wusyong:
```suggestion
      description: A link to a reproduction repo or steps to reproduce the behaviour.
```

wusyong:
@simonhyll I think it's okay to add this. Users who don't provide such info will still find a way to do it.
@amrbashir Maybe if someone skip this, we just inform them to add it and add `need more info` status label for it.

wusyong (approved):

PR #6861, opened 2023-05-04
perf: remove lto setting from CLI

CrabNejonas:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

This PR removes the `lto=true` setting for the CLI. This drastically improves the install performance (the time it takes to run `cargo install tauri-cli`).
Since the CLI is not doing performance-critical work and size is *not important* here since the cli will be compiled locally this does not have a downside it it afaik and only improves performance for rust users.

To give you an idea here are the install times on my MacBook Pro m2:

| build           | time        |
|------------|---------|
| with lto       | 2:55.31 |
| without lto | 1:05.46 |

so as you can see disabling lto reduced the compile-time by 3x.


FabianLars:
what do you think about moving the lto (and the new stuff you added in the js api) into a custom profile that inherits the release profile so that we can use that for the cargo-binstall CI?

amrbashir:
what about binstall?

lucasfernog:
I had to use patch-package because napi-rs does not support custom profiles :(

lucasfernog:
Also @JonasKruckenberg why did you target `next` here instead of `dev`?

amrbashir:
> I had to use patch-package because napi-rs does not support custom profiles :(

I would probably look into upstreaming this to `napi-rs`, the maintainer is pretty responsive and would probably publish it soon after merging.

Also if I understand correctly, this change wants to improve `cargo install tauri-cli` which iirc will use `release` profile by default which doesn't have this change and users would need to `cargo install tauri-cli --profile release-size-optimized` which is bad DX

lucasfernog:
He's focusing on v3 so I'm not sure if he wants to patch for v2. I'll try anyway.

lucasfernog:
> > I had to use patch-package because napi-rs does not support custom profiles :(
> 
> I would probably look into upstreaming this to `napi-rs`, the maintainer is pretty responsive and would probably publish it soon after merging.
> 
> Also if I understand correctly, this change wants to improve `cargo install tauri-cli` which iirc will use `release` profile by default which doesn't have this change and users would need to `cargo install tauri-cli --profile release-size-optimized` which is bad DX

I think you misunderstood. We don't want users to use `release-size-optimized`, we want them to have faster builds. The size optimization will be used only on the cargo-binstall and NPM assets.

amrbashir:
ah right, I was even more confused by the name. LGTM then.

amrbashir (dismissed):

FabianLars (dismissed):

lucasfernog:
@CrabNejonas we also need to rebase this branch with commit signing :cry:

lucasfernog (approved):

PR #6842, opened 2023-05-03
add workspace inheritance

JonasKruckenberg:
### What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

### Does this PR introduce a breaking change?

- [ ] Yes, and the changes were approved in issue #___
- [ ] No

### Checklist
- [ ] When resolving issues, they are referenced in the PR's title (e.g `fix: remove a typo, closes #___, #___`)
- [ ] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md).
- [ ] I have added a convincing reason for adding this feature, if necessary

PR #6885, opened 2023-05-06
use cargo workspace package table

chippers:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
We can use the [Cargo Workspace Package Table](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table) in order to set the same values that are used for many `[package]` keys in most packages. I think we could also use this for tooling and examples, but we need to figure out how we want to structure the workspace before that.


chippers:
@FabianLars will this also run into the problem you described in https://github.com/tauri-apps/tauri/pull/6842#issuecomment-1532789897?

FabianLars:
Yeah, covector will convert the syntax to
```toml
  [package.authors]
  workspace = true
```
and will remove the comment.

Edit: covector seems to rewrite the whole file and not just the parts it actually has to modify.

lucasfernog:
I actually had to use the `dep = { workspace = true }` syntax instead of `dep.workspace = true` to not get that ugly format from covector.

chippers:
Unfortunate, is this a problem with covector or a dependency of it? Is this tracked somewhere?

lucasfernog:
It's a problem with https://github.com/tauri-apps/tauri-toml @chippers

lucasfernog:
I'll see if updating the fork fixes it.
EDIT: nope, no easy way to adjust it :(

lucasfernog (approved):

PR #6933, opened 2023-05-10
feat(bundler/NSIS): uninstall previous wix if exist, closes #6859

### LINKED ISSUE #6859: [feat] Upgrade from WiX to NSIS bundle
### Describe the problem

I'd like to switch my users that currently use the `.msi` bundle over to the new nsis target, preferably via manual installation _and_ via the updater.

I _think_ it would be better to ignore the old installation path because of the new default perUser install scope -> Of course we need to document this if we decide to to this.

This was also asked a few times on discord already. [One person](https://discord.com/channels/616186924390023171/731495028677148753/1103512731304333382) tried it with the current implementation and reported this:
> in my testing, it correctly updates the app but now there are system-wide registry keys for the msi install and user-local keys for the user install. so now there are two entries in the Windows uninstall panel and uninstalling either will obviously delete the application.

So it sounds like the wix uninstaller isn't executed? which would make sense to me since afaik we didn't really think much about this use case before, right?

See above x)

### Alternatives considered

_No response_

### Additional context

_No response_

amrbashir:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


YichenZhou113:
I think the changes were checked in as I saw in `Cargo.lock`
```@@ -3180,8 +3180,7 @@ checksum = "8ae9980cab1db3fceee2f6c6f643d5d8de2997c58ee8d25fb0cc8a9e9e7348e5"
 [[package]]
 name = "tauri"
 version = "1.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d42ba3a2e8556722f31336a0750c10dbb6a81396a1c452977f515da83f69f842"
+source = "git+https://github.com/tauri-apps/tauri?branch=feat/nsis/uninstall-wix#7879e663dc7baf27ccc8587ed4572b457ea0f550"
```
`.msi` was created by Wix and `.exe` was created by NSIS, I tried:

- Installing my tauri app using `.msi` in default C:\drive and `.exe` in default D:\drive
- Installing my tauri app using `.msi`  and `.exe` to same folder in D:\ drive

The NSIS `.exe` installer did not detect the `.msi` installed application and did not ask for uninstall/upgrade, and there are still two uninstallers.

amrbashir:
@YichenZhou113 thanks for testing.

This change doesn't affect the rust crate `tauri` but rather the CLI, you can install it using `cargo install tauri-cli --git https://github.com/tauri-apps/tauri --branch feat/nsis/uninstall-wix` then do `cargo tauri build` to build the nsis and then run it.

Raphiiko:
I've attempted to test it, and mostly it seems to work! The new NSIS installer detects another version installed with the WIX installer, and gives the user the option to uninstall it first. 

However, I run into a weird issue with it:

If I uninstall the application installed with the NSIS installer, it fails to remove itself from the registry after uninstalling, only if during the installation I uninstalled another version installed with the WIX installer. 

To clarify:
1. Install old version of application with WIX installer
2. Start installation of new version with NSIS installer
3. NSIS installer prompts to uninstall older version with WIX installer
4. old version gets uninstalled (and removed from registry just fine)
5. NSIS installer installs new version
6. Uninstall new version 
7. Registry entry for new version fails to be removed.

The registry entry is removed during uninstallation just fine if no prior WIX-based installation is uninstalled during the installation

I've not tested this with the updater. Only by directly running the bundled installers.

amrbashir:
@Raphiiko thank you very much for testing, could you tell me the path of the registry key that failed to be rmeoved?

Raphiiko:
I believe it is `Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\<MyAppName>`.

It's the one that keeps the application listed in the Add/Remove programs list in Windows' settings (even though it's already uninstalled): 

image

amrbashir:
`Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\<MyAppName>` is for `perMachine` installation, did you configure your NSIS installer installmode to use `perMachine` or `both`?

Raphiiko:
I configured it as `both`, and only used 'Install for anyone using this computer' in the NSIS installer while testing.

I picked `both` as I assumed this keeps the installation directory the same for users upgrading from their existing wix-based installations (`perMachine`), but gives new users the option to install it just for their own user account (`currentUser`)

amrbashir:
@Raphiiko thanks, this should be fixed in the last commit

Raphiiko:
@amrbashir I can confirm that this solved the issue on my end. Thanks a lot!

lucasfernog (approved):

PR #6935, opened 2023-05-10
fix(nsis): prefill $INSTDIR with previous install path and respect `/D` flag, closes #6928

### LINKED ISSUE #6928: [feat] Overriding installation path for NSIS installation packages during silent instalation doesn't work
### Describe the problem

Usually the installation package generated with NSIS allows to pass the installation directory via command line. This is especially important for silent installations. The command line looks like this:
```
installer.exe /S /D=C:\Users\Public\installDir
```
In order the installer to support this parameter the code in the installer must look something like this:
```nsis
${If} $InstDir == "" ; /D not used
    ; custom logic to set installation path
${EndIf}
```
NSIS template in the Tauri codebase (\tooling\bundler\src\bundle\windows\templates\installer.nsi ) doesn't check against if the $InstDir is already set (via command line) and just sets the $InstDir value based on install mode.

I think the NSIS installation template must be modified so that the $InstDir  will be set only if it wasn't set via command line parameter. Something like this:
```nsis
${If} $InstDir == "" ; /D not used
  !if "${INSTALLMODE}" == "perMachine"
    ; Set default install location
    ${If} ${RunningX64}
      !if "${ARCH}" == "x64"
        StrCpy $INSTDIR "$PROGRAMFILES64\${PRODUCTNAME}"
      !else if "${ARCH}" == "arm64"
        StrCpy $INSTDIR "$PROGRAMFILES64\${PRODUCTNAME}"
      !else
        StrCpy $INSTDIR "$PROGRAMFILES\${PRODUCTNAME}"
      !endif
    ${Else}
      StrCpy $INSTDIR "$PROGRAMFILES\${PRODUCTNAME}"
    ${EndIf}
  !else if "${INSTALLMODE}" == "currentUser"
    StrCpy $INSTDIR "$LOCALAPPDATA\${PRODUCTNAME}"
  !endif
!endif
```

### Alternatives considered

_No response_

### Additional context

_No response_

amrbashir:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/.changes/nsis-restore-installation-path.md
+++ b/.changes/nsis-restore-installation-path.md
@@ -1,0 +1,5 @@
+---
+'tauri-bundler': 'patch'
+---
+
+Fix NSIS installer not using the old installation path as a default.
--- a/.changes/nsis-restore-installation-path.md
+++ b/.changes/nsis-restore-installation-path.md
@@ -2,4 +2,5 @@
 'tauri-bundler': 'patch'
 ---
 
-Fix NSIS installer not using the old installation path as a default.
+- Fix NSIS installer not using the old installation path as a default when using `perMachine` or `currentUser` install modes.
+- NSIS will now respect the `/D` flag which used to set the installation directory from command line.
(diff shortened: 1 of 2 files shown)


lucasfernog (approved):
Is it intentional that even if I uninstall the app, it still has the last install path restored instead of using the default one?

amrbashir:
yeah that's what I intended, but I could change that if users wants to.

PR #6945, opened 2023-05-11
docs: Fix api.js docs pipeline with updated typedoc dependencies

FabianLars:

- [x] Bugfix
- [ ] Feature
- [x] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
This was what broke the publishing pipeline when we tried to release 1.3 - it should work now but i'll also need to see if the docs are still correct by integrating it into tauri-docs before we merge this since the js-api.json file is much bigger now and i'm not sure yet if that's correct...


FabianLars:
Okay looks like they just changed the output format a bit. Content itself looks fine and since we don't actively use the json output yet the change won't affect us.

The actual docs on tauri.app will still be generated with 0.23 though since our fork of the markdown plugin diverged from upstream too much and i can't (and don't want to) figure it out how to update it right now.

wusyong (approved):

lucasfernog (approved):

PR #6960, opened 2023-05-13
feat(core): fill and check allowlist for plugin

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


lucasfernog:
Maybe we should use [cargo:rustc-cfg=feature="feature-flag-name"](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorustc-cfgkeyvalue) instead. (though it could also be too magic and harder to trace issues).

amrbashir (approved):
tbh, I don't think we need to add an allowlist for plugins and instead we should rely on feature flags because that's what it all boils down to eventually. I know it is gonna be a bit hard but we are already going to require users to modify Cargo.toml to add the plugins they need so it should be fine.

Also this would allow flexibility for non-official plugins where they can provide their own feature flags that don't necessarily start with `allow-*`. 


This will also have the added benefits of removing the need for our CLI to manage the feature flags (which is not pretty and can mess up formatting) for each plugin which would result in less maintenance burden for us, less code and maybe fewer build-dependencies.

Other than that, the PR looks good to me.

lucasfernog:
We won't really force the users to touch cargo.toml to add plugins, we'll do that with the tauri add CLI command.

simonhyll:
> Maybe we should use [cargo:rustc-cfg=feature="feature-flag-name"](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorustc-cfgkeyvalue) instead. (though it could also be too magic and harder to trace issues).

As in instead of adding features to Cargo.toml just making the Tauri CLI add the appropriate features? Because that I do love in no small part because my IDE and Tauri don't agree on the formatting of Cargo.toml, so anytime there's a change to the features I need to reformat Cargo.toml

The main potential issue I'd have with that is that then running `cargo build` may fail, right? Because the ability to build without the tauri cli is to me more important than ease of use in this case (when I develop in a workspace I'll often run `cargo update && cargo build` and I wouldn't want that to fail)

lucasfernog:
> > Maybe we should use [cargo:rustc-cfg=feature="feature-flag-name"](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorustc-cfgkeyvalue) instead. (though it could also be too magic and harder to trace issues).
> 
> As in instead of adding features to Cargo.toml just making the Tauri CLI add the appropriate features? Because that I do love in no small part because my IDE and Tauri don't agree on the formatting of Cargo.toml, so anytime there's a change to the features I need to reformat Cargo.toml
> 
> The main potential issue I'd have with that is that then running `cargo build` may fail, right? Because the ability to build without the tauri cli is to me more important than ease of use in this case (when I develop in a workspace I'll often run `cargo update && cargo build` and I wouldn't want that to fail)

@simonhyll actually each plugin's build script would be enabling the feature instead of the CLI messing with the Cargo.toml file.

lucasfernog:
I remember @tweidinger also had some ideas for an allowlist refactor.

lucasfernog:
We're investigating other approaches for the plugin allowlist. A static Cargo feature based check is not enough to support multi-window configurations.

PR #6966, opened 2023-05-15
fix(core): unpin all dependencies, closes #6944

### LINKED ISSUE #6944: [feat] Less strict MSRV enforcement
#### tldr: Let's unpin our dependencies :)

For context the discord discussion i started that lead to this request: https://discord.com/channels/616186924390023171/986184168998330371/1106155646769954867

Note that this request is not about the MSRV itself, though imo we should consider something like "last 1 year of rust releases" in V1, which means we could upgrade to 1.61 in a week. (i don't think 1 year is the healthiest approach, but we can take a more drastic change in v2 instead).

----

### Problem

Since v1.0 we regularly had to lock dependencies on minor and patch versions because they upgraded their MSRV to a more recent version than what we promised to our users. In v1.3 we even had to pin `time` to 0.3.15. After one week we already had at least 2 users facing issues because of this. In this case
1) was caused by `gix` requiring a min time version of 0.3.19
2) was caused by user-side code being written with APIs from more recent time versions in tauri v1.2 who then tried to upgrade to v1.3

As time goes on we will have to lock more and more dependencies, causing more and more issues for end-users, just because we tried to make it easy for users that actually required a low MSRV (if we're being honest there probably isn't even one that needs a MSRV _that_ low but whatever).

### Suggested solution

Going forward i think our approach should be a bit less "forceful":
- No more dependency minor/patch pinning
- Our code should still adhere to our MSRV, so no std apis introduced in newer Rust versions than what our MSRV allows
- Set the **minimum** version of our dependencies to a version that is compatible to our MSRV. Of course we can't use apis of these crates that were introduced in a version that raised their MSRV too high.
  - Taking `time` as an example the minimum version should be less or equal to `0.3.15` since that's the last version with a compatible MSRV - in this case `0.3` would be enough.

This way end-users can actually "choose" which MSRV they want:
- If they want the latest of everything and their toolchain is always up to date, then they don't need to care about anything.
- If they want to adhere to Tauri's MSRV for example, they have to make sure all their other dependencies have a compatible MSRV themselves, or with https://github.com/rust-lang/cargo/issues/9930 i guess cargo can potentially take care of that too.
  - For example they could run `cargo update --package time --precise 0.3.15` to update their lockfile to make all dependencies use that version of `time`. Of course this requires that other crates also have to set appropriate _minimum_ dependency versions, **but this is not and should not be our problem**.

----

One issue with this approach is that we'll likely have to start committing `Cargo.lock`, primarily to make sure the MSRV workflow keeps working, but i guess this could potentially "soft lock" the dependencies for users too which _should_ be a good thing in this context since that _should_ mean cargo will try to adhere to the lockfile and give users a MSRV compatible version until they run `cargo update` - this needs confirmation but this is what comments in and around the linked issue above sounded like to me.

Alternatively there's https://doc.rust-lang.org/cargo/reference/unstable.html#minimal-versions which we _may_ be able to use too if we don't want to commit a lock file.

----

This issue was written in the heat of the moment so i may update it later.

FabianLars:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: No idea tbh

- [x] No, unless users actually depend on our MSRV and run `cargo update` without pinning the deps themselves

### Checklist

### DIFF
--- a/.changes/unpin-deps.md
+++ b/.changes/unpin-deps.md
@@ -1,0 +1,9 @@
+---
+'tauri': patch
+'tauri-build': patch
+'tauri-codegen': patch
+'tauri-runtime': patch
+'tauri-runtime-wry': patch
+---
+
+Unpin `time`, `ignore`, `winnow`, and `ignore` crate versions. Developers now have to pin crates if needed themselves. A list of crates that need pinning to adhere to Tauri's MSRV will be visible in Tauri's GitHub workflow: https://github.com/tauri-apps/tauri/blob/dev/.github/workflows/test-core.yml#L85.
--- a/.github/workflows/test-core.yml
+++ b/.github/workflows/test-core.yml
@@ -82,5 +82,13 @@ jobs:
           workspaces: core -> ../target
           save-if: ${{ matrix.features.key == 'all' }}
 
+      - name: Downgrade crates with MSRV conflict
+        # The --precise flag can only be used once per invocation.
+        run: |
+          cargo update -p time --precise 0.3.15
+          cargo update -p ignore --precise 0.4.18
+          cargo update -p winnow --precise 0.4.1
+          cargo update -p raw-window-handle --precise 0.5.0
+
       - name: test
         run: cargo test --target ${{ matrix.platform.target }} ${{ matrix.features.args }}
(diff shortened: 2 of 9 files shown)


FabianLars:
As an addition to this PR we should look into changing the versions of the dependencies to not include patch versions (or minor version) unless necessary like we did for `tokio` and a few more when someone complained about our high minimum `tokio` version. If you want we can do it in this PR too i guess :)

wusyong (approved):
I think we can pass that. We should loose the restriction first.

PR #6968, opened 2023-05-15
refactor(android): fix race conditions processing Android plugin project as dependency

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


lucasfernog:
Hopefully I got it right this time. Unfortunately I couldn't find a way to force a gradle project sync when the custom task finishes, so the initial build will fail since Gradle won't pick up the dependency changes :(

lucasfernog:
I've had to readd the script that runs an initial build to fill the project dependencies before running the actual build. It sucks, but it's the best solution right now.

amrbashir:
Is there a reason to move it to the CLI, if there is not much to gain from this, I think we should keep it in `tauri-build` so one could potentially use `cargo build --target aarch64-linux-android` and use something other than our CLI to build the apk.

> Hopefully I got it right this time. Unfortunately I couldn't find a way to force a gradle project sync when the custom task finishes, so the initial build will fail since Gradle won't pick up the dependency changes :(

Could you explain a bit more? I will ask a friend of mine if he can help with this

lucasfernog:
We need to include each plugin project to the application's Android project as a dependency. We do this by changing the project's `build.gradle.kts` and `settings.gradle` files to include the plugin project as a path dependency. The Rust way of doing this is through the build script, which is the only way the CLI can know where the plugin project is located in the file system.

The problem with this is that when the build script is executed, Gradle already processed all the dependencies. So we actually need to trigger a build ahead of tauri dev / tauri build.

lucasfernog:
> Is there a reason to move it to the CLI, if there is not much to gain from this, I think we should keep it in `tauri-build` so one could potentially use `cargo build --target aarch64-linux-android` and use something other than our CLI to build the apk.

I'll still testing this, I think I finally found a way to fix the race conditions.

lucasfernog:
I'll test more tomorrow. One of the issues i'm seeing is that the app build script does not process all plugins when the project is deleted and initialized again.

lucasfernog:
I gave up and used cargo's built in communication way for build scripts (the `links` manifest field). The problem is that cargo builds the dependencies concurrently, so there's no good way to read each plugin output via files.

The only downside to this is that it forces the plugins to be a direct dependency of the app crate (i.e. you can't have a crate that re-exports other plugins, unless you replicate the Cargo output).

It's either this or we get the logic to rewrite the Gradle files back to the CLI.

https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key

amrbashir:
should we just publish the android projects to maven/jitpack? if not, `links` sounds a bit awful in terms of DX and would just perfer moving the logic to the CLI

lucasfernog:
If we publish the Android projects, we also need to provide a mechanism to update the dependencies on the app's project. And keeping the Rust crate and the Android dependency in sync will be a nightmare, since probably two different versions will be incompatible (I think that's also why Capacitor choose to use this path dependency approach).

At least the `links` approach does not affect the end user, only plugin developers. And the CLI one has the minimum impact, though it introduces another layer of dependency between the CLI and the tauri-build.

amrbashir:
> At least the `links` approach does not affect the end user, only plugin developers

How so? won't the users need to declare `links` in their tauri project `Cargo.toml`?

lucasfernog:
The `links` field only need to be declared on the plugin Cargo.toml. That wat it can communicate with the app's build script via `println!("cargo:key=value")` which can be read by the app build script via `std::env::var("DEP_{dep_name}_KEY")`, which we do in tauri-build.

amrbashir:
Ah gotcha, that should be fine then

amrbashir (approved):

lucasfernog:
This approach is also safer since indirect dependencies cannot declare deps to be added to the Android project. I've also added `links` to the plugin template.

PR #6982, opened 2023-05-17
fix(cli): improve vs build tools detection

FabianLars:
This makes the msvc/visual studio detection a bit more reliable since it now looks for the actually required components that are _optional_ to the C++ workload we tested for before.

This fixes 2 problems:
1) Users installed the C++ workload but not the sdk and msvc optional components
2) Users did not install the whole workload but only msvc and the sdk (that's what i always do, and iirc rustup's auto installer does too)

Most of these changes are ~~copied~~ heavily inspired by rustup's detection.

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


tooling/cli/src/info/env_system.rs
```diff
@@ -30,23 +58,24 @@ fn build_tools_version() -> crate::Result<Option<Vec<String>>> {
...
-      "-requires",
-      "Microsoft.VisualStudio.Workload.VCTools",
+      "Microsoft.VisualStudio.Component.VC.Tools.x86.x64",
       "-format",
       "json",
     ])
     .output()?;
-  Ok(if output.status.success() {
+
+  Ok(if output.status.success() && has_windows_sdk_libs() {
```
amrbashir: maybe separate this check into its own item with its own error message?

amrbashir (changes requested):

tooling/cli/src/info/env_system.rs
FabianLars: Yes, we _could_ do that, but all errors of build_tools_version are ignored (unwrap_or_default) anyway so we'd need to overhaul the whole error handling here, especially since the other specific tests don't return any errors either, so far it was only used for `?` on the `Command` and `serde_json::from_str` lines.

( -> We probably should have went with return Option<> instead of Result<Option> in the first place now that i look at it)

amrbashir: `unwrap_or_default` is fine here since we are only looking for installed instances of visual studio, and we only print the error if we couldn't find any.

`has_windows_sdk_libs` on the other hand, could be `false` while vs is installed and so it should be its own section item and have its own error message

FabianLars: I see where you're coming from but shouldn't we then also split up the VC.Tools check from the general VS instance check with that logic? I mean, we don't even actually care about VS itself but just these 2 components. (p.s. i just understood what you initially asked for i thought you meant how we return stuff from build_tools_version, not that you wanted to split it into its own standalone function)

The only reason we're checking for the sdk libs here like this is because there is no defined name of the sdk component because it contains the version in its ID

FabianLars: ~~actually you _can_ get the sdk libs by using `Microsoft.VisualStudio.Component.Windows10SDK.*` || `Microsoft.VisualStudio.Component.Windows11SDK.*` but there is _one_ wrong component that can also trigger this...~~
Nevermind, still doesn't work. vswhere just had a hickup

FabianLars: I'll try to change it according to your suggestion and change the wording a bit to be more about the components, mentioning VS as the way to get them

FabianLars:
I switched back to use vswhere only, without the custom search because the one component i mentioned in the comment above should hopefully be rare enough (IpOverUsb).

Anyway, vswhere was really wonky with weird hickups when i started out which is why i switched to the custom search. Later on it seemed to work fine but i'd prefer if a few people with good internet (or patience) can test it a few times...

tooling/cli/src/info/env_system.rs
```diff
@@ -20,7 +20,9 @@ struct VsInstanceInfo {
 const VSWHERE: &[u8] = include_bytes!("../../scripts/vswhere.exe");
 
 #[cfg(windows)]
-fn build_tools_version() -> crate::Result<Option<Vec<String>>> {
+fn build_tools_version() -> crate::Result<Vec<String>> {
+  use itertools::Itertools;
```
amrbashir: not sure why you need this import?

amrbashir (changes requested):

tooling/cli/src/info/env_system.rs
FabianLars: i used its `unique()` method to filter out duplicate VS version, in case users have both, a win10 and win11 sdk. I'll sort it and dedup it instead so we can get rid of that import :)

Fun fact: rust-analyzer actually imported it for `extend()` because i added that to the iterator by accident. Then i just went with that import for deduping the vec/iterator 🤷

amrbashir: ah didn't see the `unique` usage, it is fine to keep then, I just thought it was accidentally imported or imported for `extend`

FabianLars: > I just thought it was accidentally imported or imported for extend

You know me too well 😂

amrbashir (approved):

PR #6994, opened 2023-05-18
feat(cli): use `native-tls-vendored` on mobile instead of `rustls-tls`

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/tooling/cli/src/interface/rust.rs
+++ b/tooling/cli/src/interface/rust.rs
@@ -349,19 +349,15 @@ fn shared_options(
   features: &mut Option<Vec<String>>,
   app_settings: &RustAppSettings,
 ) {
-  if mobile {
-    features
-      .get_or_insert(Vec::new())
-      .push("tauri/rustls-tls".into());
-  } else {
-    let all_features = app_settings
-      .manifest
-      .all_enabled_features(if let Some(f) = features { f } else { &[] });
-    if !all_features.contains(&"tauri/rustls-tls".into()) {
-      features
-        .get_or_insert(Vec::new())
-        .push("tauri/native-tls".into());
-    }
+  let all_features = app_settings
+    .manifest
+    .all_enabled_features(if let Some(f) = features { f } else { &[] });
+  if !all_features.contains(&"tauri/rustls-tls".into()) {
+    features.get_or_insert(Vec::new()).push(if mobile {
+      "tauri/native-tls-vendored".into()
+    } else {
+      "tauri/native-tls".into()
+    });
   }
 }
 
--- a/.changes/android-vendored-openssl.md
+++ b/.changes/android-vendored-openssl.md
@@ -1,0 +1,6 @@
+---
+"cli.rs": patch
+"cli.js": patch
+---
+
+Use `native-tls-vendored` instead of `rustls-tls` by default on mobile.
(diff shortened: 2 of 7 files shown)


tooling/cli/src/interface/rust.rs
```diff
@@ -349,19 +349,15 @@ fn shared_options(
...
-    }
+  let all_features = app_settings
+    .manifest
+    .all_enabled_features(if let Some(f) = features { f } else { &[] });
+  if !all_features.contains(&"tauri/rustls-tls".into()) {
+    features.get_or_insert(Vec::new()).push(if mobile {
+      "tauri/native-tls-vendored".into()
+    } else {
+      "tauri/native-tls".into()
+    });
```
amrbashir, line 360: do we really need to manage these features on the CLI? can't we just let the users activate the feature he needs? 
and in the CLI template we should probably be using:
```toml
[target."cfg(not(any(\"android\", \"ios\")))".dependencies]
tauri = { version = "2.0.0-alpha.9", features = []}

[target."cfg(any(\"android\", \"ios\"))".dependencies]
tauri = { version = "2.0.0-alpha.9", features = ["native-tls-vendored"]}
```

lucasfernog: That has more implications, like the manifest rewrite we currently have (we automatically enable "system-tray, macos-private-api, isolation and protocol-asset). This duplication is also ugly :(

amrbashir: yeah, it is probably fine to keep in the CLI, but maybe only apply the feature flag if none of the 3 flags is used?

lucasfernog: We actually only need to check for the rustls-tls feature. Enabling `native-tls` when the vendored flag is enabled does not change anything.

lucasfernog:
And seems like we'll need to stick with rustls-tls, Windows still doesn't work because or Perl.

amrbashir:
lets keep it as is for now, is there any disadvantage to using `rustls-tls`?

lucasfernog:
Not yet. But the lack of option can bite us in the future.

lucasfernog:
It would be nice if we could use `native-tls-vendored` because it would mean less feature flag management on the plugin side, see the [API example](https://github.com/tauri-apps/plugins-workspace/pull/375).

amrbashir:
I wasted so much time in the past on openssl/perl issue with Windows with no avail.

lucasfernog:
Alright, let's punt this again.

wusyong:
@amrbashir I think PR is because of https://github.com/tauri-apps/plugins-workspace/pull/375
My concern is Rustls uses ring which is pure assembly. It doesn't support many platform you would imagine.
Did you confirm it rustls can work on most emulator and device platform? (like 32/64 bit, x86-64/arm, apple/android etc)

amrbashir:
I have been using tauri on emulator and real android devices, and I never had any issues (assuming tauri was always using `rustls-tls`)

wusyong:
Alright if you both have no problem with it then let's close this one.
I'll review the PR in plugin workspace later.

PR #7026, opened 2023-05-21
feat: current_version feature

### LINKED ISSUE #6995: [feat] Expose Wry::webview_version through tauri
### Describe the problem

I want to collect webview versions in my app for logging purposes so if user reports a rendering problem or similar we know how to reproduce.

some sort of reexport from tauri, like `tauri::wry::webview_version()`. There may be other related functions that I'm not thinking about as well.

### Alternatives considered

I can add a direct dependency on wry which works. What isn't clear is any risks of this. how do I ensure I get the same version of wry? is this going to bloat my binary at all?

### Additional context

_No response_

amirHdev:
fixes : #6995 

As I see in this example : 
```rust
  let _webview = WebViewBuilder::new(window)?
    .with_url("https://tauri.studio")?
    .build()?;
```
it will return WindowWrapper and suppose current_version can fetch with it . but with collaborator guide can move it .

### DIFF
--- a/.changes/add-webview-version.md
+++ b/.changes/add-webview-version.md
@@ -1,0 +1,5 @@
+---
+'tauri': 'patch'
+---
+
+Added `tauri::webview_version` , to get webview version.
--- a/.changes/add-webview-version.md
+++ b/.changes/add-webview-version.md
@@ -1,5 +1,5 @@
 ---
-'tauri': 'patch'
+'tauri': 'minor'
 ---
 
 Added `tauri::webview_version` , to get webview version.
--- a/core/tauri-runtime-wry/src/lib.rs
+++ b/core/tauri-runtime-wry/src/lib.rs
@@ -67,6 +67,7 @@ use wry::{
   webview::{FileDropEvent as WryFileDropEvent, Url, WebContext, WebView, WebViewBuilder},
 };
 
+pub use wry::webview::webview_version;
 pub use wry::application::window::{Window, WindowBuilder as WryWindowBuilder, WindowId};
 
 #[cfg(windows)]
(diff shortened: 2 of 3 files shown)


lucasfernog:
To add a new webview/window function you actually need to change the tauri-runtime and tauri crates. So the webview_version function would need to be added under [here](https://github.com/tauri-apps/tauri/blob/ac183948d610a01ac602846cddd80563c86c45af/core/tauri-runtime/src/lib.rs#L619), [here](https://github.com/tauri-apps/tauri/blob/ac183948d610a01ac602846cddd80563c86c45af/core/tauri-runtime-wry/src/lib.rs#L1325) and [here](https://github.com/tauri-apps/tauri/blob/ac183948d610a01ac602846cddd80563c86c45af/core/tauri/src/window.rs#L1122).

amrbashir:
It doesn't need to be a method on the `Window` type, it just needs to be re-exported in `tauri-runtime-wry` then re-exported in `tauri`

lucasfernog:
Ahh right. Easier that way.

amirHdev:
@lucasfernog you can review it !

amrbashir:
you need to also add a file like this one https://github.com/tauri-apps/tauri/blob/dev/.changes/appimage-follow-symlinks.md with a quick summary about this change/addition.

amirHdev:
@amrbashir Done .

amrbashir (dismissed):

.changes/add-webview-version.md
```diff
@@ -0,0 +1,5 @@
+---
+'tauri': 'patch'
```
amrbashir:
```suggestion
'tauri': 'minor'
```

amrbashir (approved):

lucasfernog:
Nice one @Mehrbod2002 !

PR #7058, opened 2023-05-25
feat(cli/mobile/init): skip installing already installed targets, closes #7044

### LINKED ISSUE #7044: [feat] skip rustup installation for cargo tauri android init (for nix user)
### Describe the problem

I'm getting a nix flake set up for my project.  I want to avoid using rustup as that is a system wide (or at least user-wide) tool.  I'd like to have as many build tools as possible specified in my flake.nix.  

I already have the rustc android targets installed and rustc and so forth.  But cargo tauri android init fails because rustup is not found.  I don't want to install the compiler and etc with rustup because I already have that.  

I'd like the tool to skip the rustup step.  Some ways to do that:

- have an optional --no-rustup flag
- check for compile targets being present before trying to call rustup to install them

### Alternatives considered

just install things with rustup during project generation, then delete the rustup installed compiler again when its through.  probably will do this but would rather not long-term.  

### Additional context

_No response_

amrbashir:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/.changes/cli-skip-targets-install.md
+++ b/.changes/cli-skip-targets-install.md
@@ -1,5 +1,6 @@
 ---
 'cli.rs': 'patch'
+'cli.js': 'patch'
 ---
 
 Add `--skip-targets-install` flag for `tauri android init` and `tauri ios init` to skip installing needed rust targets vie rustup.
--- a/.changes/cli-skip-targets-install.md
+++ b/.changes/cli-skip-targets-install.md
@@ -3,4 +3,4 @@
 'cli.js': 'patch'
 ---
 
-Add `--skip-targets-install` flag for `tauri android init` and `tauri ios init` to skip installing needed rust targets vie rustup.
+Skip Rust target installation if they are already installed.
--- a/.changes/cli-skip-targets-install.md
+++ b/.changes/cli-skip-targets-install.md
@@ -1,0 +1,5 @@
+---
+'cli.rs': 'patch'
+---
+
+Add `--skip-targets-install` flag for `tauri android init` and `tauri ios init` to skip installing needed rust targets vie rustup.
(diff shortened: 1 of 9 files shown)


.changes/cli-skip-targets-install.md
```diff
@@ -0,0 +1,5 @@
+---
+'cli.rs': 'patch'
```
lucasfernog, line 2:
```suggestion
'cli.rs': 'patch'
'cli.js': 'patch'
```

lucasfernog:
@amrbashir I think it's easier if we detect installed targets instead. Let me know if you have something against it.

amrbashir:
It is a good addition but I still think we should provide the flag anyways. Some users may want to install only install 1 or 2 targets and the flag would allow them to do so.

lucasfernog:
The flag sounds a little feature creep to me, I would avoid unless it's really necessary.

This check is already addressing a super particular use case :D usually you'll just want everything installed for you to just run `android build` and expect it to work flawlessly.

amrbashir:
Ideally, `tauri dev`, `tauri build`, `tauri android dev`, `tauri android build`...etc would check and install the need requirements if possible but would, at the same time, provide a flag to skip this part, otherwise unusual environments like Nix will fail.

lucasfernog:
Yeah but if there's a good way to detect instead of providing a skip flag.. :wink:

amrbashir:
Detection always has the possibility of missing 0.01% of users and these are the users that open github issues that will ask for a flag. I like detecting and auto fixes but at the same time I also like being in control and as tauri-cli is considered a build tool, everything it does should be configurable IMO. And I don't really consider it a feature creep since the implementation is actually simple and doesn't affect anything else.

lucasfernog:
I tend do go with the `wait and see` approach in these cases.. like only adding that check when it's requested and there's no better way to do so. But we can open another PR to continue the discussion.

lucasfernog (approved):

PR #7098, opened 2023-05-31
feat(cli): Build the CLI for Windows ARM64, closes #7015

FabianLars:
This worked locally, i bet 10 bucks that it will fail in CI

### What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [x] Other, please describe: I have no idea which boxes to tick here

### Does this PR introduce a breaking change?

- [ ] Yes, and the changes were approved in issue #___
- [x] No

### Checklist
- [x] When resolving issues, they are referenced in the PR's title (e.g `fix: remove a typo, closes #___, #___`)
- [x] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md).
- [x] I have added a convincing reason for adding this feature, if necessary: We can't make Microsoft delete the arm64 windows version and ring takes ages for a new version so this is the next best thing.

### Other information

btw, building tauri apps _for_ arm still breaks as soon as people use _something_ that depends on ring (like everything that uses rustls) 🥳 

Closes #7015

PR #7106, opened 2023-06-01
feat(cli): Build the CLI for Windows ARM64, closes #7015 (#7098)

### LINKED ISSUE #7015: [feat] Provide @tauri-apps/cli-win32-arm64-msvc NPM package
### Describe the problem

Nowadays I can only manually compile Tauri everytime.

Provide @tauri-apps/cli-win32-arm64-msvc NPM package

### Alternatives considered

_No response_

### Additional context

_No response_

FabianLars:
Port of https://github.com/tauri-apps/tauri/pull/7098

Opening as a draft because i'm not sure how to best combine the feature flags right now...

### DIFF
--- a/.changes/cli-windows-arm64.md
+++ b/.changes/cli-windows-arm64.md
@@ -1,0 +1,6 @@
+---
+'tauri-cli': minor
+'@tauri-apps/cli': minor
+---
+
+Provide prebuilt CLIs for Windows ARM64 targets.
--- a/.github/workflows/publish-cli-js.yml
+++ b/.github/workflows/publish-cli-js.yml
@@ -40,6 +40,10 @@ jobs:
             build: yarn build:release --target i686-pc-windows-msvc
             target: i686-pc-windows-msvc
             architecture: x64
+          - host: windows-latest
+            architecture: x64
+            target: aarch64-pc-windows-msvc
+            build: yarn build:release --target aarch64-pc-windows-msvc --features native-tls-vendored --cargo-flags="--no-default-features"
           - host: ubuntu-20.04
             target: x86_64-unknown-linux-gnu
             docker: ghcr.io/napi-rs/napi-rs/nodejs-rust:lts-debian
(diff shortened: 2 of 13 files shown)


FabianLars:
whoops, i actually meant to close/update this PR since i opened it before https://github.com/tauri-apps/tauri/pull/7107 😅

FabianLars:
looks correct to me now. i let you decide if it makes sense to integrate #7017, which splits up the native-tls and native-tls-vendored flags. 
Though i still hope that ring updates before v2 will be stable (unlikely but one can hope) so we can simplify that again 😩

lucasfernog:
I think you linked the wrong PR. Anyway I just want to get this working right now, these feature flags are a mess.

FabianLars:
yeah, swapped the 1 and 0, the one in the first message was correct https://github.com/tauri-apps/tauri/pull/7107

lucasfernog:
Isn't `native-tls` automatically enabled when you enable `native-tls-vendored`? I think it should, that's what all libraries do.

FabianLars:
Nope, that's not what all libraries do, at least not ureq (and iirc atto didn't either), which is why i changed it in dev to behave the same.

FabianLars:
or maybe i'm dumb because ureq doesn't have a vendored feature, but i ended up having to always enable 2 feature flags to not end up with tls errors 🤔 

Edit: Feel free to double check that yourself, all these flags confuse the hell out of me especially when there are 2 or 3 http clients involved

lucasfernog:
We should work on removing attohttpc from dev anyway like you suggested, I think it would benefit the codebase overall.

lucasfernog:
How were you testing this btw @FabianLars ? Build targetting arm on windows?

FabianLars:
Yes, targeting arm to make sure it builds, and targeting x64 with the same flags to make sure i don't get any runtime errors (i have no arm machine to test)

lucasfernog:
I like it this way where you don't need to specify `native-tls,native-tls-vendored` and the -vendored implies the native-tls feature. I can't really compile for arm64 on my VM (stupid Ring?) but it should work the same as the dev stuff you pushed.

FabianLars:
Yeah I agree, i just meant to match what i thought was the default. Let's change it for dev too then :)

If ring still prevents compilation then this PR is kinda useless. Maybe there is another dep that needs disabled default features. I'll take a look at it on my Windows machine in the evening :v:

lucasfernog:
Ahh I see. Gonna see if I can find it.

lucasfernog:
We need some changes on tauri-mobile, I'll work on it.

lucasfernog:
@FabianLars now I think it's ready. I also changed the CI to test arm64 Windows builds since it is quite fragile. Let me know if you see any mistakes.

lucasfernog (approved):

FabianLars:
Yeah, looks good to me 👍

Issue #6762, opened 2023-04-20
[bug] NextJS app directory using router.push from next/navigation doesn't render page correctly.


Royceb:
### Describe the bug

I was uncertain whether to open this in Tauri or NextJS, so I opened issues in both platforms.

While utilizing a NextJS application, version 13, and the experimental app directory, I encountered an issue where using `router.push` to navigate to a new page fails to render correctly. Although the navigation occurs successfully, the page only displays code as text instead of rendering the expected content. I have not found any console logs or errors related to this issue.

This problem only arises after building and running the application. Locally, everything seems to be functioning correctly.
See video: 

https://user-images.githubusercontent.com/3469263/233442201-cdb9872f-cb3a-4bc1-a4bf-acb8ae18aea5.mov

Screenshot:
<img width="1506" alt="image" src="https://user-images.githubusercontent.com/3469263/233442935-7b08944e-5c82-4b78-be02-1f8884ae915e.png">

Simple repro: https://github.com/Royceb/nextjs-app-dir-tauri-router-push-example

### Reproduction

1. Ran `npx create-next-app@latest --experimental-app`
2. Set up Tauri

   2.1: Ran `pnpm add -D @tauri-apps/cli`

   2.2: Added `tauri` to package.json scripts. Ran `pnpm tauri init`.

3. Updated next.config.js to use `output: "export"` for SSG (Static Generation).
4. Made a simple page at `app/examplePage/page.tsx`
5. Imported `next/navigation` in `app/page.tsx` and added a button that calls `router.push('/examplePage')`.
6. Ran `pnpm tauri dev` and clicked the button. The app navigated to the example page as expected and rendered it properly.
7. Ran `pnpm tauri build` and opened the app. Clicked the button. The app did naviate to the example page, but it did not render it properly. The page rendered code as text instead of rendering the page.

### Expected behavior

The page should render correctly like it does in dev.

### Platform and versions

```text
› OS: Mac OS 13.0.1 X64
  › Node.js: 18.12.1
  › npm: 8.19.2
  › pnpm: 7.24.2
  › yarn: 1.22.19
  › rustup: 1.25.1
  › rustc: 1.67.0
  › cargo: 1.67.0
  › Rust toolchain: stable-aarch64-apple-darwin
```

Royceb:
Adding {forceOptimisticNavigation: true} to the router.push call fixed it.

https://github.com/vercel/next.js/issues/48642

brillout:
[...]

PR #7111, opened 2023-06-01
fix: correct MIME type of .txt files (ref: #6762)

reupen:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

This sets the MIME type for .txt files as `text/plain` instead of `text/html`. See #6762 for more details. (Whether `text/html` is always the right fallback is also a question, but I haven't done anything about that here.)

I also moved MP4 up to its correct place alphabetically in a few places.

### DIFF
--- a/.changes/mime-type-fallback.md
+++ b/.changes/mime-type-fallback.md
@@ -2,5 +2,4 @@
 "tauri-utils": patch
 ---
 
-Change fallback of `MimeType::parse_from_uri` and `MimeType::parse` to `MimeType::Txt`.
-
+Change fallback of `MimeType::parse` to `MimeType::Txt`.
--- a/.changes/mime-type-fallback.md
+++ b/.changes/mime-type-fallback.md
@@ -1,5 +1,0 @@
----
-"tauri-utils": patch
----
-
-Change fallback of `MimeType::parse` to `MimeType::Txt`.
--- a/.changes/mime-type-fallback.md
+++ b/.changes/mime-type-fallback.md
@@ -1,0 +1,6 @@
+---
+"tauri-utils": patch
+---
+
+Change fallback of `MimeType::parse_from_uri` and `MimeType::parse` to `MimeType::Txt`.
+
--- a/.changes/txt-mime-type.md
+++ b/.changes/txt-mime-type.md
@@ -1,0 +1,5 @@
+---
+"tauri-utils": patch
+---
+
+Correctly determine MIME type of `.txt` files.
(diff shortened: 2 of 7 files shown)


wusyong (dismissed):

FabianLars:
I wonder if we should use Self::Txt as a general fallback instead of Self::Html 🤔 Not sure what depends on that though so maybe something to think about later when Lucas is back.

reupen:
Pushed a fixup to fix the linter error. Unfortunately the 'test core' jobs are failing with:

```
error: package `cargo_toml v0.15.3` cannot be built because it requires rustc 1.64 or newer, while the currently active rustc version is 1.60.0
```

FabianLars:
i want to cry 😂

FabianLars:
Can you add the package to this list: https://github.com/tauri-apps/tauri/blob/dev/.github/workflows/test-core.yml#L95 pinning it to 0.15.2 please?

reupen:
Done. It was `cargo update -p cargo_toml:0.15.3 --precise 0.15.2` due to:

```
error: There are multiple `cargo_toml` packages in your project, and the specification `cargo_toml` is ambiguous.
Please re-run this command with `-p <spec>` where `<spec>` is one of the following:
  cargo_toml:0.11.8
  cargo_toml:0.15.3
```

reupen:
Okay, now there's that other Clippy failure.

I can reproduce the `tauri-codegen` warning on `dev` by running `cargo clippy --manifest-path core/tauri/Cargo.toml -- -D warnings`. Since it's set to fail on warnings, I assume that's the problem rather than the PR-from-fork stuff.

FabianLars:
Yeah, don't worry about it. Since your change doesn't trigger it we can fix it in another PR :)

lucasfernog:
For Tauri itself I think it's fine to change the default fallback to txt (I hope the asset will always have the .html extension), though for users it might be a little weird. The fallback for `tauri://localhost` should be indeed html instead of txt.

lucasfernog:
I'm thinking about it and it should be a separate PR (maybe even targetting v2 instead with a better way of handling this).

lucasfernog (approved):

FabianLars:
Why though? When would falling back to text be worse than falling back to html? Or different: When do we want to send html for non html files?

Genuine question btw, i'm so out of the loop on this topic by know that i can't think of a use-case right now.

lucasfernog:
`MimeType::parse_from_uri` is used when we can't determine the mime type from the file contents (which is always the case unless we do some custom html detection logic). When we receive a request to `tauri://localhost`, we want to serve it as HTML instead of plain text (I think in some contexts that is required).

FabianLars:
hmm, i thought we always tried to load index.html in these cases, or well, i guess we do but the other way around? idk, probably not much of a problem anyway now that we handle text files 🤷

Issue #6415, opened 2023-03-08
[bug] Resource response randomly lose some characters in Android dev


meowtec:
### Describe the bug

JS resource response randomly lose some characters.

```
03-10 12:02:38.404 13924 13924 E Tauri/Console: File: https://tauri.localhost/node_modules/.vite/deps/react-dom_client.js?v=0cc140e4 - Line 14117 - Msg: Uncaught SyntaxError: Unexpected identifier
```

![屏幕截图 2023-03-10 200553](https://user-images.githubusercontent.com/4006436/224311873-9e04ff67-b84b-43fd-858d-598b85271b23.png)

![屏幕截图 2023-03-08 231606](https://user-images.githubusercontent.com/4006436/223760234-09b394d1-19ac-407e-a3ae-d5c40472d8ed.png)

Code of this line should be `hostParentIsContainer = prevHostParentIsContainer;` but I got `hostParentIsContainer  prevHostParentIsContainer;`

And every time I reload the page, the location of losing character changed.

### Reproduction

[tauri-mobile-bug.zip](https://github.com/tauri-apps/tauri/files/10941768/tauri-mobile-bug.zip) (updated at Fri, 10 Mar 2023 12:00:37 GMT)

**NOTE** that the page may need reload once or more times. (E.g. using chrome://inspect/#devices to reload it)

### Platform and versions

```text
Environment
  › OS: Windows 10.0.22624 X64
  › Webview2: 110.0.1587.63
  › MSVC:
      - Visual Studio Community 2022
  › Node.js: 16.19.1
  › npm: 8.19.3
  › pnpm: 7.29.0
  › yarn: 1.22.19
  › rustup: 1.25.2
  › rustc: 1.67.1
  › cargo: 1.67.1
  › Rust toolchain: stable-x86_64-pc-windows-msvc

Packages
  › @tauri-apps/cli [NPM]: 2.0.0-alpha.2
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Error("unexpected character 't' while parsing major version number")', src\info.rs:566:67
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Additional context

Using Android simulator Pixel_XL_API_33 (x86_64)

It work well if redirect from `tauri.localhost` to the real host:

[...]

Issue #5344, opened 2022-10-04
Error after 30 seconds of fetch execution.


roiiiu:
### Describe the bug

I have a long executing request on my Go backend and once the fetch request takes more than 30 seconds, I get this error. With axios everything works fine.
<img width="987" alt="image" src="https://user-images.githubusercontent.com/11494827/193809166-f12eeb01-62b5-4557-9819-db4604164477.png">

```ts
await fetch('http://localhost:8080/g4', {
    method: 'POST',
  })
```
```go
r := gin.Default()
r.POST("/g4", func(c *gin.Context) {
    time.Sleep(40 * time.Second)
}
```

### Platform and versions

Environment
  › OS: Mac OS 12.6.0 X64
  › Node.js: 18.10.0
  › npm: 8.19.2
  › pnpm: 7.13.0
  › yarn: 1.22.19
  › rustup: Not installed!
  › rustc: 1.64.0
  › cargo: 1.64.0
  › Rust toolchain: 

Packages
  › @tauri-apps/cli [NPM]: 1.1.1
  › @tauri-apps/api [NPM]: 1.1.0
  › tauri [RUST]: 1.1.1,
  › tauri-build [RUST]: 1.1.1,
  › tao [RUST]: 0.14.0,
  › wry [RUST]: 0.21.1,

App
  › build-type: bundle
  › CSP: unset
  › distDir: ../dist
  › devPath: http://localhost:1420/
  › framework: Vue.js

App directory structure
  ├─ dist
  ├─ node_modules
  ├─ public
  ├─ src-tauri
  ├─ .git
  ├─ .vscode
  └─ src

FabianLars:
As far as i know `attohttpc` (the http client our fetch implementation uses by default) has a default timeout of 30 seconds. You could try changing the request timeout to a higher number or to 0 which _should_ disable it:
```ts
await fetch('http://localhost:8080/g4', {
    method: 'POST',
    timeout: 60
  })
```

roiiiu:
> 
I have tried but this doesn't work and the error report for timeout is not the same.
<img width="521" alt="image" src="https://user-images.githubusercontent.com/11494827/193813997-a8e21ad1-9a3c-497e-b676-fd5b1fde5d3f.png">

FabianLars:
[...]

PR #7143, opened 2023-06-06
refactor(core): Remove attohttpc in favor of reqwest

FabianLars:
ref: for example https://github.com/tauri-apps/tauri/issues/6415 and https://github.com/tauri-apps/tauri/issues/5344

- [x] Bugfix - yes i count this as a bugfix considering how many issues users had with atto
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No, well... Technically it is because of the enum variants but i think it's worth it.

### Checklist

### DIFF
--- a/.changes/remove-attohttpc.md
+++ b/.changes/remove-attohttpc.md
@@ -1,0 +1,6 @@
+---
+'tauri': patch
+'tauri-bundler': patch
+---
+
+Removed `attohttpc` in favor of `reqwest` (core) and `ureq` (tooling).
--- a/.changes/remove-attohttpc.md
+++ b/.changes/remove-attohttpc.md
@@ -1,6 +1,0 @@
----
-'tauri': patch
-'tauri-bundler': patch
----
-
-Removed `attohttpc` in favor of `reqwest` (core) and `ureq` (tooling).
--- a/core/tauri/src/api/error.rs
+++ b/core/tauri/src/api/error.rs
@@ -22,7 +22,7 @@ pub enum Error {
   #[error("user cancelled the dialog")]
   DialogCancelled,
   /// The network error.
-  #[cfg(all(feature = "http-api"))]
+  #[cfg(feature = "http-api")]
   #[error("Network Error: {0}")]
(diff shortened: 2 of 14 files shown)


FabianLars:
that clippy json output gives me a headache lol. Also, no idea why i couldn't trigger that warning locally 🤔

lucasfernog:
Usually that happens when I don't have latest cargo. Or not using the right feature flags.

core/tauri/src/api/error.rs
```diff
@@ -22,20 +22,13 @@ pub enum Error {
   #[error("user cancelled the dialog")]
   DialogCancelled,
   /// The network error.
-  #[cfg(all(feature = "http-api", not(feature = "reqwest-client")))]
-  #[error("Network Error: {0}")]
-  Network(#[from] attohttpc::Error),
-  /// The network error.
-  #[cfg(feature = "reqwest-client")]
-  #[cfg_attr(doc_cfg, doc(cfg(feature = "reqwest-client")))]
+  #[cfg(all(feature = "http-api"))]
```
lucasfernog:
```suggestion
  #[cfg(feature = "http-api")]
```

FabianLars:
ahhhhh, i was in the root dir and not in core/tauri ...

.changes/remove-attohttpc.md
```diff
@@ -0,0 +1,6 @@
+---
+'tauri': patch
+'tauri-bundler': patch
+---
+
+Removed `attohttpc` in favor of `reqwest` (core) and `ureq` (tooling).
```
lucasfernog: I prefer a separate change file for each package in this case.

lucasfernog (changes requested):

.changes/remove-attohttpc.md
FabianLars: right, makes sense! Still trying to get a feeling for changefiles

lucasfernog (approved):

PR #7157, opened 2023-06-07
fix(core): Send updater status events when default dialog is enabled, closes #7128

### LINKED ISSUE #7128: [bug] can't get the updater status
### Describe the bug

how to get the status of emit('tauri://update')? onUpdaterEvent seems not work.
I found that the dialog will show when there is a new version available, but nothing happens when not.

1. if there exists a newer version, dialog will show
2. if currentVersion.json not exist, nothing happens. In fact, `Promise.reject` expected
3. if cuurentVersion.json exists but version in it  not greater than currentVersion, nothing happens. In fact, `Promise.reject` expected

### Reproduction

_No response_

### Expected behavior

_No response_

### Platform and versions

```text
windows
tauri 1.3.0
```

### Stack trace

_No response_

### Additional context

_No response_

FabianLars:
I'm not sure if i catched everything, quite tricky to grasp the whole flow here tbh and like we said in the issue i didn't do much more than the bare minimum here.

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/core/tauri/src/updater/mod.rs
+++ b/core/tauri/src/updater/mod.rs
@@ -423,8 +423,10 @@ pub(crate) async fn check_update_with_dialog<R: Runtime>(handle: AppHandle<R>) {
       Ok(updater) => {
         let pubkey = updater_config.pubkey.clone();
 
-        // if dialog enabled only
-        if updater.should_update && updater_config.dialog {
+        if !updater.should_update {
+          send_status_update(&handle, UpdaterEvent::AlreadyUpToDate);
+        } else if updater_config.dialog {
+          // if dialog enabled only
           let body = updater.body.clone().unwrap_or_else(|| String::from(""));
           let dialog =
             prompt_for_install(&updater.clone(), &package_info.name, &body.clone(), pubkey).await;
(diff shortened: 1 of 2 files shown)


core/tauri/src/updater/mod.rs
```diff
@@ -423,8 +423,10 @@ pub(crate) async fn check_update_with_dialog<R: Runtime>(handle: AppHandle<R>) {
       Ok(updater) => {
         let pubkey = updater_config.pubkey.clone();
 
-        // if dialog enabled only
-        if updater.should_update && updater_config.dialog {
+        if !updater.should_update {
+          send_status_update(&handle, UpdaterEvent::AlreadyUpToDate);
+        } else if updater_config.dialog {
```
amrbashir, line 428:
```suggestion
        if !updater.should_update {
          send_status_update(&handle, UpdaterEvent::AlreadyUpToDate);
        }
        if updater_config.dialog {
```

amrbashir (changes requested):

core/tauri/src/updater/mod.rs
FabianLars: This would change the logic to always install the update even if it doesn't have to. If we don't want to have an if/else we can have 2 ifs, first one with `if !updater.should_update`, second one with `if updater.should_update && updater_config.dialog` (the old one) - this is what i first wanted to push.

Alternatively an early return if we don't wanna check for updater.should_update twice

lucasfernog: I don't see the issue with the logic in the PR. What is wrong with the if/else?

amrbashir: nvm my bad, I misinterpreted the logic

amrbashir (dismissed):

core/tauri/src/updater/mod.rs
```diff
@@ -591,22 +593,41 @@ Release Notes:
...
+        },
+        move || {
+          send_status_update(&handle_, UpdaterEvent::Downloaded);
+        },
+      )
+      .await;
+
+    if let Err(err) = &update_result {
+      // emit {"status": "ERROR", "error": "The error message"}
+      send_status_update(&update.app, UpdaterEvent::Error(err.to_string()));
```
lucasfernog: It's ok for `prompt_for_install` to return an error, the caller already checks the error and emits the event.

lucasfernog (changes requested):

lucasfernog (approved):

PR #7164, opened 2023-06-08
feat(android): disable action bar in `themes.xml`

### LINKED ISSUE #7052: [feat] sync webview theme with window theme on Android
### Describe the problem

I am building an android app using tauri, and there is a white screen on start, it looks like the following:
Screenshot_20230524_230429
but actually the webpage supports dark mode, after the webview is loaded, the background turns dark.

I would like to make the app theme match with the webpage, not only the background color, but also the action bar's color 
Screenshot_20230524_231230.

### Alternatives considered

_No response_

### Additional context

A pr with similar functionality, but it's a Windows implementation: https://github.com/tauri-apps/tauri/pull/5874

amrbashir:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/.changes/android-translucent-statusbar.md
+++ b/.changes/android-translucent-statusbar.md
@@ -1,6 +1,0 @@
----
-'tauri-cli': 'patch'
-'@tauri-apps/cli': 'patch'
----
-
-On Android, make Android status bar translucent. You can change this behaviour in `gen/android/app/src/main/res/values/themes.xml`
--- a/.changes/android-translucent-statusbar.md
+++ b/.changes/android-translucent-statusbar.md
@@ -1,0 +1,6 @@
+---
+'tauri-cli': 'patch'
+'@tauri-apps/cli': 'patch'
+---
+
+On Android, make Android status bar translucent. You can change this behaviour in `gen/android/app/src/main/res/values/themes.xml`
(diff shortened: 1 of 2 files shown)


wusyong (dismissed):

lucasfernog:
Is this really what most people will want? AFAIK flutter and capacitor doesn't do this by default.

amrbashir:
I don't really know, but I kinda hate the hard coded values.

amrbashir:
Btw I was going to make it fully transparent but backed out of it because it would look awful without also making the navigation bar also transparent and instead decided to just make it translucent which should work fine in most of cases and blend good enough with the user app by default.

Users can still remove this bit and customize it as they want for their app.

Also I have seen more and more Android apps moving towards transparent status bar (and navigation bar). I have a friend who works in Android native and he has noticed this trend as well.

wusyong:
I added previous stuff because that's the only sensible default that many android dev could agree on.
But that's indeed not the case in many scenarios. I think the best we should do is advise users to edit the theme they prefer.
Therese not such thing for default value and we can't control it.

amrbashir:
Alright, I just removed all the colors and kept the default. I only added the change to hide the action bar.

lucasfernog:
Transparent navigation bar looks amazing, but I think it might also impact the app more (last time I wrote a native Android app I took me ages to get it perfect).

amrbashir:
Yeah, I think it will either be a good community plugin or we could just link to a documentation or something and users implement it if they needed it

lucasfernog (approved):

PR #7165, opened 2023-06-08
feat(core): allow extending the shell scope, closes #5910

### LINKED ISSUE #5910: [feat] allow any shell command scope
### Describe the problem

It's really very lacking. I can't ask the user to specify the path to the executable, tauri just won't let me do it. Why can't I allow this in the same way as for `fs` for example.

Make similar to `fs`, something like `**`

### Alternatives considered

I hope I get these alternatives here, I can't bear to invent what's already there.

### Additional context

What are the reasons? Why doesn't it still work that way? What workarounds can I use?

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/core/tauri/src/scope/shell.rs
+++ b/core/tauri/src/scope/shell.rs
@@ -9,8 +9,27 @@ use crate::api::shell::Program;
 
 use regex::Regex;
 use tauri_utils::{config::Config, Env, PackageInfo};
+use uuid::Uuid;
 
-use std::collections::HashMap;
+use std::{
+  collections::HashMap,
+  path::{Path, PathBuf},
+  sync::{Arc, Mutex},
+};
+
+/// Scope change event.
+#[derive(Debug, Clone)]
+pub enum Event {
+  /// A command has been allowed.
+  CommandAllowed {
+    /// The command key.
+    name: String,
+    /// The command scope definition.
+    scope: ScopeAllowedCommand,
+  },
+}
+
+type EventListener = Box<dyn Fn(&Event) + Send>;
 
 /// Allowed representation of `Execute` command arguments.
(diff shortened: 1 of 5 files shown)


core/tauri/src/scope/shell.rs
```diff
@@ -71,13 +90,35 @@ pub struct ScopeConfig {
...
   /// The shell command to be called.
-  pub command: std::path::PathBuf,
+  command: PathBuf,
 
   /// The arguments the command is allowed to be called with.
-  pub args: Option<Vec<ScopeAllowedArg>>,
+  args: Option<Vec<ScopeAllowedArg>>,
 
   /// If this command is a sidecar command.
-  pub sidecar: bool,
```
amrbashir, line 80: this is a breaking change, any reason to hide these?

lucasfernog: This struct was only used on codegen, no public API returned or takes it as input. I've hidden this so no one use it instead of the builder, which is the stable API.

amrbashir: It still shows in documentation https://docs.rs/tauri/latest/tauri/scope/struct.ShellScopeAllowedCommand.html so even if no functions take it as input, we can't break it

amrbashir: And if it is not meant to be used by users anyway, it won't hurt to keep the fields public anyways and maybe add a TODO to hide the type from docs and make the fields private

FabianLars:
I assume this won't be merged into v1 anymore, right? Should we close it then and queue it for v2 migration?

PR #4752, opened 2022-07-24
feat(core): expose test utils

lucasfernog:
### What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

### Does this PR introduce a breaking change?

- [ ] Yes, and the changes were approved in issue #___
- [ ] No

### Checklist
- [ ] When resolving issues, they are referenced in the PR's title (e.g `fix: remove a typo, closes #___, #___`)
- [x] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md).
- [ ] I have added a convincing reason for adding this feature, if necessary

PR #7170, opened 2023-06-08
refactor(core): use webview's URI schemes for IPC

### LINKED ISSUE #4875: [bug] naming an argument `cmd` causes unexpected and undocumented error
### Describe the bug

I named an argument for a custom command `cmd`, which used to work previously but now it seems to be a special name that gets treated differently by tauri. In fact it seems to override the command specified as the first parameter to `invoke()`.

```rust
#[tauri::command]
pub fn my_command(
  cmd: String
) -> Result<String, String> {
    info!("This never gets logged");
    Ok(String::new())
}
```
I invoked it from JS like this:
```
window.__TAURI__.invoke("my_command", {cmd: "test"})
```
and instead of calling my_command and passing that argument, it returns the error message `command "test" not found` in the frontend and the backend command was never executed.

This seems really strange to me. Why would certain argument names be considered special?
Is this documented anywhere?

Given that there are such special argument names, could these please be documented here? or at least somewhere prominent?
https://tauri.app/v1/guides/features/command

Even better, I'd argue that any "special" args internal to tauri should be an optional 3rd parameter to invoke() rather than overloading the user-supplied args list.

### Reproduction

```rust
#[tauri::command]
pub fn my_command(
  cmd: String
) -> Result<String, String> {
    info!("This never gets logged");
    Ok(String::new())
}
```

Now try to invoke the above command. This doesn't work:
```
window.__TAURI__.invoke("my_command", {cmd: "test"})
```
Instead it gives the error `Uncaught command test not found`

### Expected behavior

I'd expect all of the arguments given in the 2nd parameter to `invoke()` to be delivered to my function regardless, and not to have any "magic" args that are treated differently. Any such special internal args should be in a separate parameter given to `invoke()`

### Platform and versions

Environment
  › OS: Windows 10.0.22000 X64
  › Webview2: 103.0.1264.77
  › MSVC:
  › Node.js: 16.13.0
  › npm: 8.1.0
  › pnpm: Not installed!
  › yarn: Not installed!
  › rustup: 1.25.1
  › rustc: 1.62.0
  › cargo: 1.62.0

### Stack trace

_No response_

### Additional context

_No response_

lucasfernog:
closes #4875

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


lucasfernog:
A command returning a 150MB file now takes less than 60ms to resolve. Previously: almost 50 seconds.

amrbashir:
image

lucasfernog:
Needs https://github.com/tauri-apps/wry/pull/970

lucasfernog:
Also needs #4752 to fix the unit tests.

lucasfernog:
We'll detect and use the custom protocol IPC on Linux if available when https://github.com/tauri-apps/wry/pull/969 lands. On older webkit2gtk, we'll need to stick with the current approach (JSON :( )

lucasfernog:
With the latest changes returning a 150MB file takes 2s on linux using the current IPC (the new one based on custom protocols will require a feature flag).

lucasfernog:
With the latest changes returning a 150MB file takes 1.3s on Linux with both postMessage and custom protocols (the latter will require a feature flag because it requires webkit2gtk 2.40).

The existing postmessage responder was changed to check if the command return value is either an object, an array or a raw `tauri::ipc::Response`. In these cases, it'll use the `tauri::ipc::Channel` to send the response. The Channel evaluates a JS script that runs the custom protocol IPC to fetch the response stored in a Tauri state.

Example command:

```rust
use tauri::{command, ipc::Response};

#[command]
pub fn read_file() -> Response {
  Response::new(include_bytes!("../../../../test_video.mp4").to_vec())
}
```

lucasfernog:
Updates on the 150mb command response benchmark:
macOS: 70-80ms
Linux: 1.3-1.5s (ipc.postMessage), 1.3-1.4s (custom protocol)
Windows: 1.8s

kinda scary how much worse webview2 is... or maybe WKWebView is just too fast?

FabianLars:
I wonder if webview2's custom uri scheme (which only works on win10+ :/ ) is more optimized for this :thinking: Not that it would matter for us until we drop support for 7&8 :sweat_smile: 
Edit: Nevermind, it works differently than on macos/linux and it will keep using the same webresourcerequested event we already use.

Also, the measurement includes reading the file from disk right? Maybe windows is just really bad at that part too

lucasfernog:
The file is included in the app with include_bytes!

Android performance is in the 700-800ms range.
iOS 70-100ms.

lucasfernog:
This PR is now ready, though we need a new wry release to publish.

wusyong:
wry 0.29 is published! You could update the dep to it.

core/tauri/src/app.rs
```diff
@@ -820,9 +850,20 @@ impl<R: Runtime> Builder<R> {
...
+      invoke_responder: None,
+      invoke_initialization_script: InvokeInitializationScript {
+        process_ipc_message_fn: crate::manager::PROCESS_IPC_MESSAGE_FN,
+        #[cfg(ipc_custom_protocol)]
+        use_post_message_condition: "false",
+        #[cfg(not(ipc_custom_protocol))]
+        use_post_message_condition: &format!(
+          "(navigator.appVersion.includes('Linux') || navigator.appVersion.includes('Android')) && !cmd.startsWith('{}')",
+          crate::ipc::FETCH_CHANNEL_DATA_COMMAND_PREFIX
+        ),
```
amrbashir: could we change this to detect platform on rust side? `navigator.appVersion` is not reliable anymore, it is deprecated and I think if users change the useragent string it will affect `appVersion` and the detection will fail.

amrbashir (changes requested):

core/tauri/src/app.rs
lucasfernog: AFAIK the userAgent does not affect navigator.appVersion (at least last time I checked) but I'll make this change.

amrbashir: yeah even if `appVersion` is not affected, it is better to calculate once on rust at compile time

amrbashir (dismissed):

JonasKruckenberg:
So my question right out of the gate is: Why is this not implemented in wry directly?

lucasfernog:
> So my question right out of the gate is: Why is this not implemented in wry directly?

we actually need both the current IPC (postMessage) and custom protocol IPC.. wry already gives us the tools we need to so need to interfere there. Since wry is low level I don't think we should get this mess over there like handling Linux/Android/webkit2gtk-2.40 stuff.

amrbashir (dismissed):

core/tauri/src/ipc/protocol.rs
```diff
@@ -0,0 +1,272 @@
...
+        response.set_mimetype(Some(content_type.into()));
+
+        response
+      }
+
+      Method::OPTIONS => {
+        let mut r = HttpResponse::new(Vec::new().into());
+        r.headers_mut().insert(
+          ACCESS_CONTROL_ALLOW_HEADERS,
+          HeaderValue::from_static("Content-Type, Tauri-Callback, Tauri-Error, Tauri-Channel-Id"),
```
lucasfernog, line 64: @tweidinger @chippers can I set the allow headers to "*" here? The headers can be accessed like this:

```rust
#[command]
fn my_cmd(request: tauri::ipc::Request) {
  let headers = request.headers();
}
```

core/tauri/scripts/core.js
```diff
@@ -13,6 +13,13 @@
     })
   }
 
+  window.__TAURI__.convertFileSrc = function convertFileSrc(filePath, protocol = 'asset') {
+    const path = encodeURIComponent(filePath)
+    return navigator.userAgent.includes('Windows') || navigator.userAgent.includes('Android')
+      ? `https://${protocol}.localhost/${path}`
+      : `${protocol}://localhost/${path}`
```
chippers: Using `Window::with_webview` could make this logic fail if they edit the webview settings on Windows to not include the string Windows, and for macOS/Linux if they add "Windows" or "Android"

tooling/api/src/tauri.ts
```diff
@@ -9,17 +9,6 @@
...
-declare global {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  interface Window {
-    __TAURI_IPC__: (message: any) => void
-    ipc: {
-      postMessage: (args: string) => void
-    }
-  }
-}
-
```
FabianLars, line 22: without this the build command fails for me. idk why though because i do something similar (only declaring it in index) in my own projects and there it works fine 🤔

lucasfernog: build failed on your own project?

FabianLars: trying to build&pack the api package itself.

FabianLars: Also, i may be just lost in so many changes here, but using event.listen from it doesn't work for me either, failing with:
```
Uncaught (in promise) invalid args `event` for command `listen`: command listen missing required key event
```
using `window.__TAURI_INVOKE__` directly with the same args `listen` injects it works though so i think i am doing something wrong.

FabianLars: and lastly, i get stuff like this on Windows:
grafik

lucasfernog: Weird that the build works for me and CI.

I'll check the Windows stuff, seems like the request body isn't being recognized.

lucasfernog: Ahh ok I get the same build error on Windows :D

lucasfernog: Should be working now @FabianLars

FabianLars: yep, builds fine now and ipc on windows works too. Only the `listen` error still happens. The error on mobile: https://github.com/tauri-apps/tauri/pull/7170#discussion_r1260023780 and on desktop it's still the same one as in the first half of the screenshot weirdly enough.

lucasfernog: Can you share the code?

FabianLars:
```ts
import { listen } from "@tauri-apps/api/event";
listen("deep-link://new-url", console.log);
```
repro steps:
- cd tooling/api
- yarn build
- yarn pack
- copy that tgz to wherever needed
- yarn install that tgz in that test project
- add code above, see error in console
- change code to this (copy pasted from event.ts changing invoke to the window api) and it works

```ts
window.__TAURI_INVOKE__<number>("plugin:event|listen", {
    event: "deep-link://new-url",
    windowLabel: "main",
    handler: transformCallback(handler), // handler is just a wrapper around console.log here
  });
```

lucasfernog: Works on my end anyway. Something we could try is changing the invoke API to use __TAURI_INVOKE__

FabianLars: Yeah, works for me in "normal" projects too, just not in npm workspaces because apparently the tgz gets deduplicated&replaced with the released npm package from the workspace.....

lucasfernog: ok so not our fault :D

core/tauri-utils/src/pattern/isolation.js
```diff
@@ -85,8 +103,14 @@
...
-    const encrypted = await encrypt(data)
-    sendMessage(encrypted)
+    const { cmd, callback, error, payload, options } = data
+    sendMessage({
+      cmd,
+      callback,
+      error,
+      payload: await encrypt(payload),
+      options
+    })
```
chippers: This should probably be written with `Object.create(null)` like all other objects are created with in the ipc and isolation scripts.
```suggestion
const message = Object.create(null)
message.cmd = data.cmd
message.callback = data.callback
message.error = data.error
message.options = data.options
message.payload = await encrypt(payload)
sendMessage(message)
```

core/tauri/src/ipc/protocol.rs
tillmann-crabnebula: IMHO if there are no secrets meant to be sent via headers it is totally fine as default.
In the default case there are only the `Content-Type`, `Tauri-Callback`, `Tauri-Error`, `Tauri-Channel-Id` header set. These are not containing secret values from my perspective. Leaking the `Tauri-Channel-Id` could be abused but I can not really think of a scenario where it could be leaked without already having more privileges.

If there are going to be secrets in the headers or cookies attached to the request the developer would need to modify the `ACCESS_CONTROL_ALLOW_HEADERS` to suit their usage as the `*` value will be treated as a header name instead of a wildcard.

Can the developer modify this value at runtime or via configuration?
It doesn't seem supported on webkit https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers#browser_compatibility

core/tauri/scripts/ipc-protocol.js
```diff
@@ -0,0 +1,52 @@
+// Copyright 2019-2023 Tauri Programme within The Commons Conservancy
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: MIT
+
+(function () {
```
chippers: probably want to open with a semicolon here
```suggestion
;(function () {
```

core/tauri/scripts/process-ipc-message-fn.js
```diff
@@ -0,0 +1,29 @@
+// Copyright 2019-2023 Tauri Programme within The Commons Conservancy
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: MIT
+
+(function (message) {
```
chippers, line 7: probably want to open with a semicolon here
```suggestion
;(function (message) {
```

core/tauri/src/ipc/protocol.rs
lucasfernog: we can add an API in Rust for this @tillmann-crabnebula though I'd wait for it to be requested actually.

core/tauri/scripts/ipc-protocol.js
```diff
@@ -0,0 +1,52 @@
...
+  const osName = __TEMPLATE_os_name__
+  const fetchChannelDataCommand = __TEMPLATE_fetch_channel_data_command__
+  const useCustomProtocol = __TEMPLATE_use_custom_protocol__
+
+  Object.defineProperty(window, '__TAURI_POST_MESSAGE__', {
+    value: (message) => {
+      const { cmd, callback, error, payload, options } = message
+
+      // use custom protocol for IPC if the flag is set to true, the command is the fetch data command or when not on Linux/Android
+      if (useCustomProtocol || cmd == fetchChannelDataCommand || (osName !== 'linux' && osName !== 'android')) {
```
chippers: The second comparison should probably use strict equality operator.
```suggestion
      if (useCustomProtocol || cmd === fetchChannelDataCommand || (osName !== 'linux' && osName !== 'android')) {
```

tooling/api/src/tauri.ts
```diff
@@ -156,15 +173,16 @@ async function invoke<T>(cmd: string, args: InvokeArgs = {}): Promise<T> {
...
+      options
     })
   })
 }
 
 /**
  * Convert a device file path to an URL that can be loaded by the webview.
  * Note that `asset:` and `https://asset.localhost` must be added to [`tauri.security.csp`](https://tauri.app/v1/api/config/#securityconfig.csp) in `tauri.conf.json`.
- * Example CSP value: `"csp": "default-src 'self'; img-src 'self' asset: https://asset.localhost"` to use the asset protocol on image sources.
+ * Example CSP value: `"csp": "default-src 'self' ipc:; img-src 'self' asset: https://asset.localhost"` to use the asset protocol on image sources.
```
chippers: This should probably match the migration tool, which is also how assets is done.
```suggestion
 * Example CSP value: `"csp": "default-src 'self' ipc: https://ipc.localhost; img-src 'self' asset: https://asset.localhost"` to use the asset protocol on image sources.
```

examples/streaming/tauri.conf.json
```diff
@@ -47,7 +47,7 @@
       }
     ],
     "security": {
-      "csp": "default-src 'self'; media-src stream: https://stream.localhost asset: https://asset.localhost",
+      "csp": "default-src 'self' ipc:; media-src stream: https://stream.localhost asset: https://asset.localhost",
```
chippers, line 50: does this not also need the windows version?
```suggestion
      "csp": "default-src 'self' ipc: https://ipc.localhost; media-src stream: https://stream.localhost asset: https://asset.localhost",
```

core/tauri/src/lib.rs
```diff
@@ -120,6 +120,13 @@ macro_rules! android_binding {
...
+    ::tauri::wry::application::android_fn!(
+      app_tauri,
+      plugin,
+      PluginManager,
+      sendChannelData,
+      [i64, JString],
+    );
 
     #[allow(non_snake_case)]
     pub unsafe fn handlePluginResponse(
```
chippers: Even if this is for private internal use, this should still have a safety comment so that developers are aware what things need to be taken care of in order to stay safe.

core/tauri/src/lib.rs
```diff
@@ -131,12 +138,17 @@ macro_rules! android_binding {
     ) {
       ::tauri::handle_android_plugin_response(env, id, success, error);
     }
+
+    #[allow(non_snake_case)]
+    pub unsafe fn sendChannelData(env: JNIEnv, _: JClass, id: i64, data: JString) {
```
chippers: Even if this is for private internal use, this should still have a safety comment so that developers are aware what things need to be taken care of in order to stay safe.

core/tauri/src/ipc/channel.rs
```diff
@@ -0,0 +1,149 @@
...
+
+use super::{CallbackFn, InvokeBody, InvokeError, IpcResponse, Request, Response};
+
+pub const IPC_PAYLOAD_PREFIX: &str = "__CHANNEL__:";
+pub const CHANNEL_PLUGIN_NAME: &str = "__TAURI_CHANNEL__";
+// TODO: ideally this const references CHANNEL_PLUGIN_NAME
+pub const FETCH_CHANNEL_DATA_COMMAND: &str = "plugin:__TAURI_CHANNEL__|fetch";
+
+#[derive(Default, Clone)]
+pub struct ChannelDataCache(pub(crate) Arc<Mutex<HashMap<u32, InvokeBody>>>);
```
tillmann-crabnebula: This is rather a `ChannelDataQueue` or `ChannelDataUnsortedQueue` than `ChannelDataCache` as entries are not cached and taken out of the hashmap as soon as they are used.

core/tauri/src/ipc/protocol.rs
```diff
@@ -0,0 +1,272 @@
...
+    #[cfg(all(feature = "isolation", ipc_custom_protocol))]
+    if let crate::Pattern::Isolation { crypto_keys, .. } = manager.pattern() {
+      match crate::utils::pattern::isolation::RawIsolationPayload::try_from(&body)
+        .and_then(|raw| crypto_keys.decrypt(raw))
+      {
+        Ok(data) => body = data,
+        Err(e) => {
+          return Err(e.to_string());
+        }
+      }
```
chippers: not a security concern, but this could be written a little cleaner IMO
```suggestion
      body = crate::utils::pattern::isolation::RawIsolationPayload::try_from(&body)
        .and_then(|raw| crypto_keys.decrypt(raw))
        .map_err(|e| e.to_string())?;
```

core/tauri/src/ipc/protocol.rs
```diff
@@ -0,0 +1,272 @@
...
+    }
+  }
+}
+
+fn handle_ipc_request<R: Runtime>(
+  request: &HttpRequest,
+  manager: &WindowManager<R>,
+  label: &str,
+) -> std::result::Result<InvokeResponse, String> {
```
chippers, line 193: is there a reason we are using strings as errors instead of known enums? is this something we are planning to keep or refactor before v2

core/tauri/src/ipc/channel.rs
tillmann-crabnebula: Arguably it still fits in the definition of a cache, but for me it was confusing.

core/tauri/src/ipc/protocol.rs
```diff
@@ -0,0 +1,272 @@
...
+        .map_err(|_| "Tauri-Error header value must be a numeric string")?,
+    );
+
+    let content_type = request
+      .headers()
+      .get(reqwest::header::CONTENT_TYPE)
+      .and_then(|h| h.to_str().ok())
+      .unwrap_or("application/octet-stream");
+    let body = match content_type {
+      "application/octet-stream" => body.into(),
```
chippers: I think it would be nice that everywhere we use standardized mime types, we use the constants available in `mime` - which is already in our dependency tree due to the non-optional `reqwest`.

chippers (changes requested):
We should think about using constants for the `Tauri-XYZ` headers. They are typically only used directly in 3 places, but it never hurts to have un-typo-able constants.

core/tauri/src/ipc/protocol.rs
lucasfernog: this is just the first layer of validation on the IPC request handler, it shouldn't fail in most cases (mostly if you request without using the `__TAURI__.invoke` helper). There's no need to have an enum here IMO.

core/tauri-utils/src/pattern/isolation.js
lucasfernog: nice catch thanks

core/tauri/src/ipc/channel.rs
```diff
@@ -0,0 +1,150 @@
...
+}
+
+#[command(root = "crate")]
+fn fetch(
+  request: Request<'_>,
+  cache: State<'_, ChannelDataIpcQueue>,
+) -> Result<Response, &'static str> {
+  if let Some(id) = request
+    .headers()
+    .get("Tauri-Channel-Id")
```
chippers: This header was missed for the constant refactoring

core/tauri/src/ipc/protocol.rs
```diff
@@ -0,0 +1,275 @@
...
+    );
+
+    let content_type = request
+      .headers()
+      .get(reqwest::header::CONTENT_TYPE)
+      .and_then(|h| h.to_str().ok())
+      .map(|mime| mime.parse())
+      .unwrap_or(Ok(mime::APPLICATION_OCTET_STREAM))
+      .map_err(|_| "unknown content type")?;
+    let body = if content_type == mime::APPLICATION_OCTET_STREAM {
```
chippers, line 247: I think we can turn this back into a match statement. I'll submit a change

core/tauri/src/ipc/protocol.rs
```diff
@@ -0,0 +1,275 @@
...
+
+pub fn get<R: Runtime>(manager: WindowManager<R>, label: String) -> UriSchemeProtocolHandler {
+  Box::new(move |request| {
+    let mut response = match *request.method() {
+      Method::POST => {
+        let (mut response, content_type) = match handle_ipc_request(request, &manager, &label) {
+          Ok(data) => match data {
+            InvokeResponse::Ok(InvokeBody::Json(v)) => (
+              HttpResponse::new(serde_json::to_vec(&v)?.into()),
+              "application/json",
```
chippers: Some mime types up here did not get changed. I can submit the changes

chippers (changes requested):

core/tauri/src/ipc/protocol.rs
chippers: Since the value is manually implemented instead of derived, we cannot use the constant as the pattern

tillmann-crabnebula (approved):

chippers (approved):

ghost:
Having lurked in this thread since the PR was first made I'm quite delighted to see this merged! I have been using the alpha releases for a while now and wonder how I would go about using this new feature together with the new v2 api plugins? 

Naively switching over the tauri and tauri-build dependencies to use the dev branch of course creates version conflicts with the fs plugin that I use. From where should I pull in the plugins?

amrbashir:
> Naively switching over the tauri and tauri-build dependencies to use the dev branch of course creates version conflicts with the fs plugin that I use. From where should I pull in the plugins?

The v2 branch of plugins-workspace repo https://github.com/tauri-apps/plugins-workspace/tree/v2 or from the published crates on crates.io
but probably you should wait for the next alpha release (soon to be released).

kris-ava:
On MacOS this breaks IPC using Remote Domains. Safari blocks the requests to localhost from other domains.

Xiaobaishushu25:
I want to know how to use this new IPC？i update to v2,but the file transder between js and rust is also slow in Windows OS, Here is my `Cargo. toml`：
```toml
[build-dependencies]
tauri-build = { version = "2.0.0-beta", features = [] }

[dependencies]
tauri = { version = "2.0.0-beta.13", features = ["custom-protocol"] }
serde = { version = "1", features = ["derive"] }
serde_json = "1"
chrono = "0.4.31"
tokio = { version = "1.20", features = ["macros", "rt-multi-thread"] }
tauri-plugin-dialog = "2.0.0-beta.3"
tauri-plugin-http = "2.0.0-beta.3"
tauri-plugin-fs = "2.0.0-beta.3"


[features]
# This feature is used for production builds or when a dev server is not specified, DO NOT REMOVE!!
custom-protocol = ["tauri/custom-protocol"]
```
this is my rust code:
```rust
#[tauri::command]
async fn append_chunk_to_file(
    window: Window,
    path: String,
    chunk: Vec<u8>,
    end: bool,
) -> Result<(), String> {
    let current_time = Local::now().time();
    println!("enter rust time: {}", current_time);
    println!("start{:?}", Instant::now()); //收到函数时间Instant { t: 644913.1384745s }
    tokio::spawn(async move {
        let mut file = OpenOptions::new()
            .create(true)
            .append(true)
            .open(&path)
            .map_err(|e| e.to_string())
            .unwrap();
        file.write_all(&chunk).map_err(|e| e.to_string()).unwrap();
        if end {
            window.emit("insert", Payload { message: path }).unwrap();
        }
    });
    let instant = Instant::now();
    println!("end{:?}", instant); //结束函数时间Instant { t: 644913.1386845s }
    let current_time = Local::now().time();
    println!("return time: {}", current_time);
    Ok(())
}
```
i use it in js:
```
import {mkdir, readFile, writeFile, BaseDirectory} from "@tauri-apps/plugin-fs";
import {convertFileSrc, invoke} from "@tauri-apps/api/core";
const content = await file.arrayBuffer()
let content1 = new Uint8Array(content);
await invoke("append_chunk_to_file", {path: url, chunk: chunk1,end:true})
```
It took 8 seconds to transfer a 23MB image using the above code(most time take in se and de),but @lucasfernog  say A command returning a 150MB file now takes less than 60ms to resolve. Previously: almost 50 seconds. 
Where did I go wrong? Do you have the correct example code?

PR #7189, opened 2023-06-13
fix(cli): warn about bundling updater target without appropriate targets, closes #7181

### LINKED ISSUE #7181: [bug] tauri build not create updater files
### Describe the bug

I followed the updater instruction in - https://tauri.app/v1/guides/distribution/updater/ 
in tauri.conf.json changed bundle.targets to "updater" 
and when I run tauri build got this:

   **Compiling updaterwork v0.0.0 (C:\Users\rivka.shaked\Downloads\UPDATERWORK\src-tauri)
    Finished release [optimized] target(s) in 20.56s
    Finished 1 bundle at:

        Info 0 updater archives at:**

no file created in \target\release\bundle

### Reproduction

[UPDATERWORK.zip](https://github.com/tauri-apps/tauri/files/11721499/UPDATERWORK.zip)

### Expected behavior

Updater package will be created with zip and sig file

### Platform and versions

```text
[✔] Environment
    - OS: Windows 10.0.19045 X64
    ✔ WebView2: 114.0.1823.43
    ✔ rustc: 1.67.1 (d5a82bbd2 2023-02-07)
    ✔ Cargo: 1.67.1 (8ecd4f20a 2023-01-10)
    ✔ rustup: 1.25.2 (17db695f1 2023-02-01)
    - node: 18.14.0
    - npm: 9.3.1

[-] Packages
    - tauri [RUST]: 1.3.0
    - tauri-build [RUST]: 1.3.0
    - wry [RUST]: 0.24.3
    - tao [RUST]: 0.16.2
    - @tauri-apps/cli [NPM]: 1.3.1

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../src
    - devPath: ../src
```

### Stack trace

_No response_

### Additional context

_No response_

amrbashir:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/tooling/bundler/src/bundle.rs
+++ b/tooling/bundler/src/bundle.rs
@@ -48,21 +48,6 @@ pub fn bundle_project(settings: Settings) -> crate::Result<Vec<Bundle>> {
     return Ok(Vec::new());
   }
 
-  if package_types.contains(&PackageType::Updater)
-    && !package_types.iter().any(|p| {
-      matches!(
-        p,
-        PackageType::AppImage
-          | PackageType::MacOsBundle
-          | PackageType::Nsis
-          | PackageType::WindowsMsi
-      )
-    })
-  {
-    warn!("The updater bundle target exists but couldn't find any updater-enabled target, so the updater artifacts won't be generated. Please add one of these targets as well: app, appimage, msi, nsis");
-    return Ok(Vec::new());
-  }
-
   let mut bundles: Vec<Bundle> = Vec::new();
 
   let target_os = settings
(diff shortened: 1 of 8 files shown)


lucasfernog:
On Linux and macOS we're actually triggering the AppImage/app bundling. We just need to do the same on Windows.
https://github.com/tauri-apps/tauri/blob/696d77c3ce480f4a3b4c1c57dae64ff2bb7842ce/tooling/bundler/src/bundle/updater_bundle.rs#L121
https://github.com/tauri-apps/tauri/blob/696d77c3ce480f4a3b4c1c57dae64ff2bb7842ce/tooling/bundler/src/bundle/updater_bundle.rs#L78

amrbashir:
I knew I saw this code somewhere lol.

Anyways, while this works fine for Linux and macOS, it is not the same case for Windows since there is **two** updater targets and not just **one**, so should we build both targets? doesn't sound intuitive to me, instead we should be disabling building `updater` target without another updater-enabled target, what do you think?

lucasfernog:
What if we assume NSIS in this case? I think that should be our focus going forward, and maybe we can even drop WiX on v2 (probably move it out of the monorepo so custom users have time to migrate).

amrbashir:
We can assume NSIS as default for now but it is not future-proof, if we decided to add another bundler target like `Squirrel` or maybe `MSIX`

lucasfernog:
Hmm ok. Safer to go with this idea then.

lucasfernog (approved):

tooling/bundler/src/bundle.rs
```diff
@@ -93,9 +97,23 @@ pub fn bundle_project(settings: Settings) -> crate::Result<Vec<Bundle>> {
...
+            PackageType::AppImage
+              | PackageType::MacOsBundle
+              | PackageType::Nsis
+              | PackageType::WindowsMsi
+          )
+        }) {
+          warn!("The updater bundle target exists but couldn't find any updater-enabled target, so the updater artifacts won't be generated. Please add one of these targets as well: app, appimage, msi, nsis");
+          continue;
+        }
+        updater_bundle::bundle_project(&settings, &bundles)?
```
AlexTMjugador, line 113: I have just found out that `package_types` is never updated when a macOS DMG bundle is generated (see lines 77-85). Therefore, the updater bundle will not be generated if the macOS .app bundle is not explicitly set in `package_types`, even though the DMG bundle implies the generation of an app bundle and thus `package_types` does not need to explicitly contain a macOS .app bundle.

I'd suggest adding a `| PackageType::Dmg` branch to the `matches!` macro here to better handle this edge case.

PR #7214, opened 2023-06-15
chore: merge `next` into `dev`

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


lucasfernog:
This PR will be merged with a merge commit to preserve all `next` commits with authors.

.github/workflows/covector-version-or-publish-next.yml
```diff
@@ -0,0 +1,88 @@
+# Copyright 2019-2023 Tauri Programme within The Commons Conservancy
```
amrbashir, line 1: shouldn't this file be removed now and instead port the changes to `covector-version-or-publish.yml`?

core/tauri/src/asset_protocol.rs
```diff
@@ -2,9 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-License-Identifier: MIT
 
-#![cfg(protocol_asset)]
-
-use crate::api::file::SafePathBuf;
+use crate::path::SafePathBuf;
```
amrbashir, line 5:
```suggestion
#![cfg(feature = "protocol-asset")]
use crate::path::SafePathBuf;
```

tooling/cli/src/build.rs
```diff
@@ -238,13 +143,6 @@ pub fn command(mut options: Options, verbosity: u8) -> Result<()> {
...
     #[cfg(target_os = "linux")]
     {
-      use crate::helpers::config::ShellAllowlistOpen;
-      if matches!(
-        config_.tauri.allowlist.shell.open,
-        ShellAllowlistOpen::Flag(true) | ShellAllowlistOpen::Validate(_)
-      ) {
-        std::env::set_var("APPIMAGE_BUNDLE_XDG_OPEN", "1");
-      }
       if config_.tauri.system_tray.is_some() {
```
amrbashir, line 146: Did we move this to the shell plugin?

amrbashir (changes requested):

tooling/cli/src/build.rs
lucasfernog: not yet, we still need to figure out a safe way to handle this since the env var is set on the CLI.

.github/workflows/covector-version-or-publish-next.yml
lucasfernog: Yeah, I wasn't sure if I should do this right now or after merging the PR.

core/tauri/src/asset_protocol.rs
lucasfernog: Gotta be a PR instead since we can't push directly to next 😢

.github/workflows/covector-version-or-publish-next.yml
amrbashir: yeah lets do this modifications afterwards

tooling/cli/src/build.rs
amrbashir: how about adding it in the plugin build script?

amrbashir (approved):

tooling/cli/src/build.rs
lucasfernog: Yeah but to pass a security review it would have to be specific to that env var I think

core/tauri/src/asset_protocol.rs
lucasfernog: Actually probably don't even need this one since the feature flag is used on the module declaration

tooling/cli/src/build.rs
amrbashir: well, the build script could read the `TAURI_CONFIG` env var, validate the scope using this same logic and then set the env var if appropriate

lucasfernog: the build script is running on its own process though, so it actually needs a way to export the env var to tauri instead. something like what cargo does with the cargo:something printed to stdout.

core/tauri/src/asset_protocol.rs
amrbashir: yeah, this is just mainly to ensure that if the module declaration flag ever get deleted by accident, it will fail to build.

tooling/cli/src/build.rs
amrbashir: oh TIL.

Alright, I will keep this on my TODO list so I don't forget so we can discuss it later.

PR #7235, opened 2023-06-17
feat: add navigate method

jhutchins:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/core/tauri/src/window.rs
+++ b/core/tauri/src/window.rs
@@ -777,9 +777,6 @@ pub struct Window<R: Runtime> {
   manager: WindowManager<R>,
   pub(crate) app_handle: AppHandle<R>,
   js_event_listeners: Arc<Mutex<HashMap<JsEventListenerKey, HashSet<usize>>>>,
-
-  #[cfg(test)]
-  pub(crate) current_url: url::Url,
 }
 
 unsafe impl<R: Runtime> raw_window_handle::HasRawWindowHandle for Window<R> {
@@ -795,8 +792,6 @@ impl<R: Runtime> Clone for Window<R> {
       manager: self.manager.clone(),
       app_handle: self.app_handle.clone(),
       js_event_listeners: self.js_event_listeners.clone(),
-      #[cfg(test)]
-      current_url: self.current_url.clone(),
     }
   }
 }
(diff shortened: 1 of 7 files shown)


lucasfernog:
I think we can target this for the `1.x` branch instead so it lands on v1.5.

core/tauri/src/window.rs
```diff
@@ -1648,8 +1648,14 @@ impl<R: Runtime> Window<R> {
...
+  /// Navigates the webview to the defined url.
+  #[cfg(not(test))]
+  pub fn navigate(&mut self, url: Url) {
+    self.window.dispatcher.navigate(url).unwrap();
+  }
+
   #[cfg(test)]
-  pub(crate) fn navigate(&mut self, url: Url) {
+  pub fn navigate(&mut self, url: Url) {
     self.current_url = url;
```
lucasfernog: Now that we have an actual `navigate` function, we can remove the [current_url field](https://github.com/tauri-apps/tauri/blob/1542ad17d647460ec7e2750182a502debc110e28/core/tauri/src/window.rs#L782) and implement `fn navigate` under the [mock runtime](https://github.com/tauri-apps/tauri/blob/1542ad17d647460ec7e2750182a502debc110e28/core/tauri/src/test/mock_runtime.rs#L553). Basically you'll need to change the [Dispatcher#url](https://github.com/tauri-apps/tauri/blob/1542ad17d647460ec7e2750182a502debc110e28/core/tauri/src/test/mock_runtime.rs#L193) to a `Mutex<String>` and mutate it on the `navigate` impl.

With this change we have less `#[cfg(test)] #[cfg(not(test))]` to deal with.

lucasfernog (changes requested):

core/tauri/src/window.rs
jhutchins: I'll get that added.

jhutchins: I've reworked how url is handled in test contexts.

lucasfernog (approved):

lucasfernog:
Perfect!

kanatapple:
@lucasfernog 
Will this feature ship in v1.5?
If so, when will v1.5 be released?

PR #7305, opened 2023-06-26
feat(core): use handler instead of javascript for on load, closes  #7255

### LINKED ISSUE #7255: [bug] on_page_load doesn't work the same on Android
### Describe the bug

It appears like the initialization javascript that invokes the `__initialized` message is loaded on every page load for the desktop version of an application, but for an Android version that's only load on the initial page and subsequent page loads don't. This results in any callback loaded to the `on_page_load` gets invoked on every page load for desktop, but only on the initial load for Android.

### Reproduction

_No response_

### Expected behavior

I would expect that the `on_page_load` callback be called on every page load.

### Platform and versions

```text
❯ cargo tauri info

[✔] Environment
    - OS: Mac OS 12.6.6 X64
    ✔ Rust toolchain: stable-x86_64-apple-darwin (environment override by RUSTUP_TOOLCHAIN)
    - node: 19.9.0
    - pnpm: 7.27.0
    - yarn: 1.22.19
    - npm: 9.6.3

[-] Packages
    - tauri [RUST]: 2.0.0-alpha.10
    - tauri-build [RUST]: no manifest (2.0.0-alpha.5, 2.0.0-alpha.6)
    - wry [RUST]: 0.28.3
    - tao [RUST]: 0.19.1
    - @tauri-apps/cli [NPM]: 2.0.0-alpha.9

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../dist
```

### Stack trace

_No response_

### Additional context

I wonder if loading javascript into the page to handle this is the best way of doing it. It might make more sense to leverage the navigation handler or some other native hook on the various platforms and have that trigger the callback instead. It just seems like there's a lot more control interacting directly with the web viewer than trying to inject javascript into the page.

I would potentially be willing to investigate this if core team members think it's a good direction.

jhutchins:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

Depends on new features add to wry in [tauri/wry#980](https://github.com/tauri-apps/wry/pull/980)

### DIFF
--- a/core/tauri-runtime-wry/src/lib.rs
+++ b/core/tauri-runtime-wry/src/lib.rs
@@ -3027,6 +3027,7 @@ fn create_webview<T: UserEvent>(
     menu_ids,
     #[cfg(target_os = "android")]
     on_webview_created,
+    on_page_load_handler,
     ..
   } = pending;
   let webview_id_map = context.webview_id_map.clone();
@@ -3173,6 +3174,11 @@ fn create_webview<T: UserEvent>(
         })
       });
     }
+  }
+
+  if let Some(on_page_load_handler) = on_page_load_handler {
+    webview_builder =
+      webview_builder.with_on_load_handler(move |_, url| on_page_load_handler(todo!(), url));
   }
 
   let webview = webview_builder
--- a/core/tauri-runtime-wry/src/lib.rs
+++ b/core/tauri-runtime-wry/src/lib.rs
@@ -3027,7 +3027,7 @@ fn create_webview<T: UserEvent>(
     menu_ids,
     #[cfg(target_os = "android")]
     on_webview_created,
-    on_page_load_handler,
+    page_load_handler,
     ..
   } = pending;
   let webview_id_map = context.webview_id_map.clone();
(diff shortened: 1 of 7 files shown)


core/tauri-runtime-wry/src/lib.rs
```diff
@@ -3175,6 +3176,12 @@ fn create_webview<T: UserEvent>(
     }
   }
 
+  let label_c = label.clone();
+  if let Some(on_page_load_handler) = page_load_handler {
+    webview_builder =
+      webview_builder.with_on_load_handler(move |_, url| on_page_load_handler(&label_c, url));
+  }
+
```
jhutchins, line 3184: I don't love how this interfaces with the Wry API. I feel like there should be something better that could be done to get the Window object needed from the webview passed as a parameter instead of ignoring it and looking up the Window later by label.

jhutchins: This is better with the new API.

amrbashir:
Thanks for contribution but some how we forgot about this PR and it was implmenetd in https://github.com/tauri-apps/tauri/pull/8116

PR #7371, opened 2023-07-06
feat: ability to reset allowed/forbidden fs scope paths, closes #7366

### LINKED ISSUE #7366: [feat] Ability to disallow previously added files/directories in FsScope
### Describe the problem

An application might need to switch to a different workspace / folder which has been allowed via the FsScope api during runtime. This feature would allow the removal of previously added paths from the allowed scope.

Add two functions `disallow_directory` and `disallow_file` which resembles the `allow_file` and `allow_directory` functions found in fs.rs but instead of adding the paths to the allowed set, the disallow functions would remove the path.

### Alternatives considered

An alternative solution would be to have a function to reset the scope to the FsAllowlistScope configuration (initial config from app start). This could also solve the problem described above.

### Additional context

I wrote some code that does this and tested it on Windows. It's not too big of a change and I would be happy to make a pull request if this is something that should be added to Tauri. The "reset" alternative might be a simpler solution for my problem, but there might be a use case for removing certain added paths but not others.

LacombeJ:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
Issue: https://github.com/tauri-apps/tauri/issues/7366


core/tauri/src/scope/fs.rs
```diff
@@ -252,6 +262,21 @@ impl Scope {
...
     }
   }
+
+  /// This removes any paths added to allowed and forbidden scope by resetting
+  /// to the scope state to the given FsScope config
+  pub fn reset<R: crate::Runtime, M: crate::Manager<R>>(
+    &self,
+    manager: &M,
+    scope: &tauri_utils::config::FsScope,
+  ) -> crate::Result<()> {
```
amrbashir: instead of taking scope as an argument, you could provide a `clear` method which will empty allowed and forbidden hash sets, and `reset` which only takes one argument `Manager<R>` to reset to default

amrbashir (changes requested):

amrbashir (dismissed):
Thank you

amrbashir:
uh I should've noticed this earlier, @LacombeJ  could you rebase this PR to target `1.x` branch instead?

LacombeJ:
> uh I should've noticed this earlier, @LacombeJ could you rebase this PR to target `1.x` branch instead?

Yeah sure, I was actually going to ask about that

LacombeJ:
@amrbashir Let me know if I rebased properly.

This reset signature takes two arguments because 1.x has two possible scopes unlike dev/2.0.

```rs
pub fn reset<R: crate::Runtime>(&self, &App<R>, scope: &FsAllowlistScope) -> crate::Result<()> 

scope.reset(&app, &app.config().tauri.allowlist.fs.scope).unwrap();
scope.reset(&app, &app.config().tauri.allowlist.protocol.asset_scope).unwrap();
```

core/tauri/src/scope/fs.rs
```diff
@@ -258,6 +271,32 @@ impl Scope {
       false
     }
   }
+
+  /// This removes any paths added to allowed and forbidden scope by resetting
+  /// to the scope state to the given FsAllowlistScope config
+  pub fn reset<R: crate::Runtime>(
+    &self,
+    app: &App<R>,
+    scope: &FsAllowlistScope,
```
amrbashir: instead of accepting the default scope as an arugment, we should probably store the default (the one passed to `for_fs_api`) on the `Scope` struct and reuse it here

amrbashir (changes requested):

core/tauri/src/scope/fs.rs
LacombeJ: @amrbashir I wrote the changes but before I push I wanted to ask your opinion on this:

What if we only had one function `reset` that takes no arguments? We can store the initial allowed/forbidden patterns whenever Scope is created, similar to how you described it. The reset function will just restore allowed/forbidden scope back to it's original state.

```rs
pub fn reset(&self) -> crate::Result<()>
```

My reasoning:
1. If the app describes a set of forbidden directories in configuration, it probably shouldn't be cleared (the developer shouldn't include it otherwise). And I am thinking the same for any allowed paths
2. For version 2, I believe scopes are created with a manager, so the reset signature would change from accepting an `App<R>` to `Manager<R>` (probably not a big deal since App implements Manager)

amrbashir: > ```rs
> pub fn reset(&self) -> crate::Result<()>
> ```
yeah this sounds better

core/tauri/src/scope/fs.rs
```diff
@@ -36,6 +36,14 @@ pub struct Scope {
   forbidden_patterns: Arc<Mutex<HashSet<Pattern>>>,
   event_listeners: Arc<Mutex<HashMap<Uuid, EventListener>>>,
   match_options: glob::MatchOptions,
+  original_scope: OriginalScope,
```
amrbashir: lets not create a new type, just add two fields instead, `default_allowed` and `default_forbidden`

amrbashir (changes requested):

core/tauri/src/scope/fs.rs
```diff
@@ -258,6 +278,15 @@ impl Scope {
       false
     }
   }
+
+  /// This removes any paths added to allowed and forbidden scope by resetting
+  /// to the scope state to it's initial/default configuration
```
amrbashir:
```suggestion
  /// This removes any allowed or forbidden paths by resetting
  /// the scope to its initial state which matches the scope defined in the configuration file
```

amrbashir (dismissed):

lucasfernog:
Now I see it's kinda hard to use the ResetScope values so we could actually add a function that pulls a ResetScope and readds the patterns to the Scope.

core/tauri/src/scope/fs.rs
```diff
@@ -258,6 +275,48 @@ impl Scope {
       false
     }
   }
+
+  /// This removes any allowed or forbidden paths by resetting
+  /// the scope to its initial state which matches the scope defined in the configuration file.
+  ///
+  /// Returns the removed allowed and forbidden patterns.
+  pub fn reset(&self) -> ResetScope {
```
amrbashir, line 283: Maybe we could return the difference as a `Scope` type instead of a new type? and add a `clear` function that returns the removed scope also and then we can `impl Extend<Scope> for Scope`

amrbashir (changes requested):

core/tauri/src/scope/fs.rs
lucasfernog: I thought about that too, but Scope has a lot more information..

I don't see the usage of a clear() function. And implementing Extend would need us to implement a custom iterator logic too.

lucasfernog: Aalso the scope isn't really mutable (it uses interior mutability) and Extend expects it to be mutable.

amrbashir: ah right, then a custom extend function it is

amrbashir: nvm about the clear function also, I couldn't think of a use-case different than reset

lucasfernog:
This now just needs an approval from the security team.

amrbashir (approved):

PR #7382, opened 2023-07-08
fix(core/api): ignore children errors for `read_dir`, closes #7379

### LINKED ISSUE #7379: [bug] fs `readDir` fails for `BaseDirectory.Home` root with "Access is denied. (os error 5)", even as Admin
### Describe the bug

**What doesn't work:**
Using `readDir('', { dir: BaseDirectory.Home, recursive: true })` on Windows throws  
`path: <homeDirPath>: Access is denied. (os error 5)`.

**What this means:**
This suggests that the user does not have access to `<homeDirPath>` which can not be the case, since using `recursive: false` succeeds.

**Possible cause:**  
The error is thrown because the user doesn't have access to some subdirectory of `<homeDirPath>` and trying to read that directory failed.

**Why this needs fixing:**  
- The error doesn't tell which directory failed; it always uses the `<homeDirPath>`. This makes debugging impossible.
- There is no way to obtain the entries that _didn't_ fail. This makes a `readDir` invocation useless.

### Reproduction

```js
import { readDir, BaseDirectory } from '@tauri-apps/api/fs';

async function test() {
  // Empty string as dir parameter
  await readDir('', { dir: BaseDirectory.Home, recursive: true }); // Throws "path: <homeDirPath>: Access is denied. (os error 5)"
  // Dot as dir parameter
  await readDir('.', { dir: BaseDirectory.Home, recursive: true }); // Throws "path: <homeDirPath>.: Access is denied. (os error 5)"
}

await test();
```

### Expected behavior

- Only throw an error if the root directory of `readDir(rootDir, { recursive: true })` can not be read
- When throwing an error, use the path of the directory that _actually_ failed instead of the root directory
- Include the problematic subdirectory in the result but set `subdirectory.children = [];` (`'empty'`)
- Exclude the problematic subdirectory from the result (`'exclude'`)
- For backward compatibility, add `errors?: 'throw' | 'empty' | 'exclude'` to type `FsDirOptions` (defaulting to `'throw'`)

### Platform and versions

```text
[✔] Environment
    - OS: Windows 10.0.22621 X64
    ✔ WebView2: 114.0.1823.67
    ✔ MSVC:
    - node: 20.4.0
    - npm: 9.7.2

[-] Packages
    - tauri [RUST]: 1.4.1
    - tauri-build [RUST]: 1.4.0
    - wry [RUST]: 0.24.3
    - tao [RUST]: 0.16.2

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../build
    - framework: Svelte
    - bundler: Vite
```

### Stack trace

```text
Uncaught (in promise) path: <homeDirPath>: Access is denied. (os error 5)
```

### Additional context

_No response_

amrbashir:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/.changes/read_dir_recursive_errors.md
+++ b/.changes/read_dir_recursive_errors.md
@@ -1,0 +1,5 @@
+---
+'tauri': patch
+---
+
+Fix `tauri::api::read_dir` and the JS API `fs.readDir` failing if one of the children was not accessible when `recursive` option was `true`.
--- a/.changes/read_dir_recursive_errors.md
+++ b/.changes/read_dir_recursive_errors.md
@@ -1,5 +1,5 @@
 ---
-'tauri': patch
+'tauri': patch:bug
 ---
 
 Fix `tauri::api::read_dir` and the JS API `fs.readDir` failing if one of the children was not accessible when `recursive` option was `true`.
--- a/core/tauri/src/api/dir.rs
+++ b/core/tauri/src/api/dir.rs
@@ -62,8 +62,8 @@ pub(crate) fn read_dir_with_options<P: AsRef<Path>>(
   options: ReadDirOptions<'_>,
 ) -> crate::api::Result<Vec<DiskEntry>> {
   let mut files_and_dirs: Vec<DiskEntry> = vec![];
-  for entry in fs::read_dir(path)? {
-    let path = entry?.path();
+  for entry in fs::read_dir(path)?.flatten() {
+    let path = entry.path();
     let path_as_string = path.display().to_string();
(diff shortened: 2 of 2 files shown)


wusyong (approved):

lucasfernog:
We can't ignore errors, we end up losing information. We should tell the user which path failed instead.

PR #7432, opened 2023-07-17
feat: allow more flexible http requests in updater closes #7422

### LINKED ISSUE #7422: [feat] More flexibility in HTTP requests used by Updater
### Describe the problem

We need to be able to construct different headers for the HTTP request used to check for available updates, and the HTTP  request used to download the update. 

For example, if you set AWS SIGv4 headers on the request to check for updates, the same headers are not appropriate to attempt to download the update from AWS S3 (in fact the S3 request is impossible) - but I can find no way to modify the headers between checking and downloading.  

Sending signed requests, using user provided credentials for example,  would also be significantly easier if the updater endpoints could be modified after application start (different from the solution suggested https://github.com/tauri-apps/tauri/issues/2595#issuecomment-1160772317) 

Minimally, setters on `UpdateResponse` that allowed the headers to be changed. 
Ideally, also setters on `UpdateBuilder` that allowed the endpoints to be changed dynamically. 

### Alternatives considered

_No response_

### Additional context

_No response_

gmchale79:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist
- [x ] When resolving issues, they are referenced in the PR's title (e.g `fix: remove a typo, closes #___, #___`)

### Other information
These small changes allow us to use AWS SigV4 authenticated requests for checking for updates, and pre-signed S3 URLs for updater downloads - a call flow that was previously not possible. 
I'm aware that there is other work in this area in 2.x, but we need to use 1.x for the foreseeable future.


core/tauri/src/updater/core.rs
```diff
@@ -505,6 +505,33 @@ impl<R: Runtime> Clone for Update<R> {
...
+  where
+    HeaderName: TryFrom<K>,
+    <HeaderName as TryFrom<K>>::Error: Into<http::Error>,
+  {
+    let key: std::result::Result<HeaderName, http::Error> = key.try_into().map_err(Into::into);
+    self.headers.remove(key?);
+    Ok(self)
+  }
+}
+
```
amrbashir: why open a new `impl` block? just reuse the one below on line 535

core/tauri/src/updater/core.rs
```diff
@@ -505,6 +505,33 @@ impl<R: Runtime> Clone for Update<R> {
   }
 }
 
+impl<R: Runtime> Update<R> {
+  pub fn header<K, V>(mut self, key: K, value: V) -> Result<Self>
```
amrbashir, line 505:
```suggestion
  pub(crate) fn header<K, V>(mut self, key: K, value: V) -> Result<Self>
```

core/tauri/src/updater/mod.rs
```diff
@@ -275,6 +275,12 @@ impl<R: Runtime> UpdateBuilder<R> {
     Ok(self)
   }
 
+  /// Add endpoints.
```
amrbashir:
```suggestion
  /// Set the endpoints.
```

core/tauri/src/updater/mod.rs
```diff
@@ -395,6 +402,28 @@ impl<R: Runtime> UpdateResponse<R> {
...
+
+  /// Removes a header from the update request.
+  pub fn remove_header<K>(mut self, key: K) -> Result<Self>
+  where
+    HeaderName: TryFrom<K>,
+    <HeaderName as TryFrom<K>>::Error: Into<http::Error>,
+  {
+    self.update = self.update.remove_header(key)?;
+    Ok(self)
+  }
```
amrbashir: would you explain why would you need this?

amrbashir (changes requested):
Thank you for the contribution, this looks okay but just needs a few modifications.

Also please add a change file in `.changes` directory

core/tauri/src/updater/mod.rs
```diff
@@ -275,6 +275,12 @@ impl<R: Runtime> UpdateBuilder<R> {
     Ok(self)
   }
 
+  /// Add endpoints.
+  pub fn endpoints(mut self, urls: &[String]) -> Result<Self> {
+    self.inner = self.inner.urls(urls);
+    Ok(self)
+  }
```
amrbashir:
```suggestion
  pub fn endpoints(mut self, urls: &[String]) -> Self {
    self.inner = self.inner.urls(urls);
    self
  }
```

core/tauri/src/updater/mod.rs
lucasfernog: probably because he needs a header for the JSON fetch request, but not for the install request

lucasfernog (approved):

gmchale79:
Yes, exactly. If we set signed headers for the fetch request, they are not valid for the subsequent install request to S3 and S3 rejects them. Apologies for the delayed response.

core/tauri/src/updater/mod.rs
gmchale79: Yes, exactly. If we set signed headers for the fetch request, they are not valid for the subsequent install request to S3 and S3 rejects them. Apologies for the delayed response.

amrbashir (approved):

PR #7439, opened 2023-07-17
feat: Add ordered navigation handler for plugins, closes #7306

### LINKED ISSUE #7306: [feat] navigation handler hook for plugins
### Describe the problem

I have some use cases where it would be useful for a plugin to be able to add a navigation handler for it own usage.

I would like a `with_navigation_handler` function or similar added to the `Plugin` interface.

### Alternatives considered

Something similar could probably be achieved by adding an event hook into wry that created a `WindowEvent` when navigation started and then using the pre-existing `on_event` method.

The on page load kind of works, but it only triggers after a page starts loading instead of preventing navigation. 

### Additional context

_No response_

jhutchins:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


jhutchins:
This should resolve  #7306

jhutchins:
@lucasfernog these changes are great. I thought about making some similar ones, but didn't want to be responsible for breaking changes.

lucasfernog:
No need to worry about it right now since it targets v2. Thanks for the contribution!

amrbashir (approved):

Issue #7461, opened 2023-07-19
[feat] Allow window labels regex or prefix for remote domain access scope


james138138:
### Describe the problem

Currently, according to the [RemoteDomainAccessScope doc](https://tauri.app/v1/api/config#remotedomainaccessscope) and code(https://github.com/tauri-apps/tauri/blob/7b45e7fa1b02b75f9aff06e22f4a0d7071bf71ae/core/tauri/src/scope/ipc.rs#L139), tauri does strict equal check for window labels to enable tauri api.

We have an application where multiple windows(e.g. feature1-window1, feature1-window2) are opened dynamically, and we want them to be able to call `invoke` api to commute with the host application.

### Describe the solution you'd like

I would like to be able to check window label with regex or prefix instead of strict equal.

### Alternatives considered

Create a websocket server to proxy, which seems little overkill for the above use case.

### Additional context

This is patch I'm using
```
      let mut matches_window = false;
      for window_label_pattern in &s.windows {
        if Pattern::new(window_label_pattern).unwrap_or_default().matches(&label) {
          matches_window = true;
          break;
        }
      }
```

PR #7468, opened 2023-07-21
feat(core): allow window label pattern for remote domain access scope

james138138:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/core/tauri-utils/src/config.rs
+++ b/core/tauri-utils/src/config.rs
@@ -1232,7 +1232,7 @@ pub struct RemoteDomainAccessScope {
   pub scheme: Option<String>,
   /// The domain to allow.
   pub domain: String,
-  /// The list of window labels this scope applies to.
+  /// The list of window label glob patterns(`**` ignored) this scope applies to.
   pub windows: Vec<String>,
   /// The list of plugins that are allowed in this scope.
   #[serde(default)]
--- a/core/tauri-utils/src/config.rs
+++ b/core/tauri-utils/src/config.rs
@@ -1223,6 +1223,26 @@ impl Default for DisabledCspModificationKind {
   }
 }
 
+fn de_remote_domain_access_scope_windows<'de, D>(deserializer: D) -> Result<Vec<String>, D::Error>
+where
+  D: Deserializer<'de>,
+{
+  let windows = Option::<Vec<String>>::deserialize(deserializer)?.unwrap_or_default();
+  for w in &windows {
+    if w == "*" || w == "**" {
+      return Err(serde::de::Error::custom(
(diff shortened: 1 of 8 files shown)


james138138:
This should resolve https://github.com/tauri-apps/tauri/issues/7461

james138138:
Doc PR is https://github.com/tauri-apps/tauri-docs/pull/1330

lucasfernog:
I've pushed some changes to enhance this but they are breaking changes (good thing this PR targets v2 anyway). If we want to backport this for v1, we need to stick to your original PR.

Thanks for your contribution! Let's wait for a security audit on this.

james138138:
@lucasfernog Thanks for the updating, there was a [build error](https://github.com/tauri-apps/tauri/actions/runs/5645255457/job/15290657552), I've pushed a [commit](https://github.com/tauri-apps/tauri/pull/7468/commits/81c6dced5549a8d3726c12499236f281cd4b08f6), please take a look. Thanks.

core/tauri-utils/src/config.rs
```diff
@@ -1223,6 +1223,26 @@ impl Default for DisabledCspModificationKind {
   }
 }
 
+fn de_remote_domain_access_scope_windows<'de, D>(deserializer: D) -> Result<Vec<String>, D::Error>
+where
+  D: Deserializer<'de>,
+{
+  let windows = Option::<Vec<String>>::deserialize(deserializer)?.unwrap_or_default();
+  for w in &windows {
+    if w == "*" || w == "**" {
```
tweidinger, line 1232: I think this check is not sufficient as devs can just write `?*` or `[?*]` to bypass this full wildcard prevention check.
Also from my perspective allowing `**` anywhere doesn't make sense.  The `glob` crate is imho meant for  file paths and we should use a more simpler approach here.

From my perspective we should use simple string comparisons like `starts_with` or `ends_with` to allow something like `foo-` or `-remote` for labels instead of using complex matching pattern. We could expose these two patterns via a flag (e.g: `"fuzzy_matching" : true`)

If there is a reason to go for regex patterns I'm all ears but my assumption is most devs want to allow remote access to certain window types where they can prefix with `remote-`. Using the glob crate would allow for more fine grained matching but will be a footgun for most devs.

tweidinger (changes requested):

lucasfernog:
We redesigned the whole remote URL scope on #8428 

Thank you for this PR!

PR #7481, opened 2023-07-22
fix(bundler): do not skip updater bundle target if only a macOS DMG is generated

AlexTMjugador:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

As explained in https://github.com/tauri-apps/tauri/pull/7189#discussion_r1271316115, this little change addresses recent, unintended behavior changes in how the updater bundle operates.

### DIFF
--- a/.changes/bundler-macos-updater-dmg-only-fix.md
+++ b/.changes/bundler-macos-updater-dmg-only-fix.md
@@ -1,0 +1,5 @@
+---
+'tauri-bundler': 'patch:bug'
+---
+
+Fix bundler skipping updater artifacts if only a macOS DMG bundle target is specified.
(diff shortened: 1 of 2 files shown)


amrbashir:
I think we only support updating `.app` on macOS, and so `DMG` should be ignored if it is the only macOS target specified. cc @lucasfernog

AlexTMjugador:
As far as I know, DMGs are containers for `.app` bundles, so generating a DMG implies generating an `.app` bundle and it is technically possible to update an application that was installed from a DMG by replacing its `.app`. But I stumbled upon this on CI when updating Tauri, and I don't have a Mac to confirm, so please take this with a grain of salt!

lucasfernog:
@AlexTMjugador you're right. Thanks for catching this and contributing!

lucasfernog (approved):

Issue #6888, opened 2023-05-07
[bug] Sidecar not working after signing app - MAC OS


brainbook-space:
### Discussed in https://github.com/tauri-apps/tauri/discussions/2269

<div type='discussions-op-text'>

<sup>Originally posted by **NitinRamnani** July 21, 2021</sup>
Did anyone tried to use sidecar with signed app? It seems sidecar doesn't work.</div>

brainbook-space:
I tried to build the binary in examples/sidecar on Mac OS M1 CPU.
The bundled dmg installed and work fine from my machine.
If I uploaded the bundled dmg to my AWS server and download it. Then it does not work. the sidecar binary is not code signed.

I have tried my own sidecar app, same issue when I download the bundled dmg from my AWS server.

PR #7487, opened 2023-07-24
fix(bundler): sign the exe before the bundler step

FabianLars:
https://discord.com/channels/616186924390023171/1010253864110407800

This way there's better support for standalone exe or in case devs create their own bundle formats.

At first i wanted to add that to the cli but thought that'd be a bit awkward since all the signing happens in the bundler and also in case we really end up making the bundler a more general thing for other projects to use, but i'm fine with changing that again.

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/tooling/bundler/src/bundle.rs
+++ b/tooling/bundler/src/bundle.rs
@@ -63,6 +63,19 @@ pub fn bundle_project(settings: Settings) -> crate::Result<Vec<Bundle>> {
     warn!("Cross-platform compilation is experimental and does not support all features. Please use a matching host system for full compatibility.");
   }
 
+  // Sign main windows binary before the bundling step in case neither wix and nsis bundles are enabled
+  #[cfg(target_os = "windows")]
+  {
+    let main_binary = settings
+      .binaries()
+      .iter()
+      .find(|bin| bin.main())
+      .ok_or_else(|| anyhow::anyhow!("Failed to get main binary"))?;
+    let app_exe_source = settings.binary_path(main_binary);
+
+    windows::sign::try_sign(&app_exe_source, &settings)?;
+  }
+
   for package_type in &package_types {
     // bundle was already built! e.g. DMG already built .app
     if bundles.iter().any(|b| b.package_type == *package_type) {
--- a/tooling/bundler/src/bundle.rs
+++ b/tooling/bundler/src/bundle.rs
@@ -63,17 +63,28 @@ pub fn bundle_project(settings: Settings) -> crate::Result<Vec<Bundle>> {
     warn!("Cross-platform compilation is experimental and does not support all features. Please use a matching host system for full compatibility.");
   }
 
-  // Sign main windows binary before the bundling step in case neither wix and nsis bundles are enabled
   #[cfg(target_os = "windows")]
   {
-    let main_binary = settings
-      .binaries()
-      .iter()
-      .find(|bin| bin.main())
-      .ok_or_else(|| anyhow::anyhow!("Failed to get main binary"))?;
-    let app_exe_source = settings.binary_path(main_binary);
+    // Sign windows binaries before the bundling step in case neither wix and nsis bundles are enabled
+    for bin in settings.binaries() {
+      let bin_path = settings.binary_path(bin);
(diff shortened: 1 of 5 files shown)


tooling/bundler/src/bundle.rs
```diff
@@ -63,6 +63,19 @@ pub fn bundle_project(settings: Settings) -> crate::Result<Vec<Bundle>> {
     warn!("Cross-platform compilation is experimental and does not support all features. Please use a matching host system for full compatibility.");
   }
 
+  // Sign main windows binary before the bundling step in case neither wix and nsis bundles are enabled
+  #[cfg(target_os = "windows")]
+  {
+    let main_binary = settings
+      .binaries()
```
lucasfernog: should we take this opportunity to sign all binaries instead of just the main one?

FabianLars: yeah, why not. gonna look into that asap

FabianLars: 2 Questions:
- Do all the binaries in settings.binaries need to be signed? I don't know what binaries it bundles other than the main app
- I think we need a setting to skip signing sidecars (or some of them) in case they were already signed beforehand by a third party or something idk

lucasfernog: it bundles all binaries defined in the [Cargo manifest](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#binaries). The sidecars are in a separate getter, `external_binaries`.

FabianLars: > it bundles all binaries defined in the [Cargo manifest](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#binaries)

~~So i assume we care about them? aka they need to be signed too? I can't think of usecase rn~~ nvm, yeah we should sign it.

That then leaves us with the sidecar skip setting.

lucasfernog: I guess no one is even using this otherwise it would probably have issues.

It's pretty much like a sidecar, but written in Rust in the same Cargo project.

lucasfernog: Like how some projects implement the updater as a separate process. But I don't remember seeing anyone using it (except whoever requested this during our alpha, I think it was like a local server/client app).

FabianLars: Sorry for the delay, the change is windows-only for now, i'll try to at least add the settings.binaries for macos but i'm unsure how to handle sidecars. Mainly how to prevent re-signing already signed binaries and if they still have to be individually notarized or not and if that works if the sidecar was signed by a third party.

Anyway, here is what i added on Windows:
All the binaries in settings.binaries are code signed before the wix/nsis bundlers are invoked.
All the sidecar binaries are also code signed before the wix/nsis bundlers are invoked just to keep the logic together and to not have to duplicate it. The sidecar binaries get treated differently since we use signtool to check for existing signatures. I added an env var to skip that check and forcefully sign them.

lemme know what you think

FabianLars:
_Posting this as a normal ocmment instead of an answer for better visibility_

Okay, nevermind about the macos changes, they need a bit more exploration which i think is outside the scope of this PR:

We currently use the `--deep` option which should already make it sign all the binaries in the app bundle, but there are 2 issues with that.
1) It doesn't seem to actually work because https://github.com/tauri-apps/tauri/issues/6888 wouldn't be a thing then, and
2) all resources i found heavily discourage the use of `--deep` and instead recommend a more manual approach.

Lastly, there's an issue with how we handle plist files. Currently we apply it to all signed files and not just the main binary, which also seems to be frowned upon. This behavior is currently inherited from the `--deep` option but still something we have to keep in mind when we change the approach. Then we'll need to discuss whether we're okay with that or if we want to have a more granular (sidecar specific) plist appoach

----

I am interested in working on that, but not until i have a working macos setup 🙃 

So in that context i consider this PR ready for review.

lucasfernog (approved):

PR #7512, opened 2023-07-28
feat(core): Manager::emit_filter and optimize serialization

icambron:
I emit a lot of events to multiple--but specific--windows, and one of the bottlenecks there is serialization. There are two problems:

1. There is no public API for emitting events to specific, multiple windows
2. The private implementation of `emit_filter` serializes on a per-window basis

This fixes both of those. A couple of design choices:

 1. I figured `emit_filter` was the most flexible option, compared to taking a HashSet of window ids or similar. It's easy enough to pass `|w| some_window_ids.contains(w.label())`.
 2. There's a small potential perf downside if none of the windows match filter, because we've already serialized the payload. I think this is a minor edge case but I could be wrong.

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


core/tauri/src/manager.rs
```diff
@@ -1102,12 +1102,13 @@ impl<R: Runtime> WindowManager<R> {
     S: Serialize + Clone,
     F: Fn(&Window<R>) -> bool,
   {
+    let emit_args = WindowEmitArgs::from(event, source_window_label, payload)?;
     assert_event_name_is_valid(event);
     self
       .windows_lock()
       .values()
       .filter(|&w| filter(w))
-      .try_for_each(|window| window.emit_internal(event, source_window_label, payload.clone()))
```
icambron, line 1110: I don't think this clone was ever needed. Even when we serialized once per window, it's just creating a string, so using `&payload` would have been better.

core/tauri/src/manager.rs
```diff
@@ -1102,12 +1102,13 @@ impl<R: Runtime> WindowManager<R> {
     S: Serialize + Clone,
     F: Fn(&Window<R>) -> bool,
   {
+    let emit_args = WindowEmitArgs::from(event, source_window_label, payload)?;
```
icambron, line 1108: serialize once, emit many times

core/tauri/src/manager.rs
```diff
@@ -1102,12 +1102,13 @@ impl<R: Runtime> WindowManager<R> {
...
     assert_event_name_is_valid(event);
     self
       .windows_lock()
```
icambron: One question I had here was whether it made more sense to call `windows()` instead of `windows_lock()`. This would have the effect of cloning all the windows (bad) but also of not holding the lock open (good). That seems to matter for two reasons. First, plausibly someone could write a filter that needs the lock, which would immediately cause a painful and unintuitive deadlock. Second, emitting might take a while, and ideally you could do other window things while that was happening.

lucasfernog: When triggering events we usually clone the windows, it's safer. We should change this one.

icambron: @lucasfernog done

lucasfernog (dismissed):

lucasfernog (approved):

PR #7534, opened 2023-07-31
Extend context.rs to dynamically find the OUT_DIR for certain assets

sinistersnare:
Hello, my team is moving to Bazel, and we can not use tauri-cli or any of your custom cargo-based tooling. We would like to contribute our changes to you all, so you can better support Bazel. 

These changes should be non-breaking, as it just recalculates the OUT_DIR at build.rs-time and compile-time separately. So it should not affect Cargo builds.

I think that TODO, if this is going to be fully generalized we need to add this functionality to some other parts of at least this file.

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/core/tauri-codegen/src/context.rs
+++ b/core/tauri-codegen/src/context.rs
@@ -470,15 +470,20 @@ fn ico_icon<P: AsRef<Path>>(
   let width = entry.width();
   let height = entry.height();
 
-  let out_path = out_dir.join(path.file_name().unwrap());
+  let icon_file_name = path.file_name().unwrap();
+  let out_path = out_dir.join(icon_file_name);
   write_if_changed(&out_path, &rgba).map_err(|error| EmbeddedAssetsError::AssetWrite {
     path: path.to_owned(),
     error,
   })?;
 
-  let out_path = out_path.display().to_string();
-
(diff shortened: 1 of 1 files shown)


lucasfernog:
@sinistersnare is this PR ready or do you want to change some other OUT_DIR usage?

sinistersnare:
I think its fine for now. It can be further extended as needed using the code here as a template for future contributors.

lucasfernog:
By the way @sinistersnare this PR is targeting v2. So if you want this to be part of v1.5, it needs to target the `1.x` branch instead.

sinistersnare:
Ah thanks for the check on that. I rebased on 1.x

sinistersnare:
I extended the PR to include another case useful to my project. Overall, this kind of stuff is really useful for non-Cargo users, and the codegen crate would be wise to generalize from any build tool.

lucasfernog (dismissed):

lucasfernog (approved):

PR #7605, opened 2023-08-14
docs: Be more specific about plugin names in ipc scope

FabianLars:
This is a surprisingly common issue.

- [ ] Bugfix
- [ ] Feature
- [x] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/core/tauri-utils/src/config.rs
+++ b/core/tauri-utils/src/config.rs
@@ -1248,6 +1248,7 @@ pub struct RemoteDomainAccessScope {
   /// The list of window labels this scope applies to.
   pub windows: Vec<String>,
   /// The list of plugins that are allowed in this scope.
+  /// The names should be without the `tauri-plugin-` prefix and `-api` suffix, for example `"store"` for `tauri-plugin-store`.
   #[serde(default)]
   pub plugins: Vec<String>,
   /// Enables access to the Tauri API.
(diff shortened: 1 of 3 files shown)


lucasfernog (approved):

core/tauri-utils/src/config.rs
```diff
@@ -1248,6 +1248,7 @@ pub struct RemoteDomainAccessScope {
   /// The list of window labels this scope applies to.
   pub windows: Vec<String>,
   /// The list of plugins that are allowed in this scope.
+  /// The names should be without the `tauri-plugin-` prefix, for example `"store"` for `tauri-plugin-store`.
```
amrbashir, line 1251: Not a blocker but this still is a bit confusing because a plugin could have the name `tauri-plugin-foo` but in code it is actually, `tauri::plugin::Builder::new("bar").build()`, it is highly unlikely now but could be a problem down the road when some plugins have conflicting names so we should note where to find the correct plugin name, or have a convention around it.

FabianLars: Yeah, i thought about that too, but no idea at all how to make this easy to grasp for non-plugin authors & non-rust users so i went with the hotfix like wording for now :/

amrbashir (approved):

core/tauri-utils/src/config.rs
amrbashir: yeah...

PR #7616, opened 2023-08-14
refactor(bundler): switch to notarytool, closes #4300

### LINKED ISSUE #4300: (bundler): Switch to notarytool for notarization before Fall 2023
For reference: https://developer.apple.com/videos/play/wwdc2022/10109/

Relevant file: https://github.com/tauri-apps/tauri/blob/dev/tooling/bundler/src/bundle/macos/sign.rs

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/.changes/config.json
+++ b/.changes/config.json
@@ -8,7 +8,8 @@
     "pref": "Performance Improvements",
     "changes": "What's Changed",
     "sec": "Security fixes",
-    "deps": "Dependencies"
+    "deps": "Dependencies",
+    "breaking": "Breaking Changes"
   },
   "defaultChangeTag": "changes",
   "pkgManagers": {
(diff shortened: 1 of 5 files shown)


amrbashir (commented):
releasing this as a minor release might break some users CI, should it be added behind a flag and the default would warn users that it is to be discontinued soon and they should migrate?

lucasfernog:
It will break anyway in a couple months...

amrbashir:
true but we could have a soft migration at the moment instead of breaking their CI early on.

FabianLars:
tbh i feel like it's too late for that now, especially with our slow-ish release cycle

lucasfernog:
We could also look for the API key file on the known paths Apple used to look for it...

amrbashir (approved):

PR #7640, opened 2023-08-16
feature: extend tauri builder with updater_header method

Atrox:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information
This change extends `tauri::Builder` with a new method `updater_header` (in the same vein as the existing `updater_target`) to set headers on the default updater instance.

### DIFF
--- a/.changes/tauri-builder-updater-header.md
+++ b/.changes/tauri-builder-updater-header.md
@@ -1,0 +1,5 @@
+---
+"tauri": minor:feat
+---
+
+Added `tauri::Builder::updater_header` to add headers to the tauri updater.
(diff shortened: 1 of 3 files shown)


amrbashir:
Thanks for your contribution, but since updater in v2 has been moved to its own plugin, and since we already provide a similar API https://docs.rs/tauri/latest/tauri/updater/struct.UpdateBuilder.html#method.header, I will close this PR in order to avoid another breaking change.

Issue #3007, opened 2021-12-01
How to make tauri use http://tauri.localhost/ instead of https://tauri.localhost/ ? The page was blocked since the page was loaded over HTTPS.


xuchaoqian:
---
 
> **Updated @2023-09-17:** If anyone has encountered this problem, please pay attention to this PR: [Allow http origin on Windows](https://github.com/tauri-apps/tauri/pull/7645)

---

**Is your feature request related to a problem? Please describe.**
WebView console shows the following errors when we connect to endpoint 'ws:<URL>/': 
`Mixed Content: The page at '<URL>' was loaded over HTTPS, but attempted to connect to the insecure WebSocket endpoint 'ws:<URL>/'. This request has been blocked; this endpoint must be available over WSS.
`

**Describe the solution you'd like**
We should allow to set an option in tauri.conf.json to use http://tauri.localhost/ instead of https://tauri.localhost/.

**Describe alternatives you've considered**
None.

**Additional context**
1) We can NOT use WSS for historical reasons. 2) WS works fine on macOS, but does NOT works on Windows.

lity:
I met the same problem too.

mzy2240:
Me too. Electron supports this mode, and it is extremely useful for development and testing, even in some rare production env.

amrbashir:
Can you post `tauri info` please ? and whether you added `ws:` to your csp configuration ?

Also please explain your feature request a bit further and how the api for it should look like, as I am having some trouble understanding your request.

mzy2240:
I don't want to hijack the thread but here is mine:
```
Operating System - Windows, version 10.0.19044 X64
Webview2 - 96.0.1054.41

Node.js environment
  Node.js - 14.17.6
  @tauri-apps/cli - 1.0.0-beta.10
  @tauri-apps/api - Not installed

Global packages
  npm - 6.14.15
  yarn - Not installed

Rust environment
  rustc - 1.56.1
  cargo - 1.56.0

App directory structure
/.git
/dist
/dist_electron
/node_modules
/public
/py
/pydist
/src
/src-tauri

App
  tauri.rs - 1.0.0-beta.8
  build-type - bundle
  CSP - default-src blob: data: filesystem: ws: wss: http: https: tauri: 'unsafe-eval' 'unsafe-inline' 'self' img-src: 'self'
  distDir - Set automatically by Vue CLI plugin
  devPath - Set automatically by Vue CLI plugin
  framework - Vue.js (Vue CLI)
[...]

PR #7645, opened 2023-08-17
feat(core): Allow http origin on Windows, fixes: #3007

FabianLars:
Sooooo, first i wanted to make it per window but there was one place where i had no idea how to get a per-window config (click on the first commit and search for `FIXME:`) so i gave up on that for now and switched to a per-app setting on the security config. In theory i would prefer a per-window setting but i also don't think it's worth the headache. Apart from that, the only real annoyence left is convertFileSrc requiring the user to select the scheme rn. I'm not sure if anyone has a better idea that doesn't involve an ipc request, the only thing i have is a init_script that sets a global var on the window object but i'm not sure if that also works for external urls.

Anyway, mostly putting this up to gain feedback (wry isn't realeased yet anyway).

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


xuchaoqian:
Great job! 

per-app setting is enough for most use cases; per-window setting will make things be complicated.

amrbashir:
For the `convertFileSrc` you can use `window.location.protocol` and it will be automatically correct whether `https` or `http` is used

FabianLars:
That won't be reliable for non-standard origins like the localhost plugin or external urls tho.

amrbashir:
That's still fine, you could detect if it is http or https and fallback to https if it is another protocol with an option to explicitly specify whether to use https or http

FabianLars:
I went for a location.origin check. That's the only thing that came to mind which shouldn't have any false positives or whatever. Checking just for http/https isn't enough since http origins like the ones from the localhost plugin can still use https custom protocols.

tooling/api/src/tauri.ts
```diff
@@ -126,10 +127,24 @@ async function invoke<T>(cmd: string, args: InvokeArgs = {}): Promise<T> {
...
   const path = encodeURIComponent(filePath)
+
+  let scheme =
+    window.location.origin === 'http://tauri.localhost' ? 'http' : 'https'
+  if (http === true) {
+    scheme = 'http'
+  }
+  if (http === false) {
+    scheme = 'https'
+  }
```
amrbashir:
```suggestion
  let scheme =
    window.location.origin === 'http://tauri.localhost' ? 'http' : 'https'
  if (http) {
    scheme = http ? 'http' : 'https'
  }
```

amrbashir (changes requested):
@FabianLars for the per-window thing, you can trace how `window_origin` is passed to this `prepare_uri_scheme_protocol` function https://github.com/tauri-apps/tauri/blob/1.x/core/tauri/src/manager.rs#L703 and do the same, then you can pass it to `get_asset` and then to `set_csp`

tooling/api/src/tauri.ts
FabianLars: With your code a user can't use `false` to explicitly overwrite it (also no need to check for http's value again in the if since it can only be true at that point). The closest thing to my intentionally stupid code would be to use `typeof boolean` maybe?

To be clear, my intention is to have the user to _always_ be able to overwrite it with true/false even tho i can't think of a concise example where false would be actually needed...

FabianLars:
> for the per-window thing, you can trace how window_origin is passed to this prepare_uri_scheme_protocol function https://github.com/tauri-apps/tauri/blob/1.x/core/tauri/src/manager.rs#L703 and do the same, then you can pass it to get_asset and then to set_csp

Okay, i will have nightmares about the codebase now but at least i know that it's possible now, thanks :D (Edit: srsly i know i'm super full of myself but i still think that it's a bit concerning i didn't see that on third sight)

So the question left is whether we want it to be per-window or per-app. I like the modularity (which is why i started with per-window) but i think realistically nobody will need it and a per-app setting, which iirc was easier on the eyes in the codebase, covers it all. But def open for discussion (aka tell me what to do😊 ) on that one.

tooling/api/src/tauri.ts
amrbashir: right, my bad, the check should be `if (http !== undefined)`

amrbashir:
I think per-window is better to align with wry implementation and give user flexibility.

lucasfernog:
I don't really see the need to have the complexity of the per-window configuration. I would go with the per-app option, and leave it open to be requested by the community.

If we want to be extremely configurable, it would even need to be per-protocol instead.

tweidinger:
I agree with @FabianLars and @amrbashir that having it configurable for specific windows would reduce the impact with multi window applications.
I can imagine that this `http` scheme use case is not that common but we should consider per-window setting to be in line with upcoming allowlist changes and increased modularity with plugins.

Don't see this as a blocker and the current implementation looks good enough for a first iteration to me 👍

xuchaoqian:
We use localhost-plugin to get http worked, but some users can't open our app at all, because of some security policies or other reasons.

This PR comes to the rescue. The current implementation is good enough for us and most use cases. 

And we can think of per-app setting as a global setting, and per-window setting overrides the per-app setting.  So the current design will still work in the future, IMHO.

FabianLars:
Soooo, i personally kinda changed my mind here (but also not because it's already a per-app impl) and i think we should not introduce it in v2 at all and have tihs PR's implementation be temporary so per-app would be enough. The reasoning behind not adding it to v2 are:

1. I don't think there's anything that works on a https scheme but not on a http scheme, and _if_ there is it should be broken on the tauri:// scheme too.
2. Once we officially drop windows 7 support we can switch to `tauri://` on windows too and we always wanted that, right?

Instead we should have what i talked about _somewhere_ before: a _good_ resource requested api users can use to block all http requests or whatever they wanna block.

lucasfernog-crabnebula:
Sounds good to me @FabianLars

.changes/windows-http-scheme.md
```diff
@@ -0,0 +1,5 @@
+---
+'tauri': 'patch:enhance'
```
lucasfernog-crabnebula, line 2:
```suggestion
'tauri': 'patch:enhance'
'tauri-runtime': 'patch:enhance'
'tauri-runtime-wry': 'patch:enhance'
'tauri-utils': 'patch:enhance'
```

lucasfernog (approved):

amrbashir (approved):

Dirreke:
It seems that tauri v1.5 has been merged into v2. But how can I use this feature in v2?

FabianLars:
@Dirreke The http scheme is now the default in v2. We didn't add a config for it yet because we figured that if http:// doesn't work, tauri:// (used on macos and linux) won't work either. We're not completely opposed to adding it back though but didn't want to do it without it having valid use-cases. Sooo if you do have a requirement for it, please open a seperate issue, thanks :)

PR #7691, opened 2023-08-25
fix: MacOS Frameworks not signed, closes #7690 and #7710

### LINKED ISSUE #7690: [bug] MacOS Codesign doesn't handle frameworks
### Describe the bug

I'm developing an application on MacOS that uses a 3rd Party Framework that I package with my application. When code signing and then distributing the app, I get crash reports on boot regarding the Framework and it having discrepancies in the team id 

```
Not valid for use in process: mapping process and mapped file (non-platform) have different Team IDs
```

During my testing and research today, I found that Apple recommends signing included frameworks along with your app itself. So I tried this on the `.app` file, first signing the internal Framework and then signing the overall App itself and it cleared up the issue.

I'm guessing this is something that needs to be added to the bundler?

### Reproduction

_No response_

### Expected behavior

_No response_

### Platform and versions

```text
[✔] Environment
    - OS: Mac OS 13.2.1 X64
    - node: 18.15.0
    - yarn: 1.22.19
    - npm: 9.5.0

[-] Packages
    - tauri [RUST]: 1.4.1 (no manifest)
    - tauri-build [RUST]: 1.4.0 (no manifest)
    - wry [RUST]: 0.24.3 (no manifest)
    - tao [RUST]: 0.16.2 (no manifest)

[-] App
    - build-type: bundle
    - CSP: style-src 'unsafe-inline' 'self'; default-src 'self'; img-src 'self' asset: https://asset.localhost; connect-src 'self' asset: https://asset.localhost
    - distDir: ../dist
    - devPath: http://localhost:3010/
    - framework: React
    - bundler: Vite
```

### Stack trace

_No response_

### Additional context

_No response_

tr3ysmith:

- [X] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] No (however it does fix an issue I found with notarytool, which now seems to require a team-id. I added a new env variable called APPLE_TEAM_ID for notarizing

### Checklist

### DIFF
--- a/core/tauri-build/src/lib.rs
+++ b/core/tauri-build/src/lib.rs
@@ -370,6 +370,17 @@ pub fn try_build(attributes: Attributes) -> Result<()> {
   }
 
   if target_triple.contains("darwin") {
+    // If we have frameworks, we need to set the @rpath
+    // https://github.com/tauri-apps/tauri/issues/7710
+    let frameworks = &config.tauri.bundle.macos.frameworks;
+    if frameworks
+      .as_ref()
+      .and_then(|f| Some(!f.is_empty()))
+      .is_some()
+    {
+      println!("cargo:rustc-link-arg=-Wl,-rpath,@executable_path/../Frameworks");
+    }
+
     if let Some(version) = &config.tauri.bundle.macos.minimum_system_version {
       println!("cargo:rustc-env=MACOSX_DEPLOYMENT_TARGET={version}");
     }
--- a/tooling/bundler/src/bundle/macos/sign.rs
+++ b/tooling/bundler/src/bundle/macos/sign.rs
@@ -14,7 +14,8 @@ use std::{
 
 use crate::{bundle::common::CommandExt, Settings};
 use anyhow::Context;
-use log::info;
+
+use log::{error, info};
 use serde::Deserialize;
 
 const KEYCHAIN_ID: &str = "tauri-build.keychain";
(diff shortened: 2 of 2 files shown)


tr3ysmith:
https://github.com/tauri-apps/tauri/issues/7690

tr3ysmith:
So this morning, I came in to build my app, and it failed signing... So I'm looking into this further, I've also submitted a support ticket with Apple to get more clarity

tr3ysmith:
Linking in this article, looks like Apple codesign can fail if extra attributes exist on files inside the bundle....
https://developer.apple.com/library/archive/qa/qa1940/_index.html

Wondering if we should add the command to remove all extra attributes on the app bundle before signing?

tr3ysmith:
Added a function to strip extra attributes and it seems to have fixed an error I saw this morning

tr3ysmith:
**FROM APPLE**

When embedding frameworks there are two important things to note:

* You must copy the framework in a way that preserves symlinks.  I often see problems where folks copy a framework using `cp -r`, which expands the symlinks, which effectively breaks the code.  I generally recommend copying code with `ditto`.

* The notary service requires that all code be signed with a valid Developer ID, so it’s possible to build an app from components signed by different developers.  My general advice, however, is that you re-sign any nested code with your Developer ID.  Sign each code item separately, from the inside out.

tr3ysmith:
#7710

tr3ysmith:
So in testing, I think this is actually all good to go for now. I think we need to ultimately address assigning specific entitlement.plist files to each included binary in the app, both sidecar and frameworks. Apple recommends not using the same plist file for everything as it could break stuff. We should probably also look at removing the --deep argument on the codesign tool in the future and manually signing stuff

tr3ysmith:
@FabianLars if this pull request gets approved, we'll need to add APPLE_TEAM_ID as an argument to the GitHub action once its released, since it seems like notarytool requires it now.

tr3ysmith:
@FabianLars I've added signing for sidecar binaries, but I haven't been able to fully test it as I don't have a sidecar project setup. I've also removed --deep in this latest commit

lucasfernog:
@tr3ysmith I see you added a fix for 3 separate issues in this PR (and you also added a breaking change with the `APPLE_TEAM_ID` environment variable). Can you break it into 3 different PRs? it'll also help our auditing team.

Thank you :heart:

lucasfernog:
> @FabianLars if this pull request gets approved, we'll need to add APPLE_TEAM_ID as an argument to the GitHub action once its released, since it seems like notarytool requires it now.

From my tests the team id isn't required. We also already merged notarytool usage into 1.x.

tr3ysmith:
@lucasfernog ah sounds good, so I can adjust this to allow team id but not require it.

I'll close this PR and open 3 separate ones.

tr3ysmith:
#7773 - add rpath
#7774 - fix codesigning
#7775 - Apple Team ID

PR #7709, opened 2023-08-29
feat: add tray and menu javascript APIs and events, closes #6617

### LINKED ISSUE #6617: [feat] JS window menu API
### Describe the problem

Window menu can't be set from JS.

A JS window menu API would allow setting window menu from JS.

### Alternatives considered

None

### Additional context

None

---
Thanks

amrbashir:
closes #6617

TODOs:
- [x] JS docs
 

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


amrbashir:
Now that the implementation is done without depending on the `window` plugin, I still believe the `window` plugin (plus some others) should be added back to core, also notice that the `tray` plugin is dependent on the `menu` plugin so both will need to be in coupled together.

Also the `menu` plugin ideally should use the `Window` and `Position` types from the `window` plugin and the `window` plugin should depend on the `menu` plugin so that we can provide a window config option to use a specific menu when creating the window.

All this could be solved by just adding them all into one single package in core. These are essential to any tauri app and so should be provided by default in core.

core/tauri-macros/src/lib.rs
```diff
@@ -89,3 +90,64 @@ pub fn default_runtime(attributes: TokenStream, input: TokenStream) -> TokenStre
   let input = parse_macro_input!(input as DeriveInput);
   runtime::default_runtime(attributes, input).into()
 }
+
+/// Accepets a closure-like syntax to call arbitray code on a menu item
```
lucasfernog:
```suggestion
/// Accepts a closure-like syntax to call arbitrary code on a menu item
```

core/tauri-macros/src/lib.rs
```diff
@@ -89,3 +90,64 @@ pub fn default_runtime(attributes: TokenStream, input: TokenStream) -> TokenStre
...
+/// after mathcing against `kind` and retreiveing it from `resources_table` using `rid`.
```
lucasfernog:
```suggestion
/// after matching against `kind` and retrieving it from `resources_table` using `rid`.
```

core/tauri-macros/src/lib.rs
```diff
@@ -89,3 +90,64 @@ pub fn default_runtime(attributes: TokenStream, input: TokenStream) -> TokenStre
...
+///
+/// You can optionally pass a third paraemeter to select which item kinds
```
lucasfernog:
```suggestion
/// You can optionally pass a third parameter to select which item kinds
```

core/tauri-macros/src/lib.rs
```diff
@@ -89,3 +90,64 @@ pub fn default_runtime(attributes: TokenStream, input: TokenStream) -> TokenStre
...
+/// to match agains, by providing a `|` separated list of item kinds
```
lucasfernog:
```suggestion
/// to match against, by providing a `|` separated list of item kinds
```

core/tauri-macros/src/menu.rs
```diff
@@ -0,0 +1,104 @@
...
+  if kinds.is_empty() {
+    kinds.extend(defaults.clone());
+  }
+
+  let mut has_negated: bool = false;
+  for NegatedIdent(negated, _) in &kinds {
+    if *negated && !has_negated {
+      has_negated = true;
+    }
+  }
```
lucasfernog:
```suggestion
  let has_negated = kinds.iter().any(|n| n.0);
```

core/tauri-macros/src/menu.rs
```diff
@@ -0,0 +1,104 @@
...
+    .filter_map(|nident| match nident.1 {
+      i if i == "MenuItem" && !nident.0 => Some((i, Ident::new("MenuItem", Span::call_site()))),
+      i if i == "Submenu" && !nident.0 => Some((i, Ident::new("Submenu", Span::call_site()))),
+      i if i == "Predefined" && !nident.0 => {
+        Some((i, Ident::new("PredefinedMenuItem", Span::call_site())))
+      }
+      i if i == "Check" && !nident.0 => Some((i, Ident::new("CheckMenuItem", Span::call_site()))),
+      i if i == "Icon" && !nident.0 => Some((i, Ident::new("IconMenuItem", Span::call_site()))),
+      _ => None,
+    })
```
lucasfernog:
```suggestion
    .filter_map(|nident| {
      if nident.0 {
        None
      } else {
        match nident.1 {
          i if i == "MenuItem" => Some((i, Ident::new("MenuItem", Span::call_site()))),
          i if i == "Submenu" => Some((i, Ident::new("Submenu", Span::call_site()))),
          i if i == "Predefined" => Some((i, Ident::new("PredefinedMenuItem", Span::call_site()))),
          i if i == "Check" => Some((i, Ident::new("CheckMenuItem", Span::call_site()))),
          i if i == "Icon" => Some((i, Ident::new("IconMenuItem", Span::call_site()))),
          _ => None,
        }
      }
    })
```

core/tauri-macros/src/menu.rs
```diff
@@ -0,0 +1,104 @@
...
+    .unzip();
+
+  quote! {
+    match kind {
```
lucasfernog: I don't like how it assumes `kind`, `rid` and `resources_table` exists instead of taking them as arguments.. but this is internal so it's ok.

core/tauri/src/resources/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+use std::{
+  any::{type_name, Any, TypeId},
+  borrow::Cow,
+  collections::BTreeMap,
+  sync::Arc,
+};
+
+/// Resources are Rust objects that are stored in [ResourceTable] and managed by tauri.
+/// They are identified in JS by a numeric ID (the resource ID, or rid).
+/// Resources can be created in endpoints. Resources can also be retrieved in endpoints by
```
lucasfernog:
```suggestion
/// Resources can be created in commands. Resources can also be retrieved in commands by
```

core/tauri/src/resources/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+  fn is<T: Resource>(&self) -> bool {
+    self.type_id() == TypeId::of::<T>()
+  }
+
+  #[inline(always)]
+  #[allow(clippy::needless_lifetimes)]
+  pub(crate) fn downcast_arc<'a, T: Resource>(self: &'a Arc<Self>) -> Option<&'a Arc<T>> {
+    if self.is::<T>() {
+      let ptr = self as *const Arc<_> as *const Arc<T>;
+      // TODO(piscisaureus): safety comment
```
lucasfernog: let's resolve this :)

core/tauri/src/resources/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+  pub(crate) fn add<T: Resource>(&mut self, resource: T) -> ResourceId {
+    self.add_arc(Arc::new(resource))
+  }
+
+  /// Inserts a `Arc`-wrapped resource into the resource table.
+  ///
+  /// The resource type is erased at runtime and must be statically known
+  /// when retrieving it through `get()`.
+  ///
+  /// Returns a unique resource ID, which acts as a key for this resource.
```
lucasfernog, line 98:
```suggestion
  /// Returns an unique resource ID, which acts as a key for this resource.
```

core/tauri/src/resources/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+    let removed_resource = self.index.insert(rid, resource);
+    // TODO: add replace method if needed
+    assert!(removed_resource.is_none());
+    self.next_rid += 1;
+    rid
+  }
+
+  /// Returns a reference counted pointer to the resource of type `T` with the
+  /// given `rid`. If `rid` is not present or has a type different than `T`,
+  /// this function returns `None`.
```
lucasfernog:
```suggestion
  /// this function returns [`Error::BadResourceId`].
```

core/tauri/src/resources/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+  // /// If a resource with the given `rid` exists but its type does not match `T`,
+  // /// it is not removed from the resource table. Note that the resource's
+  // /// `close()` method is *not* called.
+  // ///
+  // /// Also note that there might be a case where
+  // /// the returned `Arc<T>` is referenced by other variables. That is, we cannot
+  // /// assume that `Arc::strong_count(&returned_arc)` is always equal to 1 on success.
+  // /// In particular, be really careful when you want to extract the inner value of
+  // /// type `T` from `Arc<T>`.
+  // pub(crate) fn take<T: Resource>(&mut self, rid: ResourceId) -> Result<Arc<T>, Error> {
```
lucasfernog: should we delete this?

core/tauri/src/resources/plugin.rs
```diff
@@ -0,0 +1,22 @@
...
+use crate::{
+  command,
+  plugin::{Builder, TauriPlugin},
+  AppHandle, Runtime,
+};
+
+use super::ResourceId;
+
+#[command(root = "crate")]
+fn close<R: Runtime>(app: AppHandle<R>, rid: ResourceId) -> crate::Result<()> {
```
lucasfernog, line 14: this is problematic, an attacker can close all resources by iterating on IDs until an error is returned.

maybe we should make the ID random instead?

core/tauri/src/menu/plugin.rs
```diff
@@ -0,0 +1,637 @@
+// Copyright 2019-2023 Tauri Programme within The Commons Conservancy
```
lucasfernog, line 1: this file is kinda big, can we move the commands to their own module?

core/tauri/src/tray/plugin.rs
```diff
@@ -0,0 +1,202 @@
...
+  options: TrayIconOptions,
+) -> crate::Result<(ResourceId, String)> {
+  let mut builder = if let Some(id) = options.id {
+    TrayIconBuilder::<R>::with_id(id)
+  } else {
+    TrayIconBuilder::<R>::new()
+  };
+
+  builder = builder.on_tray_event(|tray, e| {
+    let _ = tray.app_handle().emit_all("tauri://tray", e);
```
lucasfernog: I think we should use a channel here instead, so we know only the tray icon caller have access to these events

also simplifies the JS APi a lot :)

core/tauri/src/menu/plugin.rs
```diff
@@ -0,0 +1,637 @@
...
+  let resources_table = app.manager.resources_table();
+  let icon_item = resources_table.get::<IconMenuItem<R>>(rid)?;
+  icon_item.set_native_icon(icon)
+}
+
+pub(crate) fn init<R: Runtime>() -> TauriPlugin<R> {
+  Builder::new("menu")
+    .on_event(|app, e| {
+      if let RunEvent::MenuEvent(e) = e {
+        let _ = app.emit_all("tauri://menu", e);
```
lucasfernog: consider using a channel instead (maybe have a dedicated list of channels in the resources table)

lucasfernog (changes requested):

core/tauri-macros/src/menu.rs
amrbashir: I srsly don't know what I was thinking writing that code, just kill me already 😅

core/tauri/src/resources/mod.rs
amrbashir: I left it there in case we ever need but I guess we can re-implement it

core/tauri-macros/src/menu.rs
amrbashir: yeah I don't like it either, maybe I need to spend some time thinking about improving it without adding much noise in the call site

core/tauri/src/resources/plugin.rs
amrbashir: Closing resource is not that bad right?

Also for reference, deno is using the same ID strategy and though they also provide `.close()` you can't really invoke the underlying mechanism of it (compared to `invoke("plugin:resources|close"`). 

So in order to simplify the implementation and keep using incrementing ID numbers, we can hide the invoke endpoint and randomize it on each launch of the app.

core/tauri/src/menu/plugin.rs
amrbashir: the commands make most of the file, so moving them into another module will still be a big file

amrbashir:
I think this PR needs to wait until the refactors PR are merged in.

core/tauri/src/resources/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+  pub(crate) fn add<T: Resource>(&mut self, resource: T) -> ResourceId {
+    self.add_arc(Arc::new(resource))
+  }
+
+  /// Inserts a `Arc`-wrapped resource into the resource table.
+  ///
+  /// The resource type is erased at runtime and must be statically known
+  /// when retrieving it through `get()`.
+  ///
+  /// Returns an unique resource ID, which acts as a key for this resource.
```
FabianLars: idk why the github notification showed me a single change (`a` to `an`) and literally nothing else, but that's the only reason i noticed it i guess. `a unique` was correct because it starts with a `yu` sound.
And people say german is complicated... 😡 
```suggestion
  /// Returns a unique resource ID, which acts as a key for this resource.
```

lucasfernog:
We're missing a lot of documentation here, but I wanna merge it anyway because we need to do a complete pass on API examples.

I like how this can be dynamic with the MenuItem classes (so you can update, share items), or static with raw objects after https://github.com/tauri-apps/tauri/pull/7709/commits/d0eeaba4198c3b04b5f252bd87f5ff87b3359fe2

lucasfernog (approved):

PR #7722, opened 2023-08-30
fix(windows): respect `focused: false` for webview, closes #7519

### LINKED ISSUE #7519: [bug] Windows do not respect "focus" property
### Describe the bug

Hi all,

First off, just wanted to say that I'm really enjoying using Tauri and have had great feedback from users so far!

I seem to be having an issue with focus behaviour on all windows on the Windows platform. I've noticed it in quite a few different scenarios, and my app's users are also seeing the same behaviour, so thought I would report it!

It seems that because the default focus of all windows is set to `true`, this will always either override a `"focus"` setting of `false` in newly created windows at runtime, or will be just be ignored.

Additionally, I've found through testing that the focus state of new windows appears to be linked to the initial focus state of the main window, which can be seen in the linked reproduction.

However, I've noticed caveats to this behaviour - if running `npm run tauri dev` from Command Prompt and unfocusing the console window before the app starts, the app window will follow the focus state of the Command Prompt window - i.e. if it is unfocused before starting, the main app window will be unfocused too. Conversely, if the Command Prompt window is focused, the main window will then be focused.

This causes an issue for my app in particular. I have a background process that triggers new "notification" windows to be spawned on an event, and these windows are meant to overlay a currently running game. But upon spawning, will take focus away from the game to focus themselves, which is obviously not intended behaviour, as these "notification" windows should never be intended to focus in the first place with `"focus": false` set in the Window Options.

I believe a fair workaround to this issue would be to reset the default focus state of all windows to be `false` instead of `true`, as windows have methods to focus themselves (via `setFocus`), but no methods to unfocus them from what I have seen (although I'm happy to be wrong if there is!). If this is not possible, a method to unfocus the window before spawning would be massively beneficial, or to incorporate the `focusable` property from Electron as a new Window Option, which specifies that the spawned window may never take focus if set to `false`.

### Reproduction

I've created a [reproduction](https://github.com/Jackson0ne/tauri-focus-bug/tree/main) that demonstrates at least some of these issues.

If the first window is initially unfocused before the second window appears, the second window's focus property will be reported as `false` in the DevTools Console. But if the first window **is** focused when the second window is spawned, its focus will then be reported as `true`.

### Expected behavior

The `"focus"` property in both `tauri.conf.json` and via window creation in Rust and JS will be followed and applied independently to any other window's focus state.

### Platform and versions

```text
[✔] Environment
    - OS: Windows 10.0.19045 X64
    ✔ WebView2: 115.0.1901.183
    ✔ MSVC:
        - Visual Studio Community 2017
    ✔ rustc: 1.66.1 (90743e729 2023-01-10)
    ✔ Cargo: 1.66.1 (ad779e08b 2023-01-10)
    - node: 14.18.1
    - npm: 6.14.15

[-] Packages
    - tauri [RUST]: 1.4.1
    - tauri-build [RUST]: 1.4.0
    - wry [RUST]: 0.24.3
    - tao [RUST]: 0.16.2

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../src
    - devPath: ../src
```

### Stack trace

_No response_

### Additional context

_No response_

amrbashir:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/core/tauri-runtime-wry/Cargo.toml
+++ b/core/tauri-runtime-wry/Cargo.toml
@@ -13,7 +13,7 @@ exclude = [ "CHANGELOG.md", "/target" ]
 readme = "README.md"
 
 [dependencies]
-wry = { version = "0.24.1", default-features = false, features = [ "file-drop", "protocol" ] }
+wry = { git = "https://github.com/tauri-apps/wry", branch = "v0.24.1", default-features = false, features = [ "file-drop", "protocol" ] }
 tauri-runtime = { version = "0.14.0", path = "../tauri-runtime" }
 tauri-utils = { version = "1.4.0", path = "../tauri-utils" }
 uuid = { version = "1", features = [ "v4" ] }
--- a/core/tauri-runtime-wry/Cargo.toml
+++ b/core/tauri-runtime-wry/Cargo.toml
@@ -13,7 +13,7 @@ exclude = [ "CHANGELOG.md", "/target" ]
 readme = "README.md"
 
 [dependencies]
-wry = { git = "https://github.com/tauri-apps/wry", branch = "v0.24.1", default-features = false, features = [ "file-drop", "protocol" ] }
+wry = { version = "0.24.4", default-features = false, features = [ "file-drop", "protocol" ] }
 tauri-runtime = { version = "0.14.0", path = "../tauri-runtime" }
 tauri-utils = { version = "1.4.0", path = "../tauri-utils" }
 uuid = { version = "1", features = [ "v4" ] }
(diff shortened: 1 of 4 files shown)


core/tauri-runtime-wry/Cargo.toml
```diff
@@ -13,7 +13,7 @@ exclude = [ "CHANGELOG.md", "/target" ]
 readme = "README.md"
 
 [dependencies]
-wry = { version = "0.24.1", default-features = false, features = [ "file-drop", "protocol" ] }
+wry = { git = "https://github.com/tauri-apps/wry", branch = "v0.24.1", default-features = false, features = [ "file-drop", "protocol" ] }
```
amrbashir: TODO: switch to published version when WRY is published and before this PR is merged

lucasfernog (approved):

PR #7751, opened 2023-09-04
fix(core): use postMessage IPC for remote URLs on macOS closes #7662

### LINKED ISSUE #7662: [bug] refactor of IPC using URI schemes is breaking some usecase
### Describe the bug

Nice to see an improvement on performances with the new IPC mechanism of #7170 .

It is obviously better to have data exchanged between the frontend and tauri, as body of POST requests instead of JSON in URLparam of a GET request. 

But we shall note that this improvement is limited on platforms that are not android nor linux.

For me, it breaks a special use-case that I am having. I am loading a remote page (https) in a popup window from the tauri app, using the window API. In the popup window (not main) the javascript that is loaded contains the `@tauri-apps/api` and `@tauri-apps/plugin-window` modules, and makes use of the `Window.getByLabel("main").emit()` API call to send a message to the main window.

Until the breaking changes of `2.0.0-alpha.11`, this was working fine, except on android, because of the way messages are passed on that platform. 

Since the new IPC mechanism that uses `ipc://` fetch, I get a security exception on the client side of the popup window.

No matter how much CSP header I injected on that page (via the HTTP server serving it), the error remains the same, as the page has an origin with `https://` and `ipc://` is considered insecure no matter what the CSP is saying.

So my `https` page cannot send messages back to the main tauri window anymore.

I guess this is a rare usecase, but I do make use of `ipc_scope().configure_remote_access()` API in the configuration of the tauri app, and I suppose that this API was meant specifically for this kind of use case, when a webpage making use of the client API is not served by tauri but by a remote server.

So it seems that the new IPC mechanism is breaking any possible usage of `ipc_scope().configure_remote_access()`.

I first decided to freeze the tauri version i am using to `2.0.0-alpha.10` and to cherrypick some bug fixes that appeared in the latest version, and that I needed. 

But then I thought this would not be future proof, so I decided to introduce a feature in tauri (in a fork of it, obviously) that would deactivate the new IPC mechanism and fall back to the old way using postMessage.

This crate feature that i named `no-ipc-custom-protocol` is quite minimalistic and seems to work on every platform i tested (all of them, except iOS that i don't have a dev env for, yet)

Would you consider introducing this feature in the next alpha release ? 

Also I would really like if @lucasfernog, the author of the refactor, could find another way to circumvent this limitation and allow `ipc://` to be fetched on any `https://` page, but I am not aware of any solution for that. Another way might be to dynamically fall back to the older IPC mechanism when the calling webpage is from a remote origin, and let the new mechanism work on webpages that are served by tauri. That would allow the `ipc_scope().configure_remote_access()` to make sense again in the new IPC mechanism context.

The `no-ipc-custom-protocol` feature that I implemented can be found here :
https://git.nextgraph.org/NextGraph/tauri/commit/ec9b62b3379f5353c6d8b9484e12ec82ef4fa34f

Let me know if you want a merge request.

And thanks for your amazing work!

### Reproduction

unfortunately I dont have enough time left for this issue to create a minimalist project that would reproduce this issue. Also, it would be a bit hard to setup as you need an external https server...
But if you have some work on a branch that you want me to test, I will do it happily.

### Expected behavior

See description of bug

### Platform and versions

```text
[✔] Environment
    - OS: Mac OS 10.15.6 X64
    ✔ rustc: 1.69.0 (84c898d65 2023-04-16)
    ✔ Cargo: 1.69.0 (6e9a83356 2023-04-12)
    ✔ Rust toolchain: stable-x86_64-apple-darwin (environment override by RUSTUP_TOOLCHAIN)
    - node: 16.13.1
    - pnpm: 7.15.0
    - yarn: 1.22.18
    - npm: 8.1.2

[-] Packages
    - tauri [RUST]: (2.0.0-alpha.11)
    - tauri-build [RUST]: no manifest (2.0.0-alpha.6, 2.0.0-alpha.7)
    - wry [RUST]: 0.31.1
    - tao [RUST]: 0.22.2
    - @tauri-apps/api [NPM]: 2.0.0-alpha.6
    - @tauri-apps/cli [NPM]: 2.0.0-alpha.10

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../dist
    - framework: Svelte
    - bundler: Vite
```

### Stack trace

_No response_

### Additional context

_No response_

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/core/tauri/scripts/ipc-protocol.js
+++ b/core/tauri/scripts/ipc-protocol.js
@@ -18,8 +18,6 @@
         options
       } = message
 
-      console.log(osName, location.protocol)
-
       // use custom protocol for IPC if:
       // - the flag is set to true or
       // - the command is the fetch data command or
--- a/core/tauri/scripts/ipc-protocol.js
+++ b/core/tauri/scripts/ipc-protocol.js
@@ -10,11 +10,35 @@
 
   Object.defineProperty(window, '__TAURI_POST_MESSAGE__', {
     value: (message) => {
-      const { cmd, callback, error, payload, options } = message
+      const {
+        cmd,
+        callback,
+        error,
+        payload,
+        options
+      } = message
 
-      // use custom protocol for IPC if the flag is set to true, the command is the fetch data command or when not on Linux/Android
-      if (useCustomProtocol || cmd === fetchChannelDataCommand || (osName !== 'linux' && osName !== 'android')) {
-        const { contentType, data } = processIpcMessage(payload)
+      console.log(osName, location.protocol)
+
+      // use custom protocol for IPC if:
+      // - the flag is set to true or
+      // - the command is the fetch data command or
+      // - when not on Linux/Android
+      // AND
+      // - when not on macOS with an https URL
+      if (
+        (
+          useCustomProtocol ||
+          cmd === fetchChannelDataCommand ||
+          !(osName === 'linux' || osName === 'android')
+        ) &&
+        !(osName === 'macos' && location.protocol === 'https:')
+      ) {
+        console.log('process')
+        const {
+          contentType,
+          data
+        } = processIpcMessage(payload)
         fetch(window.__TAURI__.convertFileSrc(cmd, 'ipc'), {
           method: 'POST',
           body: data,
@@ -44,7 +68,15 @@
         })
       } else {
         // otherwise use the postMessage interface
-        const { data } = processIpcMessage({ cmd, callback, error, options, payload })
+        const {
+          data
+        } = processIpcMessage({
+          cmd,
+          callback,
+          error,
+          options,
+          payload
+        })
         window.ipc.postMessage(data)
       }
     }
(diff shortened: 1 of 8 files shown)


core/tauri/scripts/ipc-protocol.js
```diff
@@ -10,11 +10,35 @@
...
+        error,
+        payload,
+        options
+      } = message
 
-      // use custom protocol for IPC if the flag is set to true, the command is the fetch data command or when not on Linux/Android
-      if (useCustomProtocol || cmd === fetchChannelDataCommand || (osName !== 'linux' && osName !== 'android')) {
-        const { contentType, data } = processIpcMessage(payload)
+      console.log(osName, location.protocol)
+
```
amrbashir:
```suggestion
```

amrbashir (approved):

PR #7747, opened 2023-09-04
Resolver 1.x 

martinfrances107:
This PR resolves  a warning seen while running cargo update

warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest

PR #7743, opened 2023-09-04
chore: fixed resolver issue

martinfrances107:
This PR resolves  a warning seen while running cargo update

warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest

### What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

### Does this PR introduce a breaking change?

- [ ] Yes, and the changes were approved in issue #___
- [x] No

### Checklist
- [x] When resolving issues, they are referenced in the PR's title (e.g `fix: remove a typo, closes #___, #___`)
- [x] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md).
- [x] I have added a convincing reason for adding this feature, if necessary

PR #7753, opened 2023-09-05
chore: resolves a warning seen while running cargo update

martinfrances107:
This my third attempt to generate a PR with a "verified" signature.
Sorry for the noise, and thanks for all the technical support/advice 
#7747 
#7743 

I can see the "Verified" badge on this PR. Anyway back to the reason for the PR

... 

This PR resolves a warning seen while running cargo update

warning: some crates are on edition 2021 which defaults to resolver = "2", but virtual workspaces default to resolver = "1"
note: to keep the current resolver, specify workspace.resolver = "1" in the workspace root's manifest
note: to use the edition 2021 resolver, specify workspace.resolver = "2" in the workspace root's manifest

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "1"
 members = [
   # core
   "core/tauri",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-resolver = "1"
+resolver = "2"
 members = [
   # core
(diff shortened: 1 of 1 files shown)


FabianLars:
Now it's verified, thanks :D 

about the change itself, i'm not sure what the others think but imo we should go for resolver 2 nowadays 🤔

martinfrances107:
Now using resolver 2

FabianLars (approved):
Thank you!

PR #7754, opened 2023-09-05
refactor(core): allow custom protocol handler to resolve async

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


.changes/custom-protocol-response-refactor.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri": patch:breaking
+---
+
+Changed `Builder::register_uri_scheme_protocol` to take a second parameter to resolve the request instead of returning a response object.
```
amrbashir: Could we provide both APIs like in wry?

.changes/http-drop-export.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri": patch:breaking
+---
+
+No longer re-export the `http` crate.
```
amrbashir: why remove the re-export?

core/tauri/src/ipc/channel.rs
```diff
@@ -78,7 +78,7 @@ impl Channel {
         .unwrap()
         .insert(data_id, body);
       window.eval(&format!(
-        "__TAURI_INVOKE__('{FETCH_CHANNEL_DATA_COMMAND}', null, {{ headers: {{ '{CHANNEL_ID_HEADER_NAME}': {data_id} }} }}).then(window['_' + {}]).catch(console.error)",
+        "__TAURI_INVOKE__('{FETCH_CHANNEL_DATA_COMMAND}', null, {{ headers: {{ '{CHANNEL_ID_HEADER_NAME}': '{data_id}' }} }}).then(window['_' + {}]).catch(console.error)",
```
amrbashir, line 81: does this require a change file?

core/tauri/src/manager.rs
```diff
@@ -628,30 +627,35 @@ impl<R: Runtime> WindowManager<R> {
...
+      pending.register_uri_scheme_protocol("asset", move |request, responder| {
+        match crate::protocol::asset::get(request, asset_scope.clone(), window_origin.clone()) {
+          Ok(response) => responder(response),
+          Err(e) => responder(
+            http::Response::builder()
+              .status(http::StatusCode::BAD_REQUEST)
+              .body(e.to_string().as_bytes().to_vec().into())
+              .unwrap(),
+          ),
+        }
```
amrbashir: why not modify the `crate::protocol::asset::get` instead? to keep consistency with the 2 protocols above it

core/tauri/src/manager.rs
```diff
@@ -663,13 +667,13 @@ impl<R: Runtime> WindowManager<R> {
...
+
       let assets = assets.clone();
-      let schema_ = schema.clone();
-      let url_base = format!("{schema_}://localhost");
       let aes_gcm_key = *crypto_keys.aes_gcm().raw();
 
-      pending.register_uri_scheme_protocol(schema, move |request| {
-        match request_to_path(request, &url_base).as_str() {
+      pending.register_uri_scheme_protocol(schema, move |request, responder| {
+        let response = match request_to_path(&request).as_str() {
```
amrbashir: we should probably move this to `crate::protocol::isolation_pattern`

core/tauri/src/menu/icon.rs
```diff
@@ -207,7 +207,7 @@ impl<R: Runtime> IconMenuItem<R> {
   /// - **Windows / Linux**: Unsupported.
   pub fn set_native_icon(&mut self, _icon: Option<NativeIcon>) -> crate::Result<()> {
     #[cfg(target_os = "macos")]
-    return run_main_thread!(self, |mut self_: Self| self_
+    return run_main_thread!(self, |#[allow(unused_mut)] mut self_: Self| self_
```
amrbashir: `set_native_icon` should not take `&mut self`, I fixed it in https://github.com/tauri-apps/tauri/pull/7709/files#diff-41ef79e158d69fe78154d145fa88e594ef8a3daf35e727db8ad9fa73cc23a663R211-R213, so either revert this or make the same changes to avoid conflicts in the future

core/tauri/src/window.rs
```diff
@@ -2067,7 +2069,7 @@ impl<R: Runtime> Window<R> {
   }
 
   /// Handles this window receiving an [`InvokeRequest`].
-  pub fn on_message(self, request: InvokeRequest) -> Receiver<InvokeResponse> {
+  pub fn on_message(self, request: InvokeRequest, responder: Box<OwnedInvokeResponder<R>>) {
```
amrbashir, line 2072: change file also?

amrbashir (changes requested):

.changes/http-drop-export.md
lucasfernog: the http crate is not stable, so we actually can't re-export it (if we're talking semver strict).

amrbashir: how so? last release was from last February and it has a minor version that we can use, if they updated the minor or major, we can choose not to upgrade. It is as stable as wry/tao imo.

lucasfernog: though we all know it can't have a breaking change without the entire ecosystem major bumping though i guess i'll revert this

amrbashir (approved):

Issue #7480, opened 2023-07-22
[bug] Symlinks inside updater bundle are not preserved on update


Brendonovich:
### Describe the bug

While #3933 describes an issue with creating updater bundles that contain symlinks, this is a slightly different problem.
We bundle FFMpeg as a macOS framework inside `Spacedrive.app`, and that framework contains symlinks to files within it.
When the updater extracts the downloaded update bundle, it doesn't preserve said symlinks, instead creating empty files.
This causes the app to instantly crash on launch as what should be symlinks are invaild files.

The top Finder window below is what the updater bundle should extract to. I downloaded the updater bundle from where it was being hosted (Firebase Storage) and extracted it, just to make sure it wasn't being affected by something other than the updater.
The bottom Finder window is the contents of the app bundle after running the updater. What were symlinks have become zero size files.
![image](https://github.com/tauri-apps/tauri/assets/14191578/3c5886bf-844e-4b91-881f-67103e84cc29)

### Reproduction

Will try to make a min repro, but it's a bit tricky since it requires having access to an update server

### Platform and versions

```text
[✔] Environment
    - OS: Mac OS 12.5.0 X64
    ✔ Xcode Command Line Tools: installed
    ✔ rustc: 1.70.0 (90c541806 2023-05-31)
    ✔ Cargo: 1.70.0 (ec8a8a0ca 2023-04-25)
    ✔ rustup: 1.25.1 (2022-07-12)
    ✔ Rust toolchain: stable-aarch64-apple-darwin (default)
    - node: 18.12.1
    - pnpm: 8.6.1
    - yarn: 1.22.19
    - npm: 8.19.2

[-] Packages
    - tauri [RUST]: 1.3.0
    - tauri-build [RUST]: 1.3.0
    - wry [RUST]: 0.24.3
    - tao [RUST]: 0.16.2
    - @tauri-apps/api [NPM]: not installed!
    - @tauri-apps/cli [NPM]: 1.3.0 (outdated, latest: 1.4.0)

[-] App
    - build-type: bundle
    - CSP: default-src spacedrive: webkit-pdfjs-viewer: asset: https://asset.localhost blob: data: filesystem: ws: wss: http: https: tauri: 'unsafe-eval' 'unsafe-inline' 'self' img-src: 'self'
    - distDir: ../dist
    - devPath: http://localhost:8001/
    - framework: React
    - bundler: Vite
```

lucasfernog:
[...]

PR #7773, opened 2023-09-07
fix: rpath missing from app, closes #7710

### LINKED ISSUE #7710: [feat] Add rpath for Frameworks on MacOS
### Describe the problem

When adding frameworks for MacOS in the tauri.conf.json, the rpath is not automatically set during the build process

I would like for Tauri to automatically apply the rpath to the executable if needed:

```rust
println!("cargo:rustc-link-arg=-Wl,-rpath,@executable_path/../Frameworks");
```

### Alternatives considered

_No response_

### Additional context

_No response_

tr3ysmith:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ X] No

### Other information
#7710


Brendonovich:
Seems like this also closes #7480. Symlinks in the `Frameworks` folder weren't being preserved but using this PR they are.

afonsojramos (approved):

core/tauri-build/src/lib.rs
```diff
@@ -370,6 +370,17 @@ pub fn try_build(attributes: Attributes) -> Result<()> {
...
   if target_triple.contains("darwin") {
+    // If we have frameworks, we need to set the @rpath
+    // https://github.com/tauri-apps/tauri/issues/7710
+    let frameworks = &config.tauri.bundle.macos.frameworks;
+    if frameworks
+      .as_ref()
+      .and_then(|f| Some(!f.is_empty()))
+      .is_some()
+    {
+      println!("cargo:rustc-link-arg=-Wl,-rpath,@executable_path/../Frameworks");
```
lucasfernog: I assume this needs to be set to a different value on development, since we won't be running inside the .app bundle.

lucasfernog (changes requested):

lucasfernog:
Can you rebase your branch to sign the [`ef8a58d` ](https://github.com/tauri-apps/tauri/pull/7773/commits/ef8a58df7dd8ff5cdbcb34eefb59e7ab03566cb8) commit?

core/tauri-build/src/lib.rs
tr3ysmith: So in my specific case, on our project, our Framework we're including comes from another library crate, so in my build script I have a function that goes and fetches it and then puts it in the target/Frameworks folder if `debug_assertions` is true. This basically works because that folder path is automatically included in the search path of the binary. Maybe we just need to automatically do that similar to how the resources folder is done?

tr3ysmith:
> Can you rebase your branch to sign the [`ef8a58d` ](https://github.com/tauri-apps/tauri/pull/7773/commits/ef8a58df7dd8ff5cdbcb34eefb59e7ab03566cb8) commit?

Should be good now

core/tauri-build/src/lib.rs
lucasfernog: Yeah sounds good to me

tr3ysmith: @lucasfernog are you able to point me in the right direction on where this should be implemented? I'd be happy to do it, but I'm assuming its in a different spot?

tweidinger:
Only concerns from sec perspective would be if `@executable_path/../Frameworks` is (world/all users) writeable, which from my understanding is not the case for apps properly deployed in macos (development seems a little different) and the application loading the relative libraries is not a suid binary, which is very rarely the case for a tauri application.

I think this would need to be documented for MacOS apps as risks for privilege escalation in case some privileged calls/libs/permissions are used/called from the Tauri app otherwise LGTM

tr3ysmith:
> Only concerns from sec perspective would be if `@executable_path/../Frameworks` is (world/all users) writeable, which from my understanding is not the case for apps properly deployed in macos (development seems a little different) and the application loading the relative libraries is not a suid binary, which is very rarely the case for a tauri application.
> 
> 
> 
> I think this would need to be documented for MacOS apps as risks for privilege escalation in case some privileged calls/libs/permissions are used/called from the Tauri app otherwise LGTM

No it would not be writable, anything inside the app package once signed cannot be modified otherwise the code signing is invalid.

lucasfernog:
@tr3ysmith can you double check my latest commit and see if [`af13f20` (#7773)](https://github.com/tauri-apps/tauri/pull/7773/commits/af13f20965c6b6c8a0d72036b7494d3870c1c2bd) works fine on your setup?

core/tauri-build/src/lib.rs
lucasfernog: pushed it :)

tr3ysmith:
> @tr3ysmith can you double check my latest commit and see if [`af13f20` (#7773)](https://github.com/tauri-apps/tauri/pull/7773/commits/af13f20965c6b6c8a0d72036b7494d3870c1c2bd) works fine on your setup?

Yeah that seems to work great, I was able to remove my build script and it worked.

lucasfernog (approved):

PR #7779, opened 2023-09-07
refactor(core): custom protocol on Windows now uses the http scheme

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


amr-crabnebula (approved):

amrbashir (approved):

FabianLars:
Am i right to assume that we don't want this to be configurable? I'm asking because i'd lean towards closing https://github.com/tauri-apps/tauri/pull/7645 if that's the case.

lucasfernog:
I think for v1 it's valuable to add a config option for it (since it's a popular issue). For v2 I don't really want to bloat the code with this check, though it could be added if the community / security team feels it's a nice addition (though it sucks that it only helps windows/android).

FabianLars:
Yeah, it's just that adding a new but deprecated API has a bit of a weird vibe but alright, at least the PR wasn't wasted time then :D

And in v2 I think a better request API (basically on_navigation but for everything) would solve the problem better than the https scheme since it would be cross platform.

percy507:
Hi @lucasfernog,

Why Windows use `http` scheme now instead of `https`? I mean no harm. Just curious.

amrbashir:
allows for connecting to insecure protocols like `ws` instead of `wss` which has been the main reason for this change

PR #7813, opened 2023-09-11
refactor: remove webview2 focus listener events

amrbashir:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist


lucasfernog:
I don't really like this change. WIth the current code we get correct events, but with this change sometimes there's a glitch when I focus the window, where I get three different focused events:
```
WindowEvent { label: "main", event: Focused(true) }
MainEventsCleared
WindowEvent { label: "main", event: Focused(false) }
WindowEvent { label: "main", event: Focused(true) }
```

it's a little weird tbh

amrbashir:
Yeah but there is no way around it atm, other than keeping the current logic which is somewhat okay.

PR #7880, opened 2023-09-21
feat: add `beforeEachBundleCommand`

amrbashir:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/core/tauri-utils/src/config.rs
+++ b/core/tauri-utils/src/config.rs
@@ -1713,7 +1713,8 @@ pub struct BuildConfig {
   #[serde(alias = "before-bundle-command")]
   pub before_bundle_command: Option<HookCommand>,
   /// Similar to `before_bundle_command`, a shell command to run before the bundling phase
-  /// of each bundle format.
+  /// of each package type. Check `TAURI_BUNDLER_PACKAGE_TYPE` env var to know
+  /// which package type is being bundled.
   ///
   /// The TAURI_PLATFORM, TAURI_ARCH, TAURI_FAMILY, TAURI_PLATFORM_VERSION, TAURI_PLATFORM_TYPE and TAURI_DEBUG environment variables are set if you perform conditional compilation.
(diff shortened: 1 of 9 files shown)


FabianLars:
What's the use case for this? Or rather, what's the issue with the current beforeBundleCommand?

amrbashir:
This is allows for some conditional logic based on which package type is being bundled.

Also maybe we should also add `beforeEachTargetBuildCommand`? but I don't see a use-case for it immediately, and probably will only affect when building for Android as it has more than one target being built in the same invoke of `tauri android build`?

I don't really think this PR is urgent or needed that much so I am fine to just keep it open until someone requests it or close it.

FabianLars:
Ahh okay, my bad, i missed the `TAURI_BUNDLER_PACKAGE_TYPE` at the bottom (it's missing in the config description). Totally get it then 👍

amrbashir:
good catch, I will add it to the docs

core/tauri-utils/src/config.rs
```diff
@@ -1712,6 +1712,12 @@ pub struct BuildConfig {
   /// The TAURI_PLATFORM, TAURI_ARCH, TAURI_FAMILY, TAURI_PLATFORM_VERSION, TAURI_PLATFORM_TYPE and TAURI_DEBUG environment variables are set if you perform conditional compilation.
   #[serde(alias = "before-bundle-command")]
   pub before_bundle_command: Option<HookCommand>,
+  /// Similar to `before_bundle_command`, a shell command to run before the bundling phase
+  /// of each bundle format.
```
```suggestion
  /// of each package type. Check `TAURI_BUNDLER_PACKAGE_TYPE` env var to know
  /// which package type is being bundled.
```

amrbashir:
Closing until we see a use-case for this

PR #7888, opened 2023-09-22
Yes no cancel dialogbuttons, closes #7855

### LINKED ISSUE #7855: [feat] Add support for YesNoCancel button option on Message Dialog
### Describe the problem

Would love to see YesNoCancel option added to the Message Dialog API. This would allow for File Save Alerts on exiting the program.

Upgrade dependencies for RFD since they now support this in their crate:
https://github.com/PolyMeilex/rfd/blob/master/examples/message-custom-buttons/src/main.rs

Add two variants to MessageDialogButtons
```rust
MessageDialogButtons::YesNoCancel,
MessageDialogButtons::YesNoCancelWithLabels(yes_text, no_text, cancel_text)
```

Support this through the API just like the other button options

### Alternatives considered

_No response_

### Additional context

Unfortunately it looks like RFD is using gtk-sys 0.18 which currently conflicts with Tauri, so this isn't just a simple version bump of that and then adding this code. I've looked into this, at high level, it looks like the webkit2gtk crate needs to get version bumped on its various gtk deps to v0.18. And then Tauri's dependency for webkit2gtk gets upgraded. 

Starting this issue to track this, hopefully someone can lend a hand, I'll do what I can to assist!

tr3ysmith:
Adds YesNoCancel option for Dialogs so that things like exiting the software and asking if a user want's to "save, don't save, or cancel" can be implemented.

Issue: #7855

Also requires this PR to be completed first:
https://github.com/tauri-apps/webkit2gtk-rs/pull/149

- [ ] Bugfix
- [X] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] No

### Checklist

### DIFF
--- a/core/tauri/Cargo.toml
+++ b/core/tauri/Cargo.toml
@@ -88,11 +88,11 @@ encoding_rs = "0.8.31"
 sys-locale = { version = "0.2.3", optional = true }
 
 [target."cfg(any(target_os = \"macos\", windows, target_os = \"linux\", target_os = \"dragonfly\", target_os = \"freebsd\", target_os = \"openbsd\", target_os = \"netbsd\"))".dependencies]
-rfd = { version = "0.10", optional = true, features = [ "gtk3", "common-controls-v6" ] }
+rfd = { version = "0.12", optional = true, features = [ "gtk3", "common-controls-v6" ] }
 notify-rust = { version = "4.5", optional = true }
 
 [target."cfg(any(target_os = \"linux\", target_os = \"dragonfly\", target_os = \"freebsd\", target_os = \"openbsd\", target_os = \"netbsd\"))".dependencies]
-gtk = { version = "0.15", features = [ "v3_20" ] }
+gtk = { version = "0.18.1", features = [ "v3_20" ] }
 glib = "0.15"
 webkit2gtk = { version = "0.18.2", features = [ "v2_22" ] }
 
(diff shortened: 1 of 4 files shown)


amrbashir:
I believe this has been implemented in tauri-plugin-dialog for v2 and since we can't use gtk 0.18 in v1, I will go ahead and close this PR. Thanks for your contributions.

HumanWannabe:
@amrbashir I cannot find this anywhere in v2. Are you sure it was implemented? If not, could you maybe re-open it? 
It would be a shame for something so basic to be missed out by accident. 
If on the other hand it is indeed implemented and I missed it, could you please point me to the documentation for it?

amrbashir:
@HumanWannabe I believe this is what you are looking for https://docs.rs/tauri-plugin-dialog/latest/tauri_plugin_dialog/struct.MessageDialogBuilder.html#method.buttons and https://docs.rs/tauri-plugin-dialog/latest/tauri_plugin_dialog/enum.MessageDialogButtons.html#variant.OkCustom

HumanWannabe:
@amrbashir Thank you for your answer, but if I read this correct, it is not what the
issue was about. This allows the customization of the buttons, but it still
only allows for two buttons. The request was to allow for three, for
example "Yes, No, Cancel".

amrbashir:
I see, this is indeed not implemented, can you open an issue in plugins-worspace repo?

HumanWannabe:
yes, thank you

PR #7919, opened 2023-09-29
Fix strict mode: arguments -> args

birkskyum:
Currently, the cli breaks when running in strict mode, because a variable is named `arguments` which is a reserved word. I've renamed it to args. A similar change was recently applied to the create-tauri-app [here](https://github.com/tauri-apps/create-tauri-app/pull/478). A user hitting this issue with the `bun` runtime can be found [here](https://github.com/oven-sh/bun/issues/3237#issuecomment-1734476373).

- [X] Bugfix

- [X] No

### Checklist

### DIFF
--- a/.changes/js-strict.md
+++ b/.changes/js-strict.md
@@ -1,0 +1,5 @@
+---
+"@tauri-apps/cli": 'patch:bug'
+---
+
+Fix running `bun --bun run tauri dev` in JS environments in strict mode.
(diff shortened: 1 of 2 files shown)


amrbashir:
This has already been fixed and published, we will merge 1.x branch to `dev` later before publishing a new alpha version. Thanks for the contribution.

PR #7928, opened 2023-10-01
feat: "Using an SVG instead of png icons"

imabhinavdev:

- [x] Feature

- [ ] No

### Information

I have made changes that enable Tauri to use SVG icons instead of PNG icons. This enhancement allows for better scalability and resolution independence, making Tauri more versatile when it comes to icon handling. SVG icons can be scaled without loss of quality, which is especially valuable for applications running on various devices and display resolutions.


Beanow:
Looks like a breaking change in the CLI helper tool. So that needs a closer look.

I'd like a bit more motivation as well. The positive of a single SVG is also the negative of a single SVG. You give up control on how it scales and renders. Going in manually to fine tune how your icon looks at 32x32 is sometimes needed to get it to look crisp.

But maybe my views on that are outdated, if you have any good resource to say SVGs are now objectively better on the big 3 desktop platforms I think that'd help motivate using this as the default option.

(Of course in the end devs should be allowed to make the tradeoff themselves, so this is more about defaults, maybe even educational value of the example code.)

FabianLars:
@Beanow So the idea was to intially only have support in the `tauri icon`: https://github.com/tauri-apps/tauri/issues/6389 - it currently works by re-scaling a single png into multiple pngs so svg input would be a win here (i'd still love to implement support for multiple input icons). I think only flatpaks actually support svgs rn.

@madabhi It looks like you aimed to make tauri itself work with svgs at runtime? The main problem with that is missing native support for svgs on all platforms, so we'd either need to change tauri itself to convert svgs to something the os understand at runtime, or go with what i wrote above and keep using mostly pngs.
In that context, every single change in this PR is broken.
- The tauri.conf.json changes should break at the bundling point unless we have a fallback i forgot about, but should cause runtime problems later either way.
- The json.icns change simply breaks the icon.icns generator for the macos icon, or maybe it has 0 effect since the file names don't actually effect the file type (still png).
- The change in icon.rs is not broken by itself but doesn't solve anything either.

Sooooo, where do we go from here? I think adding runtime support for svgs in tauri is a bit too ambitious and maybe even inefficient, so imo we should stick with the `tauri icon` approach which was also worked on here: https://github.com/tauri-apps/tauri/pull/6444. That PR was already pretty close to what i imagined but it's still missing some work on the feedback provided until it could be merged, maybe you'd like to pick that up instead seeing that the author doesn't work on it anymore?

Issue #7756, opened 2023-09-05
[meta] Breaking changes before v2-beta release


amrbashir:
#### Public API changes:

- [X] Remove custom protocol types and replace with `http` crate types - https://github.com/tauri-apps/tauri/pull/7754
- [x] Remove `tauri::api` module, not sure if it is needed, it still has hidden exports but ultimately it should be removed and replace with a public API if needed - https://github.com/tauri-apps/tauri/pull/7874
- [x] `path` and `plugin` modules both, have a `Result` type alias, where they should be using `tauri::Result` instead and not have Error types on their own - https://github.com/tauri-apps/tauri/pull/7875
- [x] ~~Make almost all callbacks (where it makes sense), return a `tauri::Result` because most of our APIs return `tauri::Result` and would be easier to use `?` operator.~~
- [x] Remove `rand` and `uuid` crates, our usages of it could be replaced with an atomically incremented counter - https://github.com/tauri-apps/tauri/pull/7939
- [x] change `Env.args` to pull from `std::env::args_os` instead of `std::env::args` - https://github.com/tauri-apps/tauri/pull/7876
- [x] Consistent naming
  - [x] Scope struct names, we have `FsScope`, then we have `ipc::Scope` - https://github.com/tauri-apps/tauri/pull/7944
  - [x] `TrayIconBuilder/TrayIcon::on_tray_event` should be renamed to `TrayIconBuilder/TrayIcon::on_tray_icon_event` - https://github.com/tauri-apps/tauri/pull/7943
  - [x] [Environment variables](https://github.com/tauri-apps/tauri/blob/29818de682146a75cd9a886f7b215154cad1d42d/tooling/cli/ENVIRONMENT_VARIABLES.md) used by tauri and tauri-cli is inconsistent and needs major renaming with possibly prefixing each var with `TAURI_<SCOPE>_<VAR>` - https://github.com/tauri-apps/tauri/pull/7949
  A few examples but not limited to: 
   	- `TAURI_FIPS_COMPLIANT` should be `TAURI_BUNDLER_WIX_FIPS_COMPLIANT`
   	- `TAURI_PRIVATE_KEY` relies on `TAURI_KEY_PASSWORD` and so it should be renamed to `TAURI_PRIVATE_KEY_PASSWORD`
   	- `APPLE_ID` and `APPLE_PASSWORD` should probably be prefixed with `TAURI_` unless these env vars are somewhat of a convention used by other tools, in that case, it should be fine to keep as is.
[...]

PR #7939, opened 2023-10-02
refactor!: remove uuid and rand dependencies where applicable

amrbashir:
ref: https://github.com/tauri-apps/tauri/issues/7756

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [X] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] Yes, and the changes were approved in issue #7756
- [ ] No

### Checklist


amrbashir:
There is a couple of places that still use `rnd` and `uuid` crates:
1. isolation pattern still uses `uuid`, not sure if we can replace it with a simple counter, cc @chippers
2. csp nonce and asset protocol range requests -> migrated to the lighter `getrandom` crate in [d398300](https://github.com/tauri-apps/tauri/pull/7939/commits/d398300ef0e0354ce18a48663e9e5f7fc9f19bd3)

chippers:
> There is a couple of places that still use `rnd` and `uuid` crates:
> 
> 1. isolation pattern still uses `uuid`, not sure if we can replace it with a simple counter, cc @chippers
> 2. csp nonce and asset protocol range requests -> migrated to the lighter `getrandom` crate in [d398300](https://github.com/tauri-apps/tauri/pull/7939/commits/d398300ef0e0354ce18a48663e9e5f7fc9f19bd3)

it's used so that it's unpredictable and because the output chars are always valid as a schema suffix. using `getrandom` to generate 128 bits and using lowercase hex as the output should have the same effect here.

i've noticed some code that used uuid is replaced with smaller sizes of things, so take care that it's not problematic there. uuid v4 are often used as trusted random unique numbers because they are 128 bytes (well 122 since some are fixed) that you don't need to check for collisions for in most applications because it's so incredibly unlikely to get another of the same. u32s are much more likely to collide

some places they have been replaced with constants (like the event listener object and function names) which was seen as undesirable before to discourage people from using it manually and going through the proper api instead.

amrbashir:
> it's used so that it's unpredictable and because the output chars are always valid as a schema suffix. using getrandom to generate 128 bits and using lowercase hex as the output should have the same effect here

I will leave that one using uuid for now, at least it is only active when using isolation pattern and the main goal for this PR was to remove it from public API.

> i've noticed some code that used uuid is replaced with smaller sizes of things, so take care that it's not problematic there. uuid v4 are often used as trusted random unique numbers because they are 128 bytes (well 122 since some are fixed) that you don't need to check for collisions for in most applications because it's so incredibly unlikely to get another of the same. u32s are much more likely to collide

the u32s are only incrementing so it shouldn't collide with itself, there maybe other ids that could have the same value but they are used in other places.

> some places they have been replaced with constants (like the event listener object and function names) which was seen as undesirable before to discourage people from using it manually and going through the proper api instead.

Yeah, I wasn't aware of the history behind that, I could use `getrandom` in this case but may I ask why it was discouraged? I mean we don't ever provide support for those who uses it directly and we can break it without any notice.

chippers:
> Yeah, I wasn't aware of the history behind that, I could use `getrandom` in this case but may I ask why it was discouraged? I mean we don't ever provide support for those who uses it directly and we can break it without any notice.

just to avoid the typical "spacebar heating" problem if the behavior does change - and we should explicitly mark somewhere that it is unstable

amrbashir:
could change the const from
```rs
function_name: "_listeners_function_id_",
object_name: "_listeners_object_id_",
```
to
```rs
function_name: "_internal_listeners_function_id_",
object_name: "_internal_listeners_object_id_",
```

.changes/tauri-runtime-uuid.md
```diff
@@ -0,0 +1,5 @@
+---
+'tauri-runtime': 'patch:breaking'
+---
+
+Added `WnidowEventId` type and Changed `Dispatch::on_window_event` return type from `Uuid` to `WindowEventId`
```
lucasfernog:
```suggestion
Added `WindowEventId` type and Changed `Dispatch::on_window_event` return type from `Uuid` to `WindowEventId`
```

.changes/tauri-uuid-rand.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Removed `EventHandler` type.
+- Added `EventId` type
+- Changed `Manager::listen_global` and `Window::listen` to return the new `EventId` type instead of `EventHandler`.
+- Removed the return type of `Manager::once_global` and `Window::once`
+- Changed `Manager::unlisten` and `Window::unlisten` to take he new `EventId` type.
+- Added `tauri::scope::ScopeEventId`
+- Changed `FsScope::listen` to return the new `ScopeEventId` instead of `Uuid`.
+- Added `FsScope::unlisten`
+-
```
lucasfernog:
```suggestion

```

core/tauri/src/window.rs
```diff
@@ -2228,15 +2228,15 @@ impl<R: Runtime> Window<R> {
...
     handler: CallbackFn,
-  ) -> crate::Result<usize> {
-    let event_id = rand::random();
+  ) -> crate::Result<EventId> {
+    let event_id = self.manager.listeners().next_event_id();
 
     self.eval(&crate::event::listen_js(
-      self.manager().event_listeners_object_name(),
-      format!("'{}'", event),
+      self.manager().listeners().function_name(),
```
lucasfernog: this one was broken, changed listeners_object_name to function_name messed up the evaluated JS. I'm pushing a fix

lucasfernog (dismissed):

lucasfernog (approved):

Issue #7074, opened 2023-05-27
[bug] Debian packages are not compliant and dangerous


debug-richard:
### Describe the bug

I used the tool [lintian](https://wiki.debian.org/Lintian), the offical debian linter, to check the compliance of the tauri generated deb archives.
Unfortunately, they contain not only some trivial issues, but also some security problems.
```
$ lintian target/release/bundle/deb/tauri-app_0.0.0_amd64.deb 
E: tauri-app: control-file-has-bad-owner md5sums 1000/1000 != root/root (or 0/0)
E: tauri-app: control-file-has-bad-permissions md5sums 0664 != 0644
E: tauri-app: malformed-contact Maintainer you
E: tauri-app: missing-dependency-on-libc needed by usr/bin/tauri-app
E: tauri-app: no-changelog usr/share/doc/tauri-app/changelog.gz (native package)
E: tauri-app: no-copyright-file
E: tauri-app: unstripped-binary-or-object usr/bin/tauri-app
E: tauri-app: wrong-file-owner-uid-or-gid usr/ 1000/1000
E: tauri-app: wrong-file-owner-uid-or-gid usr/bin/ 1000/1000
E: tauri-app: wrong-file-owner-uid-or-gid usr/bin/tauri-app 1000/1000
E: tauri-app: wrong-file-owner-uid-or-gid ... use --no-tag-display-limit to see all (or pipe to a file/program)
W: tauri-app: description-synopsis-starts-with-article
W: tauri-app: desktop-entry-lacks-main-category usr/share/applications/tauri-app.desktop
W: tauri-app: no-manual-page usr/bin/tauri-app
W: tauri-app: non-standard-dir-perm usr/ 0775 != 0755
W: tauri-app: non-standard-dir-perm usr/bin/ 0775 != 0755
W: tauri-app: non-standard-dir-perm usr/share/ 0775 != 0755
W: tauri-app: non-standard-dir-perm ... use --no-tag-display-limit to see all (or pipe to a file/program)
W: tauri-app: non-standard-executable-perm usr/bin/tauri-app 0775 != 0755
W: tauri-app: non-standard-file-perm usr/share/applications/tauri-app.desktop 0664 != 0644
W: tauri-app: non-standard-file-perm usr/share/icons/hicolor/128x128/apps/tauri-app.png 0664 != 0644
W: tauri-app: non-standard-file-perm usr/share/icons/hicolor/256x256@2/apps/tauri-app.png 0664 != 0644
[...]

PR #7980, opened 2023-10-07
fix(bundler): In .deb packages, set uid=0 for all files

olivierlemasle:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

This solves one of the issues reported in #7074:

`lintian` (the Debian package linter) reported these errors:
```
E: tauri-app: control-file-has-bad-owner 1000/1000 != root/root (or 0/0) [md5sums]
...
E: tauri-app: wrong-file-owner-uid-or-gid 1000/1000 [usr/]
E: tauri-app: wrong-file-owner-uid-or-gid 1000/1000 [usr/bin/]
E: tauri-app: wrong-file-owner-uid-or-gid 1000/1000 [usr/bin/tauri-app]
...
```

The Debian package file contains two archive files (`control.tar.gz` and `data.tar.gz`) which are created with file metadata copied from the build filesystem: mode, time, uid, gid, etc.

That caused the package to use the uid of the build user (typically an unprivileged user with `uid` 1000) instead of the `root` user as required.

With this commit, the metadata are still copied, except for `uid` and `gid`, which are overridden and set to `0` (=`root`).

### DIFF
--- a/.changes/bundler-deb-fix-owner.md
+++ b/.changes/bundler-deb-fix-owner.md
@@ -1,0 +1,5 @@
+---
+'tauri-bundler': 'patch:bug'
+---
+
+In Debian packages, set `root` the owner of control files and package files.
--- a/tooling/bundler/src/bundle/linux/debian.rs
+++ b/tooling/bundler/src/bundle/linux/debian.rs
@@ -367,10 +367,20 @@ fn create_tar_from_dir<P: AsRef<Path>, W: Write>(src_dir: P, dest_file: W) -> cr
     }
     let dest_path = src_path.strip_prefix(src_dir)?;
     if entry.file_type().is_dir() {
-      tar_builder.append_dir(dest_path, src_path)?;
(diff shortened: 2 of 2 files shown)


FabianLars:
Thanks for the PR! Since we didn't merge 1.x back into dev yet (should happen once the 1.5 release calmed down a bit) i _think_ this should target the 1.x branch instead so 1.x users get the fix faster.

olivierlemasle:
Thanks @FabianLars, I've rebased the PR to target branch `1.x`.

FabianLars (approved):
Thank you!

tweidinger (approved):

tweidinger:
Thanks for this security relevant improvement 👍

lucasfernog (approved):

PR #7990, opened 2023-10-10
refactor(cli): enhance plugin subcommand, closes #7749

### LINKED ISSUE #7749: [feat] Add CLI option to `tauri plugin init` to enable/disable iOS/Android projects
### Describe the problem

Right now when running `tauri plugin init` it will always create a respective Swift and Kotlin project. To make a desktop-only plugin this is an awkward experience as the developer would have to then remove the respective projects.

I personally prefer opt-in, so I'd see the default of `tauri plugin init` to NOT create a mobile project (or to prompt the dev if they'd like to). This could give the option to add flags such as `tauri plugin init [--with-ios|--with-andoird]` to bypass that prompt.

I'm not sure if running `tauri plugin init --with-ios` would still prompt the user if they'd like to init android, or if that would only init the base desktop project with iOS and skip android completely. 

### Alternatives considered

_No response_

### Additional context

Originally brought up in https://github.com/tauri-apps/tauri-docs/pull/1412/files/be2b0c2e42ca05842674bdb5cd5cc55e1f551596#diff-2e62f026cfd52587aa884598003549fdb22855e8637da24c74c0ece364922bd8R36

amrbashir:

- [ ] Bugfix
- [X] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [X] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] Yes, and the changes were approved in issue #___
- [ ] No

### Checklist

### DIFF
--- a/.changes/cli-plugin-init.md
+++ b/.changes/cli-plugin-init.md
@@ -1,6 +1,6 @@
 ---
-'tauri-cli': 'patch'
-'@tauri-apps/cli': 'patch'
+'tauri-cli': 'patch:breaking'
+'@tauri-apps/cli': 'patch:breaking'
 ---
 
 The `tauri plugin` subcommand is receving a couple of consitency and quality of life improvements:
--- a/.changes/cli-plugin-init.md
+++ b/.changes/cli-plugin-init.md
@@ -7,4 +7,6 @@ The `tauri plugin` subcommand is receving a couple of consitency and quality of 
 
 - Renamed `tauri plugin android/ios add` command to `tauri plugin android/ios init` to match the `tauri plugin init` command.
 - Removed the `-n/--name` argument from the `tauri plugin init`, `tauri plugin android/ios init`, and is now parsed from the first positional argument.
+- Added `tauri plugin new` to create a plugin in a new directory.
+- Changed `tauri plugin init` to initalize a plugin in an existing directory (defaults to current directory) instead of creating a new one.
 - Changed `tauri plugin init` to NOT generate mobile projects by default, you can opt-in to generate them using `--android` and `--ios` flags or `--mobile` flag or initalize them later using `tauri plugin android/ios init`.
--- a/.changes/cli-plugin-init.md
+++ b/.changes/cli-plugin-init.md
@@ -1,0 +1,10 @@
+---
+'tauri-cli': 'patch'
+'@tauri-apps/cli': 'patch'
+---
+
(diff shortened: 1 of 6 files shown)


.changes/cli-plugin-init.md
```diff
@@ -0,0 +1,12 @@
+---
+'tauri-cli': 'patch'
+'@tauri-apps/cli': 'patch'
```
amrbashir:
```suggestion
'tauri-cli': 'patch:breaking'
'@tauri-apps/cli': 'patch:breaking'
```

tooling/cli/src/plugin.rs
```diff
@@ -25,17 +28,46 @@ pub struct Cli {
...
   }
 
   Ok(())
 }
+
+fn infer_plugin_name<P: AsRef<Path>>(directory: P) -> Result<String> {
+  let dir = directory.as_ref();
+  let carg_toml_path = dir.join("Cargo.toml");
+  let name = if carg_toml_path.exists() {
+    let contents = std::fs::read(carg_toml_path)?;
```
lucasfernog:
```suggestion
  let cargo_toml_path = dir.join("Cargo.toml");
  let name = if cargo_toml_path.exists() {
    let contents = std::fs::read(cargo_toml_path)?;
```

tooling/cli/src/plugin.rs
```diff
@@ -25,17 +28,46 @@ pub struct Cli {
 
 #[derive(Subcommand)]
 enum Commands {
+  New(new::Options),
```
lucasfernog, line 31: I prefer `create` instead of `new` what do you think?

lucasfernog (changes requested):

tooling/cli/src/plugin.rs
amrbashir: `new` is the usually the term I see in bootstrapping CLIs when you have both `init` and `new`, for example, `cargo new` and `cargo init` (I honestly can't think of another example, but I am quite certain I saw it somewhere else)

amrbashir: or we could have both

lucasfernog (approved):

tooling/cli/src/plugin.rs
lucasfernog: sold me on the cargo example

PR #8025, opened 2023-10-17
build(api.js): simplify output JS files

amrbashir:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

We no longer minify the NPM package, it should be the responsibility of the user bundler system to minify it, this allows the user to easily modify and patch our package locally if he wishes so.


amrbashir:
This should be ready now

tooling/api/tsup.config.ts
```diff
@@ -1,34 +0,0 @@
...
-
-import { defineConfig } from 'tsup'
-
-export default defineConfig(() => [
-  {
-    entry: ['src/*.ts'],
-    outDir: 'dist',
-    format: ['esm', 'cjs'],
-    clean: true,
-    minify: true,
```
lucasfernog, line 13: we used to minify everything, but now the terser plugin is only applied on the bundle script, is there a reason for that?

lucasfernog (changes requested):

tooling/api/tsup.config.ts
amrbashir: The goal of this PR was to simplify the npm package files to enhance the DX of modifying and applying patches inside node_modules, and minifing the output makes that significantly harder. This is also a softer approach than going full JSDoc

lucasfernog (approved):

PR #8042, opened 2023-10-18
fix(cli): duplicated newlines on child process output

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/tooling/bundler/src/bundle/common.rs
+++ b/tooling/bundler/src/bundle/common.rs
@@ -8,7 +8,7 @@ use log::debug;
 use std::{
   ffi::OsStr,
   fs::{self, File},
-  io::{self, BufReader, BufWriter},
+  io::{self, BufRead, BufReader, BufWriter},
   path::Path,
   process::{Command, ExitStatus, Output, Stdio},
   sync::{Arc, Mutex},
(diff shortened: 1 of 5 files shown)


amrbashir (dismissed):

tooling/bundler/src/bundle/common.rs
```diff
@@ -172,9 +172,12 @@ impl CommandExt for Command {
         if let Ok(0) = tauri_utils::io::read_line(&mut stdout, &mut buf) {
           break;
         }
-        debug!(action = "stdout"; "{}", String::from_utf8_lossy(&buf));
+        let mut line = String::from_utf8_lossy(&buf);
+        if line.ends_with('\n') {
+          line.to_mut().pop();
+        }
+        debug!(action = "stdout"; "{line}");
```
amrbashir: How about just always ignoring the last character? since `read_line` will wait until `\n`
```suggestion
        let line = String::from_utf8_lossy(&buf);
        debug!(action = "stdout"; "{}",  &line[0..line.len() - 1]);
```

lucasfernog: I think that is slower (isn't that creating another string?)

amrbashir: Updated the snippet, it is just taking a reference to a slice while keeping the original String

amrbashir: I'd also suggest using `String::new` with a `BufReader` and read the lines into the string, slice into it when printing and then extending the `lines` vector with the value of `string.as_bytes()`

amrbashir (approved):

Issue #4315, opened 2022-06-10
[bug] Could not determine the accessibility bus address


Elanis:
### Describe the bug

Hello,

Might be linked to https://github.com/tauri-apps/tauri/issues/4174, not sure as it's the same error but not the same case at all.

When trying to run tauri dev for benchmark purpose (https://github.com/Elanis/web-to-desktop-framework-comparison/pull/73), I'm having an error saying `Could not determine the accessibility bus address`. Google isn't helping a lot this time :/

This ends like this:
```
Building [=======================> ] 366/367: app(bin)                    
Finished dev [unoptimized + debuginfo] target(s) in 7m 31s
Running `target/debug/app`
Could not determine the accessibility bus address
```
([Link to full logs, with specific line](https://github.com/Elanis/web-to-desktop-framework-comparison/runs/6834052075?check_suite_focus=true#step:11:4771))

### Reproduction

- Create an empty app
- Create a github action CI on ubuntu-latest
   - Install recommended linux packages (https://tauri.studio/v1/guides/getting-started/prerequisites#setting-up-linux)
   - npx tauri dev

### Platform and versions

```shell
Packages
  › @tauri-apps/cli [NPM]: 1.0.0-rc.13
  › @tauri-apps/api [NPM]: 1.0.0-rc.6
  › tauri [RUST]: 1.0.0-rc.14,
  › tauri-build [RUST]: 1.0.0-rc.12,
  › tao [RUST]: 0.9.1,
  › wry [RUST]: 0.17.0,

App
  › build-type: bundle
  › CSP: unset
  › distDir: ../dist
  › devPath: ../dist

App directory structure
  ├─ dist
  ├─ node_modules
  └─ src-tauri
```

### Additional context

Here is my pull request where I try to fix that behavior: https://github.com/Elanis/web-to-desktop-framework-comparison/pull/73

Elanis:
I tried adding `at-spi2-core` with apt in addition to the recommended tauri packages, based on Google search, but same result

lucasfernog:
Might be worth looking into how we run our own benchmarks: https://github.com/tauri-apps/tauri/blob/8a63489567b9fa86e404ad42b5b30c64361efe85/.github/workflows/bench.yml#L107

Regarding the issue itself, i've never seen it :/

Elanis:
Oh, didn't see you had your own benchmark, will look into it ! Thanks ;)

lucasfernog:
[...]

PR #8045, opened 2023-10-18
fix(ci): Fix bench job

olivierlemasle:
Fixes `bench` CI job, which is failing on `dev` branch (`webkit2gtk-4.1` is not available for `ubuntu-20.04`).

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -17,7 +17,7 @@ on:
 env:
   RUST_BACKTRACE: 1
   CARGO_PROFILE_DEV_DEBUG: 0 # This would add unnecessary bloat to the target folder, decreasing cache efficiency.
-  LC_ALL: en_US.UTF-8 # This prevents strace from changing it's number format to use commas.
+  LC_ALL: en_US.UTF-8 # This prevents strace from changing its number format to use commas.
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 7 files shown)


lucasfernog:
Our benchmark CI is completely broken :( it runs forever.

olivierlemasle:
Well, that wasn't expected :man_shrugging: 

I'll look at it.

olivierlemasle:
@lucasfernog I've fixed the benchmarks, which were running forever because they are normally stopped by a Tauri command, and `window.__TAURI__.invoke` was used, so the command was never invoked.

This time, I tested it in a GitHub fork ;-)

I've also fixed some Clippy warnings, and I've added the package `at-spi2-core` to remove these warnings (cf #4315):

> AT-SPI: Error retrieving accessibility bus address: org.freedesktop.DBus.Error.ServiceUnknown: The name org.a11y.Bus was not provided by any .service files

---

If you want to make sure a future Tauri breakage will not make the benchmarks run forever as it did, I can add a timeout. Hyperfine does not support this (cf. https://github.com/sharkdp/hyperfine/issues/106) but we can wrap the command by a `timeout` command.

lucasfernog (approved):

PR #8076, opened 2023-10-21
feat(plugins): typed invoke arguments for mobile plugins

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

### Checklist

### Other information

This PR changes the plugin command API to enforce typed arguments instead of relying on on-demand argument processing via `invoke.getString`, `invoke.getInt`, `invoke.getObject` etc.

New API for Android:

```kotlin
class Options(val recursive: Boolean = false)

class ExampleArgs(val path: String?, val options: Options?, val enabled: Boolean = true)

@TauriPlugin
class ExamplePlugin(private val activity: Activity): Plugin(activity) {
    @Command
    fun run(invoke: Invoke) {
        val args = invoke.parseArgs(ExampleArgs::class.java)
        invoke.resolveObject(args)
    }
}

```

New API for iOS:

```swift
class Options: Decodable {
  let recursive: Bool = false
}

class ExampleArgs: Decodable {
  let path: String?
  let enabled: Bool = true
  let options: Options?
}

class ExamplePlugin: Plugin {
  @objc public func run(_ invoke: Invoke) throws {
    let args = try invoke.parseArgs(ExampleArgs.self)
    invoke.resolve(args)
  }
}
```

Note that you can also resolve with an object instead of the generic `JSObject` class / dictionary.


lucasfernog:
seems like the jackson dependency costs 1MB on the AAB size, maybe we can optimize that or find a better package for this (though AFAIK there's no Java expert on the team :joy: )

core/tauri/mobile/android/src/main/java/app/tauri/plugin/Plugin.kt
```diff
@@ -167,7 +185,8 @@ abstract class Plugin(private val activity: Activity) {
 
       // If call was made with a list of specific permission aliases to request, save them
       // to be requested
-      val providedPerms: JSArray = invoke.getArray("permissions", JSArray())
+      // TODO val providedPerms: JSArray = invoke.getArray("permissions", JSArray())
```
amrbashir: Why was this changed to TODO? should it be solved in this PR?

amrbashir (dismissed):
Android sounds OK to me

core/tauri/mobile/android/src/main/java/app/tauri/plugin/Plugin.kt
lucasfernog: oops missed this one, thanks for catching it @amrbashir

amrbashir (approved):

amrbashir:
> seems like the jackson dependency costs 1MB on the AAB size, maybe we can optimize that or find a better package for this (though AFAIK there's no Java expert on the team 😂 )

Java yikes, do these developers exist any more :joy:?

PR #8117, opened 2023-10-26
fix(macos): use BTreeMap for windows map to prevent crash on idle

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


core/tauri-runtime-wry/src/lib.rs
```diff
@@ -2305,7 +2306,7 @@ impl<T: UserEvent> Runtime<T> for Wry<T> {
 pub struct EventLoopIterationContext<'a, T: UserEvent> {
   pub callback: &'a mut (dyn FnMut(RunEvent<T>) + 'static),
   pub webview_id_map: WebviewIdStore,
-  pub windows: Rc<RefCell<HashMap<WebviewId, WindowWrapper>>>,
+  pub windows: Arc<RefCell<BTreeMap<WebviewId, WindowWrapper>>>,
```
wusyong:
```suggestion
  pub windows: Arc<Mutex<BTreeMap<WebviewId, WindowWrapper>>>,
```

We should use Mutex at this point even you want to use it in single thread.
This is already a control flow that has to work with tokio.
Adding lock instead of runtime check then panic is preferred behaviour imho.

wusyong (changes requested):

core/tauri-runtime-wry/src/lib.rs
lucasfernog: isn't RefCell better for performance? we also had some issues with deadlock before, and since using refcell (which is also a lot easier to track that kind of issue - mutating a borrowed value in this case) we've been fine

lucasfernog:
Discord thread for this issue: https://discord.com/channels/616186924390023171/1166431206104109157

amrbashir (approved):

MGough:
I had a build without this patch where I was able to reproduce the crashing behaviour reliably. Since applying this patch (including the most recent commits) I've been unable to reproduce the crashing.

I appreciate that's a fairly anecdotal comment!

MGough:
~~I'm afraid I've now been able to reproduce it once more with this patch applied on my project. See the discord thread. Happy to help out more to find the cause.~~

MGough:
Update on the previous comment, I was incorrect! It hadn't been applied correctly. Having applied it correctly (checking my `Cargo.lock` to make sure it's applied throughout rather than `unused`) it has resolved our persistent crashing issue. We'd love to see this merged so that we can stay up to date!

amrbashir (approved):

PR #8124, opened 2023-10-27
refactor(core): break manager struct into smaller ones

lucasfernog:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist


core/tauri/src/manager/window.rs
```diff
@@ -0,0 +1,810 @@
...
+/// Uses a custom URI scheme handler to resolve file requests
+pub struct UriSchemeProtocol<R: Runtime> {
+  /// Handler for protocol
+  #[allow(clippy::type_complexity)]
+  pub protocol:
+    Box<dyn Fn(&AppHandle<R>, http::Request<Vec<u8>>, UriSchemeResponder) + Send + Sync>,
+}
+
+#[cfg(desktop)]
+pub struct MenuManager<R: Runtime> {
```
amrbashir: I think `MenuManager` could also be split into its own file and should be a field in `AppManager` instead as it also affects `App/AppHandle`.

core/tauri/src/manager/mod.rs
```diff
@@ -0,0 +1,792 @@
...
+  /// The asset bytes.
+  pub bytes: Vec<u8>,
+  /// The asset's mime type.
+  pub mime_type: String,
+  /// The `Content-Security-Policy` header value.
+  pub csp_header: Option<String>,
+}
+
+#[default_runtime(crate::Wry, wry)]
+pub struct AppManager<R: Runtime> {
```
amrbashir, line 180: should probably also move this to `manager/app.rs`

amrbashir (changes requested):
I think my main problem with this PR is the excessive usage of `Arc`, do you think we can somehow remove the need for it?

Edit: the `Arc` usage was excessive before as well, I guess I misspoke, what I mean, can we reduce the `Arc` usage now that we have grouped multiple fields together?

core/tauri/src/manager/window.rs
lucasfernog: good one, pushed

core/tauri/src/manager/mod.rs
lucasfernog: i'm not really sure, since AppManager is the root manager it can live in mod.rs IMO

lucasfernog:
@amrbashir I've done my best with the Arc removals, some are still around because there's some types that are cloned and moved to another thread / closure (mostly callbacks).

core/tauri/src/manager/mod.rs
amrbashir: Yeah I agree but app.rs gives more clarity as it is not clear from first glance where AppManager lives.

amrbashir: anyways, this is just a nit so go with your gut.

amrbashir (approved):

PR #8168, opened 2023-11-04
fix(build): always invoke resource compiler on windows, fixes #8164

### LINKED ISSUE #8164: [bug] [alpha.17] [windows] completely broken if config.package.version is None
### Describe the bug

In case of no `version` field in the config, the whole app never loads on Windows. According to the docs this field can be set to None in which case it should default to whatever is present in `package.json`

### Reproduction

Run `npm create tauri-app@latest --alpha`.
Remove `version: "0.0.0"` from tauri.conf.json (and the preceeding comma)
Run `tauri dev`

### Expected behavior

_No response_

### Platform and versions

```text
[✔] Environment
    - OS: Windows 10.0.20348 X64
    ✔ WebView2: 118.0.2088.76
    - node: 18.13.0
    - yarn: 1.22.19
    - npm: 8.19.3

[-] Packages
    - tauri [RUST]: 2.0.0-alpha.17
    - tauri-build [RUST]: 2.0.0-alpha.11
    - wry [RUST]: 0.34.2
    - tao [RUST]: 0.23.0
    - @tauri-apps/api [NPM]: 2.0.0-alpha.11
    - @tauri-apps/cli [NPM]: 2.0.0-alpha.17

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../dist
    - bundler: Vite
```

### Stack trace

_No response_

### Additional context

_No response_

FabianLars:
This fixes a simple v1 merge error that caused winres/embed-resource to only be compiled if package.version was set in tauri.conf.json.
While i was at it i also removed the duplicate `if windows` check.

We may want to remove the icon path check to (== check for it only when setting the icon) so that the resource compiler still runs even if no icon is provided - may as well wait for a usecase for that though 🤷 
Edit: Changed the `if` location to only wrap the set_icon call to make that change a tiny bit easier.

fixes #8164

### DIFF
--- a/.changes/tauri-build-resource-compiler.md
+++ b/.changes/tauri-build-resource-compiler.md
@@ -1,0 +1,5 @@
+---
+"tauri-build": 'patch:bug'
+---
+
+Fixed an issue that caused the resource compiler to not run on Windows when `package.version` was not set in `tauri.conf.json` preventing the app from starting.
--- a/core/tauri-build/src/lib.rs
+++ b/core/tauri-build/src/lib.rs
@@ -534,45 +534,43 @@ pub fn try_build(attributes: Attributes) -> Result<()> {
       .window_icon_path
       .unwrap_or_else(|| find_icon(&config, |i| i.ends_with(".ico"), "icons/icon.ico"));
 
-    if target_triple.contains("windows") {
-      if window_icon_path.exists() {
-        let mut res = WindowsResource::new();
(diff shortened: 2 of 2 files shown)


FabianLars:
Also, the actually weird thing here is that the app fails to start because of the missing dialog manifest, so we may have messed something up elsewhere too that now this is always required even without using the dialog plugin.

amrbashir (approved):

PR #8200, opened 2023-11-09
feat(cli): merge user-defined plist with the iOS plist file

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### Other information

Merge the src-tauri/Info.plist and src-tauri/Info.ios.plist files with the iOS project Info.plist file to allow extending the IPA configuration, similar to the macOS implementation.

### DIFF
--- a/.changes/merge-ios-plist.md
+++ b/.changes/merge-ios-plist.md
@@ -1,0 +1,6 @@
+---
+"tauri-cli": patch:feat
+"@tauri-apps/cli": patch:feat
+---
+
+Merge `src-tauri/Info.plist` and `src-tauri/Info.ios.plist` with the iOS project plist file.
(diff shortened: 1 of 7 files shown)


amrbashir (approved):

lucasfernog:
@FabianLars raised a possible issue with merging `Info.plist` for both macOS and iOS, but I couldn't think of anything really.. I feel like we could also read `Info.macos.plist` and let the user decide what's get shared to both platforms.

FabianLars:
even if there aren't any actual consequences (that'd crash something or something) they'd still pollute each other right? I mean, we're only talking about a few (k)bytes but v2 already has much larger binaries so every bit counts i guess?

but yeah, as long as there aren't any actual issues i don't really care ig

PR #8228, opened 2023-11-15
feat(core): add test::test::get_ipc_response, closes #8199

### LINKED ISSUE #8199: [feat] Opening access to `tauri::test::MockRuntime` to enable windowless fuzzing of Tauri applications
### Describe the problem

Hello everyone,

I've been working on a framework to fuzz Tauri applications. 
The goal is to be able to detect vulnerabilities that could exist in an application backend code. 
Currently I'm fuzzing the Tauri commands of applications for potential exploits due to wrong input validation or 

To improve performance of fuzzing I have been using Tauri `MockRuntime` to run a Tauri application without a webview. 
For this to be possible I need to be able to access and use the `MockRuntime` from external code. 

I need to open accessibility to Tauri `MockRuntime`. 
I also need few helper functions to facilitate the fuzzing process. 
All these code can be hidden behind a `fuzz` feature. 

### Alternatives considered

Another possibility is to write another completely separated runtime rather than using the `MockRuntime`.
This may happen in the future if the fuzzer requires more features that what is currently provided by the `MockRuntime`. 

### Additional context

I already have a working forked version of Tauri on a private repo. 
I'm writing this issue to ask for permissions and recommendations for a PR that would be accessible on Tauri official repo.

vdang-crabnebula:
#8199 
Some helper functions added to the  `tauri::test::MockRuntime` that can help to fuzz Tauri applications.

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### DIFF
--- a/core/tauri/src/test/mod.rs
+++ b/core/tauri/src/test/mod.rs
@@ -85,7 +85,7 @@ use tauri_utils::{
   config::{CliConfig, Config, PatternKind, TauriConfig},
 };
 
-/// Hash structure f
+/// A key for an [`Ipc`] call.
 #[derive(Eq, PartialEq)]
 pub struct IpcKey {
   /// callback
--- a/core/tauri/src/test/mod.rs
+++ b/core/tauri/src/test/mod.rs
@@ -85,10 +85,13 @@ use tauri_utils::{
   config::{CliConfig, Config, PatternKind, TauriConfig},
 };
 
+/// Hash structure f
 #[derive(Eq, PartialEq)]
 pub struct IpcKey {
-  callback: CallbackFn,
-  error: CallbackFn,
+  /// callback
+  pub callback: CallbackFn,
+  /// error callback
+  pub error: CallbackFn,
 }
 
 impl Hash for IpcKey {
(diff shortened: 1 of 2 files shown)


lucasfernog:
We have [assert_ipc_response](https://github.com/tauri-apps/tauri/blob/4b6a602a89b36f24d34d6ccd8e3c9b7ce202c9eb/core/tauri/src/test/mod.rs#L255) which works in a similar way, maybe we can add a variant to only return the response instead of asserting it. Does it really need to be coupled with `app.run` and `app.run_iteration`? Can't that be done on your end?

vdang-crabnebula:
> We have [assert_ipc_response](https://github.com/tauri-apps/tauri/blob/4b6a602a89b36f24d34d6ccd8e3c9b7ce202c9eb/core/tauri/src/test/mod.rs#L255) which works in a similar way, maybe we can add a variant to only return the response instead of asserting it. Does it really need to be coupled with `app.run` and `app.run_iteration`? Can't that be done on your end?

Well you're right, it shouldn't be in the Tauri code. I will do another proposal and move most code into my fuzzing lib

core/tauri/src/test/mod.rs
```diff
@@ -82,10 +85,13 @@ use tauri_utils::{
   config::{CliConfig, Config, PatternKind, TauriConfig},
 };
 
+/// Hash structure f
```
lucasfernog:
```suggestion
/// A key for an [`Ipc`] call.
```

lucasfernog (approved):

PR #8233, opened 2023-11-15
fix(bundler/nsis): calculate estimated size on build system

FabianLars:
So the project from the devs that reported this issues includes 200+ folders and 1000+ files via the resources config. In this case it's a node_modules folder (intentionally) and putting aside whether or not one should bundle a node_modules folder this can also easily happen in non-nodejs contexts.

Anyway, said app took ~8 minutes to install on my machine. Btw, the files are ~20mb total.

1st commit: Pointing the GetSize call directly at the file instead of the whole folder with a mask reduces that to ~12 seconds - btw the GetSize docs are super confusing, i also thought it can only point to a folder so i looked at other things before i tried this first...

2nd commit: This i'm a bit unsure about whether we can add it to v1 or only to v2 which is why i split it up into another commit. Basically the folder structure is now calculated upfront on the build system, sorted and deduplicated and in the installer the dir creation and file extraction are now 2 steps to not create the same dir again and again. This reduced the install time to ~10 seconds.

### Further "improvements"

- SetCompress off reduces it to ~3 to 4 seconds so that _may_ be something worth to add as a config.
- If we don't check the install size it can be basically instant (without compression) so my initial idea was to calculate file sizes on the build system and use them for the progress report. That _may_ lead to slightly inaccurate progress bars but i don't think that matters (we can just set it too 100% at the end i guess). Buuut that's a slightly bigger change i didn't want to implement without talking about it and this PR is basically a hotfix for the devs that reached out.

### DIFF
--- a/.changes/nsis-slow-resources-installation.md
+++ b/.changes/nsis-slow-resources-installation.md
@@ -1,0 +1,7 @@
+---
+"@tauri-apps/cli": patch:bug
+"tauri-cli": patch:bug
+"tauri-bundler": patch:bug
+---
+
+Fixes an issue in the NSIS installer which caused the installation to take much longer than expected when many `resources` were added to the bundle.
--- a/tooling/bundler/src/bundle/windows/nsis.rs
+++ b/tooling/bundler/src/bundle/windows/nsis.rs
@@ -303,6 +303,9 @@ fn build_nsis_app_installer(
   data.insert("out_file", to_json(out_file));
 
   let resources = generate_resource_data(settings)?;
+  let resources_dirs =
+    std::collections::HashSet::<PathBuf>::from_iter(resources.values().map(|r| r.0.to_owned()));
+  data.insert("resources_dirs", to_json(resources_dirs));
   data.insert("resources", to_json(resources));
 
   let binaries = generate_binaries_data(settings)?;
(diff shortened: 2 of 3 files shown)


amrbashir:
> 2nd commit: This i'm a bit unsure about whether we can add it to v1 or only to v2 which is why i split it up into another commit. Basically the folder structure is now calculated upfront on the build system, sorted and deduplicated and in the installer the dir creation and file extraction are now 2 steps to not create the same dir again and again. This reduced the install time to ~10 seconds.

Yeah this can be backported, we don't provide any guarantees that this file won't be broken at any time.

> * SetCompress off reduces it to ~3 to 4 seconds so that _may_ be something worth to add as a config.

We do have an option to control the algorithim, we can add a variant to disable compression.

> * If we don't check the install size it can be basically instant (without compression) so my initial idea was to calculate file sizes on the build system and use them for the progress report. That _may_ lead to slightly inaccurate progress bars but i don't think that matters (we can just set it too 100% at the end i guess). Buuut that's a slightly bigger change i didn't want to implement without talking about it and this PR is basically a hotfix for the devs that reached out.

I am not sure I am following, the progress bar doesn't depend on the `GetSize` call and caluclating the size at build time should be okay.

FabianLars:
> I am not sure I am following, the progress bar doesn't depend on the GetSize call and caluclating the size at build time should be okay.

Oh right. Nevermind then, i mixed something up.

FabianLars:
Okay turns out my GetSize change was broken (always returned `0` so ig it really only works on folders) so i'll add the rust upfront solution. Running GetSize just once at the end and let it calculate the whole dir was pretty slow too (not 8 minutes though)

amrbashir (approved):

Issue #8173, opened 2023-11-05
[bug] Bundle fails on Apple Accelerate framework


codesoda:
### Describe the bug

Hi all, I have a Tauri macOS app that uses Apple's Accelerate.framework and CoreML.framework.  I've recently upgraded from 1.4 to 1.5, and now the bundler is failing when running `xattr -cr myapp.app` to clear any attributes on the frameworks. This likely has some relation to the work done for #7690.

### Reproduction

- Create a new Tauri app
- Add the Accelerate.framework to the project via tauri.conf.json under bundle > macos > frameworks.
- Try build a release build with apple developer credentials in the environment.
- Build fails on `xattr -rc myapp.app` command.

Example repo can be found at https://github.com/codesoda/accelerate_bug.

### Expected behavior

I would expect that the `xattr` command succeeds and the app bundle is then code-signed and passed to Apple to be notarized.

### Platform and versions

```text
➜  accelerate_test git:(main) yarn tauri info
yarn run v1.22.19
$ tauri info

[✔] Environment
    - OS: Mac OS 12.6.3 X64
    ✔ Xcode Command Line Tools: installed
    ✔ rustc: 1.73.0 (cc66ad468 2023-10-03)
    ✔ cargo: 1.73.0 (9c4383fb5 2023-08-26)
    ✔ rustup: 1.26.0 (5af9b9484 2023-04-05)
    ✔ Rust toolchain: stable-aarch64-apple-darwin (default)
    - node: 20.5.0
    - pnpm: 8.9.2
    - yarn: 1.22.19
    - npm: 9.8.0

[-] Packages
    - tauri [RUST]: 1.5.2
    - tauri-build [RUST]: 1.5.0
    - wry [RUST]: 0.24.4
    - tao [RUST]: 0.16.5
    - @tauri-apps/api [NPM]: 1.5.1
    - @tauri-apps/cli [NPM]: 1.5.6

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../dist
    - devPath: http://localhost:1420/
    - framework: React
    - bundler: Vite
✨  Done in 6.40s.
```

### Stack trace

```text
Running [tauri_cli] Command `cargo  build --features custom-protocol --release`
   Compiling accelerate_test v0.0.0 (/Users/me/projects/accelerate_test/src-tauri)
    Finished release [optimized] target(s) in 4.22s
[...]

Issue #8075, opened 2023-10-21
[feat] Sign .dylib files inside the Frameworks


morajabi:
### Describe the problem

We embed FFMpeg framework inside the app. In tauri v1.5 and latest v2 alpha Tauri signs the frameworks and sidebar binaries. But it does not sign the sign-able files (eg. .dylib) inside the Framework. This results in notarization to fail if the included binaries are signed using a different signature.

### Describe the solution you'd like

I would like Tauri CLI to sign all files that need to be signed in its build steps to ensure there's no mistake. 

### Alternatives considered

We're currently running a pre-build step and signing the dylibs. Same in other Tauri apps like Spacedrive and Recut. But this might result in using two different developer IDs thus failing the notarization or resulting in a crash on startup. 

### Additional context

Context:
https://discord.com/channels/616186924390023171/986184094050316358/1165228387334373387

therealpurplemana:
Did you find any solution to this?

PR #7798, opened 2023-09-09
feat: Add files field in macos build config, closes #3290

repalash:
### What kind of change does this PR introduce?

Added a new field, `appContents` to tauri macos bundle config. Any files and directories added here are copied to the app bundle under the `Contents` directory. 

This makes it possible to add custom `Library` items, `Plugins` built separately with code etc as requested in #3290

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

### Does this PR introduce a breaking change?

- [ ] Yes, and the changes were approved in issue #___
- [x] No

### Checklist
- [x] When resolving issues, they are referenced in the PR's title (e.g `fix: remove a typo, closes #___, #___`)
- [ ] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md).
- [x] I have added a convincing reason for adding this feature, if necessary

### Other information
It's my first PR here. Let me know if you want to change the field name or copy logic, or if these kinds of PRs are not accepted.

PR #8259, opened 2023-11-20
feat(bundler): codesign nested code on macos

pewsheen:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

### Other information

Related: #8075 #8173 

References:
- [Code Signing Tasks](https://developer.apple.com/library/archive/documentation/Security/Conceptual/CodeSigningGuide/Procedures/Procedures.html#//apple_ref/doc/uid/TP40005929-CH4-SW2)

Notes:
- This patch does not cover all the cases but could be a start that we can sign nested code.
- Only signs codes inside standard location (`/MacOS`, `/Frameworks`, `/PlugIns`, `/XPCServices`, `/Helpers`)
- The risk of the patch is that if we are trying to sign something that is not signable, then the whole process will be stopped.
- Successfully codesign and notarized the reproduced repo from #8173 and the `test.framework` in tauri testing.


lucasfernog:
I think we should take this one step forward and apply this logic to the whole .app bundle instead, so it also signs user resources and [macOS custom files on v2](#7798).

pewsheen:
> I think we should take this one step forward and apply this logic to the whole .app bundle instead, so it also signs user resources and [macOS custom files on v2](#7798).

After doing some article search, I think I was just reimplementing the `codesign --deep`, which was [deprecated for reasons](https://developer.apple.com/forums/thread/129980).

Maybe we should take other approaches to sign those files, like providing a `sign_files` field in `tauri.config.json` and letting the user fill the path, entitlements, and is_executable while we can still auto codesign the files that generate from tauri itself.

hjmallon:
I'm really interested in solving this issue (probably in 1.x branch since it is for a live project). Would we rather have code signing on macOS:

1. Where the main app is automatically signed as currenlty, but all ofher files need a config line like @pewsheen says?
2. Something that does a `--deep` equivalent on everything, which is not very fine grained as mentioned?
3. Something that does a `--deep` equivalent on everuthing, but with a new config key for arguments for non standard parts?
4. Something else?

.changes/mac-bundler-nested-code-sign.md
```diff
@@ -0,0 +1,6 @@
+---
+"tauri-cli": patch:feat
+"tauri-bundler": patch:feat
+---
+
+On macOS, support for signing nested code under the framework.
```
lucasfernog:
```suggestion
On macOS, support for signing nested .dylib, .app, .xpc and .framework under predefined directories inside the bundled frameworks ("MacOS", "Frameworks", "Plugins", "Helpers", "XPCServices" and "Libraries").
```

lucasfernog (approved):

raphaelmenges:
> Maybe we should take other approaches to sign those files, like providing a `sign_files` field in `tauri.config.json` and letting the user fill the path, entitlements, and is_executable while we can still auto codesign the files that generate from tauri itself.

Perhaps I overlook it in the documentation, but I am facing this exact issue and you could point me out to such `sign_files` field - if it exists. I am bundling a `.dylib` with my application that must be signed alongside the Tauri binaries to pass the notarization. Right now, notarization fails with `"The binary is not signed with a valid Developer ID certificate."` for that `.dylib`.

Worst case I can code sign the `.dylib` manually, yet, I of course prefer a Tauri-based solution.

therealpurplemana:
Did you find a solution to this? I'm running into this also.

raphaelmenges:
No, still a problem to my knowledge. I plan to use <https://github.com/lando/code-sign-action> to sign the `.dylib` manually. I would highly prefer to let the complete signing happen via Tauri.

therealpurplemana:
This isn't great.  It's causing issues with onnx and opencv libraries if you want to make release builds using them.  Isn't there a way to disable the unsafe library check and it still passes notarization?

raphaelmenges:
I found some workaround using the feature to add "frameworks" in the `tauri.conf.json`: https://github.com/tauri-apps/tauri/discussions/12001#discussioncomment-12417379

therealpurplemana:
> I found some workaround using the feature to add "frameworks" in the `tauri.conf.json`: [#12001 (comment)](https://github.com/tauri-apps/tauri/discussions/12001#discussioncomment-12417379)

This is exactly how I solved it also.  If you specify dylibs in Frameworks, Tauri will a) code sign it, b) properly @rpath it into the Frameworks folder.  I can confirm it passes Apple Notarization also and suspect even Gatekeeper.

PR #8263, opened 2023-11-20
fix(bundler/nsis): remove empty resources folders on uninstall

FabianLars:
This is a follow up of https://github.com/tauri-apps/tauri/pull/8233, so please read that PR description for context.

Currently the uninstaller only deletes the top-level resources folders. NSIS' rmdir function has a `/R` flag for recursive deletion but i thought that it has to follow the "Delete application data" checkbox, even if the installdir is not the data dirs we expose with the bundle id. This makes it quite a bit more ~~complicated~~ spammy since now we have to delete each folder level seperately...

If we prefer to use the /R over a much bigger uninstaller then i'll be happy to adapt this PR, this is just how i understood it was meant to work :)

p.s. The uninstaller now recursively deletes the installdir too _if the checkbox is ticked_, just in case the dev added files in there too.

### DIFF
--- a/.changes/nsis-leftover-dirs.md
+++ b/.changes/nsis-leftover-dirs.md
@@ -1,0 +1,7 @@
+---
+"@tauri-apps/cli": patch:bug
+"tauri-cli": patch:bug
+"tauri-bundler": patch:bug
+---
+
+Fixes an issue in the NSIS installer which caused the uninstallation to leave empty folders on the system if the `resources` feature was used.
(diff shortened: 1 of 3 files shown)


FabianLars:
a tiny bit offtopic but does somebody know how to get rid of this one?

grafik

I can't run clippy locally because of it and have to wait for CI to tell me where i did something dumb.

FabianLars:
Also because of the reverse order this looks like an awful suggestion
grafik

amrbashir (approved):

Issue #2972, opened 2021-11-26
Child webviews


ghost:
**Is your feature request related to a problem? Please describe.**
I'm trying to create a browser. Tauri fits all my needs except that it hasn't support for webviews in windows. See for example [Electron's BrowserView](https://www.electronjs.org/docs/latest/api/browser-view).

**Describe the solution you'd like**
Webview support.

FabianLars:
Duplicate of #2709.

ghost:
Oh! Sorry. Forgot to check for duplicates.

Issue #2709, opened 2021-10-04
A `BrowserView` that can be used to embed additional web content into `webview` window


SOVLOOKUP:
**Describe the solution you'd like**
A `BrowserView` that can be used to embed additional web content into `WebviewWindow` like [electron](https://www.electronjs.org/docs/latest/api/browser-view)

wusyong:
Native webviews don't have such tag. It's a non-standard attribute.
There are others alternative that could also do the job such as iframe and shadow DOM.
They are also standard on the other hand.

ghost:
> Native webviews don't have such tag. It's a non-standard attribute. There are others alternative that could also do the job such as iframe and shadow DOM. They are also standard on the other hand.

Iframes aren't really isolated or support every website. Webviews on the other hand do.

xuchaoqian:
+1 for this feature.
It's very useful to support Embed WebView. With this feature, to create apps which support safety multitasking(such as TABs, multithread but still isolated) can be very easy.

wusyong:
[iframe](https://caniuse.com/?search=iframe) and [shadow DOM](https://caniuse.com/?search=shadow%20Dom) are supported by all modern browsers while `BrowserView` is an electron specific feature.
I strongly recommend you consider these standard approaches. This could make your app compatible to all kind of browser engines.

ghost:
> [iframe](https://caniuse.com/?search=iframe) and [shadow DOM](https://caniuse.com/?search=shadow%20Dom) are supported by all modern browsers while `BrowserView` is an electron specific feature. I strongly recommend you consider these standard approaches. This could make your app compatible to all kind of browser engines.

As I said, iframe doesn't support most sites and shadow DOM is not isolated.

FabianLars:
@dragonDScript Don't ignore the first part of the earlier message:
> Native webviews don't have such tag.

And maybe additionally that one too:
> while BrowserView is an electron specific feature

Well i can see that this sentence(s) might not be quite informative/understandable enough (i blame the language barrier(s) lol). But it's still the most important message here, iframe and shadow dom are more or less just given as a consolidation.

Anyway, there are 2 difficulties here:
1) `BrowserView` is not a native html tag and we can't really create new ones "native" enough for truly embedded webviews -> the custom html tags most browsers already supports is not enough here. Especially since those aren't enabled in webkit by default _iirc_...
[...]

Issue #2076, opened 2021-06-25
multiwindow to side by side 


cheng990629:
**Is your feature request related to a problem? Please describe.**
multiwindow to side by side 
or
create subWindow() 
or 
create Plugin html()

nothingismagick:
This makes no sense to me, sorry.

cheng990629:
> This makes no sense to me, sorry.

how to \<webview\>?

nothingismagick:
I am sorry if it sounded rude, but I am still not sure what you want. Have you looked in the documentation, for example here:

https://tauri.studio/en/docs/guides/splashscreen

Alternatively, please write in your primary spoken language and I will try to translate.

cheng990629:
非常感谢，最后问一下，怎样使用electron 中的\<webview\> tabs功能，独立并行两个app

Laegel:
I guess they'd like to use a webview HTML element in a webview window, a bit like Electron does.

cheng990629:
> I guess they'd like to use a webview HTML element in a webview window, a bit like Electron does.

YES

wusyong:
[Webview tag](https://developer.chrome.com/docs/extensions/reference/webviewTag/) is already deprecated by Chrome which electron uses it.

lysdexic-audio:
Being able to [do this ](https://www.electronjs.org/docs/latest/api/browser-view) would make me jump from Electron to Tauri immediately. 

I am assuming this what the original request was about. I can't see why this was closed, actually. Please consider reopening.

FabianLars:
@lysdexic-audio Please take a look at #2709 which is explicitly about BrowserView.

PR #8280, opened 2023-11-21
refactor(core): add support to multiple webviews on a Tauri window

lucasfernog:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### Other information

Currently the Tauri's `Window` type (both the Rust struct and JS class) are actually a mix of the `tao` Window and the `wry` Webview. In order to support multiple webviews on a given window, we actually have to split those types. This PR makes that change so now `WindowBuilder` and `Window` only exposes the window APIs, and we added `WebviewBuilder` and `Webview`.

To create a window with a webview attached to it, I've added [WindowBuilder::with_webview](https://github.com/tauri-apps/tauri/blob/41c2246d81b85f625aa3e8c3fb25b3bd95b0ed84/core/tauri/src/window/mod.rs#L265) on the Rust side and [WebviewWindow class](https://github.com/tauri-apps/tauri/blob/41c2246d81b85f625aa3e8c3fb25b3bd95b0ed84/tooling/api/src/webview.ts#L520) on the JS side. This reduces the breaking changes on the JS side (which I think it's the most common way of using multiwindow on Tauri).

To add webviews to a window, [Window::add_child](https://github.com/tauri-apps/tauri/blob/41c2246d81b85f625aa3e8c3fb25b3bd95b0ed84/core/tauri/src/window/mod.rs#L852) was added. The API isn't finished yet, so I propose we either hide it for v2 or keep it behind an `unstable` feature flag. Specially the auto resize thing needs more work and configuration.

Run `cargo run --example multiwebview` to play with it:
Screenshot 2023-11-21 at 16 57 36

Ref #2972 #2709 #2076


lucasfernog:
We could also add a WebviewWindow struct on the Rust side if we want to minimize breaking changes, but that comes it a lot of duplication between the builder and window/webview types.

amrbashir:
I think `WebviewWindow` is the correct way and I'd rather add it now than later because we already removed `WebviewWindow` in v2-alpha and would be annoying to bring it back in `v2.1` or `v3`

ruoxiaomo:
I'm excited about this! I hope this feature can be released soon, as there is a high demand for it, including myself. My application urgently needs this functionality to meet the needs of my users.

MessageDream:
截屏2023-11-29 15 17 59
I noticed in `core/tauri-runtime-wry/src/lib.rs` from line 2820 to 2831, there is code for adjusting the webview size along with the window size. I suggest considering the addition of a configuration option to control the execution of this code. If the configuration option is enabled, then execute this proportionally scaling code; if the configuration option is not enabled, then do not execute this code. The option could be enabled or disabled by default. This design can provide more flexibility to users, as sometimes we might not want to scale proportionally, such as the distance between two webviews, or the distance from the webview to the top or bottom.

deminearchiver:
This feature is so underrated. It has a bunch of uses. I plan to use it to separate a custom titlebar from the content.

Stuntlover-TM:
would love this to be in tauri

core/tauri-runtime/src/lib.rs
```diff
@@ -194,13 +195,20 @@ pub trait RuntimeHandle<T: UserEvent>: Debug + Clone + Send + Sync + Sized + 'st
   /// Creates an `EventLoopProxy` that can be used to dispatch user events to the main event loop.
   fn create_proxy(&self) -> <Self::Runtime as Runtime<T>>::EventLoopProxy;
 
-  /// Create a new webview window.
+  /// Create a new window.
   fn create_window<F: Fn(RawWindow) + Send + 'static>(
     &self,
     pending: PendingWindow<T, Self::Runtime>,
-    before_webview_creation: Option<F>,
+    before_window_creation: Option<F>,
```
amrbashir, line 217: this new name is incorrect and should now be renamed `after_window_creation` or removed.

boot2linux:
does this refactor support mobile platform ?

lucasfernog:
mobile isn't supported on wry

core/tauri-runtime/src/lib.rs
```diff
@@ -194,13 +195,20 @@ pub trait RuntimeHandle<T: UserEvent>: Debug + Clone + Send + Sync + Sized + 'st
...
   ) -> Result<DetachedWindow<T, Self::Runtime>>;
 
+  /// Create a new window.
```
amrbashir:
```suggestion
  /// Create a new webview.
```

core/tauri/src/menu/plugin.rs
```diff
@@ -80,7 +80,7 @@ enum Predefined {
   Hide,
   HideOthers,
   ShowAll,
-  CloseWindow,
+  CloseWebview,
```
amrbashir: Is this intentional? it should stay `CloseWindow`

core/tauri/src/menu/plugin.rs
```diff
@@ -769,7 +769,7 @@ fn set_accelerator<R: Runtime>(
 }
 
 #[command(root = "crate")]
-fn set_as_windows_menu_for_nsapp<R: Runtime>(
+fn set_as_webviews_menu_for_nsapp<R: Runtime>(
```
amrbashir: Another one

amr-crabnebula (approved):

amrbashir (approved):

thewh1teagle:
Opened related issue
https://github.com/tauri-apps/tauri/issues/8676

jkinggg:
Hi @lucasfernog, this is an awesome feature you've been working on. I've been playing around with it, and the only issue I've had is that it completely overlays the main tauri webview, so things can get hidden behind it. It makes things like managing drop downs and headers harder. Is it possible to position this webview so it sits under the main webview, enabling layering things on top?

nileshtrivedi:
Is there an example of adding a new child webview to the main window when processing a command?

PR #8288, opened 2023-11-22
feat(core): add AssetManager::iter

lucasfernog-crabnebula:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### Other information

This new function allows users to iterate on all embedded assets, important if you want to AssetManager::get an asset you are not sure exists.

### DIFF
--- a/core/tauri-utils/src/assets.rs
+++ b/core/tauri-utils/src/assets.rs
@@ -109,6 +109,9 @@ pub trait Assets: Send + Sync + 'static {
   /// Get the content of the passed [`AssetKey`].
   fn get(&self, key: &AssetKey) -> Option<Cow<'_, [u8]>>;
 
+  /// Iterator for the assets.
+  fn iter(&self) -> Box<dyn Iterator<Item = (&&str, &&[u8])> + '_>;
+
   /// Gets the hashes for the CSP tag of the HTML on the given path.
   fn csp_hashes(&self, html_path: &AssetKey) -> Box<dyn Iterator<Item = CspHash<'_>> + '_>;
 }
@@ -161,6 +164,10 @@ impl Assets for EmbeddedAssets {
       .get(key.as_ref())
       .copied()
       .map(|a| Cow::Owned(a.to_vec()))
+  }
+
+  fn iter(&self) -> Box<dyn Iterator<Item = (&&str, &&[u8])> + '_> {
+    Box::new(self.assets.into_iter())
   }
 
   fn csp_hashes(&self, html_path: &AssetKey) -> Box<dyn Iterator<Item = CspHash<'_>> + '_> {
--- a/core/tauri/src/manager.rs
+++ b/core/tauri/src/manager.rs
@@ -627,6 +627,10 @@ impl<R: Runtime> WindowManager<R> {
     })
   }
 
+  pub fn asset_iter(&self) -> Box<dyn Iterator<Item = (&&str, &&[u8])> + '_> {
+    self.inner.assets.iter()
+  }
+
   pub fn get_asset(&self, mut path: String) -> Result<Asset, Box<dyn std::error::Error>> {
     let assets = &self.inner.assets;
     if path.ends_with('/') {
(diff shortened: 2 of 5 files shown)


core/tauri/src/manager.rs
```diff
@@ -627,6 +627,10 @@ impl<R: Runtime> WindowManager<R> {
     })
   }
 
+  pub fn asset_iter(&self) -> Box<dyn Iterator<Item = (&&str, &&[u8])> + '_> {
+    self.inner.assets.iter()
+  }
```
amrbashir, line 632: I think we don't need to expose this one here, the `AssetResolver::iter` is enough

core/tauri-utils/src/assets.rs
```diff
@@ -163,6 +166,10 @@ impl Assets for EmbeddedAssets {
       .map(|a| Cow::Owned(a.to_vec()))
   }
 
+  fn iter(&self) -> Box<dyn Iterator<Item = (&&str, &&[u8])> + '_> {
+    Box::new(self.assets.into_iter())
```
amrbashir, line 170: Do we really need the box here? can't we implement `InotIterator` directly on the type and it construct a new type `AssetEntries` that is an iterator?

core/tauri/src/manager.rs
lucasfernog-crabnebula: this is private, it's just a helper otherwise we need to make more Manager fields public

core/tauri-utils/src/assets.rs
lucasfernog-crabnebula: it needs to be `dyn` because we're in the context of a trait.. so it needs Box too to be sized. Writing another type doesn't help this. It's also the solution used for csp_hashes.

core/tauri/src/manager.rs
amrbashir: right, I though this was on the `Manager` trait, this is fine then.

core/tauri-utils/src/assets.rs
amrbashir: If we return an `AssetsEntries` struct, we don't need to deal with `dyn` and can just implement `Iterator` on `AssetsEntries`
```rs
      fn iter<'a>(&'a self) -> AssetsEntries<'a, &'_ str, &'_ [u8]>;
```

or we can return `phf::map::Entries` directly
```rs
      fn iter<'a>(&'a self) -> phf::map::Entries<'a, &'_ str, &'_ [u8]>;
```

This should also be done for `csp_hashes` as well but to avoid breaking changes, we can delay that change to v2

lucasfernog-crabnebula: but then the AssetEntries need to hold the iterator anyway, unless we force the type to be a HashMap which isn't a good idea really for a trait

amrbashir: The only downside I see with this, is that `AssetsEntries` will need to use `phf::map::Entries` as a concrete type under the hood and now the `trait Assets` is not so generic and that makes me wonder, do we need this trait? seems like our usage of it, could be just replaced with the concrete `EmbeddedAssets` type

edit: GitHub didn't show me your previous message about the HashMap

lucasfernog-crabnebula: for tests it's nice to have this trait

but we can also put the author @chippers on the line

amrbashir: this is not a blocker btw, just would be a nice improvement for v2 (unless the trait is really required)

amrbashir (approved):

core/tauri-utils/src/assets.rs
chippers: Late on seeing this @amrbashir @lucasfernog-crabnebula , the asset trait allows people to implement their own asset handling, such as if they want to fetch them from disk or network. it's one of the original reasons it became possible to change the items in a `Context`

PR #8289, opened 2023-11-22
feat(core): add tracing for vital functionality closes #5204

### LINKED ISSUE #5204: [feat] Help for debugging updater JSON errors
### Describe the problem

When update process fails, Tauri's updater only fires an event, but doesn't log the error anywhere. I've used `"dialog": true` option to rely on the build-in UI for updates, so I did not build any event handling. I didn't even know that's a thing. As a result, I had no way of knowing why my updates aren't working. The built-in UI failed silently without reporting any errors anywhere.

I had to modify tauri itself to sprinkle `eprintln!` all over updater's code to discover the problem.

I'd expect it to print errors to stderr. Possibly if launched with `--verbose` or via `RUST_LOG=debug cargo run`.

### Alternatives considered

Having to build a custom error-reporting UI based on JS events is more work than I expected, especially given that no-code `dialog:true` option exists.

### Additional context

BTW, my problem with the update JSON was 

```text
SerdeJson(Error("invalid value for `pub_date`: a character literal was not valid", line: 0, column: 0))
```

which didn't like: `"pub_date": "2022-09-15T23:54:33+0100"`.

lucasfernog-crabnebula:

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### Other information

Adds tracing to the most important Tauri interfaces (IPC, custom protocol request handling, event system, updater, JS eval, window creation, plugins).

Needs https://github.com/tauri-apps/wry/pull/1093


core/tauri-runtime-wry/Cargo.toml
```diff
@@ -13,12 +13,13 @@ exclude = [ "CHANGELOG.md", "/target" ]
 readme = "README.md"
 
 [dependencies]
-wry = { version = "0.24.4", default-features = false, features = [ "file-drop", "protocol" ] }
+wry = { git = "https://github.com/crabnebula-dev/wry", branch = "feat/0.24-tracing", default-features = false, features = [ "file-drop", "protocol" ] }
```
JonasKruckenberg:
```suggestion
wry = { git = "https://github.com/tauri-apps/wry", default-features = false, features = [ "file-drop", "protocol" ] }
```

This shouldn't point to the fork 😉

JonasKruckenberg (changes requested):

core/tauri-macros/Cargo.toml
```diff
@@ -16,7 +16,7 @@ readme = "README.md"
 proc-macro = true
 
 [dependencies]
-proc-macro2 = "1"
+proc-macro2 = { version = "1.0.60", features = ["span-locations"] }
```
amrbashir:
```suggestion
proc-macro2 = { version = "1", features = ["span-locations"] }
```

amrbashir: or at the very least pin to `1.0`

core/tauri-runtime-wry/src/lib.rs
```diff
@@ -2590,14 +2625,16 @@ fn handle_user_message<T: UserEvent>(
       }
     }
     Message::Webview(id, webview_message) => match webview_message {
-      WebviewMessage::EvaluateScript(script) => {
+      WebviewMessage::EvaluateScript(script, tx, span) => {
+        let _guard = span.entered();
```
amrbashir:
```suggestion
        let _span = span.entered();
```

core/tauri-runtime-wry/src/lib.rs
```diff
@@ -2590,14 +2625,16 @@ fn handle_user_message<T: UserEvent>(
...
         if let Some(WindowHandle::Webview { inner: webview, .. }) =
           windows.borrow().get(&id).and_then(|w| w.inner.as_ref())
         {
           if let Err(e) = webview.evaluate_script(&script) {
             debug_eprintln!("{}", e);
           }
         }
+        tx.send(()).unwrap();
```
amrbashir, line 2667: I am not aware of the performance cost that we would get from this extra logic so I think we should add all tracing logic behind a feature flag (disabled by default)

core/tauri/src/lib.rs
```diff
@@ -625,6 +625,7 @@ pub trait Manager<R: Runtime>: sealed::ManagerBase<R> {
   ///   app.emit_all("synchronized", ());
   /// }
   /// ```
+  #[tracing::instrument("app::emit::all", skip(self, payload))]
```
amrbashir: Why are we skipping payload? shouldn't the payload be instrumented as well? to differentiate which emit call are we tracking?

amrbashir (changes requested):

core/tauri-runtime-wry/Cargo.toml
lucasfernog-crabnebula: it has to point to the fork until it's published

core/tauri-macros/Cargo.toml
lucasfernog-crabnebula: i think we need something on .60 right @JonasKruckenberg

core/tauri-runtime-wry/src/lib.rs
lucasfernog-crabnebula: it just has to wait for the message to arrive on the main thread so it shouldn't be that bad but we can do the flag

core/tauri/src/lib.rs
lucasfernog-crabnebula: payload doesn't implement Debug and doing so is a breaking change

lucasfernog-crabnebula: the payload is added on a child span though

amrbashir: I guess we should add a TODO then for v2

core/tauri-runtime-wry/src/lib.rs
amrbashir: yeah ideally the flag should be added for those who want to squeeze each last piece of performance

core/tauri-macros/Cargo.toml
CrabNejonas: Not that I'm aware of no

core/tauri-runtime-wry/Cargo.toml
CrabNejonas: Then lets merge the wry PR first, it should be easier to merge that one anyway

core/tauri-macros/Cargo.toml
lucasfernog-crabnebula: ok let's get it back to 1

core/tauri-runtime-wry/src/lib.rs
lucasfernog-crabnebula: @amrbashir @JonasKruckenberg I didn't test yet but I pushed so you can take a look.. tbh I think it bloats the code a lot.

amrbashir (dismissed):

core/tauri-runtime-wry/src/lib.rs
amrbashir: it does bloat it a bit but not that much IMO

amrbashir (dismissed):

JonasKruckenberg (approved):

PR #8292, opened 2023-11-23
fix(bundler): migrate WebView2 offline installer to shorturl

i-c-b:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### DIFF
--- a/tooling/bundler/src/bundle/windows/msi/wix.rs
+++ b/tooling/bundler/src/bundle/windows/msi/wix.rs
@@ -11,7 +11,7 @@ use crate::bundle::{
     sign::try_sign,
     util::{
       download, download_and_verify, extract_zip, HashAlgorithm, WEBVIEW2_BOOTSTRAPPER_URL,
-      WEBVIEW2_INSTALLER_X64_URL, WEBVIEW2_INSTALLER_X86_URL, WIX_OUTPUT_FOLDER_NAME,
+      WEBVIEW2_OFFLINE_INSTALLER_X64_URL, WEBVIEW2_OFFLINE_INSTALLER_X86_URL, WIX_OUTPUT_FOLDER_NAME,
       WIX_UPDATER_OUTPUT_FOLDER_NAME,
     },
   },
@@ -489,9 +489,9 @@ pub fn build_wix_app_installer(
         arch.to_uppercase()
       ));
       let webiew2_installer_url = if arch == "x64" {
-        WEBVIEW2_INSTALLER_X64_URL
+        WEBVIEW2_OFFLINE_INSTALLER_X64_URL
       } else {
-        WEBVIEW2_INSTALLER_X86_URL
+        WEBVIEW2_OFFLINE_INSTALLER_X86_URL
       };
       std::fs::write(&webview2_installer_path, download(webiew2_installer_url)?)?;
       data.insert("webview2_installer_path", to_json(webview2_installer_path));
--- a/tooling/bundler/src/bundle/windows/msi/wix.rs
+++ b/tooling/bundler/src/bundle/windows/msi/wix.rs
@@ -484,7 +484,7 @@ pub fn build_wix_app_installer(
       );
     }
     WebviewInstallMode::OfflineInstaller { silent: _ } => {
-      let webview2_installer_path = tauri_tools_path.join(format!(
+      let webview2_installer_path = output_path.join(format!(
         "MicrosoftEdgeWebView2RuntimeInstaller{}.exe",
         arch.to_uppercase()
       ));
--- a/tooling/bundler/src/bundle/windows/msi/wix.rs
+++ b/tooling/bundler/src/bundle/windows/msi/wix.rs
@@ -10,9 +10,9 @@ use crate::bundle::{
   windows::{
     sign::try_sign,
     util::{
-      download, download_and_verify, extract_zip, HashAlgorithm, WEBVIEW2_BOOTSTRAPPER_URL,
-      WEBVIEW2_OFFLINE_INSTALLER_X64_URL, WEBVIEW2_OFFLINE_INSTALLER_X86_URL, WIX_OUTPUT_FOLDER_NAME,
-      WIX_UPDATER_OUTPUT_FOLDER_NAME,
+      download, download_and_verify, extract_zip, webview2_guid_path, HashAlgorithm,
+      WEBVIEW2_BOOTSTRAPPER_URL, WEBVIEW2_OFFLINE_INSTALLER_X64_URL,
+      WEBVIEW2_OFFLINE_INSTALLER_X86_URL, WIX_OUTPUT_FOLDER_NAME, WIX_UPDATER_OUTPUT_FOLDER_NAME,
     },
   },
 };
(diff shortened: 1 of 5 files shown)


tooling/bundler/src/bundle/windows/msi/wix.rs
```diff
@@ -484,31 +484,16 @@ pub fn build_wix_app_installer(
...
-          &webview2_installer_path,
-          download(
-            &format!("https://msedge.sf.dl.delivery.mp.microsoft.com/filestreamingservice/files/{}/MicrosoftEdgeWebView2RuntimeInstaller{}.exe",
-              guid,
-              arch.to_uppercase(),
-            ),
-          )?,
-        )?;
-      }
+      std::fs::write(&webview2_installer_path, download(webiew2_installer_url)?)?;
```
amrbashir: why was the logic, to check if it was already downloaded, got removed?

amrbashir (changes requested):

tooling/bundler/src/bundle/windows/msi/wix.rs
i-c-b: It was based on the input URL which is now static. We can get the final URL from the download `response` with a separate function and determine if the GUID has changed that way but this seems overkill if we were already prepared to fetch the bootstrapper each time.

amrbashir: no no, we download the offline installer in `%LOCALAPPDATA%/tauri` directory, and we skip downloading it second time if this file exists

amrbashir: Ah wait, I see what you mean, we don't need to store it in `%LOCALAPPDATA%/tauri/guid/arch` anymore IMO, it can just be `%LOCALAPPDATA%/tauri/arch`

amrbashir: and then keep the same check, because the offline installer is 100mb iirc and downloading that each time is a no no.

i-c-b: A HEAD request would be sufficient to get the final URL and crawl the existing GUID paths. If we don't find it, then do the full GET to download.

amrbashir: we don't need the GUID anymore, we can skip it

i-c-b: Wouldn't skipping the GUID cause an outdated installer to be used until the `target` directory is cleaned up? It would be fine for end users in online environments because Edge would update itself but an outdated installer in an offline environment could lead to confusion.

amrbashir: yeah right, then I guess your suggesting about querying the final URL is the way to go

tooling/bundler/src/bundle/windows/msi/wix.rs
```diff
@@ -473,41 +473,33 @@ pub fn build_wix_app_installer(
       );
     }
     WebviewInstallMode::EmbedBootstrapper { silent: _ } => {
-      let webview2_bootstrapper_path = output_path.join("MicrosoftEdgeWebview2Setup.exe");
-      std::fs::write(
-        &webview2_bootstrapper_path,
-        download(WEBVIEW2_BOOTSTRAPPER_URL)?,
-      )?;
+      let (guid, filename) = webview2_guid_path(WEBVIEW2_BOOTSTRAPPER_URL)?;
```
amrbashir: this logic of downloading the installers seems to be repeated between nsis, and wix, let's move it into a function in `util.rs`

amrbashir (changes requested):

tooling/bundler/src/bundle/windows/util.rs
```diff
@@ -24,6 +24,15 @@ pub const NSIS_UPDATER_OUTPUT_FOLDER_NAME: &str = "nsis-updater";
 pub const WIX_OUTPUT_FOLDER_NAME: &str = "msi";
 pub const WIX_UPDATER_OUTPUT_FOLDER_NAME: &str = "msi-updater";
 
+pub fn webview2_arch_url(arch: &str) -> crate::Result<&str> {
+  let url = if arch == "x64" {
+    WEBVIEW2_OFFLINE_INSTALLER_X64_URL
+  } else {
```
i-c-b: Should `arm64` be included here? Microsoft provides a separate arm64 version of the offline installer.

amrbashir: as @FabianLars mentioned, windows 11 on arm uses x86 so we will stick with that for now I guess

tooling/bundler/src/bundle/windows/nsis.rs
```diff
@@ -370,40 +370,17 @@ fn build_nsis_app_installer(
 
   match webview2_install_mode {
     WebviewInstallMode::EmbedBootstrapper { silent: _ } => {
-      let webview2_bootstrapper_path = tauri_tools_path.join("MicrosoftEdgeWebview2Setup.exe");
-      std::fs::write(
-        &webview2_bootstrapper_path,
-        download(WEBVIEW2_BOOTSTRAPPER_URL)?,
-      )?;
+      let webview2_bootstrapper_path =
+        download_webview2(tauri_tools_path, WEBVIEW2_BOOTSTRAPPER_URL)?;
```
amrbashir: I think this can be just refactored to be
```rs
download_webview2_bootstrapper(tauri_tools_path);
```

Becasue the bootstrapper doesn't need to be stored in `<arch>/<guid>`, when invoked, it will always download the latest runtime, so it is okay if this file gets a bit outdated.

tooling/bundler/src/bundle/windows/nsis.rs
```diff
@@ -370,40 +370,17 @@ fn build_nsis_app_installer(
...
-            &format!("https://msedge.sf.dl.delivery.mp.microsoft.com/filestreamingservice/files/{}/MicrosoftEdgeWebView2RuntimeInstaller{}.exe",
-              guid,
-              arch.to_uppercase(),
-            ),
-          )?,
-        )?;
-      }
+      let url = webview2_arch_url(arch)?;
+      let base_path = tauri_tools_path.join(arch);
+      let webview2_installer_path = download_webview2(&base_path, url)?;
```
amrbashir: same here, this could be 
```rs
download_webview2_offline_installer(tauri_tools_path.join(arch));
```

amrbashir (changes requested):
I apologize for requesting a lot of changes but I think we are close to merge this

amrbashir (changes requested):
I apologize for requesting a lot of changes but I think we are close to merge this

tooling/bundler/src/bundle/windows/nsis.rs
i-c-b: Wouldn't splitting it into two functions with identical code except the URL be similar to the duplication we had before in NSIS and WiX?

amrbashir: I was only concerned about the duplication between nsis and wix but not between the two installer types download logic, and realistically the code that will be duplicated would probably just be the request call which is fine IMO.

amrbashir (approved):
Thanks, LGTM now

PR #8318, opened 2023-11-29
perf: skip large `Window` and `AppHandle` fields on traces

Beanow:

- [ ] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: performance

- [x] No

### Checklist

### Other information

These contain large fields like image buffers, causing spans/events to be very large when serialized.

Especially the `window::emit::eval` one which is in a hot code path.


Beanow:
Seems all the test failures are because of MSRV :grimacing:

Also (cc @JonasKruckenberg)

amrbashir (approved):

PR #8324, opened 2023-12-01
@tauri-apps/api  - Feature. Declaration merging.

veatla:
- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [x] No

### Checklist

#### I have added global interface called `TauriInvokeTypes `. 
btw I tried to create interface inside of `namespace` but I didn't figured out. Eslint throws errors :/

# How it work?
You can declare interface in global scope and set invoke types that will throw error if typeof assignable value or argument doesn't matches.

For example:
```ts
declare global {
  export interface TauriInvokeTypes {
    login: (args: { password: string; username: string }) => Promise<boolean>;
  }
}

```

Screenshot 12-02-2023 04 08 21


amrbashir:
Honestly, I don't think this is something we should add directly in the package to keep the types simple and reduce maintenance burden. Also the user can create a wrapper function around `invoke` and change the types as your proposed. but I will let @lucasfernog to decide

veatla:
Ok

lucasfernog:
I think this adds too much overhead. We're thinking about moving to tauri-bindgen in the future instead, and maybe implement [tauri-specta](https://github.com/oscartbeaumont/tauri-specta) in the meantime, we just didn't get to it yet.

veatla:
Okay

PR #8393, opened 2023-12-14
fix(runtime-wry): propagate wry file drop event to tao event loop, closes #8206

### LINKED ISSUE #8206: [bug] v2 file-drop doesn't trigger drop event listener in fs plugin
### Describe the bug

The fs plugin listens for a drop event to modify the fs scope. The event is fired by the listener never receives it.

### Reproduction

`cargo create-tauri-app --alpha`
`cargo add tauri-plugin-fs`
Insert `.plugin(tauri_plugin_fs::init())` before `.run` in `main.rs`.
Replace `main.js` with the following:
```js
const { getCurrent } = window.__TAURI__.window;
const { readDir } = window.__TAURI__.fs;
window.addEventListener("DOMContentLoaded", () => {
  getCurrent().onFileDropEvent(async (event) => {
    if (event.payload.type === "drop") {
      let entries = await readDir(event.payload.paths.paths[0], { recursive: true });
      console.debug("entries", entries);
    }
  });
});
```
`cargo tauri dev`

### Expected behavior

The event is received and the fs scope is adjusted to allow the path.

### Platform and versions

```text
[✔] Environment
    - OS: Linux Mint 21.2 X64
    ✔ webkit2gtk-4.1: 2.42.1
    ✔ rsvg2: 2.52.5
    ✔ rustc: 1.73.0 (cc66ad468 2023-10-03)
    ✔ cargo: 1.73.0 (9c4383fb5 2023-08-26)

[-] Packages
    - tauri [RUST]: 2.0.0-alpha.17
    - tauri-build [RUST]: 2.0.0-alpha.11
    - wry [RUST]: 0.34.2
    - tao [RUST]: 0.23.0
    - tauri-cli [RUST]: 2.0.0-alpha.17

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../src
    - devPath: ../src
```

### Stack trace

_No response_

### Additional context

_No response_

amrbashir:

- [X] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [X] No

### Checklist

### DIFF
--- a/.changes/runevent-filedrop.md
+++ b/.changes/runevent-filedrop.md
@@ -1,0 +1,6 @@
+---
+'tauri': 'patch:bug'
+'tauri-runtime-wry': 'patch'
+---
+
+Fix `RunEvent::WindowEvent(event: WindowEvent::FileDrop(FileDropEvent))` never triggered.
--- a/.changes/runevent-filedrop.md
+++ b/.changes/runevent-filedrop.md
@@ -3,4 +3,4 @@
 'tauri-runtime-wry': 'patch'
 ---
 
-Fix `RunEvent::WindowEvent(event: WindowEvent::FileDrop(FileDropEvent))` never triggered.
+Fix `RunEvent::WindowEvent(event: WindowEvent::FileDrop(FileDropEvent))` never triggered and always prevent default OS behavior when `disable_file_drop_handler` is not used.
--- a/core/tauri-runtime-wry/src/lib.rs
+++ b/core/tauri-runtime-wry/src/lib.rs
@@ -104,7 +104,6 @@ use std::{
 
 pub type WebviewId = u32;
 type IpcHandler = dyn Fn(String) + 'static;
-type FileDropHandler = dyn Fn(WryFileDropEvent) -> bool + 'static;
 
 mod webview;
 pub use webview::Webview;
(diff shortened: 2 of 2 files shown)


lucasfernog (approved):

core/tauri-runtime-wry/src/lib.rs
```diff
@@ -2734,10 +2735,21 @@ fn create_webview<T: UserEvent, F: Fn(RawWindow) + Send + 'static>(
...
-      .with_file_drop_handler(create_file_drop_handler(window_event_listeners.clone()));
+    let proxy = proxy.clone();
+    let window_event_listeners = window_event_listeners.clone();
+    webview_builder = webview_builder.with_file_drop_handler(move |event| {
+      let event: FileDropEvent = FileDropEventWrapper(event).into();
+      let _ = proxy.send_event(Message::Webview(
+        window_id,
+        WebviewMessage::WebviewEvent(WebviewEvent::FileDrop(event)),
+      ));
+      let has_listeners = !window_event_listeners.lock().unwrap().is_empty();
```
lucasfernog: I think we should always return true now, we can't know for sure if the event was handled or not (that was the case with the previous implementation too - which is also wrong on v1)

lucasfernog (changes requested):

core/tauri-runtime-wry/src/lib.rs
lucasfernog: unless... we add a new API (I'll take a look)

amrbashir: An API would be nicer indeed

lucasfernog: as I suspected we can't really determine this unless we create a closure just to handle the file drop event, and remove it from the event loop event enum. yikes. I think we should always block the default OS behavior instead.

.changes/runevent-filedrop.md
```diff
@@ -0,0 +1,6 @@
+---
+'tauri': 'patch:bug'
+'tauri-runtime-wry': 'patch'
+---
+
+Fix `RunEvent::WindowEvent(event: WindowEvent::FileDrop(FileDropEvent))` never triggered.
```
lucasfernog:
```suggestion
Fix `RunEvent::WindowEvent(event: WindowEvent::FileDrop(FileDropEvent))` never triggered and always prevent default OS behavior when `disable_file_drop_handler` is not used.
```

lucasfernog (approved):

PR #8402, opened 2023-12-14
fix(core): Replace Rc with Arc to prevent crashes when sending events

### LINKED ISSUE #8177: [bug] Event emit crashes app with high call rate
### Describe the bug

Too many calls to [`tauri::Window::emit`](<https://docs.rs/tauri/latest/tauri/window/struct.Window.html#method.emit>) in a short amount of time causes the app to crash.

### Reproduction

Initial testing was conducted using a [minimal reproduction](<https://github.com/i-c-b/tauri-download-test>) with the `upload` plugin and downloading the latest [daily Blender build](<https://builder.blender.org/download/daily/archive/>). It was later narrowed down to be related to the event system and an [event flooder](<https://github.com/i-c-b/tauri-event-flood>) was created which exhibits the same behaviour on a more consistent basis with 100,000 or more calls.

### Expected behavior

_No response_

### Platform and versions

```text
[✔] Environment
    - OS: Windows 10.0.22000 X64
    ✔ WebView2: 119.0.2151.44
    - node: 20.9.0
    - pnpm: 8.10.2
    - yarn: 1.22.19
    - npm: 10.1.0

[-] Packages
    - tauri [RUST]: 1.5.2
    - tauri-build [RUST]: 1.5.0
    - wry [RUST]: 0.24.4
    - tao [RUST]: 0.16.5

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../dist
    - bundler: Vite
```

### Stack trace

_No response_

### Additional context

This issue was previously discussed on [Discord](<https://discord.com/channels/616186924390023171/1166423822061285419>). @FabianLars demonstrated a mitigation for the `upload` plugin by reducing the frequency of calls to `emit` on the [`download-throttled-events` branch](<https://github.com/tauri-apps/plugins-workspace/tree/download-throttled-events/plugins/upload>) which led to reduced crashing but didn't eliminate it entirely. The impact this issue has on the `upload` plugin would be reduced in environments with slow drives and fast networks as there is more time between messages and fewer overall messages.

FabianLars:
I feel like this is more of a workaround than a fix but this reduced the crashes on macos for me. That said, the issue i was trying to fix was on windows but when i worked on this commit i couldn't repro it with and without this change. The issue seems to be influenced by available system resources or something because it's not the first time i wasn't able to reproduce it.

This _may_ close #8177 but i'd like @i-c-b and/or @JJeris to try it out (since they were able to reproduce it more reliably than me).

p.s. i don't think i'm capable enough of doing so but we should still find out what's _actually_ happening here or alternatively change the types so that clippy doesn't complain about Arc when it's actually needed (if possible).


JJeris:
What was this workaround? The iterator or the new thing discovered by the other user?
Ill test it out today, if i can.

i-c-b:
This seems to fix the issue for Linux. On Windows, it seems that emitting more than 12725 events at time causes `emit` to fail with `runtime error: failed to send message to the webview`; unsure what's going on there.

[event-flooder](<https://github.com/i-c-b/tauri-event-flood>) crashes with v1.5.2, v1.5.3, and the bleeding-edge 1.x branch which includes the clipboard changes. All had nearly instant crashes when I tried 1,000,000 events. The moment I switched to the `event-crash` branch, the crashing stopped and I could no longer reproduce it no matter how many events I emitted.

> What was this workaround? The iterator or the new thing discovered by the other user?

In Tauri v1.5.0, the event system switched from `std::sync::Arc` to `std::rc::Rc`. This workaround reverts that change as a (hopefully) temporary measure while a more permanent solution is found. The main difference is that Arc (Atomically Reference Counted) is thread-safe and Rc (Reference Counted) isn't.

lucasfernog:
Windows issue is `Tried to wake up a closed EventLoop`. I'm investigating.

lucasfernog (approved):

PR #8420, opened 2023-12-16
fix(core): scope checks on Android

lucasfernog:

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] No

### Checklist

### Other information

On Android, when we call canonicalize() on "/data/user/0/appid" (which is the data dir), the result is a "/data/data/appid" path, so we need to adjust our scope for that.

### DIFF
--- a/core/tauri/src/scope/fs.rs
+++ b/core/tauri/src/scope/fs.rs
@@ -79,55 +79,54 @@ fn push_pattern<P: AsRef<Path>, F: Fn(&str) -> Result<Pattern, glob::PatternErro
 ) -> crate::Result<()> {
   let path: PathBuf = pattern.as_ref().components().collect();
   list.insert(f(&path.to_string_lossy())?);
-  #[cfg(any(windows, target_os = "android"))]
-  {
-    let mut path = path;
-    let mut checked_path = None;
 
-    // attempt to canonicalize parents in case we have a path like `/data/user/0/appid/**`
-    // where `**` obviously does not exist but we need to canonicalize the parent
-    //
-    // example: given the `/data/user/0/appid/assets/*` path,
-    // it's a glob pattern so it won't exist (canonicalize() fails);
-    //
-    // the second iteration needs to check `/data/user/0/appid/assets` and save the `*` component to append later.
-    //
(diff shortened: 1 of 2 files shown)


core/tauri/src/scope/fs.rs
```diff
@@ -79,11 +79,33 @@ fn push_pattern<P: AsRef<Path>, F: Fn(&str) -> Result<Pattern, glob::PatternErro
...
+      if let Some(mut p) = last {
+        path.pop();
+        if let Some(buf) = &buf {
+          p.push(buf);
+        }
+        buf.replace(p);
+      } else {
+        break None;
+      }
+    };
```
amrbashir: I'm having a hard time following this code logic tbh so might help adding more comments explaining a bit

core/tauri/src/scope/fs.rs
```diff
@@ -79,11 +79,33 @@ fn push_pattern<P: AsRef<Path>, F: Fn(&str) -> Result<Pattern, glob::PatternErro
 ) -> crate::Result<()> {
   let path: PathBuf = pattern.as_ref().components().collect();
   list.insert(f(&path.to_string_lossy())?);
-  #[cfg(windows)]
+  #[cfg(any(windows, target_os = "android"))]
```
amrbashir: I don't think windows needs all this extra logic, can we just leave this logic for android only and keep windows logic separate?

amrbashir (changes requested):

core/tauri/src/scope/fs.rs
lucasfernog: actually all targets needs this, because we always canonicalize() the path when checking it on is_allowed and if the path has symlinks we need to resolve that here too (unless we find a better way to check paths)

core/tauri/src/scope/fs.rs
lucasfernog: pushed

amrbashir (approved):

PR #8428, opened 2023-12-18
Tauri ACL/Allowlist v2 Implementation and Plugin System Refactor

tweidinger:
This PR contains several (breaking) changes related to:
- Tauri plugins
- The new allowlist
- Tauri build process

The current proposal and extended documentation for these changes can be found in a public HackMD document [here](https://hackmd.io/@secdude/BJQS7u_rn?view). This document will be undergoing several changes during the implementation but shows the general approach and architecture. **We strongly recommend reading this document before commenting.**

We currently believe this is the one big major change needed to make v2.0 feature complete for a beta/rc/release,
therefore work on this PR should be prioritized.

### Tauri Plugins

In v2 of Tauri most of the commands were moved into dedicated plugins and outside of the core of Tauri.
The move to plugins rendered the existing allowlist unusable, as the system was not designed to be compatible with plugins or code outside of Tauri. This PR addresses this by adding a new allowlist and access control for Tauri commands written by app or plugin developers.

The [plugin-workspace](https://github.com/tauri-apps/plugins-workspace) repository contains all of the previous (and more) functionality now available as plugins (see the v2 branch). This reduces the amount of complexity in Tauri itself. Additionally, it allows more fast paced changes of system interaction functionality, as no complete understanding of Tauri itself is needed to contribute. It also allows breaking releases/changes to specific plugins independently from Tauri.

To further define what Tauri plugins are and what they need to implement the `tauri-plugin` crate was created.
It can be found in `core/tauri-plugin` in this PR branch.
Plugin developers need to depend on this crate and implement the defined traits, define default permissions and optionally define scope types.

A lot of work (not tracked here) is required to upgrade plugins from v1 to v2 and convenience tooling like `tauri plugin init`, `tauri permission init/add/remove` needs to be created, while existing tooling like `tauri-cli` and `create-tauri-app` need to be changed to be compatible with the new permissions/capabilities/plugins.

### Allowlist

In v1 of Tauri most of the commands were Tauri internal APIs built into the core of Tauri. The [allowlist](https://tauri.app/v1/api/config/#allowlistconfig) was used to restrict the access from the Webview to the Tauri core and system resources.

With this PR the new allowlist is built in a way to be used by all plugins and application developers and is no longer exclusive for Tauri inbuilt functionality. It also allows more fine grained control, while being able to abstract away a lot of things.

We introduce several new naming conventions and move the allowlist to a capability driven configuration. Permissions define command enablement and scope, while capabilities link permissions with windows of the application.

A simplified example allowlist/capabilities configuration could look like:

```json
"capabilities": [
    {
        "context": "local",
        "windows": [
            "*"
        ],
        "permissions": [
            "fs:full-homefolder-access",
            "fs:block-homefolder-sensitive-access"
        ]
    }
]
```

with an example permission, which could also be inbuilt/defined by the `fs` plugin:

```json
{
    "version": 1,
    "identifier": "full-homefolder-access",
    "description": "This allows read write access to the complete $HOME folder.",
    "commands": {
        "allow": [
            "fs:readDirectory",
            "fs:readFile"
        ]
    },
    "scope": {
        "allow": [
            {
                "path": "$HOME/**"
            }
        ]
    }
}
```

or even simpler when the capabilities are further abstracted:

```json
"capabilities": [
        "default",
        "admin-windows"
]
```

### Tauri Build System

The build system will be enhanced (most likely) in a non-breaking way to understand metadata information provided by plugins and will allow us to pass information from plugins into the Tauri application. 
It also will allow us to highlight issues to the application developer when commands are enabled in the capabilities but the plugin implementing these commands is not initialized/used. Also the other way around when no commands of a plugin are enabled but the plugin is initialized.

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [x] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

- [x] Yes
- [ ] No

### Checklist

### Other information

Related issues:
-  #7461 the new v2 implementation will deprecate the `dangerousRemoteIPCAccess` as it will be part of the capability definition


core/tauri-plugin/src/build/mod.rs
```diff
@@ -0,0 +1,82 @@
+use cargo_metadata::{Metadata, MetadataCommand};
+use tauri::utils::acl::{Permission, PermissionSet};
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum Error {
+  #[error("expected build script env var {0}, but it was not found - ensure this is called in a build script")]
+  BuildVar(String),
```
lucasfernog: could use a &'static str here

core/tauri-plugin/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+authors = { workspace = true }
+homepage = { workspace = true }
+repository = { workspace = true }
+categories = { workspace = true }
+license = { workspace = true }
+edition = { workspace = true }
+rust-version = { workspace = true }
+
+[features]
+build = ["dep:anyhow", "dep:cargo_metadata", "dep:thiserror"]
```
lucasfernog:
```suggestion
build = ["dep:anyhow", "dep:cargo_metadata", "dep:thiserror"]
runtime = []
```

core/tauri-utils/src/acl/capability.rs
```diff
@@ -0,0 +1,35 @@
...
+  permissions: Vec<PermissionId>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum CapabilityContext {
+  Local,
+  Remote { dangerous_remote: Vec<String> },
+}
+
+impl Default for CapabilityContext {
```
lucasfernog: could use https://doc.rust-lang.org/std/default/trait.Default.html#enums instead

core/tauri-utils/src/acl/resolved.rs
```diff
@@ -0,0 +1,45 @@
+use matchit::Router;
+use serde::de::DeserializeOwned;
+use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
+
+/// TODO: the current proof of concept uses [`matchit`], but we should use a simplified
+/// router router that __only__ supports static and wildcard values (without wildcard params)
```
lucasfernog:
```suggestion
/// router that __only__ supports static and wildcard values (without wildcard params)
```

lucasfernog-crabnebula:
We have states now!

```toml
[[permission]]
identifier = "global-scope"
description = "This sets a global scope."
[permission.scope.allow]
path = "global"
```

```toml
[[permission]]
identifier = "allow-ping-scoped"
description = "This enables the ping command with a test scope."
commands.allow = ["ping"]
[permission.scope.allow]
path = "x"
```

```rust
#[allow(dead_code)]
#[derive(Debug, Deserialize)]
struct PingScope {
  path: PathBuf,
}

#[allow(dead_code)]
#[derive(Debug, Deserialize)]
struct SampleScope {
  path: PathBuf,
}

#[tauri::command]
fn ping<R: tauri::Runtime>(
  app: tauri::AppHandle<R>,
  value: Option<String>,
  scope: tauri::command::CommandScope<PingScope>,
  global_scope: tauri::command::GlobalScope<SampleScope>,
) -> std::result::Result<PingResponse, String> {
  println!("local scope {:?}", scope);
  println!("global scope {:?}", global_scope);
  todo!()
}
```

I'm not sure how useful command-specific scopes are going to be (probably too confusing for users?) but it's there

chippers (dismissed):
This looks good as a basis for the ACL. There are some improvements I have in a project for this rep that we ideally land before stable, but I think we can mark this PR as complete and a good MVP

chippers (dismissed):

chippers (dismissed):

lucasfernog (approved):

PR #8439, opened 2023-12-19
fix(cli): expand globs in workspace member paths

### LINKED ISSUE #8403: dev command does not correctly watch glob member patterns
### Describe the bug

Running `cargo tauri dev` with this Cargo.toml config:

```toml
[workspace]
resolver = "2"
members = ["bin/iron", "crates/*"]
```

live reloading does not work for anything in `crates/*`. Manually adding one explicit item, such as:

```
members = ["bin/iron", "crates/settings", "crates/*"]
```

makes it work again for that one, leading me to think this is the root cause

### Reproduction

Create a tauri app using a cargo workspace where `workspace.members` uses a glob pattern.
try to edit a file in that glob while developing. live reloading should fail to trigger

### Expected behavior

Live reloading should catch all rust members of the workspace

### Platform and versions

```text
❯ cargo tauri info

[✔] Environment
    - OS: Manjaro 23.1.0 X64
    ✔ webkit2gtk-4.0: 2.42.3
    ✔ rsvg2: 2.57.0
    ✔ rustc: 1.74.1 (a28077b28 2023-12-04)
    ✔ cargo: 1.74.1 (ecb9851af 2023-10-18)
    ✔ rustup: 1.26.0 (2023-11-14)
    - node: 20.9.0
    - yarn: 4.0.2
    - npm: 10.1.0

[-] Packages
    - tauri [RUST]: 1.5.2 (no manifest)
    - tauri-build [RUST]: 1.5.0 (no manifest)
    - wry [RUST]: 0.24.6 (no manifest)
    - tao [RUST]: 0.16.5 (no manifest)
    - tauri-cli [RUST]: 1.5.7
    - @tauri-apps/cli [NPM]: 1.5.7 (outdated, latest: 1.5.8)
```

### Stack trace

_No response_

### Additional context

_No response_

FabianLars:
fixes #8403

Edit: Alternatively we could also re-use the output of cargo metadata but i didn't want to bother trying to parse `"my-package 0.1.0 (path+file:///path/to/my-package)"` unless you think it'd be better than trying to be cargo again.

### DIFF
--- a/tooling/cli/src/interface/rust.rs
+++ b/tooling/cli/src/interface/rust.rs
@@ -19,8 +19,10 @@ use std::{
 };
 
 use anyhow::Context;
+use glob::glob;
 use heck::ToKebabCase;
 use ignore::gitignore::{Gitignore, GitignoreBuilder};
+use itertools::Itertools;
 use log::{debug, error, info};
 use notify::RecursiveMode;
 use notify_debouncer_mini::new_debouncer;
@@ -334,6 +336,18 @@ fn lookup<F: FnMut(FileType, PathBuf)>(dir: &Path, mut f: F) {
   }
 }
 
+// Copied from https://github.com/rust-lang/cargo/blob/69255bb10de7f74511b5cef900a9d102247b6029/src/cargo/core/workspace.rs#L665
+fn expand_member_path(path: &Path) -> crate::Result<Vec<PathBuf>> {
+  let Some(path) = path.to_str() else {
+    return Ok(Vec::new());
+  };
+  let res = glob(path).with_context(|| format!("could not parse pattern `{}`", &path))?;
+  let res = res
+    .map(|p| p.with_context(|| format!("unable to match path to pattern `{}`", &path)))
+    .collect::<Result<Vec<_>, _>>()?;
+  Ok(res)
+}
+
 impl Rust {
   fn run_dev<F: Fn(ExitStatus, ExitReason) + Send + Sync + 'static>(
     &mut self,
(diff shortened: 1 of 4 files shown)


tooling/cli/src/interface/rust.rs
```diff
@@ -420,7 +433,21 @@ impl Rust {
         .unwrap_or_else(|| vec![tauri_path])
     };
 
+    let watch_folders = watch_folders
+      .iter()
```
amrbashir:
```suggestion
      .into_iter()
```

Should remove the need for `.to_owned` later, no?

tooling/cli/src/interface/rust.rs
```diff
@@ -334,6 +335,18 @@ fn lookup<F: FnMut(FileType, PathBuf)>(dir: &Path, mut f: F) {
   }
 }
 
+// Copied from https://github.com/rust-lang/cargo/blob/69255bb10de7f74511b5cef900a9d102247b6029/src/cargo/core/workspace.rs#L665
+fn expand_member_path(path: &Path) -> crate::Result<Vec<PathBuf>> {
+  let Some(path) = path.to_str() else {
+    return Ok(Vec::new());
```
amrbashir: Seems like an unnecessary allocation, either return an Option or Err here, we catch it anyways.

amrbashir (changes requested):

tooling/cli/src/interface/rust.rs
FabianLars: first time using rust.

tooling/cli/src/interface/rust.rs
FabianLars: man, let me lazily copy paste for once 😂

amrbashir: naughty fabian 😏

amrbashir (dismissed):

amrbashir (approved):

PR #8321, opened 2023-11-30
feat(cli): improve error message with invalid target for `build`

zacharyftw:
### What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

### Does this PR introduce a breaking change?

- [ ] Yes, and the changes were approved in issue #___
- [x] No

### Checklist
- [ ] When resolving issues, they are referenced in the PR's title (e.g `fix: remove a typo, closes #___, #___`)
- [ ] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md).
- [ ] I have added a convincing reason for adding this feature, if necessary

PR #8448, opened 2023-12-20
fix(cli): Handle target triples with 4 components

FabianLars:
Follow up of #8321 which broke the `TAURI_ENV_` vars for linux, windows and ios-sim

p.s. there are 2 targets, fuchsia and wasi that only have 2 components but ig we don't need to care about those for now.

### DIFF
--- a/tooling/cli/src/interface/rust.rs
+++ b/tooling/cli/src/interface/rust.rs
@@ -230,8 +230,11 @@ impl Interface for Rust {
 
     let target_triple = &self.app_settings.target_triple;
     let target_components: Vec<&str> = target_triple.split('-').collect();
-    let (arch, host) = match target_components.as_slice() {
-      [arch, _, host] => (*arch, *host),
+    let (arch, host, host_env) = match target_components.as_slice() {
+      // 3 components like aarch64-apple-darwin
+      [arch, _, host] => (*arch, *host, None),
+      // 4 components like x86_64-pc-windows-msvc and aarch64-apple-ios-sim
+      [arch, _, host, host_env] => (*arch, *host, Some(*host_env)),
       _ => {
         log::warn!("Invalid target triple: {}", target_triple);
         return env;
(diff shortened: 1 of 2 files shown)


tooling/cli/src/interface/rust.rs
```diff
@@ -248,12 +251,12 @@ impl Interface for Rust {
...
-      match host {
+      match (host, host_env) {
         // keeps compatibility with old `std::env::consts::OS` implementation
-        "darwin" => "macos".into(),
-        "ios-sim" => "ios".into(),
-        "androideabi" => "android".into(),
-        h => h.into(),
+        ("darwin", _) => "macos".into(),
+        ("ios", Some("sim")) => "ios".into(),
+        ("androideabi", _) => "android".into(),
```
amrbashir: lets' remove these compatibilities and use them as is.

amrbashir (changes requested):

tooling/cli/src/interface/rust.rs
FabianLars: did you mean like this? https://github.com/tauri-apps/tauri/pull/8448/commits/c3f12cd7851efa8b9b5b78f482691b47cff73aa4

amrbashir: yeah, except maybe keep the conversion from `ios-sim` -> `ios`, what do you think @lucasfernog ?

FabianLars: yeah, i left that one in. By splitting it into host and host_env this is done above already. I also didn't see much value in a seperate ios-sim target (unless androideabi is only used in sims?)

amrbashir: Ah nice, didn't notice that. if anyone wants `-sim` part, they can parse `TAURI_ENV_TARGET_TRIPLE` themselves.

amrbashir (approved):

lucasfernog (approved):

lucasfernog:
Test failures aren't part of this PR so i'll open a separate PR for it

PR #8474, opened 2023-12-27
feat: re-export `Url`

amrbashir:
`Url` is used/returned from public API, we should re-export it


.changes/export-url.md
```diff
@@ -0,0 +1,5 @@
+---
+'tauri': 'patch'
```
amrbashir:
```suggestion
'tauri': 'patch:feat'
```

lucasfernog (approved):

PR #8497, opened 2023-12-29
feat: add `libraries` flag to `MacConfig` (1.x)

dongcarl:
For macOS apps leveraging `SMAppService`, we need to place a codesigned `plist` under the app's `Contents/Library/LaunchDaemons` directory. This is outside of the `Contents/Resources` directory so we can't use the `resources` flag.

The PR as it stands adds a `libraries` flag to `MacConfig` to allow this

### DIFF
--- a/core/tauri-config-schema/schema.json
+++ b/core/tauri-config-schema/schema.json
@@ -1380,6 +1380,16 @@
             "string",
             "null"
           ]
+        },
+        "libraries": {
+          "description": "MacOS library files that need to be bundled with the app.",
+          "type": [
+            "object",
+            "null"
+          ],
+          "additionalProperties": {
+            "type": "string"
+          }
         }
       },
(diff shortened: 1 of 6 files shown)


FabianLars:
Thanks for contributing! We recently merged a similar PR into dev (v2) here https://github.com/tauri-apps/tauri/pull/7798 and i think it'd make more sense to backport this / match its name and behavior :)
( -> i'm not sure that a simple git cherry pick would be easy so _i_ would be fine if it's recreated instead 🤷 )

Also, we require all commits to be signed to merge a PR. Here's a guide for that: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#ssh-commit-signature-verification - i recommend the ssh approach because i think it's easier but a guide for gpg is right above that though.

PR #8520, opened 2024-01-01
fix(cli): Watch workspace members if tauri dir is workspace root

### LINKED ISSUE #8509: [bug] After tuari project dependency is installed, running `pnpm tauri dev` cannot run,
### Describe the bug

After tuari project dependency is installed, running `pnpm tauri dev` cannot run,
report: thread '<unnamed>' panicked at src\interface\rust.rs:451:79: called `Option::unwrap()` on a `None` value stack backtrace: note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace. * The terminal will be reused by the task, press any key to close.

### Reproduction

_No response_

### Expected behavior

Can operate normally

### Full `tauri info` output

```text
thread '<unnamed>' panicked at src\interface\rust.rs:451:79: called `Option::unwrap()` on a `None` value stack backtrace: note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace. * The terminal will be reused by the task, press any key to close.
```

### Stack trace

_No response_

### Additional context

_No response_

FabianLars:
See title. This PR also includes a fix/workaround for paths with funny characters that may not make the glob expansion panic.

Fixes #8509 -> That's about the funny characters and the reason i looked into the code again in the first place.

### DIFF
--- a/tooling/cli/src/interface/rust.rs
+++ b/tooling/cli/src/interface/rust.rs
@@ -347,6 +347,43 @@ fn expand_member_path(path: &Path) -> crate::Result<Vec<PathBuf>> {
   Ok(res)
 }
 
+fn get_watch_folders(tauri_path: &Path, workspace_path: &Path) -> crate::Result<Vec<PathBuf>> {
+  // We always want to watch the main tauri folder.
+  let mut watch_folders = vec![tauri_path.to_path_buf()];
+
+  // We also try to watch workspace members, no matter if the tauri cargo project is the workspace root or a workspace member
+  let cargo_settings = CargoSettings::load(workspace_path)?;
+  if let Some(folders_iter) = cargo_settings.workspace.as_ref().map(|w| {
+    w.members
+      .clone()
+      .unwrap_or_default()
+      .into_iter()
+      .flat_map(|p| {
+        let p = workspace_path.join(p);
+        match expand_member_path(&p) {
+          // Sometimes expand_member_path returns an empty vec, for example if the path contains `[]` as in `C:/[abc]/project/`.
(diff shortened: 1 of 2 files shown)


FabianLars:
thinking about it i'm not sure anymore if it actually made a difference since it should watch the workspace members anyway since they are inside the root (if tauri dir is ws root) but ig being mroe explicit about it is good? 🙃

tooling/cli/src/interface/rust.rs
```diff
@@ -415,40 +415,43 @@ impl Rust {
...
-    let watch_folders = watch_folders
-      .into_iter()
-      .flat_map(|p| {
-        match expand_member_path(&p) {
-          Ok(p) => p,
-          Err(err) => {
-            // If this fails cargo itself should fail too. But we still try to keep going with the unexpanded path.
-            error!("Error watching {}: {}", p.display(), err.to_string());
-            vec![p]
+    // We also try to watch workspace members, no matter if the tauri cargo project is the workspace root or a workspace member
```
amrbashir: this logic now needs a function on its own IMO.

amrbashir (changes requested):

tooling/cli/src/interface/rust.rs
FabianLars: Just the "get workspace member paths" part or the whole "get watch folders" block? It's only a one line difference so excuse the stupid question but i don't wanna do it twice :D

amrbashir: hard to say tbh, just go with your gut, I am fine with either way

amrbashir (approved):

Issue #8519, opened 2024-01-01
[bug] V2 Tauri custom titlebar unable to resize when decorations is false


Sir-Thom:
### Describe the bug

I was trying v2 of tauri and I noticed that if you set decorations to false in the tauri.conf.json  the ability to resize the application is impossible even if you set resizable to true.

### Reproduction

1.  `npm create tauri-app@latest -- --alpha`
2.  add the demo titlebar to the project :
3.  make those configuration in the tauri.conf.json : 
```
{
  "build": {
    "beforeDevCommand": "yarn dev",
    "beforeBuildCommand": "yarn build",
    "devPath": "http://localhost:1420",
    "distDir": "../dist",
    "withGlobalTauri": true
  },
  
  "package": {
    "productName": "tauri-app",
    "version": "0.0.0"
  },
  "tauri": {
    "bundle": {
      "active": true,
      "targets": "all",
      "identifier": "com.tauri.dev",
      "icon": [
        "icons/32x32.png",
        "icons/128x128.png",
        "icons/128x128@2x.png",
        "icons/icon.icns",
        "icons/icon.ico"
      ]
    },
    "security": {
      "csp": null
    },
    
    "windows": [
      {
        "decorations": false,
        "fullscreen": false,
        "resizable": true,
        "title": "tauri-app",
        "width": 800,
        "height": 600
      }
    ]
  },
  "plugins": {
    "shell": {
      "open": true
    },
    "allowlist":{
      "all": true,
      "window": {
        "center": false,
        "close": true,
        "create": false,
        "hide": false,
        "maximize": true,
        "minimize": true,
        "print": false,
        "requestUserAttention": false,
        "setAlwaysOnTop": false,
        "setCursorGrab": true,
        "setCursorIcon": true,
        "setCursorPosition": false,
        "setCursorVisible": false,
        "setDecorations": false,
        "setFocus": true,
        "setFullscreen": true,
        "setIcon": true,
        "setMaxSize": false,
        "setMinSize": false,
        "setPosition": false,
        "setResizable": true,
        "setSize": true,
        "setSkipTaskbar": false,
        "setTitle": true,
        "show": true,
        "startDragging": true,
        "unmaximize": true,
        "unminimize": true

      }
    }
  }
}
```

### Expected behavior

the application is able to be resize when decorations is set to false

### Full `tauri info` output

```text
[✔] Environment
    - OS: openSUSE 20231226 X64
[...]

PR #8537, opened 2024-01-04
fix(core): fix undecorated window resizing,  #8519

amrbashir:


tooling/api/src/window.ts
```diff
@@ -45,6 +45,16 @@ export interface Monitor {
 type Theme = 'light' | 'dark'
 type TitleBarStyle = 'visible' | 'transparent' | 'overlay'
 
+type ResizeDirection =
+  | 'East'
```
amrbashir, line 49: should it be camelCase?

FabianLars: I think the current casing makes sense for directions even though it'll be different from the other defined types. Not sure i'd like to rename the other types though so maybe a mix is fine?
Or maybe we should add an alias on the rust side so non ts users won't get a stroke?

FabianLars (dismissed):

FabianLars (approved):

PR #8539, opened 2024-01-04
fix(core): Command::output suspend while wait for response

dyxushuai:
https://github.com/tauri-apps/tauri/blob/67d7877f27f265c133a70d48a46c83ffff31d571/core/tauri/src/api/process/command.rs#L273-L286

These two methods will hold the `tx(Sender)` and never drop them because the method `spawn_pipe_reader ` will be lost in the loop and unable to get out without `break;` 

https://github.com/tauri-apps/tauri/blob/67d7877f27f265c133a70d48a46c83ffff31d571/core/tauri/src/api/process/command.rs#L401-L425

So, the loop in `output` method will never receive a `None` value from `Reciver`.

https://github.com/tauri-apps/tauri/blob/67d7877f27f265c133a70d48a46c83ffff31d571/core/tauri/src/api/process/command.rs#L362-L377


dyxushuai:
Maybe resolves #8572

amrbashir (approved):

PR #8586, opened 2024-01-11
chore: Commit Cargo.lock

FabianLars:
If i have to deal with msrv ci issues caused by unrelated dependency updates one more time i will actually lose my mind

p.s. we've had the lockfile in plugins-workspace since day 1 without any issues :)


.github/workflows/test-core.yml
```diff
@@ -82,47 +82,5 @@ jobs:
           workspaces: core -> ../target
           save-if: ${{ matrix.features.key == 'all' }}
 
-      - name: Downgrade crates with MSRV conflict
```
amrbashir, line 85: This was serving as a reference to those who want keep the MSRV, do we have an alternative after this is removed?

FabianLars: Other than the commit link to that, no. I hoped that'd be enough since as far as i've gathered nobody actually cared.

Alternatively we could add another workflow that runs every 2-4 weeks or so or make the current workflow remove the lock file on the dev branch so it's a bit more obvious to spot. Or maybe as part of the covector PR makes probably the most sense?
As long as it doesn't keep breaking PRs i don't really care, we could also add it to the faq as a starting point and update it whenever we feel like it.

amrbashir: We can run it before covector publish, similar to the integration tests

FabianLars: like this?

amrbashir: yeah

amrbashir (dismissed):

lucasfernog (dismissed):

FabianLars:
how the hell are the tests failing now, i didn't touch them 😭

lucasfernog:
arboard crate broke it. i'm checking

FabianLars:
ohhhh, i may have used the wrong feature flags when i created the lockfile which would maybeeee explain the error in the first failed workflow 🤔

lucasfernog (approved):

PR #8621, opened 2024-01-17
refactor(core): refactor and fix event system following multiwebview support

### LINKED ISSUE #8435: [bug] The CloseRequestedEvent cannot prevent the closing of a dynamically created window.
### Describe the bug

I created a new `WebviewWindow (B)` at runtime within an existing `WebviewWindow (A)` and set up a listener for the `CloseRequestedEvent` of the new webview, intending to execute certain actions before it closes. However, despite calling `preventDefault()`, it doesn't seem to work. Here is my code:
```
<template>
  <div class="container">
    <p>NewWindow</p>
    <el-button @click="createNewWindow">Create</el-button>
  </div>
</template>
<script setup lang="ts">
import { WebviewWindow } from "@tauri-apps/api/window";
const createNewWindow = async () => {
  const webview = new WebviewWindow("media-login-window", {
    url: "https://google.com",
  });
  webview.once("tauri://created", function () {
    // webview window successfully created
    console.log("tauri created");
  });
  webview.once("tauri://error", function (e) {
    console.log("window create error", e);
    // an error occurred during webview window creation
  });

  await webview.onCloseRequested(async (event) => {
    console.log("tauri close-requested", event);
    event.preventDefault();
    // i want do something here, but the new webview closed.
  });
};
</script>
```
While debugging the source code, I found that in the `on_window_event` function, the window instance for checking `has_js_listener` is the newly created `WebviewWindow(B)`. However, the window instance used for registering the event listener is the already existing `WebviewWindow(A)`. Is this a bug, or is there an issue with my usage?
image
image

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.22000 X64
    ✔ WebView2: 114.0.1823.51
    ✔ MSVC: Visual Studio ���ɹ��� 2019
    ✔ rustc: 1.74.1 (a28077b28 2023-12-04)
    ✔ cargo: 1.74.1 (ecb9851af 2023-10-18)
    - node: 20.9.0
    - yarn: 1.22.19
    - npm: 10.1.0

[-] Packages
    - tauri [RUST]: 1.5.3
    - tauri-build [RUST]: 1.5.0
    - wry [RUST]: 0.24.6
    - tao [RUST]: 0.16.5
    - @tauri-apps/api [NPM]: 1.5.2
    - @tauri-apps/cli [NPM]: 1.5.7 (outdated, latest: 1.5.8)

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../dist
    - framework: Vue.js
    - bundler: Vite
```

### Stack trace

_No response_

### Additional context

_No response_

amrbashir:
closes #8435


.changes/tauri-event-after-multiwebview.md
```diff
@@ -0,0 +1,10 @@
+---
+'tauri': 'patch:breaking'
+---
+
+Refactored the event system to better accommodate the new window types:
+
+- Added `EventTarget` enum.
+- `App/AppHandle/Window/Webview/WebviewWindow::emit` will not emit to all event listeners.
```
lucasfernog-crabnebula:
```suggestion
- `App/AppHandle/Window/Webview/WebviewWindow::emit` will now emit to all event listeners.
```
I think that's what you meant?

lucasfernog-crabnebula:
I like how this simplifies the event system (most complex part of our system tbh.. even though it shouldn't be). But it changes a lot about the mental model when developing compared to v1. On v1 you treat the object that you can `emit()` as the source of the event, but now it always triggers to all listeners and you need to use `emit_to` instead (opposite of the current approach).

I'm fine with it (it's weird to me because I'm so used to the current approach but I can live with it and learn it) but this needs to be the last breaking change on this API. ever. And I'd like to get input from others @FabianLars @simonhyll

.changes/tauri-event-after-multiwebview.md
amrbashir: yeah maybe, just `will emit`

amrbashir: actualy let's go with your wording

amrbashir:
I have been also thinking about maybe adding a spec or a diagram of how the event system emits and listens to events but don't hold this PR on it, I can add it later when I get to it.

FabianLars:
I think it makes sense. It was quite messy before, and the event system (for us and the user) is indeed quite complicated so i like the explicitness in this approach.

In v1 many people tripped up over the emit vs trigger approach and the current v2 version is a bit ambiguous.

Looking at the diagram, one last thing that feels a bit meh is the listen_global api. I can't really put it into words (i'll try to later once my headache is gone lol), but maybe make the App/Handle an actual target too?
Then you'd also be able to emit events strictly to the apphandle without bothering the window/webview listeners.

I think we could then get rid of listen_global directly, or make it an "event sniffer" (that can listen to events no matter the target).

amrbashir:
> I think we could then get rid of listen_global directly, or make it an "event sniffer" (that can listen to events no matter the target).

That's what it currently does and should be kept that way IMO.

Edit: actually no it doesn't sniff, let's make it a sniffer then

>  I can't really put it into words (i'll try to later once my headache is gone lol), but maybe make the App/Handle an actual target too?

While we can definitely do this, and uses can use `emit` or `emit_filter` to emit to App only, it won't be possible using `emit_to` unless we change `emit_to` to take an `EventTarget` which will make the UX of the function a little worse IMO. But I definitely like this idea, I will think about it but if there is better suggestion, let me know.

amrbashir:
Alright, I have added `App` event target and thanks to Rust trait system we can still keep `emit_to` UX as is, I have also removed the diagram and made a simple spec markdown file.

FabianLars:
I really like what you came up with, nice work!

lucasfernog:
Looks amazing. I'm only wondering if we really need to keep the app/window/webview/webviewWindow emit and emit_to functions now that they all run the same logic. Having a single "entry point" for emitting events might be easier to understand and simplify migration too.

amrbashir:
They are all the same now, it is just a method on `Manager` trait

lucasfernog (approved):

canxin121:
#8916

PR #8623, opened 2024-01-17
[feat] add arch support  for loong64

Kiri2002:

### DIFF
--- a/core/tauri/Cargo.toml
+++ b/core/tauri/Cargo.toml
@@ -28,6 +28,7 @@ rustc-args = [ "--cfg", "docsrs" ]
 rustdoc-args = [ "--cfg", "docsrs" ]
 default-target = "x86_64-unknown-linux-gnu"
 targets = [
+  "loongarch64-unknown-linux-gnu",
   "x86_64-pc-windows-msvc",
   "x86_64-unknown-linux-gnu",
   "x86_64-apple-darwin",


Kiri2002:
since the loong64 already support the basic tool-chains and rust etc.
I try to simply add the ''loongarch64-unknown-linux-gnu" to the target.
Magically, works well.
From archlinux(for loong64)

lucasfernog:
This only adds documentation for the `loongarch64-unknown-linux-gnu` target. It doesn't change anything about how the crate is compiled. Actually compiling it is a little more complicated since you need to cross compile.

Since we don't have any particular documentation item specific to `loongarch64-unknown-linux-gnu`, I'll close this one. Thank you!

Kiri2002:
> This only adds documentation for the `loongarch64-unknown-linux-gnu` target. It doesn't change anything about how the crate is compiled. Actually compiling it is a little more complicated since you need to cross compile.
> 
> Since we don't have any particular documentation item specific to `loongarch64-unknown-linux-gnu`, I'll close this one. Thank you!

It works because rust is already support loong64. Once added loongarch64-unknown-linux-gnu target, it can be complied in loong64 machines. It doesn't need to have loong64 release in github, just for source code support, so downstream(Archlinux etc.)  can compile without add patch.

Tauri is already compiled successfully in loong64.

We don't need "need to cross compile", we just compile in loong64 machine.

If it can be merged, just for source code, and don't need release compile, it will help a lot!

Kiri2002:
@lucasfernog Thank your review. I test again just now to build ```cargo-tauri``` in ```loong64```(Archlinux), built and tested successfullt. The only thing I change is to add ```loongarch64-unknown-linux-gnu``` target.
[cargo-tauri-1.5.9-1-loong64-build.log](https://github.com/tauri-apps/tauri/files/15151688/cargo-tauri-1.5.9-1-loong64-build.log)
[cargo-tauri-1.5.9-1-loong64-check.log](https://github.com/tauri-apps/tauri/files/15151689/cargo-tauri-1.5.9-1-loong64-check.log)

Still hope it can add to source(convenient for loong64 user to build in their local)

FabianLars:
I am still not sure what you want us to do here to be honest. This PR only added this target to the docs.rs instructions, so that it shows up in this dropdown:
grafik
This does not change anything about the actual platform support and the docs will be exactly the same as in the already existing x86_64-unknown-linux-gnu target docs.

To give you another example, even though the docs.rs targets only include x86_64, you can compile tauri to 32bit targets and even arm (aarch64).

Kiri2002:
> I am still not sure what you want us to do here to be honest. This PR only added this target to the docs.rs instructions, so that it shows up in this dropdown: grafik This does not change anything about the actual platform support and the docs will be exactly the same as in the already existing x86_64-unknown-linux-gnu target docs.
> 
> To give you another example, even though the docs.rs targets only include x86_64, you can compile tauri to 32bit targets and even arm (aarch64).
 
Yes, that's right, without this change, it can be built.
Now I totally understand this!
Sorry for my misunderstand before, and Thank you for your perfect explain!!

PR #8666, opened 2024-01-24
feat(cli): update template with permissions and capabilities

lucasfernog:

### DIFF
--- a/tooling/cli/templates/app/src-tauri/capabilities/default.json
+++ b/tooling/cli/templates/app/src-tauri/capabilities/default.json
@@ -1,0 +1,16 @@
+{
+  "$schema": "./schemas/desktop-schema.json",
+  "identifier": "default-plugins",
+  "description": "enables the default permissions",
+  "windows": ["main"],
+  "permissions": [
+    "path:default",
+    "event:default",
+    "window:default",
+    "webview:default",
+    "app:default",
+    "resources:default",
+    "menu:default",
+    "tray:default",
+  ]
+}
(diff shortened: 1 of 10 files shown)


tooling/cli/templates/app/src-tauri/capabilities/default.json
```diff
@@ -0,0 +1,16 @@
...
+  "identifier": "default-plugins",
+  "description": "enables the default permissions",
+  "windows": ["main"],
+  "permissions": [
+    "path:default",
+    "event:default",
+    "window:default",
+    "webview:default",
+    "app:default",
+    "resources:default",
```
amrbashir, line 12: do we want to enable this by default? or better can `menu:default` and `tray:default` include this permission? because resources is kinda internal and users should not need to deal with it.

lucasfernog: right now there's no way for a plugin to enable commands on other plugins.. but we can adapt core to ignore permission checks for the resources plugin worst case

amrbashir: is that for a security reason or just not implemented?

lucasfernog: doesn't make much sense really.. but i can include a set of permissions that are enabled by default

lucasfernog: we can include all core plugin default permissions automatically.. then the user can deny the functions he doesn't want.. what do you think @chippers

amrbashir (approved):

PR #8670, opened 2024-01-24
fix: improvements and ipc fixes for loading window content using custom protocol

### LINKED ISSUE #5478: [bug] can not load custom protocol content when creating window
### Describe the bug

```rust
let app_store = WindowBuilder::new(
    &app,
    "appstore",
    WindowUrl::External("package://localhost/".parse().unwrap()),
).build().unwrap();
```

after looking inside the tauri source code, found the problem.

https://github.com/tauri-apps/tauri/blob/e4292ce7be198f6265dde9b6ae54e73b7c41e7da/core/tauri/src/manager.rs#L1128-L1138

is that ok to support something likes `WindowUrl::CustomProtocol("package://localhost/".parse().unwrap())` and treats as local content or adding a config to remove the limitation by the user's will?

### Reproduction

_No response_

### Expected behavior

_No response_

### Platform and versions

Environment
  › OS: Windows 10.0.19044 X64
  › Webview2: 106.0.1370.52
  › MSVC:
      - Visual Studio 15 ���ɹ��� 2017
  › Node.js: 16.17.0
  › npm: 8.19.2
  › pnpm: 7.13.5
  › yarn: 1.22.19
  › rustup: 1.25.1
  › rustc: 1.63.0
  › cargo: 1.63.0

Packages
  › @tauri-apps/cli [NPM]: 1.0.5
  › @tauri-apps/api [NPM]: 1.0.2
  › tauri [RUST]: 1.1.1,
  › tauri-build [RUST]: 1.1.1,
  › tao [RUST]: 0.14.0,
  › wry [RUST]: 0.21.1,

App
  › framework: Svelte

App directory structure
  ├─ .git
  ├─ .vscode
  ├─ embedded
  ├─ local-cargo
  ├─ node_modules
  ├─ public
  ├─ src
  └─ src-tauri

### Stack trace

_No response_

### Additional context

_No response_

amrbashir:
closes #5478

- made `distDir` and `devPath` optional so if someone wants to load all their windows using custom protocols they don't need to specify and create an empty directory for `distDir` when building
- added `WebviewUrl::CustomProtocol` which is the same as `WebviewUrl::External` but more explicit.
- fixed ipc calls from a window that loaded over custom protocol


core/tauri/src/webview/mod.rs
```diff
@@ -1044,17 +1044,47 @@ fn main() {
...
+
+    // or if relative to `distDir` or `devPath`
+      self
+          .manager()
+          .get_url()
+          .make_relative(current_url)
+          .is_some()
+
+      // TODO: what use-case is covered by this?
       || (cfg!(dev) && current_url.domain() == Some("tauri.localhost"))
```
amrbashir: @lucasfernog what use-case is being covered by this check?

amrbashir:
There is one more improvement we can make but don't know if it makes sense? currently users will need to always explicitly specify the url for each window either in code or in tauri.conf.json, would it make sense if we allowed them to set a base url for all windows instead of `tauri://` which is the default for all windows at the moement?

core/tauri/src/webview/mod.rs
lucasfernog: IIRC it's for mobile.. should've documented it my bad

lucasfernog:
> There is one more improvement we can make but don't know if it makes sense? currently users will need to always explicitly specify the url for each window either in code or in tauri.conf.json, would it make sense if we allowed them to set a base url for all windows instead of `tauri://` which is the default for all windows at the moement?

I think this would complicate the codebase and introduce some bugs likely.. let's keep that until someone actually requests it

amrbashir:
I agree

core/tauri/src/webview/mod.rs
amrbashir: won't this be covered by the `tauri:://` case as well?

lucasfernog (approved):

PR #8672, opened 2024-01-25
fix(cli): add MaybeWorkspace to license field

Andreybest:
Got the bug in a project after updating to latest v2-alpha.
And got this error:
```
> pnpm tauri dev

       Error failed to load cargo settings: failed to parse Cargo.toml: TOML parse error at line 6, column 11
  |
6 | license = { workspace = true }
  |           ^^^^^^^^^^^^^^^^^^^^
invalid type: map, expected a string

 ELIFECYCLE  Command failed with exit code 1.
```

From what I saw in a code, license field was expected to be either String or None, but not workspace. Made it same as other fields (_by copy-pasting code from them_).

I think that it should work with any other case, but at least it fixes my issue. Let me know if I made something wrong. Will try to be fast with responses.

### Checks
- `cargo test` passes
- `cargo clippy` passes
- `cargo fmt` _locally_ passes
- if cli installed with `cargo install` and used via `cargo tauri dev` with a license field like `license = { workspace = true }`, exception is not being thrown

### DIFF
--- a/.changes/cli-license-field-workspace.md
+++ b/.changes/cli-license-field-workspace.md
@@ -1,0 +1,6 @@
+---
+"tauri-cli": patch:bug
+"@tauri-apps/cli": patch:bug
+---
+
+Allow license field in Cargo.toml to be `{ workspace = true }`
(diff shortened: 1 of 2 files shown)


FabianLars:
Thanks for contributing! Could you also add a changefile similar to https://github.com/tauri-apps/tauri/blob/dev/.changes/cli-config-target-mobile.md please?

Andreybest:
@FabianLars Added, hope my phrasing is correct here.

FabianLars (approved):

Issue #862, opened 2020-07-19
[docs] differences in dev and bundle, running an exec process


gregbenner:
Very cool library I just wrote a [post](https://gregbenner.life/web-view-apps-on-your-os/) looking at similar

Looking at the docs I can't find a difference with dev and bundle for running a bundled binary like:
```
"build": {
    "distDir": ".",
    "devPath": "http://localhost:3666",
    "beforeDevCommand": "vue ui --headless --port 3666",
    "beforeBuildCommand": "vue ui --headless --port 3666"
  },
```

The dev works fine but the release hangs, I'm assuming it''s waiting for the server to exit. Do I need to include it as an "externalBinary" instead or run it as a spawned exec process? I'm not Rust savy, more node.js hopefully this isnt a stupid question.

Thanks!

PR #8675, opened 2024-01-25
fix(acl): command scope should not error out if missing

lucasfernog:
The command-specific scope is not required and it should be easier for plugin devs to reference them without having to deal with that:
```rust
#[tauri::command]
async fn execute(scope: tauri::command::CommandScope<'_, ScopeEntry>)...
```
so it should not fail to retrieve the value if the scope was not set by the app user


lucasfernog:
Usage example: [tauri-apps/plugins-workspace@`1f307c7` (#862)](https://github.com/tauri-apps/plugins-workspace/pull/862/commits/1f307c788b6c20824064494ccb75ca177122ba12)

core/tauri/src/command/authority.rs
```diff
@@ -112,9 +112,27 @@ impl<T: Debug + DeserializeOwned + Send + Sync + 'static> ScopeValue<T> {
   }
 }
 
+#[derive(Debug)]
+enum OwnedOrRef<'a, T: Debug> {
```
amrbashir, line 116: Can't we use `Cow` ?

amrbashir (dismissed):

core/tauri/src/command/authority.rs
lucasfernog: Cow requires `T: Clone` and we don't really need that :(

amrbashir: I see, it is fine then

amrbashir (approved):

PR #8677, opened 2024-01-25
fix(core/window): set window_effects from config

AuTa:
The #8280 pull request cannot set window effects when use the config `windowEffects.effects`.

`window_effects` is always `None`, and the if judgment on line 410 will only be false.
```rust
    if let Some(effects) = self.window_effects {
      crate::vibrancy::set_window_effects(&window, Some(effects))?;
    }
```
Advance the `window_effects` because the `config` was moved at `window_builder`.


FabianLars (approved):
thank you :)

PR #8680, opened 2024-01-25
feat: Add support for deep links

### LINKED ISSUE #323: Feature: Custom URI scheme support (deep linking)
Custom URL schemes provide a way to reference resources inside an app. Users tapping a custom URL e.g. `yourapp://foo?bar` in an email, for example, launch the app in a specified context. Other apps can also trigger the app to launch with specific context data; for example, a photo library app might display a specified image.

Tauri could have a simplified API to register/unregister the custom URI scheme protocol a listener.

A Rust package https://github.com/maidsafe/system_uri exists that does System App URI registration for macOS, Linux and Windows. Unfortunately, this package does not support passing also the url params to the application - it only triggers the application launch. An example implementation together with Tauri  -https://github.com/iotaledger/spark-wallet/blob/master/src-tauri/src/main.rs

FabianLars:
Together with https://github.com/tauri-apps/plugins-workspace/issues/916 this will resolve #323.

One semi major question we never **clearly** answered is how we want to align it with the file association feature. If we plan to add helper apis to the file assocs to make handling os differences easier i think it makes sense to pull that out into it's own plugin. I _thought_ we have code apis for it, which is why i always talked about the app module, but seems like i was wrong. In this case moving it into a plugin could be done later in a non breaking way too, we'd just have 2 duplicate config entries 🤷

I just feel like they are similar enough that it may be a bit weird having one in core and one as a plugin but maybe i'm overthinking it idk.

Also, after we got v2 to stable i'd like to work on the plugin/cli instructions thingy to get rid of the plugin specific stuff again.

Speaking of that, this PR also adds xdg-open if the shell plugin is used.


aleksey-hoffman:
@FabianLars thanks mate, a lot of Tauri apps would benefit from this! 
Can't wait for it to be merged so I can add it inito Sigma File Manager

core/tauri-utils/src/config.rs
```diff
@@ -833,14 +833,28 @@ pub struct FileAssociation {
...
 }
 
+/// File association
+#[derive(Debug, PartialEq, Eq, Clone, Deserialize, Serialize)]
+#[cfg_attr(feature = "schema", derive(JsonSchema))]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct DeepLinkProtocol {
+  /// URL schemes to associate with this app without `://`. For example `my-app`
+  pub schemes: Vec<String>,
+  /// The protocol name. macOS-only and maps to `CFBundleTypeName`. Defaults to `<bundle-id>.<schemes[0]>`
```
```suggestion
  /// The protocol name. **macOS-only** and maps to `CFBundleTypeName`. Defaults to `<bundle-id>.<schemes[0]>`
```

core/tauri-utils/src/config.rs
```diff
@@ -833,14 +833,28 @@ pub struct FileAssociation {
...
+  pub name: Option<String>,
+  /// The app's role for these schemes. macOS-only and maps to `CFBundleTypeRole`.
```
```suggestion
  /// The app's role for these schemes. **macOS-only** and maps to `CFBundleTypeRole`.
```

tooling/bundler/src/bundle/settings.rs
```diff
@@ -904,6 +906,12 @@ impl Settings {
     &self.bundle_settings.file_associations
   }
 
+  /// Return the list of deep link protocols to be registered for
+  /// this bundle.
+  pub fn deep_link_protocols(&self) -> &Option<Vec<DeepLinkProtocol>> {
+    &self.bundle_settings.deep_link_protocols
```
amrbashir: Optional reference is better semantically than a reference to an option

```suggestion
  pub fn deep_link_protocols(&self) -> Option<&Vec<DeepLinkProtocol>> {
    self.bundle_settings.deep_link_protocols.as_ref()
```

tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -652,6 +660,15 @@ Section Uninstall
     {{/each}}
   {{/each}}
 
+  ; Delete deep links
+  {{#each deep_link_protocol as |protocol| ~}}
+    ReadRegStr $DL0 SHCTX "Software\Classes\{{protocol}}\shell\open\command" ""
```
amrbashir: What is `$DL0`? can't find any mention of it in NSIS docs

tooling/bundler/src/bundle/windows/templates/main.wxs
```diff
@@ -53,7 +41,8 @@
 
         <Icon Id="ProductIcon" SourceFile="{{icon_path}}"/>
         <Property Id="ARPPRODUCTICON" Value="ProductIcon" />
-        <Property Id="ARPNOREPAIR" Value="yes" Secure="yes" />      <!-- Remove repair -->
+        <Property Id="ARPNOREPAIR" Value="yes" Secure="yes" />
+        <!-- Remove repair -->
```
amrbashir: Formatting messed up where the comment points at, move it up one line

tooling/bundler/src/bundle/windows/templates/main.wxs
```diff
@@ -74,17 +63,9 @@
...
-                     Control="Next"
-                     Event="NewDialog"
-                     Value="InstallDirDlg"
-                     Order="2">1</Publish>
-            <Publish Dialog="InstallDirDlg"
-                     Control="Back"
-                     Event="NewDialog"
-                     Value="WelcomeDlg"
-                     Order="2">1</Publish>
+                                                                        <!-- Skip license dialog -->
```
amrbashir: fomratting issue

tooling/bundler/src/bundle/windows/templates/main.wxs
```diff
@@ -111,6 +92,19 @@
                 <RegistryKey Root="HKCU" Key="Software\\{{manufacturer}}\\{{product_name}}">
                     <RegistryValue Name="InstallDir" Type="string" Value="[INSTALLDIR]" KeyPath="yes" />
                 </RegistryKey>
+                {{#each deep_link_protocols as |protocol| ~}}
+                                                                                                    <!-- Change the Root to HKCU for perUser installations -->
```
amrbashir: formatting and just to not sound as a broken record, can we revert most of these formatting changes?

tooling/cli/src/build.rs
```diff
@@ -156,6 +156,17 @@ pub fn command(mut options: Options, verbosity: u8) -> Result<()> {
       if config_.tauri.bundle.appimage.bundle_media_framework {
         std::env::set_var("APPIMAGE_BUNDLE_GSTREAMER", "1");
       }
+
+      // TODO: Consider using the plugin crate directly
```
amrbashir: not sure what this means but the way it is implemented right now is okay (until a better approach to extend the bunder)

amrbashir (changes requested):

FabianLars:
i guess i'm the only one with vscode's auto formatter enabled lol. i'll try to revert that shit

tooling/cli/src/build.rs
FabianLars: i meant importing the crate to be able to use the actual config struct. but yeah, not worth the headache here.

tooling/bundler/src/bundle/settings.rs
FabianLars: i copied this from the file associations. should i change that too?

amrbashir: yeah

tooling/bundler/src/bundle/windows/templates/installer.nsi
FabianLars: afaik nsis doesn't care what it's called and i think it meant DeepLink0 for me (don't ask what the 0 is for). I can change it to $R7 if you prefer that. We use $R_ for other ReadRegStr calls and 0-6 is used already.

amrbashir: > afaik nsis doesn't care what it's called

wild, didn't know that



> i think it meant DeepLink0 for me (don't ask what the 0 is for). I can change it to $R7 if you prefer that. We use $R_ for other ReadRegStr calls and 0-6 is used already.

$R7 is fine, also we can use any other variable, they are not used anymore after this call iirc.

amrbashir (approved):

PR #8681, opened 2024-01-25
feat(ipc): improve error message when plugin command is not allowed

lucasfernog:
On debug, we'll give more information to the app developer on why the IPC call was rejected:
- what permissions he can use to enable the command
- if the command is enabled but for a different window, we present that info
- if the command is enabled but for a different remote, we also display that info
- if the command was denied by one of the capabilities, we tell that to the user too

### DIFF
--- a/core/tauri-utils/src/acl/plugin.rs
+++ b/core/tauri-utils/src/acl/plugin.rs
@@ -109,20 +109,7 @@ mod build {
   use std::convert::identity;
 
   use super::*;
-  use crate::tokens::*;
-
-  /// Write a `TokenStream` of the `$struct`'s fields to the `$tokens`.
-  ///
-  /// All fields must represent a binding of the same name that implements `ToTokens`.
-  macro_rules! literal_struct {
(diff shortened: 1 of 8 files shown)


core/tauri-utils/src/acl/plugin.rs
```diff
@@ -104,3 +101,66 @@ impl Manifest {
...
+  use quote::{quote, ToTokens, TokenStreamExt};
+  use std::convert::identity;
+
+  use super::*;
+  use crate::tokens::*;
+
+  /// Write a `TokenStream` of the `$struct`'s fields to the `$tokens`.
+  ///
+  /// All fields must represent a binding of the same name that implements `ToTokens`.
+  macro_rules! literal_struct {
```
amrbashir: Can't we share this macro, seems replicated a couple of times

amrbashir (dismissed):

amr-crabnebula (approved):

amrbashir (approved):

PR #8688, opened 2024-01-27
fix(cli): disable global gitignore during tauri.conf.* lookup

wvffle:
There are cases when people use git to manage their dotfiles in the home directory. When a tauri projects uses other name than `src-tauri` for the rust source code, the lookup may fail if there's a global gitignore.

This PR aims to fix that.

### DIFF
--- a/.changes/cli-global-gitignore.md
+++ b/.changes/cli-global-gitignore.md
@@ -1,0 +1,6 @@
+---
+"tauri-cli": patch:enhance
+"@tauri-apps/cli": patch:enhance
+---
+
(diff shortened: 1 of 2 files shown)


amrbashir (commented):
Could you also add a change file in `.changes` directory?

amrbashir (approved):

Qix-:
I'm not sure what your current backport strategy is but could this be back ported to v1 perchance? Or is v1 completely frozen?

FabianLars:
v1 is still open for backports, we just don't work on backports that much ourselves and rely more on PRs.

PR #8697, opened 2024-01-29
fix(cli): Ignore query parameter in dev server

### LINKED ISSUE #8148: [bug] Cannot use URL parameters with window.location.href
### Describe the bug

I am unable to use URL parameters with window.location.href

This works:
window.location.href = 'test.html'

But this doesn't work:
window.location.href = 'test.html?test=test' 

The console outputs "Navigated to http://127.0.0.1:1430/test.html?test=test" but the page does not change.

I am using plain HTML with no build tools or frameworks. I have also tried using the following but it returns the same result:

window.location.href = encodeURIComponent(`test.html?test=test, 'UTF-8')}`)

### Reproduction

I was able to reproduce the problem in a freshly created Tauri HTML app.

### Expected behavior

It should route to test.html and the URL params should be available in the test.html page.

### Platform and versions

```text
[✔] Environment
    - OS: Windows 10.0.19045 X64
    ✔ WebView2: 118.0.2088.76
    - node: 18.18.0
    - yarn: 1.22.19
    - npm: 10.1.0

[-] Packages
    - tauri [RUST]: 1.5.2
    - tauri-build [RUST]: 1.5.0
    - wry [RUST]: 0.24.4
    - tao [RUST]: 0.16.5

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../src
    - devPath: ../src
```

### Stack trace

_No response_

### Additional context

_No response_

FabianLars:
fixes #8148
additional ref: https://discord.com/channels/616186924390023171/1201199918379974766

imho query parameters have no business being in frontend file names (same for `#` but the browser doesn't send that part to the server so doesn't matter here).

This should now also match the behavior after `tauri build`.

The asset protocol should be able to handle it fine due to the url encoding, so devs can still load files with weird filenames that aren't part of the frontend.

### DIFF
--- a/.changes/cli-devserver-queryparam.md
+++ b/.changes/cli-devserver-queryparam.md
@@ -3,4 +3,4 @@
 "@tauri-apps/cli": patch:bug
 ---
 
-The built-in dev watcher now ignores the `?` query part of URLs.
+Fix the built-in dev server failing to serve files when URL had queries `?` and other url components.
--- a/.changes/cli-devserver-queryparam.md
+++ b/.changes/cli-devserver-queryparam.md
@@ -1,0 +1,6 @@
+---
+"tauri-cli": patch:bug
+"@tauri-apps/cli": patch:bug
+---
+
+The built-in dev watcher now ignores the `?` query part of URLs.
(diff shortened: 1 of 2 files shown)


amrbashir (dismissed):

.changes/cli-devserver-queryparam.md
```diff
@@ -0,0 +1,6 @@
+---
+"tauri-cli": patch:bug
+"@tauri-apps/cli": patch:bug
+---
+
+The built-in dev watcher now ignores the `?` query part of URLs.
```
amrbashir:
```suggestion
Fix the built-in dev server failing to serve files when URL had queries `?` and other url components.
```

amrbashir (approved):

PR #8699, opened 2024-01-29
refactor(core): allow scope deserialization to run setup code

lucasfernog:
Some plugins runs some checks and processing on the scope type (like the fs plugin replacing variables on the paths in a scope). This changes the scope value to a simple `T: DeserializeOwned` to a custom struct the plugin writer can implement.

### DIFF
--- a/core/tauri/src/command/authority.rs
+++ b/core/tauri/src/command/authority.rs
@@ -216,7 +216,7 @@ pub struct ScopeManager {
   global_scope_cache: TypeMap![Send + Sync],
 }
 
-/// Marks a type as an scope object.
+/// Marks a type as a scope object.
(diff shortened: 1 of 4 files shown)


core/tauri/src/command/authority.rs
```diff
@@ -220,9 +216,28 @@ pub struct ScopeManager {
   global_scope_cache: TypeMap![Send + Sync],
 }
 
+/// Marks a type as an scope object.
```
amrbashir:
```suggestion
/// Marks a type as a scope object.
```

amrbashir (dismissed):

amrbashir (approved):

PR #8708, opened 2024-01-30
refactor(core): exit() and restart() triggers ExitRequested and Exit

lucasfernog:
This changes the AppHandle::exit and AppHandle::restart implementations to actually trigger RunEvent::ExitRequested and RunEvent::Exit instead of forcing exit with std::process::exit

### DIFF
--- a/.changes/refactor-exit.md
+++ b/.changes/refactor-exit.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch:breaking
+---
+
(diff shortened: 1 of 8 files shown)


FabianLars (approved):

amrbashir (approved):
Can we also do the same for `Window::close`? so custom titlebars can trigger `CloseRequested` ?

lucasfernog:
Gonna try that

PR #8718, opened 2024-01-31
feat(bundler): add support for EV codesigning certs via azuresigntool

tr3ysmith:
This feature adds support for EV code signing via azuresigntool.exe. The original method using using signtool.exe still works. But if the environment variables are set properly before running `tauri build`, the Tauri bundler will attempt to use AzureSignTool.

To use AzureSignTool, the user must have an EV certificate hosted in Azure Key Vault.


tr3ysmith:
due to the complexity of the azuresigntool package, I removed it from the binary and it'll need to currently be preinstalled on the host machine. With GitHub runners, this should be possible by running  `dotnet tool install --global AzureSignTool` before executing the Tauri build command

tr3ysmith:
@lucasfernog should we bump the rust min version on the main crate too? I had to bump it to 1.72 on the CLI crate because of the dependency for jsonschema -> aHash (aHash requires 1.72)

https://github.com/tkaitchuck/aHash/pull/196

tooling/cli/Cargo.toml
```diff
@@ -6,7 +6,7 @@ name = "tauri-cli"
 version = "1.5.9"
 authors = [ "Tauri Programme within The Commons Conservancy" ]
 edition = "2021"
-rust-version = "1.60"
```
amrbashir, line 9: Even though this is a CLI and a MSRV shouldn't matter that much but I think we plan to keep 1.60 for 1.x. Anyways, this PR probably needs to target `dev` not `1.x` so this shouldn't be a problem.

tooling/bundler/src/bundle/windows/sign.rs
```diff
@@ -17,14 +19,164 @@ use winreg::{
...
-pub struct SignParams {
+/// Enum to hold the different signing params depending on the current setup. (Signtool or AzureSignTool)
+pub enum SignParams {
+  SignTool(SignToolParams),
+  Azure(AzureSignToolParams),
+}
+
+impl SignParams {
+  /// Locates the signing tool executable
+  fn locate_tool(&self) -> crate::Result<PathBuf> {
```
amrbashir: Seems like is never used, let's remove for now

tooling/bundler/src/bundle/windows/msi/wix.rs
```diff
@@ -788,8 +785,9 @@ pub fn build_wix_app_installer(
       &msi_output_path,
     )?;
     rename(&msi_output_path, &msi_path)?;
-    try_sign(&msi_path, settings)?;
-    output_paths.push(msi_path);
```
amrbashir, line 792: this line shouldn't be removed

tooling/bundler/src/bundle/windows/sign.rs
```diff
@@ -88,93 +240,84 @@ fn locate_signtool() -> crate::Result<PathBuf> {
...
+    // We'll attempt to create a SignParams struct from the environment variables and the settings
+    // First we'll start with AzureSignTool, if we have all the required environment variables set, then we'll use that.
+    // If not, we'll fallback to Signtool.
+    // But if we don't have a certificate thumbprint set, we can't sign at all and we'll return None.
+    match (
+      var_os("AZURE_KEYVAULT_URL"),
+      var_os("AZURE_CLIENT_ID"),
+      var_os("AZURE_TENANT_ID"),
+      var_os("AZURE_CLIENT_SECRET"),
+      var_os("AZURE_CERTIFICATE_NAME"),
```
amrbashir, line 258: we need to list all of these new env variables in `tooling/cli/ENVIRONMENT_VARIABLES.md`

amrbashir (changes requested):
I am not sure if we should support `azuresigntool` tbh since it is not an official tool (correct me if I am wrong), should we instead provide a generic option to specify a signing command, similar to this https://github.com/crabnebula-dev/cargo-packager/blob/ec1a3a12296275341fd59c083d98202de316805f/crates/packager/src/config/mod.rs#L1206-L1214?

tooling/bundler/src/bundle/windows/sign.rs
tr3ysmith: Added!

tooling/bundler/src/bundle/windows/msi/wix.rs
tr3ysmith: @amrbashir its still there, it just comes from sign_params now right below it. I added back `output_paths.push(msi_path);`

tooling/bundler/src/bundle/windows/sign.rs
tr3ysmith: removed!

tooling/cli/Cargo.toml
tr3ysmith: removed!

tr3ysmith:
@amrbashir do I need to do anything to fix the clippy workflow failure?

amrbashir:
if you want, that would be appreciated, otherwise we can fix it in another PR

amrbashir:
cc @lucasfernog about https://github.com/tauri-apps/tauri/pull/8718#pullrequestreview-1870698873

lucasfernog:
> I am not sure if we should support `azuresigntool` tbh since it is not an official tool (correct me if I am wrong), should we instead provide a generic option to specify a command to signing command, similar to this [crabnebula-dev/cargo-packager@`ec1a3a1`/crates/packager/src/config/mod.rs#L1206-L1214](https://github.com/crabnebula-dev/cargo-packager/blob/ec1a3a12296275341fd59c083d98202de316805f/crates/packager/src/config/mod.rs#L1206-L1214)?

It gets complicated when you think about how complicated it is to sign a Windows app.. we need a better solution, and this might be one (I still didn't run through it yet, hopefully when I get back from vacation).

lucasfernog:
> @lucasfernog should we bump the rust min version on the main crate too? I had to bump it to 1.72 on the CLI crate because of the dependency for jsonschema -> aHash (aHash requires 1.72)
> 
> [tkaitchuck/aHash#196](https://github.com/tkaitchuck/aHash/pull/196)

Yeah we should.. we need to do a better job on the msrv for the CLI

tr3ysmith:
> > @lucasfernog should we bump the rust min version on the main crate too? I had to bump it to 1.72 on the CLI crate because of the dependency for jsonschema -> aHash (aHash requires 1.72)
> > [tkaitchuck/aHash#196](https://github.com/tkaitchuck/aHash/pull/196)
> 
> Yeah we should.. we need to do a better job on the msrv for the CLI

@lucasfernog @amrbashir should we then bump both the CLI and Core, or just CLI?

tr3ysmith:
> if you want, that would be appreciated, otherwise we can fix it in another PR

I think we should do it in another PR, since it looks to be completely unrelated to this change

lucasfernog:
that's ok we can do that separately

tr3ysmith:
@lucasfernog MSRV for cli is updated

WalrusSoup:
> I am not sure if we should support `azuresigntool` tbh since it is not an official tool (correct me if I am wrong), should we instead provide a generic option to specify a signing command, similar to this https://github.com/crabnebula-dev/cargo-packager/blob/ec1a3a12296275341fd59c083d98202de316805f/crates/packager/src/config/mod.rs#L1206-L1214?

+1. We just got our new digicert token and it would be nice to be able to just call our own signing script. Otherwise, I could see this just growing and growing to accommodate the dozens of signing approaches.

amrbashir:
I agree, @tr3ysmith would a generic command work for you as well?

tr3ysmith:
@amrbashir sorry ya'll. I've had some crazy life stuff pop up. Yes a generic command would work for me as long as I can pass all the required arguments through via the tauri-action. I can work on that here in the next few days.

amrbashir:
thanks @tr3ysmith hope you're doing great, let me know if I can help with anything

amrbashir:
Hi @tr3ysmith I have opened https://github.com/tauri-apps/tauri/pull/9865 to implement the custom signing command, gonna close this PR now in favor of that one. Thanks for your contributions

tr3ysmith:
@amrbashir no problem, is this something that'll get merge into 1.x at all?

amrbashir:
there is a PR to backport it in #9902

PR #8723, opened 2024-02-01
feat(core): implement new config structure

### LINKED ISSUE #8055: [feat] v2 - Move bundle configs into respective platform objects.
### Describe the problem

Currently the config is a bit of a mess, we have deb and appimage configs at the root but also a windows config that contains the nsis and wix configs.

Move appimage and deb into a linux config. Something similar will probably need to happen on macos with the open dmg PR and potentially upcoming formats (pkg?).

I can work on this myself (unless a community member _wants_ to work on this :) ) but wanted to discuss this here first, so don't take this issue as a "Please fix this for me" thing 😅

P.S. i am aware that cargo-package doesn't have OS configs but tauri's config covers way more stuff (other than bundler related things) so imo a bit more structure would be valuable, especially since not every user may know each bundle name so maybeee this clears it up a bit for a few people 🤷 

### Alternatives considered

pulling out wix, nsis, etc into the root or just leaving it as is. The first one is too messy and the second one looks too much like an "accident" (for the lack of a better word)

### Additional context

imo this should happen after the rpm and dmg prs so that the community members don't have to deal with it, _unless_ we think we can't get those PR in tauri in time for the audit, then we should change it asap and let the PRs deal with the change.

PRs that would be affected:
rpm https://github.com/tauri-apps/tauri/pull/5202 (adds new configs but at the root like deb)
dmg https://github.com/tauri-apps/tauri/pull/7964 (adds new configs but at the root like deb)
snap https://github.com/tauri-apps/tauri/pull/6532 (this one doesn't have configs yet)
pacman https://github.com/tauri-apps/tauri/pull/4301 (this one doesn't have configs yet)

### LINKED ISSUE #8727: [bug] Multiple License options with misleading option name
### Describe the bug

Currently, there are multiple separate licence options in Tauri Configuration. They should be clubbed into 1 or 2 option, with consistent naming.
Also, the name is also misleading. Current Situation:

1. `PackageSettings` : It has a `license` option and expects a string.
2. `RpmSettings`: It has license option as an alternate to `PackageSettings`
3. **`MacOsSettings`**: It has a license option, that is the path to the LICENSE file. However, it is taken as `Option<String>` instead of `Option<PathBuf>`
4. `WixSettings`: It also has a license option, which takes path to the LICENSE file, but uses ` Option<PathBuf>`
5. `NsisSettings`: Similar to WixSettings.

So, it forms a bit chaos for bundling application.

### Reproduction

_No response_

### Expected behavior

1. The License should be clubbed into 1 universal option under package config.
2. `license_file` option should be used instead of `license` appropiately.

### Full `tauri info` output

```text
Latest Tauri dev branch
```

### Stack trace

_No response_

### Additional context

_No response_

amrbashir:
RFC#5 https://github.com/tauri-apps/rfcs/blob/f3e82a6b0c5390401e855850d47dc7b7d9afd684/texts/0005-tauri-config-restructure.md

Closes #8055
Closes #8727


core/tauri-utils/src/config.rs
```diff
@@ -1607,6 +1539,26 @@ pub struct SecurityConfig {
...
   pub asset_protocol: AssetProtocolConfig,
+  /// The pattern to use.
+  #[serde(default)]
+  pub pattern: PatternKind,
+}
+
+impl SecurityConfig {
+  /// Returns all Cargo features.
+  pub fn all_features() -> Vec<&'static str> {
+    vec!["isolation"]
```
lucasfernog: we should either keep this inside AppConfig or change all references to AppConfig::all_features, app_config.features() to also check these

FabianLars:
One thing we talked about a few times which i completely forgot to mention in the rfc is to merge appimage deb rpm etc into a linux parent and dmg and app into a macos one to match the windows config. I honestly think this is really important.
https://github.com/tauri-apps/tauri/issues/8055

FabianLars:
another one maybe? https://github.com/tauri-apps/tauri/issues/8727

amrbashir:
yeah seems like a good idea

core/tauri-utils/src/config.rs
```diff
@@ -1920,28 +1773,30 @@ pub enum HookCommand {
...
   /// You can also provide a list of paths to be embedded, which allows granular control over what files are added to the binary.
   /// In this case, all files are added to the root and you must reference it that way in your HTML files.
   ///
-  /// When an URL is provided, the application won't have bundled assets
+  /// When a URL is provided, the application won't have bundled assets
   /// and the application will load that URL by default.
-  #[serde(alias = "dist-dir")]
-  pub dist_dir: Option<AppUrl>,
+  #[serde(alias = "frontend-dist")]
+  pub frontend_dist: Option<AppUrl>,
```
amrbashir: this option seems overloaded IMO and `frontend_dist` doesn't really convey all possible options, do we want to split this? maybe into:
- `frontend_dist`
- `frontend_files`
- `frontend_url`

lucasfernog: the problem is they are mutually exclusive...

lucasfernog: what if we turn it into an enum?
`production: { url: string } | { dist: pathbuf } | { files: vec<pathbuf> }`

lucasfernog: another problem with frontend_url is that it would be weird next to dev_url

.changes/tauri-bundle-settings-rfc-5.md
```diff
@@ -0,0 +1,5 @@
+---
+'tauri-bundler': 'major:breaking'
+---
+
+- Removed all license fields from `WixSettings`, `NsisSettings` and `MacOsSettings` and replaced with `license` and `license_file` fileds in `BundlerSettings`.
```
lucasfernog:
```suggestion
- Removed all license fields from `WixSettings`, `NsisSettings` and `MacOsSettings` and replaced with `license` and `license_file` fields in `BundlerSettings`.
```

.changes/config-restructure-rfc-5.md
```diff
@@ -0,0 +1,20 @@
+---
+'tauri': 'major:breaking'
+'tauri-utils': 'major:breaking'
```
lucasfernog, line 3:
```suggestion
'tauri': 'major:breaking'
'tauri-utils': 'major:breaking'
'tauri-build': 'major:breaking'
'tauri-context': 'major:breaking'
'tauri-cli': 'major:breaking'
'@tauri-apps/cli': 'major:breaking'
```

core/tauri-utils/src/config.rs
lucasfernog: spoke with @FabianLars and we agree it's better to keep it as is...

FabianLars:
does it make sense to rename AppUrl to FrontendDist too? Quite liked that about the prodFrontend variant.

lucasfernog:
Also separated the enum variants so the JSON schema description is better (and the whole type makes more sense).

amrbashir:
That's better so maybe just change the name to `frontend`?

lucasfernog:
`frontend` isn't that clear to me IMO.. unless we use the `frontend > dist` thing (which is also not perfect along with devUrl)

lucasfernog:
I'll merge this one so we can move forward with the beta. If there's any changes left to do, we can do it in a separate PR.

lucasfernog (approved):

PR #8729, opened 2024-02-01
feat(tauri-plugin): generate permissions reference markdown file for plugin

lucasfernog:
Since the amount of generated permissions is huge, documenting it at the app-level seems like the best place to do so:

- central place for every available permission
- only lists permissions that are actually available to the app (does not show plugins that are not being compiled)
- autogenerated, no need for us to keep huge lists of permissions for every plugin on the docs website


amrbashir:
I don't see any value of this tbh, most users will not look into auto-generated files, and they will most definitely search for them on the doc site

lucasfernog:
Documenting all available permissions gets really tricky :|

amrbashir (dismissed):
This is relatively small change and I am fine with it but we still need to document these on the doc site somehow.

lucasfernog:
Maybe we can generate that on the plugin level instead? I know the docs repo has a git submodule for the plugin workspace...

amrbashir:
Yeah sure, we can maybe provide a helper method in `tauri-plugin` crate to facilitate this, it should also be updated with each change to the permissions list.

amrbashir (approved):

PR #8735, opened 2024-02-02
fix(examples): multiwindow example README.md

washanhanzi:
It seems the `window-create` feature is removed as shown in 

https://github.com/tauri-apps/tauri/blob/95da1a27476e01e06f6ce0335df8535b662dd9c4/core/tauri/Cargo.toml#L181-L184

but the `README.md` didn't reflect that.


FabianLars (approved):
thanks

PR #8736, opened 2024-02-02
fix(build): ignore changes on autogenerated files

lucasfernog:
- Writing to core/tauri/permissions is not a good idea, always triggers the build script to rerun
- rerun-if-changed should not be used for autogenerated permissions otherwise plugin build script always reruns


amrbashir (approved):

lucasfernog:
This is actually fine as long as we be careful not to always write the file (check if the contents changed first) and not rerun-if-changed it. That's why i'm keeping the autogenerated commands at the plugin crate itself and adding the reference markdown for each core plugin at #8737

PR #8739, opened 2024-02-03
fix(core): do not embed assets on dev when using a URL

lucasfernog:


lucasfernog:
Small regression from #8723

amrbashir (approved):

PR #8740, opened 2024-02-03
fix(cli): OsString::as_encoded_bytes is only available on Rust 1.74

lucasfernog:


amrbashir (approved):

FabianLars:
wait, do we have a msrv for the cli too? I always kinda thought we'd just go with latest. I guess it makes sense though since it's even more likely to be added to package managers etc with msrv restrictions.

lucasfernog:
it's always latest, I was just pissed at myself for having an outdated Rust on my Windows VM :joy: let's ignore this

PR #8743, opened 2024-02-03
fix(core): Rename FrontendDist::Dist to Dir

FabianLars:
Dist::Dist kinda triggers me and i'd like to rename it to Folder, Dir, or Directory. I started with Dir because Folder sounds weird to me and Directory is so long, but ultimately i'm fine with all 3 so please voice your preference :)

p.s. if you want to stick with Dist::Dist just close this.


FabianLars:
Actually wait, now that i see the pr preview i don't like Dir so much anymore D:

lucasfernog:
Dist::Dist is weird indeed, i didn't notice that

I don't like the word Folder but i use it anyway
I guess I'd go with Directory since Dist::Dir is also too close :joy:

lucasfernog (approved):

PR #8744, opened 2024-02-03
fix(cli): build command `features` and `bundles` can have multiple values

lucasfernog:

### DIFF
--- a/tooling/cli/src/build.rs
+++ b/tooling/cli/src/build.rs
@@ -54,7 +54,7 @@ pub struct Options {
   /// If `none` is specified, the bundler will be skipped.
   ///
   /// Note that the `updater` bundle is not automatically added so you must specify it if the updater is enabled.
-  #[clap(short, long, action = ArgAction::Append, num_args(0..))]
+  #[clap(short, long, action = ArgAction::Append, num_args(0..), value_delimiter = ',')]
   pub bundles: Option<Vec<String>>,
   /// JSON string or path to JSON file to merge with tauri.conf.json
   #[clap(short, long)]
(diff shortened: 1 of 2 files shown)


tooling/cli/src/build.rs
```diff
@@ -46,15 +46,15 @@ pub struct Options {
...
+  #[clap(short, long, action = ArgAction::Append, num_args(0..))]
   pub features: Option<Vec<String>>,
   /// Space or comma separated list of bundles to package.
   ///
   /// Each bundle must be one of `deb`, `rpm`, `appimage`, `msi`, `app` or `dmg` on MacOS and `updater` on all platforms.
   /// If `none` is specified, the bundler will be skipped.
   ///
   /// Note that the `updater` bundle is not automatically added so you must specify it if the updater is enabled.
-  #[clap(short, long, value_delimiter = ',')]
+  #[clap(short, long, action = ArgAction::Append, num_args(0..))]
```
amrbashir: Not sure what this solves tbh, `Option<Vec<T>>` will use `ArgAction::Append` and `num_args(0..)`, This is kinda redundant. Also not using value_delimiter, means we need to handle it ourselves which I removed in favor of clap built-in value delimiter

lucasfernog: that's not really true, just try it yourself, `tauri build --features a b` won't work

lucasfernog: plus there's more args using this

amrbashir: Ah maybe clap docs is not clear about the num_args thing but they do use append by default `--features a,b,g --features s` should work

amrbashir: Anyways, if we go with manual delimiter, we need to add the code to handle it

amrbashir (approved):

PR #8756, opened 2024-02-04
refactor(core): enhance app rerun-if-changed for capabilities and frontend dist

lucasfernog:
Spoke about this with @FabianLars ... currently we do not rerun the build script when a new capability file is created, leaving the app in a weird state where the permissions are not parsed and the user has no idea why. The idea was to move the schemas from `capabilities/schemas` to `gen/schemas` so we can `rerun-if-changed=capabilities`.

I'm also applying the same logic to the plugin permissions, which currently does not trigger a rebuild if a new permission is added.


FabianLars:
Would it be bad to simply print the rerun instruction with `CAPABILITIES_SCHEMA_FOLDER_NAME` directly instead of going with the files parent() method.

I think all the templates contain a file already so your way should work but if there is no file yet then adding one does still not trigger a rebuild.

Maybe we call this more of an edge case idk, since as i said the v2 templates contain it and migrating from v1 to v2 causes a rebuild anyway. I just used an "empty" v2 beta project to test this and noticed it that way.

Edit: hmmm i assume this wasn't done to prevent constant rebuilds if the folder doesn't exist :/

lucasfernog:
we can either assume the folder is there (it should) but i was also trying to support custom capability folder path which are possible via the tauri-build attributes

lucasfernog:
i'll make a change to be smarter about it

lucasfernog:
Pushed the same change for plugins (which does force a rebuild on the first **local** build since we now commit all autogenerated files).

lucasfernog:
first local build after any permission change, but i think it's a good trade off between feature and DX

FabianLars (approved):
My test flow still doesn't reallyyy work but at least manually re-running tauri dev triggers a rebuild now and that's good enough, thanks for that :)

hjmallon:
I have just updated to v2 in an existing application. 

1. Is there any reason we dont generate the schemas to the `target` folder which is already gitignored etc?
2. This change has not been reproduced in tauri-docs and create-tauri-app

(e.g. https://github.com/tauri-apps/tauri-docs/blob/6d9239c2f6df6d0d0690d6dd83c0bc17de524ae7/src/content/docs/references/v2/acl.mdx#L21)

PR #8761, opened 2024-02-04
fix(core): build on unix failing

lucasfernog:

### DIFF
--- a/.changes/fix-tauri-build-unix.md
+++ b/.changes/fix-tauri-build-unix.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch:bug
+---
+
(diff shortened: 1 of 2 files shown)


lucasfernog:
I'm not sure what happened.. zbus required [nix::unistd::Uid](https://docs.rs/nix/0.26.4/nix/unistd/struct.Uid.html) but it does not enable the user feature.. I think some other crate was enabling this and hiding the issue?

FabianLars (approved):
wild

simonhyll (approved):

marc2332:
I think this could be reverted once zbus releases a patch, context: https://github.com/dbus2/zbus/pull/568

PR #8766, opened 2024-02-04
fix(multiwebview): remove webview from store on close

lucasfernog:
Kinda hacky? should we think about implementing event loop events for webviews instead?


FabianLars (approved):

lucasfernog:
FYI I'm currently playing with an app that needs the multiwebview support, expect to see a bunch of improvements soon!

Issue #8760, opened 2024-02-04
[bug] HTML encoding issue for data-url


huangmingg:
### Describe the bug

Html content containing special characters are not properly encoded when creating a new webview window using data url.

### Reproduction

```
let html = r#"
    <html>
      <body>
       special character incoming ###
       thai characters incoming สระ
      </body>
    </html>
  "#;
  let data = format!("data:text/html,{}", html);
  let _docs_window = tauri::WindowBuilder::new(
    &handle,
    "myWindow",
    tauri::WindowUrl::External(data.parse().unwrap())
  )

  .build()
  .unwrap();
```

does not work, while reading from an external file

```
  let html = r#"
    <html>
      <body>
       special character incoming ###
       thai characters incoming สระ
      </body>
    </html>
  "#;
  std::fs::write(path.clone(), html).expect("Unable to write file");
  let data = format!("data:text/html,{}", html);
  let _docs_window = tauri::WindowBuilder::new(
    &handle,
    "myWindow",
    tauri::WindowUrl::App("external.html".parse().unwrap())
  )
  .build()
  .unwrap();
```

works

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.19045 X64
    ✔ WebView2: 120.0.2210.144
    ✔ MSVC: Visual Studio Build Tools 2022
    ✔ rustc: 1.74.1 (a28077b28 2023-12-04)
    ✔ cargo: 1.74.1 (ecb9851af 2023-10-18)
    ✔ rustup: 1.26.0 (5af9b9484 2023-04-05)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (default)
    - node: 16.16.0
    - pnpm: 6.11.0
    - yarn: 1.22.15
    - npm: 8.11.0

[-] Packages
    - tauri [RUST]: 1.5.3
    - tauri-build [RUST]: 1.5.0
    - wry [RUST]: 0.24.6
    - tao [RUST]: 0.16.5
    - @tauri-apps/api [NPM]: 1.5.1 (outdated, latest: 1.5.3)
    - @tauri-apps/cli [NPM]: 1.5.6 (outdated, latest: 1.5.9)

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: Set automatically by Vue CLI plugin
    - devPath: Set automatically by Vue CLI plugin
    - framework: Vue.js (Nuxt)
    - bundler: Webpack
```

### Additional context

[...]

PR #8779, opened 2024-02-05
fix(core) Fix HTML encoding in webview rendered via data url

huangmingg:
closes https://github.com/tauri-apps/tauri/issues/8760 in 1.x


huangmingg:
Tested on windows:
image

If anyone can test the example on the other platforms
`cargo run --example data-url --features window-data-url`

core/tauri-runtime-wry/src/lib.rs
```diff
@@ -3246,13 +3246,35 @@ fn create_webview<T: UserEvent>(
...
+  let mut webview_builder =
+      WebViewBuilder::new(window).map_err(|e| Error::CreateWebview(Box::new(e)))?;
+
+  // use with_html method if html content can be extracted from url.
+  // else defaults to with_url method
+  webview_builder = if let Some(html_string) = tauri_utils::html::extract_html_content(&url_str) {
+      // we run a decode here on the html_string for sanity check
+      // the content should be fully decoded when passing to WRY, which
+      // will be responsible for handling the encoding based on the OS.
+      let html =
```
amrbashir: I don't think `tauri-runtime-wry` should handle this case, it should be handled by `tauri` crate and passed as a valid url.

core/tauri/Cargo.toml
```diff
@@ -358,3 +359,8 @@ path = "../../examples/streaming/main.rs"
 name = "isolation"
 path = "../../examples/isolation/main.rs"
 required-features = [ "isolation" ]
+
+[[example]]
+name = "data-url"
+path = "../../examples/data-url/main.rs"
+required-features = [ "window-data-url" ]
```
amrbashir: I don't think we need to have an example just for data urls

core/tauri/Cargo.toml
```diff
@@ -96,6 +96,7 @@ encoding_rs = "0.8.31"
 sys-locale = { version = "0.2.3", optional = true }
 tracing = { version = "0.1", optional = true }
 indexmap = { version = "1", features = [ "std", "serde" ], optional = true }
+urlencoding = "2.1.3"
```
amrbashir: No need to bring another crate, we already have a similar crate `percent-encoding`

amrbashir (changes requested):

core/tauri-runtime-wry/src/lib.rs
huangmingg: I feel like this should be the responsibility of `wry` instead of `tauri `or even `tauri-runtime-wry`, as it is dependent on the underlying web engine on each platform on how the html string is parsed. For windows, `wry `does the following to render html content on webview2:

```
      unsafe {
        webview
          .NavigateToString(PCWSTR::from_raw(encode_wide(html).as_ptr()))
          .map_err(webview2_com::Error::WindowsError)?;
      }
```

which encodes the HTML content. So, by passing in an already encoded uri, it will result in a double encoding. This is not an issue on linux where it seems to handle this issue by default.

I am rather unfamiliar with the underlying technologies, and this is based on my rudimentary understanding from looking into this issue thus far. Would greatly appreciate any corrections or insights on this.

core/tauri-runtime-wry/src/lib.rs
```diff
@@ -3247,13 +3247,27 @@ fn create_webview<T: UserEvent>(
...
+  // else defaults to with_url method
+  webview_builder = if let Some(html_string) = tauri_utils::html::extract_html_content(&url) {
+    webview_builder
+      .with_html(html_string)
+      .map_err(|e| Error::CreateWebview(Box::new(e)))?
+  } else {
+    webview_builder
+      .with_url(&url)
+      .map_err(|e| Error::CreateWebview(Box::new(e)))?
+  };
```
amrbashir, line 3264: `tauri-runtime-wry` shouldn't do these checks at all, instead, `PendingWindow` should have a field `html: Option<String>` and if it is `Some`, we use `with_html`

core/tauri-runtime-wry/Cargo.toml
```diff
@@ -21,6 +21,7 @@ rand = "0.8"
 raw-window-handle = "0.5"
 tracing = { version = "0.1", optional = true }
 arboard = { version = "3", optional = true }
+percent-encoding = "2.1"
```
amrbashir, line 24: this is only needed on Linux, let's revert this back

core/tauri-runtime/Cargo.toml
```diff
@@ -50,3 +50,4 @@ system-tray = [ ]
 macos-private-api = [ ]
 global-shortcut = [ ]
 clipboard = [ ]
+window-data-url = [ "tauri-utils/window-data-url" ]
```
amrbashir, line 53: and revert this as well

core/tauri-utils/Cargo.toml
```diff
@@ -38,6 +38,7 @@ semver = "1"
 infer = "0.13"
 dunce = "1"
 log = "0.4.20"
+data-url = { version = "0.3.1", optional = true }
```
amrbashir, line 41: ditto

core/tauri-utils/Cargo.toml
```diff
@@ -54,3 +55,4 @@ process-relaunch-dangerous-allow-symlink-macos = [ ]
 config-json5 = [ "json5" ]
 config-toml = [ "toml" ]
 resources = [ "glob", "walkdir" ]
+window-data-url = [ "data-url" ]
```
amrbashir, line 58: ditto

core/tauri-utils/src/html.rs
```diff
@@ -286,6 +286,20 @@ pub fn inline_isolation(document: &mut NodeRef, dir: &Path) {
...
+}
+
+/// Temporary naive method to extract data from html data string
+pub fn extract_html_content(input: &str) -> Option<&str> {
+  if input.starts_with("data:text/html,") {
+    Some(&input[15..])
+  } else {
+    None
+  }
+}
```
amrbashir, line 301: let's move these back to `tauri` crate as they are only used there for now

core/tauri/src/manager.rs
```diff
@@ -1005,23 +1007,46 @@ impl<R: Runtime> WindowManager<R> {
...
-    if let Some(csp) = self.csp() {
-      if url.scheme() == "data" {
-        if let Ok(data_url) = data_url::DataUrl::process(url.as_str()) {
-          let (body, _) = data_url.decode_to_vec().unwrap();
-          let html = String::from_utf8_lossy(&body).into_owned();
-          // naive way to check if it's an html
-          if html.contains('<') && html.contains('>') {
+    match (
+      url.scheme(),
+      tauri_utils::html::extract_html_content(url.as_str()),
```
amrbashir, line 1012: we shouldn't call `extract_html_content` unless we are sure it is a data url
```rs
#[cfg(feature = "window-data-url")]
if ur.scheme() == "data" {
  let html = extract_html_content();
  // ...
}
```

core/tauri/src/manager.rs
```diff
@@ -1005,23 +1007,46 @@ impl<R: Runtime> WindowManager<R> {
...
+    ) {
+      #[cfg(feature = "window-data-url")]
+      ("data", Some(html_string)) => {
+        // There is an issue with the external DataUrl where HTML containing special characters
+        // are not correctly processed. A workaround is to first percent encode the html string,
+        // before it processed by DataUrl.
+        let encoded_string = percent_encoding::utf8_percent_encode(html_string, percent_encoding::NON_ALPHANUMERIC).to_string();
```
amrbashir, line 1019: we shouldn't encode the URL, we should expect the user has already encoded it as per https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs 
which makes this PR kinda uneeded?

amrbashir (changes requested):

core/tauri/src/manager.rs
gentean: I found that data URLs are not encoded by default on Linux, but they are encoded by default on macOS. I haven't tried on Windows yet. I expect consistency across platforms, with no encoding.

The following content will be encoded on macOS, which is not what I expected.
```js
const webview = new WebviewWindow('print', {
        url: `data:text/html,<html><body>你好，世界</body></html>`,
        center: true,
        visible: true,
        width: 300,
        height: 300,
});

```

PR #8783, opened 2024-02-05
fix(cli): regression on --config not accepting file paths

lucasfernog:

### DIFF
--- a/tooling/cli/src/build.rs
+++ b/tooling/cli/src/build.rs
@@ -6,7 +6,7 @@ use crate::{
   helpers::{
     app_paths::{app_dir, tauri_dir},
     command_env,
-    config::{get as get_config, FrontendDist, HookCommand},
+    config::{get as get_config, ConfigHandle, FrontendDist, HookCommand},
     updater_signature::{secret_key as updater_secret_key, sign_file},
   },
   interface::{AppInterface, AppSettings, Interface},
@@ -81,7 +81,7 @@ pub fn command(mut options: Options, verbosity: u8) -> Result<()> {
     options.target.clone(),
   )?;
 
-  setup(target, &interface, &mut options, false)?;
+  setup(&interface, &mut options, config.clone(), false)?;
 
   let config_guard = config.lock().unwrap();
   let config_ = config_guard.as_ref().unwrap();
@@ -266,13 +266,11 @@ pub fn command(mut options: Options, verbosity: u8) -> Result<()> {
 }
 
 pub fn setup(
-  target: Target,
   interface: &AppInterface,
   options: &mut Options,
+  config: ConfigHandle,
   mobile: bool,
 ) -> Result<()> {
-  let config = get_config(target, options.config.as_ref().map(|c| &c.0))?;
-
   let tauri_path = tauri_dir();
   set_current_dir(tauri_path).with_context(|| "failed to change current working directory")?;
 
--- a/tooling/cli/src/build.rs
+++ b/tooling/cli/src/build.rs
@@ -6,12 +6,11 @@ use crate::{
   helpers::{
     app_paths::{app_dir, tauri_dir},
     command_env,
-    config::{get as get_config, FrontendDist, HookCommand, MERGE_CONFIG_EXTENSION_NAME},
-    resolve_merge_config,
+    config::{get as get_config, FrontendDist, HookCommand},
     updater_signature::{secret_key as updater_secret_key, sign_file},
   },
   interface::{AppInterface, AppSettings, Interface},
-  CommandExt, Result,
+  CommandExt, ConfigValue, Result,
 };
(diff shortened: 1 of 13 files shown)


tooling/cli/src/build.rs
```diff
@@ -272,22 +271,17 @@ pub fn setup(
   options: &mut Options,
   mobile: bool,
 ) -> Result<()> {
-  let (merge_config, merge_config_path) = resolve_merge_config(&options.config)?;
-  options.config = merge_config;
-
-  let config = get_config(target, options.config.as_deref())?;
+  let config = get_config(target, options.config.as_ref().map(|c| &c.0))?;
```
amrbashir: can't we just config as an argument to the `setup` function?

tooling/cli/src/dev.rs
```diff
@@ -165,10 +164,7 @@ pub fn setup(
   options: &mut Options,
   mobile: bool,
 ) -> Result<()> {
-  let (merge_config, _merge_config_path) = resolve_merge_config(&options.config)?;
-  options.config = merge_config;
-
-  let config = get_config(target, options.config.as_deref())?;
+  let config = get_config(target, options.config.as_ref().map(|c| &c.0))?;
```
amrbashir: ditto

tooling/cli/src/dev.rs
```diff
@@ -345,14 +341,17 @@ pub fn setup(
...
           c.build.dev_url = dev_url.clone();
-          options.config = Some(serde_json::to_string(&c).unwrap());
+          options.config = Some(crate::ConfigValue(serde_json::to_value(&c).unwrap()));
         } else {
-          options.config = Some(format!(r#"{{ "build": {{ "devUrl": "{server_url}" }} }}"#))
+          options.config = Some(crate::ConfigValue(
+            serde_json::from_str(&format!(r#"{{ "build": {{ "devUrl": "{server_url}" }} }}"#))
+              .unwrap(),
+          ))
         }
```
amrbashir: We should probably refactor this to change the value in-place without deserializing and serializing again.

tooling/cli/src/mobile/mod.rs
```diff
@@ -178,13 +182,15 @@ fn setup_dev_config(
...
-        config_extension.replace(serde_json::to_string(&c).unwrap());
+        config_extension.replace(ConfigValue(serde_json::to_value(&c).unwrap()));
       } else {
-        config_extension.replace(format!(r#"{{ "build": {{ "devUrl": "{url}" }} }}"#));
+        config_extension.replace(ConfigValue(
+          serde_json::from_str(&format!(r#"{{ "build": {{ "devUrl": "{url}" }} }}"#)).unwrap(),
+        ));
       }
-      reload_config(config_extension.as_deref())?;
+      reload_config(config_extension.as_ref().map(|c| &c.0))?;
```
amrbashir: ditto

amrbashir (changes requested):

tooling/cli/src/build.rs
lucasfernog: we do not do this overall in the codebase, get_config is cached so it's cheap if it's already been resolved, otherwise we need to pass the config around all codebase

tooling/cli/src/dev.rs
lucasfernog: good idea, done

tooling/cli/src/build.rs
amrbashir: yeah, but config is parsed right before calling `setup` so it is better to just pass the config to it, than getting from cache again

tooling/cli/src/mobile/mod.rs
```diff
@@ -178,13 +182,20 @@ fn setup_dev_config(
...
+          .and_then(|root| root.get_mut("build"))
+          .and_then(|build| build.as_object_mut())
+        {
+          build.insert("devUrl".into(), url.to_string().into());
+        }
       } else {
-        config_extension.replace(format!(r#"{{ "build": {{ "devUrl": "{url}" }} }}"#));
+        config_extension.replace(ConfigValue(
+          serde_json::from_str(&format!(r#"{{ "build": {{ "devUrl": "{url}" }} }}"#)).unwrap(),
+        ));
```
amrbashir: would it be faster to use serde_json::json! macro here?

lucasfernog: oops i checked it and couldn't find how to add dynamic values, turns out it was in the macro docs itself :joy:

lucasfernog:
stupid android test CI still broken..

tooling/cli/src/build.rs
lucasfernog: you're right it wasn't that complicated in the end.. thanks

amrbashir (approved):

PR #8797, opened 2024-02-06
refactor(core): allow referencing capabilities on the Tauri config file

lucasfernog:
The current capability file design automatically adds all files in the capabilities folder to the application. While this is convenient for the developer, it restricts dynamic use cases (for instance plugins that are only enabled on debug or with an associated Cargo feature). After talking with @tweidinger we agreed we should add a reference on tauri.conf.json where the dev must specify which capabilities to include.

This PR adds a capabilities array inside app > security > capabilities where you can reference capabilities by identifier or inline capability objects (though this one isn't ideal since it lacks JSON schema support).

I'll open an issue later today detailing the dynamic capability implementation we can support.

### DIFF
--- a/core/tauri-codegen/src/context.rs
+++ b/core/tauri-codegen/src/context.rs
@@ -15,7 +15,7 @@ use tauri_utils::acl::capability::Capability;
 use tauri_utils::acl::plugin::Manifest;
 use tauri_utils::acl::resolved::Resolved;
 use tauri_utils::assets::AssetKey;
-use tauri_utils::config::{Config, FrontendDist, PatternKind};
+use tauri_utils::config::{CapabilityEntry, Config, FrontendDist, PatternKind};
 use tauri_utils::html::{
   inject_nonce_token, parse as parse_html, serialize_node as serialize_html_node,
 };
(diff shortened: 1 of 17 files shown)


lucasfernog:
Maybe we should default to include all capabilities if the config array is empty? I know @FabianLars would like that so we won't deal with another "hidden" breaking change.

FabianLars:
Yup, probably. But it's still beta so if it's objectively better than just break it 🤷 

(i also don't really get the issue itself tbh so idk)

lucasfernog:
I like how the latest change is a balance between DX and configurability.

tooling/cli/templates/tauri.conf.json
```diff
@@ -19,7 +19,8 @@
       }
     ],
     "security": {
-      "csp": null
+      "csp": null,
+      "capabilities": ["default-plugins"]
```
amrbashir: I don't like the `default-plugins` identifier, could we change that to just `default`?
```suggestion
      "capabilities": ["default"]
```
or should we inline the default capability here, so users used to allowlist, see something familiar? and we could try to find a way to list default plugins in the schema so they have auto-completion

core/tauri-codegen/src/context.rs
```diff
@@ -381,14 +381,35 @@ pub fn context_codegen(data: ContextData) -> Result<TokenStream, EmbeddedAssetsE
...
   let capabilities_file_path = out_dir.join(CAPABILITIES_FILE_NAME);
-  let capabilities: BTreeMap<String, Capability> = if capabilities_file_path.exists() {
+  let mut capabilities_from_files: BTreeMap<String, Capability> = if capabilities_file_path.exists()
+  {
     let capabilities_file =
       std::fs::read_to_string(capabilities_file_path).expect("failed to read capabilities");
     serde_json::from_str(&capabilities_file).expect("failed to parse capabilities")
   } else {
     Default::default()
   };
```
amrbashir, line 391: A small optimization is to read from disk only if capabilities from config is empty

amrbashir (changes requested):

core/tauri-codegen/src/context.rs
lucasfernog: not possible since the config can also reference capabilities defined in the filesystem

tooling/cli/templates/tauri.conf.json
lucasfernog: let's make this change when we change the config schema to be stored in the project itself (so we can modify it)

core/tauri-codegen/src/context.rs
amrbashir: ah, didn't see that one

amrbashir (approved):

PR #8801, opened 2024-02-06
refactor: webview as child for WebviewWindow, enhance resizing closes #6322

### LINKED ISSUE #6322: [bug] Resizing on Windows is much slower in Tauri than in Wry
### Describe the bug

We all know that Chromium is super slow to resize its content, it's present in all Chromium browsers too, but while trying out a few workarounds to make it less noticeable, like setting the window background color via win32 apis (changing the webview2 background doesn't change anything btw because the webview itself is not resized in time), i noticed that this is pretty much no problem in Wry but it got even worse in Tauri than the last time i really paid attention to it.
Here's a recording. Its resolution and fps makes it look a bit worse than it actually is.

https://user-images.githubusercontent.com/30730186/210389507-8367d374-c65f-4238-9e76-7aea3d65ffb4.mp4

Note that i already disabled Tauri's window event handlers in its source code for that recording - no noticeable difference. The 2 apps in the video are release builds pointing to my local dev server. debug builds, and release builds with bundled artifacts have the same behavior.

I could not reproduce this issue on Linux.

### Reproduction

Build literally any app on Windows. It's present even in the most lightweight vanillajs apps.

### Expected behavior

i don't expect Tauri to be as fast as Wry but it shouldn't feel like my computer is crashing.

### Platform and versions

Environment
  › OS: Windows 10.0.19045 X64
  › Webview2: 108.0.1462.54
  › MSVC:
  › Node.js: 18.12.1
  › npm: 8.19.2
  › pnpm: 7.21.0
  › yarn: 1.22.19
  › rustup: 1.25.1
  › rustc: 1.66.0
  › cargo: 1.66.0

Packages
  › @tauri-apps/cli [NPM]: 1.2.2
  › @tauri-apps/api [NPM]: 1.2.0
  › tauri [RUST]: 1.2.3, (overwritten with local path for testing but released version has the same problem)
  › tauri-build [RUST]: 1.2.1,
  › tao [RUST]: 0.15.8,
  › wry [RUST]: 0.23.4,

App
  › build-type: bundle
  › CSP: unset
  › distDir: ../dist
  › framework: React
  › bundler: Vite

App directory structure
  ├─ dist
  ├─ node_modules
  ├─ src
  └─ src-tauri

### Stack trace

_No response_

### Additional context

I also added some window event logging in Wry's source code for testing and that looks jittery too when used in Tauri.

lucasfernog:
Doesn't work 100% yet :(


lucasfernog:
I've been getting help from @FabianLars to get this to work but no luck yet.

The example from https://github.com/tauri-apps/tauri/issues/6322#issuecomment-1872491965 also has a window.request_redraw() but I don't think that helps.

core/tauri-runtime-wry/src/lib.rs
```diff
@@ -3342,27 +3400,7 @@ fn create_webview<T: UserEvent>(
...
-        target_os = "ios",
-        target_os = "android"
-      )))]
-      let builder = {
-        let vbox = window.default_vbox().unwrap();
-        WebViewBuilder::new_gtk(vbox)
-      };
-      builder
-    }
+    WebViewBuilder::new_as_child(&window)
```
amrbashir, line 3443: iirc, using `new_as_child` has shortcomings on macOS and probably should not be used by default
cc @wusyong

core/tauri-runtime-wry/src/lib.rs
```diff
@@ -117,6 +118,8 @@ type IpcHandler = dyn Fn(String) + 'static;
 mod webview;
 pub use webview::Webview;
 
+const RESIZE_INTERVAL: Duration = Duration::from_millis(1000 / 120);
```
amrbashir, line 121: is there a reason for this specific interval? or should we get it from monitor refresh rate?

amrbashir (changes requested):
Tbh I am not super sold on implementing this inside tauri and would recommend that this be kept up to users to implement (I believe this will require exposing more of these events to `App::run` method).

And while we are at it, I don't also like the auto-resize attribute, if user opted explicitly into multiwebview, they should be expected to handle that on their own but I am not opposed to it completely.

core/tauri-runtime-wry/src/lib.rs
FabianLars: Anything faster was getting quite laggy in my testing especially times 4 for the multiwebview example.

But making it variable is still useful, at least to get it lower than 120 for slower displays.

FabianLars:
I'm not super happy with it either, at least with how it currently is implemented and how less of an effect it has on a multiwebview app, but I disagree that it should be handled on the user side. Without anything on our side resizing on windows is really horrible and basically unusable for multiwebview windows.

Maybe have it opt-in as a compromise?

amrbashir:
No need for an option, let's roll with it and see if there is any negative feedback down the road.

lucasfernog:
@FabianLars @amrbashir can you check what else we can do on Windows to enhance this?

lucasfernog:
Probably needs a better approach.

Issue #8799, opened 2024-02-06
[feat] add capabilities at runtime


lucasfernog:
### Describe the problem

The ACL system implemented by #8428 is a static implementation of access control for IPC commands. Dynamic usage like starting a localhost server that needs access to the IPC is not supported securely.

An example is the [localhost plugin](https://github.com/tauri-apps/plugins-workspace/tree/v2/plugins/localhost) which starts a localhost server to serve your app assets.
To support this use case, we need to add a runtime API that lets you allow commands on a specific execution context.

### Describe the solution you'd like

Add a new function in the runtime authority to enable new permissions.

lucasfernog:
This is similar to #8798 but adding the capability at runtime is better for performance and validation.

PR #8802, opened 2024-02-06
feat(codegen): allow defining additional capabilities, closes #8798

### LINKED ISSUE #8798: [feat] conditionally add capabilities
### Describe the problem

The current capability design automatically pulls every capability file and resolves to the app runtime authority. While this is a good approach for developer experience, it is static and does not support more dynamic applications such as:
- nightly releases including new features
- feature-flag based multi application mode (for instance an admin vs user shared app that uses Cargo features to toggle what is compiled)
- debug-only plugins (like developer productivity tools)

we should add an option to define capabilities in Rust code. Ideally this should still be done at compile time. Since the current capabilities resolution is done in the Context generation, we could extend it to allow defining capabilities conditionally:

```rust
tauri::generate_context!(
  capabilities = [
	  #[cfg(debug_assertions)]
	  "./capabilities/debug.json",
	  #[cfg(feature = "admin")]
	  "./capabilities/admin.json",
  ]
)
```

#8797 implements the first step for this feature, without that change users need to define conditionally added capabilities in a separate folder otherwise they would be enabled automatically.

### Alternatives considered

We could also allow defining the capability at runtime instead, but I'll open a separate issue for that one.

### Additional context

_No response_

lucasfernog:
Changes the `tauri::generate_context` proc macro to parse the `capabilities` attribute:
```rust
tauri::generate_context!(
  capabilities = [
    "../capabilities/base.json",
    #[cfg(feature = "beta")]
    "../capabilities/beta.json"
  ]
```

In this case, the additional capabilities are parsed at runtime (limitation of proc-macro, cannot resolve the `#[cfg]` attributes :(

Additionally, the tauri-build context codegen method now has a `capability` method:

```rust
fn main() {
  let mut codegen = tauri_build::CodegenContext::new();
  if !cfg!(feature = "custom-protocol") {
    codegen = codegen.dev();
  }

  #[cfg(feature = "admin")]
  codegen = codegen.capability("./capabilities/admin.json");

  tauri_build::try_build(tauri_build::Attributes::new().codegen(codegen))
    .expect("failed to run tauri-build");
}
```

This API is more powerful since it can resolve the capability at compile time.


lucasfernog-crabnebula:
RuntimeAuthority::add_capability implements the base API for #8799

core/tauri/src/ipc/authority.rs
```diff
@@ -64,15 +65,15 @@ impl Origin {
 }
 
 impl RuntimeAuthority {
-  pub(crate) fn new(resolved_acl: Resolved) -> Self {
+  #[doc(hidden)]
```
amrbashir, line 68: do we need to hide this?

core/tauri/src/ipc/authority.rs
```diff
@@ -84,6 +85,93 @@ impl RuntimeAuthority {
     }
   }
 
+  #[doc(hidden)]
```
amrbashir, line 88: does this need to be hidden also? and have this internal name? why not just expose it?

core/tauri-macros/src/context.rs
```diff
@@ -26,51 +27,98 @@ impl Parse for ContextItems {
...
+                      }
+                    } else {
+                      Err(syn::Error::new(
+                        input.span(),
+                        "unexpected expression for capability",
+                      ))
+                    }
+                  })
+                  .collect::<Result<Vec<_>, syn::Error>>()?,
+              );
```
amrbashir: too much indentation imo, we should extract this into a function or try to reduce indentation levels

core/tauri-utils/src/acl/resolved.rs
```diff
@@ -315,8 +315,7 @@ struct ResolvedCommandTemp {
   pub scope: Vec<ScopeKey>,
   pub resolved_scope_key: Option<ScopeKey>,
 }
-
-fn resolve_command(
+fn resolve_command_temp(
```
amrbashir: why the `_temp` suffix?

core/tauri-utils/src/acl/resolved.rs
```diff
@@ -307,6 +248,65 @@ fn parse_window_patterns(windows: HashSet<String>) -> Result<Vec<glob::Pattern>,
   Ok(patterns)
 }
 
+fn resolve_capability<F: FnMut(&str, &str, Commands, Scopes)>(
```
amrbashir: the function name doesn't really convey what it does, should probably rename to `with_resolved_commands`?

amrbashir (changes requested):

core/tauri/src/ipc/authority.rs
lucasfernog: I don't want anyone to use this API (though they shouldn't.. it's useless without attaching to a manager - which you can't - and it's only used during codegen)

core/tauri/src/ipc/authority.rs
lucasfernog: not stable yet.. maybe we move it under the unstable feature flag, but I don't like the idea of allowing devs to allow an specific command, they should go through the whole capabilities > permissions folow, this was just added for our examples which can't use a build script so they do not go through our ACL definitions
cc @tweidinger

core/tauri-macros/src/context.rs
lucasfernog: there's a way to improve this one, i'll push

lucasfernog:
Needs #8797

lucasfernog:
I would be ok with removing the option from the `generate_context` macro too since it leads you into thinking it's resolved at compile time but it isn't (the tauri-build option is much better).

core/tauri/src/ipc/authority.rs
amrbashir: Ah, nvm, I thought it had more methods that users could use for their own ACL implementation or something.

amrbashir:
> In this case, the additional capabilities are parsed at runtime (limitation of proc-macro, cannot resolve the `#[cfg]` attributes :(

Can't we just provide an option to specify a capability file without `#[cfg]` and so we can parse this at compile-time? if they want to use `#[cfg]`, they should do so on the whole macro

```rs
#[cfg(not(beta))]
let context = tauri::generate_context!(capabilities = ["path/to/capability.json"]);
#[cfg(beta)]
let context = tauri::generate_context!(capabilities = ["path/to/capability-beta.json"]);
```

lucasfernog:
Let's roll with that approach for now until Rust supports expanding cfg on the proc macro.

amrbashir (approved):

Issue #8780, opened 2024-02-05
[bug] [v2.0.0-beta] tauri generates invalid schema json


eikowagenknecht:
### Describe the bug

Just upgraded from Tauri v2-alpha14 to -beta1, ran `npm run tauri migrate` to update the `tauri.conf.json`, started the application once (looked fine on first sight) and then my precommit hook reminded my that there were some new non-valid .json files:

- /src-tauri/gen/schemas/desktop-schema.json contains only `{schema_str}`
- /src-tauri/gen/schemas/windows-schema.json as well

The files are regenerated on every start of the application.

Is this to be expected?

### Expected behavior

Generate valid .json files.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.22635 X64
    ✔ WebView2: 121.0.2277.98
    ✔ MSVC: Visual Studio Community 2022
    ✔ rustc: 1.74.0 (79e9716c9 2023-11-13)
    ✔ cargo: 1.74.0 (ecb9851af 2023-10-18)
    ✔ rustup: 1.26.0 (5af9b9484 2023-04-05)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (default)
    - node: 20.10.0
    - npm: 10.2.4

[-] Packages
    - tauri [RUST]: 2.0.0-beta.2
    - tauri-build [RUST]: 2.0.0-beta.1
    - wry [RUST]: 0.35.2
    - tao [RUST]: 0.25.0
    - @tauri-apps/api [NPM]: 2.0.0-beta.0
    - @tauri-apps/cli [NPM]: 2.0.0-beta.1

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:1420/
    - framework: React
    - bundler: Vite
```

FabianLars:
I assume this was fixed in https://github.com/tauri-apps/tauri/pull/8782 and we forgot to link it. Thanks for the report :)

PR #8803, opened 2024-02-06
fix: creation of schemas

kehrazy:
currently, tauri generates schemas like this:
```json
{schema_str}
```

rust seems to treat the value like a string literal, not a format statement.

this commit should fix this error.


olivierlemasle:
It should already be fixed by #8782 (already merged, but not released - cf #8775 and #8780)

amrbashir:
As mentioned by @olivierlemasle , this is already fixed, thanks for your contribution

PR #8806, opened 2024-02-07
feat(core): add Manager::add_capability, closes #8799

### LINKED ISSUE #8799: [feat] add capabilities at runtime

lucasfernog:
The tricky part here was making the scope mutable. Had to use an Arc, but I'm not sure if there's a better way to do it.


amrbashir (dismissed):

lucasfernog:
Needs #8802

amrbashir (approved):

PR #8810, opened 2024-02-07
fix(cli): fix code output of `tauri plugin android init`

giripriyadarshan:
double quotes `"` wasn't matching in plugin android init command


amrbashir (approved):
Thank you for your contributions

PR #8793, opened 2024-02-06
fix: correct invoke command for toggling devtools by hotkey, closes #8771

pewsheen:
Issue: [tauri#8771](https://github.com/tauri-apps/tauri/issues/8771)

The scope used by webview toggle_devtools script was set to `window`, it should be `webview` now.

PR #8819, opened 2024-02-08
fix(core): use event.code to check which key pressed

pewsheen:
Hitting `CMD + ALT + I`, `event.key` may be `^, Dead, i, I` depends on which modified key was hit at last.
However, `event.code` will always give you `KeyI` and I think this is more suitable for the use case.


FabianLars (dismissed):
Is the changefile needed if https://github.com/tauri-apps/tauri/pull/8793/files with a similar changefile wasn't released yet?

amrbashir (approved):

PR #8826, opened 2024-02-09
fix: target specific window for window events

vonPB:
When using the [onCloseRequested JS Api](https://github.com/tauri-apps/tauri/blob/cf3e40cc473dd0bb77f759fd59c1594af263429b/tooling/api/src/window.ts#L1689C29-L1689C39), the event was triggered even when closing other windows, not just the one of interest.

I believe it would be more logical for this event to be dispatched exclusively for the current window's close requests.

Before:

https://github.com/tauri-apps/tauri/assets/65032978/81220baa-6c0e-4677-bb65-e6cc797e0a8a

After:

https://github.com/tauri-apps/tauri/assets/65032978/d557beec-2c18-46cc-9643-fc9202c4f2de

### DIFF
--- a/.changes/tauri-close-requested-target-specific.md
+++ b/.changes/tauri-close-requested-target-specific.md
@@ -1,0 +1,5 @@
+---
+'tauri': 'patch:bug'
+---
+
+Fix JS `onCloseRequested` catching close event from other windows.
--- a/core/tauri/src/lib.rs
+++ b/core/tauri/src/lib.rs
@@ -713,21 +713,23 @@ pub trait Manager<R: Runtime>: sealed::ManagerBase<R> {
     #[cfg(feature = "tracing")]
     tracing::Span::current().record("target", format!("{target:?}"));
 
-    self.manager().emit_filter(event, payload, |s| match s {
-      t @ EventTarget::Window { label }
-      | t @ EventTarget::Webview { label }
-      | t @ EventTarget::WebviewWindow { label } => {
-        if let EventTarget::AnyLabel {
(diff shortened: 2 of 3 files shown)


amrbashir (changes requested):
Nice catch, I believe all window events should use `emit_filter` and target, `EventTarget::Any`, `EventTarget::Window` and `EventTarget::WebviewWindow`

vonPB:
> Nice catch, I believe all window events should use `emit_filter` and target, `EventTarget::Any`, `EventTarget::Window` and `EventTarget::WebviewWindow`

Thanks for the feedback @amrbashir, I've changed the `emit_to` to `emit_filter`.
However, I'm curious about your suggestions to use `EventTarget::Any` as well. Wouldn't `Any` already target the others inherently? This would also continue to send the close requested event to all windows...

Or did you just mean to say that all events should use emit_filter?

amrbashir:
You're right, no need to target `EventTarget::Any`, I suggested that so the event would still trigger listeners from `app.listen_any` but it will be emitted to it anyway. For this PR, `EventTarget::Window` and `EventTarget::WebviewWindow` should be enough.

vonPB:
Ah, gotcha. So even if there is no matching`Window` or `WebviewWindow` the event should still be emitted internally for `listen_any` to catch, right? https://github.com/tauri-apps/tauri/pull/8826/commits/257e95b218ca1bf9b543d644a94f1136f4614b53

amrbashir (approved):
Thank you

PR #8852, opened 2024-02-13
fix(core): Incorrect resource_dir when app run from new style target/PLATFORM+ARCH/(debug|release) dir

samwillis:
V2 changed the target dir from `/target/(debug|release)` to `target/PLATFORM+ARCH/(debug|release)`. When the app is run from the target dir, `resource_dir` fails to correctly resolve the resource directory on Mac (and maybe Linux) as the cwd doesn't match the check for the `/target/(debug|release)`.

This changes the check to support both `/target/(debug|release)` and `target/PLATFORM+ARCH/(debug|release)`


core/tauri-utils/src/platform.rs
```diff
@@ -239,9 +239,14 @@ pub fn resource_dir(package_info: &PackageInfo, env: &Env) -> crate::Result<Path
...
-    || curr_dir.ends_with(format!("{MAIN_SEPARATOR}target{MAIN_SEPARATOR}release").as_str())
-    || cfg!(target_os = "windows")
+  let parts: Vec<&str> = curr_dir.split(MAIN_SEPARATOR).collect();
+
+  if (cfg!(target_os = "windows") || parts.contains(&"target"))
+    && (parts[parts.len() - 1] == "debug" || parts[parts.len() - 1] == "release")
+    && (parts[parts.len() - 2] == "target"
+      || parts
+        .get(parts.len() - 3)
+        .map_or(false, |&dir| dir == "target"))
```
amrbashir: I believe we can simplify this a bit and we also need to check if indexes are valid

```suggestion
  let parts: Vec<&str> = curr_dir.split(MAIN_SEPARATOR).collect();
  let len = parts.len();
  if cfg!(target_os = "windows")
    || ((len >= 2 && parts[len - 2] == "target") || (len >= 3 && parts[len - 3] == "target"))
      && (parts[len - 1] == "debug" || parts[len - 1] == "release")
```

core/tauri-utils/src/platform.rs
```diff
@@ -239,9 +239,14 @@ pub fn resource_dir(package_info: &PackageInfo, env: &Env) -> crate::Result<Path
   let exe_dir = exe.parent().expect("failed to get exe directory");
   let curr_dir = exe_dir.display().to_string();
 
-  if curr_dir.ends_with(format!("{MAIN_SEPARATOR}target{MAIN_SEPARATOR}debug").as_str())
-    || curr_dir.ends_with(format!("{MAIN_SEPARATOR}target{MAIN_SEPARATOR}release").as_str())
```
amrbashir, line 243: This check is broken anyways, it doesn't take into account that cargo supports arbitray profile names, cc @lucasfernog

amrbashir (changes requested):

core/tauri-utils/src/platform.rs
samwillis: Thanks @amrbashir
I've test that here in my app
(Rust newbie here)

lucasfernog (dismissed):

core/tauri-utils/src/platform.rs
lucasfernog: yeah, pushed a change. I think checking if .cargo-lock exists should be enough

amrbashir (approved):

PR #8844, opened 2024-02-13
refactor(core): add webview events

amrbashir:

PR #8866, opened 2024-02-15
fix frontend event listeners for `EventTarget::Webview`

i-c-b:


core/tauri/src/manager/window.rs
```diff
@@ -135,7 +135,9 @@ impl<R: Runtime> Window<R> {
   fn emit_to_window<S: Serialize + Clone>(&self, event: &str, payload: S) -> crate::Result<()> {
     let window_label = self.label();
     self.emit_filter(event, payload, |target| match target {
-      EventTarget::Window { label } | EventTarget::WebviewWindow { label } => label == window_label,
+      EventTarget::Window { label }
+        | EventTarget::Webview { label }
+        | EventTarget::WebviewWindow { label } => label == window_label,
```
amrbashir, line 140: These events are specifically targeted at `Window` and `WebviewWindow`'s only, the file drop events are fixed in https://github.com/tauri-apps/tauri/pull/8844 where we introduce new webview-only events.

PR #8873, opened 2024-02-16
feat(bundler/nsis): allow including additional NSIS plugins

Pylogmon:
This PR aims to add an `additionalPlugins` configuration option for the NSIS bundler, which allowing developers to use various functionalities provided by NSIS Plugins in the NSIS template. I apologize for not being very familiar with the tauri project structure, if there are any omissions, please kindly point them out.

The result I expect is to configure something like the following in `tauri.conf.json`, and then tauri can copy the dll file written in `additionalPlugins` to the nsis plugin directory before packaging nsis.

```json
{
  "tauri": {
    "bundle": {
      "windows": {
        "nsis": {
          "template": "./template/installer.nsi",
          "additionalPlugins": ["plugins/SimpleSC.dll"]
        }
      }
    }
  }
}
```

### DIFF
--- a/.changes/nsis-additional-plugins.md
+++ b/.changes/nsis-additional-plugins.md
@@ -1,0 +1,8 @@
+---
+'tauri-utils': 'minor'
+'tauri-bundler': 'minor'
+'cli.rs': 'minor'
+'cli.js': 'minor'
+---
+
+Add `nsis > additionalPlugins` option to include additional NSIS plugins.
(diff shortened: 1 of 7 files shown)


Pylogmon:
@amrbashir Would you take a look at this?

amrbashir:
if you're using a custom NSIS template, you can just include additional plugins in your template, without having to add this option

Pylogmon:
Oh, I'm sorry, I don't know enough about NSIS, You are right.

PR #8876, opened 2024-02-16
docs: Fix PathResolver doc comment

FabianLars:

### DIFF
--- a/core/tauri/src/path/desktop.rs
+++ b/core/tauri/src/path/desktop.rs
@@ -6,7 +6,7 @@ use super::{Error, Result};
 use crate::{AppHandle, Manager, Runtime};
 use std::path::PathBuf;
 
-/// A helper class to access the mobile camera APIs.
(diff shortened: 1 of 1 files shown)


core/tauri/src/path/desktop.rs
```diff
@@ -6,7 +6,7 @@ use super::{Error, Result};
 use crate::{AppHandle, Manager, Runtime};
 use std::path::PathBuf;
 
-/// A helper class to access the mobile camera APIs.
+/// The path resolver is a helper class for general and application-specific path APIs.
```
FabianLars, line 9:
```suggestion
/// The path resolver is a helper class to get general and application-specific path APIs.
```
or is that better?

lucasfernog: `for` here makes more sense with the ending `path APIs` IMO

lucasfernog (approved):

Issue #8476, opened 2023-12-27
[bug] Tauri V2: CSP issues on external window URLs


Glitch752:
### Describe the bug

Hi! I'm using Tauri v2 and have a window with an external URL. This worked fine with injecting scripts and calling to rust when I was previously on v1, but I recently switched to v2. Now, I get content security policy errors like the following when using any window.__TAURI__ API; for example:
```
Refused to connect to 'http://ipc.localhost/test' because it violates the document's Content Security Policy.
```

I'm sorry if I'm overlooking something. Thank you!

### Reproduction

- Create a new Tauri v2 project
- Add the following code to `main.rs`:
```rust
// Prevents additional console window on Windows in release, DO NOT REMOVE!!
#![cfg_attr(not(debug_assertions), windows_subsystem = "windows")]

#[tauri::command]
fn test() {
    println!("Called from Rust!");
}

fn main() {
    tauri::Builder::default()
        .setup(|app| {
            let window = tauri::WindowBuilder::new(app, "main", tauri::WindowUrl::External("https://twitter.com/TauriApps".parse().unwrap()))
                .initialization_script(r#"
                document.addEventListener('DOMContentLoaded', () => {
                    window.__TAURI__.core.invoke('test');
                });
                "#)
                .build()
                .expect("failed to create window");
            window.open_devtools();
            Ok(())
        })
        .invoke_handler(tauri::generate_handler![test])
        .run(tauri::generate_context!())
        .expect("error while running tauri application");
}
```
- In the console, you should see `Refused to connect to 'http://ipc.localhost/test' because it violates the document's Content Security Policy.`

### Expected behavior

The Tauri APIs should ideally work regardless of the site's content security policy, but I understand that's likely not an option. This setup worked in Tauri v1, so something is different. From what I can tell, Tauri used `window.ipc.postMessage` but now uses these custom protocols on Windows and MacOS; I assume it's not as easy as overriding `ipc_custom_protocol` to false, but doing that along with overriding the check in `ipc-protocol.js` seems to work in most situations.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.19045 X64
    ✔ WebView2: 120.0.2210.91
    ✔ MSVC: Visual Studio Community 2022
    ✔ rustc: 1.73.0-nightly (ad963232d 2023-07-14)
    ✔ cargo: 1.73.0-nightly (694a57956 2023-07-11)
[...]

PR #8887, opened 2024-02-18
add option to force v1 postMessage ipc protocol

thewh1teagle:
It's important to have the ability to control whether to use the old IPC protocol or the newer one. For instance, because the new method uses fetch, we have issues like https://github.com/tauri-apps/tauri/issues/8476.

So, I added the ability to force using the v1 `postMessage` `IPC` protocol by activating the `force-ipc-v1-protocol` feature in the Tauri dependency.

I tested it on `MacOS` / `Windows` both when activated and not, and it works.
Also it solved for me the `CSP` issues in windows.

https://discord.com/channels/616186924390023171/1201542436040757248

## Update
Looks like it doesn't work on `Windows` when returning JSON value from rust, checking it. (it works with strings)

## Update1 
Solved in second commit. need review

### DIFF
--- a/core/tauri/Cargo.toml
+++ b/core/tauri/Cargo.toml
@@ -128,6 +128,7 @@ http-range = "0.1.5"
 [features]
 default = [ "wry", "compression", "objc-exception", "common-controls-v6" ]
 unstable = [ ]
+force-ipc-v1-protocol = [ ]
 common-controls-v6 = [ "tray-icon?/common-controls-v6", "muda/common-controls-v6" ]
 tray-icon = [ "dep:tray-icon" ]
 tracing = [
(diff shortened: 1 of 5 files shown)


lucasfernog:
I think this should be resolved at runtime instead, so you can have both the v1 and the v2 IPC implementations if you app use both local and remote URLs.

thewh1teagle:
@lucasfernog

I think it might be much better to handle it at runtime / per webview.
Do you have any ideas for a convenient interface or examples that we could follow?

lucasfernog:
After investigating this I believe it's easier to handle the CSP issue at the IPC layer instead of letting the user to configure it (with is a lot more complicated). I'll open a PR soon. 

Thank you for this PR though, you've done a great job investigating the problem and finding a solution!

PR #8932, opened 2024-02-21
chore(deps): update to tao 0.26

lucasfernog:

### DIFF
--- a/core/tauri-utils/src/lib.rs
+++ b/core/tauri-utils/src/lib.rs
@@ -447,6 +447,6 @@ pub struct ProgressBarState {
   pub status: Option<ProgressBarStatus>,
   /// The progress bar progress. This can be a value ranging from `0` to `100`
   pub progress: Option<u64>,
-  /// The identifier for your app to communicate with the Unity desktop window manager **Linux Only**
-  pub unity_uri: Option<String>,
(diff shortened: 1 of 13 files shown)


core/tauri-utils/src/lib.rs
```diff
@@ -447,6 +447,6 @@ pub struct ProgressBarState {
   pub status: Option<ProgressBarStatus>,
   /// The progress bar progress. This can be a value ranging from `0` to `100`
   pub progress: Option<u64>,
-  /// The identifier for your app to communicate with the Unity desktop window manager **Linux Only**
-  pub unity_uri: Option<String>,
+  /// The `.desktop` filename with the Unity desktop window manager, for example `myapp.desktop` **Linux Only**
+  pub desktop_filename: Option<String>,
```
amrbashir: Shouldn't we move these struct and related enum out of utils crate? Not sure why they are there in the first place.

lucasfernog: it's needed in tauri-runtime and tauri-runtime-wry :(

lucasfernog: ah yeah it can live in the runtime now

amrbashir (approved):

PR #8939, opened 2024-02-21
feat(webview): add reparent API

lucasfernog:
Adds a new API that can be used to move a child webview to another window.


core/tauri/src/webview/plugin.rs
```diff
@@ -159,6 +159,20 @@ mod desktop_commands {
   setter!(set_webview_position, set_position, Position);
   setter!(set_webview_focus, set_focus);
 
+  #[cfg(any(debug_assertions, feature = "devtools"))]
```
amrbashir: any reason why this is only enabled on these features

lucasfernog: copypasta lol

amrbashir (approved):

PR #8942, opened 2024-02-22
fix: center window ignores scale factors

vonPB:


core/tauri-runtime-wry/src/lib.rs
```diff
@@ -3282,8 +3282,9 @@ fn create_window<T: UserEvent, F: Fn(RawWindow) + Send + 'static>(
         .inner_size_constraints
         .clamp(desired_size, monitor.scale_factor())
         .to_logical::<i32>(monitor.scale_factor());
-      let screen_size = monitor.size();
-      let monitor_pos = monitor.position();
+      let scale_factor = monitor.scale_factor();
```
amrbashir: scale_factor is also used in `desired_size` and `window_size` variables above, let's move these variables above them and use it there

amrbashir (dismissed):
Thank you

amrbashir (approved):

PR #8948, opened 2024-02-22
fix(bundler): remove fallback for license_file

### LINKED ISSUE #8944: [bug] [NSIS] License agreement screen cannot be omitted
### Describe the bug

There seems to be a change from v2-alpha to v2-beta in the NSIS bundler that causes the license agreement screen to be shown (with empty license text field) even if there's no `license` or `licenseFile` file set or set to `null` in [BundleConfig](https://beta.tauri.app/references/v2/config/#bundleconfig). I don't see any configurable option in [NsisConfig](https://beta.tauri.app/references/v2/config/#nsisconfig) either.

In our case we don't want to show the End-user license agreement (EULA) screen as it's a German software product and showing the user the agreement after he bought the product has no legal effect, see [Endbenutzer-Lizenzvertrag](https://de.wikipedia.org/wiki/Endbenutzer-Lizenzvertrag).

Translated quote:
> License terms that are only made available to the buyer after purchase (for example **during installation** or as a printed insert in the packaging) are ineffective for the buyer.

### Reproduction

Create a new Tauri v2-beta app and configure a NSIS bundle to be built. Execute the resulting NSIS installer and advance to second screen. License agreement screen is shown with empty license text.

### Expected behavior

If no license and/or license file is set, don't show the license agreement screen. Or give an option in NsisConfig to omit the screen explicitly.

### Full `tauri info` output

```text
$ yarn tauri info
yarn run v1.22.21
$ tauri info

[✔] Environment
    - OS: Windows 10.0.17763 X64
    ✔ WebView2: 121.0.2277.128
    ✔ Rust toolchain: stable-i686-pc-windows-msvc (default)
    - node: 20.10.0
    - pnpm: 8.15.2
    - yarn: 1.22.21
    - npm: 10.2.3

[-] Packages
    - tauri [RUST]: 2.0.0-beta.2
    - tauri-build [RUST]: 2.0.0-beta.1
    - wry [RUST]: 0.35.2
    - tao [RUST]: 0.25.0
    - @tauri-apps/api [NPM]: 2.0.0-beta.2
    - @tauri-apps/cli [NPM]: 2.0.0-beta.1

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - framework: Vue.js
    - bundler: Vite
Done in 10.53s.
```

### Stack trace

_No response_

### Additional context

_No response_

amrbashir:
closes #8944

### DIFF
--- a/.changes/bundler-license.md
+++ b/.changes/bundler-license.md
@@ -1,5 +1,5 @@
 ---
-'tauri-bundler': 'patch'
+'tauri-bundler': 'patch:bug'
 ---
 
 Fix NSIS installer always containing a license page even though `licenseFile` option is not set in the config.
--- a/.changes/bundler-license.md
+++ b/.changes/bundler-license.md
@@ -1,0 +1,5 @@
+---
+'tauri-bundler': 'patch'
+---
+
+Fix NSIS installer always containing a license page even though `licenseFile` option is not set in the config.
--- a/.changes/bundler-rpm-license.md
+++ b/.changes/bundler-rpm-license.md
@@ -2,4 +2,4 @@
 "tauri-bundler": patch:bug
 ---
 
-Don't fallback to `liceneseFile` and use only `license` field when building RPM.
+Don't fallback to `licenseFile` and use only `license` field when building RPM.
--- a/.changes/bundler-rpm-license.md
+++ b/.changes/bundler-rpm-license.md
@@ -1,0 +1,5 @@
+---
+"tauri-bundler": patch:bug
+---
+
+Don't fallback to `liceneseFile` and use only `license` field when building RPM.
--- a/tooling/bundler/src/bundle/linux/rpm.rs
+++ b/tooling/bundler/src/bundle/linux/rpm.rs
@@ -47,7 +47,8 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
 
   let license = settings
     .license_file()
-    .map(|l| std::fs::read_to_string(l).expect("failed to read license"))
+    .and_then(|l| std::fs::read_to_string(l).ok())
+    .or_else(|| settings.license())
     .unwrap_or_default();
   let mut builder = rpm::PackageBuilder::new(name, version, &license, arch, summary)
     .epoch(epoch)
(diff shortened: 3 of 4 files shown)


.changes/bundler-license.md
```diff
@@ -0,0 +1,5 @@
+---
+'tauri-bundler': 'patch'
```
amrbashir:
```suggestion
'tauri-bundler': 'patch:bug'
```

tooling/bundler/src/bundle/linux/rpm.rs
```diff
@@ -47,7 +47,8 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
 
   let license = settings
     .license_file()
```
olivierlemasle: This should be `license`, not `license_file`

olivierlemasle (changes requested):
I was actually doing the same pull request just now 😄

Or actually, I was changing the RPM bundle to use the restored `license()` function instead of `license_file()`, as RPM `License` tag is meant to contain a license identifier (e.g SPDX identifier), not a license file content.

It might also be good to add this RPM change in the changelog too.

tooling/bundler/src/bundle/linux/rpm.rs
amrbashir: nice catch

olivierlemasle (approved):
👍

Just add perhaps the RPM bug fix in the changelog (since the implementation of the new settings structure, the License tag of RPM bundles included the whole time when `license_file` is defined)

lucasfernog (dismissed):
we should document this behavior in the future

amrbashir:
> I was actually doing the same pull request just now 😄

That's unfortunate, in the future could you reach out in discord or the issue and I will assign you? just so we don't work on the same thing. We appreciate your help <3.

amrbashir:
It is partially my fault too, I will also start assigning myself before working on something

.changes/bundler-rpm-license.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri-bundler": patch:bug
+---
+
+Don't fallback to `liceneseFile` and use only `license` field when building RPM.
```
lucasfernog:
```suggestion
Don't fallback to `licenseFile` and use only `license` field when building RPM.
```

lucasfernog (approved):

PR #8950, opened 2024-02-22
refactor(core): allow configuring both local and remote URLs on capability

lucasfernog:
The current capability format allows you to configure **either** a capability that applies to local URLs **or** to a list of remote URLs. This forces users to have to define two identical capability files if they choose to enable a set of permissions both locally and to remote URLs.
This PR changes the model from `context: local | { remote: { urls: string[] } }` to `remote: { urls: string[] }, local: bool`

### DIFF
--- a/.changes/capability-context-refactor.md
+++ b/.changes/capability-context-refactor.md
@@ -1,0 +1,7 @@
+---
+"tauri-utils": patch:breaking
+"tauri-cli": patch:breaking
+"@tauri-apps/cli": patch:breaking
+---
+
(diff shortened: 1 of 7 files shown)


amrbashir (approved):

amrbashir:
Could we maybe change it to
```json
{
  "context": ["local", { "remote": ["https://url.com"] }]
}
```

Seems like it would allow for future additions if needed

lucasfernog:
that was my first idea, but that also allows to weird configurations like `[local, { remote }, local]` and it looks ugly on json/toml formats.

PR #8971, opened 2024-02-24
fix(docs): fix API webview/window class doc example

i-c-b:
Does this need a `.changes` file?


FabianLars:
so far we've not added changefiles for pure docs changes

FabianLars (approved):

amrbashir (approved):

PR #8988, opened 2024-02-26
feat(core) enhance `.plugin()` inside the app `Builder`

p1mo:
adding os and device specific wrappers for .plugin()

### DIFF
--- a/core/tauri/Cargo.toml
+++ b/core/tauri/Cargo.toml
@@ -159,6 +159,7 @@ config-toml = [ "tauri-macros/config-toml" ]
 icon-ico = [ "infer", "ico" ]
 icon-png = [ "infer", "png" ]
 macos-proxy = [ "tauri-runtime-wry/macos-proxy" ]
+os-plugins = [ ]
 
 [[example]]
 name = "commands"
--- a/core/tauri/Cargo.toml
+++ b/core/tauri/Cargo.toml
@@ -159,7 +159,6 @@ config-toml = [ "tauri-macros/config-toml" ]
 icon-ico = [ "infer", "ico" ]
 icon-png = [ "infer", "png" ]
 macos-proxy = [ "tauri-runtime-wry/macos-proxy" ]
-os-plugins = [ ]
 
 [[example]]
 name = "commands"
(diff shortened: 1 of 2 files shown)


core/tauri/Cargo.toml
```diff
@@ -159,6 +159,7 @@ config-toml = [ "tauri-macros/config-toml" ]
 icon-ico = [ "infer", "ico" ]
 icon-png = [ "infer", "png" ]
 macos-proxy = [ "tauri-runtime-wry/macos-proxy" ]
+os-plugins = [ ]
```
amrbashir: there is no need for a specific feature for it, we can just add the helper methods without it directly

core/tauri/src/app.rs
```diff
@@ -1753,6 +1753,76 @@ tauri::Builder::default()
   }
 }
 
+// enhance core `.plugin()` for mobile and desktop
+impl<R: Runtime> Builder<R> {
+  /// same behavior like `.plugin()` but skips on non desktop builds and ignores `target_os`
```
amrbashir: Let's change all these docs to start with capital
```suggestion
  /// Same behavior like `.plugin()` but skips on non desktop builds and ignores `target_os`
```

amrbashir (changes requested):

lucasfernog:
this doesn't help much.. target-specific plugins usually cannot be compiled on other platforms

instead what we could do is have a `plugin_maybe` function that takes an `Option<Plugin>` so you can use it like this:
```rust
tauri::Builder::default()
  .plugin_maybe(
    #[cfg(mobile)] Some(tauri_plugin_nfc::init()),
    #[cfg(not(mobile))] None,
  )
```

amrbashir:
I believe they all compile just fine, but fail when an API is used

p1mo:
my apps are compiling fine with that ... but yeah when you use the api it errors but thats on the dev ...

i just made the fn's to solve my little problems ... and maybe someone needs these kind of wrappers too.

but idk doing it like that ?

```rust
  .plugin_maybe(
      #[cfg(mobile)] Some(tauri_plugin_nfc::init()),
      #[cfg(not(mobile))] None,
    )
```

a lot of plugins and a lot of None's ^^

amrbashir:
I believe both can be added with a note about runtime crashes and recommending usage of plugin_maybe instead

PR #9008, opened 2024-02-27
feat(core): allow defining permissions for the app commands

lucasfernog:
Current ACL only applies to plugin commands. This PR adds a mechanism to autogenerate permissions for the app commands - though it needs to be manually provided by the user :( - and also pull permission from a directory, similar to how plugin permissions work.

By default all app commands are still allowed, so if you do not use `tauri_build::Attributes::app_manifest` you're not affected by this change.


core/tauri-build/src/acl.rs
```diff
@@ -28,43 +29,118 @@ const CAPABILITIES_SCHEMA_FILE_NAME: &str = "schema.json";
...
+  /// Define a list of commands that gets permissions autogenerated in the format of `allow-$command` and `deny-$command`
+  /// where $command is the command name in snake_case.
+  pub fn commands(mut self, commands: &'static [&'static str]) -> Self {
+    self.commands = commands;
+    self
+  }
+
+  /// Sets a glob pattern that is used to find the permissions of the app.
+  ///
+  /// **Note:** You must emit [rerun-if-changed] instructions for the permissions directory.
```
amrbashir, line 96: can't we do that internally?

core/tauri-build/src/lib.rs
```diff
@@ -514,54 +490,21 @@ pub fn try_build(attributes: Attributes) -> Result<()> {
...
-        &plugin_out_dir,
-      )?);
-    }
 
-    let manifest = tauri_utils::acl::plugin::Manifest::new(permission_files, None);
-    plugin_manifests.insert(name.into(), manifest);
-  }
+  let mut acl_manifests = acl::get_manifests_from_plugins()?;
+  acl_manifests.insert(
+    "".into(),
```
amrbashir: this seems sort of magical key and can be easily forgot about? we should add an explicit field or marker for it

core/tauri-build/src/acl.rs
lucasfernog: if you set a custom glob pattern it's hard to determine which path should be used to rerun-if-changed

core/tauri-build/src/lib.rs
lucasfernog: you're right, pushed

core/tauri-build/src/acl.rs
amrbashir: makes sense

amrbashir (approved):

PR #9009, opened 2024-02-27
fix(cli): CLI path issues on mobile project initialization

amrbashir:
closes #8531

PR #9015, opened 2024-02-28
fix(cli): Truncate BuildTask.kt before writing

i-c-b:


i-c-b:
I also noticed the same code for [`tauri ios init`](<https://github.com/tauri-apps/tauri/blob/719093568073a5649bc9157b25a61b506bf75591/tooling/cli/src/mobile/ios/project.rs#L158>) and [`tauri plugin init --android` / `tauri plugin android init`](<https://github.com/tauri-apps/tauri/blob/719093568073a5649bc9157b25a61b506bf75591/tooling/cli/src/plugin/init.rs#L277>) but neither seems to have BuildTask.kt in their template files. Should I remove their checks for the file?

amrbashir:
I think Lucas fixed the android part in #9009 , feel free to just truncate on other platforms.

lucasfernog:
can't we always truncate?
EDIT: ahh we didn't want to mess with existing files.. let's go with this then

lucasfernog (approved):

PR #9036, opened 2024-02-29
feat(core): dynamic runtime capability

lucasfernog:
Some capabilities must be dynamic (e.g. allow a dynamically started localhost server, runtime-defined remote URLs, fine-grained permissions that depend on user configuration like access token scopes..). This can be done already by creating a string at runtime and leaking it so it becomes a &'static str, but it's not ideal so we now support both runtime strings and added a new capability builder aswell so you can either `include_str!()` a static capability file or create/modify one at runtime.

### DIFF
--- a/core/tauri/src/ipc/authority.rs
+++ b/core/tauri/src/ipc/authority.rs
@@ -164,34 +164,31 @@ impl CapabilityBuilder {
       .try_into()
       .unwrap_or_else(|_| panic!("invalid permission identifier '{permission}'"));
 
+    let allowed_scope = allowed
+      .into_iter()
+      .map(|a| {
+        serde_json::to_value(a)
+          .expect("failed to serialize scope")
+          .into()
+      })
+      .collect();
+    let denied_scope = denied
+      .into_iter()
+      .map(|a| {
+        serde_json::to_value(a)
+          .expect("failed to serialize scope")
+          .into()
+      })
+      .collect();
+    let scope = Scopes {
+      allow: Some(allowed_scope),
+      deny: Some(denied_scope),
+    };
+
     self
       .0
       .permissions
-      .push(PermissionEntry::ExtendedPermission {
-        identifier,
-        scope: Scopes {
-          allow: Some(
-            allowed
-              .into_iter()
-              .map(|a| {
-                serde_json::to_value(a)
-                  .expect("failed to serialize scope")
-                  .into()
-              })
-              .collect(),
-          ),
-          deny: Some(
-            denied
-              .into_iter()
-              .map(|a| {
-                serde_json::to_value(a)
-                  .expect("failed to serialize scope")
-                  .into()
-              })
-              .collect(),
-          ),
-        },
-      });
+      .push(PermissionEntry::ExtendedPermission { identifier, scope });
     self
   }
 }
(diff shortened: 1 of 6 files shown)


core/tauri/src/ipc/authority.rs
```diff
@@ -62,6 +65,128 @@ impl Origin {
...
+      .get_or_insert_with(Default::default)
+      .urls
+      .push(url);
+    self
+  }
+
+  /// Do not apply this capability on local app URLs.
+  pub fn skip_local(mut self) -> Self {
+    self.0.local = false;
+    self
```
amrbashir: Probably better to just call it local and take a bool?
```suggestion
  pub fn local(mut self, local: bool) -> Self {
    self.0.local = local;
    self
```

core/tauri/src/ipc/authority.rs
```diff
@@ -62,6 +65,128 @@ impl Origin {
...
+  pub fn window(mut self, window: impl Into<String>) -> Self {
+    self.0.windows.push(window.into());
+    self
+  }
+
+  /// Link this capability to the given webview label.
+  pub fn webview(mut self, webview: impl Into<String>) -> Self {
+    self.0.webviews.push(webview.into());
+    self
+  }
```
amrbashir: would it be worth to also add iterator version of these two APIs as well?

core/tauri/src/ipc/authority.rs
```diff
@@ -62,6 +65,128 @@ impl Origin {
...
+    mut self,
+    permission: impl Into<String>,
+    allowed: Vec<T>,
+    denied: Vec<T>,
+  ) -> Self {
+    let permission = permission.into();
+    let identifier = permission
+      .clone()
+      .try_into()
+      .unwrap_or_else(|_| panic!("invalid permission identifier '{permission}'"));
```
amrbashir, line 165: I would try to avoid panicking and return a result instead

core/tauri/src/ipc/authority.rs
```diff
@@ -62,6 +65,128 @@ impl Origin {
...
+      .permissions
+      .push(PermissionEntry::ExtendedPermission {
+        identifier,
+        scope: Scopes {
+          allow: Some(
+            allowed
+              .into_iter()
+              .map(|a| {
+                serde_json::to_value(a)
+                  .expect("failed to serialize scope")
```
amrbashir: ditto

core/tauri/src/ipc/authority.rs
```diff
@@ -62,6 +65,128 @@ impl Origin {
...
+                  .into()
+              })
+              .collect(),
+          ),
+          deny: Some(
+            denied
+              .into_iter()
+              .map(|a| {
+                serde_json::to_value(a)
+                  .expect("failed to serialize scope")
```
amrbashir: ditto

core/tauri/src/ipc/authority.rs
```diff
@@ -62,6 +65,128 @@ impl Origin {
...
+                  .into()
+              })
+              .collect(),
+          ),
+        },
+      });
+    self
```
amrbashir: let's spread these onto multiple lines to avoid crazy indentation

amrbashir (changes requested):

lucasfernog:
@amrbashir it's quite rare for those functions to panic (most identifiers are valid, and it's also almost impossible for serde_json::to_value to fail) and I don't really like the pattern of returning Result in a builder
plus we're also panicking on add_capability itself which I think is ok since if you mess up in the capability tou shouldn't ship your app at all

amrbashir:
That's fair I guess

amrbashir (approved):

PR #7764, opened 2023-09-06
fix(core): use postMessage IPC for remote URLs on iOS ref #7751

lucasfernog:
### What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

### Does this PR introduce a breaking change?

- [ ] Yes, and the changes were approved in issue #___
- [ ] No

### Checklist
- [ ] When resolving issues, they are referenced in the PR's title (e.g `fix: remove a typo, closes #___, #___`)
- [x] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md).
- [ ] I have added a convincing reason for adding this feature, if necessary

PR #9038, opened 2024-02-29
fix(ipc): fallback to postMessage if protocol fails, closes #8476

### LINKED ISSUE #8476: [bug] Tauri V2: CSP issues on external window URLs

lucasfernog:
This PR changes the IPC mechanism to always support the custom-protocol and postMessage variants. The frontend now handles the custom protocol fetch call failing (usually due to CSP or webview denying access to the protocol) and fallbacks to the postMessage API).

The fixes for #7662 (#7751, #7754 and #7764) are still valid.

This is a more generic approach to #8887

### DIFF
--- a/.changes/ipc-post-message-fallback.md
+++ b/.changes/ipc-post-message-fallback.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch:enhance
+---
+
+Fallback to the postMessage IPC interface if we cannot reach the IPC custom protocol.
--- a/core/tauri/build.rs
+++ b/core/tauri/build.rs
@@ -217,11 +217,6 @@ fn main() {
   alias("desktop", !mobile);
   alias("mobile", mobile);
 
-  alias(
-    "ipc_custom_protocol",
-    target_os != "android" && (target_os != "linux" || has_feature("linux-ipc-protocol")),
-  );
-
   let out_dir = PathBuf::from(var("OUT_DIR").unwrap());
 
   let checked_features_out_path = out_dir.join("checked_features");
--- a/core/tauri/scripts/ipc-protocol.js
+++ b/core/tauri/scripts/ipc-protocol.js
@@ -6,72 +6,73 @@
   const processIpcMessage = __RAW_process_ipc_message_fn__
   const osName = __TEMPLATE_os_name__
   const fetchChannelDataCommand = __TEMPLATE_fetch_channel_data_command__
-  const useCustomProtocol = __TEMPLATE_use_custom_protocol__
+  const linuxIpcProtocolEnabled = __TEMPLATE_linux_ipc_protocol_enabled__
+  let customProtocolIpcFailed = false
 
-  Object.defineProperty(window.__TAURI_INTERNALS__, 'postMessage', {
-    value: (message) => {
-      const { cmd, callback, error, payload, options } = message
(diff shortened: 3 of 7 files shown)


amrbashir (approved):

thewh1teagle:
@lucasfernog 
Looks like the IPC fallback doesn't work in case of CSP errors when plugins used (eg. event.listen).
Is there a way to control it manually and force use of old IPC only? 

Also I when it fallback to postMessage if I return any object value (eg serde_json::Value) the promise doesn't resolve on webview2

PR #9042, opened 2024-03-01
feat(api): add `Image` class

amrbashir:


lucasfernog (dismissed):

lucasfernog (dismissed):

lucasfernog:
@amrbashir let me know if you agree with my changes :P

core/tauri/src/image/plugin.rs
```diff
@@ -5,46 +5,41 @@
...
-  let rid = resources_table.add(image);
-  Ok(rid)
-}
-#[command(root = "crate")]
-fn from_ico_bytes<R: Runtime>(app: AppHandle<R>, image: JsImage<'_>) -> crate::Result<ResourceId> {
-  let image: Image<'_> = image.try_into()?;
-  let image = image.to_owned();
-  let mut resources_table = app.resources_table();
-  let rid = resources_table.add(image);
+  let rid = resources_table.add(Arc::try_unwrap(image).unwrap().to_owned());
```
amrbashir: I guess we can just use `Image::new(&rgba, width, height).to_owned()` instead or wrapping in an Arc and unwraping

core/tauri/src/image/plugin.rs
```diff
@@ -5,46 +5,41 @@
...
   Ok(rid)
 }
+
 #[command(root = "crate")]
-fn from_bytes<R: Runtime>(app: AppHandle<R>, image: JsImage<'_>) -> crate::Result<ResourceId> {
-  let image: Image<'_> = image.try_into()?;
-  let image = image.to_owned();
+fn from_bytes<R: Runtime>(app: AppHandle<R>, bytes: Vec<u8>) -> crate::Result<ResourceId> {
+  let image = JsImage::Bytes(&bytes).into_img(&app)?;
```
amrbashir: ditto

core/tauri/src/image/plugin.rs
```diff
@@ -5,46 +5,41 @@
...
+  let rid = resources_table.add(Arc::try_unwrap(image).unwrap().to_owned());
   Ok(rid)
 }
+
 #[command(root = "crate")]
-fn from_path<R: Runtime>(app: AppHandle<R>, image: JsImage<'_>) -> crate::Result<ResourceId> {
-  let image: Image<'_> = image.try_into()?;
-  let image = image.to_owned();
+fn from_path<R: Runtime>(app: AppHandle<R>, path: std::path::PathBuf) -> crate::Result<ResourceId> {
+  let image = JsImage::Path(path).into_img(&app)?.to_owned();
```
amrbashir: ditto

amrbashir (commented):
Any reason to remove `fromIcoBytes` and `fromPngBytes`?

lucasfernog:
they're the same as fromBytes, we could create aliases on the JS side in case they differ in the future, but for now i don't see a reason to have three separate commands specially since if you allow the from_bytes one you can "bypass" the allowlist pretty much

amrbashir:
> they're the same as fromBytes, we could create aliases on the JS side in case they differ in the future

Well that was just a mistake in my implementation, they should be different as `fromBytes` will use infer to determine the bytes and fromIcoBytes and fromPngBytes should not.

lucasfernog (approved):

PR #9049, opened 2024-03-01
fix(cli): migration for http plugin ACL

i-c-b:
`tauri migrate` creates an array of strings but http plugin expects an array of objects with a key of `url` and a string value.

### DIFF
--- a/.changes/fix-cli-migration-http-acl.md
+++ b/.changes/fix-cli-migration-http-acl.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch:bug
+---
+
+Fix `tauri migrate` for http plugin ACL.
--- a/tooling/cli/src/migrate/config.rs
+++ b/tooling/cli/src/migrate/config.rs
@@ -14,7 +14,7 @@ use tauri_utils::{
 };
 
 use std::{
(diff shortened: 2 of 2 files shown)


amrbashir:
seems like the better change here would be to change the http plugin scope to take array of url patterns instead of objects, cc @lucasfernog

lucasfernog:
IIRC I kept the object way of defining scopes because it looked better on TOML files.. if you guys wanna double check it's fine

lucasfernog (approved):

amrbashir:
Just checked this and there is a few things to improve:
1. ~~permissions in a capability file using toml need to use `[[permissions]]` synatx, you can't `permissions = ["allow"]`~~ Edit: this isn't true, only the generated schema doesn't match this variant
2. the scope in toml will need to be defined as `allow = [{ url = "hello" }]` which is not very nice

Here is an example of the whole scope file:
```toml
"$schema" = "../gen/schemas/desktop-schema.json"
identifier = "run-app-base"
description = "Base permissions to run the app"
windows = ["main"]

[[permissions]]
identifier = "http:default"
allow = [{ url = "hello" }]
```

PR #9059, opened 2024-03-03
feat(core): always use child webviews on the unstable feature

lucasfernog:
This PR changes the WebviewWindow implementation to always use wry's child webview implementation when the `unstable` feature flag is enabled. This enhances flexibility for app developers, allowing a WebviewWindow to be transformed to a multiwebview window and vice versa, reparent a webview from a WebviewWindow to another window etc.


core/tauri-runtime-wry/src/lib.rs
```diff
@@ -3427,6 +3449,9 @@ fn create_window<T: UserEvent, F: Fn(RawWindow) + Send + 'static>(
 
   if let Some(webview) = webview {
     webviews.push(create_webview(
+      #[cfg(feature = "unstable")]
+      WebviewKind::WindowChild,
```
amrbashir, line 3453: Why do we want to use `WindowChild` for `WebviewWindow` ? I think we still has issues with that on macOS

lucasfernog: it's explained in the PR description, note this is all behind the unstable feature flag

amrbashir: yeah but in the future we plan to remove this `unstable` feature, no?

lucasfernog: yeah when we sort all the bugs and stabilize the API (that's why i mentioned the flag)

amrbashir: cool

amrbashir (approved):

PR #9068, opened 2024-03-04
fix(acl): scope resolution should be per window

lucasfernog:
The ACL resolution has a bug where it is grouping all scopes per command and execution context (the allowed_commands map key) and ignoring the capability window. This PR changes the resolution to only use the command name as map key, and use a vector to collect the resolved permissions (windows/webviews/context/scope).

This comes with a small breaking change because the scope is now resolved at runtime instead of being merged at compile time. Parsing via the ScopeObject trait is still only done once for each block of scope, but we now push copies of each scope entry (now stored as Arc<T>) to the returned scope value. Previously an Arc<Vec<T>> was used which isn't enough to copy the scopes around now that they are isolated to be referenced per window/webview.


core/tauri-utils/src/acl/resolved.rs
```diff
@@ -214,45 +182,17 @@ impl Resolved {
...
     };
 
     Ok(resolved)
   }
 }
 
-fn parse_glob_patterns(raw: HashSet<String>) -> Result<Vec<glob::Pattern>, Error> {
+fn parse_glob_patterns(raw: Vec<String>) -> Result<Vec<glob::Pattern>, Error> {
   let mut raw = raw.into_iter().collect::<Vec<_>>();
   raw.sort();
```
amrbashir: Unnecessary iterator
```suggestion
fn parse_glob_patterns(mut raw: Vec<String>) -> Result<Vec<glob::Pattern>, Error> {
  raw.sort();
```

amrbashir (approved):

PR #9069, opened 2024-03-04
chore: remove unstable log features & use qualified `log::<macro>` path

amrbashir:

### DIFF
--- a/.changes/utils-debug-eprintln.md
+++ b/.changes/utils-debug-eprintln.md
@@ -1,0 +1,5 @@
+---
+'tauri-utils': 'major:breaking'
+---
+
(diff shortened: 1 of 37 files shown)


lucasfernog (approved):
we should consider adding tauri-plugin-log by default (on debug) on the app template.. it's now required to see the log::error messages (plus it's kinda required on mobile, otherwise app logs/stdout aren't visible on android and iOS)

amrbashir:
maybe we could add an internal logger that is activated on debug only. I don't want to pollute the template with another plugin, I am already dissatisfied with the shell plugin being there by default.

lucasfernog:
shell is there only for CTA right? and yes it sucks

amrbashir:
yeah only in CTA template

PR #9079, opened 2024-03-05
feat: enhance multiple capabilities file format & fix mixed permissions schema

amrbashir:
This enhances the DX of defining multiple definitions in a single JSON file, previously the only way was:
```json
{
  "capabilities": [ {...}, {...} ] 
}
```
now we can also support 
```json
[
  {...},
  {...}
]
```

toml however can't take advantage of this due to the nature of toml but doesn't matter that much
```toml
[[capabilities]]
identifier = "id"

[[capabilities]]
identifier = "another-one"
```

### DIFF
--- a/.changes/build-schema-generation.md
+++ b/.changes/build-schema-generation.md
@@ -1,0 +1,5 @@
+---
+"tauri-build": patch:bug
+---
+
+Fixed generation of capability schema for permissions field which previously disallowed mixed (strings and objects) permission definition.
--- a/.changes/utils-named-capability-file.md
+++ b/.changes/utils-named-capability-file.md
@@ -1,0 +1,5 @@
+---
+"tauri-utils": major:breaking
+---
+
+Changed `CapabiltyFile::List` enum variant to be a tuple-struct and added `CapabiltyFile::NamedList`. This allows more flexibility when parsing capabilties from JSON files. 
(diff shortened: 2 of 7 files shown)


amrbashir:
This PR also fixes generation of `PermissionEntry` which made the generated schema not accept the following valid permission array:
```json
{
  "permissions": ["perm1", { "identifier": "perm2" }]
}
```

Even though mixed permission array are parsed correctly in toml, the `Event Better TOML` extension in vscode still flags the following tomls as incorrect:
```toml
identifier = "allow-http"
windows = ["main"]

[[permissions]]
identifier = "allow-read"
```
and
```toml
identifier = "allow-http"
windows = ["main"]
permissions = ["app:allow-app-hide", { identifier = "allow-read" }]
```

This maybe a bug in `Event Better TOML` extension but I am not of aware of any other tool I can use to test and compare

Edit: tried to use `taplo-cli` to validate the schema and it worked fine, so not sure where the problem is. Anyways, since the format is supported by us and the schema generated is correct, this IDE problem shouldn't concern us.

Edit2: after closing and re-opening vscode, the validation seems to accept it and work fine so maybe after this PR schema fix, users won't have to deal with this at all.

lucasfernog:
the Even Better TOML extension schema validation isn't perfect, at least autocompletion doesn't always work. That's why we use JSON by default for capabilities.

lucasfernog:
a change file would be nice here i think

amrbashir:
Added

lucasfernog (approved):

PR #9105, opened 2024-03-07
[feat][nsis] update installer.nsi template check $INSTDIR

linguofeng:
When the installation path is selected under D:\apps, the actual installation path should be D:\apps\myapp Because if you don't do this, the D:\apps directory will be emptied and deleted when the program is uninstalled.

So we should check if INSTDIR path contains ${PRODUCTNAME}


amrbashir (commented):
I don't like forcing the user to install to a directory with a specific name, if they wish to install to `D:/apps` they should be allowed to even if it is not empty. Installers on windows have always allowed this behavior.

linguofeng:
Then when the user does not understand that the wrong installation directory has been selected, all data is deleted when the uninstallation is performed. Because many Windows users don’t know the system very well, what we can do is protect users’ data.

https://github.com/tauri-apps/tauri/blob/78a905e47e8dce3cd3ccdb95aaef476c36b229c6/tooling/bundler/src/bundle/windows/templates/installer.nsi#L294

linguofeng:
https://github.com/electron-userland/electron-builder/issues/3201

FabianLars:
If it's just about the /R flag we could simply remove the "Delete user data" checkbox, or let it work on the known paths for tauri's appX apis only (which currently means folders with the bundle id). If the dev saves data outside of that in the install dir directly then so be it 🤷

linguofeng:
Sorry, wrong line number quoted

https://github.com/tauri-apps/tauri/blob/78a905e47e8dce3cd3ccdb95aaef476c36b229c6/tooling/bundler/src/bundle/windows/templates/installer.nsi#L664

Not about appdata dir，It INSTDIR

FabianLars:
ah okay, the issue you linked was about the /r flag which we call for the instdir if the app data box is checked. Either way, my comment is still valid because RmDir will only remove a directory if it's completely empty, _unless_ the /R flag is applied (which we only do if the box is checked).

Itsusinn:
Could you please allow us customize this nsis behavior by adding a `forceInstallDir` option in
https://tauri.app/v1/api/config#nsisconfig ? @FabianLars 

One of our users delete all his files under `D:\Apps` : (

FabianLars:
A config makes no sense to me honestly. It should be changed for everyone.

PR #9107, opened 2024-03-07
fix(core): fix emit skipping webview listeners if filter wasn't provided

amrbashir:

### DIFF
--- a/.changes/core-emit-js-all-targets.md
+++ b/.changes/core-emit-js-all-targets.md
@@ -1,0 +1,5 @@
+---
+'tauri': 'patch:bug'
+---
+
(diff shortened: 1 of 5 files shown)


.changes/core-emit-js-all-targets.md
```diff
@@ -0,0 +1,5 @@
+---
+'tauri': 'patch:bug'
+---
+
+Fix `emit` and `emit_to` (when used with `EventTarget::Any`) alawys skipping the webview listeners.
```
lucasfernog:
```suggestion
Fix `emit` and `emit_to` (when used with `EventTarget::Any`) always skipping the webview listeners.
```

lucasfernog (approved):

PR #9112, opened 2024-03-07
fix(iOS): Update .gitignore of template to exclude *.xcodeproj/ for iOS projects

MikeChongCan:
We don't need to track `xcodeproj` in git because it is generated using xcodegen

### DIFF
--- a/tooling/cli/templates/mobile/ios/.gitignore
+++ b/tooling/cli/templates/mobile/ios/.gitignore
@@ -1,3 +1,4 @@
 xcuserdata/
 build/
 Externals/
+*.xcodeproj/


lucasfernog:
Well the whole gen folder is generated, and if you want to commit it, why wouldn't you want to commit the xcodeproj too?

MikeChongCan:
In this case, do we need to ignore the whole gen folder ?

FabianLars:
> In this case, do we need to ignore the whole gen folder ?

No, if you do ignore it then you'll always have to run `ios init` every time you clone the project. It's similar for .xcodeproj too since it contains settings you probably want to share with others/your team.

PR #9113, opened 2024-03-07
feat(core): add dev function

lucasfernog:


lucasfernog:
particularly needed by the [notification plugin](https://github.com/tauri-apps/plugins-workspace/blob/644eb448bb5d32ee837c35902bfd14e3914e2bc5/plugins/notification/src/desktop.rs#L189)

amrbashir (approved):

PR #9116, opened 2024-03-07
refactor(acl): use URLPattern instead of glob for remote URLs

lucasfernog:
Changes the remote URL configuration to use URLPattern instead of glob patterns which works better for URLs.


core/tauri-utils/Cargo.toml
```diff
@@ -33,6 +33,8 @@ json5 = { version = "0.4", optional = true }
 toml = { version = "0.8", features = [ "parse" ] }
 json-patch = "1.2"
 glob = "0.3"
+urlpattern = "0.2"
+regex = "1"
```
amrbashir, line 37: I hate having to depend on regex, but there is no way around it apparently

amrbashir (approved):

PR #9122, opened 2024-03-07
feat(core): expose the image module

lucasfernog:
This exposes the JsImage type so we can use it on our plugins :)

### DIFF
--- a/core/tauri/src/image/mod.rs
+++ b/core/tauri/src/image/mod.rs
@@ -198,13 +198,13 @@ impl TryFrom<Image<'_>> for tray_icon::Icon {
 #[derive(serde::Deserialize)]
 #[serde(untagged)]
 #[non_exhaustive]
-pub enum JsImage<'a> {
+pub enum JsImage {
   /// A reference to a image in the filesystem.
   #[non_exhaustive]
   Path(std::path::PathBuf),
   /// Image from raw bytes.
   #[non_exhaustive]
-  Bytes(&'a [u8]),
+  Bytes(Vec<u8>),
   /// An image that was previously loaded with the API and is stored in the resource table.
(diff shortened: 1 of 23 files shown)


core/tauri/src/image/mod.rs
```diff
@@ -186,20 +188,39 @@ impl TryFrom<Image<'_>> for tray_icon::Icon {
...
+/// The counterpart of the `@tauri-apps/api/image` Image class.
+///
+/// # Stability
+///
+/// The stability of the variants are not guaranteed, and matching against them is not recommended.
+/// Use [`JsImage::into_img`] instead.
 #[derive(serde::Deserialize)]
 #[serde(untagged)]
+#[non_exhaustive]
 pub enum JsImage<'a> {
```
amrbashir: `Image` struct is re-exported from root, do we want to do the same for `JsImage` or remove the re-export from root? or just keep different re-exports?

core/tauri/src/image/mod.rs
```diff
@@ -186,20 +188,39 @@ impl TryFrom<Image<'_>> for tray_icon::Icon {
   }
 }
 
+/// The counterpart of the `@tauri-apps/api/image` Image class.
```
amrbashir: This is not exactly true though, I believe this is equivalent to `string | number[] | Unit8Array | ArrayBuffer | Image` so maybe needs to change the wording and mention `transformImage`?

amrbashir (changes requested):

core/tauri/src/image/mod.rs
lucasfernog: i feel like we have too many root re-exports tbh

amrbashir (approved):

PR #9123, opened 2024-03-07
docs: adding 'desktop only' pill to menu module

kandrelczyk:
It's a silly change but I've noticed that menu module is missing 'desktop only' pill. Tray module has this pill and it's confusing that menu doesn't.


amrbashir (approved):
Thank you, these little things matter ❤️️.

PR #9141, opened 2024-03-10
feat(core): allow swapping the assets implementation

lucasfernog:
The current `Context::assets_mut` is useless since you must provide a type that matches the generic arg in Context, so we're dropping it in favor of a Box<dyn Assets> instead.

Additionally, during codegen you can now define your custom assets implementation (by default we embed the frontendDist assets):
```rust
struct MyAssets {
...
}
impl tauri::Assets for MyAssets {
...
}

fn main() {
  let assets = MyAssets {};
  tauri::generate_context(
    assets = assets
  );
}
```

### DIFF
--- a/core/tauri/src/lib.rs
+++ b/core/tauri/src/lib.rs
@@ -343,9 +343,9 @@ pub fn dev() -> bool {
 /// # Stability
 /// This is the output of the [`generate_context`] macro, and is not considered part of the stable API.
 /// Unless you know what you are doing and are prepared for this type to have breaking changes, do not create it yourself.
-pub struct Context<A: Assets> {
+pub struct Context {
   pub(crate) config: Config,
-  pub(crate) assets: Box<A>,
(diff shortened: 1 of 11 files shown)


core/tauri/src/lib.rs
```diff
@@ -387,13 +387,14 @@ impl<A: Assets> Context<A> {
 
   /// The assets to be served directly by Tauri.
   #[inline(always)]
-  pub fn assets(&self) -> &A {
+  #[allow(clippy::borrowed_box)]
+  pub fn assets(&self) -> &Box<dyn Assets> {
```
amrbashir: any reason not to return `&dyn Assets` here?

amrbashir (dismissed):

core/tauri/src/lib.rs
lucasfernog: oops forgot i could use as_ref, gonna push

PR #9144, opened 2024-03-11
fix(core): race between drop old JS listeners and create new listeners on page load

i-c-b:
@amrbashir mentioned removing a `for` loop (not sure which one) as an alternate solution on [Discord](<https://discord.com/channels/616186924390023171/986184094050316358/1216134672317284445>) so let's go with that if it's the better solution.


amrbashir (approved):

amrbashir:
I will open another PR for the alternative fix.

PR #9147, opened 2024-03-11
refactor(core): add setup() to the Assets trait

lucasfernog:
Follow up for #9141 where we allow custom asset providers to access the App instance for initialization logic.

### DIFF
--- a/core/tauri/src/lib.rs
+++ b/core/tauri/src/lib.rs
@@ -343,9 +343,9 @@ pub fn dev() -> bool {
 /// # Stability
 /// This is the output of the [`generate_context`] macro, and is not considered part of the stable API.
 /// Unless you know what you are doing and are prepared for this type to have breaking changes, do not create it yourself.
-pub struct Context<A: Assets> {
+pub struct Context {
   pub(crate) config: Config,
-  pub(crate) assets: Box<A>,
+  pub(crate) assets: Box<dyn Assets>,
   pub(crate) default_window_icon: Option<image::Image<'static>>,
   pub(crate) app_icon: Option<Vec<u8>>,
   #[cfg(all(desktop, feature = "tray-icon"))]
@@ -356,7 +356,7 @@ pub struct Context<A: Assets> {
   pub(crate) runtime_authority: RuntimeAuthority,
 }
 
-impl<A: Assets> fmt::Debug for Context<A> {
+impl fmt::Debug for Context {
   fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
     let mut d = f.debug_struct("Context");
     d.field("config", &self.config)
@@ -372,7 +372,7 @@ impl<A: Assets> fmt::Debug for Context<A> {
   }
 }
 
-impl<A: Assets> Context<A> {
+impl Context {
   /// The config the application was prepared with.
   #[inline(always)]
   pub fn config(&self) -> &Config {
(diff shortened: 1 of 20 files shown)


core/tauri/src/lib.rs
```diff
@@ -338,14 +339,47 @@ pub fn dev() -> bool {
...
+  /// Iterator for the assets.
+  fn iter(&self) -> Box<dyn Iterator<Item = (&str, &[u8])> + '_>;
+
+  /// Gets the hashes for the CSP tag of the HTML on the given path.
+  fn csp_hashes(&self, html_path: &AssetKey) -> Box<dyn Iterator<Item = CspHash<'_>> + '_>;
+}
+
+impl<R: Runtime> Assets<R> for EmbeddedAssets {
+  fn get(&self, key: &AssetKey) -> Option<Cow<'_, [u8]>> {
+    EmbeddedAssets::get(self, key)
```
amrbashir, line 361: any reason not to do it like this?
```suggestion
   self.get(key)
```

core/tauri/src/lib.rs
```diff
@@ -387,14 +421,20 @@ impl Context {
...
   #[inline(always)]
-  pub fn assets_mut(&mut self) -> &mut Box<dyn Assets> {
-    &mut self.assets
+  pub fn assets_mut(&mut self) -> &mut Box<dyn Assets<R>> {
+    self.assets.as_mut().unwrap()
+  }
+
+  ///
+  #[inline(always)]
+  pub fn replace_assets(&mut self, assets: Box<dyn Assets<R>>) -> Box<dyn Assets<R>> {
```
amrbashir: I think we should name this `set_assets` and remove `assets_mut` to match `tray_icon` and `set_tray_icon`, `default_window_icon` and `set_default_window_icon`

core/tauri/src/lib.rs
```diff
@@ -387,14 +421,20 @@ impl Context {
...
+  pub fn assets(&self) -> &dyn Assets<R> {
+    self.assets.as_ref().unwrap().as_ref()
   }
 
   /// A mutable reference to the assets to be served directly by Tauri.
   #[inline(always)]
-  pub fn assets_mut(&mut self) -> &mut Box<dyn Assets> {
-    &mut self.assets
+  pub fn assets_mut(&mut self) -> &mut Box<dyn Assets<R>> {
+    self.assets.as_mut().unwrap()
```
amrbashir: should we return Options here instead of unwrapping?

amrbashir (changes requested):

core/tauri/src/lib.rs
lucasfernog: it might call the trait implementation instead of the embeddedassets public method, better to be explicit or change the method name

core/tauri/src/lib.rs
lucasfernog: cool

core/tauri/src/lib.rs
lucasfernog: it's an Option internally only for us to replace the value and return the previous one (so custom asset providers can fallback to the embedded assets if they desire)

lucasfernog: but they'll never be None

core/tauri/src/lib.rs
amrbashir: cool

core/tauri/src/lib.rs
amrbashir: but we provide `set_assets` method and so it could change by users and then panic on this unwrap later, better to be explicit here and also match other methods like `tray_icon` and `set_tray_icon`

lucasfernog: we'll remove this one, and even if the user use set_assets the value will never be None, we just want to be able to take() the previous value

amrbashir: ah right, I was under the assumption that `set_assets` will take an `Option<Box<dyn Assets>>` and users will be allowed to set `None`.

amrbashir: Should probably remove the `Option` internally and do the `.replace()` manually by calling `mem::replace`

amrbashir (approved):

PR #8930, opened 2024-02-21
 fix(event): clear residual js listeners (#8916 )

canxin121:
fix clear residual listeners, close #8916 

I noticed that in v1, `js_listeners` were stored on the frontend, so refreshing the page cleans up the original `js_listeners` (note that this doesn't mean that useEffect cleans up the side-effects, as it doesn't have a way to clean up the side-effects before refreshing the page).

However, in v2, `js_listeners` are stored not only in the front end, but also in the back end, and the number of events triggered is determined by the `js_listeners` that stored in the backend, which results in a pileup of `js_listeners` that can't be cleaned up when the page is refreshed.

I added a new js command `unlisten_all` and invoke it at the end of `core.js` to solve this problem. `unlisten_all` only cancels the `js_listeners` at the frontend and backend of the *current window*. 

This fix works fine on my machine, in both single-window and multi-window scenarios.

PR #9151, opened 2024-03-11
enhance(core/event): filter js listeners on rust side only then emit filtered ids to be triggered

amrbashir:
fix regression introduced in https://github.com/tauri-apps/tauri/pull/8930 , and reported in https://github.com/tauri-apps/tauri/pull/8930#issuecomment-1986279046


.changes/core-js-unlisten-all-regression.md
```diff
@@ -0,0 +1,5 @@
+---
+'tauri': patch:bug
+---
+
+Fix a regression where new listeners regirested in JS are cleared on page load.
```
lucasfernog:
```suggestion
Fix a regression where new listeners registered in JS are cleared on page load.
```

lucasfernog:
I need more time to test this one, but we also need to release what we already have (important bug fixes). Can you recreate the change file later?

amrbashir:
removed the change file since an alternative fix has been released now

lucasfernog (approved):

PR #9156, opened 2024-03-12
feat(core): allow defining global API script on plugin build

lucasfernog:
Adds `tauri_plugin::Builder::global_api_script_path` so plugin authors can define the JavaScript global API bindings (supposed to be injected to `window.__TAURI__`) at compile time, so the string is only part of the binary when withGlobalTauri is true. Currently this needs to be done manually at runtime (and it's always added to the binary via include_str).

### DIFF
--- a/core/tauri-utils/src/plugin.rs
+++ b/core/tauri-utils/src/plugin.rs
@@ -1,0 +1,60 @@
+// Copyright 2019-2024 Tauri Programme within The Commons Conservancy
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: MIT
+
+//! Compile-time and runtime types for Tauri plugins.
+#[cfg(feature = "build")]
+pub use build::*;
+
+#[cfg(feature = "build")]
+mod build {
+  use std::{
+    env::vars_os,
+    path::{Path, PathBuf},
+  };
+
+  const GLOBAL_API_SCRIPT_PATH_KEY: &str = "GLOBAL_API_SCRIPT_PATH";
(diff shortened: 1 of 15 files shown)


core/tauri/src/manager/mod.rs
```diff
@@ -188,6 +188,9 @@ pub struct AppManager<R: Runtime> {
   /// Application pattern.
   pub pattern: Arc<Pattern>,
 
+  /// Global API script.
+  pub global_api_script: Arc<Option<String>>,
```
amrbashir: should probably prefeix with `plugin` like this `plugin_global_api_script` to make it clear that these are collected from plugins which is separate from global api defined by tauri crate itself.

core/tauri-utils/src/plugin.rs
```diff
@@ -0,0 +1,60 @@
...
+        let script = std::fs::read_to_string(&script_path).unwrap_or_else(|e| {
+          panic!(
+            "failed to read global script path at {}: {e}",
+            script_path.display()
+          )
+        });
+
+        merged_script.push_str(";(function () {\n");
+        merged_script.push_str(&script);
+        merged_script.push_str("})()");
```
amrbashir: This should be handled by the user, we shouldn't also merge all the plugins into a single script, instead we should save them as a vector of scripts and load them one by one so they have a different execution context, otherwise different scripts could have a similar intermediate variable and cause conflicts and errors.

amrbashir (changes requested):

amrbashir (approved):

PR #9166, opened 2024-03-12
fix(core/path): remove suffix in basename only once

amrbashir:
ref: #9064

### DIFF
--- a/tooling/api/src/path.ts
+++ b/tooling/api/src/path.ts
@@ -726,9 +726,9 @@ async function join(...paths: string[]): Promise<string> {
  * Returns the directory name of a `path`. Trailing directory separators are ignored.
  * @example
  * ```typescript
- * import { dirname, appDataDir } from '@tauri-apps/api/path';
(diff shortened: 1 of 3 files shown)


tooling/api/src/path.ts
```diff
@@ -769,10 +768,8 @@ async function extname(path: string): Promise<string> {
  * Returns the last portion of a `path`. Trailing directory separators are ignored.
  * @example
  * ```typescript
- * import { basename, resolveResource } from '@tauri-apps/api/path';
- * const resourcePath = await resolveResource('app.conf');
- * const base = await basename(resourcePath);
- * assert(base === 'app.conf');
+ * import { basename } from '@tauri-apps/api/path';
+ * const base = await basename('path/to/app.conf');
```
lucasfernog, line 772:
```suggestion
 * const base = await basename('path/to/app.conf');
 * assert(base === 'app.conf');
```

lucasfernog (approved):

Issue #8363, opened 2023-12-10
[bug] build fail with Chinese productName in  tauri.conf.json  failed to bundle project: error running light.exe


makao007:
### Describe the bug

I edit the tauri.conf.json file with a Chinese ProductName and run `pnpm tauri build` but fail
```
Error failed to bundle project: error running light.exe
```

### Reproduction

 edit the tauri.conf.json, update ProductName  to "你好"
 pnpm tauri build

### Expected behavior

build successfuly

### Platform and versions

```text
[✔] Environment
    - OS: Windows 10.0.22621 X64
    ✔ WebView2: 120.0.2210.61
    ✔ MSVC: Visual Studio Community 2022
    ✔ rustc: 1.74.0 (79e9716c9 2023-11-13)
    ✔ cargo: 1.74.0 (ecb9851af 2023-10-18)
    ✔ rustup: 1.26.0 (5af9b9484 2023-04-05)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (default)
    - node: 20.10.0
    - pnpm: 8.11.0
    - npm: 10.2.3
```

### Stack trace

```text
Compiling tauri-app v0.0.0 (D:\code\web\tauri-app\src-tauri)
    Finished release [optimized] target(s) in 35.70s
        Info Target: x64
     Running candle for "main.wxs"
     Running light to produce D:\code\web\tauri-app\src-tauri\target\release\bundle\msi\你好_0.0.0_x64_en-US.msi
       Error failed to bundle project: error running light.exe
 ELIFECYCLE  Command failed with exit code 1.
```

FabianLars:
As far as i know WiX (the .msi creation tool) simply doesn't support chinese characters in this context, in which case https://github.com/tauri-apps/tauri/issues/8109 may be the best workaround.

Can you run the build command again with the `--verbose` flag (`pnpm tauri build --verbose`) so we get an actual error message please?

makao007:
Thanks for your advise.
I run `pnpm tauri build --verbose` and found below error:
```
D:\code\web\tauri-app\src-tauri\target\release\wix\x64\main.wxs(20) : error LGHT0311 : A string was provided with characters that are not available in the specified database code page '1252'. Either change these characters to ones that exist in the database's code page, or update the database's code page by modifying one of the following attributes: Product/@Codepage, Module/@Codepage, Patch/@Codepage, PatchCreation/@Codepage, or WixLocalization/@Codepage.
```
[...]

Issue #8349, opened 2023-12-06
[bug] Tauri unwantedly converts PascalCase package product name to kebab-case


nekename:
### Describe the bug

The string configured as `package.productName` in tauri.conf.json in my project is "OpenDeck". This is the string I would like to be shown in start menus, docks, app launchers, software stores (e.g. GNOME Software), etc. 

However, Tauri unwantedly converts this string to "open-deck" for the output artifact names (resulting in `open-deck_2.0.0_amd64.deb` and so forth) and for the executable name (meaning that it's `open-deck` to run it from a shell). I would like these both to be `opendeck`. However, the docs don't appear to mention any property that can be changed to override this behaviour.

### Expected behavior

I should be able to specify an alternative string to use for these properties in my `tauri.conf.json` instead of having to deal with an inaccurate case conversion.

### Platform and versions

```text
[✔] Environment
    - OS: Ubuntu 22.04 X64
    ✔ webkit2gtk-4.0: 2.42.2
    ✔ rsvg2: 2.52.5
    ✔ rustc: 1.73.0 (cc66ad468 2023-10-03)
    ✔ cargo: 1.73.0 (9c4383fb5 2023-08-26)
    ✔ rustup: 1.26.0 (5af9b9484 2023-04-05)
    ✔ Rust toolchain: stable-x86_64-unknown-linux-gnu (environment override by RUSTUP_TOOLCHAIN)
    - node: 18.18.0
    - yarn: 1.22.21
    - npm: 10.2.4

[-] Packages
    - tauri [RUST]: 1.5.2
    - tauri-build [RUST]: 1.5.0
    - wry [RUST]: 0.24.4
    - tao [RUST]: 0.16.5
    - tauri-cli [RUST]: 1.5.6
    - @tauri-apps/api [NPM]: 1.5.1
    - @tauri-apps/cli [NPM]: 1.5.6 (outdated, latest: 1.5.7)

[-] App
    - build-type: bundle
    - CSP: default-src 'self';
    - distDir: ../build
    - devPath: http://localhost:5173/
    - framework: Svelte
    - bundler: Vite
```

FabianLars:
I think <https://github.com/tauri-apps/tauri/issues/8109> covers this right? You could set `opendeck` as the packageName and `OpenDeck` as displayName then.

nekename:
Yep, but it's a shame it doesn't exist ;)

nekename:
@FabianLars Any update?

[...]

Issue #8109, opened 2023-10-26
[feat] displayName for package


moranbw:
### Describe the problem

This is probably nitpicking, but I am looking to have two separate "names" for my Tauri application.  Right now, I understand that in `tauri.conf.json` I can define `package` -> `productName`.  This is what is used to name installers, exe's, log files (if using the [plugin](https://github.com/tauri-apps/plugins-workspace/tree/v1/plugins/log)), etc.  

But it also is what is used for the "user-facing" display name, for example what it is called in the Start Menu on Windows.  So for `my-app`, if I want the user to have a "user-friendly" name to search for, I need to set `productName` to `My App`.  But then the underlying executable is called `My App.exe`, which just kind of irks me lol.  I think it should be `my-app.exe`.  It just seems more professional, and more in line with what I see in other apps I have installed on my system.

### Describe the solution you'd like

Keep `productName` as the default way of naming your app.  But add an option in `package` called `displayName`, which determines the "user-facing" name of the application, while keeping file-system friendly `productName` for .exe's, etc.

### Alternatives considered

I am thinking that this would be an app manifest thing on Windows.  Not sure about Linux and Mac.

NXTaar:
Hope it will be implemented. Have the same problem here for a Windows app

eythaann:
x3, waiting for this feature.

mariotaku:
Hi, I have created a PR to this issue, please try and let me know if it works for you!

hengkx:
Is there any progress? I really need this function

FabianLars:
The PR is linked right above your comment -> https://github.com/tauri-apps/tauri/pull/9375
We can't backport that exact solution to v1 though because it's a breaking change.

amrbashir:
Just a note that we have decided to add `mainBinaryName` option in https://github.com/tauri-apps/tauri/pull/10977 to control the main binary name, this makes `productName` essentially the displayName (which is what it was supposed to) and the new option for main binary name only.

PR #9175, opened 2024-03-13
feat(bundler): display name for bundle

mariotaku:
This PR Is for #8109, #8349 and #8363.

For Linux, productName needs to be converted to lower-kebab-case. To make sure app desktop entry displays user friendly name, `displayName` has been added.

Example:

```json
{
  "productName": "myapp",
  "displayName": "My App"
}
```

### DIFF
--- a/.changes/display-name-tauri-conf.md
+++ b/.changes/display-name-tauri-conf.md
@@ -1,0 +1,6 @@
+---
+"tauri-cli": patch:enhance
+"tauri-build": patch:enhance
+---
+
+Added a new configuration option `tauri.conf.json > displayName` to specify a user-friendly name for the bundle. If it is empty, the default value is the `productName` field.
--- a/core/tauri-config-schema/schema.json
+++ b/core/tauri-config-schema/schema.json
@@ -19,6 +19,13 @@
       ],
       "pattern": "^[^/\\:*?\"<>|]+$"
     },
+    "displayName": {
+      "description": "App display name, will use `productName` if not set.",
+      "type": [
+        "string",
+        "null"
+      ]
+    },
     "version": {
       "description": "App version. It is a semver version number or a path to a `package.json` file containing the `version` field. If removed the version number from `Cargo.toml` is used.",
       "type": [
--- a/core/tauri-utils/src/config.rs
+++ b/core/tauri-utils/src/config.rs
@@ -1973,6 +1973,9 @@ pub struct Config {
   #[serde(alias = "product-name")]
   #[cfg_attr(feature = "schema", validate(regex(pattern = "^[^/\\:*?\"<>|]+$")))]
   pub product_name: Option<String>,
+  /// App display name, will use `productName` if not set.
+  #[serde(alias = "display-name")]
+  pub display_name: Option<String>,
   /// App version. It is a semver version number or a path to a `package.json` file containing the `version` field. If removed the version number from `Cargo.toml` is used.
   #[serde(deserialize_with = "version_deserializer", default)]
   pub version: Option<String>,
@@ -2578,6 +2581,7 @@ mod build {
     fn to_tokens(&self, tokens: &mut TokenStream) {
       let schema = quote!(None);
       let product_name = opt_str_lit(self.product_name.as_ref());
+      let display_name = opt_str_lit(self.display_name.as_ref());
       let version = opt_str_lit(self.version.as_ref());
       let identifier = str_lit(&self.identifier);
       let app = &self.app;
@@ -2590,6 +2594,7 @@ mod build {
         ::tauri::utils::config::Config,
         schema,
         product_name,
+        display_name,
         version,
         identifier,
         app,
--- a/core/tauri/src/test/mod.rs
+++ b/core/tauri/src/test/mod.rs
@@ -99,6 +99,7 @@ pub fn mock_context<R: Runtime, A: Assets<R>>(assets: A) -> crate::Context<R> {
     config: Config {
       schema: None,
       product_name: Default::default(),
+      display_name: Default::default(),
       version: Default::default(),
       identifier: Default::default(),
       app: AppConfig {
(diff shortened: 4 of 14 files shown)


FabianLars:
Hi, thanks for contributing! This is not a full review yet, as i don't have time today to actually try it out but for a start 2 remarks:
1) This may have been added by accident but please remove the deb/rpm changes and open another PR for this if you want to see this in Tauri.
2) in my opinion we shouldn't add this kind of platform specific configs but keep relying on https://tauri.app/v1/api/config#platform-specific-configuration
3) Also, once this is not a draft anymore we'll need a changefile similar to those here: https://github.com/tauri-apps/tauri/tree/dev/.changes

mariotaku:
@FabianLars Hi! Thanks for the quick comment.

1. I'll move deb/rpm changes to another PR. 
2. Sorry, I didn't notice we already have platform specific configuration already, I'll utilize it and add `displayName` only.
3. Sure! I'll be making this change too.

mariotaku:
Done updating the change. Please let me know if there are anything missing :)

mariotaku:
Hi @FabianLars, is it possible to include this change in 2.0 release?

FabianLars:
I hope so, yes. I didn't try it yet but your changes look correct and reasonable as far as i can see.

Paging @amrbashir for a double check on the windows installers

xuchaoqian:
Is it easy to confuse "product name" and "display name"?
How about using "bundle name" as the name for the bundle?

mariotaku:
@xuchaoqian Sounds good too, especially for Linux

amrbashir:
I am not keen on adding another option that is almost identical to `productName`.

#8109 can be fixed by keeping the executable name generated from cargo and avoid renaming it.

#8349 can be fixed by removing the implicit kebab-case conversion and ensure that `productName` is used as is in the generated `.desktop` file.

#8363 can be fixed by modifying the codepage and/or using a Chinese locale.

FabianLars:
sounds good to me too ig 🤷

amrbashir:
cc @lucasfernog

lucasfernog:
I like @amrbashir idea. But we need to at least make a copy of the binary to change its name, otherwise it'll look weird on linux and windows.

If we're going to break the bundler output paths, please do so asap.

amrbashir:
> But we need to at least make a copy of the binary to change its name, otherwise it'll look weird on linux and windows

Could you explain a bit more, not sure if I understand the issue

mariotaku:
@amrbashir Thanks for your input, and I have some ideas about the solutions you've mentioned.

> #8109 can be fixed by keeping the executable name generated from cargo and avoid renaming it.

This sound great, will you make bundler respect binary name defined in `Cargo.toml`? Or I could try making a PR.
 
> #8349 can be fixed by removing the implicit kebab-case conversion and ensure that `productName` is used as is in the generated `.desktop` file.

I'm afraid this will have some issues, because debian package names are having lower-kebab-case. If you override `productName` in `tauri.conf.json`, title in desktop file generated will become lower-kebab-case too.

amrbashir:
> This sound great, will you make bundler respect binary name defined in `Cargo.toml`? Or I could try making a PR.

Feel free to open a PR. The CLI renames the binary after it gets generated by Cargo, we just need to remove that behavior.

> I'm afraid this will have some issues, because debian package names are having lower-kebab-case. If you override `productName` in `tauri.conf.json`, title in desktop file generated will become lower-kebab-case too.

Not sure I quite follow what you mean, but we only need to use kebab-case where required and keep everything else as the normal non-kebab-case productName.

mariotaku:
@amrbashir 


>>     I'm afraid this will have some issues, because debian package names are having lower-kebab-case. If you override productName in tauri.conf.json, title in desktop file generated will become lower-kebab-case too.
> Not sure I quite follow what you mean, but we only need to use kebab-case where required and keep everything else as the normal non-kebab-case productName.

For bundles like `deb`, the name [must be kebab-case](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-source). So this is where the conversion required. But this PR was trying to override the automatic conversion when it can't produce desired one. 

For example, when the product name is `webOS Dev Manager`, it will be converted to `web-os-dev-manager` while the desired one would be `webos-dev-manager`, like `OpenDeck` in #8349.

Can this case be covered?

amrbashir:
I believe that could be covered by using a custom `.desktop` template.

nekename:
> For example, when the product name is `webOS Dev Manager`, it will be converted to `web-os-dev-manager` while the desired one would be `webos-dev-manager`, like `OpenDeck` in #8349.

Exactly

mariotaku:
> > For example, when the product name is `webOS Dev Manager`, it will be converted to `web-os-dev-manager` while the desired one would be `webos-dev-manager`, like `OpenDeck` in #8349.
> 
> Exactly

I solved it by doing this:

1. Create `tauri.linux.conf.json` and set `productName` to the one you want (for me, webos-dev-manager`) [example](https://github.com/webosbrew/dev-manager-desktop/blob/e76afaac1a9ecf98aef023aa0e246471de08f810/src-tauri/tauri.linux.conf.json#L2)
2. Create a `.desktop` entry with the display name you want, and configure to make tauri use that. [example](https://github.com/webosbrew/dev-manager-desktop/blob/e76afaac1a9ecf98aef023aa0e246471de08f810/src-tauri/tauri.conf.json#L44)

Not very straightforward but works fine

nekename:
> I solved it by doing this:

Oh, awesome! Do you know if that works on Tauri v1 as well (otherwise I can just https://tryitands.ee)?

Edit: it seems to work! I had to surround `productName` in `package` for it to work in Tauri v1.

amrbashir:
Just a note that we have decided to add `mainBinaryName` option in https://github.com/tauri-apps/tauri/pull/10977 to control the main binary name, this makes `productName` essentially the displayName (which is what it was supposed to) and the new option for main binary name only.

PR #9179, opened 2024-03-13
refactor(image): expose constructor, unify size getters

lucasfernog:
- Expose Image JS constructor for internal use (when a plugin returns a resource id for instance)
- Added Image::size() instead of Image::width() and Image::height() (1 IPC call instead of 2 - usually you want both width and height)


.changes/image-size-refactor.md
```diff
@@ -0,0 +1,6 @@
+---
+"@tauri-apps/api": patch:breaking
+"tauri": patch:breaking
+---
+
+Removed `width` and `height` commands from the `Image` class, use `size` instead.
```
amrbashir:
```suggestion
Removed `width` and `height` methods on the JS `Image` class, use `size` instead.
```

amrbashir (approved):

PR #9219, opened 2024-03-19
fix(core): menu `remove` js binding not working

Legend-Master:


lucasfernog (approved):
Nice catch, thank you!!Nice catch, thank you!!

PR #9225, opened 2024-03-19
fix(core): Fix menu.remove trying to remove the item from itself instead of from the menu

FabianLars:
follow up to https://github.com/tauri-apps/tauri/pull/9219/ - only noticed because clippy screamed at me in anothe pr.

don't mind the commit message, i was tired and mixed it up.

btw, i had to disable the isolation pattern in the api example app, otherwise _nothing_ worked (it even had css issues)


Legend-Master:
Oops, was editing from vscode remote repository, didn't notice it without the help of rust analyzer, sorry for the broken "fix" >.<

lucasfernog (approved):

PR #9246, opened 2024-03-22
fix(windows): changing WebView visibility on hide/show/minimize

### LINKED ISSUE #6864: [bug] visibilitychange event not working correctly
### Describe the bug

The ``visibilitychange`` event only fires when the tauri app is minimized or maximized.

### Reproduction

in a react project. App.tsx:
```js
    document.addEventListener('visibilitychange', (e) => {
        console.log('visibilitychange', e);
    });
```
then hide the tauri window with another window.

### Expected behavior

this event should also be registered if the tauri window is being obscured by another one (that is exactly what happens in the browser). Example https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API

### Platform and versions

```text
Environment
  › OS: Windows 10.0.19045 X64
  › Webview2: 112.0.1722.64
  › MSVC: 
  › Node.js: 16.17.1
  › npm: 8.15.0
  › pnpm: Not installed!
  › yarn: 1.22.19
  › rustup: 1.25.2
  › rustc: 1.68.0
  › cargo: 1.68.0

Packages
  › @tauri-apps/cli [NPM]: 1.1.1
  › @tauri-apps/api [NPM]: 1.3.0
  › tauri [RUST]: 1.2.5,
  › tauri-build [RUST]: 1.3.0,
  › wry [RUST]: 0.23.4,

App
  › build-type: bundle
  › CSP: unset
  › distDir: ../dist
  › devPath: http://localhost:1420/
  › framework: React

App directory structure
  ├─ .vscode
  ├─ node_modules
  ├─ public
  ├─ src
  └─ src-tauri
```

### Stack trace

_No response_

### Additional context

_No response_

Legend-Master:
Reference: #6864

We need to call `SetIsVisible` manually according to https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2controller?view=webview2-winrt-1.0.2365.46#isvisible

> Determines whether to show or hide the WebView. If IsVisible is set to false, the WebView is transparent and is not rendered. However, this does not affect the window containing the WebView (the ParentWindow parameter that was passed to [CoreWebView2Environment.CreateCoreWebView2ControllerAsync](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2environment?view=webview2-winrt-1.0.2365.46#createcorewebview2controllerasync)). WebView as a child window does not get window messages when the top window is minimized or restored. For performance reasons, developers should set the IsVisible property of the WebView to false when the app window is minimized and back to true when the app window is restored. The app window does this by handling SIZE_MINIMIZED and SIZE_RESTORED command upon receiving WM_SIZE message. There are CPU and memory benefits when the page is hidden. For instance Chromium has code that throttles activities on the page like animations and some tasks are run less frequently. Similarly, WebView2 will purge some caches to reduce memory usage.

This is a quick dirty fix to make it work at least for minimize / restore / show / hide (no fully covered by other windows support yet) (`Resized` fires on minimized and restored, and since we don't have an is visible event yet, the best we can do is to hook onto those methods for now), we probably want to separate this event in https://github.com/tauri-apps/tao (issue https://github.com/tauri-apps/tao/issues/892) in the future

### DIFF
--- a/core/tauri-runtime-wry/src/lib.rs
+++ b/core/tauri-runtime-wry/src/lib.rs
@@ -2637,12 +2637,12 @@ fn handle_user_message<T: UserEvent>(
           WindowMessage::Show => {
             window.set_visible(true);
             #[cfg(windows)]
-            let _ = update_webview_visibility(&window, &webviews);
+            let _ = update_webview_visibility(&webviews, !window.is_minimized());
           }
           WindowMessage::Hide => {
             window.set_visible(false);
             #[cfg(windows)]
-            let _ = update_webview_visibility(&window, &webviews);
+            let _ = update_webview_visibility(&webviews, false);
           }
           WindowMessage::Close => {
             panic!("cannot handle `WindowMessage::Close` on the main thread")
@@ -3227,7 +3227,8 @@ fn handle_event_loop<T: UserEvent>(
                 }
               }
               #[cfg(windows)]
-              let _ = update_webview_visibility(&window, &webviews);
+              let _ =
+                update_webview_visibility(&webviews, window.is_visible() && !window.is_minimized());
             }
           }
           _ => {}
@@ -3965,10 +3966,9 @@ fn clear_window_surface(
 
 #[cfg(windows)]
 fn update_webview_visibility(
-  window: &Window,
   webviews: &[WebviewWrapper],
+  is_visible: bool,
 ) -> windows::core::Result<()> {
-  let is_visible = window.is_visible() && !window.is_minimized();
   for webview in webviews {
     let controller = webview.controller();
     unsafe { controller.SetIsVisible(is_visible) }?;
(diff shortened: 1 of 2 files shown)


core/tauri-runtime-wry/src/lib.rs
```diff
@@ -3952,3 +3962,16 @@ fn clear_window_surface(
     let _ = buffer.present();
   }
 }
+
+#[cfg(windows)]
+fn update_webview_visibility(
+  window: &Window,
+  webviews: &[WebviewWrapper],
+) -> windows::core::Result<()> {
+  let is_visible = window.is_visible() && !window.is_minimized();
```
amrbashir: I think this should be passed `is_visible` as an argument, because in the case of when calling `window.hide()` we already know that it is not visible and so we don't need to call `window.is_visible()`

amrbashir (changes requested):

core/tauri-runtime-wry/src/lib.rs
Legend-Master: Ok, but I feel like that will put the responsibility of calculating `is_visible` to the caller, which is more error prone

amrbashir (approved):

PR #9262, opened 2024-03-25
fix(cli): fix migration from globalShortcut to global-shortcut

i-c-b:
Reported in Discord: <https://discord.com/channels/616186924390023171/731495028677148753/1221393137659023390>.


tooling/cli/src/migrate/config.rs
```diff
@@ -39,19 +39,27 @@ pub fn migrate(tauri_dir: &Path) -> Result<MigratedConfig> {
...
+          remote: None,
+          windows: vec!["main".into()],
+          webviews: vec![],
+          permissions,
+          platforms: None,
+        })?,
+      )?;
+    } else {
+      log::info!(
+        "Skipping capabilities migration; migrated capabilities found in {}",
```
lucasfernog, line 59: should we also print the migrated capability json here? so the user can compare both files (or maybe write to `migrates.json.tmp`)

i-c-b: The established pattern so far has been to print any manual changes. Rather than a temporary file that the user is expected to act on, would it be better to check if [`migrated.permissions`](<https://github.com/tauri-apps/tauri/blob/2398b3ff90e03afbd4598e779ba0147e1f684b36/tooling/cli/src/migrate/config.rs#L38>) is empty and, if not, make another file without the `*:default`s?

amr-crabnebula:
I believe #9931 fixed this. thanks @i-c-b

amrbashir:
I believe #9931 fixed this. thanks @i-c-b

PR #9267, opened 2024-03-25
fix(core): Deadlock when using JsImage::Resource

FabianLars:
Just showing the problem so we can discuss how to solve this.

The problem is that JsImage.into_img() locks the resource table (if it's rid) but we're also doing the same before so we end up with a deadlock. I've fixed this for the icon menu item but then realised that the menus themselves have the same problem if they have an icon child.

The easiest way to fix this is to change into_img to take a reference to the resources_table instead of a Manager but that's inconsistent with our other apis.

Maybe changing the (Sub)Menu's create_item and PayloadKind with_item to use a Manager reference instead of a resources table works too but i assume there was a reason you did this.

If you have another idea i'm all ears :) (if you wanna take over the pr, that's also fine 🤷)

P.S. This pr started as a js-only fix here: https://discord.com/channels/616186924390023171/1220447433914454156 (using Image in js for the icon didn't work because we didn't convert it from `{ rid }` to `rid`

### DIFF
--- a/core/tauri/src/menu/plugin.rs
+++ b/core/tauri/src/menu/plugin.rs
@@ -351,10 +351,11 @@ fn new<R: Runtime>(
   handler: Channel,
 ) -> crate::Result<(ResourceId, MenuId)> {
   let options = options.unwrap_or_default();
-  let mut resources_table = app.resources_table();
 
   let (rid, id) = match kind {
     ItemKind::Menu => {
+      // TODO: This doesn't work
+      let mut resources_table = app.resources_table();
       let mut builder = MenuBuilder::new(&app);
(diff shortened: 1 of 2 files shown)


lucasfernog:
thoughts on this @amrbashir ?

amrbashir:
we can just provide a separate method 
```rs
impl JsImage {
  fn into_img2(&self, resources_table: &ResourceTable) {}
}
```

can be kept private to the crate and made public later if needed.

amrbashir:
closing as I am already refactoring resources in https://github.com/tauri-apps/tauri/pull/9272 which should also fix this issue.

PR #9272, opened 2024-03-26
refactor(core): scope JS resources to the webview

amrbashir:


lucasfernog:
I think it's easier if we added a state that is specific to a window/webview so each target has its own resource table.

Currently there's a StateManager in the manager, but we can easily include one in the window or webview too (WebviewWindow can just use the window one).

plus we should still use random resource IDs

amrbashir:
> I think it's easier if we added a state that is specific to a window/webview so each target has its own resource table.
> 
> Currently there's a StateManager in the manager, but we can easily include one in the window or webview too (WebviewWindow can just use the window one).

yeah this sounds better, however I still would like to keep a global resources table so I guess I will just remove `Manager::resources_table` and add a `resources_table` method on `Window/Webview/WebviewWindow/App/AppHandle` types

> plus we should still use random resource IDs

if the resources are scoped, we don't have to use random ids, is there any other reason to use random ids?

lucasfernog:
we can check with the auditors later but I feel it's safer with random IDs. cc @chippers @tweidinger

amrbashir:
I did confirm with the auditors before opening the PR, they still need to re-test though

tweidinger:
> if the resources are scoped, we don't have to use random ids, is there any other reason to use random ids?

Hardening in depth and reducing risk. I would strongly suggest to have it randomized as we don't know all the special use cases in finished apps.

lucasfernog:
doesn't cost much to randomize it (we should already have a dependency on getrandom somewhere)

amrbashir:
> we should already have a dependency on getrandom somewhere

totally forgot we already have a dependency on it, pushed the random changes

lucasfernog (approved):

PR #9297, opened 2024-03-29
feat: add `cursor_position` getter

### LINKED ISSUE #9250: [feat] Expose methods to get cursor_position
### Describe the problem

I have a use case where I'd like to be able to detect if the mouse intersects with a tauri window.

The reason is that I use `set_ignore_cursor_events(true)` to create an overlay displaying something useful to a user.

However, I'd like to be able to allow the user to move their mouse over the window and control the opacity so they can see what behind the window.

Here is an MVP repo that shows you can use javascript to do this but only if the window is in focus.

https://github.com/Hacksore/mouse-intersection-when-ignore-mouse-events

A way in rust to listen globally for the mouse intersecting events of windows.

- intersecting (called once)
- no longer intersecting (called once)

These events would let you run some "callback" rust code (unsure if that's rusty) but for my use case I'd want to toggle some body class on the app.

tl;dr `set_ignore_cursor_events(true)`, you mouse over a tauri window you can run rust code on enter/exit.

### Alternatives considered

_No response_

### Additional context

As mentioned initially you can do this in JS but it requires the window to be focused.

Demo:

https://github.com/tauri-apps/tauri/assets/996134/cae75ec0-682d-4234-86f7-bac88352547c

amrbashir:
closes #9250


core/tauri/src/webview/webview_window.rs
```diff
@@ -1219,6 +1208,21 @@ impl<R: Runtime> WebviewWindow<R> {
...
+/// Desktop window getters.
+#[cfg(desktop)]
+impl<R: Runtime> WebviewWindow<R> {
+  /// Get the cursor position  relative to the top-left hand corner of the desktop.
+  ///
+  /// Note that the top-left hand corner of the desktop is not necessarily the same as the screen.
+  /// If the user uses a desktop with multiple monitors,
+  /// the top-left hand corner of the desktop is the top-left hand corner of the monitor at the top-left of the desktop.
+  ///
+  /// The coordinates can be negative if the top-left hand corner of the window is outside of the visible screen region.
```
FabianLars: At least on windows 0,0 will be on the top left of the _primary_ monitor, so if you have monitors to the left of your main one (like i do), they will also have negative coordinates while being in the visible screen region.

FabianLars (changes requested):

core/tauri/src/webview/webview_window.rs
amrbashir: this was copied from `tao` which is copied from `winit` ofc, maybe need to adjust the wording a bit, maybe just simplify it to:
> Get the cursor position relative to the top-left hand corner of the main monitor.

lucasfernog (dismissed):

core/tauri/src/webview/webview_window.rs
FabianLars: Sorry for the late reply (out of office), but iirc the comment was correct for x11 == it's always a positive number on x11.

No idea about macos.

I can test both tomorrow when i'm back.

pewsheen: Mac has negative coordinates as well. Just like Windows, it's relative to the main monitor.

FabianLars: sorry for the delay. X11 indeed starts with 0,0 on the leftmost monitor regardless of whether it's the primary one or not.

FabianLars:
Sorry, didn't see that my review was requested. 2 things:
- The .ts file wasn't updated yet
- My last comment above was not addressed: `X11 indeed starts with 0,0 on the leftmost monitor regardless of whether it's the primary one or not.`

amrbashir:
updated

FabianLars (approved):

PR #9309, opened 2024-03-30
fix(windows) window center doesn't consider taskbar size

### LINKED ISSUE #9259: [bug] window center doesn't consider taskbar position and size
### Describe the bug

Window center doesn't consider task bar position and size

image

### Reproduction

_No response_

### Expected behavior

image

### Full `tauri info` output

```text
Hello world example at current head (7898b601d14ed62053dd24011fabadf31ec1af45)
```

### Stack trace

_No response_

### Additional context

_No response_

Legend-Master:
Reference: #9259

Current implementation doesn't consider title bar size on creating window, which makes the window off center a little bit on creating window

Note: I didn't test this on multiple monitors


core/tauri-runtime-wry/src/lib.rs
```diff
@@ -2612,12 +2612,10 @@ fn handle_user_message<T: UserEvent>(
           // Setters
           WindowMessage::Center => {
             if let Some(monitor) = window.current_monitor() {
-              let window_size = inner_size(&window, &webviews, has_children);
-              let screen_size = monitor.size();
-              let monitor_pos = monitor.position();
-              let x = (screen_size.width as i32 - window_size.width as i32) / 2 + monitor_pos.x;
-              let y = (screen_size.height as i32 - window_size.height as i32) / 2 + monitor_pos.y;
-              window.set_outer_position(TaoPhysicalPosition::new(x, y));
+              window.set_outer_position(calculate_window_center_position(
```
amrbashir: on macOS, I think we should use objc crate to call `center` method on the `NSWindow` instead of setting the position manually.

core/tauri-runtime-wry/src/lib.rs
```diff
@@ -3400,13 +3398,10 @@ fn create_window<T: UserEvent, F: Fn(RawWindow) + Send + 'static>(
...
-        .to_logical::<i32>(scale_factor);
-      let screen_size = monitor.size().to_logical::<i32>(scale_factor);
-      let monitor_pos = monitor.position().to_logical::<i32>(scale_factor);
-      let x = (screen_size.width - window_size.width) / 2 + monitor_pos.x;
-      let y = (screen_size.height - window_size.height) / 2 + monitor_pos.y;
-
-      window_builder = window_builder.position(x as f64, y as f64);
+        .to_physical::<u32>(scale_factor);
+      let position = calculate_window_center_position(window_size, monitor);
+      let logical_position = position.to_logical::<f64>(scale_factor);
```
amrbashir, line 3559: on Windows, we can get the height of titlebar using `GetThemeSysSize(SM_CYSIZE) + GetThemeSysSize(SM_CXPADDEDBORDER) * 2` and subtract it from the Y position returned from `calculate_window_center_position`

amrbashir (changes requested):

core/tauri-runtime-wry/src/lib.rs
Legend-Master: I'll take a look

core/tauri-runtime-wry/src/lib.rs
Legend-Master: I don't have a Mac, would you mind doing that part?

core/tauri-runtime-wry/src/lib.rs
Legend-Master: With `AdjustWindowRect`, I think we can even calculate the menu bar size but menu is behind private `platform_specific` in `WindowBuilder` from `tao`, so maybe not

core/tauri-runtime-wry/src/lib.rs
amrbashir: yeah sure

core/tauri-runtime-wry/src/lib.rs
amrbashir: If you want to use `AdjustWindowRect`, then treat menu as always `false` when creating the window because the menu is added later after the window is created using `muda` crate.

amrbashir (approved):

Issue #3310, opened 2022-01-30
[feat] zoom in and out functionality (ctrl/ctrl+shift + "=" and ctrl/ctrl+shift+ "-")


pinnick:
### Describe the problem

Browsers such as edge and chrome have a zoom in and out functionality with hotkeys mentioned in the title, and I would like to include that in my application with tauri, but the hotkeys aren't working for me.

### Describe the solution you'd like

add the hotkey support

FabianLars:
tracking issue at https://github.com/tauri-apps/wry/issues/388

martpie:
It would seem the upstream has been fixed, I guess now it's just about extending the `MenuItem` enum and implementing what has been done on `wry` right?

I may help with that if this is begginer-friendly enough

keiya01:
I think this issue is related to https://github.com/tauri-apps/wry/issues/569 .
And it is work in progress in https://github.com/tauri-apps/wry/pull/574 .

candrewlee14:
Looks like the upstream issues this depends on have been resolved. What can I do to enable this?

amrbashir:
even though it landed in upstream, tauri still needs to expose it.

rajatkulkarni95:
Bumping this up. Any idea when we'd get this into the framework?

Looky1173:
Bumping this up again! This would be a welcome addition to Tauri, mainly for accessibility concerns.

salmenf:
Is there any workaround? I hope this gets added soon, would be great to have for accessibility.

mxismean:
<img width="929" alt="image" src="https://user-images.githubusercontent.com/44690729/217980767-30d1e592-cd63-4350-b9c2-ef91db935009.png">
Can tauri add a new `zoom` configuration item in the tauri.conf.json configuration file. @FabianLars 

This allows users to set the default zoom of the window

mxismean:
A year has passed. I want to know how the progress of this bug is?
When can TAURI support Zoom?
@FabianLars

abose:
In the meantime, there is a pure js based way to get Zoom working in Tauri.
 using `document.body.style.zoom = 1.5`  from JS worked flawlessly in Windows, mac, and Linux Taui 1.5 builds. So there is a pure browser-based way to do this for now as its support in both Chromium (Tauri windows) and web-kit(Tauri mac/linux) (But not Firefox.) 

[...]

PR #9352, opened 2024-04-03
feat(windows): Expose webview2 zoom hotkeys from wry

Legend-Master:
Reference: #3310

### DIFF
--- a/core/tauri-utils/src/config.rs
+++ b/core/tauri-utils/src/config.rs
@@ -1276,11 +1276,7 @@ pub struct WindowConfig {
   ///
   /// - **macOS**: Requires the `macos-proxy` feature flag and only compiles for macOS 14+.
   pub proxy_url: Option<Url>,
-  /// Whether page zooming by hotkeys is enabled
-  ///
-  /// ## Platform-specific
-  ///
-  /// **macOS / Linux / Android / iOS**: Unsupported
+  /// Whether page zooming by hotkeys is enabled **Windows Only**
   #[serde(default)]
   pub zoom_hotkeys_enabled: bool,
 }
--- a/core/tauri-utils/src/config.rs
+++ b/core/tauri-utils/src/config.rs
@@ -1276,6 +1276,13 @@ pub struct WindowConfig {
   ///
   /// - **macOS**: Requires the `macos-proxy` feature flag and only compiles for macOS 14+.
   pub proxy_url: Option<Url>,
+  /// Whether page zooming by hotkeys is enabled
+  ///
+  /// ## Platform-specific
+  ///
+  /// **macOS / Linux / Android / iOS**: Unsupported
+  #[serde(default)]
+  pub zoom_hotkeys_enabled: bool,
 }
 
 impl Default for WindowConfig {
@@ -1321,6 +1328,7 @@ impl Default for WindowConfig {
       incognito: false,
       parent: None,
       proxy_url: None,
+      zoom_hotkeys_enabled: false,
     }
   }
 }
@@ -2242,6 +2250,7 @@ mod build {
       let window_effects = opt_lit(self.window_effects.as_ref());
       let incognito = self.incognito;
       let parent = opt_str_lit(self.parent.as_ref());
+      let zoom_hotkeys_enabled = self.zoom_hotkeys_enabled;
 
       literal_struct!(
         tokens,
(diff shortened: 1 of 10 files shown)


core/tauri-utils/src/config.rs
```diff
@@ -1276,6 +1276,13 @@ pub struct WindowConfig {
   ///
   /// - **macOS**: Requires the `macos-proxy` feature flag and only compiles for macOS 14+.
   pub proxy_url: Option<Url>,
+  /// Whether page zooming by hotkeys is enabled
+  ///
+  /// ## Platform-specific
+  ///
+  /// **macOS / Linux / Android / iOS**: Unsupported
```
```suggestion
  /// Whether page zooming by hotkeys is enabled **Windows Only**
```

core/tauri/src/webview/mod.rs
```diff
@@ -775,6 +775,17 @@ fn main() {
     self.webview_attributes.auto_resize = true;
     self
   }
+
+  /// Whether page zooming by hotkeys is enabled
+  ///
+  /// ## Platform-specific
+  ///
+  /// **macOS / Linux / Android / iOS**: Unsupported
```
amrbashir:
```suggestion
  /// Whether page zooming by hotkeys is enabled **Windows Only**
```

core/tauri/src/webview/webview_window.rs
```diff
@@ -838,6 +838,17 @@ fn main() {
     self.webview_builder = self.webview_builder.proxy_url(url);
     self
   }
+
+  /// Whether page zooming by hotkeys is enabled
+  ///
+  /// ## Platform-specific
+  ///
+  /// **macOS / Linux / Android / iOS**: Unsupported
```
amrbashir:
```suggestion
  /// Whether page zooming by hotkeys is enabled **Windows only**
```

tooling/api/src/webview.ts
```diff
@@ -648,6 +648,14 @@ interface WebviewOptions {
    * - **macOS**: Requires the `macos-proxy` feature flag and only compiles for macOS 14+.
    * */
   proxyUrl?: string
+  /**
+   * Whether page zooming by hotkeys or gestures is enabled
+   *
+   * ## Platform-specific
+   *
+   * - **macOS / Linux / Android / iOS**: Unsupported
```
amrbashir:
```suggestion
   * Whether page zooming by hotkeys or gestures is enabled **Windows Only**
```

amrbashir (changes requested):
Just a few documentation nitpicks. You will need to regenerate the schema file.

lucasfernog:
Can we use [WebView::zoom](https://docs.rs/wry/latest/wry/struct.WebView.html#method.zoom) to manually implement this feature on macOS/Linux?

amrbashir (approved):

amrbashir:
This PR is for only allowing users to enable the builtin one on Webview2, we need another PR to expose [WebView::zoom](https://docs.rs/wry/latest/wry/struct.WebView.html#method.zoom) and users can create a UI for it themselves.

amrbashir:
hmm actually this doesn't enable any UI, just the hotkeys, maybe we can indeed emulate the behavior on other platforms and also expose [WebView::zoom](https://docs.rs/wry/latest/wry/struct.WebView.html#method.zoom) at the same time

@Legend-Master would you like to add that too in this PR?

Legend-Master:
I think it's fine to expose the webview2 one first in this pr (it's in the webview2 api but we don't have a way to use it), and expose zoom function and polyfill for this functionality in another pr

If we want to put them in a single pr, I can take a look tomorrow, but I'm not familiar with WebKit zoom controls (is it the same as Chromium?)

And emulating it would be injecting some JavaScript to the webview?

amrbashir:
> If we want to put them in a single pr, I can take a look tomorrow

let's make them in another PR.

> but I'm not familiar with WebKit zoom controls (is it the same as Chromium?)

Yeah should be the same keys, `ctrl +/-`.

> And emulating it would be injecting some JavaScript to the webview?

yeah, take a look at https://github.com/tauri-apps/tauri/blob/36b4c12497fbe636066f4848c6877b3ab6cc892e/core/tauri/src/webview/scripts/toggle-devtools.js

Legend-Master:
Seems like webview2's zoom popup is disabled without a way to getting it back https://github.com/MicrosoftEdge/WebView2Feedback/issues/95#issuecomment-591590945, also the zoom factor is not persisted across page load (reset to 100% on reload/restart), pretty sad that webview2's zoom control is a downgraded version of the browser zoom control 😥

I'll take a look at exposing `zoom` function now

PR #9375, opened 2024-04-04
refactor(cli&bundler): avoid renaming main executable and preserve cargo name

### LINKED ISSUE #8109: [feat] displayName for package

### LINKED ISSUE #8349: [bug] Tauri unwantedly converts PascalCase package product name to kebab-case

### LINKED ISSUE #9221: [feat] firendly error output when make nsis bundle on linux but got (os error 2)
### Describe the problem

I am using `tauri cli 1.5.11`, i always get `Error failed to bundle project: No such file or directory (os error 2)` when i build nsis bundle on linux, i don't know why, then i found a issue #6743 about this error, so i think my tauri version is too low, but after i upgraded to `2.0.0-beta.9`, this error still exist.

I am so confused, so i debug the tauri cli line by line, i found my binary file name is not nsis needs, just because my `package.productionName` is `BDSM` but nsis needs `bdsm`, the error occoured at:
https://github.com/tauri-apps/tauri/blob/d349558abbefc78baf98f7837e5237f370758931/tooling/bundler/src/bundle/windows/nsis.rs#L636

Here is the error log:
```
warning: `bdsm` (bin "bdsm") generated 4 warnings
    Finished release [optimized] target(s) in 4m 08s
    Warn Cross-platform compilation is experimental and does not support all features. Please use a matching host system for full compatibility.
    Info Verifying NSIS package
    Downloading https://github.com/tauri-apps/binary-releases/releases/download/nsis-plugins-v0/NSIS-ApplicationID.zip
    Info extracting NSIS ApplicationID plugin
    Downloading https://github.com/tauri-apps/nsis-tauri-utils/releases/download/nsis_tauri_utils-v0.2.2/nsis_tauri_utils.dll
    Info validating hash
    Info Target: x64
    Info Code signing is currently only supported on Windows hosts, skipping...
    Error failed to bundle project: `No such file or directory (os error 2)`
```

I think, can we provide which file name caused `No such file or directory` like this:
```
cannot get metadata for /app/src-tauri/target/x86_64-pc-windows-msvc/release/bdsm.exe: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

### Alternatives considered

_No response_

### Additional context

_No response_

amrbashir:
closes #8109
closes #8349


amrbashir:
cc @FabianLars this will affect tauri-action, can you please give this a test

FabianLars:
> cc @FabianLars this will affect tauri-action, can you please give this a test

We already have logic in place to read the cargo.toml file so this should be an easy change. I will open a pr once this one here is merged (or earlier if i'm bored lol)

nekename:
Hi,

I'm close to releasing a stable version of my app, and I would like to have my app's proper name in releases :P (#8349)

Could this be merged ASAP?

Thanks

FabianLars:
We can (almost) never promise when something gets merged, but you can always use tauri from git if you can't wait for us :) <https://tauri.app/v1/guides/faq#how-can-i-use-unpublished-tauri-changes> (just replace dev/1.x with this PR's branch)

lucasfernog:
Pushed a small change to use the product name for the installers. I'm not sure about the Linux ones though - maybe we keep using the binary name? @FabianLars @amrbashir

amrbashir:
It is fine for the installers, not sure about how it will affect linux bundle as well but should be fine to change as well, even macOS if possible. It is just a filename

lucasfernog (approved):

FabianLars:
@olivierlemasle Sorry for the random ping but since you added rpm support, do you think applying this change to the rpm bundle is fine too? For reference, this is the output we have after this PR (deb changed, rpm not):
grafik

amrbashir:
I think we can rename it manually if the underlying library doesn't support using a custom file name

FabianLars:
I didn't look at any code, i assumed we are currently setting the output name ourselves tbh, just wasn't sure about conventions and maybe even issues with dnf. Conventions probably don't matter because afaik the new .deb naming scheme also breaks with conventions(?). I think i'd prefer the consistency this time idk

amrbashir:
Looks like we do, I will open a PR to change that

bukowa:
Trying to find out what's the new way to change application binary name correctly:
-  so all bundlers etc. work fine
- without using package.name as I want to publish it under different name as a crate

We can use:

```toml
[package]
name="my-crate-name"
default-run= "my-application-name"

[[bin]]
name = "my-application-name"
path = "src/main.rs"
```

Is this the way?

bukowa:
> Trying to find out what's the new way to change application binary name correctly:
> 
> * so all bundlers etc. work fine
> * without using package.name as I want to publish it under different name as a crate
> 
> We can use:
> 
> ```toml
> [package]
> name="my-crate-name"
> default-run= "my-application-name"
> 
> [[bin]]
> name = "my-application-name"
> path = "src/main.rs"
> ```
> 
> Is this the way?

I think this is problematic, cuz of https://github.com/tauri-apps/tauri/blob/dev/tooling/cli/src/interface/rust.rs#L846

PR #9415, opened 2024-04-09
fix(runtime-wry): don't hide/show webview when showing/hiding parent

### LINKED ISSUE #9393: [bug] v2 A white screen appears when calling show()
### Describe the bug

When I updated to the latest v2 version dependency, I found that when my application called show(), an inevitable white screen appeared.

new dependencies: https://github.com/Borber/Tran

https://github.com/tauri-apps/tauri/assets/30563826/d5b619bc-5da0-46b7-b927-875ebaf9d443

old dependencies: https://github.com/Borber/Tran/tree/old_dependence

https://github.com/tauri-apps/tauri/assets/30563826/641d0fac-f9d9-47ad-949c-a319c3b8efc2

Their code is exactly the same, but the dependency versions are different

### Reproduction

Compile the two codes from different branches above

### Expected behavior

Don’t show a white screen

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.19044 X64
    ✔ WebView2: 123.0.2420.65
    ✔ MSVC: Visual Studio 生成工具 2022
    ✔ rustc: 1.79.0-nightly (88c2f4f5f 2024-04-02)
    ✔ cargo: 1.79.0-nightly (a59aba136 2024-03-28)
    ✔ Rust toolchain: nightly-x86_64-pc-windows-msvc (default)
    - node: 20.11.0
    - pnpm: 8.15.6
    - yarn: 1.22.22
    - npm: 10.5.1

[-] Packages
    - tauri [RUST]: 2.0.0-beta.13
    - tauri-build [RUST]: 2.0.0-beta.10
    - wry [RUST]: 0.37.0
    - tao [RUST]: 0.26.2
    - @tauri-apps/api [NPM]: 2.0.0-beta.7
    - @tauri-apps/cli [NPM]: 2.0.0-beta.12

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:5046/
    - framework: SolidJS
    - bundler: Vite
```

### Stack trace

_No response_

### Additional context

Other dependencies basically have nothing to do with windows, so I can only think that it is a bug introduced when Tauri was updated.

amrbashir:
This fixes a regression introduced in #9246 where previously users would wait for `DOMContentLoaded` before showing the window to avoid having a white flash, also introduce a regression for apps that is hidden by default and is shown on global shortcuts or some other action.

Also accoding to the MSDN docs for webview2's `IsVisible` they only recommend to change the webview visibility on when un/minimizing.

closes #9393
ref: https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2controller?view=webview2-winrt-1.0.2365.46#isvisible

### DIFF
--- a/.changes/wry-visiblit-change-regression.md
+++ b/.changes/wry-visiblit-change-regression.md
@@ -1,0 +1,6 @@
+---
+"tauri-runtime-wry": "patch"
+---
+
+Fix regression on Windows, where calling `window.show` would flash a white screen before showing the webview.
+
--- a/core/tauri-runtime-wry/src/lib.rs
+++ b/core/tauri-runtime-wry/src/lib.rs
@@ -2699,16 +2699,8 @@ fn handle_user_message<T: UserEvent>(
           WindowMessage::Unmaximize => window.set_maximized(false),
           WindowMessage::Minimize => window.set_minimized(true),
           WindowMessage::Unminimize => window.set_minimized(false),
-          WindowMessage::Show => {
-            window.set_visible(true);
-            #[cfg(windows)]
-            let _ = set_webview_visibility(&webviews, !window.is_minimized());
-          }
-          WindowMessage::Hide => {
-            window.set_visible(false);
-            #[cfg(windows)]
-            let _ = set_webview_visibility(&webviews, false);
-          }
+          WindowMessage::Show => window.set_visible(true),
+          WindowMessage::Hide => window.set_visible(false),
           WindowMessage::Close => {
             panic!("cannot handle `WindowMessage::Close` on the main thread")
           }
@@ -3380,10 +3372,10 @@ fn handle_event_loop<T: UserEvent>(
                     log::error!("failed to autoresize webview: {e}");
                   }
                 }
+
+                #[cfg(windows)]
+                let _ = unsafe { webview.controller().SetIsVisible(!window.is_minimized()) };
               }
-              #[cfg(windows)]
-              let _ =
-                set_webview_visibility(&webviews, window.is_visible() && !window.is_minimized());
             }
           }
           _ => {}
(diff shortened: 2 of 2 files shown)


Legend-Master:
I still think we should `SetIsVisible` on show and hide (similar to https://www.electronjs.org/docs/latest/api/browser-window#page-visibility), I'll take a look to see if I can understand the white screen flashing

Legend-Master:
Found the problem, we just need to swap the order of `set_webview_visibility` and `window.set_visible`

amrbashir:
> Found the problem, we just need to swap the order of `set_webview_visibility` and `window.set_visible`

tried that and it was unreliable, also tried delaying `window.set_visible` for `10ms` and didn't work either.

Legend-Master:
Weird, it was fine moment ago, now I see white screen flashing on maximize/minimize as well, seems quite random, I'll take a closer look

Legend-Master:
I can reproduce the no content flashing on restore sometimes and on show most of the times, I think we should revert #9246 entirely first and try to re-land it when we find a way to avoid it

> also tried delaying window.set_visible for 10ms and didn't work either.

I tried it with 20ms, and that seemed to solve the problem, but I doubt it be a good idea though

I didn't notice this problem in the original pr because I was using the hello world example with white background, and `Welcome to Tauri!` flashing was not that noticeable

I think https://github.com/MicrosoftEdge/WebView2Feedback/issues/1412 maybe related

Legend-Master:
I had a pretty wacky idea, we'll be having a JavaScript script running and reporting visibility changes and we block restore/show until we get the event report from JavaScript side 🤣

PR #9420, opened 2024-04-09
feat(core/windows): Convert UNC paths to simple paths in JS apis.

### LINKED ISSUE #9407: [feat] Remove UNC prefix from paths returned in public apis
### Describe the problem

Currently our APIs (most of them moved to plugins i guess) still return UNC paths on Windows. For those unaware, those paths start with `\\?\`.

Most ecosystems seem to have problems with that, this includes the js frontend but also a multitude of sidecars. We've also already received reports about things built into Windows not being able to handle those.

use dunce everywhere to safely remove the prefix. I doubt that this will cause any issues but even if, it'd be still less than the unc paths cause lol

### Alternatives considered

do nothing and keep blaming the other languages' ecosystems for not being able to handle unc paths.

### Additional context

_No response_

FabianLars:
partially fixes #9407 

Not much in here i guess. That's all i found looking at the plugins / searching for various formats of the `#[command]` macro.

### DIFF
--- a/.changes/api-simplify-unc-paths.md
+++ b/.changes/api-simplify-unc-paths.md
@@ -1,0 +1,5 @@
+---
+tauri: patch:enhance
+---
+
+Tauri's built-in commands for the JS api will now return simplified paths on Windows, removing the `\\?\` prefix.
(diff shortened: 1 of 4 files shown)


FabianLars:
I forgot to look at events brb

FabianLars:
i can't see any events needing adjustments in this repo :)

amrbashir (approved):

PR #9211, opened 2024-03-18
fix(core): Announce new webviews and windows

FabianLars:
fixes https://github.com/tauri-apps/tauri/issues/9200
fixes https://github.com/tauri-apps/tauri/issues/8144

PR #9429, opened 2024-04-09
fix(core): deadlock on window create

lucasfernog:
The change from #9211 causes a deadlock on Webview::eval when the tracing feature is enabled (because on this case we wait for the eval to complete to keep the tracing span information).

### DIFF
--- a/core/tauri/src/window/mod.rs
+++ b/core/tauri/src/window/mod.rs
@@ -426,7 +426,7 @@ tauri::Builder::default()
 
     let app_manager = self.manager.manager_owned();
     let window_label = window.label().to_string();
-    // run on the main thread to fix deadlock on webview.eval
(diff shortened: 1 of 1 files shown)


core/tauri/src/window/mod.rs
```diff
@@ -424,17 +424,22 @@ tauri::Builder::default()
...
-
-    app_manager.emit(
-      "tauri://window-created",
-      Some(crate::webview::CreatedEvent {
-        label: window.label().into(),
-      }),
-    )?;
+    let app_manager = self.manager.manager_owned();
+    let window_label = window.label().to_string();
+    // run on the main thread to fix deadlock on webview.eval
```
FabianLars:
```suggestion
    // run on the main thread to fix a deadlock on webview.eval if the tracing feature is enabled
```

FabianLars (approved):

PR #9435, opened 2024-04-10
feat: adds rust-toolchain.toml

simonhyll:
> In relation to the CVE about Rust it's about time we pulled the trigger on adding rust-toolchain.toml in project templates. The reasoning behind this is pretty simple, it's good practice in projects in general to have one to enforce all developers to use the same Rust version, but more importantly it makes it very easy for projects to update their toolchain version when things like this happens. Furthermore it makes it more relevant for us to develop a tauri audit --fix command that checks the version in rust-toolchain.toml and recommends updating it.

Reflects the proposed update to create-tauri-app: https://github.com/tauri-apps/create-tauri-app/pull/664


FabianLars:
Does the toolchain file also support a plain "stable" channel.

I already hear the screams when we raise our msrv too high x)

lucasfernog (approved):

lucasfernog:
> Does the toolchain file also support a plain "stable" channel.
> 
> I already hear the screams when we raise our msrv too high x)

@FabianLars yeah: https://rust-lang.github.io/rustup/overrides.html#channel

FabianLars:
then i'd be heavily in favor of using that instead of a specific number. imo we have too many users that have no idea about cargo/rustup etc and don't want to learn about it either. For those who do know about it, they can just modify/remove the file...

lucasfernog:
maybe i'm missing the point.. isn't the idea that we can automate updating the toolchain version easily?

lucasfernog:
using `stable` is the same as not having the file at all I believe

FabianLars:
> using stable is the same as not having the file at all I believe

Ah yeah maybe, hoped that it'd just auto update lol.

> maybe i'm missing the point.. isn't the idea that we can automate updating the toolchain version easily?

Then we should wait with this PR until we have the automation in place.

PR #9436, opened 2024-04-10
Replace unmaintained dependency `zip` with `zip_next`

Pr0methean:
`zip` hasn't been updated in almost a year, and it has known bugs that can cause panics when trying to open an invalid ZIP file, or an encrypted file without the password. This PR replaces it with `zip_next`, a fork I actively maintain.


ahqsoftwares:
https://github.com/Pr0methean/zip-next-wrapper
This repository has been archived by the owner on Apr 21, 2024. It is now read-only.

PR #9457, opened 2024-04-13
fix(cli/migrate): tolerate non-UTF-8 in migration

kornelski:
When I ran `cargo tauri migrate` I got only:

> Error Stream did not contain valid UTF-8

which is a completely unhelpful dead-end. 

* I've made stdout/stderr captures tolerate non-UTF-8 text.
* I've added context to migrator's errors that help locate the root cause.
* I've made migrator skip and warn about non-UTF-8 JavaScript files instead of aborting the entire operation.

### DIFF
--- a/.changes/cli-migrate-non-utf8.md
+++ b/.changes/cli-migrate-non-utf8.md
@@ -1,0 +1,6 @@
+---
+'tauri-cli': 'patch:bug'
+'@tauri-apps/cli': 'patch:bug'
+---
+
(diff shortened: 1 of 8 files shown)


amrbashir:
Instead of skipping non-utf8 files, I changed it to replace bytes instead, tested by appending invalid utf8 bytes to a JS file `echo -n -e '\x00\x9F\x92\x96' >> src/App.jsx`, let me know if I missed anything

amrbashir (approved):

amrbashir (approved):

PR #9459, opened 2024-04-13
feat(cli): show full expected path of `frontendDist` if not found

kornelski:
It was unclear to me what the `frontendDist` path is relative to, so Tauri reporting it can't find it doesn't tell me much. I've added absolute path to the error, which clarifies where the directory is supposed to be exactly.


FabianLars:
Thanks for the PR! Could you also please add a changefile similar to https://github.com/tauri-apps/tauri/blob/dev/.changes/cli-empty-responses.md please?

amrbashir (approved):

PR #9486, opened 2024-04-17
fix(api): add type definitions to package.json exports

ankoh:
This commit includes the typescript type definitions in the package exports of @tauri-apps/api.


ankoh:
Without this commit, tsc will just fail to resolve the type definitions in some cases (for me it occurred with NodeNext module resolution through ts-jest)

lucasfernog (dismissed):

amrbashir (approved):

PR #9526, opened 2024-04-21
feat(nsis): add an option to opt out tauri utils plugin

Legend-Master:
Add an option to replace nsis util plugin with built-in ones, this can reduce the final bundle size by ~0.8 MB

Downsides including:

- Can't compare semver like `2.0.0-alpha.1`
- No Windows 7 support when using [downloadBootstrapper](https://tauri.app/v1/guides/building/windows/#downloaded-bootstrapper) option due to Windows 7 ~~doesn't have curl installed by default~~ doesn't have tls 1.2 enabled by default (this option is not recommended for Windows 7 anyway)

### DIFF
--- a/core/tauri-utils/src/config.rs
+++ b/core/tauri-utils/src/config.rs
@@ -744,6 +744,16 @@ pub struct NsisConfig {
   ///
   /// See <https://nsis.sourceforge.io/Reference/SetCompressor>
   pub compression: Option<NsisCompression>,
+  /// Whether include and use tauri nsis plugin for
+  /// compare semver, check and terminate running app, and download webview2 bootstrap,
+  /// this will increase the installer size by around 0.8 MB
+  /// when disabled, uses tools provided by the OS and nsis,
+  /// this option has some downsides:
+  /// - can't compare semver like 2.0.0-alpha.1
+  /// - no Windows 7 support when using webviewInstallMode downloadBootstrapper
+  /// default to true
+  #[serde(default = "default_true")]
+  pub use_tauri_plugin: bool,
 }
 
 /// Install Modes for the NSIS installer.
--- a/core/tauri-utils/src/config.rs
+++ b/core/tauri-utils/src/config.rs
@@ -744,14 +744,16 @@ pub struct NsisConfig {
   ///
   /// See <https://nsis.sourceforge.io/Reference/SetCompressor>
   pub compression: Option<NsisCompression>,
-  /// Whether include and use tauri nsis plugin for
-  /// compare semver, check and terminate running app, and download webview2 bootstrap,
-  /// this will increase the installer size by around 0.8 MB
-  /// when disabled, uses tools provided by the OS and nsis,
-  /// this option has some downsides:
-  /// - can't compare semver like 2.0.0-alpha.1
-  /// - no Windows 7 support when using webviewInstallMode downloadBootstrapper
-  /// default to true
+  /// Whether to includede and use nsis-tauri-utils.dll plugin to
+  /// compare semver, check and terminate running app, and download webview2 bootstrapper.
+  /// This will increase the installer size by around 0.8 MB.
+  ///
+  /// If disabled, the installer will use alternative tools provided by the OS
+  /// and built-in NSIS plugins but this comes with downsides:
+  /// - Can't compare semver with pre-release labels like `2.0.0-alpha.1`
+  /// - Using `webviewInstallMode: downloadBootstrapper` may fail on Windows 7.
+  ///
+  /// Defaults to `true`.
   #[serde(default = "default_true")]
   pub use_tauri_plugin: bool,
 }
--- a/tooling/bundler/src/bundle/settings.rs
+++ b/tooling/bundler/src/bundle/settings.rs
@@ -415,14 +415,16 @@ pub struct NsisSettings {
   pub display_language_selector: bool,
   /// Set compression algorithm used to compress files in the installer.
   pub compression: Option<NsisCompression>,
-  /// Whether include and use tauri nsis plugin for
-  /// compare semver, check and terminate running app, and download webview2 bootstrap,
-  /// this will increase the installer size by around 0.8 MB
-  /// when disabled, uses tools provided by the OS and nsis,
-  /// this option has some downsides:
-  /// - can't compare semver like 2.0.0-alpha.1
-  /// - no Windows 7 support when using webviewInstallMode downloadBootstrapper
-  /// default to true
+  /// Whether to includede and use nsis-tauri-utils.dll plugin to
+  /// compare semver, check and terminate running app, and download webview2 bootstrapper.
+  /// This will increase the installer size by around 0.8 MB.
+  ///
+  /// If disabled, the installer will use alternative tools provided by the OS
+  /// and built-in NSIS plugins but this comes with downsides:
+  /// - Can't compare semver with pre-release labels like `2.0.0-alpha.1`
+  /// - Using `webviewInstallMode: downloadBootstrapper` may fail on Windows 7.
+  ///
+  /// Defaults to `true`.
   pub use_tauri_plugin: bool,
 }
 
--- a/tooling/bundler/src/bundle/settings.rs
+++ b/tooling/bundler/src/bundle/settings.rs
@@ -415,6 +415,15 @@ pub struct NsisSettings {
   pub display_language_selector: bool,
   /// Set compression algorithm used to compress files in the installer.
   pub compression: Option<NsisCompression>,
+  /// Whether include and use tauri nsis plugin for
+  /// compare semver, check and terminate running app, and download webview2 bootstrap,
+  /// this will increase the installer size by around 0.8 MB
+  /// when disabled, uses tools provided by the OS and nsis,
+  /// this option has some downsides:
+  /// - can't compare semver like 2.0.0-alpha.1
+  /// - no Windows 7 support when using webviewInstallMode downloadBootstrapper
+  /// default to true
+  pub use_tauri_plugin: bool,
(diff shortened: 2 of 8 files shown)


Legend-Master:
I need to take a look at the CheckIfAppIsRunning part, it's not actually detecting the app right now

Legend-Master:
> I need to take a look at the CheckIfAppIsRunning part, it's not actually detecting the app right now

Should be fixed now

core/tauri-utils/src/config.rs
```diff
@@ -744,6 +744,16 @@ pub struct NsisConfig {
...
   /// See <https://nsis.sourceforge.io/Reference/SetCompressor>
   pub compression: Option<NsisCompression>,
+  /// Whether include and use tauri nsis plugin for
+  /// compare semver, check and terminate running app, and download webview2 bootstrap,
+  /// this will increase the installer size by around 0.8 MB
+  /// when disabled, uses tools provided by the OS and nsis,
+  /// this option has some downsides:
+  /// - can't compare semver like 2.0.0-alpha.1
+  /// - no Windows 7 support when using webviewInstallMode downloadBootstrapper
+  /// default to true
```
```suggestion
  /// Whether to includede and use nsis-tauri-utils.dll plugin to
  /// compare semver, check and terminate running app, and download webview2 bootstrapper.
  /// This will increase the installer size by around 0.8 MB.
  ///
  /// If disabled, the installer will use alternative tools provided by the OS
  /// and built-in NSIS plugins but this comes with downsides:
  /// - Can't compare semver with pre-release labels like `2.0.0-alpha.1`
  /// - Using `webviewInstallMode: downloadBootstrapper` may fail on Windows 7.
  ///
  /// Defaults to `true`.
```

tooling/bundler/src/bundle/settings.rs
```diff
@@ -415,6 +415,15 @@ pub struct NsisSettings {
...
   /// Set compression algorithm used to compress files in the installer.
   pub compression: Option<NsisCompression>,
+  /// Whether include and use tauri nsis plugin for
+  /// compare semver, check and terminate running app, and download webview2 bootstrap,
+  /// this will increase the installer size by around 0.8 MB
+  /// when disabled, uses tools provided by the OS and nsis,
+  /// this option has some downsides:
+  /// - can't compare semver like 2.0.0-alpha.1
+  /// - no Windows 7 support when using webviewInstallMode downloadBootstrapper
+  /// default to true
```
```suggestion
  /// Whether to includede and use nsis-tauri-utils.dll plugin to
  /// compare semver, check and terminate running app, and download webview2 bootstrapper.
  /// This will increase the installer size by around 0.8 MB.
  ///
  /// If disabled, the installer will use alternative tools provided by the OS
  /// and built-in NSIS plugins but this comes with downsides:
  /// - Can't compare semver with pre-release labels like `2.0.0-alpha.1`
  /// - Using `webviewInstallMode: downloadBootstrapper` may fail on Windows 7.
  ///
  /// Defaults to `true`.
```

tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -451,7 +461,11 @@ Section WebView2
   !if "${INSTALLWEBVIEW2MODE}" == "downloadBootstrapper"
     Delete "$TEMP\MicrosoftEdgeWebview2Setup.exe"
     DetailPrint "$(webview2Downloading)"
-    nsis_tauri_utils::download "https://go.microsoft.com/fwlink/p/?LinkId=2124703" "$TEMP\MicrosoftEdgeWebview2Setup.exe"
+    !if "${USETAURIPLUGIN}" == "true"
+      nsis_tauri_utils::download "https://go.microsoft.com/fwlink/p/?LinkId=2124703" "$TEMP\MicrosoftEdgeWebview2Setup.exe"
+    !else
+      nsExec::Exec 'curl --location "https://go.microsoft.com/fwlink/p/?LinkId=2124703" --output "$TEMP\MicrosoftEdgeWebview2Setup.exe"'
```
amrbashir: curl is only included in newer windows 10 versions, we should either use powershell to call `InvokeWebRequest` or at least have it as a fallback.

amrbashir (changes requested):
I still would like to explore reducing the size of our semverCompare and process utils, I believe we can probably make these two plugins really small that we wouldn't need to use the alternatives.

tooling/bundler/src/bundle/windows/templates/installer.nsi
Legend-Master: curl is included at the end of the 2017, 2 years after Windows 10 comes out, I'm not sure how many people would use a version in that gap

Anyway, I'll take a look to use `InvokeWebRequest` instead

Legend-Master:
> I still would like to explore reducing the size of our semverCompare and process utils, I believe we can probably make these two plugins really small that we wouldn't need to use the alternatives.

I compiled it without the download one, it's around 200 KB, not too bad but definitely something we could improve

amrbashir:
> > I still would like to explore reducing the size of our semverCompare and process utils, I believe we can probably make these two plugins really small that we wouldn't need to use the alternatives.
> 
> I compiled it without the download one, it's around 200 KB, not too bad but definitely something we could improve

A 100kb of that are related to static CRT iirc

amrbashir:
Sorry for the long delay, I wanted to experiment a bit and see if we can avoid having this option.

I have been doing some [experiments](https://github.com/tauri-apps/nsis-tauri-utils/tree/feat/no-std) and I have managed to reduce the size of some of our nsis plugins:

|                  | nsis-semvercompare | nsis-process |  nsis-semvercompare & nsis-process combined in one DLL|
|------------------|--------------------|--------------|---------------------------------------------|
| dev              | 198kb              | 208kb        | 200kb                                       |
| no_std      | 35kb               | 48kb         | 53kb                                        |
| no_std (UPX) | 20kb               | 26kb         | 29kb                                        |

So I think we can keep using our `nsis-process` and `nsis-semvercompare` without compromising while keeping the size very small.


The remaining bit here is the `nsis-download` plugin which is the most bloated of them all and unfortunately can't be written in a `no_std` environment, at least without too much work. 

So I am proposing to remove it all together and use `powershell` to download the bootstrapper, this means it will fail on Windows7 and so we will need to recommend users to use `embedBootstrapper` when targeting Windows7.

If that's not desired, we can make this PR add an opt-out option specifically for the `nsis-download` plugin.

cc @lucasfernog @FabianLars

FabianLars:
I think requiring embedBootstrapper for windows 7 is fine. We shouldn't do more than the minimum for win7 at this point imo (and remove support for it in v3).

One thing though, i am fairlyyy sure i've had issues without our download plugin on win10 too so we should triple check it! We should also check if using powershell will open a terminal window again which tbh would be a blocker for me too.

amrbashir:
> One thing though, i am fairlyyy sure i've had issues without our download plugin on win10 too so we should triple check it! We should also check if using powershell will open a terminal window again which tbh would be a blocker for me too.

using `nsExec::Exec` with `PowerShell.exe -WindowStyle hidden` should not show it.

But the one thing I hate about this is that we lose the progress bar.

Legend-Master:
no std version seem to be promising, 50 KB is acceptable

To be honest, if we can't reduce the download plugin's size enough, there'll not be much benefit doing it this way, the bootstrap is only 1.5 MB

> using nsExec::Exec with PowerShell.exe -WindowStyle hidden should not show it.

I'm pretty sure `nsExec::Exec powershell -Command` alone creates a hidden window

FabianLars:
The problem i am thinking about is that when ms terminal is the default console host then the hidden window style doesn't work anymore. On win11 this isn't a problem as long as we're only downloading webview2 since all win11 installations should have that, but win10 added support for changing the default to ms terminal in later versions too and there we can't expect webview2 to be installed.

Legend-Master:
I'm pretty sure `nsExec::Exec powershell -Command` will create a hidden window even when you set windows terminal as default console host, I have tested this on my machine (windows 10) as least

amrbashir:
I have been tinkering over the weekend a bit and I managed to get the dlls even smaller

|                  | nsis-semvercompare | nsis-process |  nsis-semvercompare & nsis-process combined in one DLL|
|------------------|--------------------|--------------|---------------------------------------------|
| dev              | 198kb              | 208kb        | 200kb                                       |
| no_std      | 35kb               | 48kb         | 53kb                                        |
| no_std+upx| 20kb               | 26kb         | 29kb                                        |
| no_std/no_msvcrt      | 10kb               | 22kb         | 27kb                                        |
| no_std/no_msvcrt+upx| 7kb               | 13kb         | 16kb                                        |

I don't think I can get them smaller than this, so I will start looking into the `download` plugin

amrbashir:
I don't know how I missed this before but looks like nsis has a [built-in download plugin](https://nsis.sourceforge.io/Builtin_NSISdl_plug-in) that is able to download the webview2 installer which is also really small in size

amrbashir:
I apologize for initially asking you to open this PR then rejecting it. I have opened https://github.com/tauri-apps/nsis-tauri-utils/pull/26 for the small dll size

Legend-Master:
> I don't know how I missed this before but looks like nsis has a [built-in download plugin](https://nsis.sourceforge.io/Builtin_NSISdl_plug-in) that is able to download the webview2 installer which is also really small in size

Its first line says

> This built-in plugin allows you to download files via HTTP (but not **HTTPS**).

I did come across this but it says no HTTPS

> I apologize for initially asking you to open this PR then rejecting it.

It's fine, glad to accept a better solution 😃

FabianLars:
> I did come across this but it says no HTTPS

That's probably why it works on win7 because https will fail, though i am surprised a http request against a https url works, probably only a matter of time until this breaks so we should keep an eye on it.

PR #9527, opened 2024-04-21
refactor(nsis): use nsis's built-in com plugin instead of ApplicationID plugin

Legend-Master:
Use nsis's built-in com plugin instead of ApplicationID plugin, this reduces the installer size by 150-200 KB, and also fixes pinned shortcut not getting cleaned up on uninstall

### DIFF
--- a/tooling/bundler/src/bundle/windows/templates/installer.nsi
+++ b/tooling/bundler/src/bundle/windows/templates/installer.nsi
@@ -13,6 +13,8 @@ ManifestDPIAware true
 !include WordFunc.nsh
 !include "FileAssociation.nsh"
 !include "StrFunc.nsh"
+!include "Win\COM.nsh"
+!include "Win\Propkey.nsh"
 ${StrCase}
 ${StrLoc}
 
@@ -679,8 +681,9 @@ Section Uninstall
   {{/each}}
   RMDir "$INSTDIR"
 
-  WinShell::UninstAppUserModelId "${BUNDLEID}"
-  WinShell::UninstShortcut "$DESKTOP\${MAINBINARYNAME}.lnk"
+  !insertmacro DeleteAppUserModelId
+  !insertmacro UnpinShortcut "$SMPROGRAMS\$AppStartMenuFolder\${MAINBINARYNAME}.lnk"
+  !insertmacro UnpinShortcut "$DESKTOP\${MAINBINARYNAME}.lnk"
 
   ; Remove start menu shortcut
   !insertmacro MUI_STARTMENU_GETFOLDER Application $AppStartMenuFolder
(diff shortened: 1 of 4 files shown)


tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -679,6 +679,9 @@ Section Uninstall
   {{/each}}
   RMDir "$INSTDIR"
 
+  WinShell::UninstAppUserModelId "${BUNDLEID}"
+  WinShell::UninstShortcut "$DESKTOP\${MAINBINARYNAME}.lnk"
```
amrbashir: We already remove the desktop shortcut on Line 691

We shouldn't remove the shortcut on uninstall, otherwise when updating it will be removed and never re-created again. We should detect if we are updating and skip removing the shortcut.

amrbashir:
I believe this will require us to upload `WinShell` to our binary-releases repo but looking at https://nsis.sourceforge.io/WinShell_plug-in there seems to be an alternative way using NSIS built-in COM plugin, can we try using that instead?

Legend-Master:
I could take a look tomorrow, but I'm really getting tired of fighting nsi scripts 😂, this plugin is pretty small, and does the job, if we can take it I think we should

amrbashir:
> I could take a look tomorrow, but I'm really getting tired of fighting nsi scripts 😂

No problem, just let me know and I can fight it for you

tooling/bundler/src/bundle/windows/templates/installer.nsi
Legend-Master: This is what they did in their example, I don't know if we should do it or not, I can take a look tomorrow

I don't think uninstall runs on update, right?

Legend-Master: > We already remove the desktop shortcut on Line 691

Oh, was on mobile, forgot that this (`UninstShortcut`) is to unpin the shortcut

tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -722,11 +728,66 @@ FunctionEnd
...
 FunctionEnd
 
 Function CreateStartMenuShortcut
   CreateDirectory "$SMPROGRAMS\$AppStartMenuFolder"
   CreateShortcut "$SMPROGRAMS\$AppStartMenuFolder\${MAINBINARYNAME}.lnk" "$INSTDIR\${MAINBINARYNAME}.exe"
-  ApplicationID::Set "$SMPROGRAMS\$AppStartMenuFolder\${MAINBINARYNAME}.lnk" "${BUNDLEID}"
+  !insertmacro SetLnkAppUserModelId "$SMPROGRAMS\$AppStartMenuFolder\${MAINBINARYNAME}.lnk"
 FunctionEnd
+
+!macro SetLnkAppUserModelId shortcut
```
amrbashir, line 758: macros need to be defined before the place they are called

tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -722,11 +728,66 @@ FunctionEnd
...
+    ${IApplicationDestinations::SetAppID} $1 '("${BUNDLEID}")i.r0'
+    ${If} $0 >= 0
+      ${IApplicationDestinations::RemoveAllDestinations} $1 ''
+    ${EndIf}
+    ${IUnknown::Release} $1 ""
+  ${EndIf}
+!macroend
+
+; From https://stackoverflow.com/a/42816728/16993372
+!macro UnpinShortcut shortcut
```
amrbashir: ditto

tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -722,11 +728,66 @@ FunctionEnd
 
 Function CreateDesktopShortcut
   CreateShortcut "$DESKTOP\${MAINBINARYNAME}.lnk" "$INSTDIR\${MAINBINARYNAME}.exe"
-  ApplicationID::Set "$DESKTOP\${MAINBINARYNAME}.lnk" "${BUNDLEID}"
+  SetLnkAppUserModelId "$DESKTOP\${MAINBINARYNAME}.lnk"
```
amrbashir:
```suggestion
  !insertmacro SetLnkAppUserModelId "$DESKTOP\${MAINBINARYNAME}.lnk"
```

tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -722,11 +728,66 @@ FunctionEnd
...
+        ${IUnknown::Release} $2 ""
+        ${IPersistFile::Save} $1 '("${shortcut}",1)'
+      ${EndIf}
+      ${IUnknown::Release} $1 ""
+    ${EndIf}
+    ${IUnknown::Release} $0 ""
+  ${EndIf}
+!macroend
+
+!macro DeleteAppUserModelId
```
amrbashir: If I understand correctly, this removes the jump list and recent destinations from the pinned taskbar icon, any reason why we would want to remove those? I think we shouldn't call this.

amrbashir (changes requested):

tooling/bundler/src/bundle/windows/templates/installer.nsi
amrbashir: nvm, seems like firefox and other applications do the same, let's follow

amrbashir (approved):

PR #9553, opened 2024-04-23
fix(cli): pnpm detection on mobile init and xcode-script

lucasfernog:


amrbashir (approved):

aleclarson:
will this fix #8860 if I use pnpm?

PR #9568, opened 2024-04-25
ci: Remove nodejs14 test in cli.js publishing workflow

FabianLars:
The macos runners switched to arm and node 16 is the first to support arm.


lucasfernog (approved):

FabianLars:
i'm scared

PR #9588, opened 2024-04-27
fix(core): fix check temp path permission on mac os, fix #6256

### LINKED ISSUE #6256: [bug] Unhandled Promise Rejection for $TEMP: path not allowed on the configured scope
### Describe the bug

The problem occurs when opening a file from the `BaseDirectory.Temp` . Other directories work.
I tried different configurations, the only one that works is a wildcard.
```json
   "fs": {
        "readFile": true,
        "scope": [
          "$TEMP/test.txt",
          "$TEMP/*/**",
          "$TEMP/**",
          "$TEMP/*",
          "$TEMP/",
          "$TEMP",
          "**", // <- wildcard is the only one that works
      }
```
```typescript
 readBinaryFile(fileNameInputEl.value, { dir: BaseDirectory.Temp }).then(
      (data) => {
        console.log(data)
      }
    )
```
image

### Reproduction

**I created a repository for reproduction: https://github.com/gormlabenz/tauri-temp-bug**
1. Add `"$TEMP/**"` to scope
```json
 "fs": {
        "readFile": true,
        "scope": [
          "$TEMP/**",
        ]
      }
```
2. Open file from 'BaseDirectory.Temp'
```typescript
    readBinaryFile(filename, { dir: BaseDirectory.Temp }).then(
      (data) => {
        console.log(data)
      }
    )
```

### Expected behavior

Shouldn't fail when opening file from `BaseDirectory.Temp` 

### Platform and versions

Environment
  › OS: Mac OS 13.1.0 X64
  › Node.js: 16.15.0
  › npm: 9.3.1
  › pnpm: 7.27.0
  › yarn: 1.22.19
  › rustup: 1.25.1
  › rustc: 1.63.0
  › cargo: 1.63.0
  › Rust toolchain: stable-x86_64-apple-darwin 

Packages
  › @tauri-apps/cli [NPM]: 1.2.3

### Stack trace

```shell
[Error] Unhandled Promise Rejection: path not allowed on the configured scope: /var/folders/8s/xvw_7jh940d8n4cmn3wj5n8w0000gn/T/test.txt
	promiseEmptyOnRejected
	promiseReactionJob
```

### Additional context

_No response_

0x-jerry:
close #6256

### DIFF
--- a/.changes/check-permission-on-macos.md
+++ b/.changes/check-permission-on-macos.md
@@ -1,0 +1,5 @@
+---
+"tauri": "patch:bug"
+---
+
+Fix check temporary path permission on macos.
--- a/core/tauri/src/api/path.rs
+++ b/core/tauri/src/api/path.rs
@@ -293,7 +293,7 @@ pub fn resolve_path<P: AsRef<Path>>(
       BaseDirectory::App => app_config_dir(config),
       #[allow(deprecated)]
       BaseDirectory::Log => app_log_dir(config),
-      BaseDirectory::Temp => Some(temp_dir()),
+      BaseDirectory::Temp => temp_dir().canonicalize().ok(),
       BaseDirectory::AppConfig => app_config_dir(config),
       BaseDirectory::AppData => app_data_dir(config),
       BaseDirectory::AppLocalData => app_local_data_dir(config),
--- a/core/tauri/src/endpoints/operating_system.rs
+++ b/core/tauri/src/endpoints/operating_system.rs
@@ -42,7 +42,7 @@ impl Cmd {
   }
 
   fn tempdir<R: Runtime>(_context: InvokeContext<R>) -> super::Result<PathBuf> {
-    Ok(std::env::temp_dir())
+    Ok(std::env::temp_dir().canonicalize()?)
   }
 
   fn locale<R: Runtime>(_context: InvokeContext<R>) -> super::Result<Option<String>> {
--- a/core/tauri/src/scope/fs.rs
+++ b/core/tauri/src/scope/fs.rs
@@ -391,4 +391,27 @@ mod tests {
       assert!(scope.is_allowed("C:\\home\\tauri\\anyfile"));
     }
   }
+
+  #[cfg(unix)]
+  #[test]
+  fn check_temp_dir() {
+    use std::{
+      env::temp_dir,
+      fs::{remove_file, write},
+    };
+
+    let scope = new_scope();
+    scope.allow_directory(temp_dir().canonicalize().unwrap(), true).unwrap();
+
+    let test_temp_file = temp_dir().canonicalize().unwrap().join("tauri_test_file");
+    if test_temp_file.exists() {
+      remove_file(test_temp_file.clone()).unwrap();
+    }
+
+    assert!(scope.is_allowed(test_temp_file.clone()));
+
+    write(test_temp_file.clone(), ".").unwrap();
+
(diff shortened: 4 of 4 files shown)


tweidinger:
Will check this PR in the upcoming days but am wondering if https://github.com/tauri-apps/tauri/pull/9072/files#diff-903ee43f2eb846686e5baf219fc515ff00de27369cef1e81b37164de847d2e90 does fix the underlying issue, as it resolves symlinks now before checking. This is in `dev`, so not sure if it can/is/will be back ported and if other changes influence behavior.

0x-jerry:
@tweidinger I tested how path::is_symlink works, and seems it only checks the exact symbolic path, eg.

```rs
use std::{
    env,
    path::PathBuf,
};

fn main() {
    let p_var = PathBuf::from("/var");

    println!("{:?}", p_var.is_symlink()); // => true

    let p_var = env::temp_dir();

    println!("{:?}", p_var.is_symlink()); // => false
}

```
image

This means any subpath of a symbolic folder will return false, so https://github.com/tauri-apps/tauri/pull/9072/files#diff-903ee43f2eb846686e5baf219fc515ff00de27369cef1e81b37164de847d2e90 may not be able to fix subpath permission check.

lucasfernog:
The issue happens because the allowed path is actually a symlink so the full path does not match the path we're checking against (which is canonicalized so it resolves to the actual folder).

There is an easier way to fix this which is to canonicalize the temp dir when resolving the path variables.
basically changing https://github.com/tauri-apps/tauri/blob/44e3335da8c1a811a2ebe72f2036dfad2381da24/core/tauri/src/api/path.rs#L296 to `BaseDirectory::Temp => temp_dir().canonicalize().ok()`

it's probably more secure this way since we would still check canonicalized paths in the end, resolving symlinks and path traversals

cc @tweidinger

0x-jerry:
Thanks for the guidance, done!

0x-jerry:
> The issue happens because the allowed path is actually a symlink so the full path does not match the path we're checking against (which is canonicalized so it resolves to the actual folder).
> 
> There is an easier way to fix this which is to canonicalize the temp dir when resolving the path variables. basically changing
> 
> https://github.com/tauri-apps/tauri/blob/44e3335da8c1a811a2ebe72f2036dfad2381da24/core/tauri/src/api/path.rs#L296
> 
> to `BaseDirectory::Temp => temp_dir().canonicalize().ok()`
> it's probably more secure this way since we would still check canonicalized paths in the end, resolving symlinks and path traversals
> 
> cc @tweidinger

When the user creates a temporary file, the js code should be like this:

```js
import { fs, path } from '@tauri-apps/api'
import { tempdir } from '@tauri-apps/api/os'

const tempFile = await path.join(await tempdir(), 'not-exits-file')

await fs.writeFile(tempFile, 'some data')
```

it will use `os.tempdir()` to get the temporary path, when writing the file, the `tempFile` should not exist, so check permission will only check the path, not the canonicalized path, this will fail because `os.tempdir()` is not canonicalized, while the permission check is checking the canonicalized temporary path.

So, we may also change `os.tempdir()` API, and make it return the canonicalized temporary path.

tweidinger (dismissed):

tweidinger:
LGTM. The only blocker is that not all commits are signed @0x-jerry could you please amend signatures to previous commits or rewrite to a single commit? The history will be squashed into a single commit anyway.

0x-jerry:
Done!

lucasfernog (approved):

PR #9612, opened 2024-04-30
fix(windows): hide windows on cleanup

### LINKED ISSUE #9610: [bug] window white flashing on exit on Windows
To reproduce, launch api example from the repo, switch the app to dark mode (system preference doesn't matter here, light and dark mode results the same, at least on Windows 10), quit the app from tray icon menu

https://github.com/tauri-apps/tauri/assets/68118705/bef106c6-24c3-487d-9b09-881af2f77679

image

Version: current head 3f0805488506e013e15cfb3cc1cfc8c2c5f84bd2, api example

Legend-Master:
Fix #9610

I don't know if closing the window here will break anything, so I just used `hide`

### DIFF
--- a/core/tauri/src/app.rs
+++ b/core/tauri/src/app.rs
@@ -772,11 +772,15 @@ macro_rules! shared_app_impl {
         #[cfg(all(desktop, feature = "tray-icon"))]
         self.manager.tray.icons.lock().unwrap().clear();
         self.manager.resources_table().clear();
-        for (_, window) in self.manager.windows().iter() {
+        for (_, window) in self.manager.windows() {
           window.resources_table().clear();
         }
-        for (_, webview) in self.manager.webviews().iter() {
+        for (_, webview) in self.manager.webviews() {
           webview.resources_table().clear();
+        }
+        #[cfg(windows)]
+        for (_, window) in self.manager.windows() {
+          let _ = window.hide();
         }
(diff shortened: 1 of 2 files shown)


core/tauri/src/app.rs
```diff
@@ -772,12 +772,16 @@ macro_rules! shared_app_impl {
...
+        for (_, window) in self.manager.windows() {
           window.resources_table().clear();
         }
-        for (_, webview) in self.manager.webviews().iter() {
+        for (_, webview) in self.manager.webviews() {
           webview.resources_table().clear();
         }
+        #[cfg(windows)]
+        for (_, window) in self.manager.windows() {
+          let _ = window.hide();
```
amrbashir: how will this effect plugins like `window-state`?

amrbashir: can just be moved to the first loop without needing a new loop

Legend-Master: This should be run after all other exit event handlers so should be fine

> ```rust
> let event = on_event_loop_event(&app_handle, RuntimeRunEvent::Exit, &manager);
> callback(&app_handle, event);
> app_handle.cleanup_before_exit();
> ```

amrbashir (approved):

amrbashir (approved):

PR #9618, opened 2024-04-30
fix(core): fix incorrect `isTauri` return type

typed-sigterm:
A tiny patch to #9539: `() => unknown` to `() => boolean`

### DIFF
--- a/tooling/api/src/core.ts
+++ b/tooling/api/src/core.ts
@@ -239,7 +239,7 @@ export class Resource {
 }
 
 function isTauri() {
-  return 'isTauri' in window && window.isTauri
+  return 'isTauri' in window && !!window.isTauri
 }
 
 export type { InvokeArgs, InvokeOptions }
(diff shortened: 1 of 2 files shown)


tooling/api/src/core.ts
```diff
@@ -239,7 +239,7 @@ export class Resource {
 }
 
 function isTauri() {
-  return 'isTauri' in window && window.isTauri
+  return 'isTauri' in window && !!window.isTauri
```
amrbashir: let's also add an explicit return type

```ts
function isTauri(): boolean {
  return 'isTauri' in window && !!window.isTauri
}
```

amrbashir (changes requested):

amrbashir:
Thanks for contribution

amrbashir (dismissed):

amrbashir (approved):

PR #9631, opened 2024-05-01
fix(core): better error when deserializing plugin config

oscartbeaumont:
Before:
```
Error: PluginInitialization("updater", "invalid type: null, expected struct Config")
```

After:
```
Error: PluginInitialization("updater", "Error deserializing 'plugins.updater' within your Tauri configuration: invalid type: null, expected struct Config")
```

### DIFF
--- a/.changes/better-error-for-invalid-plugin-config.md
+++ b/.changes/better-error-for-invalid-plugin-config.md
@@ -1,0 +1,5 @@
+---
+'tauri': patch:enhance
+---
+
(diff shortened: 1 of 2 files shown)


FabianLars:
Thanks! I love seeing PRs like this one 😍 Just like the other PR, we need a small change file.

oscartbeaumont:
Fixed, sorry!

amrbashir (approved):

amrbashir (approved):

PR #9661, opened 2024-05-05
[v2] nsis: add files

### LINKED ISSUE #9652: [feat] [v2] add files to windows installer config
### Describe the problem

Unlike macOS target, Windows target (in `tauri.conf.json`) does not have `files` to bundle more stuff with the app. It would be great to add that configuration setting too, just like for all other platforms. When using custom NSI file I have to provide all the same, but I'd really love to leverage Tauri's template.

I'd like to specify files the same way I do for other platforms, like macOS.

### Alternatives considered

_No response_

### Additional context

_No response_

pronebird:
This PR adds support for `files` for NSIS installer, similar setting can be found under `macOS` configuration. The rationale for adding this is the following:

- Other platforms already support bundling additional platform-dependent files.
- Using `files` section makes it possible to bundle external binaries _and_ their resources under individual folders. External binaries setting does not provide such flexibility.

## Further thoughts

Making custom `installer.nsi` is difficult. It would be better to provide a way to have a pre and post install hooks. Maybe we could include custom `.nsh` files that would augment the behavior of `templates/installer.nsi`.

This is still work in progress and I am not able to test it. Somehow I keep stumbling upon the following error which is a dead end given that it does mention the source of this error:

> Error [tauri_cli] failed to bundle project: `The system cannot find the file specified. (os error 2)`

Closes https://github.com/tauri-apps/tauri/issues/9652

### DIFF
--- a/core/tauri-config-schema/schema.json
+++ b/core/tauri-config-schema/schema.json
@@ -2294,6 +2294,14 @@
               "type": "null"
             }
           ]
+        },
+        "files": {
+          "description": "The files to include into NSIS installer.",
+          "default": {},
+          "type": "object",
+          "additionalProperties": {
+            "type": "string"
+          }
         }
       },
       "additionalProperties": false
--- a/core/tauri-utils/src/config.rs
+++ b/core/tauri-utils/src/config.rs
@@ -744,6 +744,9 @@ pub struct NsisConfig {
   ///
   /// See <https://nsis.sourceforge.io/Reference/SetCompressor>
   pub compression: Option<NsisCompression>,
+  /// The files to include into NSIS installer.
+  #[serde(default)]
+  pub files: HashMap<PathBuf, PathBuf>,
 }
 
 /// Install Modes for the NSIS installer.
--- a/tooling/bundler/src/bundle/settings.rs
+++ b/tooling/bundler/src/bundle/settings.rs
@@ -415,6 +415,8 @@ pub struct NsisSettings {
   pub display_language_selector: bool,
   /// Set compression algorithm used to compress files in the installer.
   pub compression: Option<NsisCompression>,
+  /// The files to include into NSIS installer.
+  pub files: HashMap<PathBuf, PathBuf>,
 }
 
 /// The Windows bundle settings.
--- a/tooling/bundler/src/bundle/windows/nsis.rs
+++ b/tooling/bundler/src/bundle/windows/nsis.rs
@@ -324,6 +324,13 @@ fn build_nsis_app_installer(
   resources_ancestors.sort_by_key(|p| std::cmp::Reverse(p.components().count()));
   resources_ancestors.pop(); // Last one is always ""
 
+  let files = generate_files_data(settings)?;
+  let files_dirs =
+    std::collections::HashSet::<PathBuf>::from_iter(
+        files.values()
+        .filter_map(|r| r.0.to_owned())
+    );
+
   // We need to convert / to \ for nsis to move the files into the correct dirs
   #[cfg(not(target_os = "windows"))]
   let resources: ResourcesMap = resources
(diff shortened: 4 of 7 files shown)


FabianLars:
> Using files section makes it possible to bundle external binaries and their resources under individual folders. External binaries setting does not provide such flexibility.

The `resources` feature does though (the docs _still_ weren't updated but it accepts objects to configure source and target paths). Doesn't really matter since i agree with you that it makes sense to add this for the sake of cross platform consitency, but still wanted to mention it x)

pronebird:
> > Using files section makes it possible to bundle external binaries and their resources under individual folders. External binaries setting does not provide such flexibility.
> 
> The `resources` feature does though (the docs _still_ weren't updated but it accepts objects to configure source and target paths). Doesn't really matter since i agree with you that it makes sense to add this for the sake of cross platform consitency, but still wanted to mention it x)

That's good to know. However `resources` are located at the `bundle` level. Given that I use one `tauri.conf.json` for all platforms, all assets specified under `resources` would end up in each installer that I build.

Right now I am using `resources` to bundle assets that are used on all platforms, `macOS > files` to ship third-party binaries and their assets for macOS. That trick didn't work on windows. 

Even if `windows > nsis > files` were allowed, I'd still need a way to run custom pre-install and post-install scripts which in my case set-up windows service on Windows, or launch daemon on macOS. So for the time being, I had to resort to to the custom `installer.nsi`. The tauri `installer.nsi` template is 800 LOC long and if I could only tap into some parts of it without copying the whole thing, that would make it so much easier to build the installer.

FabianLars:
> That's good to know. However resources are located at the bundle level. Given that I use one tauri.conf.json for all platforms, all assets specified under resources would end up in each installer that I build.

> Right now I am using resources to bundle assets that are used on all platforms, macOS > files to ship third-party binaries and their assets for macOS. That trick didn't work on windows.

You can have OS specific configs (that overwrite the main config) -> https://tauri.app/v1/api/config#platform-specific-configuration

> Even if windows > nsis > files were allowed, I'd still need a way to run custom pre-install and post-install scripts which in my case set-up windows service on Windows, or launch daemon on macOS. So for the time being, I had to resort to to the custom installer.nsi. The tauri installer.nsi template is 800 LOC long and if I could only tap into some parts of it without copying the whole thing, that would make it so much easier to build the installer.

I did not disagree with the "Further thoughs" section of your PR description, i just think it's off-topic with this PR and should have its own issue/PR :)

pronebird:
> > That's good to know. However resources are located at the bundle level. Given that I use one tauri.conf.json for all platforms, all assets specified under resources would end up in each installer that I build.
> 
> > Right now I am using resources to bundle assets that are used on all platforms, macOS > files to ship third-party binaries and their assets for macOS. That trick didn't work on windows.
> 
> You can have OS specific configs (that overwrite the main config) -> https://tauri.app/v1/api/config#platform-specific-configuration
> 
> > Even if windows > nsis > files were allowed, I'd still need a way to run custom pre-install and post-install scripts which in my case set-up windows service on Windows, or launch daemon on macOS. So for the time being, I had to resort to to the custom installer.nsi. The tauri installer.nsi template is 800 LOC long and if I could only tap into some parts of it without copying the whole thing, that would make it so much easier to build the installer.
> 
> I did not disagree with the "Further thoughs" section of your PR description, i just think it's off-topic with this PR and should have its own issue/PR :)

Thanks again. The merged config could work in my case, at least I wouldn't have to copy files manually. 

Since it's not documented how to write the resources map, could you please maybe shed some light on how does one specify the glob pattern in map, i.e should it be this:

```json
"resources": {
  "tray-icons/": "tray-icons/*.png",
}
```

or that?

```json
"resources": {
  "tray-icons/*.png": "tray-icons/*.png",
}
```

I'll make sure to create [another issue](https://github.com/tauri-apps/tauri/issues/9668) in regards of nsis installer hooks.

Update: it seems that resource map works different from `files`, for example if I want to bundle a binary from `target/release` folder, the bundler errors, stating that the file already exists, which tells me that it must be copying files.

Maybe that's because when building a universal mac binary, my workspace binaries aren't compiled into `universal-apple-darwin` folder and so the copying mechanism works fine, but on Windows everything falls into `target/release`.

If I got it right, the syntax is inverted here, i.e key is path to file, value is where to put it in installer. Keys and values are flipped in that sense in `macos > files`

```json
"resources": {
  "../../target/release/my-other-bin.exe": "my-other-bin.exe"
}
```

I'll [create an issue](https://github.com/tauri-apps/tauri/issues/9666) for that too 😓

amrbashir:
I see no reason to include this option, as your use-case is already covered by `resources` map and the other `files` option (`.deb` for example) is to copy files into subdirectories within the `.deb` file not the final install location. It also doesn't have any APIs to resolve access to them.

PR #9687, opened 2024-05-07
feat(core): add an option for preventing window overflow

### LINKED ISSUE #9671: [feat] add an option to prevent window overflow
### Describe the problem

If I set the initial window size to say 1280x720, it will overflow to outside of the monitor on some screens (e.g. 1920x1080 with 1.75 scaling on a laptop)

It would be nice to have a built-in method for limiting the size to no larger than the working area (e.g. monitor - task bar)
I think this should even be on by default with an option to disable it if someone really wants to have a window bigger than the monitor/work area size

And a configurable margin like it must be smaller than the working area by 100px horizontally and 50px vertically would be nice as well

### Alternatives considered

_No response_

### Additional context

_No response_

Legend-Master:
Closes #9671


Legend-Master:
We probably need some better documentations and names, and also implementation for linux and mac

core/tauri-runtime-wry/src/lib.rs
```diff
@@ -4128,9 +4191,39 @@ fn inner_size(
...
+      return TaoPhysicalSize::new(
+        (monitor_info.rcWork.right - monitor_info.rcWork.left) as u32,
+        (monitor_info.rcWork.bottom - monitor_info.rcWork.top) as u32,
+      );
+    }
+  }
+  #[cfg(target_os = "macos")]
+  {
+    use cocoa::{appkit::NSScreen, base::id};
+    use tao::platform::windows::MonitorHandleExtMacOS;
```
pewsheen:
```suggestion
    use tao::platform::macos::MonitorHandleExtMacOS;
```

core/tauri-runtime-wry/src/lib.rs
```diff
@@ -4128,9 +4191,39 @@ fn inner_size(
...
+    if let Some(ns_screen) = target_monitor.ns_screen() {
+      let ns_screen = ns_screen as id;
+      let rect = unsafe { ns_screen.visibleFrame() };
+      return TaoPhysicalSize::new(rect.size.width as u32, rect.size.height as u32);
```
pewsheen:
```suggestion
      return TaoLogicalSize::new(rect.size.width, rect.size.height)
        .to_physical(target_monitor.scale_factor());
```

pewsheen (changes requested):

pewsheen:
Just a note: on macOS, you will need to turn off `System Settings > Desktop & Dock > Mission Control > Displays have separate Spaces` to let the window span across monitors.

core/tauri-runtime-wry/src/lib.rs
Legend-Master: Oops 🤣

Legend-Master:
Searched around a bit but didn't find a good way of doing it on linux, I guess falling back to use the monitor size is not all that bad

.changes/prevent-window-overflow.md
```diff
@@ -0,0 +1,9 @@
+---
+"tauri": "patch:feat"
+"tauri-utils": "patch:feat"
+"tauri-runtime": "patch:feat"
+"tauri-runtime-wry": "patch:feat"
+"@tauri-apps/api": "patch:feat"
+---
+
+Add an option `prevent_overflow` for preventing the window frin overflowing the monitor on creation
```
BryceBarbara:
```suggestion
Add an option `prevent_overflow` for preventing the window from overflowing the monitor on creation
```

.changes/prevent-window-overflow.md
```diff
@@ -0,0 +1,9 @@
+---
+"tauri": "patch:feat"
+"tauri-utils": "patch:feat"
+"tauri-runtime": "patch:feat"
+"tauri-runtime-wry": "patch:feat"
+"@tauri-apps/api": "patch:feat"
+---
+
+Add an option `prevent_overflow` for preventing the window from overflowing the monitor on creation
```
amrbashir:
```suggestion
Add `preventOverflow` config option to prevent the window from overflowing the monitor size on creation
```

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -3906,7 +3945,7 @@ fn create_window<T: UserEvent, F: Fn(RawWindow) + Send + 'static>(
   }
 
   #[cfg(desktop)]
-  if window_builder.center {
+  if window_builder.prevent_overflow.is_some() || window_builder.center {
     let monitor = if let Some(window_position) = &window_builder.inner.window.position {
       event_loop.available_monitors().find(|m| {
```
amrbashir, line 4184: I don't remember why we are doing it this way, but can't we change this to [`monitor_from_point`](https://docs.rs/tao/latest/tao/event_loop/struct.EventLoopWindowTarget.html#method.monitor_from_point)

crates/tauri-utils/src/config.rs
```diff
@@ -1438,6 +1438,30 @@ pub struct WindowEffectsConfig {
...
+  pub width: u32,
+  /// Vertical margin in physical unit
+  pub height: u32,
+}
+
+/// Prevent overflow with a margin
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+#[cfg_attr(feature = "schema", derive(JsonSchema))]
+#[serde(untagged)]
+pub enum PreventOverflowMarginConfig {
```
amrbashir:
```suggestion
pub enum PreventOverflowConfig {
```

crates/tauri/src/window/mod.rs
```diff
@@ -472,6 +472,31 @@ impl<'a, R: Runtime, M: Manager<R>> WindowBuilder<'a, R, M> {
...
   }
 
+  /// Prevent the window from overflowing the working area (e.g. monitor size - taskbar size) on creation
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux:** Prevent overflowing the monitor instead of workarea
+  /// - **iOS / Android:** Unsupported.
+  #[must_use]
+  pub fn prevent_overflow(mut self) -> Self {
```
amrbashir, line 486: we need similar methods on WebviewWindowBuilder

packages/api/src/window.ts
```diff
@@ -2145,6 +2153,15 @@ interface WindowOptions {
   maxWidth?: number
   /** The maximum height. Only applies if `maxWidth` is also set. */
   maxHeight?: number
+  /**
+   * Prevent initial window from getting bigger than the work area
+   *
+   * ## Platform-specific
```
amrbashir:
```suggestion
   * #### Platform-specific
```

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -4508,3 +4573,35 @@ fn inner_size(
 ) -> TaoPhysicalSize<u32> {
   window.inner_size()
 }
+
+fn get_work_area_size(target_monitor: &MonitorHandle) -> TaoPhysicalSize<u32> {
```
amrbashir: we should create a `MonitorExt` and implement it on `MonitorHandle`, all in a separate file, see `tauri-runtime-wry/src/window/mod.rs` for an example with `WindowExt`

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -907,6 +919,33 @@ impl WindowBuilder for WindowBuilderWrapper {
...
+      .replace(PhysicalSize::new(0, 0).into());
+    self
+  }
+
+  /// Prevent the window from overflowing the working area (e.g. monitor size - taskbar size)
+  /// on creation with a margin
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux:** Prevent overflowing the monitor instead of workarea
```
amrbashir: why? can't we use https://docs.rs/gdk/0.18.0/gdk/prelude/trait.MonitorExt.html#method.workarea

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -3923,39 +3962,65 @@ fn create_window<T: UserEvent, F: Fn(RawWindow) + Send + 'static>(
...
         .inner
         .window
         .inner_size_constraints
         .clamp(desired_size, scale_factor)
         .to_physical::<u32>(scale_factor);
+      let mut window_size = inner_size;
+      #[allow(unused_mut)]
+      // Left and right window shadow counts as part of the window on Windows
+      // We need to include it when calculating positions, but not size
+      let mut shadow_width = 0;
```
amrbashir, line 4216: why do we need to include it? I think we should just include the starting of the window border not shadow

amrbashir (changes requested):

crates/tauri-runtime-wry/src/lib.rs
Legend-Master: If I remembered correctly, the shadow counts as part of the window when we set the positions, say your shadow is 5px left right bottom, and you set the window's position to (0, 0), then the actual window's top left will be placed at (5, 0

crates/tauri-runtime-wry/src/lib.rs
Legend-Master: Ah, didn't know it exits, do you think we can move this `get_work_area_size` to `tao` instead? If not we'll need a way to access the inner `gdk::Monitor` from `tao`

crates/tauri-runtime-wry/src/lib.rs
Legend-Master: `window_builder.inner.window.position` is a `Position` here that we need to convert to physical position with scale factor so I don't think we can use `monitor_from_point` here, also I think `monitor_from_point` was introduced after this PR was made?

crates/tauri-runtime-wry/src/lib.rs
amrbashir: we can expose the raw `gdk::Monitor` in `tao` similar to `MonitorHandleExtWindows::hmonitor`

crates/tauri-runtime-wry/src/lib.rs
amrbashir: maybe, let's keep it as is for now then

crates/tauri-runtime-wry/src/monitor/mod.rs
```diff
@@ -0,0 +1,28 @@
...
+mod windows;
+
+#[cfg(desktop)]
+pub trait MonitorExt {
+  /// Enable or disable the window
+  ///
+  /// ## Platform-specific:
+  ///
+  /// - **Android / iOS**: Unsupported.
+  fn get_work_area_size(&self) -> PhysicalSize<u32>;
```
amrbashir: let's rename to just 
```suggestion
  fn work_area(&self) -> PhysicalSize<u32>;
```

lucasfernog (approved):
we almost celebrated this PR anniversary, sorry 😂

Issue #7684, opened 2023-08-23
[bug] Tauri's JavaScript shell Command skips stdout lines when large amounts of data is printed


mauritzn:
### Describe the bug

When running a command using Tauri's shell [Command](https://tauri.app/v1/api/js/shell#command) that returns a lot of data some lines get skipped in stdout and the process seems to keep running indefinitely in the background, preventing future Commands and also sometimes preventing the app from refreshing properly. From my testing around 20,000+ lines in stdout causes the problem appear.

### Reproduction

[https://github.com/mauritzn/tauri-shell-bug](https://github.com/mauritzn/tauri-shell-bug)

This repo was made as a minimal reproduction of this bug. It includes a Python script _(`__print_numbers.py`)_ which prints numbers until 250,000 numbers have been printed, it starts by printing `started` and ends by printing `done`.

The amount of numbers the Python script prints can be changed in the script, usually 20,000 is enough for the bug to appear, but I made the default 250,000 since it was still bearable and should be more than enough for the bug to appear.

I recommend running it using `npm run tauri dev`, this is mainly due to the automatic finding of the Python file, since it looks for it at the current execution path _(feel free to change this)_. Running it as a built binary also works, but make sure that the `__print_numbers.py` file is in the same directory.

### Expected behavior

Everything outputted from the CLI application ran using [Command](https://tauri.app/v1/api/js/shell#command) would be accessible.

### Platform and versions

```text
[✔] Environment
    - OS: Windows 10.0.22631 X64
    ✔ WebView2: 123.0.2420.65
    ✔ MSVC:
        - Visual Studio Build Tools 2022
        - Visual Studio Community 2019
    ✔ rustc: 1.77.1 (7cf61ebde 2024-03-27)
    ✔ cargo: 1.77.1 (e52e36006 2024-03-26)
    ✔ rustup: 1.27.0 (bbb9276d2 2024-03-08)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (default)
    - node: 20.10.0
    - npm: 10.5.0
    - bun: 1.1.0

[-] Packages
    - tauri [RUST]: 1.6.1
    - tauri-build [RUST]: 1.5.1
    - wry [RUST]: 0.24.7
    - tao [RUST]: 0.16.8
    - @tauri-apps/api [NPM]: 1.5.3
    - @tauri-apps/cli [NPM]: 1.5.11

[-] App
    - build-type: bundle
    - CSP: unset
[...]

PR #9706, opened 2024-05-08
fix(core/shell): speedup `Command.execute` & fix extra new lines

amrbashir:
The speed gains comes from running the Command in Rust fully and returning the result in one go instead of using events.

The extra new lines was a regression from https://github.com/tauri-apps/tauri/pull/6519

ref: https://github.com/tauri-apps/tauri/issues/7684#issuecomment-2100897383


core/tauri/src/endpoints/shell.rs
```diff
@@ -81,7 +90,76 @@ pub enum Cmd {
...
+  #[allow(unused_variables)]
+  fn execute_and_return<R: Runtime>(
+    context: InvokeContext<R>,
+    program: String,
+    args: ExecuteArgs,
+    options: CommandOptions,
+  ) -> super::Result<ChildProcessReturn> {
+    let mut command = prepare_cmd(&context, &program, args, &options)?;
+
+    #[cfg(any(shell_execute, shell_sidecar))]
```
lucasfernog: why isn't this cfg on the function itself?

lucasfernog: wait actually the cfg is not needed at all

lucasfernog (approved):

PR #9726, opened 2024-05-09
docs: added example of tauri.allowlist.protocol

angelxmoreno:
I prefer docs that do not assume I've read other parts of the docs. It took me some time to figure out the right format for the protocol allowlist. Also, from this page alone it is not clear that I should be using the runtime variables for the scope array.


amrbashir (approved):
Thank you for contributions

PR #9789, opened 2024-05-15
feat(core): add `app-region: drag` to `data-tauri-drag-region`

Legend-Master:
Use `app-region: drag` on supported platforms (currently only WebView2 123+ on Windows)

### DIFF
--- a/.changes/app-region-drag.md
+++ b/.changes/app-region-drag.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch:feat
+---
+
+Add `app-region: drag` to HTML elements with `data-tauri-drag-region` on supported platforms (currently only WebView2 123+ on Windows)
--- a/.changes/app-region-drag.md
+++ b/.changes/app-region-drag.md
@@ -2,4 +2,4 @@
 "tauri": patch:feat
 ---
 
-Add `app-region: drag` to HTML elements with `data-tauri-drag-region` on supported platforms (currently only WebView2 123+ on Windows)
+Add `app-region: drag` to HTML elements with `data-tauri-drag-region` on Windows, only WebView2 123+, which should fix dragging using touch.
--- a/core/tauri/src/window/scripts/drag.js
+++ b/core/tauri/src/window/scripts/drag.js
@@ -63,4 +63,9 @@
       }
     })
   }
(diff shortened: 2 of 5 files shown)


core/tauri/src/window/scripts/drag.js
```diff
@@ -63,4 +63,9 @@
       }
     })
   }
+  if ('adoptedStyleSheets' in document) {
```
amrbashir:
```suggestion
  if (osName === 'windows' && 'adoptedStyleSheets' in document) {
```

amrbashir (changes requested):
We probably need to get the webview version and embed it into this script at startup then disable the custom JS handler on Windows when using WebView2 123+

Legend-Master:
> We probably need to get the webview version and embed it into this script at startup then disable the custom JS handler on Windows when using WebView2 123+

I don't think we need to, we won't get events in non client region, so if `app-region: drag` is working, it should just skip the other handlers

core/tauri/src/window/scripts/drag.js
Legend-Master: We might get `app-region: drag` in the future on Linux and Mac, do we want to limit this part to only run on Windows?

amrbashir: yes, until other platforms support, we should limit to only Windows

amrbashir:
> > We probably need to get the webview version and embed it into this script at startup then disable the custom JS handler on Windows when using WebView2 123+
> 
> I don't think we need to, we won't get events in non client region, so if `app-region: drag` is working, it should just skip the other handlers

That's enough I guess

.changes/app-region-drag.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri": patch:feat
+---
+
+Add `app-region: drag` to HTML elements with `data-tauri-drag-region` on supported platforms (currently only WebView2 123+ on Windows)
```
amrbashir:
```suggestion
Add `app-region: drag` to HTML elements with `data-tauri-drag-region` on Windows, only WebView2 123+, which should fix dragging using touch.
```

amrbashir (changes requested):

amrbashir (approved):

PR #9796, opened 2024-05-15
fix(core/shell): speedup `Command.execute` & fix extra new lines

amrbashir:
Re-add of https://github.com/tauri-apps/tauri/pull/9706 but with a minor release

The speed gains comes from running the Command in Rust fully and returning the result in one go instead of using events.

The extra new lines was a regression from https://github.com/tauri-apps/tauri/pull/6519

ref: https://github.com/tauri-apps/tauri/issues/7684#issuecomment-2100897383


.changes/shell-execute-extra-newline.md
```diff
@@ -0,0 +1,6 @@
+---
+"@tauri-apps/api": "patch:bug"
```
FabianLars, line 2: does this also need to be minor?

Problem would be then that we have 1.6 in js and 1.7 in rust. Maybe we should copy the last 1.5 release into 1.6 to fix that.

amrbashir: The problematic change was in the other change file where it is bumped as minor.

FabianLars: Ah okay, that leaves the version mismatch question though. Maybe i should move that to discord though as it's not directly related to this PR.

lucasfernog (dismissed):

PR #9833, opened 2024-05-20
fix(bundler/nsis): fix registering deep-link protocols

amrbashir:


FabianLars (approved):
I don't understand how i regularly commit something different than what i tested. wtf am i doing

FabianLars (approved):

PR #9840, opened 2024-05-21
fix(bundler/rpm): Reduce compression level to gzip/6

FabianLars:
This matches the .deb builder to reduce build times. It's still much slower than .deb but this is probably something we have to live with or something to be fixed in the rpm crate itself. We could also look at how well other compression formats are supported across distro versions i guess (can't really do that for .deb yet).

p.s. the reason i do this is because rn the rpm bundler is as slow, or even slower than the appimage builder which shouldn't be possible considering how much more stuff the appimage contains...

### DIFF
--- a/tooling/bundler/src/bundle/linux/rpm.rs
+++ b/tooling/bundler/src/bundle/linux/rpm.rs
@@ -47,7 +47,9 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
   let license = settings.license().unwrap_or_default();
   let mut builder = rpm::PackageBuilder::new(name, version, &license, arch, summary)
     .epoch(epoch)
-    .release(release);
+    .release(release)
+    // This matches .deb compression. On a 240MB source binary the bundle will be 100KB larger than rpm's default while reducing build times by ~25%.
+    .compression(rpm::CompressionWithLevel::Gzip(6));
 
   if let Some(description) = settings.long_description() {
     builder = builder.description(description.trim())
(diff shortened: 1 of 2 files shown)


tooling/bundler/src/bundle/linux/rpm.rs
```diff
@@ -47,7 +47,9 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
   let license = settings.license().unwrap_or_default();
   let mut builder = rpm::PackageBuilder::new(name, version, &license, arch, summary)
     .epoch(epoch)
-    .release(release);
+    .release(release)
+    // This matches .deb compression. On a 240MB source binary the bundle will be 100KB larger than rpm's default while reducing build times by ~25%.
+    .compression(rpm::CompressionWithLevel::Gzip(6));
```
amrbashir, line 52: shouldn't we make this configurable? some users will still want to shave off those 100kb

amrbashir (changes requested):

tooling/bundler/src/bundle/linux/rpm.rs
FabianLars: it's not configurable for .deb either and since this was more like a hotfix (came up in a customer call) i didn't bother. We can totally look into that though. (i'm done for today)

amrbashir: we have config for NSIS to control the compression, maybe we should add a similar option for linux targets as well but let's do it in another PR then

amrbashir (approved):

PR #9850, opened 2024-05-22
chore: emit rustc-check-cfg for nightly

JonasKruckenberg:
The old `cargo:` instruction syntax has been deprecated with version 1.77 and newer features (such as the `check-cfg` feature introduced in 1.79) require the new double colon `cargo::` syntax. 
This change keeps the old syntax around since it is ignored on newer versions and still necessary for our MSRV policy.

### DIFF
--- a/core/tauri-build/src/lib.rs
+++ b/core/tauri-build/src/lib.rs
@@ -211,9 +211,8 @@ fn copy_frameworks(dest_dir: &Path, frameworks: &[String]) -> Result<()> {
 // creates a cfg alias if `has_feature` is true.
 // `alias` must be a snake case string.
 fn cfg_alias(alias: &str, has_feature: bool) {
-  println!("cargo::rustc-check-cfg=cfg({alias})");
+  println!("cargo:rustc-check-cfg=cfg({alias})");
   if has_feature {
-    println!("cargo::rustc-cfg={alias}");
     println!("cargo:rustc-cfg={alias}");
   }
 }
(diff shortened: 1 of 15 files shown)


core/tauri-build/src/lib.rs
```diff
@@ -211,7 +211,9 @@ fn copy_frameworks(dest_dir: &Path, frameworks: &[String]) -> Result<()> {
 // creates a cfg alias if `has_feature` is true.
 // `alias` must be a snake case string.
 fn cfg_alias(alias: &str, has_feature: bool) {
```
lucasfernog, line 213: we pretty much have this function on all crates, can you also update those? just search for `cargo:rustc-cfg`

amrbashir:
is it okay to bump MSRV to 1.77? seems too soon even for beta

JonasKruckenberg:
Its not bumping the MSRV, since the old syntax style instructions will still be emitted as well

core/tauri-build/src/lib.rs
JonasKruckenberg: sure thing

amrbashir:
ah my bad, didn't look carefully and thought we are replacing old syntax. You even wrote that in the PR description 🤦‍♂️

lucasfernog:
We must use `cargo:` instead of `cargo::` until we raise our MSRV. Also we do not need to duplicate the instructions, Cargo still supports the `cargo:` syntax. This was verified by running `cargo +nightly clippy` with latest nightly.

lucasfernog (approved):

PR #9856, opened 2024-05-22
feat(android): aware of version info from tauri config

mariotaku:
This PR makes generated Android packages having the same `versionName` as it's defined in `tauri.config.json`, and providing 2 ways to define `versionCode`:

1. Calculated: `version.major * 1000000 + version.minor * 1000 + version.patch`
2. Overridden 1 by `bundle > android > versionCode` in `tauri.config.json`


core/tauri-build/src/mobile.rs
```diff
@@ -5,17 +5,22 @@
...
 
 use anyhow::{Context, Result};
+use semver::Version;
+use tauri_utils::config::Config;
 
-pub fn generate_gradle_files(project_dir: PathBuf) -> Result<()> {
+pub fn generate_gradle_files(project_dir: PathBuf, config: &Config) -> Result<()> {
   let gradle_settings_path = project_dir.join("tauri.settings.gradle");
   let app_build_gradle_path = project_dir.join("app").join("tauri.build.gradle.kts");
+  let app_gradle_properties_path = project_dir.join("app").join("gradle.properties");
```
amrbashir: let's pick a different name, `gradle.properties` already exists in the android project even if it is in another directory
```suggestion
  let app_gradle_properties_path = project_dir.join("app").join("tauri.properties");
```

core/tauri-build/src/mobile.rs
```diff
@@ -48,13 +53,27 @@ dependencies {"
 
   app_build_gradle.push_str("\n}");
 
+  if let Some(version) = config.version.as_ref() {
+    app_gradle_properties.push_str(&format!("tauri.android.versionName={}\n", version));
+    if let Some(version_code) = config.bundle.android.version_code.as_ref() {
+      app_gradle_properties.push_str(&format!("tauri.android.versionCode={}\n", version_code));
+    } else if let Ok(version) = Version::parse(version) {
+      let version_code = version.major * 1000000 + version.minor * 1000 + version.patch;
+      app_gradle_properties.push_str(&format!("tauri.android.versionCode={}\n", version_code));
```
amrbashir: we could emit `tauri.` prefix since it will be inside `tauri.properties` file
```suggestion
    app_gradle_properties.push_str(&format!("android.versionName={}\n", version));
    if let Some(version_code) = config.bundle.android.version_code.as_ref() {
      app_gradle_properties.push_str(&format!("android.versionCode={}\n", version_code));
    } else if let Ok(version) = Version::parse(version) {
      let version_code = version.major * 1000000 + version.minor * 1000 + version.patch;
      app_gradle_properties.push_str(&format!("android.versionCode={}\n", version_code));
```

core/tauri-build/src/mobile.rs
```diff
@@ -48,13 +53,27 @@ dependencies {"
...
+    }
+  }
+
   write(&gradle_settings_path, gradle_settings).context("failed to write tauri.settings.gradle")?;
 
   write(&app_build_gradle_path, app_build_gradle)
     .context("failed to write tauri.build.gradle.kts")?;
 
+  write(&app_gradle_properties_path, app_gradle_properties)
+    .context("failed to write gradle.properties")?;
```
amrbashir:
```suggestion
    .context("failed to write tauri.properties")?;
```

tooling/cli/templates/mobile/android/app/.gitignore
```diff
@@ -2,4 +2,5 @@
 /src/main/jniLibs/**/*.so
 /src/main/assets/tauri.conf.json
 /tauri.build.gradle.kts
-/proguard-tauri.pro
\ No newline at end of file
+/proguard-tauri.pro
+/gradle.properties
```
amrbashir:
```suggestion
/tauri.properties
```

core/tauri-build/src/mobile.rs
```diff
@@ -5,17 +5,22 @@
...
   let app_build_gradle_path = project_dir.join("app").join("tauri.build.gradle.kts");
+  let app_gradle_properties_path = project_dir.join("app").join("gradle.properties");
 
   let mut gradle_settings =
     "// THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.\n".to_string();
   let mut app_build_gradle = "// THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
 val implementation by configurations
 dependencies {"
     .to_string();
+  let mut app_gradle_properties =
```
amrbashir:
```suggestion
  let mut app_tauri_properties =
```

amrbashir (changes requested):

amrbashir:
Oh looks like you didn't sign your commits. You need to setup [commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits), then you can sign past commit like [this](https://superuser.com/questions/397149/can-you-gpg-sign-old-commits/1123928#1123928) for example.

mariotaku:
Hello! I have updated & signed the commit.

core/tauri-build/src/mobile.rs
```diff
@@ -48,13 +53,27 @@ dependencies {"
...
+      app_tauri_properties.push_str(&format!("tauri.android.versionCode={}\n", version_code));
+    }
+  }
+
   write(&gradle_settings_path, gradle_settings).context("failed to write tauri.settings.gradle")?;
 
   write(&app_build_gradle_path, app_build_gradle)
     .context("failed to write tauri.build.gradle.kts")?;
 
+  write(&app_tauri_properties_path, app_tauri_properties)
```
amrbashir: maybe avoid writing if `config.version` is None and don't emit `cargo:rerun` either?

amrbashir (changes requested):

amrbashir:
Seems like you need to run `cargo fmt` and you're also missing documentation on `version_code` field and schema is not generated so run `cargo -b tauri-config-schema` to generate it,

lucasfernog (approved):

lucasfernog:
nice one!

core/tauri-build/src/mobile.rs
```diff
@@ -48,13 +52,40 @@ dependencies {"
...
 
+  if let Some(version) = config.version.as_ref() {
+    app_tauri_properties.push(format!("tauri.android.versionName={}", version));
+    if let Some(version_code) = config.bundle.android.version_code.as_ref() {
+      app_tauri_properties.push(format!("tauri.android.versionCode={}", version_code));
+    } else if let Ok(version) = Version::parse(version) {
+      let version_code = version.major * 1000000 + version.minor * 1000 + version.patch;
+      app_tauri_properties.push(format!("tauri.android.versionCode={}", version_code));
+    }
+  }
```
pewsheen, line 63: Do we need to check the final version code here? Since it might be out of  the valid range. For example, `create-tauri-app` will create with version `0.0.0` hence the VersionCode is 0 which is invalid

mariotaku: @pewsheen This is a good catch! I think it can verified after line 60.

amrbashir: I think we should fail with an error when building for android if `versionCode` is not specified in `tauri.conf.json` and app version is `0.0.0` but only in release mode. Similar to what we do when `identifier` is set to `com.tauri.dev`

pewsheen: Sounds good! I've created a PR fro this: https://github.com/tauri-apps/tauri/pull/9898

PR #9861, opened 2024-05-23
feat(bundler/windows): Update wix to fix vulnerabilities and add arm support

FabianLars:
(careful, v1 branch!)

I won't be able to actually test this until Sunday but opening this for a community member to test (and for whoever else that wants to of course)

### DIFF
--- a/.changes/feat-wix-arm64.md
+++ b/.changes/feat-wix-arm64.md
@@ -1,0 +1,6 @@
+---
+tauri-bundler: minor:feat
+tauri-cli: minor:feat
+---
+
+Upgraded the WiX version to 3.14 which fixes vulnerability issues and adds support for Arm targets.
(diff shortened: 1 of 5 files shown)


lucasfernog (dismissed):
tested on UTM, works

lucasfernog:
is this one ready @FabianLars ?

FabianLars:
Change of plans and I wasn't home yet since I opened it. If you say it's working then I think it's good (check the change files and remove the bump one if you're not gonna release it quickly, we just wanted to match minors again).

Also not the biggest fan of the new folder but I guess that's the only way to make sure it gets the new one. Maybe we should use this versioning in general and add a tauri clean command idk

FabianLars:
With 9706(?) merged the bump changefile is useless, please remove it before merge (Idk if im dumb but deleting files is not working in the app for me)

lucasfernog (approved):

Issue #9662, opened 2024-05-05
libwebkit2gtk-4.0 not available in Ubuntu 24 & Debian 13 repositories


happybeing:
### Describe the bug

The Tauri v1 pre-requisite `libwebkit2gtk-4.0-dev` is not available in Ubuntu 24 repositories. This causes an error when installing the pre-requisites specified [here](https://tauri.app/v1/guides/getting-started/prerequisites/#setting-up-linux), and when running `cargo tauri dev` on Ubuntu 24.

Note: `libwebkit2gtk-4.1-dev` is available but doesn't help with Tauri 1. 

### Reproduction

Run:
```
sudo apt update
sudo apt install libwebkit2gtk-4.0-dev \
    build-essential \
    curl \
    wget \
    file \
    libssl-dev \
    libgtk-3-dev \
    libayatana-appindicator3-dev \
    librsvg2-dev
```
reports:
```
E: Unable to locate package libwebkit2gtk-4.0-dev
E: Couldn't find any package by glob 'libwebkit2gtk-4.0-dev'
```

And `cargo tauri dev`:
```
The following warnings were emitted during compilation:

warning: javascriptcore-rs-sys@0.4.0: 

error: failed to run custom build command for `javascriptcore-rs-sys v0.4.0`

Caused by:
  process didn't exit successfully: `/home/mrh/src/safe-browser/awe/src-tauri/target/debug/build/javascriptcore-rs-sys-855dc84908b0218c/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-env-changed=JAVASCRIPTCOREGTK_4.0_NO_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG_PATH
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH
[...]

PR #9863, opened 2024-05-23
fix(Linux): Fix libwebkit2gtk-4.0 not available issue

naman-crabnebula:
Hi guys!

So, in this PR, I am embedding `libwebkit2gtk-4.0` shared objects along in the `deb` package.
This change will increase the `deb` file size by roughly 40 MB, but it seems to be best option right now.

- closes https://github.com/tauri-apps/tauri/issues/9662

It still requires more testing, so I am making it a draft PR right now.
Thanks for your patience.


amrbashir:
This size increase is not desired and is a breaking change, instead it should be an opt-in. We should also document how users can add a PPA to their system so they can install required packages.

FabianLars:
Agreed because this change would break support for building .deb bundles on non debian/ubuntu distros.

Also, make sure to test the installation on systems that don't have webkitgtk installed (via the 2204 repo workaround). I mean, running the app after the app, not just the installation.

naman-crabnebula:
> instead it should be an opt-in

Well, `deb` packages are supposed to be working on Debian, and there is no point in building one, if it can't work with latest Debian distros, imho.

> We should also document how users can add a PPA to their system so they can install required packages.

Well, I thought about this, and tried this as well. But the issue is : It can break dependency tree, and may even lead to breaking entire desktop environment for some users. So, it is quite risky.

> Agreed because this change would break support for building .deb bundles on non debian/ubuntu distros.

No, it won't. I tested it on Arch Linux, Manjaro, Ubuntu 22. I am gonna test it on Fedora as well.

> Also, make sure to test the installation on systems that don't have webkitgtk installed (via the 2204 repo workaround). I mean, running the app after the app, not just the installation.

It is not possible to run Tauri itself without `webkit2gtk`. Also, yeah, we can not build `deb` without `libwebkit2gtk`, which is expected, since you can't bundle libraries that you don't have in first place.

amrbashir:
> > instead it should be an opt-in
> 
> Well, `deb` packages are supposed to be working on Debian, and there is no point in building one, if it can't work with latest Debian distros, imho.

That still doesn't justify making this breaking changes. Also `deb` are supposed to be small and use dependencies from PPAs, so it is counter intuitive to embed dependencies and at this point, they might as well just use the AppImage.

> > We should also document how users can add a PPA to their system so they can install required packages.
> 
> Well, I thought about this, and tried this as well. But the issue is : It can break dependency tree, and may even lead to breaking entire desktop environment for some users. So, it is quite risky.

Then we just make a PPA that we maintain ourselves that contains only `libwebkit2gtk-4.0` to avoid breaking other dependencies and document how users can use this PPA (and tauri users will document the same for their users).

naman-crabnebula:
> That still doesn't justify making this breaking changes

All we need is that `libwebkit2gtk` must be installed on the building system, which is already a dependency for Tauri

> Also deb are supposed to be small and use dependencies from PPAs, so it is counter intuitive to embed dependencies and at this point, they might as well just use the AppImage.

Yeah, I understand the concerns. That's why it took me so much time in search of other alternatives.

> Then we just make a PPA that we maintain ourselves that contains only libwebkit2gtk-4.0

We have a couple of more libraries, and not only `libwebkit2gtk-4.0`. Maintaining a PPA and **asking users to add it manually** would be an overkill for this I guess, and would be *last resort*, imho.


I have another cool idea, can we just create another `deb` package, that will just contain required libraries. It can be installed separately by user.  With this

1. Size of `deb` package will be same as before.
2. Easy to maintain ( basically no extra efforts at all )
3. Easy for user to just install that package
4. No extra PPA or anything will be required.
5. Original development workflow will not be disturbed.

amrbashir:
> > That still doesn't justify making this breaking changes
> 
> All we need is that `libwebkit2gtk` must be installed on the building system, which is already a dependency for Tauri

It is not only for development, end-users need to be able to install it as well.

> > Also deb are supposed to be small and use dependencies from PPAs, so it is counter intuitive to embed dependencies and at this point, they might as well just use the AppImage.
> 
> Yeah, I understand the concerns. That's why it took me so much time in search of other alternatives.

I know, thanks for your help but I honestly can't approve this solution unless it is an opt-in.

> > Then we just make a PPA that we maintain ourselves that contains only libwebkit2gtk-4.0
> 
> We have a couple of more libraries, and not only `libwebkit2gtk-4.0`. Maintaining a PPA and **asking users to add it manually** would be an overkill for this I guess, and would be _last resort_, imho.

How many more dependencies? and excuse my ignorance but what does maintaining a PPA entails? what would I need to do on a daily basis to maintain it?


> I have another cool idea, can we just create another `deb` package, that will just contain required libraries. It can be installed separately by user. With this
> 
> 1. Size of `deb` package will be same as before.
> 2. Easy to maintain ( basically no extra efforts at all )
> 3. Easy for user to just install that package
> 4. No extra PPA or anything will be required.
> 5. Original development workflow will not be disturbed.

That could work but I am not keen on it either, it means each application will distribute a different version of libwebkit2gtk and could potentially endup with an outdated version or conflict with other versions installed from another apps.

naman-crabnebula:
> It is not only for development, end-users need to be able to install it as well.

Yeah, this PR basically copies the `webkit2gtk` library available on building system, to the `deb` file, so that, end users don't have to do anything on Latest Ubuntu versions.

> How many more dependencies?

`libjxl` and `icu` are the main cause of possible dependency conflict. `icu` is used by a lot of packages, and might cause dependency issue. 

> and excuse my ignorance but what does maintaining a PPA entails?

We would need a server, that can fetch and build the package. I am really not sure, if Github actions can be used, because I have never seen anyone using it. Not much stuff. But it is more troublesome on the users' end, since they have to add PPA manually. We can create a script for that as well ( it can be done pretty easily with `sed` )
But I don't think maintaining a server solely for this is a good idea at all.

> That could work but I am not keen on it either, it means each application will distribute a different version of libwebkit2gtk

Well, we can ship it via Tauri Website maybe.

> could potentially endup with an outdated version

Outdated version is required in this case :sweat_smile: 

> conflict with other versions installed from another apps

No, it won't conflict with other apps, as we are shipping it as a 3rd party app.

amrbashir:
> We would need a server, that can fetch and build the package.

Can't we just fetch from the old ubuntu PPA? or even just add a pre-built packages?

> But I don't think maintaining a server solely for this is a good idea at all.

Why not?

> Outdated version is required in this case 😅

I meant outdated version of this new `deb` (that encapsulates `libwebkit2gtk`)  (if we release a new version for some reason).

> No, it won't conflict with other apps, as we are shipping it as a 3rd party app.

I meant conflict with other tauri apps, not other non-tauri apps.

naman-crabnebula:
> Can't we just fetch from the old ubuntu PPA? or even just add a pre-built packages?

Yeah, we can do that.

> I meant conflict with other tauri apps, not other non-tauri apps.

Yeah, it won't conflict with any Tauri App as well, since user have to install it just once, and forget about it.

amrbashir:
I guess that could work, but I'd still like to see that as opt-in, so tauri users can know what is required on their part.

amrbashir:
I think we should try solving this by addressing all of these points:
- Build another `deb` that has these required packages to be used by any v1 tauri app.
- Document how users can use an old `PPA`
- Add an option to embed required libs inside the `deb`

naman-crabnebula:
> I think we should try solving this by addressing all of these points:
> 
> * Build another `deb` that has these required packages to be used by any v1 tauri app.
> * Document how users can use an old `PPA`
> * Add an option to embed required libs inside the `deb`

Yeah, sounds good

RoseBlume:
How about putting this in [pacstall](https://github.com/pacstall/pacstall-programs) and updating prerequisites with installation instructions via pacstall.

Kutter07 (approved):

darth-raijin:
Did anyone come around to making the deb package?

PR #9879, opened 2024-05-25
fix(nsis): can't unpin start menu shortcut on uninstall

Legend-Master:
Actually it does able to unpin the shortcut for some reason even before `$AppStartMenuFolder` is even there, anyways, this should be called after it


Legend-Master:
Fixed/included in #9559, closing this then

PR #9883, opened 2024-05-26
feat!(core): add bundle createUpdaterArtifacts configuration

### LINKED ISSUE #9839: [feat] move updater bundle target to a separate field
### Describe the problem

About updater configs, I think updater is not really a "target" anymore

I would suggest we change our config files a bit, removing updater from bundle targets, instead making it a separate field, like:

`updater: true | false | "legacyZipped"`

- `true`: only generates/reuses non-zipped installers/updaters
- `legacyZipped`: keep the v1 behavior 

### Alternatives considered

_No response_

### Additional context

_No response_

Legend-Master:
Closes #9839


FabianLars:
I know this still a draft, but so we don't forget: The `migrate` command should set the v1 setting if the updater was enabled before.

Legend-Master:
~~Note I haven't tested this much yet~~

Tested on Windows now

core/tauri-utils/src/config.rs
```diff
@@ -1049,9 +1071,12 @@ pub struct BundleConfig {
   /// Whether Tauri should bundle your application or just output the executable.
   #[serde(default)]
   pub active: bool,
-  /// The bundle targets, currently supports ["deb", "rpm", "appimage", "nsis", "msi", "app", "dmg", "updater"] or "all".
+  /// The bundle targets, currently supports ["deb", "rpm", "appimage", "nsis", "msi", "app", "dmg"] or "all".
   #[serde(default)]
   pub targets: BundleTarget,
+  #[serde(default)]
+  /// Produce updaters and their signatures or not
+  pub updater: Updater,
```
amrbashir: After some thinking I think we should call this `createUpdaterArtifacts`. This makes it clear that it generates files to be used by the updater and won't confuse users as to why this option exists here and not inside `plugins > updater`

amrbashir: cc @FabianLars @lucasfernog

core/tauri-utils/src/config.rs
```diff
@@ -1038,6 +1033,33 @@ impl BundleResources {
...
+  /// Generates lagacy zipped v1 compatible updaters
+  String(V1Compatible),
+  /// Produce updaters and their signatures or not
+  // Can't use untagged on enum field here: https://github.com/GREsau/schemars/issues/222
+  Bool(bool),
+}
+
+impl Default for Updater {
+  fn default() -> Self {
+    Self::Bool(false)
```
amrbashir, line 1082: @lucasfernog should we make `V1Compatible` the default and emit a warning that it will be changed in v3?

tooling/cli/node/index.d.ts
```diff
@@ -1,7 +1,3 @@
-// Copyright 2019-2024 Tauri Programme within The Commons Conservancy
```
amrbashir: let's revert these changes

tooling/cli/node/index.js
```diff
@@ -1,7 +1,3 @@
-// Copyright 2019-2024 Tauri Programme within The Commons Conservancy
-// SPDX-License-Identifier: Apache-2.0
-// SPDX-License-Identifier: MIT
-
```
amrbashir: and these as well

tooling/bundler/src/bundle/settings.rs
```diff
@@ -156,8 +155,10 @@ pub struct PackageSettings {
 /// The updater settings.
 #[derive(Debug, Default, Clone)]
 pub struct UpdaterSettings {
+  /// Should generate v1 compatible unzipped updater
```
```suggestion
  /// Should generate v1 compatible zipped updater
```

tooling/cli/src/bundle.rs
```diff
@@ -214,77 +229,72 @@ pub fn bundle<A: AppSettings>(
       )
     })
     .collect();
```
amrbashir, line 238:
```suggestion
    .collect();

```

tooling/cli/src/bundle.rs
```diff
@@ -214,77 +229,72 @@ pub fn bundle<A: AppSettings>(
...
-            log::warn!("The updater secret key from `TAURI_PRIVATE_KEY` does not match the public key from `plugins > updater > pubkey`. If you are not rotating keys, this means your configuration is wrong and won't be accepted at runtime when performing update.");
-          }
-          signed_paths.push(signature_path);
-        }
-      }
+  let Some(pubkey) = &update_settings.pubkey else {
+    // Maybe return error here?
+    return Ok(());
+  };
+  // get the public key
```
amrbashir:
```suggestion
  };

  // get the public key
```

tooling/cli/src/bundle.rs
```diff
@@ -214,77 +229,72 @@ pub fn bundle<A: AppSettings>(
...
+              let maybe_path = Path::new(&private_key);
+              let private_key = if maybe_path.exists() {
+                std::fs::read_to_string(maybe_path)?
+              } else {
+                private_key
+              };
+              updater_signature::secret_key(private_key, password)
+            }
+            _ => Err(anyhow::anyhow!("A public key has been found, but no private key. Make sure to set `TAURI_SIGNING_PRIVATE_KEY` environment variable.")),
+          }?;
```
amrbashir: the formatting here is weird, let's format manually if `cargo fmt` fails to format it

tooling/cli/src/bundle.rs
```diff
@@ -214,77 +229,72 @@ pub fn bundle<A: AppSettings>(
...
+
+  let pubkey = base64::engine::general_purpose::STANDARD.decode(pubkey)?;
+  let pub_key_decoded = String::from_utf8_lossy(&pubkey);
+  let public_key = minisign::PublicKeyBox::from_string(&pub_key_decoded)?.into_public_key()?;
+
+  // make sure we have our package built
```
amrbashir: outdated comment, lets' remove
```suggestion
```

tooling/bundler/src/bundle/settings.rs
```diff
@@ -156,8 +155,10 @@ pub struct PackageSettings {
...
+  pub v1_compatible: bool,
   /// Signature public key.
-  pub pubkey: String,
+  pub pubkey: Option<String>,
```
amrbashir: let's just remove the option, it is not needed anymore, it only existed to check whether updater was enabled or not

tooling/cli/src/interface/rust.rs
```diff
@@ -812,11 +812,25 @@ impl AppSettings for RustAppSettings {
...
+    let v1_compatible = matches!(config.bundle.updater, Updater::String(_));
+    let updater_settings = if updater_enabled {
+      let (pubkey, msiexec_args) =
+        if let Some(updater_plugin_config) = config.plugins.0.get("updater") {
+          let updater: UpdaterConfig = serde_json::from_value(updater_plugin_config.clone())?;
+          let pubkey = if !updater.pubkey.is_empty() {
+            Some(updater.pubkey)
+          } else {
+            None
+          };
```
amrbashir: and let's remove this too

tooling/cli/src/migrate/config.rs
```diff
@@ -773,6 +803,20 @@ mod test {
     );
   }
 
+  #[test]
+  fn migrate_updater_target() {
```
amrbashir, line 807: let's add a couple more tests for `targets: "updater"` and `targets: "all"`

amrbashir (changes requested):

amrbashir:
As always, thanks for your help

tooling/cli/node/index.d.ts
Legend-Master: We should probably take a look at why building node version of the cli removes those license headers

core/tauri-utils/src/config.rs
Legend-Master: Enable generating updaters only makes sense if they use the updater plugin, I don't think enable should be the default (or maybe we detect plugin updater field), and `V1Compatible` should never be the default, I'm so worried about we will end up having everyone still using `V1Compatible`, and warning on default value is pretty weird as well

Maybe we change the name to `LegacyV1Compatible` so that people will know it's not supposed to be used (maybe some more details in the description), and should really only be there because of `tauri migrate`

core/tauri-utils/src/config.rs
lucasfernog: makes sense to rename it to be more clear

Legend-Master: `createUpdaterArtifacts` sounds good to me

core/tauri-utils/src/config.rs
amrbashir: The problem is in order for apps built with v1, to migrate to updater v2, they need to push zips so v1 can update v2, then they can stop using them

Legend-Master: I already made `tauri migrate` to migrate the old updater to `V1Compatible`, I don't think there's any reason to make it the default, we can emit a warning though

amrbashir: that's what I am suggesting, keep the current default and constantly emit warnings that it will be removed in v3. This allows users to migrate their v1 apps to v2 and then later from v2 to v3 there won't be an issue.

Legend-Master: kk, I can add that warning tomorrow

amrbashir (dismissed):

Legend-Master:
@lucasfernog Please take a look when you got time

.changes/separate-updater-field.md
```diff
@@ -0,0 +1,8 @@
+---
+"tauri-utils": patch:breaking
+"tauri-bundler": patch:breaking
+"tauri-cli": patch:breaking
+"@tauri-apps/cli": patch:breaking
+---
+
+Move updater target from `bundle > targets` to a separate field `bundle > updater`
```
lucasfernog-crabnebula:
```suggestion
Move updater target from `bundle > targets` to a separate field `bundle > createUpdaterArtifacts`
```

lucasfernog (approved):

PR #9909, opened 2024-05-29
fix(nsis): fix regression in shortcuts pointing to a non-existing file

amrbashir:
regression introduced in: https://github.com/tauri-apps/tauri/pull/9375


lucasfernog (dismissed):

amrbashir:
still got one more thing to fix

lucasfernog (approved):

PR #9902, opened 2024-05-29
feat(bundler): support custom sign command on Windows

muwoo:
feat: Migrate v2 functionality to v1.x: support custom sign command on Windows (Discussed in https://github.com/tauri-apps/tauri/discussions/9900)

- [x] cargo test passes
- [x] cargo clippy passes

PR #9914, opened 2024-05-30
fix(windows): Fix Windows System User build failures by using the current directory for bundling tools (fix: 9895)

### LINKED ISSUE #9895: [bug] Building for Windows Fails as System User (e.g., on AWS EC2)
### Describe the bug

Near the end of a Tauri Windows build, when it comes to bundling, Tauri seems to rely on some toolsets it installs in the user's AppData. However, in the context of AWS EC2 instances, build workloads are run under the systemuser. This means that installing these toolsets (e.g., WiX, NSIS) in this directory makes them unusable/executable, due to system/security restriction. The result/error is manifested without clear meaning, whether bundling .exe (using NSIS) or .msi (using WiX).

Only got to this conclusion because if I run the build on the same code in the same place, but as any other user, then Tauri installs the bundling tools in that other user's AppData, and then those bundling tools work perfectly fine for the Tauri build to succeed.

Datapoints:
- Reproduced this issue with the Tauri quickstart project
- The same code in the same directory builds successfully when doing the build as any non-system user. In this case, above toolchains are installed and run from that user's AppData. It only fails for the systemuser wherein above toolchains are installed to the un-usable system's AppData directory
- Regardless of whichever user I am (systemuser, non-system user), I am not able to run the `candle.exe` in that WiX toolset in the systemuser's AppData. In other words, it's not just Tauri that fails to launch it; no user can launch it either.

### Reproduction

On a Windows Machine where you can access as systemuser (e.g., the only option for AWS EC2 workloads)
1. Follow [Tauri QuickStart guide](https://tauri.app/v1/guides/getting-started/setup/) to setup a Tauri project on Windows.
2. Build the application, and get above reported error.
3. [Optional] As another user (non-system), build the same application exactly where it is, and see build succeed. 

### Expected behavior

The bundling toolchains that Tauri installs are expected to be installed at a location where they are usable. This would result in the expected behavior of Tauri build and bundling success, regardless of whether the caller is a systemuser or a non-system user.

At the least, it is expected to be able to configure where the toolchains are installed/referenced.

### Full `tauri info` output

```text
[✔] Environment
     - OS: Windows 10.0.20348 X64
     ✔ WebView2: 125.0.2535.67
     ✔ MSVC:
     ✔ rustc: 1.78.0 (9b00956e5 2024-04-29)
     ✔ cargo: 1.78.0 (54d8815d0 2024-03-26)
     - node: 18.20.3
     - npm: 10.7.0
 
 [-] Packages
     - tauri [RUST]: 1.6.1
     - tauri-build [RUST]: 1.5.1
     - wry [RUST]: 0.24.7
     - tao [RUST]: 0.16.7
     - @tauri-apps/api [NPM]: 1.5.3 (outdated, latest: 1.5.6)
     - @tauri-apps/cli [NPM]: 1.5.14
 
 [-] App
     - build-type: bundle
     - CSP: unset
     - distDir: ../dist
     - framework: React
     - bundler: Vite
```

### Stack trace

Initial build logs (e.g., `TAURI_TARGET_TRIPLE`):

```
Running [tauri_cli::build] beforeBuildCommand `npm run build-app`
        Debug [tauri_cli::build] Setting environment for hook {"TAURI_ARCH": "x86_64", "TAURI_TARGET_TRIPLE": "x86_64-pc-windows-msvc", "TAURI_FAMILY": "windows", "TAURI_PLATFORM_VERSION": "10.0.20
 348", "TAURI_PLATFORM": "windows", "TAURI_PLATFORM_TYPE": "Windows_NT"}
      Running [tauri_cli] Command `cmd  /S /C npm run build-app`
```

Problem with NSIS for .exe bundling: 

```
...{bunch of random, non-error debug logs}
Running [tauri_bundler::bundle::windows::nsis] makensis.exe to produce {omitted path}_x64-setup
 .exe
      Running [tauri_bundler::bundle::common] Command `C:\Windows\system32\config\systemprofile\AppData\Local\tauri\NSIS\makensis.exe  -V3 {truncated parent path}\src-tauri\target\release\nsis\x64\installer.nsi`
 Unable to start child process, error 0x2
        Error [tauri_cli_node] failed to bundle project: `The system cannot find the file specified. (os error 2)`
```

Note on above: the file does exist (just can't be run).

Problem with WiX for .msi bundling:

```
...{bunch of random, non-error debug logs}
      Running [tauri_bundler::bundle::windows::msi::wix] candle for "main.wxs"
      Running [tauri_bundler::bundle::common] Command `C:\Windows\system32\config\systemprofile\AppData\Local\tauri\WixTools\candle.exe  -arch x64 main.wxs -dSourceDir={omitted}.exe`
        Error [tauri_cli_node] failed to bundle project: error running candle.exe
```

### Additional context

I know that there are GitHub actions for cross-compilation, but that isn't an option for us. Besides, this problem is not necessarily restricted to AWS EC2, as it would apply to any Windows host where systemuser is used for building.

AnthonyNGarcia:
Closes #9895 with details on the issue there.

As a summary: Tauri uses Wix/NSIS toolsets for bundling .msi and .exe files respectively. Currently, it installs/retrieves these toolchains from the user app directory. However, this approach fails if the current user is a System user, since their app directory (the system directory) is restricted, which results in the bundling failing due to unusable Wix/NSIS executables.

This PR solves the problem by instead installing/retrieving the Wix/NSIS toolset from the current project directory, in `target/tools`. For example, Wix would be at `target/tools/Wix`. This would:

- Be a known location that should be accessible, since it is the current project's workspace.
- Enable deterministic builds across different Windows Users, since the tools are in the project's workspace rather than being user-specific
- Enable controlled builds for the same Windows User across different Tauri projects, in case support comes later to use a specific version of these tools for a specific project.
- Better support as-is offline builds wherein these tools can be configured in the project's workspace ahead-of-time (rather than configuring them on-the-fly).

Tested this code change locally:
- `cargo test`
- `cargo clippy`
- Using it on my Windows machine with Tauri Quickstart guide example, to generate .msi and .exe (tests Wix and NSIS flows, respectively). They both built successfully. Also confirmed both of them work for app installation

### DIFF
--- a/.changes/bundler-local-tools-directory.md
+++ b/.changes/bundler-local-tools-directory.md
@@ -1,0 +1,5 @@
+---
+'tauri-bundler': 'patch:feat'
+---
+
+Add `Settings::local_tools_directory`.
--- a/.changes/cli-use-local-tools-directory.md
+++ b/.changes/cli-use-local-tools-directory.md
@@ -1,0 +1,7 @@
+---
+"tauri-cli": "patch:feat"
+"@tauri-apps/cli": "patch:feat"
+---
+
+Use cargo's target directory to store and cache bundling tools when `bundle > useLocalToolsDir` option is active.
+
--- a/.changes/fix-windows-build-failures.md
+++ b/.changes/fix-windows-build-failures.md
@@ -1,0 +1,5 @@
+---
+'tauri-bundler': 'patch:bug'
+---
+
+Fixes build failures on Windows when building as a System User.
--- a/.changes/fix-windows-build-failures.md
+++ b/.changes/fix-windows-build-failures.md
@@ -1,5 +1,0 @@
----
-'tauri-bundler': 'patch:bug'
----
-
-Fixes build failures on Windows when building as a System User.
--- a/.changes/utils-use-local-tools-directory.md
+++ b/.changes/utils-use-local-tools-directory.md
@@ -1,0 +1,5 @@
+---
+'tauri-utils': 'patch:feat'
+---
+
+Add `use_local_tools_dir` option.
--- a/core/tauri-config-schema/schema.json
+++ b/core/tauri-config-schema/schema.json
@@ -150,6 +150,7 @@
             "release": "1"
           },
           "targets": "all",
+          "useLocalToolPath": false,
           "windows": {
             "allowDowngrades": true,
             "certificateThumbprint": null,
@@ -157,7 +158,6 @@
             "nsis": null,
             "timestampUrl": null,
             "tsp": false,
-            "useLocalToolPath": false,
             "webviewFixedRuntimePath": null,
             "webviewInstallMode": {
               "silent": true,
@@ -295,6 +295,7 @@
               "release": "1"
             },
             "targets": "all",
+            "useLocalToolPath": false,
             "windows": {
               "allowDowngrades": true,
               "certificateThumbprint": null,
@@ -302,7 +303,6 @@
               "nsis": null,
               "timestampUrl": null,
               "tsp": false,
-              "useLocalToolPath": false,
               "webviewFixedRuntimePath": null,
               "webviewInstallMode": {
                 "silent": true,
@@ -1124,6 +1124,11 @@
             "null"
           ]
         },
+        "useLocalToolPath": {
+          "description": "Whether to use the project's local workspace or the current user's Tauri app data cache, for managing build tools (e.g., Wix) when building this application.\n\nIf true, installs and uses bundling tools (e.g., Wix) in the project's local workspace, under `target\\.tauri-tools`. If false, these tools are cached in the current user's platform-specific app data directory.\n\nAn example where it can be appropriate to set this to `true` is when building this application as a Windows System user (e.g., AWS EC2 workloads), because the Window system's app data directory is restricted.",
+          "default": false,
+          "type": "boolean"
+        },
         "appimage": {
           "description": "Configuration for the AppImage bundle.",
           "default": {
@@ -1189,7 +1194,6 @@
             "nsis": null,
             "timestampUrl": null,
             "tsp": false,
-            "useLocalToolPath": false,
             "webviewFixedRuntimePath": null,
             "webviewInstallMode": {
               "silent": true,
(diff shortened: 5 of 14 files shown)


amrbashir (changes requested):
We used to use `target` directory before but we changed it to use `cache_dir` because these files rarely change so it makes sense to just cache them for every project and the few files that change, we detect that and update them.

That said, I'd like to keep the current behavior, and add this either as a fallback when current cache_dir fails, or as a separate CLI flag (or environment flag)

FabianLars:
Agreed with Amr, just wanted to mention that this would need a proper target dir detection (similar to how the `bundle/` folder gets placed) because `<cwd>/target/` is simply incorrect tbh.

AnthonyNGarcia:
Great callout, thank you folks. I like the suggestion to keep the existing behavior as default given it is more optimal for most use cases. I've put the feature behind a feature flag in the Tauri config json with schema validation.  I believe this approach accomplishes the best of both worlds: gating the feature, but also minimizing risk of build errors associated with CLI usage and flags, by leveraging the feature via schema validation. After all, someone who needs this feature needs this all the time, so setting it once in the config and being safe would be preferred (at least speaking from the voice of one such person who wants to consume this fix).

Tested that the new json config property is fully optional. If not provided at all, defaults to false (aka existing behavior). Also tested if false is explicitly provided (again, existing behavior). Finally, when true is provided (new behavior, working).

As for the callout on getting to `target` directory more correctly, good callout. I updated it to traverse up one from the available `project_out_directory`. Do let me know if you are aware of a better way to get to the target directory. I also considered just putting these tools directly in `project_out_directory` but it name conflicts with existing folders there `wix` and `nsis` which are holding intermediate artifacts.

AnthonyNGarcia:
As per offline discussion in Discord, moved the feature from `bundle.windows` up to `bundle` directly as it could apply to other platforms (though this PR just implements it for Windows). Updated docs accordingly and renamed tool folder from `tools` -> `.tauri-tools`

FabianLars:
> I also considered just putting these tools directly in project_out_directory but it name conflicts with existing folders there wix and nsis which are holding intermediate artifacts.

Hmm, maybe something like this (copy pasted from somewhere else in the cli)
```rs
  let mut path = get_cargo_metadata()
    .with_context(|| "failed to get cargo metadata")?
    .target_directory;
```

Using the parent dir of project_out_directory is probably a bit awkward because it will be different if the user uses the `--target` flag, duplicating the tools at which point you may as well use project_out_directory directly (with the extra .tauri-tools folder) just to be safe.

amrbashir (approved):

MystiPanda:
I just tested this PR and didn't make any changes to `tauri.conf.json`, I got permissions error on GitHub Actions:
https://github.com/clash-verge-rev/clash-verge-rev/actions/runs/9360177885/job/25765128485
image

MystiPanda:
Well, unrelated, I found out that #9902 caused the problem.

PR #9915, opened 2024-05-30
fix(nsis): migrate old shortcuts by setting the path only instead of re-creating a new one

Legend-Master:
References: #9375, #9909

### DIFF
--- a/.changes/nsis-migrate-shortcut.md
+++ b/.changes/nsis-migrate-shortcut.md
@@ -1,5 +1,5 @@
 ---
-"tauri-bundler": "patch:bug"
+"tauri-bundler": "patch:enhance"
 ---
 
 For NSIS installer, migrate old shortcuts by setting the path only instead of re-creating a new one
--- a/.changes/nsis-migrate-shortcut.md
+++ b/.changes/nsis-migrate-shortcut.md
@@ -1,0 +1,5 @@
+---
+"tauri-bundler": "patch:bug"
+---
+
+Migrate old shortcuts by setting the path only instead of re-creating a new one
--- a/.changes/nsis-migrate-shortcut.md
+++ b/.changes/nsis-migrate-shortcut.md
@@ -2,4 +2,4 @@
 "tauri-bundler": "patch:bug"
 ---
 
-Migrate old shortcuts by setting the path only instead of re-creating a new one
+For NSIS installer, migrate old shortcuts by setting the path only instead of re-creating a new one
--- a/tooling/bundler/src/bundle/windows/templates/installer.nsi
+++ b/tooling/bundler/src/bundle/windows/templates/installer.nsi
@@ -347,7 +347,7 @@ Var AppStartMenuFolder
 !define MUI_FINISHPAGE_SHOWREADME_FUNCTION CreateDesktopShortcut
 ; Show run app after installation.
 !define MUI_FINISHPAGE_RUN "$INSTDIR\${MAINBINARYNAME}.exe"
-!define MUI_PAGE_CUSTOMFUNCTION_PRE SkipIfPassiveButUpdateShortcutIfUpdate
+!define MUI_PAGE_CUSTOMFUNCTION_PRE SkipIfPassive
 !insertmacro MUI_PAGE_FINISH
 
 ; Uninstaller Pages
@@ -590,6 +590,8 @@ Section Install
     Call CreateDesktopShortcut
   shortcut_done:
 
+  Call MigrateProductNameExeShortcuts
+
   !ifdef NSIS_HOOK_POSTINSTALL
     !insertmacro "${NSIS_HOOK_POSTINSTALL}"
   !endif
@@ -727,22 +729,25 @@ Function SkipIfPassive
   ${IfThen} $PassiveMode = 1  ${|} Abort ${|}
 FunctionEnd
 
-Function SkipIfPassiveButUpdateShortcutIfUpdate
-  ${If} $PassiveMode = 1
-    Call CreateDesktopShortcut
-    Abort
+Function CreateStartMenuShortcut
+  ; Skip creating shortcut if in update mode or no shortcut mode
+  ${If} $UpdateMode = 1
+    Return
   ${EndIf}
+  ${If} $NoShortcutMode = 1
+    Return
+  ${EndIf}
+
+  CreateDirectory "$SMPROGRAMS\$AppStartMenuFolder"
+  CreateShortcut "$SMPROGRAMS\$AppStartMenuFolder\${PRODUCTNAME}.lnk" "$INSTDIR\${MAINBINARYNAME}.exe"
+  !insertmacro SetLnkAppUserModelId "$SMPROGRAMS\$AppStartMenuFolder\${PRODUCTNAME}.lnk"
 FunctionEnd
 
 Function CreateDesktopShortcut
-  ; Skip creating shortcut if in update mode
-  ; and shortcuts doesn't exist, which means user deleted it
-  ; so we respect that.
+  ; Skip creating shortcut if in update mode or no shortcut mode
   ${If} $UpdateMode = 1
-    IfFileExists "$DESKTOP\${PRODUCTNAME}.lnk" +2 0
-      Return
+    Return
   ${EndIf}
-
   ${If} $NoShortcutMode = 1
     Return
   ${EndIf}
@@ -751,19 +756,11 @@ Function CreateDesktopShortcut
   !insertmacro SetLnkAppUserModelId "$DESKTOP\${PRODUCTNAME}.lnk"
 FunctionEnd
 
-Function CreateStartMenuShortcut
-  ; Skip creating shortcut if in update mode.
-  ; See `CreateDesktopShortcut` above.
-  ${If} $UpdateMode = 1
-    IfFileExists "$SMPROGRAMS\$AppStartMenuFolder\${PRODUCTNAME}.lnk" +2 0
-      Return
+Function MigrateProductNameExeShortcuts
+  ${If} ${FileExists} "$SMPROGRAMS\$AppStartMenuFolder\${PRODUCTNAME}.lnk"
+    SetShortcutTarget "$SMPROGRAMS\$AppStartMenuFolder\${PRODUCTNAME}.lnk" "$INSTDIR\${MAINBINARYNAME}.exe"
   ${EndIf}
-
-  ${If} $NoShortcutMode = 1
-    Return
+  ${If} ${FileExists} "$DESKTOP\${PRODUCTNAME}.lnk"
+    SetShortcutTarget "$DESKTOP\${PRODUCTNAME}.lnk" "$INSTDIR\${MAINBINARYNAME}.exe"
   ${EndIf}
-
-  CreateDirectory "$SMPROGRAMS\$AppStartMenuFolder"
-  CreateShortcut "$SMPROGRAMS\$AppStartMenuFolder\${PRODUCTNAME}.lnk" "$INSTDIR\${MAINBINARYNAME}.exe"
-  !insertmacro SetLnkAppUserModelId "$SMPROGRAMS\$AppStartMenuFolder\${PRODUCTNAME}.lnk"
 FunctionEnd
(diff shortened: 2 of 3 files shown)


.changes/nsis-migrate-shortcut.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri-bundler": "patch:bug"
```
amrbashir:
```suggestion
"tauri-bundler": "patch:enhance"
```

tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -347,7 +347,7 @@ Var AppStartMenuFolder
 !define MUI_FINISHPAGE_SHOWREADME_FUNCTION CreateDesktopShortcut
 ; Show run app after installation.
 !define MUI_FINISHPAGE_RUN "$INSTDIR\${MAINBINARYNAME}.exe"
-!define MUI_PAGE_CUSTOMFUNCTION_PRE SkipIfPassiveButUpdateShortcutIfUpdate
+!define MUI_PAGE_CUSTOMFUNCTION_PRE SkipIfPassive
```
amrbashir, line 350: why was this changed?

tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -577,6 +577,10 @@ Section Install
   WriteRegDWORD SHCTX "${UNINSTKEY}" "NoRepair" "1"
   WriteRegDWORD SHCTX "${UNINSTKEY}" "EstimatedSize" "${ESTIMATEDSIZE}"
 
+  ; We used to use product name as MAINBINARYNAME
+  ; migrate old shortcuts to target the new MAINBINARYNAME
+  Call MigrateProductNameExeShortcuts
```
amrbashir: we don't need to create a separate function for migration, in `CreateStartMenuShortcut` and `CreateDesktopShortcut` just check if it exists and update it or re-create it

amrbashir (changes requested):

tooling/bundler/src/bundle/windows/templates/installer.nsi
Legend-Master: wasn't this change for migrating shortcuts? since we do it on install now, no need to do it here anymore

tooling/bundler/src/bundle/windows/templates/installer.nsi
Legend-Master: it doesn't make sense to be in create shortcuts functions, we might skip those functions (we might not currently, but from the name it's not clear it's also responsible for migrating old shortcuts)

tooling/bundler/src/bundle/windows/templates/installer.nsi
amrbashir: No, this is for passive mode installer where it is not necessarily an update

amrbashir: Actually I realize my implementation and name was wrong

tooling/bundler/src/bundle/windows/templates/installer.nsi
amrbashir: The name could change, and instead of multiple functions we can just put all logic related to a shortcut inside one function, so we always call it but the logic inside it decides whether to create, update or skip

tooling/bundler/src/bundle/windows/templates/installer.nsi
amrbashir: Let's ignore this and remove, my brain can't handle more NSIS this week

Legend-Master: isn't passive mode installer handled in install, line 592?

tooling/bundler/src/bundle/windows/templates/installer.nsi
Legend-Master: kk

Legend-Master: Wait, we still need to separate start menu shortcut from desktop one, they have different configs from UIs

tooling/bundler/src/bundle/windows/templates/installer.nsi
amrbashir: yeah I think so, but I remember I had a case where ot wasn't covered by that and I added that code, I may have changed the code a bit after that. I really don't remember, let's just move on

tooling/bundler/src/bundle/windows/templates/installer.nsi
amrbashir: Yeah 2 functions, one for desktop and one for start menu. I think we need to extract some of the logic into another macro

Legend-Master: To be honest, I still don't think it make much sense to put migration logic in `CreateStartMenuShortcut`/`CreateDesktopShortcut`, since this is not about creating the shortcut at all (changing those function's name to something like `HandleStartMenuShortcut` doesn't make much sense to me either), migration should just be a separate function (if you don't like a function we can use a macro but I don't think that's what you mean)

amrbashir: Problem is to call said functions, we will do the conditional at the call site which sure is okay but bloats the installation section and I would love to hide these in a function so it is easier to focus on just that function and understand the logic without need of outer context that much.

Legend-Master: Do you mean extract this to a function?

https://github.com/tauri-apps/tauri/blob/040289ea52467c423bd895026d891553409e862a/tooling/bundler/src/bundle/windows/templates/installer.nsi#L582-L595

amrbashir: I think I would like it to be
```nsi 
   ; Create start menu shortcut 
   !insertmacro MUI_STARTMENU_WRITE_BEGIN Application 
     Call CreateOrUpdateStartMenuShortcut 
   !insertmacro MUI_STARTMENU_WRITE_END 
  
   ; Create desktop shortcut
   Call CreateOrUpdateDesktopShortcut
```

and each of these functions would hide the implementation inside it

Legend-Master: This wouldn't work, start menu one will be skipped on update

Migration logic here is very different than creating the shortcuts in this context, I suggest we keep shortcut migration code at a central place

amrbashir: > This wouldn't work, start menu one will be skipped on update

Why would it be skipped?

Legend-Master: Doesn't this skip the inner block if no start menu page present (e.g. on update)?

```nsis
   !insertmacro MUI_STARTMENU_WRITE_BEGIN Application 
     Call CreateOrUpdateStartMenuShortcut 
   !insertmacro MUI_STARTMENU_WRITE_END 
```

amrbashir: it still calls it, we just skipped page, so think of it as the page was shown for a split second and then clicked next immediately, so the default values are used.

Legend-Master: You're right, I thought skip if passive means not to run that part, it's actually more or less the same as clicking the next button

But still, it will not be the case anymore once we add in an option to disable that page, I still think put the migration logic together make a lot more sense than putting them in create functions (yes, the name can change, but the caller needs to be more cautious about skipping those function)

amrbashir: That's what I am trying to avoid, by always calling them and let the function itself handle the skipping.

Legend-Master:
```nsis
  ; Create start menu shortcut
  !insertmacro MUI_STARTMENU_WRITE_BEGIN Application
    Call CreateStartMenuShortcut
  !insertmacro MUI_STARTMENU_WRITE_END

  ; Create desktop shortcut for silent and passive installers
  ; because finish page will be skipped
  IfSilent create_shortcut 0
  StrCmp $PassiveMode "1" create_shortcut shortcut_done
  create_shortcut:
    Call CreateDesktopShortcut
  shortcut_done:
```

This part itself does some checking and might skip it (it might not, but let's just get it done, in the most clear way, I don't want to play with nsis anymore)

NSIS itself is so error prone (that's why we use rust to write most things not nsi script), I really don't want to rely on those things to work, calling migration here is most clear way of always running it

amrbashir: That's the thing we don't always need to migrate, like normal install without update should always recreate not update.

Legend-Master: We should always migrate, even the user checks no desktop shortcut at the end, we should still update existing shortcuts

amrbashir: I don't agree with that, we shouldn't update if we are not in update mode, otherwise just recreate, reason being if somehow the user messed up their shortcuts and wasn't very experienced to fix it, they will usually resort to reinstalling which should recreate the shortcut and fix it. They could go through the trouble of uninstalling then installing but that could potentially be more destructive.

Legend-Master: Current implementation will do exactly that, if user re-runs the installer, it will first migrate the shortcut, and then re-create it, if the user check no update mode, it will only do the migration

amrbashir: I know and that is redundant, we could be avoiding that extra work.

Legend-Master: Alright, let's don't get stuck here debating, I still think it's better to put migration code in a function, but I'll change it to doing it in create shortcut function

tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -727,43 +736,39 @@ Function SkipIfPassive
...
 FunctionEnd
 
-Function SkipIfPassiveButUpdateShortcutIfUpdate
-  ${If} $PassiveMode = 1
-    Call CreateDesktopShortcut
-    Abort
+Function CreateOrUpdateStartMenuShortcut
+  ; We used to use product name as MAINBINARYNAME
+  ; migrate old shortcuts to target the new MAINBINARYNAME
+  ${If} ${FileExists} "$DESKTOP\${PRODUCTNAME}.lnk"
```
amrbashir, line 742: We should only update if in update mode, otherwise recreate the shortcut

```suggestion
  ; Update shortcuts target path if updating
  ${If} $UpdateMode = 1
  ${AndIf} ${FileExists} "$DESKTOP\${PRODUCTNAME}.lnk"
```

tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -727,43 +736,39 @@ Function SkipIfPassive
...
-Function CreateStartMenuShortcut
-  ; Skip creating shortcut if in update mode.
-  ; See `CreateDesktopShortcut` above.
-  ${If} $UpdateMode = 1
-    IfFileExists "$SMPROGRAMS\$AppStartMenuFolder\${PRODUCTNAME}.lnk" +2 0
-      Return
+Function CreateOrUpdateDesktopShortcut
+  ; We used to use product name as MAINBINARYNAME
+  ; migrate old shortcuts to target the new MAINBINARYNAME
+  ${If} ${FileExists} "$SMPROGRAMS\$AppStartMenuFolder\${PRODUCTNAME}.lnk"
```
amrbashir, line 761: ditto
```suggestion
  ; Update shortcuts target path if updating
  ${If} $UpdateMode = 1
  ${AndIf} ${FileExists} "$SMPROGRAMS\$AppStartMenuFolder\${PRODUCTNAME}.lnk"
```

amrbashir (changes requested):

tooling/bundler/src/bundle/windows/templates/installer.nsi
Legend-Master: We don't pass in /update in the earlier versions, and this migration is exactly for those versions, we shouldn't re-create if /update is not present in the command line args

amrbashir: right, I forgot if we decided on a solution for this tbh, let's roll with this solution for now as it seems the safest

amrbashir (approved):

PR #9923, opened 2024-05-30
fix: workaround for schemars' improper handling of codeblocks

simonhyll:
There were more instances of code blocks being improperly handled.


amrbashir:
maybe it is better to handle that when processing the schema, because I am afraid this will affect how these blocks are rendered in IDE hover documentation.

simonhyll:
It's definitely better handled elsewhere, this is more of a quick fix to make sure the documentation gets built correctly pending a proper upstream fix

amrbashir:
does this affect other code blocks? or just this one?

simonhyll:
Not sure how widespread the issue is. All I know is that this affects the config reference being built properly and that these are the final places that affects that reference. I think we'll need to either patch `schemars` or add a custom implementation for serde in order to fix this issue throughout the entire codebase.

amrbashir:
There are only 3 places that use multi line code blocks which you had to modify, this solution only fixes the current ones and doesn't account for future uses.

I believe the solution here would be to either fix `schemars` upstream or when processing the schema file in `tauri-docs` we manually insert these new lines.

simonhyll:
Yea it's definitely better handled as an actual fix upstream to e.g. `schemars`. I wouldn't do it in `tauri-docs` because it's not that the docs can't parse it correctly, it's that the generated codeblocks are wrong in `schema.json`. If someone else were to depend on Tauri in order to get those schema files then they'd also face issues because the docstrings are plain wrong. So if we're doing a workaround anywhere I'd do it in the function where we save the schema files instead, it's just as annoying to deal with adding the newlines in Rust as it is in JavaScript just that in Rust we'll be actually fixing the schema files themselves at the source.

And after fixing it in `tauri` the end result will be these schema files, so might as well commit the workaround pending a more permanent fix.

amrbashir:
right, forgot about tauri-config-schema, then I'd request modifying that instead of the rust docstring

PR #9932, opened 2024-05-31
feat(nsis): add an option to disable compression

Legend-Master:

### DIFF
--- a/.changes/nsis-no-compression.md
+++ b/.changes/nsis-no-compression.md
@@ -1,8 +1,6 @@
 ---
-"tauri-utils": patch::breaking
-"tauri-bundler": patch::breaking
-"tauri-cli": patch::breaking
-"@tauri-apps/cli": patch::breaking
+"tauri-utils": patch::feat
+"tauri-bundler": patch::feat
 ---
 
-Add an option to disable NSIS compression
+Add an option to disable NSIS compression `bundle > nsis > compression: "none"`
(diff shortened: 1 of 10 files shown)


amrbashir (commented):
I am not sure what this PR is trying to solve, it is more or less looks like a refactor to me.

Legend-Master:
Before

image

After

image

amrbashir:
That still not an issue, if user specifies null we default back to `lzma`, or maybe we need to actually make `null` disable compression

Legend-Master:
Isn't this about correcting the schema?

amrbashir:
You're right, but I think the schema is kinda correct, maybe just missing a documentation to explain that using `null` will default to `lzma`.

This PR has a breaking change because it changed the field from `Option<T>` to `T` and I am afraid some users may want to be able to disable compression and if that were to be requested, we will have to make another breaking change, so maybe we should take the opportunity to cover that case as well?

Legend-Master:
Right, we can include no compression as an option here

.changes/nsis-no-compression.md
```diff
@@ -0,0 +1,8 @@
+---
+"tauri-utils": patch::breaking
+"tauri-bundler": patch::breaking
+"tauri-cli": patch::breaking
+"@tauri-apps/cli": patch::breaking
+---
+
+Add an option to disable NSIS compression
```
amrbashir:
```suggestion
---
"tauri-utils": patch::feat
"tauri-bundler": patch::feat
---

Add an option to disable NSIS compression `bundle > nsis > compression: "none"`
```

amrbashir (approved):

amrbashir:
Thank you

PR #9944, opened 2024-06-02
fix(nsis): run app as normal user

### LINKED ISSUE #9835: [bug] application launched from perMachine or both nsis installer still have admin privilege (elevated)
### Describe the bug

application launched from perMachine or both nsis installer still have admin privilege (elevated).

It's not good to have admin privilege and some process (like Unity) denies launching admin privilege.

### Reproduction

1. install some tauri app with nsis installer
2. launch the app from nsis installer
3. the launched app process is elevated

### Expected behavior

the app launched from installer should not elevated.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.22631 X64
    ✔ WebView2: 124.0.2478.97
    - node: 20.11.1
    - npm: 10.2.4

    - CSP: unset
    - distDir: out
    - framework: React (Next.js)
    - bundler: Webpack
```

### Stack trace

_No response_

### Additional context

as far as i search, this method seems to work but I didn't test yet https://stackoverflow.com/questions/11169431/how-to-start-a-new-process-without-administrator-privileges-from-a-process-with

Downstream issue: vrc-get/vrc-get#1020

Legend-Master:
Fixes: #9835

Reference: https://devblogs.microsoft.com/oldnewthing/20190425-00/?p=102443


Legend-Master:
Don't want to solve merge conflicts, waiting for https://github.com/tauri-apps/tauri/pull/9915 to get merged first

amrbashir (changes requested):
I tried so hard to review this PR and even though you did a great job documenting each call, I still find it error prone and unmaintainable. However, I couldn't find an easier alternative.

I think the right thing to do this logic in Rust in `nsis-tauri-utils` as part of the `nsis-process` plugin.

Would you like to work on that @Legend-Master?

Legend-Master:
I could give it a try, didn't work on no std rust before, but guess it wouldn't be harder than nsis 😂

amrbashir:
Don't worry I think the existing functions there can guide you but If you need help, just let me know

tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -610,7 +615,7 @@ Function .onInstSuccess
     ${GetOptions} $CMDLINE "/R" $R0
     ${IfNot} ${Errors}
       ${GetOptions} $CMDLINE "/ARGS" $R0
-      Exec '"$INSTDIR\${MAINBINARYNAME}.exe" $R0'
+      nsis_tauri_utils::RunAsUser "$INSTDIR\${MAINBINARYNAME}.exe" "$R0"
```
amrbashir, line 618:
```suggestion
      Call RunMainBinary
```

tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -346,10 +346,15 @@ Var AppStartMenuFolder
...
 ; Show run app after installation.
-!define MUI_FINISHPAGE_RUN "$INSTDIR\${MAINBINARYNAME}.exe"
+!define MUI_FINISHPAGE_RUN
+!define MUI_FINISHPAGE_RUN_FUNCTION RunMainBinary
 !define MUI_PAGE_CUSTOMFUNCTION_PRE SkipIfPassive
 !insertmacro MUI_PAGE_FINISH
 
+Function RunMainBinary
+  nsis_tauri_utils::RunAsUser "$INSTDIR\${MAINBINARYNAME}.exe" ""
+FunctionEnd
```
amrbashir, line 356: let's move this next to the functions at the end of file

amrbashir (changes requested):

tooling/bundler/src/bundle/windows/templates/installer.nsi
Legend-Master: I don't want to use `Push $R0` to be honest, it's pretty error prone, and since we need to do pop in `RunMainBinary`, I think we need another wrapper just to get `MUI_FINISHPAGE_RUN_FUNCTION` to work

amrbashir: I see, there is a hacky way to avoid it by always using $R9 as the second argument in the function which should be empty for GUI and when checking `/R`, instead of reading into $R0, we read into $R9, but that's not an ideal solution and $R9 could be used later so let's keep it this way for now or convert it into a macro later

amrbashir (approved):

PR #9952, opened 2024-06-03
feat: improve deserialization errors by using serde-untagged on enums

### LINKED ISSUE #9951: [feat] Better errors when deserializing capabilities
### Describe the problem

I missed in the docs (if it's there at all) that plugin names can't have underscores in them. I would have figured this out pretty quickly had the error message for serializing capabilities/default.json been more specific. However, all I got from the build script was
```
thread 'main' panicked at trellis/src-tauri/build.rs:14:6:
  failed to run tauri-build: failed to parse JSON: data did not match any variant of untagged enum CapabilityFile
```

Normally, serde would give a more specific error, but it can't because it doesn't know which variant of the untagged enum to compare the file to. Based on https://github.com/serde-rs/serde/issues/2157#issuecomment-1696114422, this could be mitigated by using https://docs.rs/serde-untagged/latest/serde_untagged/index.html.

I spun up a quick-and-dirty implementation for only `/core/tauri-utils/src/acl/capability.rs`. I'll open a PR for it in a moment, but it obviously needs work. Before it's merged (if it is going in the right direction), we would need to find all similar instances present in other files.

The result is an error like this:
```
  thread 'main' panicked at build.rs:11:6:
  failed to run tauri-build: failed to parse JSON: identifiers can only include lowercase ASCII, hyphens which are not leading or trailing, and a single colon if using a prefix at line 16 column 23
```
It's not perfect; it would also be nice to wrap this error with some information about which file is throwing the error.

Also, it would be nice if the "no underscore" requirement for plugin names is stated in the docs. I think it might go nicely here: https://v2.tauri.app/develop/plugins/, but I don't have time to make another PR.

### Alternatives considered

Besides writing our own deserializer without `serde_untagged`, I don't know of any other options.

### Additional context

_No response_

jsimonrichard:
This is the quick-and-dirty PR mentioned in #9951. The goal is to improve serde deserialization errors by implementing Deserialize using the crate [serde_untagged](https://docs.rs/serde_untagged). This allows serde to avoid the ambiguity introduced by using untagged enums. Without it, errors simply state: `data did not match any variant of untagged enum ...`. With this code, errors are more specific. For example:
```
  thread 'main' panicked at build.rs:11:6:
  failed to run tauri-build: failed to parse JSON: identifiers can only include lowercase ASCII, hyphens which are not leading or trailing, and a single colon if using a prefix at line 16 column 23
```

There are a few to-do items:
- [ ] Check the code base for similar usage of untagged enums and determine if `serde_untagged` should be used.
- [ ] Wrap the error above (see #9951 for details) with information about which file is producing that error.
- [ ] There are a few code smells: namely, structs duplicating the contents of `PermissionEntry::ExtendedPermission` and `CapabilityFile::NamedList` so that they can be used with the `Deserialize` derive trait. 

I don't have any time to work on this further (at least for now), but I wanted to get it into your hands just in case you'd like to do something with it.


core/tauri-utils/src/acl/capability.rs
```diff
@@ -129,6 +159,33 @@ impl CapabilityFile {
...
+impl<'de> Deserialize<'de> for CapabilityFile {
+  fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+  where
+    D: Deserializer<'de>,
+  {
+    UntaggedEnumVisitor::new()
+      .seq(|seq| seq.deserialize::<Vec<Capability>>().map(Self::List))
+      .map(|map| {
+        if let Some(len) = map.size_hint() {
+          if len == 1 {
```
amrbashir: instead of checking the length, we could just try to deserialize as `CapabilityNamedList` and if it fails, we fallback to deserializing as `Capability`? I know this is less performant but should cover the case where `map.size_hint()` is None.

core/tauri-utils/Cargo.toml
```diff
@@ -42,6 +42,7 @@ infer = "0.15"
 dunce = "1"
 log = "0.4.21"
 cargo_metadata = { version = "0.18", optional = true }
+serde-untagged = "0.1.6"
```
amrbashir:
```suggestion
serde-untagged = "0.1"
```

amrbashir (changes requested):
This is an awesome PR. I left a few comments but if you don't have enough time to work on this, just let me know and I can take over. 

thanks for contributing :heart:

jsimonrichard:
> This is an awesome PR. I left a few comments but if you don't have enough time to work on this, just let me know and I can take over.

Take it away. You should be able to make changes to my repo, but if not let me know.

amrbashir:
I think I do have access, Thank you @jsimonrichard.

core/tauri-utils/src/acl/capability.rs
amrbashir: looks like we can't because trying `.deserialize` takes ownership of the value and we can't retry again

amrbashir (dismissed):

lucasfernog (approved):

PR #9954, opened 2024-06-04
fix(core): add std feature to raw-window-handle

Legend-Master:
Add std feature to raw-window-handle crate so that no default features can work

Reference: https://github.com/tauri-apps/plugins-workspace/actions/runs/9334978097/job/25730280072?pr=1402

### DIFF
--- a/.changes/raw-window-handle-std.md
+++ b/.changes/raw-window-handle-std.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch:deps
+---
+
+Add std feature to raw-window-handle crate so that no default features can work
--- a/.changes/raw-window-handle-std.md
+++ b/.changes/raw-window-handle-std.md
@@ -1,5 +1,5 @@
 ---
-"tauri": patch:deps
+"tauri": patch:bug
 ---
 
-Add std feature to raw-window-handle crate so that no default features can work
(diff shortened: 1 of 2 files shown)


.changes/raw-window-handle-std.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri": patch:deps
+---
+
+Add std feature to raw-window-handle crate so that no default features can work
```
amrbashir:
```suggestion
---
"tauri": patch:bug
---

Add `std` feature to `raw-window-handle` crate so that using `default-features = false` on `tauri` crate can work
```

amrbashir (changes requested):
I am still not sure how does enabling `no-default-features` affects `raw-window-handle` and I also can't reproduce the bug locally

amrbashir:
NVM, I was using the wrong name for `default-features`, still not sure how are they related.

amrbashir (approved):

Legend-Master:
`tauri` uses `thiserror` to forward `raw-window-handle`'s error, but this is only doable with `std` feature, currently we have it on `tauri-runtime-wry` which has `tao` which has `raw-window-handle` with `std` feature, but with not default features, we no longer depends on `tauri-runtime-wry` -> `tao`

PR #9956, opened 2024-06-04
fix(bundler): fix error when renaming the downloaded zip of NSIS 

Pylogmon:
This PR fixes the issue mentioned in https://github.com/tauri-apps/tauri/pull/9902#pullrequestreview-2095258915. Creating an `NSIS` folder in advance will cause subsequent renaming failures after extract. Additionally, the `Plugins/x86-unicode` directory already exists in the extracted directory and does not need to be created again.

### DIFF
--- a/tooling/bundler/src/bundle/windows/nsis.rs
+++ b/tooling/bundler/src/bundle/windows/nsis.rs
@@ -122,12 +122,10 @@ fn get_and_extract_nsis(nsis_toolset_path: &Path, _tauri_tools_path: &Path) -> c
     NSIS_TAURI_UTILS_SHA1,
     HashAlgorithm::Sha1,
   )?;
-  write(
-    nsis_plugins
-      .join("x86-unicode")
-      .join("nsis_tauri_utils.dll"),
-    data,
-  )?;
+
+  let target_folder = nsis_plugins.join("x86-unicode");
+  create_dir_all(&target_folder)?;
+  write(target_folder.join("nsis_tauri_utils.dll"), data)?;
 
   Ok(())
 }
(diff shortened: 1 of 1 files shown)


tooling/bundler/src/bundle/windows/nsis.rs
```diff
@@ -71,7 +71,6 @@ pub fn bundle_project(settings: &Settings, updater: bool) -> crate::Result<Vec<P
   let nsis_toolset_path = tauri_tools_path.join("NSIS");
 
   if !nsis_toolset_path.exists() {
-    create_dir_all(&nsis_toolset_path)?;
```
amrbashir, line 74: at least we should create the parent of `nsis_toolset_path` so instead of removing, we should move this call up to line 70 and call it for `tauri_tools_path`

tooling/bundler/src/bundle/windows/nsis.rs
```diff
@@ -119,10 +118,12 @@ fn get_and_extract_nsis(nsis_toolset_path: &Path, _tauri_tools_path: &Path) -> c
     NSIS_TAURI_UTILS_SHA1,
     HashAlgorithm::Sha1,
   )?;
-
-  let target_folder = nsis_plugins.join("x86-unicode");
-  create_dir_all(&target_folder)?;
-  write(target_folder.join("nsis_tauri_utils.dll"), data)?;
```
amrbashir: `x86-unicode` could be empty when building on Linux or macOS so we need to add this back

amrbashir (changes requested):

amrbashir:
Thanks for contribution

tooling/bundler/src/bundle/windows/nsis.rs
Pylogmon: Why does this code run on Linux and MacOS?

tooling/bundler/src/bundle/windows/nsis.rs
Pylogmon: Okay, I get it.

tooling/bundler/src/bundle/windows/nsis.rs
amrbashir: because we support creating NSIS installers on macOS and Linux

amrbashir (approved):

amrbashir:
Thank you

PR #9963, opened 2024-06-04
feat(cli): iOS signing for CI usage

lucasfernog:
This PR sets up our CLI to modify the iOS project and export options to support manual code signing for CI usage.

A new crate was added, tauri-macos-sign, which was donated by our partnership with CrabNebula, which did the research for this iOS CI packaging solution. This crate is now used internally in the tauri-bundler too.

Here's what needs to be added to our docs:

```
### Certificate

After enrolling, navigate to the [Certificates](https://developer.apple.com/account/resources/certificates/list) page to create a new Apple Distribution certificate. Download the new certificate and install it to the macOS Keychain.

To export the certificate key, open the "Keychain Access" app, expand the certificate's entry, right-click on the key item and select "Export <key-name>" item. Select the path of the exported .p12 file and remember its password.
Run `base64 -i <path-to-certificate.p12> | pbcopy` to convert the certificate to base64 and copy it to the clipboard and paste that value to the `IOS_CERTIFICATE` environment variable. The certificate password must be set to the `IOS_CERTIFICATE_PASSWORD` variable.

### Provisioning Profile

Additionally, you must provide the provisioning profile for your application. In the [Identifiers](https://developer.apple.com/account/resources/identifiers/list) page, create a new App ID and make sure its "Bundle ID" value matches the identifier set in the [`identifier`] configuration.

Navigate to the [Profiles](https://developer.apple.com/account/resources/profiles/list) page to create a new provisioning profile. For App Store distribution, it must be an "App Store Connect" profile. Select the appropriate App ID and link the certificate you previously created.

After creating the provisioning profile, download it and run `base64 -i <path-to-profile.mobileprovision> | pbcopy` to convert the profile to base64 and copy it to the clipboard. Set that value to the `IOS_MOBILE_PROVISION` environment variable.
```


.changes/ios-signing.md
```diff
@@ -0,0 +1,6 @@
+---
+"tauri-cli": patch:feat
+"@tauri-apps/cli": patch:feat
+---
+
+Setup iOS signing by reading `IOS_CERTIFICATE`, `IOS_CERTIFICATE_PASSWORD` and `IOS_MOBILE_PROVISION` environment variables.
```
amrbashir, line 6: are these usually different than `APPLE_CERTIFICAT` and  `APPLE_CERTIFICATE_PASSWORD` or can we reuse those env vars?

tooling/cli/src/mobile/ios/build.rs
```diff
@@ -63,6 +63,42 @@ pub struct Options {
   /// Skip prompting for values
   #[clap(long, env = "CI")]
   pub ci: bool,
+  /// Export method.
+  #[clap(long, value_enum)]
+  pub method: Option<ExportMethod>,
```
amrbashir: Should probably be explicit and call it `--export-method`

.changes/ios-signing.md
lucasfernog: for macOS you'll usually go with a Developer ID Application certificate which doesn't apply to iOS. For iOS you'll usually go with an Apple Distribution or iOS Distribution certificate

tooling/cli/src/mobile/ios/build.rs
lucasfernog: agreed

.changes/ios-signing.md
amrbashir: cool, let's go with new env vars then

lyager:
This is a good I think. I've been in close dialogue with @lucasfernog on this one (thanks!), and it works for people who does not use a Company signature, but just a private one. For company distribution a resigning is needed - but the above is good, and I think it should me merged.

ecmel:
Any time frame for merging this?

amrbashir (approved):

ecmel:
Thank you for the great PR, I am getting `Provisioning profile is not an "iOS App Development" profile` error when building for iOS. Tried with `App Store Connect` and `Ad Hoc` distribution profiles. Only `iOS App Development` profile works.

```
** EXPORT FAILED **

    Error command ["xcodebuild"] exited with code 70
```

I was able to fix it by manually editing `src-tauri/gen/apple/ExportOptions.plist` and modifying `<key>method</key>` to 	`<string>release-testing</string>` Possible values are `development, app-store-connect, release-testing, enterprise, debugging, validation`

PR #9964, opened 2024-06-04
feat(cli): setup Android signing

lucasfernog:
Tweaks the default Android project to include a simple setup for Android signing.

Example Github Action usage:

```yml
- name: setup Android signing
        working-directory: src-tauri/gen/android
        run: |
          echo "keyAlias=${{ secrets.ANDROID_KEY_ALIAS }}" > keystore.properties
          echo "password=${{ secrets.ANDROID_KEY_PASSWORD }}" >> keystore.properties
          base64 -d <<< "${{ secrets.ANDROID_KEY_BASE64 }}" > $RUNNER_TEMP/keystore.jks
          echo "storeFile=$RUNNER_TEMP/keystore.jks" >> keystore.properties
```

### DIFF
--- a/tooling/cli/templates/mobile/android/app/build.gradle.kts
+++ b/tooling/cli/templates/mobile/android/app/build.gradle.kts
@@ -1,4 +1,5 @@
 import java.util.Properties
+import java.io.FileInputStream
 
 plugins {
     id("com.android.application")
@@ -6,6 +7,15 @@ plugins {
     id("rust")
     {{~#each android-app-plugins}}
     id("{{this}}"){{/each}}
+}
+
+val keystoreProperties = Properties()
+try {
+  // release builds require the file to be set up
+  keystoreProperties.load(FileInputStream(rootProject.file("keystore.properties")))
+} catch (e: Exception) {
+  // load a dummy file for debug purposes (signing config won't be used)
+  keystoreProperties.load(FileInputStream(rootProject.file("keystore.properties.dummy")))
 }
 
 val tauriProperties = Properties().apply {
@@ -26,6 +36,14 @@ android {
         versionCode = tauriProperties.getProperty("tauri.android.versionCode", "1").toInt()
         versionName = tauriProperties.getProperty("tauri.android.versionName", "1.0")
     }
+    signingConfigs {
+        create("release") {
+            keyAlias = keystoreProperties["keyAlias"] as String
+            keyPassword = keystoreProperties["password"] as String
+            storeFile = file(keystoreProperties["storeFile"] as String)
+            storePassword = keystoreProperties["password"] as String
+        }
+    }
     buildTypes {
         getByName("debug") {
             manifestPlaceholders["usesCleartextTraffic"] = "true"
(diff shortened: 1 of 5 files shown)


amrbashir (dismissed):

tooling/cli/templates/mobile/android/app/build.gradle.kts
```diff
@@ -8,6 +9,15 @@ plugins {
...
 }
 
+val keystoreProperties = Properties()
+try {
+  // release builds require the file to be set up
+  keystoreProperties.load(FileInputStream(rootProject.file("keystore.properties")))
+} catch (e: Exception) {
+  // load a dummy file for debug purposes (signing config won't be used)
+  keystoreProperties.load(FileInputStream(rootProject.file("keystore.properties.dummy")))
+}
```
amrbashir, line 19: maybe we should just include `keystore.properties` by default instead of the dummy one, or even just remove this altogether and just document how users might want to add signing later on if they wish.

lucasfernog: i'd be fine with this on docs instead @FabianLars @simonhyll

amrbashir: It used to be in the docs but was lost in recent refactors of tauri documentation and whether we merge this PR or not, I think we will still need to have it documented

lucasfernog:
@FabianLars what do you think about this standard approach vs a documentation based one? I'd like to get this resolved asap.. 
Personally I like giving an easier path to this for users for DX, but I also agree this is usually handled by docs instead.

amrbashir:
I was going to PR the guide today, maybe I will do it tomorrow

FabianLars:
I really like the idea (though we should also document the manual approach) but maybe it makes more sense to match the other platforms where we rely on env vars (or maybe both)?

amrbashir:
I opened a PR for signing guide https://github.com/tauri-apps/tauri-docs/pull/2337 anyways as it needs to be documented nonetheless

lucasfernog:
> I really like the idea (though we should also document the manual approach) but maybe it makes more sense to match the other platforms where we rely on env vars (or maybe both)?

in this case i think it's dangerous since we would end up writing secrets in plain text..

lucasfernog:
> I opened a PR for signing guide [tauri-apps/tauri-docs#2337](https://github.com/tauri-apps/tauri-docs/pull/2337) anyways as it needs to be documented nonetheless

if we merge this PR we also need to update that guide btw

lucasfernog:
another reason to merge this is that it helps people that do not commit the gen folder

amrbashir:
> another reason to merge this is that it helps people that do not commit the gen folder

didn't we change that behavior? at least it should be fine to not ignore `gen` folder by default, there is only a couple of files that should be ignored and we can add them in the `.gitignore`

FabianLars:
Yes, we changed it, but that doesn't mean that everyone will want to commit the folder if they don't intend on changing files in it (ignoring the context of this pr)

lucasfernog:
by default we gitignore the gen folder IIRC

amrbashir:
> by default we gitignore the gen folder IIRC

we don't, we only ignore `gen/schemas`

amrbashir:
the problem with this PR approach is it just as broken as if it wasn't configured, I would say even more broken, since without you can build your unsigned apk successfully but with this, it will always fail until signing is configured.

PR #9974, opened 2024-06-04
fix(rpm): use product name for output package

amrbashir:
ref: https://github.com/tauri-apps/tauri/pull/9375
ref: https://github.com/tauri-apps/tauri/pull/9375#issuecomment-2144997488

### DIFF
--- a/.changes/bundler-rpm-filename.md
+++ b/.changes/bundler-rpm-filename.md
@@ -1,0 +1,6 @@
+---
+"tauri-bundler": "patch:bug"
+---
+
+Use the `productName` for `rpm` package name instead of main binary name, to be consistent with other bundle types.
+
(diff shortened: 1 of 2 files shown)


amrbashir:
@FabianLars even though the product name is used for the output bundle, the main binary name is still used in the RPM manifest, I assume the main binary name affects how you call the binary from command line? can you test if this aligns with `.deb`?

FabianLars:
.deb:
- .deb file uses productName as is
- binary to invoke in console is from cargo.toml
- apt knows the installed app as productName as kebab-case (!)

.rpm:
- .rpm file uses productName as is
- binary to invoke in console is from cargo.toml
- dnf knows the installed app as name from cargo.toml (!)

amrbashir:
@FabianLars can you retest with the last commit?

FabianLars:
as far as i can tell it behaves the same now

FabianLars (approved):

PR #7964, opened 2023-10-05
feat(bundler): add dmg settings, closes #4669

Andreybest:
Added new "dmg" param to tauri.conf.json. It provides more settings to "create-dmg" cli and allows to customise DMG files with:
- background image
- Installer window size
- Installer window position
- Application position
- Application folder position

DMG view before:
image

DMG view after:
image

New part of tauri.conf.json looks like this:
```json
{
  "tauri": {
    "bundle": {
      "dmg": {
        "background": "images/background.png",
        "appPosition": {
          "x": 180,
          "y": 170
        },
        "applicationFolderPosition": {
          "x": 480,
          "y": 170
        },
        "windowSize": {
          "height": 400,
          "width": 660
        },
        "windowPosition": {
          "x": 200,
          "y": 180
        }
      }
    }
  }
}
```

Resolves #4669 

### What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [x] Build-related changes
- [ ] Other, please describe:

### Does this PR introduce a breaking change?

- [ ] Yes, and the changes were approved in issue #___
- [x] No

### Checklist
- [x] When resolving issues, they are referenced in the PR's title (e.g `fix: remove a typo, closes #___, #___`)
- [ ] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md).
- [x] I have added a convincing reason for adding this feature, if necessary

### Other information
Small note. I've never used commit signing, but saw that it's required after creating PR (after first commit). So I have only 1 out of 2 commit signed, I hope it's not a big deal 🙂

PR #8334, opened 2023-12-04
Cherry pick DMG settings to 1.x

hjmallon:
### What kind of change does this PR introduce?

Cherry pick from dev to 1.x

- [ ] Bugfix
- [x] Feature
- [ ] Docs
- [ ] New Binding issue #___
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

### Does this PR introduce a breaking change?

- [ ] Yes, and the changes were approved in issue #___
- [x] No

### Checklist
- [x] When resolving issues, they are referenced in the PR's title (e.g `fix: remove a typo, closes #___, #___`)
- [ ] A change file is added if any packages will require a version bump due to this PR per [the instructions in the readme](https://github.com/tauri-apps/tauri/blob/dev/.changes/readme.md).
- [x] I have added a convincing reason for adding this feature, if necessary

PR #9985, opened 2024-06-05
feat: add dmg settings

Pylogmon:
Added new "dmg" param to tauri.conf.json. It provides more settings to "create-dmg" cli and allows to customise DMG files.

Backport of #7964


hjmallon:
I just updated my MR which does the same thing here https://github.com/tauri-apps/tauri/pull/8334. When one of these is merged the other can be closed.

FabianLars:
Let's go with #8334 then (i prefer to go with the older one, sorry :/ ). Thank you for contributing nonetheless!

PR #9994, opened 2024-06-06
feat!(nsis): add an option to customize start menu folder

### LINKED ISSUE #9880: [feat] an option to not create a start menu folder for nsis installer
### Describe the problem

We currently use nsis's built-in start menu shortcut folder picker and put the shortcut in that folder, and most of the times, user will just spam next and resulting `app/app.exe` in the start menu, it's pretty weird to put a single shortcut inside of a folder for me

And also, I don't think most end users understand what shortcut menu is and that page is always a source of confusion (I didn't know what that is after using Windows for 10 years before I started really learning those things myself at least)

Add an option to skip that picker page and just put the shortcut in the start menu without the folder, to be honest, I think it should even be the default

### Alternatives considered

Doing it manually with custom template

### Additional context

A simple comparison (not produced by tauri's nsis installer, just as a visual reference)

image
image

Legend-Master:
Closes #9880


Legend-Master:
Current behavior should be

If old installer was using `StartMenu/App/App.lnk` and new installer uses:

- With same folder name:
  - Everything stays the same
- No additional folder or with a different folder name:
  - Nothing happens on `/UPDATE`
  - Create a new shortcut on manual install
  - Remove old and new shortcuts on uninstall

If old installer was using `StartMenu/App.lnk` and new installer uses:

- With a folder name:
  - Nothing happens on `/UPDATE`
  - Create a folder and new shortcut on manual install
  - Remove old and new shortcuts on uninstall
- No folder name:
  - Everything stays the same

core/tauri-utils/src/config.rs
```diff
@@ -776,6 +776,9 @@ pub struct NsisConfig {
   /// See <https://nsis.sourceforge.io/Reference/SetCompressor>
   #[serde(default)]
   pub compression: NsisCompression,
+  /// Create a start menu folder for grouping shortcuts together
```
```suggestion
  /// Set the folder name for the start menu shortcut.
  /// 
  /// Use this option if you have multiple apps and wish to group their shortcuts under one folder
  /// or if you generally prefer to set your shortcut inside a folder.
  /// 
  /// Examples:
  /// - `AwesomePublisher`, shortcut will be placed in `%AppData%\Microsoft\Windows\Start Menu\Programs\AwesomePublisher\<your-app>.lnk`
  /// - If unset, shortcut will be placed in `%AppData%\Microsoft\Windows\Start Menu\Programs\<your-app>.lnk`
```

tooling/bundler/src/bundle/windows/nsis.rs
```diff
@@ -309,7 +316,7 @@ fn build_nsis_app_installer(
   let main_binary_path = settings.binary_path(main_binary).with_extension("exe");
   data.insert(
     "main_binary_name",
-    to_json(main_binary.name().replace(".exe", "")),
+    to_json(main_binary.name().strip_suffix(".exe")),
```
amrbashir: Let's add a fallback, just in case it doesn't end with `.exe`
```suggestion
    to_json(main_binary.name().strip_suffix(".exe").unwrap_or(main_binary.name())),
```

tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -744,15 +751,19 @@ Function RestorePreviousInstallLocation
...
 Function SkipIfPassive
   ${IfThen} $PassiveMode = 1  ${|} Abort ${|}
 FunctionEnd
 
 Function CreateOrUpdateStartMenuShortcut
   ; We used to use product name as MAINBINARYNAME
   ; migrate old shortcuts to target the new MAINBINARYNAME
-  ${If} ${FileExists} "$DESKTOP\${PRODUCTNAME}.lnk"
-    !insertmacro SetShortcutTarget "$DESKTOP\${PRODUCTNAME}.lnk" "$INSTDIR\${MAINBINARYNAME}.exe"
+  ${If} ${FileExists} "$SMPROGRAMS\$AppStartMenuFolder\${PRODUCTNAME}.lnk"
```
amrbashir: Shouldn't we also check for `$SMPROGRAMS\${PRODUCTNAME}.lnk`, and update it?

.changes/nsis-start-menu-folder.md
```diff
@@ -0,0 +1,6 @@
+---
+"tauri-utils": "patch:feat"
+"tauri-bundler": "patch:breaking"
+---
+
+Add an option to customize start menu folder for NSIS installer, and changed default to no additional folder instead of an extra folder with the same name as the shortcut
```
amrbashir: let's separate the breaking change in another change file

amrbashir:
```suggestion

Add `bundle > nsis > startMenuFolder` option to customize start menu folder for NSIS installer
```

amrbashir (changes requested):

tooling/bundler/src/bundle/windows/templates/installer.nsi
Legend-Master: Right, was thinking that we only had shortcuts with folders, forgot some people like myself had it without the folder 😂

.changes/nsis-start-menu-folder-breaking.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri-bundler": "patch:breaking"
+---
+
+Changed start menu folder default to no additional folder instead of an extra folder with the same name as the shortcut
```
amrbashir:
```suggestion
Changed NSIS start menu shortcut to be placed directly inside `%AppData%\Microsoft\Windows\Start Menu\Programs` without an additional folder. You can get the old behavior by setting `bundle > nsis > startMenuFolder` to the same value as your `productName`
```

tooling/bundler/src/bundle/settings.rs
```diff
@@ -417,6 +417,8 @@ pub struct NsisSettings {
   pub display_language_selector: bool,
   /// Set compression algorithm used to compress files in the installer.
   pub compression: NsisCompression,
+  /// Create a start menu folder for grouping shortcuts together
```
```suggestion
  /// Set the folder name for the start menu shortcut.
  ///
  /// Use this option if you have multiple apps and wish to group their shortcuts under one folder
  /// or if you generally prefer to set your shortcut inside a folder.
  ///
  /// Examples:
  /// - `AwesomePublisher`, shortcut will be placed in `%AppData%\Microsoft\Windows\Start Menu\Programs\AwesomePublisher\<your-app>.lnk`
  /// - If unset, shortcut will be placed in `%AppData%\Microsoft\Windows\Start Menu\Programs\<your-app>.lnk`
```

tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -702,6 +708,7 @@ Section Uninstall
     !insertmacro UnpinShortcut "$SMPROGRAMS\$AppStartMenuFolder\${PRODUCTNAME}.lnk"
     Delete "$SMPROGRAMS\$AppStartMenuFolder\${PRODUCTNAME}.lnk"
     RMDir "$SMPROGRAMS\$AppStartMenuFolder"
+    Delete "$SMPROGRAMS\${PRODUCTNAME}.lnk"
```
amrbashir: We need to call `UnpinShortcut` before this as well

tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -702,6 +708,7 @@ Section Uninstall
     !insertmacro UnpinShortcut "$SMPROGRAMS\$AppStartMenuFolder\${PRODUCTNAME}.lnk"
     Delete "$SMPROGRAMS\$AppStartMenuFolder\${PRODUCTNAME}.lnk"
     RMDir "$SMPROGRAMS\$AppStartMenuFolder"
```
amrbashir: I don't think we should call these anymore, unless the installer have created it, maybe we should check if `$AppStartMenuFolder` is not empty or something or if it equals `${STARTMENUFOLDER}`.

tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -744,15 +751,22 @@ Function RestorePreviousInstallLocation
...
 Function CreateOrUpdateStartMenuShortcut
   ; We used to use product name as MAINBINARYNAME
   ; migrate old shortcuts to target the new MAINBINARYNAME
-  ${If} ${FileExists} "$DESKTOP\${PRODUCTNAME}.lnk"
-    !insertmacro SetShortcutTarget "$DESKTOP\${PRODUCTNAME}.lnk" "$INSTDIR\${MAINBINARYNAME}.exe"
+  ${If} ${FileExists} "$SMPROGRAMS\$AppStartMenuFolder\${PRODUCTNAME}.lnk"
+    !insertmacro SetShortcutTarget "$SMPROGRAMS\$AppStartMenuFolder\${PRODUCTNAME}.lnk" "$INSTDIR\${MAINBINARYNAME}.exe"
+    Return
+  ${ElseIf} ${FileExists} "$SMPROGRAMS\${PRODUCTNAME}.lnk"
+    !insertmacro SetShortcutTarget "$SMPROGRAMS\${PRODUCTNAME}.lnk" "$INSTDIR\${MAINBINARYNAME}.exe"
```
amrbashir: shouldn't we upgrade both if they exist? and not just update one or the other?

amrbashir (changes requested):

tooling/bundler/src/bundle/windows/templates/installer.nsi
Legend-Master: There're no good way of knowing that to be honest, `$AppStartMenuFolder` returns default start menu folder if no registry key is found and also when that registry key is found and is the default value

Legend-Master: There're no good way of knowing that to be honest, `$AppStartMenuFolder` returns default start menu folder if no registry key is found and also when that registry key is found and is the default value

tooling/bundler/src/bundle/windows/templates/installer.nsi
Legend-Master: Migration should only need to happen once, and unless they had both of them before, there're no reason to check and migrate both

amrbashir: after this PR, developers may change the start menu folder back and forth if they wish, so we should just update both to be safe.

tooling/bundler/src/bundle/windows/templates/installer.nsi
amrbashir: yeah, I was afraid of that, maybe we should check if it is at least equal to `${STARTMENUFOLDER}`, that at least should be an indication to remove this.

tooling/bundler/src/bundle/windows/templates/installer.nsi
Legend-Master: It shouldn't happen at all, the first time user installs it, it migrates the existing one, after the first time, we won't need the migration code to work anymore

tooling/bundler/src/bundle/windows/templates/installer.nsi
Legend-Master: I don't see any downside of just try deleting this except the potential extra attempt to be honest, unless someone else's app is in your app's folder and the shortcut is the same name

amrbashir: yeah I guess, but avoiding deleting files that wasn't created by us still better than accidentally deleting a shortcut and a folder that may or may not be created by us

tooling/bundler/src/bundle/windows/templates/installer.nsi
amrbashir: the migration code will always run on update, because if you changed the main binary name, we still need to update the shortcut

Legend-Master: it will always run on update, but it won't matter if it works or not after the first run

tooling/bundler/src/bundle/windows/templates/installer.nsi
Legend-Master: I don't see a good way of detecting if it's created by us or not to be honest

amrbashir: just checking if `$AppStartMenuFolder` is equal to `${STARTMENUFOLDER}` should be enough IMO. if there is residual files after that, we can't be blamed as that is our best effort to cleanup

tooling/bundler/src/bundle/windows/templates/installer.nsi
amrbashir: not sure I understand what you mean

amrbashir: if we are updating and found a shortcut, why update only that one, we can also update the other one as well

tooling/bundler/src/bundle/windows/templates/installer.nsi
Legend-Master: If it was `Tauri API/Tauri API.lnk`, and we changed `${STARTMENUFOLDER}` to `""`, `$AppStartMenuFolder` will not be equal to `${STARTMENUFOLDER}` and it doesn't get clean up

tooling/bundler/src/bundle/windows/templates/installer.nsi
Legend-Master: I don't think it can happen, but if you're concerned, I'll add it then

tooling/bundler/src/bundle/windows/templates/installer.nsi
amrbashir: When running the uninstaller, it will have the same values as the installer that created the uninstaller not the one trying to update, so the value of AppStartMenuFolder will be the same and same will be for STARTMENUFOLDER

Legend-Master: But it's no longer true after update if we changed `${STARTMENUFOLDER}` to `""`

Another way is to migrate old shortcuts on update, but we need to preserve the old start menu shortcuts because if we migrate them, pinned old shortcuts will be gone

Legend-Master: Well, I just realized that we didn't specify these

image

Currently it can't even clean up start menu shortcuts if the user chose another folder

Legend-Master: We didn't store anything before, there're no way of knowing if we had these or not

After this change, we'll have most of tauri apps migrating to this and if we choose not to clean it up, then almost everyone will be unhappy about installing, if we delete it, there will be some really rare occasions where some other app with exactly the same name and folder name will get affected (they'll have some other problems like desktop shortcut anyway)

amrbashir: > But it's no longer true after update if we changed `${STARTMENUFOLDER}` to `""`

correct, but at that point, I don't think we should care about it

> Another way is to migrate old shortcuts on update, but we need to preserve the old start menu shortcuts because if we migrate them, pinned old shortcuts will be gone

That's not needed, it probably will be more headache

amrbashir: > Well, I just realized that we didn't specify these
> 
> image
> 
> Currently it can't even clean up start menu shortcuts if the user chose another folder

It doesn't use defaults? bummer, we should specify those then

amrbashir: > We didn't store anything before, there're no way of knowing if we had these or not
> 
> After this change, we'll have most of tauri apps migrating to this and if we choose not to clean it up, then almost everyone will be unhappy about installing, if we delete it, there will be some really rare occasions where some other app with exactly the same name and folder name will get affected (they'll have some other problems like desktop shortcut anyway)

yes but on one hand, we are deleting, which is destructive and on the other we are being careful not to delete arbitrary files that we are not at least 50% percent sure we should delete. I'd pick the later.

In ideal scenarios, the shortcut location will be the same between updates and nothing changes for the user.

In worst case, an extra shortcut will be leftover, not the end of the world and if the user stumbles upon a broken shortcut, Windows will prompt the user to delete it.

Legend-Master: I would argue that our app simply occupies both `ProductName.lnk` and `ProductName/ProductName.lnk`, if there is a conflict, then be it

> that we are not at least 50% percent sure we should delete

After this, I would expect it to be 99%+ on existing apps

> In worst case, an extra shortcut will be leftover, not the end of the world and if the user stumbles upon a broken shortcut, Windows will prompt the user to delete it.

Yes, but I would have a very bad impression about that app

> It doesn't use defaults? bummer, we should specify those then

This thing is quite tricky to be honest, it depends on multiple things in different sections, I will leave to another day to fix, not this PR

amrbashir: > I would argue that our app simply occupies both `ProductName.lnk` and `ProductName/ProductName.lnk`, if there is a conflict, then be it
> 
> > that we are not at least 50% percent sure we should delete
> 
> After this, I would expect it to be 99%+ on existing apps

yeah but on future apps it won't be, it will bet at least 50%

> > In worst case, an extra shortcut will be leftover, not the end of the world and if the user stumbles upon a broken shortcut, Windows will prompt the user to delete it.
> 
> Yes, but I would have a very bad impression about that app

It would even be a worse impression if your app deletes another shortcut though

> > It doesn't use defaults? bummer, we should specify those then
> 
> This thing is quite tricky to be honest, it depends on multiple things in different sections, I will leave to another day to fix, not this PR

No problem, thanks for helping

Legend-Master: Well, if you persist, I wouldn't fight on this, but is there anyway to mitigate this for old apps? I can just add a check here and only remove plain shortcuts if no `${STARTMENUFOLDER}` and only remove the one in the folder other wise

> It would even be a worse impression if your app deletes another shortcut though

To be honest, if you really think about it, we try to re-target old shortcuts to our app, it's just as destructive as this (and happens a lot more frequently on very update), I don't see how we can deal with this type of conflicts anyway

amrbashir: > is there anyway to mitigate this for old apps? I can just add a check here and only remove plain shortcuts if no `${STARTMENUFOLDER}` and only remove the one in the folder other wise

I honestly hate this; we seem to be cycling to this same issue every time we make a breaking change like this, "can we migrate old app shortcut?" yes, sure but it comes at costs tbh. Ideally, we should be smart about deleting or updating existing shortcuts, I wonder if we can associate the app identifier in the shortcut properties, so that we can query them later to decide whether to update or delete a shortcut.

> > It would even be a worse impression if your app deletes another shortcut though
> 
> To be honest, if you really think about it, we try to re-target old shortcuts to our app, it's just as destructive as this (and happens a lot more frequently on very update), I don't see how we can deal with this type of conflicts anyway

that is indeed destructive as well, might as well delete them all then

Legend-Master: The problem about making changes on updaters is that we can't really make breaking changes, they're directly impacting the user, not like other breaking changes are just affecting the devs

> I wonder if we can associate the app identifier in the shortcut properties, so that we can query them later to decide whether to update or delete a shortcut

We probably can detect if the shortcut's target is on our binary or not, but I don't know how practical is it to be done in nsis

> that is indeed destructive as well, might as well delete them all then

Do you mean we just delete the shortcuts on uninstall or delete the migration code?

Legend-Master: I just pushed a commit to detect if the shortcut is actually pointing to our binary, this should prevent all those things

amrbashir (approved):

PR #9999, opened 2024-06-06
feat(schema): make acl permissions schema unique

Legend-Master:
Reference from Discord chat: https://discord.com/channels/616186924390023171/986184094050316358/1233987058213716019

> I think our schema file is missing an uniqueItems on permissions (maybe also windows and webviews)
> 
> image


amrbashir:
should we just switch to use a `HashSet` instead of `Vec`? that should produce a schema with `uniqueItems: true`, and makes defining a capability in Rust have same semantics as the JSON capability cc @lucasfernog

lucasfernog (dismissed):

lucasfernog:
not worth the trouble of using HashSet, since our snapshot tests rely on a deterministic order :|

amrbashir (approved):

PR #10007, opened 2024-06-07
fix(core): check if listener.handler is undefined

kanatapple:
Bug Details
- Event handler registered in `listen` in frontend is not executed
(Event firing from backend -> frontend is succeeding)

Cause
- When an event fires, https://github.com/tauri-apps/tauri/blob/5b769948a81cac333f64c870a470ba6525bd5cd3/core/tauri/src/event/mod.rs#L230 this function is executed, but `listener.handler` may be `undefined`, causing an exception and preventing the event handler from being executed.
- __internal_unstable_listeners_object_id__

Solution
- The root cause was that `listener.handler` was registered as `undefined`, but I could not find the cause, so I added a check to see if `listener.handler` is `undefined`.

Reference from Discord chat: https://discord.com/channels/616186924390023171/1215591289953718304/1217419467840028682

### DIFF
--- a/core/tauri/src/event/mod.rs
+++ b/core/tauri/src/event/mod.rs
@@ -231,7 +231,7 @@ pub fn event_initialization_script(function: &str, listeners: &str) -> String {
         const listeners = (window['{listeners}'] && window['{listeners}'][eventData.event]) || []
         for (const id of ids) {{
           const listener = listeners[id]
-          if (listener) {{
+          if (listener && listener.handler) {{
             eventData.id = id
             listener.handler(eventData)
           }}


amrbashir (approved):
Thanks for contribution, this check is ofc a good solution and doesn't have any downsides but if you can, could you please provide a minimal repro of the bug so I can investigate how did the handler became undefined?

kanatapple:
@amrbashir 
Thanks for the review 🙇 

It can be reproduced in this repository.
https://github.com/kanatapple/tauri-remote-access
Just load `https://www.google.co.jp` to reproduce.

kanatapple:
@amrbashir 
I know it's a little late, but was a change file not necessary?

kanatapple:
Looks like it wasn't necessary 👍

PR #10015, opened 2024-06-08
fix: missing depends for rpm package

Pylogmon:
rpm package missing `libayatana-appindicator3.so.1` dependency


amrbashir (changes requested):
LGTM, just need to add a change file in `.changes` directory

amrbashir (approved):

PR #10030, opened 2024-06-10
feat(cli): include default.toml and capabilities in plugin template

amrbashir:

### DIFF
--- a/tooling/cli/src/plugin/init.rs
+++ b/tooling/cli/src/plugin/init.rs
@@ -239,8 +239,16 @@ pub fn command(mut options: Options) -> Result<()> {
     .with_context(|| "failed to render plugin Android template")?;
   }
 
-  std::fs::create_dir(template_target_path.join("permissions"))
+  let permissions_dir = template_target_path.join("permissions");
+  std::fs::create_dir(&permissions_dir)
     .with_context(|| "failed to create `permissions` directory")?;
+
+  let default_permissions = r#"[default]
+description = "Default permissions for the plugin"
+permissions = ["ping"]
+"#;
+  std::fs::write(permissions_dir.join("default.toml"), default_permissions)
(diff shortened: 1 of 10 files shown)


tooling/cli/templates/plugin/build.rs
```diff
@@ -1,4 +1,4 @@
-const COMMANDS: &[&str] = &["ping", "execute"];
```
tillmann-crabnebula, line 1: Any specific reason to remove the execute command in this PR?
This breaks the example app UI which invokes the execute command.

tillmann-crabnebula (changes requested):

tooling/cli/templates/plugin/build.rs
amrbashir: Nice catch, I meant to remove it and replace it with `ping`, seems like I missed that part, I will update `execute` usages to be `ping` instead

amrbashir: should be fixed now

tweidinger (approved):

tooling/cli/src/plugin/init.rs
```diff
@@ -239,9 +239,17 @@ pub fn command(mut options: Options) -> Result<()> {
     .with_context(|| "failed to render plugin Android template")?;
   }
 
-  std::fs::create_dir(template_target_path.join("permissions"))
+  let permissions_dir = template_target_path.join("permissions");
+  std::fs::create_dir(&permissions_dir)
     .with_context(|| "failed to create `permissions` directory")?;
 
+  let default_permissions = r#"[default]
```
tillmann-crabnebula, line 246: One nitpick I noticed "too late" is that this should include a reference to the schema file for autocompletion and validation.

```
"$schema" = "schemas/schema.json"
```

Issue #10022, opened 2024-06-08
[bug] `cargo tauri icon` cannot find font-family for SVG <text> on Linux (Ubuntu)


cornmonger:
### Describe the bug

`cargo tauri icon` is having problems applying the **font-family** attribute in SVG files. This appears to be the case regardless of whether the font-family is indicated inline in a `<text>` element or as a `<style>` definition. It warns that it can't find the font and then does not render text when generating icons. Everything else appears to render. 

This run, for example, is just with no font-family specified at all:
```bash
$ cargo tauri icon test.svg
    Warn No match for 'sans-serif' font-family.
    Appx Creating StoreLogo.png
    Appx Creating Square30x30Logo.png
    Appx Creating Square44x44Logo.png
   [...]
```

So far, this is reproducible on a relatively fresh install of Ubuntu Desktop 24 as well as an older install of Ubuntu Desktop 23.

### Reproduction

1. Follow the beta guide to setup Tauri via Cargo on an Ubuntu Desktop 24 system.
2. Create a new Tauri beta app using cargo.
3. Run `cargo tauri icon test.svg` against any SVG with a `<text>` element.
4. ...
5. No profit.

Here is a test SVG that I can reproduce with (test.svg):
```svg
<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
<text x="0" y="50" fill="white">hello</text>
</svg>
```
System info:
```bash
$ sudo apt install [...]
libwebkit2gtk-4.1-dev is already the newest version (2.44.2-0ubuntu0.24.04.1).
build-essential is already the newest version (12.10ubuntu1).
curl is already the newest version (8.5.0-2ubuntu10.1).
wget is already the newest version (1.21.4-1ubuntu4).
file is already the newest version (1:5.45-3build1).
libxdo-dev is already the newest version (1:3.20160805.1-5build1).
libssl-dev is already the newest version (3.0.13-0ubuntu3.1).
libayatana-appindicator3-dev is already the newest version (0.5.93-1build3).
librsvg2-dev is already the newest version (2.58.0+dfsg-1build1).
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

### Full `tauri info` output

```text
WARNING: no lock files found, defaulting to npm

[✔] Environment
    - OS: Ubuntu 24.4.0 X64
    ✔ webkit2gtk-4.1: 2.44.2
    ✔ rsvg2: 2.58.0
[...]

PR #10039, opened 2024-06-11
fix(cli): icon should render explicit svg text node fonts (fix: #10022)

roylaurie:
This fixes the first problem mentioned in #10022: SVG text nodes are generally not rendering in `cargo tauri icon`.

After applying this patch, it should be observed that running `cargo tauri icon` against an SVG with a `<text>` node no longer warns about not finding fonts that exist on a (Linux) system. Further, visually reviewing the resulting image files created should now show text rendered.

However, the correct font is still not selected for generic font-families on Linux systems that are using `fontconfig`. This includes "_sans-serif_", "_sans_", "_comic_", etc. In reviewing the [fontdb](/RazrFalcon/fontdb) library, which is used for font selection, I found that the font that the library selects for generic families is not the one that `fc-match` selects. There is no warning associated here, the user would have to visually confirm that this is the case.

The fontdb team is aware of the issue and has an open issue for it:
[fontdb: Support automatic determination of default family names on Linux from fontconfig #24](/RazrFalcon/fontdb/issues/24)

For the time being, I would recommend explicitly defining the exact font to use in an SVG, which is typically the case anyway. Of course, distributing the font with the SVG would ensure better cross-platform compatibility regardless.

### DIFF
--- a/.changes/fix-cli-icon-svg-render-text.md
+++ b/.changes/fix-cli-icon-svg-render-text.md
@@ -1,0 +1,7 @@
+---
+"tauri-cli": "patch:bug"
+"@tauri-apps/cli": "patch:bug"
+---
+
+Fixed an issue that prevented `tauri icon` from rendering `<text>` nodes in SVG files.
+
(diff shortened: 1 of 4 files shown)


amrbashir (changes requested):
Thank you, this PR is almost ready to go, could you please add a change file in `.changes` directory?

roylaurie:
Sure. Added a `.changes` file.

amrbashir (approved):
Thank you

PR #10086, opened 2024-06-18
fix(nsis): esitmated size unit

Legend-Master:
Esitmated size unit should be in KB (1024 bytes), not kB (1000 bytes)


FabianLars:
Where did you find that information? All i can find right now seems like it's about 1000 - can't find anything _explicit_ rn so idk (same for when i implemented this).
Also isn't it KB (1000) and KiB (1024)? Or is that a german thing x)

Legend-Master:
From NSIS's website that they said in KB

> https://nsis.sourceforge.io/Add_uninstall_information_to_Add/Remove_Programs#:~:text=EstimatedSize%20(DWORD)%20%2D%20The%20size%20of%20the%20installed%20files%20(in%20KB)

Windows uses KB to represent 1024 bytes (you can see this in file explorer)

image

The easiest way to know this is in the installer's choose install directory page, you can see an estimated size, and that size currently doesn't match the size we write to the registry, which will be displayed in Windows's uninstall programs page (also you can install it and see the actual size excluding the uninstaller with the number in uninstall page)

KiB is 1024, kB is 1000, KB can be both XD

> https://physics.nist.gov/cuu/Units/binary.html
> https://www.nist.gov/pml/owm/metric-si-prefixes

To be honest, Windows uses MB for 1024 x 1024 bytes but in SI unit it should be 1000 x 1000 (2^6) bytes, so...

FabianLars:
thanks for the answer :)

FabianLars (approved):

PR #10093, opened 2024-06-19
fix(windows): resource_dir() return \\\?\\ prefix symbols

olexiyb:
`handle.path().resource_dir()` gives unexpected folder started with "\\\\?\\" due to current behavior of std::env::current_exe() described in https://github.com/rust-lang/rust/issues/99931

This does not allow the use of the folder in scripts or other programs called in Tauri application.

### DIFF
--- a/core/tauri-utils/src/platform.rs
+++ b/core/tauri-utils/src/platform.rs
@@ -351,6 +351,13 @@ mod tests {
   use crate::{Env, PackageInfo};
 
   #[test]
+  fn test_current_exe() {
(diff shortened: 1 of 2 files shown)


amrbashir:
This is the default behavior of Rust, and since this API is used in Rust, I think it should stay as is and if users want to remove the UNC prefix, they should use `dunce`, we only remove UNC prefix in tauri if it is being sent to JS.

olexiyb:
Can we have a second function to get resource dir with UNC prefix?

PR #10106, opened 2024-06-23
fix(nsis): delete app data button gone on higher scaling

Legend-Master:
Fix a regression after we enabled dpi awareness for NSIS installer in #8940

On 150% scaling

Before:

image

After:

image

Also added `dpiAwareness` `PerMonitorV2` to manifest for Windows 10 1607+ (note this should not affect lower versions since they should be able to ignore this and pick up `dpiAware` `true` set by `ManifestDPIAware true`)

> https://github.com/kichik/nsis/blob/5fc0b87b819a9eec006df4967d08e522ddd651c9/Docs/src/attributes.but#L286-L300


tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -1,5 +1,6 @@
 Unicode true
 ManifestDPIAware true
+ManifestDPIAwareness PerMonitorV2
```
amrbashir, line 7: could you link to the documentation of this option? I can't seem to find a reliable source for it so let's add a link here so it can be easily found later.

tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -364,19 +365,37 @@ Var DeleteAppDataCheckboxState
...
+    StrCpy $3 "${__NSD_CheckBox_EXSTYLE}"
+    IntOp $4 0 * $2
+  ${EndIf}
+  IntOp $5 100 * $2
+  IntOp $6 400 * $2
+  IntOp $7 25 * $2
+  IntOp $4 $4 / 96
+  IntOp $5 $5 / 96
+  IntOp $6 $6 / 96
+  IntOp $7 $7 / 96
```
amrbashir, line 394: I wonder if we can get the position of `Uninstall from` text and add our control at 100px downards

The width could be reliably calculated from the window returned by `FindWindow $1`

tooling/bundler/src/bundle/windows/templates/installer.nsi
Legend-Master: This option is undocumented on their website, but is in the Docs folder of the source, I would probably prefer adding in a link to this PR instead, what do you think?

> https://github.com/kichik/nsis/blob/5fc0b87b819a9eec006df4967d08e522ddd651c9/Docs/src/attributes.but#L286-L300

tooling/bundler/src/bundle/windows/templates/installer.nsi
Legend-Master: To be honest, I don't want to try to do it in NSIS 😂, this implementation may not be the most robust but realistically speaking, I don't think this layout will change anytime soon

tooling/bundler/src/bundle/windows/templates/installer.nsi
amrbashir: could add both links, no problem

amrbashir (approved):

PR #10112, opened 2024-06-24
feat: add generic to tauri::ipc::Channel

Brendonovich:
Adds a `T` generic to `ipc::Channel`. This allows for type-safe `on_message` callbacks, and makes it possible to extract the channel message type in tauri-specta.


Brendonovich:
Still needs work for mobile

PR #10117, opened 2024-06-24
fix(bundler/wix): Add `AUTOLAUNCHAPP` and `LAUNCHAPPARGS` public properties

FabianLars:
Currently the v2 updater plugin cannot restart the app after an update because we don't use powershell anymore. With this the updater will be able to tell the msi to start the app after the installation (plugin PR will be opened after https://github.com/tauri-apps/plugins-workspace/pull/1495 was merged cause i'm lazy)

### DIFF
--- a/tooling/bundler/src/bundle/windows/templates/main.wxs
+++ b/tooling/bundler/src/bundle/windows/templates/main.wxs
@@ -30,7 +30,9 @@
         <Property Id="REINSTALLMODE" Value="amus" />
 
         <!-- Check if this is an update from tauri-plugin-update to autostart the app after install -->
-        <Property Id="AUTOLAUNCHAPP" Value="no" Secure="yes" />
+        <Property Id="AUTOLAUNCHAPP" Secure="yes" />
+        <!-- Property to forward cli args to the launched app to not lose those of the pre-update instance -->
+        <Property Id="LAUNCHAPPARGS" Secure="yes" />
 
         {{#if allow_downgrades}}
             <MajorUpgrade Schedule="afterInstallInitialize" AllowDowngrades="yes" />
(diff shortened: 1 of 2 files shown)


tooling/bundler/src/bundle/windows/templates/main.wxs
```diff
@@ -334,6 +337,10 @@
         </InstallExecuteSequence>
         {{/if}}
 
+        <InstallExecuteSequence>
+          <Custom Action="LaunchApplication" After="InstallFinalize"><![CDATA[NOT(AUTOLAUNCHAPP="True")]]> AND NOT Installed</Custom>
```
amrbashir: Can this be changed to also accept a list of arguments?

FabianLars: Do you mean to be able to re-apply the CLI flags the initial app instance was started with?

It should be possible, I'll take a look if that's what you meant.

amrbashir: Yeah that's what I meant

FabianLars: whoever invented spaces can go straight to hell

FabianLars: I can't believe how long it took me to figure out the code for the updater plugin ffs (in handsight it was easy but i never would have guessed to escape `"` with `""` in arguments already encapsulated with `"`...)

https://github.com/tauri-apps/plugins-workspace/pull/1498

tooling/bundler/src/bundle/windows/templates/main.wxs
```diff
@@ -29,6 +29,11 @@
         <!-- reinstall all files; rewrite all registry entries; reinstall all shortcuts -->
         <Property Id="REINSTALLMODE" Value="amus" />
 
+        <!-- Check if this is an update from tauri-plugin-update to autostart the app after install -->
```
amrbashir:
```suggestion
        <!-- Auto launch app after installation, useful for passive mode which usually used in updates -->
```

amrbashir (approved):

Issue #6255, opened 2023-02-13
[feat] cargo profile support


flapili:
### Describe the problem

when we have multiples build variant like dev/stagging/production/beta/alpha/etc... the debug flag is not enough

### Describe the solution you'd like

add an option to the cli like `tauri build --profile foobar`

### Additional context

for now I have to use a custom py script which take a kind of profile, build tauri via subprocess.run then go to `/targets/<triples-target>/release(or debug)/bundle/msi/name-version-arch.zip` to a folder according to the given profile.

amrbashir:
try `tauri build -- --profile foobar`

amrbashir:
Looks like it will conflict with the `--release` that we pass internally, gonna open a PR to fix it

bukowa:
Any chance to get it in `v1`?

amrbashir:
this is already fixed in v1 as well

bukowa:
> this is already fixed in v1 as well

Right it just needs the `--debug` flag to work. Also it has other problems;
``` toml
[profile.testing]
inherits = "dev"
```
```sh
# after removing target dir
$ tauri -- build --debug -- --profile=testing

Finished `testing` profile [unoptimized + debuginfo] target(s) in 49.67s                         
   Error failed to rename `C:\Users\buk\WebstormProjects\ck3oop\target\debug\ck3oop-ui-rs.exe` to
`C:\Users\buk\WebstormProjects\ck3oop\target\debug\ck3oop-ui.exe`: The system cannot find the path specified. (os error 3)
```
Looks like tauri wants to rename something post building but with different profile it just can't?

```toml
[package]
name = "ck3oop-ui-rs"
```

```json
  "package": {
    "productName": "ck3oop-ui",
  },
  ```

amrbashir:
> > this is already fixed in v1 as well
> 
> Right it just needs the `--debug` flag to work. Also it has other problems;

It doesn't need `--debug`, it should work without it

> ```toml
> [profile.testing]
> inherits = "dev"
> ```
> 
> ```shell
> # after removing target dir
> $ tauri -- build --debug -- --profile=testing
> 
> Finished `testing` profile [unoptimized + debuginfo] target(s) in 49.67s                         
>    Error failed to rename `C:\Users\buk\WebstormProjects\ck3oop\target\debug\ck3oop-ui-rs.exe` to
[...]

PR #10136, opened 2024-06-27
fix(cli): parse `--profile=<profile>` syntax

amrbashir:
ref: https://github.com/tauri-apps/tauri/issues/6255#issuecomment-2192466839

### DIFF
--- a/tooling/cli/src/interface/rust.rs
+++ b/tooling/cli/src/interface/rust.rs
@@ -46,7 +46,7 @@ mod manifest;
 use cargo_config::Config as CargoConfig;
 use manifest::{rewrite_manifest, Manifest};
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Default)]
 pub struct Options {
   pub runner: Option<String>,
   pub debug: bool,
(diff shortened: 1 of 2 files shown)


tooling/cli/src/interface/rust.rs
```diff
@@ -1022,9 +1022,14 @@ pub fn get_profile(options: &Options) -> &str {
   options
     .args
     .iter()
-    .position(|a| a == "--profile")
-    .map(|i| options.args[i + 1].as_str())
-    .unwrap_or_else(|| if options.debug { "debug" } else { "release" })
+    .position(|a| a.starts_with("--profile"))
+    .map(|i| {
+      options.args[i]
+        .split_once("=")
```
amrbashir:
```suggestion
        .split_once('=')
```

bukowa:
Can we check for all flags not only profile?

amrbashir:
Our CLI arguments are parsed using `clap` crate which covers this syntax for us and so we don't need do it manually. However, we allow passing arbitrary args to the build runner (usually `cargo`) and these args could contain `--profile` which conflicts with `--release` flag that we pass internally so we need to avoid this conflict. In short, this is the only place we manually check for the flag and its value.

lucasfernog (approved):

PR #10137, opened 2024-06-27
enhance(nsis): include uninstaller size in estimated size

amrbashir:


FabianLars (approved):
(did not test it)

lucasfernog (approved):
uninstaller size was kinda irrelevant on my test but it seems like it works

PR #10139, opened 2024-06-27
feat: add TSend generic to Channel

Brendonovich:
Adds a `TSend` generic to `ipc::Channel`. This allows for type-safe calls to `send`, and makes it possible for `tauri-specta` to extract the generic type of channels.
The generic isn't used in the `on_message` function as that introduces quite a bit more complexity to do so, especially when supporting mobile. This can be implemented later if needed.


core/tauri/src/menu/plugin.rs
```diff
@@ -877,7 +877,7 @@ pub(crate) fn init<R: Runtime>() -> TauriPlugin<R> {
     .on_event(|app, e| {
       if let RunEvent::MenuEvent(e) = e {
         if let Some(channel) = app.state::<MenuChannels>().0.lock().unwrap().get(&e.id) {
-          let _ = channel.send(&e.id);
+          let _ = channel.send(e.id.clone());
```
lucasfernog, line 880: shouldn't we allow sending a reference?

Brendonovich: I did consider that but `send` immediately serializes the value, so the string would have been cloned anyway. At least this way you can serialize owned values unlike if only references were supported, but maybe there's a way to support both i haven't considered

.changes/ipc-channel-generic.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri": major:enhance
```
lucasfernog:
```suggestion
"tauri": major:breaking
```

lucasfernog (approved):

PR #10167, opened 2024-07-02
fix(core): fix raw invoke body for isolation pattern

amrbashir:
The `isolation` pattern requests are made using JSON but the payload could be raw bytes, so we send the original `Content-Type`  from frontend and make sure to deserialize the payload using that one instead of `Content-Type` from request headers


amrbashir:
@lucasfernog do you have any idea why is the macOS test failing?

FabianLars:
the embed-plist crate can only be used once per binary (by design). So maybe we have to add an option to the generate_context macro to disable it for these test 🤔 (Or imo better yet, if we disable it in test env directly but idk if that works across crates)

lucasfernog (approved):
you forgot to verify this on an actual app - the JS side was encoding the payload to string

amrbashir:
I did test it on an actual app, otherwise I wouldn't have found the problem and its fix.

I did notice that it was encoding to a string and didn't understand why so I left it as is because the failing code I was testing, worked fine with it.

lucasfernog:
gooood, well done

PR #10170, opened 2024-07-02
refactor!(core & api): rename drag events for better consistency and clarity

amrbashir:


tooling/api/src/webview.ts
```diff
@@ -542,51 +533,53 @@ class Webview {
...
             position: mapPhysicalPosition(event.payload.position)
           }
         })
       }
     )
 
-    const unlistenDrop = await this.listen<DragDropPayload>(
-      TauriEvent.DROP,
+    const unlistenDragOver = await this.listen<DragPayload>(
+      TauriEvent.DRAG_LEAVE,
```
jaydevelopsstuff: This listener should be attached to `TauriEvent.DRAG_OVER`, no?

jaydevelopsstuff (changes requested):

tooling/api/src/webview.ts
amrbashir: nice catch, thanks

.changes/drag-drop-events-renamed-api.md
```diff
@@ -0,0 +1,17 @@
+---
+"@tauri-apps/api": "patch:breaking"
+---
+
+Renamed drag and drop events in `TauriEvent` enum to better convey when they are triggered:
+
+- `TauriEvent.DRAG` -> `TauriEvent.DRAG_ENTER`
+- `TauriEvent.DROP_OVER` -> `TauriEvent.DRAG_OVER`
+- `TauriEvent.DROP` -> `TauriEvent.DRAG_DROP`
+- `TauriEvent.DROP_CANCELLED` -> `TauriEvent::DRAG_LEAVE`
```
lucasfernog, line 10: 'drag leave' makes sense on the JS world because you must define which elements receive the event; in our case 'leave' is misleading because it's not triggered when the drag operation moves out of the window

lucasfernog (changes requested):

.changes/drag-drop-events-renamed-api.md
amrbashir: It should be triggered, if not, that's a bug

lucasfernog: oohhh i didn't know we changed that on wry, let me check

lucasfernog: works indeed, but now i miss the old https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dragend_event from https://docs.rs/wry/0.29.0/wry/webview/enum.FileDropEvent.html#variant.Cancelled

lucasfernog (approved):

.changes/drag-drop-events-renamed-api.md
amrbashir: we can't differentiate between cancelled and leave so we combined them, not sure what's the best name here

PR #10171, opened 2024-07-02
feat(core): mark `AppHandle::restart` and `process::restart` as diverging functions

### LINKED ISSUE #10160: [feat] `AppHandle::restart` should be diverging function (`-> !`)
### Describe the problem

When I want to restart the tauri app, we should call `AppHandle<R>::restart` but it's not diverging function so I have to add `unreachable!()` to ensure the logic after restart will not executed

Make `AppHandle::restart` an diverging function (`-> !`).

As far as I found, it's already diverging (in tauri v1).

This change can be breaking so considering in v2 might be good.

Never type is not stable but diverging function is stable since rust 1.0.

https://doc.rust-lang.org/rust-by-example/fn/diverging.html

### Alternatives considered

just insert

### Additional context

_No response_

amrbashir:
closes #10160

### DIFF
--- a/.changes/core-restart-fn-diverging.md
+++ b/.changes/core-restart-fn-diverging.md
@@ -1,0 +1,5 @@
+---
+"tauri": "patch:enhance"
+---
+
+Mark `AppHandle::restart` and `process::restart` as [diverging functions](https://doc.rust-lang.org/rust-by-example/fn/diverging.html).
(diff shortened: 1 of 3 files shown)


Legend-Master:
Should we also mark `exit`?

amrbashir:
good call

amrbashir:
actually it is not possible for `exit` since it only sends a request to `exit`, and users are free to prevent that, so we need to either leave it as is or introduce a new method called `request_exist` and mark the current one as diverging

Legend-Master:
That makes sense, I only saw that if error process::exit part and assumed it will always exit, seems like it's detecting the error of sending the message

FabianLars (approved):

lucasfernog (approved):

PR #10201, opened 2024-07-04
ci: Fix v1 docs update trigger

FabianLars:
In the docs repo I'd prefer to keep v1 workflows in the v1 branch so the old repository_dispatch doesn't work anymore. With the gh cli we can trigger the normal workflow_dispatch though.

I have a feeling this may not work on the first try so we need to monitor this on the next v1 release :)


amrbashir (dismissed):

amrbashir:
shouldn't this PR target `1.x` branch?

FabianLars:
Probably maybe perhaps - github's UI always directed me to the dev branch so i was utterly confused. But you're right, the msrv prs also were against the 1.x branch.

FabianLars:
sadge

FabianLars:
sadge

FabianLars:
grafik
chill github...

PR #10210, opened 2024-07-06
feat(core): add a new function to set theme dynamically

### LINKED ISSUE #5279: [feat] Change Window theme dynamically at runtime
### Describe the problem

As far as I can tell from the docs, concerning the theme of the window, there are JS APIs for only (1) [getting the window's current theme](https://tauri.app/v1/api/js/window/#theme), and (2) [listening to the system theme change](https://tauri.app/v1/api/js/window#onthemechanged).

I'm trying to set the window's theme to "dark"/"light" as well when the user chooses the "dark"/"light" theme in my app.

**Is there a way to dynamically change the theme of the window through JavaScript?**

A JS API to change the theme of the window.

### Alternatives considered

_No response_

### Additional context

_No response_

Legend-Master:
Closes #5279

Waiting for https://github.com/tauri-apps/tao/pull/937


packages/api/src/window.ts
```diff
@@ -1676,6 +1676,16 @@ class Window {
     })
   }
 
+  /**
+   * Set window theme, pass in undefined or no parameters to follow system theme
```
amrbashir:
```suggestion
   * Set window theme, pass null to follow system theme
   *
   * @since 2.0.0
```

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -2973,6 +2981,13 @@ fn handle_user_message<T: UserEvent>(
               }
             };
           }
+          WindowMessage::SetTheme(theme) => {
+            window.set_theme(match theme {
```
amrbashir, line 3028: should also update the webview theme on Windows,

amrbashir (changes requested):
We should also add `app.set_theme()`

crates/tauri/src/window/mod.rs
```diff
@@ -1989,6 +1990,11 @@ tauri::Builder::default()
       .set_title_bar_style(style)
       .map_err(Into::into)
   }
+
+  /// Set the window theme.
+  pub fn set_theme(&self, theme: Option<Theme>) -> crate::Result<()> {
+    self.window.dispatcher.set_theme(theme).map_err(Into::into)
```
amrbashir: On Windows, We should also update the theme for menu if there is any set

crates/tauri-runtime-wry/src/lib.rs
Legend-Master: isn't it already handled in the theme event?

Legend-Master:
I need to get some sleep, will get back to this tomorrow

crates/tauri-runtime-wry/src/lib.rs
amrbashir: yeah, you're right

crates/tauri/src/window/mod.rs
Legend-Master: Do you know how to make the entire menu dark? Using `set_theme_for_hwnd` only makes the root menu button dark

amrbashir: it should automatically be using dark theme if the window dark theme is configured correctly, maybe we did something wrong in the recent tao commit? could you try reverting that and see if it works?

Legend-Master: Tried tao 0.30, system on light mode, app on dark mode (using `window_builder.theme(Some(Theme::Dark))`), is this correct?

image

amrbashir: Running the `tao` example in `muda` repository, I can see the menu is dark. This is using `tao@0.30.1` btw. I am on Windows 11 though. I haven't tested the dark menus on Windows 10 ever.

Legend-Master: Maybe we need to do some work there then, if I switch system theme to dark, the sub menus are now dark, and some other apps like sublime text's sub menus do turn dark on light system theme with dark app theme

Do you mind trying out the example in this PR? I can't seem to change the menu from dark back to light

Legend-Master: Oh well, it's a muda bug, we sent the data through `lparam` but we're reading the data from `wparam`

Legend-Master: Made a PR https://github.com/tauri-apps/muda/pull/224, also I think we also need to trigger a redraw here in muda, what do you think?

crates/tauri/src/app/plugin.rs
```diff
@@ -50,6 +52,11 @@ pub fn default_window_icon<R: Runtime>(
   })
 }
 
+#[command(root = "crate")]
+pub async fn set_app_theme<R: Runtime>(app: AppHandle<R>, theme: Option<Theme>) {
```
Legend-Master, line 56: This will cause a deadlock on plugins without `async`, lock happened on these two places, I'm not sure how should we handle this, seems like it's because this call triggers a `ThemeChanged` event while having a lock on the plugins when invoked

https://github.com/tauri-apps/tauri/blob/f6a9717aadb8455f193fb35f20735f586d7dc2fb/crates/tauri/src/manager/mod.rs#L462

https://github.com/tauri-apps/tauri/blob/f6a9717aadb8455f193fb35f20735f586d7dc2fb/crates/tauri/src/app.rs#L2157

packages/api/src/window.ts
```diff
@@ -1676,6 +1676,21 @@ class Window {
     })
   }
 
+  /**
+   * Set window theme, pass in `null` or `undefined` to follow system theme
+   *
+   * ### Platform-specific
```
amrbashir:
```suggestion
   * #### Platform-specific
```

packages/api/src/window.ts
```diff
@@ -1676,6 +1676,21 @@ class Window {
...
+   *
+   * - **Linux / macOS**: Theme is app-wide and not specific to this window.
+   * - **iOS / Android:** Unsupported.
```
amrbashir, line 1685:
```suggestion
   * - **iOS / Android:** Unsupported.
   *
   * @since 2.0.0
```

packages/api/src/app.ts
```diff
@@ -101,4 +102,25 @@ async function defaultWindowIcon(): Promise<Image | null> {
...
 
-export { getName, getVersion, getTauriVersion, show, hide, defaultWindowIcon }
+/**
+ * Set app's theme, pass in `null` or `undefined` to follow system theme
+ *
+ * @example
+ * ```typescript
+ * import { setTheme } from '@tauri-apps/api/app';
+ * await setTheme('dark');
+ *
```
amrbashir, line 115:
```suggestion
 *
 * @since 2.0.0
```

packages/api/src/app.ts
```diff
@@ -101,4 +102,25 @@ async function defaultWindowIcon(): Promise<Image | null> {
   )
 }
 
-export { getName, getVersion, getTauriVersion, show, hide, defaultWindowIcon }
+/**
+ * Set app's theme, pass in `null` or `undefined` to follow system theme
+ *
```
amrbashir, line 107: should mention platform specific notes here

crates/tauri/src/window/mod.rs
```diff
@@ -1989,6 +1990,36 @@ tauri::Builder::default()
...
+  pub fn set_theme(&self, theme: Option<Theme>) -> crate::Result<()> {
+    self
+      .window
+      .dispatcher
+      .set_theme(theme)
+      .map_err(Into::<crate::Error>::into)?;
+    #[cfg(windows)]
+    if let (Some(window_menu), Ok(hwnd)) = (&*self.menu_lock(), self.hwnd()) {
+      let raw_hwnd = hwnd.0 as isize;
+      let menu = window_menu.menu.clone();
```
amrbashir: instead of holding the `menu_lock` here for the rest of the scope, we can just clone and release the lock
```suggestion
    let window_menu = self.menu_lock().clone();
    if let (Some(menu), Ok(hwnd)) = (window_menu, self.hwnd()) {
      let raw_hwnd = hwnd.0 as isize;
```

crates/tauri/src/app.rs
```diff
@@ -685,6 +685,32 @@ macro_rules! shared_app_impl {
         })
       }
 
+      /// Set the app theme.
+      pub fn set_theme(&self, theme: Option<Theme>) {
+        #[cfg(windows)]
+        for window in self.manager.windows().values() {
+          if let (Some(window_menu), Ok(hwnd)) = (&*window.menu_lock(), window.hwnd()) {
+            let raw_hwnd = hwnd.0 as isize;
+            let menu = window_menu.menu.clone();
```
amrbashir: instead of holding the `menu_lock` here for the rest of the scope, we can just clone and release the lock
```suggestion
          let window_menu = window.menu_lock();
          if let (Some(menu), Ok(hwnd)) = (window_menu, window.hwnd()) {
            let raw_hwnd = hwnd.0 as isize;
```

amrbashir (changes requested):

amrbashir (approved):

wyhaya:
Will the set_theme API be ported to `Tauri 1x`? It is still very difficult to dynamically change the theme for Tauri 1x, especially on Windows, as we cannot access the Webview instance.

FabianLars:
We typically don't backport features this large _ourselves_ but we'll look at community prs doing so. The changes, especially in Tao are not that trivial so i doubt this would be worth it.

p.s. what's the reason you stay with v1?

Issue #10205, opened 2024-07-05
[bug]  A full-size window is created (without enabling fullscreen), but the window has 8px on each side where mouse events cannot be captured.


eed-cn:
### Describe the bug

A full-size window is created (without enabling fullscreen), but the window has 8px on each side where mouse events cannot be captured.

I want to create a fullscreen window, similar to a screenshot application, but the edges of the screen do not respond to mouse events. Is this a bug, or is there an issue with my code?

my code:
```js
const screenWindow = new WebviewWindow(`${screenLabel}-${m.id}`, {
        alwaysOnTop: true,
        decorations: false,
        resizable: false,
        shadow: false,
        skipTaskbar: true,
        transparent: true,        
        url,
        visible: false,
      })
      const unlistenFn1 = await screenWindow.once('tauri://created', async () => {
        unlistenFn1()
        screenWindow.setPosition(new PhysicalPosition(m.position[0], m.position[1]))
        screenWindow.setSize(new PhysicalSize(m.size[0], m.size[1]))
      })
// show
// m.position[0] =0 
// m.position[1] =0
// m.size[0] =screen width
// m.size[1] = screen hight
```

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.22635 X64
    ✔ WebView2: 126.0.2592.87
    ✔ MSVC: Visual Studio Community 2022
    ✔ rustc: 1.79.0 (129f3b996 2024-06-10)
    ✔ cargo: 1.79.0 (ffa9cf99a 2024-06-03)
    ✔ rustup: 1.27.1 (54dd3d00f 2024-04-24)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (default)
    - node: 21.7.3
    - pnpm: 9.4.0
    - yarn: 1.22.22
    - npm: 10.5.0
    - bun: 1.1.10

[-] Packages
    - tauri [RUST]: 2.0.0-beta.23
    - tauri-build [RUST]: 2.0.0-beta.18
    - wry [RUST]: 0.41.0
    - tao [RUST]: 0.28.1
    - @tauri-apps/api [NPM]: 2.0.0-beta.12
    - @tauri-apps/cli [NPM]: 2.0.0-beta.12

[-] App
    - build-type: bundle
    - CSP: default-src * filesystem: data: blob: stream: tauri: ipc: asset:
    - frontendDist: ../dist
    - devUrl: http://localhost:1420/
    - framework: Vue.js
[...]

PR #10211, opened 2024-07-06
fix(runtime-wry): window edge not working after setting resizable false and decorated false dynamically

Legend-Master:
Not sure if it's related with #10205

### DIFF
--- a/.changes/resize-false-undecorated-edge.md
+++ b/.changes/resize-false-undecorated-edge.md
@@ -1,0 +1,5 @@
+---
+"tauri-runtime-wry": "patch:bug"
+---
+
+Fix window edge not working after setting resizable false and decorated false dynamically
--- a/core/tauri-runtime-wry/src/lib.rs
+++ b/core/tauri-runtime-wry/src/lib.rs
@@ -2782,7 +2782,7 @@ fn handle_user_message<T: UserEvent>(
             } else if !window.is_decorated() {
               undecorated_resizing::attach_resize_handler(window.hwnd());
             }
-          },
+          }
           WindowMessage::SetMaximizable(maximizable) => window.set_maximizable(maximizable),
           WindowMessage::SetMinimizable(minimizable) => window.set_minimizable(minimizable),
           WindowMessage::SetClosable(closable) => window.set_closable(closable),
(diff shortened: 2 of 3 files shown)


Legend-Master:
I'm not sure if we also want to disable it on fullscreen and maximize or is it something the user should do manually

FabianLars:
> I'm not sure if we also want to disable it on fullscreen and maximize or is it something the user should do manually

I think we used to do that 🤔

Legend-Master:
Looking at the code it seems to be handled by amr's code already, ~~didn't test it though~~ seems like it's working

.changes/resize-false-undecorated-edge.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri-runtime-wry": "patch:bug"
+---
+
+Fix window edge not working after setting resizable false and decorated false dynamically
```
amrbashir, line 5: I am not sure I understand the problem, if decoration is false and resizable is false, window edge should not work

Legend-Master: Shouldn't the edges of an undercoated window with resize disabled receive events and being clickable?

amrbashir: yeah it should, is it not receiving these events?

Legend-Master: no, that part is blocked by the top level resize window and not interactable at all

amrbashir: ah I see where the bug is, I will push a better fix

amrbashir: @Legend-Master could you give 14af33bc765751b07de187884995c8066902a697 a test?

Legend-Master: Doesn't seem to work, after setting both resize and decoration to false, the entire window becomes not interactable

amrbashir: I retested and seemed like I can click just fine, outside and inside the resize region.

amrbashir: nvm, I was mislead by the hover effect, I reverted my changes

amrbashir (approved):

.changes/resize-false-undecorated-edge.md
amrbashir: kinda weird that `HTTRANSPARENT` didn't work, will need to research it more I guess. Your solution is fine for now.

PR #10212, opened 2024-07-06
feat(cli/migrate): migrate zipped updater

Legend-Master:
Reference: https://github.com/tauri-apps/plugins-workspace/pull/1521


FabianLars:
Do we still need/want this with 1521 closed? I think it wouldn't hurt to add.

Legend-Master:
I think we can add this anyway, we might change mind on that one in the future 🙃

amrbashir:
let's close it for now, and re-open it in coordination with https://github.com/tauri-apps/plugins-workspace/pull/1521

PR #10228, opened 2024-07-09
chore(example): refactor api example

Legend-Master:
- Fixed some leaking listeners on hot reload 
- Fixed some broken stylings
- Fixed some broken UI functionalities
- Unify checkbox to left side of the label
- Add accent color to checkbox
- Disable content protection by default
- Slightly changed some other styles

image
image

### DIFF
--- a/examples/api/unocss.config.js
+++ b/examples/api/unocss.config.js
@@ -70,7 +70,7 @@ export default defineConfig({
     }
   ],
   shortcuts: {
-    btn: `select-none outline-none cursor-pointer shadow-md p-2 rd-1 text-primaryText border-none font-400 dark:font-600
+    btn: `select-none outline-none shadow-md p-2 rd-1 text-primaryText border-none font-400 dark:font-600
             bg-accent hover:bg-accentDarker active:bg-accentDarkest text-accentText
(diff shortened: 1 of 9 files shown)


examples/api/unocss.config.js
```diff
@@ -70,7 +70,7 @@ export default defineConfig({
     }
   ],
   shortcuts: {
-    btn: `select-none outline-none shadow-md p-2 rd-1 text-primaryText border-none font-400 dark:font-600
+    btn: `select-none outline-none cursor-pointer shadow-md p-2 rd-1 text-primaryText border-none font-400 dark:font-600
```
amrbashir: btn by default shouldn't have `cursor: pointer`, native apps don't usually set it
```suggestion
    btn: `select-none outline-none shadow-md p-2 rd-1 text-primaryText border-none font-400 dark:font-600
```

Legend-Master: Just want to say that looking at most web based apps and also websites I use, like Chrome, Edge, Steam, Discord, vscode, Google, GitHub, ... are using cursor pointer on buttons by default

I prefer it a bit more (to be honest, I would use it on anything clickable), but yeah, most native apps don't do this

amrbashir: yeah it is kinda a preference thing, I usually prefer to stick to native behavior when possible.

amrbashir (approved):

PR #10246, opened 2024-07-10
enhance: use `EventName` on window's `once` and remove listener before calling handler

Legend-Master:
- Use `EventName` on `Window`, `Webview` and `WebviewWindow`'s `once`
- Unlisten before calling handler on once so we always unlisten the event even if the callback throws, also allow that unlisten's error to not get caught
- Also refactored some other things


Legend-Master (commented):
Note that I don't think `once` on `Window` works with `tauri://created` and `tauri://error`, the handler is not removed after it gets called, I'm not familiar with how window specific event works, so I didn't fix that part

amrbashir (approved):

PR #10250, opened 2024-07-10
add "webview:allow-internal-toggle-devtools" to default.json

Fake-User:
developers should be able to open dev tools in the default project


Legend-Master:
Isn't it included in `webview:default`?

FabianLars:
yes it is

FabianLars:
@Fake-User can you tell us a bit more about the issue you were facing? Can you check whether you had `webview:default` enabled?

Fake-User:
@FabianLars testing with a fresh Tauri v2 project [ bun create tauri-app --beta ] on macOS Sonoma 14.5.

The "webview:default" permission is in the default.json, but the shortcuts for [opening the webview console](https://v2.tauri.app/develop/debug/#webview-console) do not work without adding the "webview:allow-internal-toggle-devtools" permission.

Legend-Master:
~~ah, I see when I press ctrl + shift + i on windows, the devtools does show up but with `VM12:263  Uncaught (in promise) Command plugin:webview|internal_toggle_devtools not allowed by ACL` logged to the console, probably a bug~~ never mind, just realized the api example doesn't have `webview:default` in the allow list, adding it fixed it

I can't reproduce this on Windows at least

Fake-User:
I just tested on a coworkers m2 mbp, so there is definitely an issue on macOS.
additionally [ cmd opt i ] wont close the web view console,
even with this permission enabled : /

lucasfernog:
Just tested it on macOS and it works after adding `webview:default` (without this change).

lucasfernog:
btw that API only works on debug builds (or release built with --features tauri/devtools)

FabianLars:
Works on my macbook without this pr too 🤔

PR #10270, opened 2024-07-13
fix(bundler): warning for self contained updaters

Legend-Master:
Fix regression from #10188, bundler warns about no updater-enabled targets were built for self contained updaters like app image, nsis, msi

### DIFF
--- a/tooling/bundler/src/bundle.rs
+++ b/tooling/bundler/src/bundle.rs
@@ -161,7 +161,15 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
         package_type: PackageType::Updater,
         bundle_paths: updater_paths,
       });
-    } else {
+    } else if updater.v1_compatible
+      || !package_types.iter().any(|package_type| {
+        // Self contained updater, no need to zip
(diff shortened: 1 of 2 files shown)


tooling/bundler/src/bundle.rs
```diff
@@ -161,7 +161,15 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
         package_type: PackageType::Updater,
         bundle_paths: updater_paths,
       });
-    } else {
+    } else if updater.v1_compatible
+      || !package_types.iter().any(|package_type| {
```
FabianLars, line 165: Hmm, maybe we need this to be dynamic depending on the currently built platform, but we could also add it later idc - edit: or maybe it's too early for me rn and it doesn't matter at this point in the code

FabianLars (approved):

tooling/bundler/src/bundle.rs
Legend-Master: To be honest, the way this thing is currently structured is all about maintaining the old code path for v1 zipped updaters

Issue #4917, opened 2022-08-12
[feat] Provide print API


alimehasin:
### Describe the problem

Tauri lacks a print API and this make it very difficult for electron based apps that perform lots of work with printers to migrate to Tauri, there are many use cases 

1. Unable to print a page if it is hidden `visible: false` (`window.print()` triggered automatically after page load)
2. Unable to close the page after the printing dialog is closed (I haven't found any way to know if the dialog still open)
3. Unable to perform silent printing

### Describe the solution you'd like

Provide print API

### Alternatives considered

At least I hope to see an event the get fired when the printing dialog closed

### Additional context

I have seen `Print to Printer and related options` checked in [this issue](https://github.com/tauri-apps/tauri/issues/544)
But electron has rich API regarding to printing comparing to Tauri

alimehasin:
> Unable to close the page after the printing dialog is closed (I haven't found any way to know if the dialog still open)

It turns out that there is a JS event called `afterprint` that will solves this problem, you can read more about it [here](https://developer.mozilla.org/en-US/docs/Web/API/Window/afterprint_event)

meet-rizwan:
@alimehasin  how this event `afterprint` will fix the printing in tauri ?  i am very curious to know because i am facing this problem for very long , i am not able to launch my application because of printin . many people are trying to figure out how this process will help in providing print solution for tauri 

how this event  will tell how may printers are connected to the system?
how this event tell which printer is default printer
how to stop print dialog in chrome 
how to adjust the margin and custom print settings by this event  ?

alimehasin:
Hello @rizwan92, the `afterprint` event won't fix these issues, it will only fix one problem which is

> Unable to close the page after the printing dialog is closed (I haven't found any way to know if the dialog still open)

As quoted in my previous comment

meet-rizwan:
I do not know why this issue's priority is medium. It is a very essential part of building a desktop application. I really want to launch my  application using Tauri

Beanow:
@rizwan92 Tauri is an open source project depending on it's contributors.
If you need this feature urgently, maybe you can help get the ball rolling?

For example by finding out the cross platform support by the underlying webviews for printing, and seeing what the common ground is?

Referencing https://tauri.app/v1/references/webview-versions/ if you need more insight in the webviews.

ihekui:
I am looking forward to the silent printing API, because my app needs this API to print small tickets and price tags

meet-rizwan:
[...]

PR #10275, opened 2024-07-13
feat: first implementation of print API

mildred:
add few printing options, implemented on top of ~~https://github.com/tauri-apps/wry/pull/1317~~ https://github.com/tauri-apps/wry/pull/1324
this is a step towards https://github.com/tauri-apps/tauri/issues/4917


mildred:
On the wry side there is still an issue with PDF generation, but otherwise this PR seems to work well. Further implementation will come to wry itself (and the typescript wrapper)

PR #10277, opened 2024-07-14
fix: `Webview.reparent` pointing to wrong Rust API

### LINKED ISSUE #10274: It seems the reparent API is linking with set_webview_focus instead
I believe there is an error here:

https://github.com/tauri-apps/tauri/blob/6d965e9fbf57323b1ba25faa68b4d0b7df4f07db/tooling/api/src/webview.ts#L492-L507

Legend-Master:
Fix #10274


FabianLars (approved):

FabianLars:
btw, you can merge your own PRs yourself once approved (if you already got the permission set up), or at least if they're small if you're not that confident yet.

Just saying that because I'd like to train you differently from the start compared to how it was for me back then 😂

PR #10323, opened 2024-07-19
For linting and testing auto update the node version to "lts/*"

martinfrances107:
This deprecation notice.

```
version-or-publish
The following actions uses Node.js version which is deprecated and will be forced to run on node20: actions/setup-node@v3. For more info: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/
```

seen in this test run.

<https://github.com/martinfrances107/tauri/actions/runs/9995204078>

### DIFF
--- a/.github/workflows/lint-js.yml
+++ b/.github/workflows/lint-js.yml
@@ -23,7 +23,7 @@ jobs:
       - run: corepack enable
       - uses: actions/setup-node@v4
         with:
-          node-version: '18'
+          node-version: 'lts/*'
           cache: 'pnpm'
       - run: pnpm i --frozen-lockfile
       - run: pnpm format:check
@@ -35,7 +35,7 @@ jobs:
       - run: corepack enable
       - uses: actions/setup-node@v4
         with:
-          node-version: '18'
+          node-version: 'lts/*'
           cache: 'pnpm'
       - run: pnpm i --frozen-lockfile
       - run: pnpm eslint:check
@@ -47,7 +47,7 @@ jobs:
       - run: corepack enable
       - uses: actions/setup-node@v4
         with:
-          node-version: '18'
+          node-version: 'lts/*'
           cache: 'pnpm'
       - run: pnpm i --frozen-lockfile
       - run: pnpm ts:check


martinfrances107:
I should explain the second commit.

When I reviewed my own fix... I saw that it should be universally applied to  many actions.

.github/workflows/test-cli-js.yml
```diff
@@ -39,9 +39,9 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
 
       - name: setup node
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v4
         with:
-          node-version: 18
+          node-version: 20
```
jbolda: Please adjust this to node@18. We would like to run the tests on the oldest LTS/maintenance release. That being said, if there is an alias for it, it would be nice to not have to change it.

.github/workflows/test-android.yml
```diff
@@ -43,9 +43,9 @@ jobs:
           sudo apt-get install -y libgtk-3-dev webkit2gtk-4.1
 
       - name: setup node
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
-          node-version: 18
+          node-version: 20
```
jbolda: Please adjust this to node@18. We would like to run the tests on the oldest LTS/maintenance release. That being said, if there is an alias for it, it would be nice to not have to change it.

.github/workflows/lint-js.yml
```diff
@@ -20,9 +20,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-node@v3
+      - uses: actions/setup-node@v4
         with:
-          node-version: '18'
+          node-version: '20'
```
jbolda: We could probably set this to `lts/*` to avoid having to update it again in the future.

jbolda (changes requested):
Appreciate the effort!

martinfrances107:
Opps so I went stomping arround in clodhoopers over a pre-existing well thought out set of policies. ( thanks for being so polite with the bull in the china shop :) )

... My bad.

For reference I went looking for the node policy, I will quote it at the end. The LTS is changed annually

For our discussion the options  from actions/setup-node@v3 that I want to use seem to be "latest" "current" or "lts/*" as appropiate

For this change there is a negative aspect..

When debugging/comparing test between passes and failures... a developer might grimace to discover that the node version is not static from run to run.

But given that policy this will be a "once yearly event" 

I just wanted that converstion to be stated in public.

Anyway I will fix all my errors over the weekend.

---------------------------------------------- 

https://nodejs.org/en/about/previous-releases

Major Node.js versions enter Current release status for six months, which gives library authors time to add support for them. After six months, odd-numbered releases (9, 11, etc.) become unsupported, and even-numbered releases (10, 12, etc.) move to Active LTS status and are ready for general use. LTS release status is "long-term support", which typically guarantees that critical bugs will be fixed for a total of 30 months. Production applications should only use Active LTS or Maintenance LTS releases.

.github/workflows/docker.yml
```diff
@@ -67,9 +67,9 @@ jobs:
           targets: ${{ matrix.target.name }}
 
       - name: Setup node
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
-          node-version: 18
+          node-version: 'latest'
```
jbolda: @lucasfernog do you need a specific version here?

jbolda (changes requested):
@martinfrances107 I apologize if I caused confusion here!

My intent is our tests run with specifically defined node `18`. As you mentioned, we want to update this with an intent instead of on the LTS schedule. I believe we share the same opinion here!

Seperate from that, I believe we can `lts/*` with the lint workflows. I don't believe we will be affected by this version changing on their schedule, and it is convenient. I could be convinced to use `18` everywhere though.

martinfrances107:
Sorry for the delay here .. this had to wait until the weekend

Ok I think have more faithfully expressed the requirements .. here it is in bullets points.

* Anything publishing or deployment related
  -- A hard coded number. Ensures no surprise updates. Human will always be involved in change.

* lint/testing
  -- auto increment with a bias to the conservative "LTS"

* android... I think my best response is the backout the changes in that regard completely
and let someone else drive change.

Other more informed android devs will have opions about lastes/LTS and for example wheather those measures
should be ignored because they deviate from the long tail of the install base.

Another thing I see in a few places is a list of supported nodes version --

```
      matrix:
        node:
          - '16'
          - '18'
          - '20'
```

to my mind I think we can drop support for 16 now, but I think especially when discussing policy.
micro PRs are better.

PS 

A PR is incomming.

.github/workflows/docker.yml
martinfrances107: I am just going to back out changes here... devs with a more a more informed opinion should drive change here.

jbolda (dismissed):
Thanks! Appreciate it!

martinfrances107:
Many of the change have already filtered into the dev branch 

Mostly due to last saturdays commit 

Author: Lucas Fernandes Nogueira <lucas@tauri.app>
Date:   Sat Aug 17 08:21:27 2024 -0300
 23a912bb84d7c6088301e1ffc59adfa8a799beab

 So I have redone this PR (hence the force push) with just small items left.

In fact The only thing that remain is ...

 lint/testing
-- auto increment with a bias to the conservative "LTS"

lucasfernog (approved):

martinfrances107:
Thank you

PR #10339, opened 2024-07-21
feat(bundler/nsis): add minimum webview2 version checks

Legend-Master:
Add an option to the installer to ensure the webview2 version


tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -499,57 +500,75 @@ Section WebView2
...
-      ; $6 holds the path to the webview2 installer
-      ExecWait "$6 ${WEBVIEW2INSTALLERARGS} /install" $1
-      ${If} $1 = 0
-        DetailPrint "$(webview2InstallSuccess)"
-      ${Else}
-        DetailPrint "$(webview2InstallError)"
-        Abort "$(webview2AbortError)"
+      ${If} $R0 = 1
+        DetailPrint "$(installingWebview2)"
+        ExecWait `"C:\Program Files (x86)\Microsoft\EdgeUpdate\MicrosoftEdgeUpdate.exe" /install appguid={F3017226-FE2A-4295-8BDF-00C3A9A7E4C5}` $1
```
Legend-Master: I don't have an arm device to verify if they also use this path and also I'm not sure how webview2 user install works here (install path and doing it with system install updater without `needsadmin` or setting it to `prefers` didn't work)

FabianLars: The path is the same on arm (couldn't test the user install yet)

core/tauri-utils/src/config.rs
```diff
@@ -834,6 +834,11 @@ pub struct NsisConfig {
   /// ```
   #[serde(alias = "installer-hooks")]
   pub installer_hooks: Option<PathBuf>,
+  /// Ensure webview2 version to be newer than this major version,
+  /// if the user's webview2 is older than this version,
+  /// the installer will trigger a webview2 update
+  #[serde(alias = "ensure-webview2-version")]
+  pub ensure_webview2_version: Option<u16>,
```
amrbashir: shouldn't this be a string?

tooling/bundler/src/bundle/settings.rs
```diff
@@ -457,6 +457,10 @@ pub struct NsisSettings {
   /// !macroend
   /// ```
   pub installer_hooks: Option<PathBuf>,
+  /// Ensure webview2 version to be newer than this major version,
+  /// if the user's webview2 is older than this version,
+  /// the installer will trigger a webview2 update
+  pub ensure_webview2_version: Option<u16>,
```
amrbashir: ditto

tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -499,57 +500,77 @@ Section WebView2
...
-      ${If} $1 = 0
-        DetailPrint "$(webview2InstallSuccess)"
-      ${Else}
-        DetailPrint "$(webview2InstallError)"
-        Abort "$(webview2AbortError)"
+      ${If} $R0 = 1
+        DetailPrint "$(installingWebview2)"
+        ; Chromium updater docs: https://source.chromium.org/chromium/chromium/src/+/main:docs/updater/user_manual.md
+        ; Modified from "HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\Microsoft EdgeWebView\ModifyPath"
+        ExecWait `"C:\Program Files (x86)\Microsoft\EdgeUpdate\MicrosoftEdgeUpdate.exe" /install appguid={F3017226-FE2A-4295-8BDF-00C3A9A7E4C5}&needsadmin=true` $1
```
amrbashir: I don't like this hardcoded value, especially since `C:/` is not guaranteed to be the OS drive. We should check `path` key inside `Computer\HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\EdgeUpdate`

amrbashir (changes requested):

core/tauri-utils/src/config.rs
Legend-Master: I don't think this needs to be a full semantic version, just major version is good enough, what do you think?

tooling/bundler/src/bundle/windows/templates/installer.nsi
Legend-Master: Right, I can take a look at it this Friday probably, kinda busy working on a game mod right now, releasing in a few days

core/tauri-utils/src/config.rs
amrbashir: That's not very intuitive IMO

tooling/bundler/src/bundle/windows/templates/installer.nsi
amrbashir: no worries, whenever you have time

tooling/bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -493,63 +496,90 @@ SectionEnd
...
+        ${Else}
+          ReadRegStr $R1 HKLM "SOFTWARE\Microsoft\EdgeUpdate" "path"
+        ${EndIf}
+        ${If} $R1 == ""
+          ReadRegStr $R1 HKCU "SOFTWARE\Microsoft\EdgeUpdate" "path"
+        ${EndIf}
+        ${If} $R1 != ""
+          ; Chromium updater docs: https://source.chromium.org/chromium/chromium/src/+/main:docs/updater/user_manual.md
+          ; Modified from "HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\Microsoft EdgeWebView\ModifyPath"
+          ExecWait `"$R1" /install appguid=${WEBVIEW2APPGUID}&needsadmin=true` $1
```
Legend-Master: Just note that I can't get it working without `needsadmin` even on the user install

.changes/nsis-ensure-webview2-version.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri-build": "patch:feat"
```
amrbashir:
```suggestion
"tauri-bundler": "patch:feat"
"tauri-utils": "patch:feat"
```

tooling/bundler/src/bundle/windows/templates/installer.nsi
amrbashir: then we shouldn't abort on failure, maybe just leave a warning with a dialog box to continue?

amrbashir (changes requested):

tooling/bundler/src/bundle/windows/templates/installer.nsi
Legend-Master: Makes sense in that case, I was thinking it should fail because if the version is lower some APIs might not be available (we made wry works on earlier versions so this should be mitigated a lot now)

Do I need to translate the warning to all those languages? Not sure how this is done before

amrbashir: well, languages are translated and proof-read by the community, but if you can, maybe use google translate to have initial translation as a starting point

Legend-Master: I feel like the `Error: Installing WebView2 failed with exit code $1` message might be enough, if you think it's better to get a custom message here, I can work on it tomorrow

amrbashir: that's fine with me, let's roll with it for now and improve later if needed

core/tauri-utils/src/config.rs
```diff
@@ -838,6 +838,11 @@ pub struct NsisConfig {
   /// ```
   #[serde(alias = "installer-hooks")]
   pub installer_hooks: Option<PathBuf>,
+  /// Ensure webview2 version to be newer than this version,
+  /// if the user's webview2 is older than this version,
+  /// the installer will trigger a webview2 update
+  #[serde(alias = "ensure-webview2-version")]
+  pub ensure_webview2_version: Option<String>,
```
amrbashir: I feel like `minimum_webview2_version` is a better name?

core/tauri-utils/src/config.rs
```diff
@@ -838,6 +838,11 @@ pub struct NsisConfig {
   /// ```
   #[serde(alias = "installer-hooks")]
   pub installer_hooks: Option<PathBuf>,
+  /// Ensure webview2 version to be newer than this version,
+  /// if the user's webview2 is older than this version,
+  /// the installer will trigger a webview2 update
```
```suggestion
  /// Try to ensure WebView2 version is newer or equal to this version,
  /// if the user's WebView2 is older than this version,
  /// the installer will try to trigger a WebView2 update.
```

amrbashir (changes requested):

.changes/nsis-ensure-webview2-version.md
```diff
@@ -0,0 +1,6 @@
+---
+"tauri-bundler": "patch:feat"
+"tauri-utils": "patch:feat"
+---
+
+Add a new option `ensureWebview2Version` for Windows NSIS installer to trigger a webview2 update if the user's webview2 is older than this version
```
amrbashir:
```suggestion
Add a new option `minimumWebview2Version` for Windows NSIS installer to trigger a webview2 update if the user's webview2 is older than this version
```

amrbashir (approved):

amrbashir (approved):

PR #10356, opened 2024-07-22
feat(core): fallback to file system for AssetResolver::get, closes #8411

### LINKED ISSUE #8411: AssetResolver behavior is inconsistent between dev and production
### Describe the problem

It took me quite a while to figure out how `AssetResolver` is used because it seems to produce no result. Only after studying the source code I realized that this behavior is specific to dev mode because there are no embedded assets there. Production builds show the expected behavior however.

I ended up implementing a fallback: if `AssetResolver` doesn’t work, implement the same functionality by listing files in `config.dist_dir`. I’d expect this fallback to be part of `AssetResolver` itself, so that it delivers consistent results both in dev and production mode.

### Alternatives considered

At the very least, this footgun and the proposed fallback need to be documented. As things are right now, there isn’t a word in the documentation mentioning this issue.

### Additional context

I’m using Tauri 1.5.3. However, judging by the source code the problem hasn’t been addressed in Tauri 2.0 either.

lucasfernog:
Changes the AssetResolver::get implementation to fallback to reading the distDir directly when using the development mode pointing to a external URL (which is the way most people use the dev mode and means there's no embedded assets to pull from).


tillmann-crabnebula:
LGTM in general. Some Tests failing due to breaking change in `Context`.

chip-crabnebula (approved):

chippers (approved):

Issue #9803, opened 2024-05-17
[bug] file association issues


thewh1teagle:
### Describe the bug

I use file association in v2 app, and it works great in macOS. However in Linux it doesn't work at all and the app isn't suggested by right clicking some file.
In Windows it's partially works, with nsis: it doesn't list the app on right click but I also need to click "Choose another app" under the suggested apps context menu.
With msi it doesn't list it at all.

### Reproduction

Add file association to tauri app, compile and install in Linux/Windows and then right click some file with the desired extension 

### Expected behavior

It should list in suggested app just by right click some file and choosing open with

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.22631 X64
    ✔ WebView2: 124.0.2478.97
    ✔ MSVC: Visual Studio Community 2022
    ✔ rustc: 1.77.1 (7cf61ebde 2024-03-27)
    ✔ cargo: 1.77.1 (e52e36006 2024-03-26)
    ✔ rustup: 1.27.0 (bbb9276d2 2024-03-08)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (default)
    - node: 20.12.0
    - npm: 10.5.0
    - bun: 1.1.7

[-] Packages
    - tauri [RUST]: 2.0.0-beta.19
    - tauri-build [RUST]: 2.0.0-beta.15
    - wry [RUST]: 0.39.5
    - tao [RUST]: 0.28.0
    - @tauri-apps/api [NPM]: 2.0.0-beta.11
    - @tauri-apps/cli [NPM]: 2.0.0-beta.17

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:1420/
    - framework: React
    - bundler: Vite
```

### Additional context

Repro open source app [vibe](https://github.com/thewh1teagle/vibe)

martpie:
Also, the [file association example](https://github.com/tauri-apps/tauri/blob/dev/examples/file-associations) is pretty hard to follow and implement compared to other capabilities allowed by plugins. I wonder if there is a friendlier way of doing cross-platform file association?

metkm:
Still, this is an issue. I can't make it work for some reason using msi on Windows. It doesn't get listed at all. It looks like using NSIS works but not MSI.

metkm:
So, the workaround for this for now is to do it manually yourself and extend the .wxs file that was automatically generated. I'm going to put an example here.

[...]

PR #10358, opened 2024-07-22
fix(bundler): Add support for File Association in Linux by default

naman-crabnebula:
- Fixes Linux part of https://github.com/tauri-apps/tauri/issues/9803

Added `%F` to open the file with Tauri Application.

However, developer have to handle the argument received himself ( obviously )

### DIFF
--- a/.changes/bundler-fix-file-association.md
+++ b/.changes/bundler-fix-file-association.md
@@ -1,0 +1,5 @@
+---
+"tauri-bundler": "patch:feat"
+---
+
+Add support for File Association in Linux by default.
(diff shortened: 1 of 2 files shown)


FabianLars:
If i understand the specification correctly then you can only have one of %f %F %u %U in there but never 2. We need %u or %U for deep links though (wherever that went...). It sounds like if you use %U then it includes %F but as `file://` urls. iirc we have the same behavior on macOS so it shouldn't be too bad.

Any thoughts on that?

Also, should we make this change depend on whether file associations/deep links where configured at all?

PR #10360, opened 2024-07-23
use loopback interface on desktop dev server by default

chippers:
This makes the desktop dev server use the loopback interface by default, only exposing itself to the public network if the `--expose` flag is passed.

cc: @amrbashir do you think this change in behavior from the cli is fine?


tooling/cli/src/dev.rs
```diff
@@ -205,12 +207,18 @@ pub fn setup(
             url.set_host(Some(&local_ip_address))?;
           }
         } else {
+          let default = if options.expose {
+            "0.0.0.0"
+          } else {
+            "127.0.0.1"
+          };
+
```
amrbashir: The `--expose` flag doesn't make sense here to be honest, since this piece of logic doesn't expose anything, it just replaces `$HOST` string in the `beforeDevCommand` so the `beforeDevCommand` can know what the server IP is. 

The most common [usecase](https://github.com/tauri-apps/create-tauri-app/blob/4eddf49d3bac3714b513bf5e6c8f9753f40aba08/templates/template-angular/.manifest#L11) for `$HOST`, is when developing on mobile, where the frontend server needs to be accessed on a local network IP such as `192.168.1.12` not `127.0.0.1` and since this ip can change between differen `tauri dev` invocations, we replace `$HOST` in `beforeDevCommand`.

That said, using `127.0.0.1` as the default instead of `0.0.0.0` is fine since it doesn't affect mobile development (which is the main reason of using `0.0.0.0`). We can even skip adding `--expose` flag.

amrbashir (approved):

PR #10361, opened 2024-07-23
Fix Specta integration

oscartbeaumont:
#10139 introduced a usage of `specta::Type` (the derive macro) which is not enabled by default. This PR adds the required `derive` feature to Specta.

This seems to have caused issues generating Rust documentation for the Tauri 2.0.0-beta.24 release. Due to the nature of Cargo dependencies this will only ever cause an error when using Tauri with the `specta` feature without the user having `specta` with the `derive` feature which should be exceedingly rare.


oscartbeaumont:
Closed as duplicate of #10362

PR #10362, opened 2024-07-23
fix: use specta's derive feature

Brendonovich:
Specta recently moved the `specta::Type` macro to the `derive` feature, which broke the docs.rs build for beta 24. This PR fixes that, moving to specta rc 16 as the minimum and using the `derive` feature.

### DIFF
--- a/.changes/specta-derive-feature.md
+++ b/.changes/specta-derive-feature.md
@@ -1,5 +1,5 @@
 ---
-"tauri": patch:fix
+"tauri": patch:bug
 ---
 
 use specta rc 15's derive feature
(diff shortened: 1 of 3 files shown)


oscartbeaumont:
Published Specta 2.0.0-rc.16 which should hopefully apply the fix to the CI issues we are seeing here.

I don't have access to the branch so @Brendonovich give it a shot.

.changes/specta-derive-feature.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri": patch:fix
```
FabianLars:
```suggestion
"tauri": patch:bug
```
I'm so happy it's not just me who makes this mistake all the time 😂

FabianLars (dismissed):

FabianLars (approved):

PR #10368, opened 2024-07-23
fix: DragDropEvent should have the same structure

tance77:
The types for the DragDropEvent should be the same regardless and not present if no data. 

3. If this change requires a new version, then add a change file in `.changes` directory with the appropriate bump, see https://github.com/tauri-apps/tauri/blob/dev/.changes/README.md

Unsure if it does or not.

### DIFF
--- a/tooling/api/src/webview.ts
+++ b/tooling/api/src/webview.ts
@@ -33,11 +33,7 @@ import { Window, getCurrentWindow } from './window'
 import { WebviewWindow } from './webviewWindow'
 
 /** The drag and drop event types. */
-type DragDropEvent =
-  | { type: 'enter'; paths: string[]; position: PhysicalPosition }
-  | { type: 'over'; position: PhysicalPosition }
-  | { type: 'drop'; paths: string[]; position: PhysicalPosition }
-  | { type: 'leave' }
+type DragDropEvent = {type: 'enter' | 'over' | 'drop' | 'leave', paths?: string[], position?: PhysicalPosition}
 
 /**
  * Get an instance of `Webview` for the current webview.
--- a/tooling/api/src/webview.ts
+++ b/tooling/api/src/webview.ts
@@ -513,8 +513,8 @@ class Webview {
    * ```typescript
    * import { getCurrent } from "@tauri-apps/api/webview";
    * const unlisten = await getCurrentWebview().onDragDropEvent((event) => {
-   *  if (event.payload.type === 'hover') {
-   *    console.log('User hovering', event.payload.paths);
+   *  if (event.payload.type === 'over') {
+   *    console.log('User over', event.payload.paths);
    *  } else if (event.payload.type === 'drop') {
    *    console.log('User dropped', event.payload.paths);
    *  } else {
@@ -555,7 +555,8 @@ class Webview {
           ...event,
           payload: {
             type: 'over',
-            position: mapPhysicalPosition(event.payload.position)
+            paths: [],
+            position: mapPhysicalPosition(event.payload.position),
           }
         })
       }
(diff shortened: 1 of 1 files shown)


tooling/api/src/webview.ts
```diff
@@ -33,11 +33,7 @@ import { Window, getCurrentWindow } from './window'
 import { WebviewWindow } from './webviewWindow'
 
 /** The drag and drop event types. */
-type DragDropEvent =
-  | { type: 'enter'; paths: string[]; position: PhysicalPosition }
-  | { type: 'over'; position: PhysicalPosition }
-  | { type: 'drop'; paths: string[]; position: PhysicalPosition }
-  | { type: 'leave' }
+type DragDropEvent = {type: 'enter' | 'over' | 'drop' | 'leave', paths?: string[], position?: PhysicalPosition}
```
amrbashir, line 36: this new type definition doesn't match what the events actually is, and will make this snippet
```ts
if (event.type === 'enter') {
	console.log(event.position);
	console.log(event.paths);
}
```
turn into
```ts
if (event.type === 'enter') {
	if (event.position) {
		console.log(event.position);
	}
	if (event.paths) {
		console.log(event.paths);
	}
}
```

amrbashir (changes requested):

tooling/api/src/webview.ts
tance77: There are a few issues either way:

- if we use the old types, then we end up with it not understanding which properties are even set
- if we use the new types, we do need to check if they are nullable, but at least typescript understands they are present

Unless there is another way to handle this in TS? Currently, even if we cast it to the DragDropEvent, we still have to do checks everywhere because it does not seem to understand which of the object forms we're working with.

amrbashir: > * if we use the old types, then we end up with it not understanding which properties are even set

how so? are you using `listen` or `window.onDragDropEvent`?

here is a typescript playground for how it behaves with the old types https://www.typescriptlang.org/play/?target=99#code/JYOwLgpgTgZghgYwgAgKIDcLgDwBUB8yA3gFDLISbgBcyAzmFKAOZnLAAmtIArgLYAjaGwAOcAJ4AbAPZwuyXCQC+JEmHEiUGLGADCcSZIGIA1nkIBeZAApKO2tpwEAlMguF00zqrUaUABQALcTpgBAN-aVCwYGkQN2IAD25+ISgAGmRxFMFoZBV1TWQAESg4ZlLpEUcwNzYAH2JkQohaAHIdaDaAbmQxMEC6WgYmEGYAbQBdXpEo4Bi42iCQsIi5hfiVckaiZr926UwoHr712JAl4NDwyUjo8-yGppb2jigqk-7B4cYWKZmzotkMtrmt7nFHttnvtkG1JBA4Jg2o8SHA6OIQAhkDAeJiNsg4qVypVqlQwNY2ORAnAQBx4VAHGT9IZjAgzESKu9STp8GxnEt3nxgHQINhPJxCKQVCRCWVOVUatZbGTXO5iGxgDAbHZwAA6MRSWQcXUtNxWABEnSg5tcpHI5AQcTo0nhupkzGVOn1Ehkcm9AzozjY0vImu1ZO9hr9posFvhiIgNvV9uQjpAztd7s9eoNvuNX0DyAA9EXkCBpDEkM1AihoO8MsghOEeCLkAADeOYNsUMnIDjSCB0EBtWrUzB9OAB4PKZxAA

tance77: When updating our dependencies our typescript server didn't update to the new types and was stuck in a frozen state. This example you gave works after restarting our Typescript server. 

I am going to close this as a nonissue.

PR #10390, opened 2024-07-25
refactor: core plugin permissions are now prefixed core:, closes #10359

### LINKED ISSUE #10359: [feat] Tauri Core Plugin Namespace Prefix
### Describe the problem

Currently all Tauri core APIs are pseudo plugins.

They implement the plugin trait but are not individual crates following the `tauri-plugin-` naming scheme, as this would cause circular dependencies for the code depending on tauri and the tauri code depending on the "plugin".

This means currently we:

* Need to reserve/block all tauri-plugin-<core plugin name> plugins

* Can not differentiate between core and external plugins

To mitigate some problems we propose a `core-` prefix for all core plugins and to document and enforce that the `core-` plugin "namespace"  is only able to be used from within the tauri core and not from any other plugin.

- [ ] All `tauri-plugin-core-*` crates should be blocked from being initialized outside of the Tauri core.
- [ ] No `tauri-plugin-core` crate should be valid. Maybe we reserve the crate name to avoid confusion?
- [ ] The core plugins should be renamed as `core-<core-plugin-name>` (e.g. `core-path`, `core-window`, `core-webview`, ...).

The documentation should be updated to reflect this behavior.

As this is a breaking change we need to test this explicitly and involve multiple people for a review.
This will add minimal boilerplate in the code and permission files but should be mostly mitigated by having a `core:default` permission set for all core plugins.

### Alternatives considered

We considered changing the architecture to be able to handle more than one namespace like `core:app:reload` instead of `app:reload` but this is also a breaking change involving a bigger refactor and we still have the `core` prefix/namespace to be reserved.

### Additional context

Was initially discussed with @chippers and @lucasfernog

lucasfernog:


tillmann-crabnebula:
Do we want to introduce the `core:default` permission set in this PR as well?
This one would combine all core plugins default permissions.

.changes/core-plugin-namespace.md
```diff
@@ -0,0 +1,9 @@
+---
+"tauri": patch:breaking
+"tauri-plugin": patch:breaking
+"@tauri-apps/cli": patch:breaking
+"tauri-cli": patch:breaking
+---
+
+Core plugin permissions are now prefixed with `core:` and the `core` plugin name is reserved.
+The `tauri migrate` tool will automate the migration process, which involves prefixing all `app`, `event`, `image`, `menu`, `resources`, `tray`, `webview` and `window` permissions with `core:`.
```
tillmann-crabnebula: Also missing the `path` pseudo plugin.

core/tauri-utils/src/acl/capability.rs
```diff
@@ -174,16 +174,16 @@ pub struct Capability {
...
+  ///  "core:path:default",
+  ///  "core:event:default",
+  ///  "core:window:default",
+  ///  "core:app:default",
+  ///  "core:image:default",
+  ///  "core:resources:default",
+  ///  "core:menu:default",
+  ///  "core:tray:default",
+  ///  "core:shell:allow-open",
+  ///  "core:dialog:open",
```
tillmann-crabnebula: These are no core plugins afaik

tillmann-crabnebula (commented):
Looks mostly good to me. Some small observations we should fix.

Edit: Also some docstring lints to fix: https://github.com/tauri-apps/tauri/actions/runs/10101389779/job/27934877226?pr=10390#step:6:529

.changes/core-plugin-namespace.md
lucasfernog: pushed

core/tauri-utils/src/acl/capability.rs
lucasfernog: oops, pushed

lucasfernog:
> Do we want to introduce the `core:default` permission set in this PR as well? This one would combine all core plugins default permissions.

Pushed, but had to make a little workaround because we do not support cross-plugin referencing of permissions (we decided we wouldn't define plugin dependencies here to simplify the work).

tweidinger (approved):

tillmann-crabnebula (approved):

PR #10392, opened 2024-07-26
chore: emit rustc-check-cfg for rust 1.80

anatawa12:
This is almost same as #9850.
I tried cherry-picking 9ac930380a5df3fe700e68e75df8684d261ca292 but it caused many conflicts so I reimplemented manually.


lucasfernog (approved):
thank you!

PR #10404, opened 2024-07-28
fix: Conf parsing error always displays `tauri.conf.json` file name even when using toml or json5

ernstvanderlinden:
Conf parsing error always displays `tauri.conf.json` as path, even when using `Tauri.toml` or `tauri.conf.json5`

Example Error Message when using Tauri.toml:
Error unable to parse toml Tauri config file at
/Users/bla/repo/bla/bla/src-tauri/tauri.conf.json because invalid TOML value, did you mean to use a quoted string? at line 41 column 7


FabianLars:
Thank you! Would you mind adding a changefile (see .changes dir) too?

ernstvanderlinden:
Thanks, done Fabian

FabianLars (approved):

Issue #8104, opened 2023-10-25
[bug] [macOs] Tauri updater fails with "Tauri API error: Permission Denied (os error 13)" for standard (non-admin) users


0rvar:
### Describe the bug

The built in Tauri updater (and the official updater plugin for v2) fails with a cryptic error message for "Standard" users, meaning users that are not administrators on their Mac. 

### Reproduction

https://github.com/tauri-apps/tauri/assets/1098408/b7a447d8-214d-4c28-bc90-168bddf34a50

1. Create and login to a standard user on mac (not an administrator user).
2. Download the next-newest version from https://github.com/spacedriveapp/spacedrive/releases (or any other popular tauri app)
3. Install the app to /Applications, type in an administrator username and password when prompted by the OS
4. Run the app and accept the update prompt
5. Updater fails with "Tauri API error: Permission Denied (os error 13)"

### Expected behavior

The scenario should be a first-class citizen, first of all - less cryptic error message, documented.

I have a working, semi tested version of the updater function for mac (`#[cfg(target_os = "macos")] fn copy_files_and_run`) here:
https://github.com/0rvar/tauri/commit/7c740516efaa4ff8b14b27c31d3c9a2cf54e21ed

Let me know if I should create a signed PR with this approach.

It prompts for elevated privileges if needed using Applescript. There are other approaches as well:
* Using `security_AuthorizationExecuteWithPrivileges`, but this has been deprecated for a very long time
* Using SMJobless. No idea how that works

### Platform and versions

```text
Mac, any version
```

Caleb-T-Owens:
Hi! I've encountered this issue same too. It would be great to know if there has been any movement on this 👍

kurdin:
@Caleb-T-Owens  I hit the same wall, trying to figure out what to do in case of this error. Did you find any solutions ?

jamesfebin:
Facing the same issue, help is appreciated.

kurdin:
@jamesfebin 

> Facing the same issue, help is appreciated.

I could not find any good automatic solution without hacking/patching the Tauri source code, so here is what I did:

On the frontend, I check for permission errors on update, and if any, I show a notification banner to the user with a message that the update needs to be downloaded and installed manually with a user that has correct (admin) permissions.

Here is my frontend code for checking for the error :
```js
try {
      await checkUpdate();
      await installUpdate();
      get().setIsFirstRunAfterUpdate(true);
      if (availableVersionNumber.value) {
[...]

PR #10427, opened 2024-07-30
fix(macOS): Tauri Update Permission Denied Error

jLynx:
- Addressed the problem of Tauri updater failing for standard (non-admin) users on macOS with a "Permission Denied" error as stated in this issue https://github.com/tauri-apps/tauri/issues/8104.
- Modified the `copy_files_and_run` function for macOS to handle permission issues more gracefully.
- Introduced an AppleScript-based solution to request administrator privileges when needed:
  - If a permission error is detected, the script prompts for admin credentials.
  - Uses `osascript` to execute shell commands with elevated privileges.
- Added a helper function `is_permission_error` to identify permission-related errors.
- Ensured that the solution works for both standard and admin users, providing a more robust update mechanism.

image

Credit to @0rvar as this code is from them


jLynx:
Hey team!
Just wondering what the PR cycle time & release timeline for Tauri? Just wondering on average how long this will be before it goes into prod

tweidinger:
Timelines and review times are very much flexible and depend on people/code owners/working group members available. 

The initial review showed some potential issues with TOCTU (Time of Check Time of Use) attacks ([old reference](https://developer.apple.com/library/archive/documentation/Security/Conceptual/SecureCodingGuide/Articles/RaceConditions.html)) and I am still not sure if we are fine to accept the potential risk in this specific case.

Currently we do a signature validation in memory and not write to disk to prevent such TOCTU attacks, which has its issues and one of them caused this PR. With writing to the temp folder we open up to other processes of the system and just replacing the files in temp folder would allow another unprivileged application to gain admin rights in the worst case.

The solution here does use a library which limits the umask to the current user by [default](https://docs.rs/tempfile/latest/tempfile/struct.Builder.html#security), so no direct risk from other users but only other processes from the same user. The elevation is only used for moving files, so no risk of direct execution. The only direct risk would be other processes are able to manipulate the Tauri app or replace it with their own binary. This can have impact depending on how your system or organization is set up and how much untrusted binaries are running on your system :grinning:.

I'll discuss once more with @chippers and maybe @lucasfernog on what we think about this and if there are changes needed but can't give you a specific timeline.

jLynx:
I appreciate the feedback & update @tweidinger

lucasfernog:
unfortunately we can't use osakit here like https://github.com/tauri-apps/plugins-workspace/pull/2067 since it required Rust 1.71 - so let's keep using osascript :)

lucasfernog (approved):

PR #10457, opened 2024-08-02
feat(cli): allow skipping rustfmt project reformatting when adding a plugin

mmvanheusden:
Normally, each time, for example `pnpm tauri add dialog` is ran, the project is automatically reformatted using rustfmt (`cargo fmt`).

I added an option allowing it to be skipped. Maybe it is even better to have it removed all together?
I don't think adding a plugin to a project should automatically reformat all your code. Those two things are entirely unrelated.

### DIFF
--- a/tooling/cli/src/add.rs
+++ b/tooling/cli/src/add.rs
@@ -229,4 +229,3 @@ pub fn command(options: Options) -> Result<()> {
 
   Ok(())
 }
-
--- a/tooling/cli/src/add.rs
+++ b/tooling/cli/src/add.rs
@@ -86,9 +86,9 @@ pub struct Options {
   /// Git branch to use.
   #[clap(short, long)]
   pub branch: Option<String>,
-  /// Don't reformat code with rustfmt
+  /// Don't format code with rustfmt
   #[clap(long)]
-  pub no_reformatting: bool,
+  pub no_fmt: bool,
 }
 
 pub fn command(options: Options) -> Result<()> {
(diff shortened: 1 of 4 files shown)


tooling/cli/src/add.rs
```diff
@@ -86,6 +86,9 @@ pub struct Options {
   /// Git branch to use.
   #[clap(short, long)]
   pub branch: Option<String>,
+  /// Don't reformat code with rustfmt
```
lucasfernog:
```suggestion
  /// Don't format code with rustfmt
```

tooling/cli/src/add.rs
```diff
@@ -86,6 +86,9 @@ pub struct Options {
...
+  #[clap(long)]
+  pub no_reformatting: bool,
```
lucasfernog:
```suggestion
  pub no_fmt: bool,
```

tooling/cli/src/add.rs
```diff
@@ -185,12 +188,15 @@ pub fn command(options: Options) -> Result<()> {
       log::info!("Adding plugin to {}", file.display());
       std::fs::write(file, out.as_bytes())?;
 
-      // run cargo fmt
-      log::info!("Running `cargo fmt`...");
-      let _ = Command::new("cargo")
-        .arg("fmt")
-        .current_dir(&tauri_dir)
-        .status();
+      if !options.no_reformatting {
```
lucasfernog:
```suggestion
      if !options.no_fmt {
```

lucasfernog (changes requested):

lucasfernog (approved):

PR #10462, opened 2024-08-03
chore: Update dev container configuration

deadmeu:
Updates the dev container configuration to use the default Microsoft supported rust environment.

Support for GUI rendering is also included for Linux hosts using Wayland.

Tested & used on a Linux machine running Wayland, using Podman (instead of Docker).

Opening as a draft to solicit feedback and potentially develop further to add better support for X11, as well as Windows & macOS.


deadmeu:
Hi @simonhyll not sure if this is interesting to you at this stage, but thanks for your previous work and videos covering dev containers & Tauri. I think it would be great if this was updated and used as a guide for others looking to use dev containers to develop their Tauri applications, and I think it would be great if your work targeting Android was included. I think each type of target would be best handled via separate `devcontainer.json` files. If you have any thoughts or opinions please share them.

vanarok:
> Обновляет конфигурацию контейнера разработки для использования среды Rust, поддерживаемой Microsoft по умолчанию.
> 
> Поддержка рендеринга графического интерфейса также включена для хостов Linux, использующих Wayland.
> 
> Протестировано и использовано на машине Linux под управлением Wayland с использованием Podman (вместо Docker).
> 
> Открывается как черновик для сбора отзывов и потенциальной дальнейшей разработки с целью добавления лучшей поддержки X11, а также Windows и macOS.

Are you using a podman? With userns keep-id the container goes down. 
Only the host value is available in docker for userns.
https://docs.docker.com/reference/cli/docker/container/run/#userns

PR #10470, opened 2024-08-03
set `packageManager` field in package.json for corepack

chippers:
`git log -S '"packageManager"'` shows no history of this field ever being set, except in this commit. We use it in the docs repo, is there a reason we haven't used it here yet? If no explicit reason no then this makes working with corepack nicer

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,5 +22,6 @@
   },
   "dependencies": {
     "typescript": "^5.4.5"
-  }
+  },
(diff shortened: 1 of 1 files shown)


FabianLars:
> We use it in the docs repo

to be fair, iirc we only started to use it there because it's the only way to tell netlify which version to use (also works via corepack though).

chippers:
> > We use it in the docs repo
> 
> to be fair, iirc we only started to use it there because it's the only way to tell netlify which version to use (also works via corepack though).

ah ok. my recent version of corepack is automatically adding this field to every package.json if it doesn't exist for yarn, npm, pnpm for all my repos

lucasfernog (approved):

PR #10481, opened 2024-08-04
fix(cli): v1 migration should add the updater plugin if it is enabled

lucasfernog:

### DIFF
--- a/tooling/cli/src/migrate/migrations/v1/config.rs
+++ b/tooling/cli/src/migrate/migrations/v1/config.rs
@@ -111,11 +111,7 @@ fn migrate_config(config: &mut Value) -> Result<MigratedConfig> {
       }
 
       // cli
-      process_updater(tauri_config, &mut plugins)?;
-    }
-
-    if plugins.contains_key("updater") {
-      migrated.plugins.insert("updater".to_string());
+      process_updater(tauri_config, &mut plugins, &mut migrated)?;
     }
 
     config.insert("plugins".into(), plugins.into());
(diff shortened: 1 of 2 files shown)


tooling/cli/src/migrate/migrations/v1/config.rs
```diff
@@ -114,6 +114,10 @@ fn migrate_config(config: &mut Value) -> Result<MigratedConfig> {
       process_updater(tauri_config, &mut plugins)?;
     }
 
+    if plugins.contains_key("updater") {
```
FabianLars: shouldn't this be `if let Some(updater) = tauri_config.remove("updater") {` (moved into the above block)?

lucasfernog: I didn't want to add the migrated argument (lazy) but it's indeed cleaner that way.. pushed

FabianLars: ah tbh i just missed the process_updater line above that (its comment is saying `cli` btw) so i thought the code is actually incorrect.

tooling/cli/src/migrate/migrations/v1/config.rs
```diff
@@ -111,7 +111,7 @@ fn migrate_config(config: &mut Value) -> Result<MigratedConfig> {
       }
 
       // cli
```
FabianLars:
```suggestion
      // updater
```

FabianLars (approved):

PR #10495, opened 2024-08-05
feat(core): add config for the minimum iOS version

lucasfernog:


FabianLars:
Is 13 the lowest we can go nowadays? Just asking whether we should update the readme which still lists 9

lucasfernog:
good question.. i'd need to check on another machine, my main one only lets me try iOS 15+

PR #10498, opened 2024-08-05
fix(bundler): Correct nsis pre-uninstall hook to post-uninstall

catalinsh:
The NSIS_HOOK_PREUNINSTALL macro was where NSIS_HOOK_POSTUNINSTALL should have been.


FabianLars (approved):
thank you!

PR #10499, opened 2024-08-05
fix: Fix bundler&cli MSRV

FabianLars:
I added the bundler and cli to the msrv check and while trying to downgrade all the crates to the bundler's MSRV (1.60) it appeared that the bundler and cli were effectively already on 1.70/1.71 (both in code syntax and cargo.toml deps) and downgrading them required more changes than i'd like to do.

Therefore this isn't really a breaking change and imo should be fine to be part of a simple patch release.
1.70 is a bit over 1 year old now.

The bundler and cli msrv checks are seperate because a) i tried to adhere to 1.60 for the bundler first and b) we bump the cli msrv more often at which point a seperate check will be required anyway.


FabianLars:
~~I'll be afk for a few hours so if any of the ci checks fail i'll need a while to fix them~~ nevermind, see ya tomorrow (CI is happy any way)

lucasfernog (approved):

PR #10504, opened 2024-08-06
feat(cli): detect project NPM package manager on init

### LINKED ISSUE #10257: [feat] pnpm tauri init at an existing next js project
### Describe the problem

its not exactly a problem but I would like to propose a QoL feature to pnpm tauri init when run on an existing project.
Screenshot 2024-07-11 180116
as you see in the image it asks for frontend dev command and defaults to npm run dev

I propose there is some more options before asking for frontend dev command:

a question before asking for build command:
existing project? > yes / no
what do you use?
>npm
>pnpm
>yarn

and then ask for dev and build command with the correct defaults!!!

### Alternatives considered

_No response_

### Additional context

_No response_

fu050409:
Hi, there's duplicate interaction in command `cargo tauri init` (or using other package managers for `init`). Instead of repeatedly asking for the `dev` and `build` commands, we could just simply let the users choose which package manager they want to use and automatically select the corresponding commands. This works the same as `pnpm create tauri`.

Resolved: #10257


tooling/cli/src/init.rs
```diff
@@ -130,29 +130,41 @@ impl Options {
...
-      .map(|s| Ok(Some(s)))
-      .unwrap_or_else(|| {
-        prompts::input(
-          "What is your frontend dev command?",
-          Some("npm run dev".to_string()),
-          self.ci,
-          true,
-        )
-      })?;
+    let package_manager =
```
lucasfernog: we can try to infer the package manager by checking if a yarn.lock/package-lock.json/pnpm-lock.yaml exists: https://github.com/tauri-apps/tauri/blob/008b9f7481db3fcfaff5424e4b25a2399959be17/tooling/cli/src/helpers/npm.rs#L36

tooling/cli/src/init.rs
```diff
@@ -130,29 +130,41 @@ impl Options {
...
-          self.ci,
-          true,
-        )
-      })?;
+    self.before_dev_command =
+      self
+        .before_dev_command
+        .map(Some)
+        .unwrap_or(match package_manager.as_str() {
+          "npm" => Some("npm run dev".to_string()),
```
lucasfernog: using `dev` and `build` as defaults isn't desired.. some frontends use `start` instead of `dev` and if you have a monorepo you might have a `frontend:dev` script instead...

lucasfernog (changes requested):
using the create-tauri-app logic here doesn't help us since we must support more kinds of projects (CTA already knows the scripts in the project since it controls the entire project template). I would only focus on detecting the package manager and using it on the default script while still prompting the user (instead of the current `npm run` default)

tooling/cli/src/init.rs
fu050409: Yes, you're right. This is where I was thoughtless, I'll set a default command that corresponds to the package manager, but still prompt the user for the exact command.

tooling/cli/src/init.rs
fu050409: Thanks and I'll refactor the codes.

lucasfernog (approved):

PR #10569, opened 2024-08-11
fix(cli): `ios dev` and `ios build` support for `bun`

lucasfernog:

### DIFF
--- a/tooling/cli/src/mobile/ios/xcode_script.rs
+++ b/tooling/cli/src/mobile/ios/xcode_script.rs
@@ -14,7 +14,7 @@ use clap::Parser;
 
 use std::{
   collections::HashMap,
-  env::{current_dir, set_current_dir, var_os},
+  env::{current_dir, set_current_dir, var, var_os},
   ffi::OsStr,
   path::{Path, PathBuf},
   process::Command,
@@ -62,7 +62,10 @@ pub fn command(options: Options) -> Result<()> {
   }
 
   // `xcode-script` is ran from the `gen/apple` folder when not using NPM.
-  if var_os("npm_lifecycle_event").is_none() && var_os("PNPM_PACKAGE_NAME").is_none() {
+  // so we must change working directory to the src-tauri folder to resolve the tauri dir
+  if (var_os("npm_lifecycle_event").is_none() && var_os("PNPM_PACKAGE_NAME").is_none())
+    || var("npm_config_user_agent").map_or(false, |agent| agent.starts_with("bun"))
+  {
     set_current_dir(current_dir()?.parent().unwrap().parent().unwrap()).unwrap();
   }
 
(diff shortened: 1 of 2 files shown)


tooling/cli/src/mobile/ios/xcode_script.rs
```diff
@@ -62,7 +62,10 @@ pub fn command(options: Options) -> Result<()> {
   }
 
   // `xcode-script` is ran from the `gen/apple` folder when not using NPM.
-  if var_os("npm_lifecycle_event").is_none() && var_os("PNPM_PACKAGE_NAME").is_none() {
+  // so we must change working directory to the src-tauri folder to resolve the tauri dir
+  if (var_os("npm_lifecycle_event").is_none() && var_os("PNPM_PACKAGE_NAME").is_none())
+    || var("npm_config_user_agent").map_or(false, |agent| agent.starts_with("bun"))
```
amrbashir, line 67: should we remove this check and always set the directory?

lucasfernog: when using `pnpm` `npm` `yarn` etc the CWD is automatically set to the app dir, so moving up two parents would crash it.
when using cargo or bun it'll actually be under gen/apple, and we can't resolve the tauri dir from there.. so we need this condition

lucasfernog: we could instead check if CWD ends with `gen/apple` idk if it helps

amrbashir: `xcode-script` and `android-script` are only valid if they are executed from a process that was launched by `ios dev/build` or `android dev/build`, right? `ios dev/build` or `android dev/build` could pass the app_dir using either the options mechanism we have or through an environment variable.

amrbashir: also looking at https://github.com/tauri-apps/tauri/blob/028cffede293e70b4766192763dd0d033a8b75ad/tooling/cli/src/mobile/ios/xcode_script.rs#L72 , we already resolve the app_dir, why not use that and set the directory to it.

lucasfernog: we can only resolve because we changed the CWD earlier..
to use the options we need to resolve the tauri config, which means we need to resolve the tauri dir too, so it doesn't help

lucasfernog: and xcode isn't a big fan of environment variables, plus if there's already an opened instance it wouldn't receive it

amrbashir: alright, let's merge as is and hope it doesn't crash again.

amrbashir (approved):

bfish510:
I believe this may still be an issue. I changed the rust build phase of the Xcode project from `bun tauri ios xcode-script -v --platform` to `cargo tauri ios xcode-script -v --platform` to fix iOS builds. Unsure of the repercussions of this, but it's unblocked me.

PR #10572, opened 2024-08-11
fix(cli): properly exit with code 1 on panic when running with bun

lucasfernog:
Found out about this when testing #10569 (maybe bun handles this differently when using NAPI?)

### DIFF
--- a/.changes/fix-cli-panic-bun.md
+++ b/.changes/fix-cli-panic-bun.md
@@ -1,0 +1,5 @@
+---
+"@tauri-apps/cli": patch:bug
+---
+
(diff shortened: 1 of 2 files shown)


amrbashir (commented):
do we really need to do anything about it? sounds like a bug in bun and we should just report it to bun instead of adding a workaround?

lucasfernog:
> do we really need to do anything about it? sounds like a bug in bun and we should just report it to bun instead of adding a workaround?

I'm not even sure if it's really a bug in bun, or just a ""feature"" of Node.js 😂 makes sense anyway for us to catch_unwind here, we're running Rust from another language

amrbashir (approved):

PR #10577, opened 2024-08-12
refactor(bundler): output file name consistency, closes #10465

### LINKED ISSUE #10465: release asset naming inconsistencies
My {app_name} currently is 'deploy' my files/assets that are attached to the release look like the following:

```bash
deploy-1.0.1-1.x86_64.rpm

deploy_1.0.1_aarch64.dmg

deploy_1.0.1_amd64.AppImage

deploy_1.0.1_amd64.deb

deploy_1.0.1_x64-setup.exe

deploy_1.0.1_x64.dmg

deploy_1.0.1_x64_en-US.msi

deploy_aarch64.app.tar.gz

deploy_x64.app.tar.gz

Source code
(zip)

Source code
(tar.gz)
```

The first inconsistency issues I have is with the assets `deploy-1.0.1-1.x86_64.rpm` , `deploy_1.0.1_x64-setup.exe`, and `deploy_1.0.1_x64_en-US.msi` which have '-' as separators instead of '_' like the rest.

The second inconsistency issue I have is with the last two assets `deploy_aarch64.app.tar.gz` and `deploy_x64.app.tar.gz` which do not have the version number in the naming. They should look like this instead: `deploy_1.0.1_aarch64.app.tar.gz` and `deploy_1.0.1_x64.app.tar.gz`

I am not sure if the last two files aren't getting the version added because of the 
```yaml
with:
          # Those targets are only used on macos runners so it's in an `if` to slightly speed up windows and linux builds.
          targets: ${{ matrix.platform == 'macos-latest' && 'aarch64-apple-darwin,x86_64-apple-darwin' || '' }}
```
in the action or if there is something else internal that is going on with the action.
If there is something that I can change with the action let me know or if this is something internal with your action then I would like to request these to be changed for consistent naming conventions across all assets generated using this release action.

Here is my workflow yml file to help debug this issue.

```yaml
name: '🚀 Publish'

on:
  workflow_dispatch:
  push:
    branches:
      - release

# This workflow will trigger on each push to the `release` branch to create or update a GitHub release, build your app, and upload the artifacts to the release.

jobs:
  publish-tauri:
    permissions:
      contents: write
    strategy:
      fail-fast: false
      matrix:
        include:
          - platform: 'macos-latest' # for Arm based macs (M1 and above).
            args: '--target aarch64-apple-darwin'
          - platform: 'macos-latest' # for Intel based macs.
            args: '--target x86_64-apple-darwin'
          - platform: 'ubuntu-22.04' # for Tauri v1 you could replace this with ubuntu-20.04.
            args: ''
          - platform: 'windows-latest'
            args: ''

    runs-on: ${{ matrix.platform }}
    # increase this 15 minutes to something longer if your project starts taking longer to build. 
    # This is just an inital fail safe so your build doesn't use up to 6 hours of time on your 2,000/month billable actions minutes.
    timeout-minutes: 15 # Set a reasonable timeout, shorter than GitHub's default 6 hours
    steps:
      - uses: actions/checkout@v4

      - name: 🛠️ Set up Node.js
        uses: actions/setup-node@v4
        with:
          node-version: lts/*
      
      - name: 🛠️ Set up Bun.sh
        uses: oven-sh/setup-bun@v2
        with:
          bun-version: latest

      - name: 🔧 Install 🦀 Rust 🟢✨ Stable
        uses: dtolnay/rust-toolchain@stable
        with:
          # Those targets are only used on macos runners so it's in an `if` to slightly speed up windows and linux builds.
          targets: ${{ matrix.platform == 'macos-latest' && 'aarch64-apple-darwin,x86_64-apple-darwin' || '' }}
      
      - name: Rust cache
        uses: swatinem/rust-cache@v2
        with:
          workspaces: './src-tauri -> target'

      - name: 🔧 Install 📦 Dependencies (🐧Ubuntu only)
        if: matrix.platform == 'ubuntu-22.04' # This must match the platform value defined above.
        run: |
          sudo apt-get update
          sudo apt-get install -y libwebkit2gtk-4.0-dev libwebkit2gtk-4.1-dev libappindicator3-dev librsvg2-dev patchelf
        # webkitgtk 4.0 is for Tauri v1 - webkitgtk 4.1 is for Tauri v2.
        # You can remove the one that doesn't apply to your app to speed up the workflow a bit.

      - name: 🔧 Install 💻 Frontend 📦 Dependencies
        # change this to npm, pnpm or bun depending on which one you use.  
        run: bun install --frozen-lockfile # current equivalent of 'npm ci' for bun (as of 07/29/2024 see/subscribe to this issue: https://github.com/oven-sh/bun/issues/5283 for updates).
        # Info: For reproducible installs, use --frozen-lockfile . This will install the exact versions of each package specified in the lockfile. 
        
      - uses: tauri-apps/tauri-action@v0
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        with:
          tagName: v__VERSION__ # the action automatically replaces \_\_VERSION\_\_ with the app version.
          releaseName: 'v__VERSION__'
          #releaseBody: '👀 See the 📂 assets to ⬇️ download this 🆕 version and 🔧 install.'
          releaseDraft: false
          prerelease: false
          args: ${{ matrix.args }}

```

lucasfernog:
I'm still not sure if we want to keep the release tag in the RPM file name, but I'll keep it for now.
Mainly wanted to raise awareness on this issue first.
I like the consistency of this, but I'm also not really a fan of using `-` instead of `_` as separator.

Note that `-` is also the separator used on the language string currently appended to the MSI file name. I think I'll just change that to use `_` instead.

### DIFF
--- a/tooling/bundler/src/bundle.rs
+++ b/tooling/bundler/src/bundle.rs
@@ -249,3 +249,21 @@ pub fn check_icons(settings: &Settings) -> crate::Result<bool> {
     Ok(true)
   }
 }
+
+pub(crate) fn bundle_name(settings: &Settings, ext: &str) -> String {
+  format!(
+    "{}-{}-{}.{ext}",
+    settings.product_name(),
+    settings.version_string(),
+    settings.binary_arch()
+  )
+}
+
+pub(crate) fn bundle_name_with_suffix(settings: &Settings, suffix: &str, ext: &str) -> String {
+  format!(
+    "{}-{}-{}-{suffix}.{ext}",
+    settings.product_name(),
+    settings.version_string(),
+    settings.binary_arch()
+  )
+}
--- a/tooling/bundler/src/bundle.rs
+++ b/tooling/bundler/src/bundle.rs
@@ -260,6 +260,7 @@ pub(crate) fn bundle_name(settings: &Settings, ext: &str) -> String {
   )
 }
 
+#[inline]
 pub(crate) fn bundle_name_with_suffix(settings: &Settings, suffix: &str, ext: &str) -> String {
   format!(
     "{}-{}-{}-{suffix}.{ext}",
--- a/tooling/bundler/src/bundle.rs
+++ b/tooling/bundler/src/bundle.rs
@@ -250,6 +250,7 @@ pub fn check_icons(settings: &Settings) -> crate::Result<bool> {
   }
 }
 
+#[inline]
 pub(crate) fn bundle_name(settings: &Settings, ext: &str) -> String {
   format!(
     "{}-{}-{}.{ext}",
--- a/tooling/bundler/src/bundle/linux/rpm.rs
+++ b/tooling/bundler/src/bundle/linux/rpm.rs
@@ -18,8 +18,6 @@ use super::freedesktop;
 /// Bundles the project.
 /// Returns a vector of PathBuf that shows where the RPM was created.
 pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
-  let product_name = settings.product_name();
-  let version = settings.version_string();
   let release = settings.rpm().release.as_str();
   let epoch = settings.rpm().epoch;
   let arch = match settings.binary_arch() {
@@ -30,27 +28,26 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
 
   let summary = settings.short_description().trim();
 
-  let package_base_name = format!("{product_name}-{version}-{release}.{arch}");
-  let package_name = format!("{package_base_name}.rpm");
+  let rpm_name = crate::bundle::bundle_name_with_suffix(settings, release, "dmg");
 
   let base_dir = settings.project_out_directory().join("bundle/rpm");
-  let package_dir = base_dir.join(&package_base_name);
+  let package_dir = base_dir.join(Path::new(&rpm_name).file_stem().unwrap());
   if package_dir.exists() {
-    fs::remove_dir_all(&package_dir)
-      .with_context(|| format!("Failed to remove old {package_base_name}"))?;
+    fs::remove_dir_all(&package_dir).with_context(|| format!("Failed to remove old {rpm_name}"))?;
   }
   fs::create_dir_all(&package_dir)?;
-  let package_path = base_dir.join(&package_name);
+  let package_path = base_dir.join(&rpm_name);
 
-  log::info!(action = "Bundling"; "{} ({})", package_name, package_path.display());
+  log::info!(action = "Bundling"; "{} ({})", rpm_name, package_path.display());
 
   let license = settings.license().unwrap_or_default();
   let name = heck::AsKebabCase(settings.product_name()).to_string();
-  let mut builder = rpm::PackageBuilder::new(&name, version, &license, arch, summary)
-    .epoch(epoch)
-    .release(release)
-    // This matches .deb compression. On a 240MB source binary the bundle will be 100KB larger than rpm's default while reducing build times by ~25%.
-    .compression(rpm::CompressionWithLevel::Gzip(6));
+  let mut builder =
+    rpm::PackageBuilder::new(&name, settings.version_string(), &license, arch, summary)
+      .epoch(epoch)
+      .release(release)
+      // This matches .deb compression. On a 240MB source binary the bundle will be 100KB larger than rpm's default while reducing build times by ~25%.
+      .compression(rpm::CompressionWithLevel::Gzip(6));
 
   if let Some(description) = settings.long_description() {
     builder = builder.description(description);
--- a/tooling/bundler/src/bundle/linux/rpm.rs
+++ b/tooling/bundler/src/bundle/linux/rpm.rs
@@ -28,7 +28,7 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
 
   let summary = settings.short_description().trim();
 
-  let rpm_name = crate::bundle::bundle_name_with_suffix(settings, release, "dmg");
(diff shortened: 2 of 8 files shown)


tooling/bundler/src/bundle/windows/msi/wix.rs
```diff
@@ -176,37 +176,18 @@ fn copy_icon(settings: &Settings, filename: &str, path: &Path) -> crate::Result<
...
-
-  let package_base_name = format!(
-    "{}_{}_{}_{}",
-    settings.product_name(),
-    version,
-    arch,
-    language,
-  );
+  let msi_name =
+    crate::bundle::bundle_name_with_suffix(settings, &language.replace('-', "_"), "msi");
```
amrbashir: we should probably revert this change, `-` (hyphen) is the separator in language tags per the IETF standard https://en.wikipedia.org/wiki/IETF_language_tag#Syntax_of_language_tags

tooling/bundler/src/bundle/windows/msi/wix.rs
```diff
@@ -377,6 +348,18 @@ fn run_light(
   Ok(())
 }
 
+fn wix_arch(settings: &Settings) -> crate::Result<&'static str> {
```
amrbashir, line 351:
```suggestion
#[inline]
fn wix_arch(settings: &Settings) -> crate::Result<&'static str> {
```

tooling/bundler/src/bundle.rs
```diff
@@ -249,3 +249,21 @@ pub fn check_icons(settings: &Settings) -> crate::Result<bool> {
     Ok(true)
   }
 }
+
+pub(crate) fn bundle_name(settings: &Settings, ext: &str) -> String {
```
amrbashir, line 254:
```suggestion
#[inline]
pub(crate) fn bundle_name(settings: &Settings, ext: &str) -> String {
```

tooling/bundler/src/bundle.rs
```diff
@@ -249,3 +249,21 @@ pub fn check_icons(settings: &Settings) -> crate::Result<bool> {
...
+  format!(
+    "{}-{}-{}.{ext}",
+    settings.product_name(),
+    settings.version_string(),
+    settings.binary_arch()
+  )
+}
+
+pub(crate) fn bundle_name_with_suffix(settings: &Settings, suffix: &str, ext: &str) -> String {
```
amrbashir, line 264:
```suggestion
#[inline]
pub(crate) fn bundle_name_with_suffix(settings: &Settings, suffix: &str, ext: &str) -> String {
```

amrbashir (changes requested):

tooling/bundler/src/bundle/windows/msi/wix.rs
lucasfernog: well that's true, but with this change it's also the separator for the parts in the bundle file name.. I'd prefer using `_` instead, but some people prefer `-`

amrbashir: there will always people wanting it the other way around but I don't see an issue with `App-1.2.0-x64-en-US.exe`, seems better than `App-1.2.0-x64-en_US.exe`

amrbashir (dismissed):

lucasfernog:
let's at least hold on merging this one, must be communicated (ideally released as the last RC or perhaps even the first 2.0 stable release)

FabianLars:
This is missing explicit Arm + AppImage support - noticed because of https://github.com/tauri-apps/tauri/pull/10619 (arm64 currently works, but i apparently didn't test arm 32bit which doesn't work).

Expanding this condition for arm->armhf should be enough i think https://github.com/tauri-apps/tauri/blob/dev/tooling/bundler/src/bundle/linux/templates/appimage#L14

tooling/bundler/src/bundle/linux/rpm.rs
```diff
@@ -30,27 +28,26 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
 
   let summary = settings.short_description().trim();
 
-  let package_base_name = format!("{product_name}-{version}-{release}.{arch}");
-  let package_name = format!("{package_base_name}.rpm");
+  let rpm_name = crate::bundle::bundle_name_with_suffix(settings, release, "dmg");
```
FabianLars:
```suggestion
  let rpm_name = crate::bundle::bundle_name_with_suffix(settings, release, "rpm");
```

tooling/bundler/src/bundle/windows/msi/wix.rs
xuchaoqian: > there will always people wanting it the other way around but I don't see an issue with `App-1.2.0-x64-en-US.exe`, seems better than `App-1.2.0-x64-en_US.exe`

I think this is about consistency rather than standards. The hyphen is a separator between different parts of the filename, while the underscore is a separator between words within a part, such as Snow_White-v1.0.1_beta1-x86_64-en_US.deb. Under the consistency naming convention, it is easy to parse the various standard-compliant parts.

xuchaoqian: > there will always people wanting it the other way around but I don't see an issue with `App-1.2.0-x64-en-US.exe`, seems better than `App-1.2.0-x64-en_US.exe`

I think this is about consistency rather than standards. The hyphen(-) is a separator between different parts of the filename, while the underscore(_) is a separator between words within a part, such as Snow_White-v1.0.1_beta1-x86_64-en_US.deb. Under the consistent naming convention, it is easy to parse the various standard-compliant parts.

lucasfernog:
closing this one as it's too late for such a huge change that impacts everyone shipping Tauri apps.. see you in v3

Issue #10583, opened 2024-08-12
[feat] manually specify signtool.exe used by tauri


anatawa12:
### Describe the problem

With eSigner CKA by SSL.com, it may fails to sign exe with x64 version of signtool.exe.

### Describe the solution you'd like

Therefore, I want to manually specify path to signtool.exe with environment variable.

### Alternatives considered

Use custom sign command but it requires signing tool to not have whitespace, which is not true (program Files contains whitespace) for signtool.
In addition, location of signtool is relatively PC-specific so I think it's not good to specify in tauri.conf.json.

FabianLars:
> Use custom sign command but it requires signing tool to not have whitespace, which is not true (program Files contains whitespace) for signtool.

Escaped quotes should work, so `"signCommand": "\"C:/path/to/Sign Tool.exe\" some commands %1"`

>  In addition, location of signtool is relatively PC-specific so I think it's not good to specify in tauri.conf.json.

Well, an option to specify which signtool exe to use would have the same problem.

---

I don't think is something we will explicitly support and instead leave that to the existing signCommand config. Please reach out again if you can't make it work (for example if the escaped quotes don't work)

anatawa12:
>>  Use custom sign command but it requires signing tool to not have whitespace, which is not true (program Files contains whitespace) for signtool.
> 
> Escaped quotes should work, so "signCommand": "\"C:/path/to/Sign Tool.exe\" some commands %1"

https://github.com/tauri-apps/tauri/blob/20be073bc41f13bb72c442a4d789d481be6bc951/tooling/bundler/src/bundle/windows/sign.rs#L138

escaped quote looks not working

> Well, an option to specify which signtool exe to use would have the same problem.

with envronment variable , there's no such a problem.

Copied from issue:
> Therefore, I want to manually specify path to signtool.exe with **environment variable**.

PR #10589, opened 2024-08-13
feat: Support '"' quoting for custom signing tool

anatawa12:
Add support for '"' quoting in custom signing tool

https://github.com/tauri-apps/tauri/issues/10583#issuecomment-2285266577

### DIFF
--- a/tooling/bundler/src/bundle/windows/sign.rs
+++ b/tooling/bundler/src/bundle/windows/sign.rs
@@ -337,17 +337,44 @@ mod tests {
   fn test_parse_custom_sign_command() {
     // basic cases
     assert!(parse_custom_sign_command("").unwrap().is_empty());
-    assert_eq!(parse_custom_sign_command("foo").unwrap(), vec!["foo".to_string()]);
-    assert_eq!(parse_custom_sign_command("foo bar").unwrap(), vec!["foo".to_string(), "bar".to_string()]);
-    assert_eq!(parse_custom_sign_command("foo bar baz").unwrap(), vec!["foo".to_string(), "bar".to_string(), "baz".to_string()]);
-    assert_eq!(parse_custom_sign_command("foo \"bar baz\"").unwrap(), vec!["foo".to_string(), "bar baz".to_string()]);
-    assert_eq!(parse_custom_sign_command("foo \"bar baz\" qux").unwrap(), vec!["foo".to_string(), "bar baz".to_string(), "qux".to_string()]);
-    assert_eq!(parse_custom_sign_command("foo \"bar baz\"\"qux\"").unwrap(), vec!["foo".to_string(), "bar baz\"qux".to_string()]);
-    assert_eq!(parse_custom_sign_command("\"foo bar\" baz").unwrap(), vec!["foo bar".to_string(), "baz".to_string()]);
+    assert_eq!(
+      parse_custom_sign_command("foo").unwrap(),
+      vec!["foo".to_string()]
+    );
+    assert_eq!(
+      parse_custom_sign_command("foo bar").unwrap(),
+      vec!["foo".to_string(), "bar".to_string()]
+    );
+    assert_eq!(
+      parse_custom_sign_command("foo bar baz").unwrap(),
+      vec!["foo".to_string(), "bar".to_string(), "baz".to_string()]
+    );
+    assert_eq!(
+      parse_custom_sign_command("foo \"bar baz\"").unwrap(),
+      vec!["foo".to_string(), "bar baz".to_string()]
+    );
+    assert_eq!(
+      parse_custom_sign_command("foo \"bar baz\" qux").unwrap(),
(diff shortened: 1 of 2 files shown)


tooling/bundler/src/bundle/windows/sign.rs
```diff
@@ -132,10 +132,95 @@ pub fn verify(path: &Path) -> crate::Result<bool> {
...
+  let mut state = State::Idle;
+
+  for c in command.chars() {
+    match (state, c) {
+      (State::Idle, c) if c.is_ascii_whitespace() => {
+        // skip whitespace
+      }
+      (State::Idle, '"') => {
+        state = State::Quoted;
+      }
```
anatawa12, line 163: It might be good to reserve some symbols for future extension

anatawa12:
I've fixed format error. sorry

lucasfernog (approved):
looks good to me, anything against this change @amrbashir ?

tooling/bundler/src/bundle/windows/sign.rs
```diff
@@ -132,10 +132,95 @@ pub fn verify(path: &Path) -> crate::Result<bool> {
...
+        state = State::Unquoted;
+      }
+
+      (State::Unquoted, c) if c.is_ascii_whitespace() => {
+        result.push(buffer.clone());
+        buffer.clear();
+        state = State::Idle;
+      }
+      (State::Unquoted, '"') => {
+        return Err(anyhow::anyhow!("custom signing command contains an unclosed quote").into());
```
amrbashir, line 175: We shouldn't error here imo, and just continue parsing

tooling/bundler/src/bundle/windows/sign.rs
```diff
@@ -132,10 +132,95 @@ pub fn verify(path: &Path) -> crate::Result<bool> {
...
+        buffer.push('"');
+        state = State::Quoted;
+      }
+      (State::QuotedJustAfterQuote, c) if c.is_ascii_whitespace() => {
+        result.push(buffer.clone());
+        buffer.clear();
+        state = State::Idle;
+      }
+      (State::QuotedJustAfterQuote, _) => {
+        return Err(anyhow::anyhow!("custom signing command contains an unclosed quote").into());
```
amrbashir, line 198: same here

tooling/bundler/src/bundle/windows/sign.rs
```diff
@@ -132,10 +132,95 @@ pub fn verify(path: &Path) -> crate::Result<bool> {
...
+    }
+  }
+
+  match state {
+    State::Idle => {}
+    State::Unquoted => {
+      result.push(buffer);
+    }
+    State::Quoted => {
+      return Err(anyhow::anyhow!("custom signing command contains an unclosed quote").into());
```
amrbashir, line 209: and here

amrbashir:
I wonder if it is better to change the command into an object rather than introduce custom parsing logic
```json
{
	"customSignCommand": {
		"cmd": "program with space.exe",
		"args": ["list", "of", "args"],
	}
}
```

tooling/bundler/src/bundle/windows/sign.rs
anatawa12: I think it's a little unclear that `foo"bar baz` should be treated as `foo` `bar baz`, `foo"bar` `baz`, or `foo"bar baz` so I made this error. Which should I choose?

anatawa12:
I think it's better to use object form but I have no Idea how should I implement that since crate for bundler and config is diffdrent crate Shuld I create struct on both config and bundler?

tooling/bundler/src/bundle/windows/sign.rs
amrbashir: it should be `foo"bar` `baz`

amrbashir:
> I think it's better to use object form but I have no Idea how should I implement that since crate for bundler and config is diffdrent crate Shuld I create struct on both config and bundler?

yup, two configs, there is similar objects you can take inspiration from like `NsisConfig` in config and `NsisSettings` in bundler.

tooling/bundler/src/bundle/windows/sign.rs
amrbashir: but if we go with the object solution (which we should), this doesn't matter

anatawa12: I mean your decision is not clear and reasonable enough I think.
For example, with command prompt (`cmd.exe`), `foo"bar baz` become `foobar baz`.


(Sorry I forget to list this up in the previous example list. `foo"bar baz` is mistake of `foobar baz`)


Anyway I made PR for object solution so this discussion might be useless.

#10634

amrbashir: thanks, will close this PR then and move discussion there

PR #10634, opened 2024-08-15
feat: Custom sign command with object notation

anatawa12:
https://github.com/tauri-apps/tauri/pull/10589#issuecomment-2287668387

### DIFF
--- a/.changes/bundler-object-custom-sign-command.md
+++ b/.changes/bundler-object-custom-sign-command.md
@@ -1,6 +1,6 @@
 ---
-"tauri-bundler": patch:breaking
-"tauri-utils": patch:breaking
+"tauri-bundler": patch:feat
+"tauri-utils": patch:feat
 ---
 
-Custom sign command for windows executables is now with object instead of string to support whitespaces in the command path and arguments.
+Custom sign command with object notation for whitespaces in the command path and arguments.
--- a/.changes/bundler-object-custom-sign-command.md
+++ b/.changes/bundler-object-custom-sign-command.md
@@ -1,0 +1,6 @@
+---
+"tauri-bundler": patch:breaking
+"tauri-utils": patch:breaking
+---
+
+Custom sign command for windows executables is now with object instead of string to support whitespaces in the command path and arguments.
--- a/.changes/bundler-object-custom-sign-command.md
+++ b/.changes/bundler-object-custom-sign-command.md
@@ -2,7 +2,7 @@
 "tauri-bundler": patch:feat
 "tauri-utils": patch:feat
 "@tauri-apps/cli": patch:feat
-"tauri-utils": patch:feat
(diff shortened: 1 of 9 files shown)


anatawa12:
I ran cargo fmt at repository root again but no diff are generated. I feel it's weird to fail fmt check

core/tauri-utils/src/config.rs
```diff
@@ -891,6 +891,17 @@ impl Default for WebviewInstallMode {
   }
 }
 
+/// Custom Signing Command configuration.
+#[derive(Debug, PartialEq, Eq, Clone, Deserialize, Serialize)]
```
anatawa12, line 895:
```suggestion
#[derive(Debug, PartialEq, Eq, Clone, Deserialize, Serialize)]
#[cfg_attr(feature = "schema", derive(JsonSchema))]
#[serde(rename_all = "camelCase", deny_unknown_fields)]
```

.changes/bundler-object-custom-sign-command.md
```diff
@@ -0,0 +1,6 @@
+---
+"tauri-bundler": patch:breaking
+"tauri-utils": patch:breaking
+---
+
+Custom sign command for windows executables is now with object instead of string to support whitespaces in the command path and arguments.
```
amrbashir: I wonder if it is worth supporting both formats, string is more concise for simple use cases and object is better for white space paths and such

amrbashir: cc @lucasfernog

lucasfernog: supporting both formats is what we do :) use an enum with serde(untagged)

amrbashir: @anatawa12 sorry for the inconvenience, could you add support for both? and add a note in docs that users should use the object format if their binary or arguments have spaces.

anatawa12: OK.
Should support quote in string notation for whitespaces? or use old simple parser with split(' ')?

amrbashir: use old simple parser, and just add a note in docs, thanks

anatawa12:
> I ran cargo fmt at repository root again but no diff are generated. I feel it's weird to fail fmt check

I found that I have to run `cargo fmt` inside each folder in `tooling/` folder since they are not part of root workspace

.changes/bundler-object-custom-sign-command.md
anatawa12: I added string form support in 53e285a1e9b248fbd6c821a8b975c028aedee302.

Since string format is only for simpler config file, I added parsing logic in cli module and bundler module have single object form as input.

lucasfernog (approved):

PR #10641, opened 2024-08-15
refactor(core): serialize response once, closes #5641

### LINKED ISSUE #5641: [feat] improve performances of invoke by discarding use of serde_json::Value
### Describe the problem

I analysed the code of tauri because I wanted to improve the performances of the `invoke` function (on my project, it takes 30s to read 500Mb of data with a call to `invoke` while simply reading and writing it in JSON takes only 5s).

From what I understood by reading the code, here is what happen when we call the `invoke` function:

1. The Rust function is called [tauri-macros/src/command/wrapper.rs:158]
2. Its output is serialized as a `serde_json::Value`  [tauri/src/command.rs:197]
3. The `serde_json::Value` is serialized as a `serde_json::Value` (see #5639) [tauri/src/hooks.rs:196]
4. The `serde_json::Value` is serialized as a Json String [tauri/src/api/ipc.rs:69]
5. The Json String is converted to `RawValue`, potentially copying all the String in memory [tauri/src/api/ipc.rs:70]
6. The `RawValue` is converted to an escaped String [tauri/src/api/ipc.rs:85]
7. The escaped String is put in a larger JS callback function String [tauri/src/api/ipc.rs:174]
8. The String is evaluated in JS [tauri/src/window.rs:1289]

I think that three serialization is overkill, the serialization in 3. can be avoided (see #5639) and probably the serialization in 4. too (also it requires a bit more work).

String allocation might be optimized but this is not what takes the longest running time in my opinion.

Instead of serializing the result of the invoked function to a `serde_json::Value`, it should be serialized directly to a Json String.

### Alternatives considered

_No response_

### Additional context

Issue #4026 discusses memory leaks when using the `invoke` command. My tests also showed memory leaks. These two problems (performance issues and memory leaks) are related.

lucasfernog:
This change impacts both the custom protocol and the postMessage based IPC implementations. Basically it changes the whole IPC mechanism to work on raw JSON strings so we do not need to serialize a serde_json::Value after serializing to it from a user-provided type.

i benchmarked this with a 150MB file response (returning Vec<u8> instead of tauri::ipc::Response since the latter does not serialize at all) and it went from 29s to 23s (custom protocol) and from 54s to 48s (post message) on macOS.


amrbashir (dismissed):

amrbashir:
This a good improvement, nice job

amrbashir (approved):

PR #10648, opened 2024-08-16
fix(android): avoid rebuilds if nothing changed

Flakebi:
Unconditionally overwriting files where the build reruns if they changed leads to rebuilds every time.
Only overwrite a file if its content is different to not rebuild in such a case.


lucasfernog (dismissed):
nice catch! I did face this issue before but never had the time to actually dig into this, thank you!

core/tauri-utils/src/io.rs
```diff
@@ -4,7 +4,16 @@
 
 //! IO helpers.
 
-use std::io::BufRead;
+use std::{fs::read_to_string, io::BufRead, path::Path};
+
+/// Write to the given file if the content is different.
+pub fn write_if_changed(content: &str, path: &Path) -> std::io::Result<()> {
+  if content != read_to_string(path).unwrap_or_default() {
```
Flakebi: This does a probably unexpected thing if `content == ""`. In that case the file is not created if it doesn’t already exist (due to the `or_default`).
I think the solution from https://github.com/tauri-apps/wry/pull/1342 is better
```rust
if read_to_string(path).map_or(true, |c| c != content) {
  // …
}
```

lucasfernog: right right, let's fix that

lucasfernog: oh actually we already had a function with this functionality, i've pushed a change to use that one instead

lucasfernog (approved):

PR #10657, opened 2024-08-17
skip pack in publish, use check fetch

jbolda:
## Motivation

We don't need the `cargo package` or `npm pack` uploaded to the release. It likely goes unnoticed and simply increases CI time. Also, swap to the built-in fetch calls.


.changes/config.json
```diff
@@ -14,10 +14,13 @@
   "pkgManagers": {
     "rust": {
       "version": true,
-      "getPublishedVersion": "node ../../.scripts/covector/package-latest-version.js cargo ${ pkgFile.pkg.package.name } ${ pkgFile.pkg.package.version }",
+      "getPublishedVersion": {
+        "use": "fetch:check",
```
lucasfernog, line 18: I assume the built in check function only checks if the latest release is different from the current one right?
the reason we added a custom one here is to support hotfixes:
- let's say the latest release is 1.2.0
- but we open a separate branch on 1.1.0 and want to publish a hotfix for it as 1.1.1 - we want to publish that crate alone, but not others, so our custom check performs a "semver" check by matching against minors only

jbolda: It checks if _that_ specific version is published. From what you described, it seems like that still works with your workflow?

lucasfernog (approved):

PR #10658, opened 2024-08-17
skip prepublish on api

jbolda:
## Motivation

Trying to reduce the publish config. I removed the prepublish step which it can pull from the generic javascript manager docs.

### DIFF
--- a/.changes/config.json
+++ b/.changes/config.json
@@ -134,38 +134,6 @@
     "@tauri-apps/api": {
       "path": "./tooling/api",
       "manager": "javascript",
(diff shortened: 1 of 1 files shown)


lucasfernog (changes requested):
i think we agreed this isn't that simple right? the api package publish script is a lot more involved - and would require changes to work with a direct `npm publish`

jbolda:
@lucasfernog correct. I left the publish step because of that.

lucasfernog (approved):

PR #10700, opened 2024-08-20
fix(core): allow hyphens and underscores on identifiers, closes #9707

### LINKED ISSUE #9707: [bug] Inconsistent handling of app identifiers containing hyphens when building for Android
### Describe the bug

Like most app development frameworks, Tauri uses reverse domain notation for the app identifiers.

Tauri only allows letters (`a-z`, `A-Z`), digits (`0-9`) and hyphens (`-`) in app identifiers. However, Android only allows underscores (`_`) in app identifiers, not hyphens; iOS, on the other hand, does allow hyphens, but no underscores.

Right now, this means that running `cargo tauri android init` panics if the app identifier contains hyphens or alternatively only contains of two parts (I think the latter is unintended, anyway).

To illustrate, let's assume I have an app identifier like `tech.almost-senseless.jack-point`. Trying to build this app for Android fails with a panic due to the hyphen in `almost-senseless`. Renaming it to `tech.almostsenseless.jack-point` or `tech.almost-senseless` works fine, though: In the first case, the generated Android package name and identifier would be `tech.almostsenseless.jack_point`, and in the latter case it would be `tech.jack_point`.

Evidently, the last part of the app identifier gets silently dropped and replaced with the product name, and in the product name, hyphens get replaced with underscores.

This has the side effect that running `cargo tauri android init` when the app identifier is `com.tauri.dev` works, by the way, because it becomes `com.tauri.jack_point` - I think that's not intended that way, is it?

So my problem boils down to that

- Having hyphens in your app identifier shouldn't lead to a panic; rather, the hyphens should get replaced with underscores in the generated Android project.
- The handling of the identifier should be more consistent, with the last part not getting silently dropped and replaced with the product name.

### Reproduction

- Create a Tauri project from any template, for instance using `cargo create-tauri-app --beta`. Please make sure mobile support was enabled.
- Open `tauri.conf.json` and change the identifier to `com.some-domain.some-app`.
- Run `cargo tauri android init`.

### Expected behavior

The command should complete successfully, generating an Android project with the identifier `com.some_domain.some_app`.

### Full `tauri info` output

```text

[✔] Environment
    - OS: Windows 10.0.22631 X64
    ✔ WebView2: 124.0.2478.80
    ✔ rustc: 1.80.0-nightly (e82c861d7 2024-05-04)
    ✔ cargo: 1.80.0-nightly (05364cb2f 2024-05-03)
    ✔ Rust toolchain: nightly-x86_64-pc-windows-msvc (environment override by RUSTUP_TOOLCHAIN)
    - node: 20.11.0
    - npm: 10.5.0

[-] Packages
    - tauri [RUST]: 2.0.0-beta.17
    - tauri-build [RUST]: 2.0.0-beta.13
    - wry [RUST]: 0.39.3
    - tao [RUST]: 0.27.1
    - tauri-cli [RUST]: 2.0.0-beta.16
    - @tauri-apps/cli [NPM]: 2.0.0-beta.16

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
```
```


### Stack trace

```text
thread 'main' panicked at C:\Users\timjb\.cargo\registry\src\index.crates.io-6f17d22bba15001f\tauri-cli-2.0.0-beta.16\src\mobile\mod.rs:322:6:
called `Result::unwrap()` on an `Err` value: DomainInvalid { domain: "almost-senseless.tech", cause: NotAsciiAlphanumeric { bad_chars: ['-'] } }
```

### Additional context

_No response_

lucasfernog:


lucasfernog:
needs a wry fix i think

lucasfernog:
Needs https://github.com/tauri-apps/tao/pull/964

core/tauri-build/src/lib.rs
```diff
@@ -478,7 +478,7 @@ pub fn try_build(attributes: Attributes) -> Result<()> {
     if i == last {
       println!("cargo:rustc-env=TAURI_ANDROID_PACKAGE_NAME_APP_NAME={w}");
     } else {
-      android_package_prefix.push_str(w);
+      android_package_prefix.push_str(&w.replace(['_', '-'], "_1"));
```
pewsheen: `TAURI_ANDROID_PACKAGE_NAME_APP_NAME` seemed to be replaced in https://github.com/tauri-apps/tauri/blob/762cf31a113f33cd524c8f6fa52ca47db8ea3705/core/tauri-macros/src/mobile.rs#L46-L51 before it calls `android_binding!`

Should we move that replacement to L479 as well? So that we can do it in one place.

lucasfernog: good idea, pushed

pewsheen:
> needs a wry fix i think

Looks like we don't need it? Other than this one, I think it's good to merge

lucasfernog:
> > needs a wry fix i think
> 
> Looks like we don't need it? Other than this one, I think it's good to merge

yeah it was just the tao thing

PR #10707, opened 2024-08-21
chore(deps) Update Prettier config and its version to 3.3.3 (dev)

ghost:
This pull request offers suggestions that should not affect the functionality, just modernization of code.

- update Prettier version;
- avoid disaligned Prettier versions in the root package.json and in the subfolders package.json by changing from ^3.3.3 to 3.3.3;
- update Prettier config from its old format (.prettierrc) to prettier.config.cjs (.cjs because the root package.json does not have an explicit "type": "module").
- from a few versions onwards, Prettier does not require anymore to specify arguments in Scripts about where .prettierignore and .prettierrc are located. Prettier finds these configs automatically, when they are in the expected positions. So it was removed to simplify and make maintenance easier.
- add --cache to Scripts. Prettier runs faster.

---

See https://github.com/tauri-apps/tauri/issues/10719

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,8 +1,0 @@
-{
-  "singleQuote": true,
-  "semi": false,
-  "trailingComma": "none",
-  "overrides": [
-    { "files": [".changes/**.md"], "options": { "singleQuote": false } }
-  ]
-}
--- a/package.json
+++ b/package.json
@@ -11,8 +11,8 @@
     "url": "https://github.com/tauri-apps/tauri.git"
   },
   "scripts": {
-    "format": "prettier --write . --config .prettierrc --ignore-path .prettierignore",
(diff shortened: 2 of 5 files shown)


amrbashir:
> * update Prettier config from its old format (.prettierrc) to prettier.config.cjs. (.cjs is the file format because that is how the root package.json is being recognized by the editor, given this root package.json does not have an explicit "type": "module");

Can we revert this? `.prettierrc` is still supported and I don't see anywhere it being removed or deprecated, `.prettierrc.json` works too

amrbashir (changes requested):
> * update Prettier config from its old format (.prettierrc) to prettier.config.cjs. (.cjs is the file format because that is how the root package.json is being recognized by the editor, given this root package.json does not have an explicit "type": "module");

Can we revert this? `.prettierrc` is still supported and I don't see anywhere it being removed or deprecated, `.prettierrc.json` works too

PR #10743, opened 2024-08-22
chore: cleanup and simplify examples

amrbashir:
- Updated `.gitignore` file
- Simplified and cleanedup examples
- Removed examples that serve almost zero value:
  - `navigation`
  - `parent-window`
  - `web`
  - `workspace`
  - `plugins/tauri-plugin-sample`: the plugin example in `api` example is enough


examples/api/src-tauri/Cargo.toml
```diff
@@ -37,14 +37,3 @@ features = [
 [dev-dependencies.tauri]
 path = "../../../core/tauri"
 features = ["test"]
-
-[features]
-prod = ["tauri/custom-protocol"]
```
lucasfernog, line 42: hey i liked this feature.. a super convenient way to run the app in production mode

examples/run-iteration/main.rs
```diff
@@ -1,26 +0,0 @@
...
-
-fn main() {
-  let mut app = tauri::Builder::default()
-    .build(tauri::generate_context!(
-      "../../examples/run-iteration/tauri.conf.json"
-    ))
-    .expect("error while building tauri application");
-
-  loop {
-    app.run_iteration(|_app, _event| {
```
lucasfernog: i think we should keep this example.. there's no other usage of this api

examples/api/src-tauri/Cargo.toml
amrbashir: I didn't find it being used somewhere so just removed it, let's bring it back

examples/run-iteration/main.rs
amrbashir: yeah I was hesitant too, let's bring that back as well

amrbashir:
so linking the api example manifest using `tauri-winres` results in `STATUS_ENTRYPOINT_NOT_FOUND` but if directly embedding it using `cargo:rustc-link-arg=/MANIFEST:EMBED` has no issues. I hate windows

lucasfernog:
> so linking the api example manifest using `tauri-winres` results in `STATUS_ENTRYPOINT_NOT_FOUND` but if directly embedding it using `cargo:rustc-link-arg=/MANIFEST:EMBED` has no issues. I hate windows

I guess the `__TAURI_WORKSPACE__` workaround doesn't help us here right? since it's a binary that must have that one

amrbashir:
The `__TAURI_WORKSPACE__` workaround also uses `cargo:rustc-link-arg=/MANIFEST:EMBED` and that's what works

amrbashir:
I will spend sometime to test and see what's the different and why this happens.

lucasfernog (approved):

PR #10751, opened 2024-08-23
fix(cli): ensure gradlew is executable and does not use CRLF

lucasfernog:
This is required due to a limitation of our template engine, which cannot preserve file permissions.

When you generate the Android project on Android and commit it, it will fail to run in CI (if you are running on UNIX) because it will lack the executable permissions. Plus if it uses CRLF it will also fail to execute (`/usr/bin/env: ‘sh\r’: No such file or directory`).

### DIFF
--- a/tooling/cli/src/mobile/mod.rs
+++ b/tooling/cli/src/mobile/mod.rs
@@ -374,9 +374,9 @@ fn ensure_gradlew(project_dir: &std::path::Path) -> Result<()> {
     let mut permissions = metadata.permissions();
     let is_executable = permissions.mode() & 0o111 != 0;
     if !is_executable {
-      permissions.set_mode(0o755);
+      permissions.set_mode(permissions.mode() | 0o111);
       std::fs::set_permissions(&gradlew_path, permissions)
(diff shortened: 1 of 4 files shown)


tooling/cli/src/mobile/mod.rs
```diff
@@ -362,6 +365,31 @@ fn ensure_init(
...
+#[cfg(unix)]
+fn ensure_gradlew(project_dir: &std::path::Path) -> Result<()> {
+  use std::os::unix::fs::PermissionsExt;
+
+  let gradlew_path = project_dir.join("gradlew");
+  if let Ok(metadata) = gradlew_path.metadata() {
+    let mut permissions = metadata.permissions();
+    let is_executable = permissions.mode() & 0o111 != 0;
+    if !is_executable {
+      permissions.set_mode(0o755);
```
amrbashir: should we preserve previous permissions and only add `0o111` for execution bit? instead of always setting `0o755`

amrbashir (dismissed):

amrbashir (approved):

Issue #10765, opened 2024-08-25
[bug] resolve empty object change reject


rdlabo:
### Describe the bug

This code change reject at iOS

~~invoke.resolve([:])~~

```
invoke.resolve()
```

Browser error: `failed to deserialize response: invalid type: map, expected unit`

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 14.4.1 X64
    ✔ Xcode Command Line Tools: installed
    ✔ rustc: 1.80.1 (3f5fd8dd4 2024-08-06)
    ✔ cargo: 1.80.1 (376290515 2024-07-16)
    ✔ rustup: 1.27.1 (54dd3d00f 2024-04-24)
    ✔ Rust toolchain: stable-aarch64-apple-darwin (default)
    - node: 22.6.0
    - pnpm: 8.9.0
    - yarn: 1.17.3
    - npm: 10.8.2

[-] Packages
    - tauri [RUST]: 2.0.0-rc.3
    - tauri-build [RUST]: 2.0.0-rc.3
    - wry [RUST]: 0.42.0
    - tao [RUST]: 0.29.0
    - tauri-cli [RUST]: 2.0.0-alpha.7
    - @tauri-apps/api [NPM]: 2.0.0-rc.1
    - @tauri-apps/cli [NPM]: 2.0.0-rc.3

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:1420/
    - bundler: Rollup

[-] iOS
    - Developer Teams: ***
```

lucasfernog:
Can you share the command call? I think this is just a mismatch between the type you're sending from the plugin and what you are expecting in the frontend.

rdlabo:
Sorry! It seems it was a debugging issue on my part, due to a type mismatch in Rust.
I will close this issue.

PR #10766, opened 2024-08-25
fix(ios): plugin ios can resolve([:])

rdlabo:
support `resolve([:])` at plugin / ios

fixs https://github.com/tauri-apps/tauri/issues/10765

### DIFF
--- a/core/tauri/mobile/ios-api/Sources/Tauri/Invoke.swift
+++ b/core/tauri/mobile/ios-api/Sources/Tauri/Invoke.swift
@@ -46,6 +46,9 @@ import UIKit
   }
 
   public func resolve(_ data: JsonObject) {
+    if (data.isEmpty) {
+      resolve()
+    }
     resolve(.dictionary(data))
   }
 


core/tauri/mobile/ios-api/Sources/Tauri/Invoke.swift
```diff
@@ -46,6 +46,9 @@ import UIKit
   }
 
   public func resolve(_ data: JsonObject) {
+    if (data.isEmpty) {
+      resolve()
```
lucasfernog, line 50: this is misleading as you are resolving an empty object but the value send through the bridge is actually `null`..

rdlabo: Thanks for review. You are right. I will re-implement this so that it does not become a REJECT.

rdlabo: Thanks for review. You are right.

rdlabo:
Close this pull request once.

PR #10778, opened 2024-08-26
fix(ci): missing pnpm on check api action

Legend-Master:
Reference: https://github.com/tauri-apps/tauri/actions/runs/10552931246/job/29232497631?pr=10210


Legend-Master:
Seems like we missed a setup node action here as well, should I also add it here?

https://github.com/tauri-apps/tauri/blob/ee1179ee7c5e9deb91baaf47982efb234ca9f7b3/.github/workflows/covector-version-or-publish.yml#L26

amrbashir (approved):

PR #10786, opened 2024-08-26
feat: add `DoubleClick` variant for `TrayIconEvent`

amrbashir:


lucasfernog (approved):
should a double click event be emitted after two Click events? or should those be suppressed if it's a double click?

amrbashir:
> should a double click event be emitted after two Click events? or should those be suppressed if it's a double click?

it is not suppressed, both are emitted

PR #10798, opened 2024-08-27
fix(docs): typo in documentation of `setSize()` method in Webview API.

ishaan26:
The example for the `setSize()` method in the documentation for the Webview api wrongly imports `getCurrent` (I believe as was in tauri v1) instead of `getCurrentWebview`.


amrbashir (approved):

amrbashir:
Thank you

PR #10799, opened 2024-08-27
feat(api): permission types for mobile plugins

lucasfernog:
let's have a central place to keep the permission types

ref https://github.com/tauri-apps/plugins-workspace/issues/979


tooling/api/src/core.ts
```diff
@@ -123,6 +123,26 @@ async function addPluginListener<T>(
...
 
+type PermissionState = 'granted' | 'denied' | 'prompt' | 'prompt-with-rationale'
+
+/**
+ * Get permission state for a plugin.
+ *
+ * This should be used by plugin authors to wrap their actual implementation.
+ */
+async function checkPermissions<T>(plugin: string): Promise<T> {
+  return invoke(`plugin:${plugin}|check_permissions`)
```
amrbashir: won't this require an implementation in Rust? if so, then I'd prefer to keep them in the same place as the rust impl, not here

lucasfernog: these are abstract methods..
the API here is only meant to be used by plugin authors (e.g. FS, NFC, notification plugin) to share the proper types

https://github.com/tauri-apps/tauri/blob/36eee37220cff34a1fab35791dcd0775ae86ec7c/crates/tauri/mobile/android/src/main/java/app/tauri/plugin/Plugin.kt#L171

https://github.com/tauri-apps/tauri/blob/36eee37220cff34a1fab35791dcd0775ae86ec7c/crates/tauri/mobile/ios-api/Sources/Tauri/Plugin/Plugin.swift#L35

amrbashir: I see, this fine then

amrbashir (approved):

PR #10802, opened 2024-08-27
feat(cli): synchronize Tauri config and lib name with iOS Xcode project

lucasfernog:
- the Xcode project now uses a fixed output library name, which means changes to the Cargo.toml lib name won't affect it (backwards compatible change, we're checking if this new format is being used or not by reading the project.pbxproj)
- sync config identifier with the pbxproj
- sync development team config with the pbxproj

the sync runs both on dev and on build


lucasfernog:
CI fix #10803

amrbashir (approved):

PR #10805, opened 2024-08-27
chore: add root taurignore

lucasfernog:
enhances the DX of running `tauri dev` in any of the examples folder - we don't need to watch the entire workspace for changes

### DIFF
--- a/.taurignore
+++ b/.taurignore
@@ -1,0 +1,10 @@
+packages/api
+packages/cli
+crates/tauri-cli
+crates/tauri-bundler
+crates/tauri-driver
+crates/tauri-macos-sign
+crates/tauri-schema-generator
+crates/tests
+bench
+.changes
--- a/.taurignore
+++ b/.taurignore
@@ -1,3 +1,11 @@
+.changes
+.devcontainer
+.docker
+.github
+.scripts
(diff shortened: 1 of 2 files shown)


lucasfernog:
we could ignore the entire crates folder, idk about that
tbh i don't use `tauri dev` in the examples much, it's mostly mobile dev that impacts me

.taurignore
```diff
@@ -0,0 +1,10 @@
+packages/api
+packages/cli
+crates/tauri-cli
+crates/tauri-bundler
+crates/tauri-driver
+crates/tauri-macos-sign
+crates/tauri-schema-generator
+crates/tests
+bench
+.changes
```
amrbashir: if we ignoring .changes, should we also ignore `.github`, `.docker` and other directories ?

lucasfernog: how often do we change these anyway? 😅

amrbashir: never but this is about local development and it could either be accidentally saved or you could be changing a workflow while working on a feature, I'd say add them just in case

lucasfernog: cool will do that tomorrow when i wake up

amrbashir (approved):

PR #10816, opened 2024-08-28
fix(ci): publish CLI workflow wrong target path

lucasfernog:


lucasfernog:
https://github.com/tauri-apps/tauri/actions/runs/10603218524/job/29387029286

PR #10838, opened 2024-08-30
fix(windows): Handle root paths that cannot be canonicalized

### LINKED ISSUE #10815: [bug] Infinite loop in scope::fs::push_pattern
### Describe the bug

On Windows, calling `tauri::scope::fs::push_pattern` with a pattern of "\\\\?\\**" results in an infinite loop attempting to create the canonicalized path.

### Reproduction

1) In an app's capabilities JSON add to the permissions array:
```json
{ "identifier": "fs:scope", "allow": ["\\\\?\\**"] }
```
2) Perform any fs API call from the frontend. The backend will lock up in an infinite loop when creating the `Scope`.

### Expected behavior

`push_pattern` should either return an error or process the pattern successfully.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.19045 X64
    ✔ WebView2: 128.0.2739.42
    ✔ rustc: 1.80.0 (051478957 2024-07-21)
    ✔ cargo: 1.80.0 (376290515 2024-07-16)
    - node: 18.19.1
    - pnpm: 9.9.0
    - npm: 10.2.4

[-] Packages
    - tauri [RUST]: 2.0.0-rc.2
    - tauri-build [RUST]: 2.0.0-rc.2
    - wry [RUST]: 0.41.0
    - tao [RUST]: 0.28.1
    - tauri-cli [RUST]: 1.5.11
    - @tauri-apps/api [NPM]: 2.0.0-rc.0
    - @tauri-apps/cli [NPM]: 2.0.0-rc.3

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - framework: SolidJS
    - bundler: Vite
```

### Stack trace

_No response_

### Additional context

_No response_

alexmoon:
If the root path element of a pattern could not be canonicalized, `scope::fs::push_pattern` would enter an infinite loop. This is because `path.pop()` does nothing and returns `false` when called on a path with no parent, which can happen even when the path element iterator is non-empty. This could occur on Windows, for example, with a pattern of `\\?\**`.

Fixes #10815

### DIFF
--- a/crates/tauri/src/scope/fs.rs
+++ b/crates/tauri/src/scope/fs.rs
@@ -77,7 +77,7 @@ fn push_pattern<P: AsRef<Path>, F: Fn(&str) -> Result<Pattern, glob::PatternErro
   pattern: P,
   f: F,
 ) -> crate::Result<()> {
-  let path: PathBuf = pattern.as_ref().components().collect();
+  let path: PathBuf = dunce::simplified(pattern.as_ref()).components().collect();
   list.insert(f(&path.to_string_lossy())?);
 
   let mut path = path;
@@ -514,29 +514,23 @@ mod tests {
       assert!(!scope.is_allowed("\\\\.\\COM2"));
     }
 
-    // This case does not work because a drive letter without a backslash (like "C:")
-    // canonicalizes to the current working directory of that drive. On the other hand,
-    // adding a backslash (like "C:\") results in the patterns being constructed with
-    // double backslahes (e.g. "C:\\**") which then do not match.
-    // let scope = new_scope();
-    // {
-    //   // Drive root
-    //   scope.allow_directory("C:", true).unwrap();
-    //   assert!(scope.is_allowed("C:\\Windows"));
-    //   assert!(scope.is_allowed("C:\\Windows\\system.ini"));
-    //   assert!(scope.is_allowed("C:\\NonExistentFile"));
-    //   assert!(!scope.is_allowed("D:\\home"));
-    // }
+    let scope = new_scope();
+    {
+      // Drive root
+      scope.allow_directory("C:\\", true).unwrap();
+      assert!(scope.is_allowed("C:\\Windows"));
+      assert!(scope.is_allowed("C:\\Windows\\system.ini"));
+      assert!(scope.is_allowed("C:\\NonExistentFile"));
+      assert!(!scope.is_allowed("D:\\home"));
+    }
 
     let scope = new_scope();
     {
       // Verbatim drive root
-      scope.allow_directory("\\\\?\\C:", true).unwrap();
+      scope.allow_directory("\\\\?\\C:\\", true).unwrap();
       assert!(scope.is_allowed("C:\\Windows"));
       assert!(scope.is_allowed("C:\\Windows\\system.ini"));
-      // Does not work because non-existent files can't be canonicalized, therefore
-      // the verbatim prefix ("\\?\") is not added.
-      // assert!(scope.is_allowed("C:\\NonExistentFile"));
+      assert!(scope.is_allowed("C:\\NonExistentFile"));
       assert!(!scope.is_allowed("D:\\home"));
     }
 
--- a/crates/tauri/src/scope/fs.rs
+++ b/crates/tauri/src/scope/fs.rs
@@ -77,7 +77,21 @@ fn push_pattern<P: AsRef<Path>, F: Fn(&str) -> Result<Pattern, glob::PatternErro
   pattern: P,
   f: F,
 ) -> crate::Result<()> {
-  let path: PathBuf = dunce::simplified(pattern.as_ref()).components().collect();
+  let mut path: PathBuf = dunce::simplified(pattern.as_ref()).components().collect();
+
+  if cfg!(windows) {
+    // Canonicalize disk-relative paths before inserting into the list
+    use std::path::{Component, Prefix};
+    let mut components = path.components();
+    if let Some(Component::Prefix(prefix)) = components.next() {
+      if matches!(prefix.kind(), Prefix::Disk(_) | Prefix::VerbatimDisk(_))
+        && !matches!(components.next(), Some(Component::RootDir))
+      {
+        path = dunce::simplified(&path.canonicalize()?).to_path_buf();
+      }
+    }
+  }
+
   list.insert(f(&path.to_string_lossy())?);
 
   let mut path = path;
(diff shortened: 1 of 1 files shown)


crates/tauri/src/scope/fs.rs
```diff
@@ -123,7 +125,7 @@ fn push_pattern<P: AsRef<Path>, F: Fn(&str) -> Result<Pattern, glob::PatternErro
 
   if let Some(p) = canonicalized {
     list.insert(f(&p.to_string_lossy())?);
-  } else if cfg!(windows) {
+  } else if cfg!(windows) && !path.to_string_lossy().starts_with("\\\\") {
```
FabianLars, line 141: My brain is still in vacation mode but couldn't we end up with no allowed paths with this check?

Also i think this whole issue calls for new test cases if you wouldn't mind adding them :)

alexmoon: The un-canonicalized path is added to the list above (line 81). A Windows path that starts with `\\` is already a UNC, device, or verbatim path and prefixing it with `\\?\` doesn't make sense.

I'll work on some test cases.

FabianLars: Ahh damn, missed that line, sry.

alexmoon:
I've added a test case for the various flavors of Windows root paths. I ran into a number of edge cases that are not currently handled well, which I documented in comments in the test case. I can open an issue for those if you would like. A few examples:

- Canonicalization differs depending on if the path exists. For example "C:\Exists" canonicalizes to "\\\\?\C:\Exists" while "C:\DoesNotExit" is left as is.
- A similar issue affects UNC network share paths. If the network share is present/reachable when a pattern is added you will get different patterns added to the scope than if the share is not present/reachable.
- The path "C:" is different from "C:\\" (the former refers to the current working directory of the C drive, while the second refers to the root directory of the C drive). There's effectively no way to pass "C:\\" to `add_directory` because it unconditionally adds a backslash before the wildcards which leads to a pattern like "C:\\\\**" which doesn't match anything.
- If you add a verbatim path (one that starts with "\\\\?\\"), it won't match non-verbatim paths that refer to the same target.

crates/tauri/src/scope/fs.rs
```diff
@@ -470,4 +472,79 @@ mod tests {
...
+    let scope = new_scope();
+    {
+      // UNC network path
+      scope.allow_directory("\\\\localhost\\c$", true).unwrap();
+      assert!(scope.is_allowed("\\\\localhost\\c$"));
+      assert!(scope.is_allowed("\\\\localhost\\c$\\Windows"));
+      // Does not work because push_pattern adds a trailing backslash to the base
+      // pattern before "\**" is added leading to a pattern of "\\localhost\c$\\**"
+      // which does not match.
+      // assert!(scope.is_allowed("\\\\localhost\\c$\\NonExistentFile"));
```
amrbashir: this case should definitely work, maybe just check that path doesn't end with `\` before appending `\**`

crates/tauri/src/scope/fs.rs
```diff
@@ -470,4 +472,79 @@ mod tests {
...
+    {
+      // Verbatim UNC network path
+      scope
+        .allow_directory("\\\\?\\UNC\\localhost\\c$", true)
+        .unwrap();
+      assert!(scope.is_allowed("\\\\localhost\\c$"));
+      assert!(scope.is_allowed("\\\\localhost\\c$\\Windows"));
+      // Does not work because non-existent files can't be canonicalized, therefore
+      // the verbatim prefix ("\\?\UNC\") is not added.
+      // assert!(scope.is_allowed("\\\\localhost\\c$\\Windows\\NonExistentFile"));
```
amrbashir: Not sure I understand why this won't work, I see that `"\\\\localhost\\c$\\Windows"` works but why not `"\\\\localhost\\c$\\Windows\\NonExistentFile"`, isn't it just one level deeper?

crates/tauri/src/scope/fs.rs
```diff
@@ -470,4 +472,79 @@ mod tests {
...
+    // double backslahes (e.g. "C:\\**") which then do not match.
+    // let scope = new_scope();
+    // {
+    //   // Drive root
+    //   scope.allow_directory("C:", true).unwrap();
+    //   assert!(scope.is_allowed("C:\\Windows"));
+    //   assert!(scope.is_allowed("C:\\Windows\\system.ini"));
+    //   assert!(scope.is_allowed("C:\\NonExistentFile"));
+    //   assert!(!scope.is_allowed("D:\\home"));
+    // }
```
amrbashir: I think we can fix this by calling `std::fs::canonicalize` which will resolve `C:` to its working directory, and we add that to our allowed list. However, it will contain `\\?\` prefix, so we can use `dunce::simplified` to remove this prefix and append the un-prefixed path to our allowed list.

amrbashir (commented):
Thank you and apologies for the late reply

crates/tauri/src/scope/fs.rs
alexmoon: The problem is if you allow a directory with a verbatim path (e.g. `\\\\?\UNC\localhost\c$`), then the equivalent non-verbatim version of the path (e.g. `\\\\localhost\c$`) is not allowed. In general, I'm not sure it's safe to convert a verbatim path to a non-verbatim path so this is probably desirable behavior but it is a bit surprising. Especially when combined with the behavior of canonicalizing existing paths into their verbatim equivalents.

crates/tauri/src/scope/fs.rs
alexmoon: Ok, I've updated it using `dunce::simplified` to make most of the remaining edge cases work. Using a bare drive letter (like `C:` or `\\?\C:`) still won't work do to the canonicalization issue, but it seems like maybe that should be disallowed in `allow_file` and `allow_directory` since it makes the permission dependent on an the environment which could have an unexpected value.

crates/tauri/src/scope/fs.rs
alexmoon: Note that `dunce::simplified()` doesn't simplify UNC network paths, only disk based paths.

crates/tauri/src/scope/fs.rs
amrbashir: Calling `std::fs::canonicalize` with `C:` works, but doesn't work with `\\?\C:` is what is stated in [MSDN](https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats#apply-the-current-directory):
> If a path isn't fully qualified, Windows applies the current directory to it. UNCs and device paths do not have the current directory applied. Neither does a full drive with separator C:\.

crates/tauri/src/scope/fs.rs
amrbashir: I would just ignore this case then and leave it up to the user to specify both versions of the path

crates/tauri/src/scope/fs.rs
alexmoon: That's correct. However, `\\?\C:` is simplified by `dunce` to `C:` which is then canonicalized to the current path of that drive.

I've added canonicalization of disk-relative paths to `push_pattern`. I think this should cover everything we've discussed.

amrbashir (approved):
Thank you. Unfortunately, you don't seem to have commit signing enabled and I can't merge the PR until 
it is signed.

You need to setup [commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits), then you can sign past commit like [this](https://superuser.com/questions/397149/can-you-gpg-sign-old-commits/1123928#1123928). `git rebase --exec 'git commit --amend --no-edit -n -S' -i dev`

alexmoon:
Sorry about that. My Windows machine doesn't have signing set up.

amrbashir:
I have a few changes locally while testing this PR, that simplifies the code a bit and adds more documentation. Since I can't push to your branch, I will open a follow up PR. Thanks for your help and the test cases as well.

amrbashir (approved):

PR #10840, opened 2024-08-31
feat(core): always use the custom protocol IPC on Linux

lucasfernog:
We can always use the custom protocol based IPC on Linux since raising the minimum required webkit2gtk does not affect the minimum Linux distros supported by Tauri


lucasfernog:
requested by @tweidinger and @FabianLars :D

lucasfernog:
i'm on my mac laptop now but i can test this one tomorrow

amrbashir (approved):

FabianLars (approved):

PR #10844, opened 2024-08-31
fix(core): clippy warnings, simplify embed_plist usage

lucasfernog:


examples/file-associations/src-tauri/src/main.rs
```diff
@@ -29,6 +29,9 @@ fn handle_file_associations(app: AppHandle, files: Vec<PathBuf>) {
 fn main() {
   tauri::Builder::default()
     .setup(|app| {
+      #[cfg(not(any(windows, target_os = "linux")))]
+      let _app = app;
```
amrbashir: this complicates the example a bit, I'd rather just add `#![allow(unused)]` at the top

amrbashir (dismissed):

examples/file-associations/src-tauri/src/main.rs
lucasfernog: i can add it inline.. i hate suppressing warns globally ;D

amrbashir (approved):

PR #10866, opened 2024-09-02
feat(bundler): add `TAURI_BUNDLER_TOOLS_GITHUB_MIRROR`to specify a GitHub mirror

thep0y:
This feature is designed to facilitate the use of `tauri` for developers in **Mainland China** or other regions where **GitHub is not accessible**.

This commit introduces a feature that checks for the presence of an environment variable `TAURI_BUNDLER_TOOLS_GITHUB_MIRROR` before downloading the webview2 offline installer. If `TAURI_BUNDLER_TOOLS_GITHUB_MIRROR` is set, the download URL is rewritten to use a specified GitHub mirror, which can improve download speeds for users in regions where direct access to GitHub might be slow or unreliable. This change allows users to benefit from faster downloads without needing to configure proxy servers, particularly beneficial for those in mainland China where alternative mirrors are commonly used.

This commit helps resolve the issue encountered in https://github.com/tauri-apps/tauri/issues/7338.

PowerShell:
```powershell
$Env:TAURI_BUNDLER_TOOLS_GITHUB_MIRROR='https://mirror.ghproxy.com/'
bun run tauri build
pnpm run tauri build
npm run tauri build
```


crates/tauri-bundler/src/bundle/windows/util.rs
```diff
@@ -70,8 +70,20 @@ pub fn download_webview2_offline_installer(base_path: &Path, arch: &str) -> crat
 pub fn download(url: &str) -> crate::Result<Vec<u8>> {
   log::info!(action = "Downloading"; "{}", url);
 
-  let agent = ureq::AgentBuilder::new().try_proxy_from_env(true).build();
-  let response = agent.get(url).call().map_err(Box::new)?;
+  let (agent, url) = match std::env::var("GITHUB_CDN") {
```
amrbashir: I think `GITHUB_CDN` is a bit vague, maybe `TAURI_BUNDLER_TOOLS_DOWNLOAD_HOST` is better here. You also need to document that env var in `crates/tauri-cli/ENVIRONEMENT_VARIABLES.md` file

crates/tauri-bundler/src/bundle/windows/util.rs
```diff
@@ -70,8 +70,20 @@ pub fn download_webview2_offline_installer(base_path: &Path, arch: &str) -> crat
...
+    Ok(cdn) => {
+      let mut parsed_cdn = url::Url::parse(&cdn).expect("Invalid cdn");
```
amrbashir: We should propagate the error using `?` instead of panicing with `.expect` or better use it with `.context()` from `anyhow` crate.

```suggestion
      let mut parsed_cdn = url::Url::parse(&cdn).context("invalid tools download host URL")?;
```

You will need to import `anyhow::Context` trait

crates/tauri-bundler/src/bundle/windows/util.rs
```diff
@@ -70,8 +70,20 @@ pub fn download_webview2_offline_installer(base_path: &Path, arch: &str) -> crat
...
+      parsed_cdn.set_path(url);
+
+      (ureq::AgentBuilder::new().build(), parsed_cdn.to_string())
+    }
+    Err(_) => (
+      ureq::AgentBuilder::new().try_proxy_from_env(true).build(),
```
amrbashir: we can keep using `AgentBuilder` with `try_proxy_from_env` in the two cases, no need to make distinction, only `url` should be changed based on the new env var

amrbashir (changes requested):
Thank you, could you also add a change file in `.changes` directory?

crates/tauri-bundler/src/bundle/windows/util.rs
thep0y: Currently, tauri-bundler downloads files from GitHub, which is why the variable name `GITHUB_CDN` is used. 

If future versions source downloads from other websites, the variable `TAURI_BUNDLER_TOOLS_DOWNLOAD_HOST` might cause confusion or errors. To avoid this, would it be more appropriate to rename it to `TAURI_BUNDLER_TOOLS_DOWNLOAD_GITHUB_CDN` or `TAURI_BUNDLER_TOOLS_DOWNLOAD_GITHUB_TEMPLATE`?

Or is it unlikely that future versions of tauri-bundler will download files from any site other than GitHub?

crates/tauri-bundler/src/bundle/windows/util.rs
thep0y: > we can keep using `AgentBuilder` with `try_proxy_from_env` in the two cases, no need to make distinction, only `url` should be changed based on the new env var

Once a proxy is set in the environment variables along with a CDN (still referred to as a CDN for now), some CDNs might refuse to respond. Therefore, I believe the CDN and the `try_proxy_from_env` should be handled separately. If both a proxy and a CDN are set in the environment variables, it's likely that the user intends to use the CDN rather than the proxy, and both should not be used simultaneously.

crates/tauri-bundler/src/bundle/windows/util.rs
FabianLars: webview2, which this pr is about, isn't downloaded from github though?

thep0y: In accordance with internet regulations, users in mainland China are unable to directly download files from GitHub. A proxy is typically required. However, some users are not proficient in using proxies. Hence, I desire to add this feature—to facilitate file downloads from GitHub via public reverse proxy servers (CDN).

FabianLars: Yes, i understood this. I am not against this PR/change. I'm just saying that the GITHUB_CDN variable is used here for files that are not downloaded from github but from microsoft.

thep0y: I understand what you mean now, so I should add a source check, prioritizing CDN usage for files coming from GitHub, and for files from other source websites, downloading them based on `try_proxy_from_env`.

crates/tauri-bundler/src/bundle/windows/util.rs
amrbashir: makes sense

crates/tauri-bundler/src/bundle/windows/util.rs
```diff
@@ -67,11 +69,26 @@ pub fn download_webview2_offline_installer(base_path: &Path, arch: &str) -> crat
   Ok(file_path)
 }
 
+fn create_agent_and_url(url: &str) -> crate::Result<(ureq::Agent, String)> {
+  match std::env::var("TAURI_BUNDLER_TOOLS_DOWNLOAD_GITHUB_CDN") {
```
amrbashir: How about `TAURI_BUNDLER_TOOLS_GITHUB_MIRROR`?

thep0y: I think it's fine; the name of the variable doesn't matter to me, as it doesn't affect this feature.

amrbashir: cool, then we should rename it to `TAURI_BUNDLER_TOOLS_GITHUB_MIRROR`

crates/tauri-bundler/src/error.rs
```diff
@@ -58,6 +58,10 @@ pub enum Error {
   #[cfg(windows)]
   #[error("`{0}`")]
   Glob(#[from] glob::GlobError),
+  /// Failed to parse the github cdn
```
amrbashir:
```suggestion
  /// Failed to parse the URL
```

crates/tauri-cli/ENVIRONMENT_VARIABLES.md
```diff
@@ -15,6 +15,7 @@ These environment variables are inputs to the CLI which may have an equivalent C
 - `TAURI_CLI_NO_DEV_SERVER_WAIT` — Skip waiting for the frontend dev server to start before building the tauri application.
 - `TAURI_LINUX_AYATANA_APPINDICATOR` — Set this var to `true` or `1` to force usage of `libayatana-appindicator` for system tray on Linux.
 - `TAURI_BUNDLER_WIX_FIPS_COMPLIANT` — Specify the bundler's WiX `FipsCompliant` option.
+- `TAURI_BUNDLER_TOOLS_DOWNLOAD_GITHUB_CDN ` - Specify a GitHub CDN to speed up the download of necessary files.
```
amrbashir:
```suggestion
- `TAURI_BUNDLER_TOOLS_GITHUB_MIRROR` - Specify a GitHub mirror to download files and tools used by tauri bundler.
```

.changes/bundler-github-cdn-from-env.md
```diff
@@ -0,0 +1,6 @@
+---
+"tauri-bundler": patch:feat
+"tauri-cli": patch:feat
+---
+
+Use the environment variable `TAURI_BUNDLER_TOOLS_DOWNLOAD_GITHUB_CDN` to specify a GitHub CDN for faster downloads of necessary files. This is designed for areas like Mainland China where GitHub access can be unreliable.
```
amrbashir:
```suggestion
Add `TAURI_BUNDLER_TOOLS_GITHUB_MIRROR` environment variable to specify a GitHub mirror to download files and tools used by tauri bundler. This is designed for areas like Mainland China where GitHub access can be unreliable.
```

crates/tauri-bundler/src/bundle/windows/util.rs
amrbashir:
```suggestion
  match std::env::var("TAURI_BUNDLER_TOOLS_GITHUB_MIRROR") {
```

amrbashir (changes requested):

crates/tauri-bundler/src/error.rs
```diff
@@ -58,6 +58,10 @@ pub enum Error {
   #[cfg(windows)]
   #[error("`{0}`")]
   Glob(#[from] glob::GlobError),
+  /// Failed to parse the URL
+  #[cfg(windows)]
```
amrbashir:
```suggestion
```

amrbashir (approved):

PR #10871, opened 2024-09-03
feat: add tauri-schema-worker

amrbashir:
ref: https://github.com/tauri-apps/create-tauri-app/issues/725#issuecomment-2185859045


crates/tauri-schema-worker/src/config.rs
```diff
@@ -0,0 +1,142 @@
+use anyhow::Context;
```
amrbashir, line 5:
```suggestion
// Copyright 2019-2024 Tauri Programme within The Commons Conservancy
// SPDX-License-Identifier: Apache-2.0
// SPDX-License-Identifier: MIT

use anyhow::Context;
```

.github/workflows/deploy-schema-worker.yml
```diff
@@ -0,0 +1,25 @@
+# Copyright 2019-2022 Tauri Programme within The Commons Conservancy
```
amrbashir:
```suggestion
# Copyright 2019-2024 Tauri Programme within The Commons Conservancy
```

.changes/config.json
```diff
@@ -212,16 +203,16 @@
         "tauri-runtime-wry",
         "tauri-build"
       ],
-      "postversion": "node ../../.scripts/ci/sync-cli-metadata.js ${ pkg.pkg } ${ release.type }"
+      "postversion": [
+        "node ../../.scripts/ci/sync-cli-metadata.js ${ pkg.pkg } ${ release.type }",
+        "cargo build --manifest-path ../tauri-schema-generator/Cargo.toml"
```
lucasfernog, line 208: should probably do this only when tauri-utils bumps instead

lucasfernog: nvm when utils bump tauri bumps too anyway.. and the schema is tied to the tauri version

crates/tauri-schema-worker/src/config.rs
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  let Some(version) = releases.into_iter().find(|r| version.matches(&r.version)) else {
+    return max_stable_schema().await;
+  };
+
+  schema_file_for_version(version.version).await
+}
+
+async fn crate_releases(crate_: &str) -> anyhow::Result<Vec<CrateRelease>> {
```
lucasfernog: what do you think about caching these for a while? maybe in the future even add a webhook for tauri publishes?

.github/workflows/deploy-schema-worker.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: cloudflare/wrangler-action@3
+        with:
+          command: deploy
+          workingDirectory: 'packages/schema-worker'
```
lucasfernog: isn't this in the crates folder?

amrbashir: yeah that's correct, I initially wrote the worker in JS but later switched to rust so the directory changed

crates/tauri-schema-worker/src/config.rs
amrbashir: that's a good idea, will have to read more about workers 😅

lucasfernog: maybe the worker can cache the JSON schema per version, then when we request `/config/2` we just redirect to the latest version `/config/2.1.2` or whatever, so we don't really need to deal with any cache on our end
David knows how to do it, he wrote some cloudflare caching in the past

amrbashir: `/config/2` is resolved to `2.1.2` or whatever latest is, and response will be cached for resolved version so `/config/2` or `/config/2.1` or `/config/2.1.2` or `/config/latest` (and other aliases) will use the same file from cache.

lucasfernog (approved):

lucasfernog:
this is just a beauty

Issue #9969, opened 2024-06-04
[feat] add visionOS support


eugenehp:
### Describe the problem

Add support for Apple Vision Pro

### Describe the solution you'd like

same as iOS 

### Alternatives considered

Working on better support in Tao

### Additional context

Current PRs submitted:

- [ ] https://github.com/tauri-apps/wry/pull/1286
- [ ] https://github.com/SSheldon/rust-objc/pull/123
- [ ] https://github.com/SSheldon/rust-block/pull/20
- [ ] https://github.com/tauri-apps/cargo-mobile2/pull/340
- [x] https://github.com/Brendonovich/swift-rs/pull/63
- [x] https://github.com/oconnor663/os_pipe.rs/pull/32
- [x] https://github.com/Byron/open-rs/pull/99
- [x] https://github.com/rust-lang/socket2/pull/503
- [ ] https://github.com/briansmith/ring/pull/2032

eugenehp:
`objc` is least likely to merge the PR, so what would be a good course of action here? patch or put a `objc2` under `Tauri`?

FabianLars:
> objc is least likely to merge the PR, so what would be a good course of action here? patch or put a objc2 under Tauri?

@pewsheen is working on porting wry to objc2 already but it's a lot of work.

eugenehp:
Nice! There's also [`objc2` ](https://github.com/madsmtm/objc2) by @madsmtm but it doesn't have full API parity with the original `objc` as far as I know.

So for the time being, what would be the appropriate way to patch this in Tauri (Wry and Tao) @FabianLars

FabianLars:
tbh I'd probably wait and see if we can get wry ported in time, especially considering that we're still waiting for other projects than objc. Devs that want to use it earlier can use cargo's patch feature to overwrite the crates.io version with your forks.

eugenehp:
Sounds good. That's what we're using right now – patches. Thinking maybe to put some temporary crates on crates.io to make things easier later, if no news on merging/porting main repos.

madsmtm:
Linking the issue in Wry for tracking the migration to `objc2`: https://github.com/tauri-apps/wry/issues/1239

> Nice! There's also [`objc2` ](https://github.com/madsmtm/objc2) by @madsmtm but it doesn't have full API parity with the original `objc` as far as I know.

Could you expand on what it's missing compared to `objc`? It's originally a fork of that project, so it should have full feature-parity (and more, ofc ;) ).

[...]

PR #10924, opened 2024-09-08
feat: Migrate to `objc2`

madsmtm:
I've migrated a lot of the crates Tauri depends on, see backrefererences to https://github.com/tauri-apps/wry/issues/1239,  let's do it for Tauri as well.

The benefit to this part of Tauri's codebase is honestly not that big, I'm doing this mostly for consistency, though it does unblock https://github.com/tauri-apps/tauri/issues/9969.

I haven't signed the commit, it's kind of a hazzle that I don't really want to deal with, feel free to rebase it into oblivion, I don't need the attribution.


madsmtm:
Not sure I did the changelog entries correctly

crates/tauri-runtime-wry/src/webview.rs
```diff
@@ -13,25 +13,17 @@ mod imp {
...
-
-  pub struct Webview {
-    pub webview: id,
-    pub manager: id,
-    pub view_controller: id,
+    pub webview: *mut c_void,
+    pub manager: *mut c_void,
+    #[cfg(target_os = "macos")]
+    pub ns_window: *mut c_void,
+    #[cfg(not(target_os = "macos"))]
```
lucasfernog:
```suggestion
    #[cfg(target_os = "ios")]
```
just for consistency with the usage

lucasfernog: also makes it easier for us to locate it when we want to support visionOS

lucasfernog (approved):
this is awesome!

PR #10938, opened 2024-09-10
feat: Add support for using Product Name 

### LINKED ISSUE #10548: [feat] `tauri migrate` shoud migrate main bin name to product name
### Describe the problem

#9375 removed renaming binary to product name specified in `Tauri.toml`.

However, this is not migrated with `tauri migrate` and unexpectedly broke user installed shortcuts

I think it's much better to migrate main bin name to product name in `tauri migrate` command.

I'm going to create issue for documentation.

### Alternatives considered

_No response_

### Additional context

_No response_

Geometrically:
This PR is based off the work here https://github.com/gitbutlerapp/tauri/pull/1 by @krivi. (most of this PR description was taken from there- it was very well written!)

## Background:
This change in Tauri https://github.com/tauri-apps/tauri/pull/9375/files, which is noted as a refactor is in fact a breaking change as far as the app updater is concerned.

The impact is that because the binary now has a different name, it is not possible to smoothly update a Tauri application via the updater if the initial version of the app is using `v1.x` and the new version is using `v2.x`.

## Approach:
I'm introducing here an option for the `tauri-cli` (`tauri-bundler`) for producing the same binary names it did in `v1`, allowing users to update the app via the updater. In order to enable this behavior, set the environment variable

```
V1_COMPATIBLE_BIN_NAME=true
```

Previously, the bundler did the rename just after building the binary. I poked and tried doing the same once again, but because quite a lot of the code in that bundler changed since that was merged, it felt possibly more error prone (since now the rest of the code can safely assume that the binary is the crate name)

So, instead, my approach is to do the rename the binary at the last moment (i.e. in the `bundle_project` function in eg. app.rs)

fixes #10548

## Testing
This PR has been tested across Windows, MacOS, and Linux with the Modrinth App. You can find test builds built using this PR here: https://github.com/modrinth/code/actions/runs/10785759488

### DIFF
--- a/crates/tauri-bundler/Cargo.toml
+++ b/crates/tauri-bundler/Cargo.toml
@@ -42,6 +42,7 @@ sha1 = "0.10"
 sha2 = "0.10"
 zip = { version = "2.0", default-features = false, features = ["deflate"] }
 dunce = "1"
+heck = "0.5"
 
 [target."cfg(target_os = \"windows\")".dependencies]
 uuid = { version = "1", features = ["v4", "v5"] }
@@ -63,7 +64,6 @@ tauri-macos-sign = { version = "0.1.1-rc.0", path = "../tauri-macos-sign" }
 regex = "1"
 
 [target."cfg(target_os = \"linux\")".dependencies]
-heck = "0.5"
 ar = "0.9.0"
 md5 = "0.7.0"
 rpm = "0.15.0"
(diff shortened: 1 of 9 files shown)


amrbashir:
I am looking into all issues related to v1 -> v2 updater so bear with me until I get this all sorted out.

Geometrically:
No worries- I fully tested this on Windows/MacOS/Linux and everything seems to be working as expected :)

lucasfernog:
I believe we can close this one now that we have #10939 #10996 #10977 #10969 #10967 #10962 #10952

amrbashir:
yeah, that correct, this is not longer needed and if `productName` needs to be used, you can set the new `mainBinaryName` option to the same value as `productName`.

Apologies for the inconvenience.

PR #10939, opened 2024-09-10
feat(updater): support updating using non-zipped files

amrbashir:
This backported from tauri-plugin-updater for v2


core/tauri/src/updater/core.rs
```diff
@@ -622,12 +626,12 @@ impl<R: Runtime> Update<R> {
...
-      "Accept",
-      HeaderValue::from_str("application/octet-stream").unwrap(),
+      header::ACCEPT,
+      HeaderValue::from_str("application/octet-stream")?,
     );
     headers.insert(
-      "User-Agent",
-      HeaderValue::from_str("tauri/updater").unwrap(),
+      header::USER_AGENT,
+      HeaderValue::from_str(UPDATER_USER_AGENT)?,
```
lucasfernog, line 634: we shouldn't change the user agent - just in case it breaks someone's server

amrbashir: cool, reverted in [`035093e` (#10939)](https://github.com/tauri-apps/tauri/pull/10939/commits/035093ea14058c7c437f2b626b407900297411f8)

lucasfernog (approved):

PR #10940, opened 2024-09-10
feat(cli): generate signatures for non-zipped targets with `createUpdaterArtifacts: v1Compatible`

amrbashir:

### DIFF
--- a/crates/tauri-cli/src/bundle.rs
+++ b/crates/tauri-cli/src/bundle.rs
@@ -226,17 +226,10 @@ fn sign_updaters(
   let update_enabled_bundles: Vec<&tauri_bundler::Bundle> = bundles
     .iter()
     .filter(|bundle| {
-      if update_settings.v1_compatible {
-        matches!(bundle.package_type, PackageType::Updater)
-      } else {
-        matches!(
-          bundle.package_type,
-          PackageType::Updater
-            | PackageType::Nsis
-            | PackageType::WindowsMsi
-            | PackageType::AppImage
-        )
-      }
+      matches!(
+        bundle.package_type,
+        PackageType::Updater | PackageType::Nsis | PackageType::WindowsMsi | PackageType::AppImage
+      )
     })
     .collect();
 
--- a/.changes/sign-non-zipped-updater-with-v1-compatible.md
+++ b/.changes/sign-non-zipped-updater-with-v1-compatible.md
@@ -1,0 +1,6 @@
+---
+"tauri-cli": "patch:enhance"
+---
+
+Generate `.sig` signature files for installers and bundles when `createUpdaterArtifacts` is set to `v1Compatible`
+


crates/tauri-cli/src/bundle.rs
```diff
@@ -226,17 +226,10 @@ fn sign_updaters(
   let update_enabled_bundles: Vec<&tauri_bundler::Bundle> = bundles
     .iter()
     .filter(|bundle| {
-      if update_settings.v1_compatible {
-        matches!(bundle.package_type, PackageType::Updater)
```
amrbashir, line 230: cc @Legend-Master since you were the one to add this check, was there a specific reason I should be aware of? or removing the check sounds okay with you?

Legend-Master: This was to match the v1 behavior so that we only generate the signatures for updaters, not the non-zipped ones, if we decide to also generate it for non-zipped targets, removing this should be fine

Legend-Master: Just note that we pick the updater based on the signature file in tauri-actions and if we generate it for both (it used to be only a single signature file), we might need to make some tweaks there

Legend-Master: Maybe I ask what's this change for? I saw another PR for supporting non-zipped updaters in v1, but how does it relate to the v2 cli?

amrbashir: It is not related to anything, just a behavior I noticed while working on non-zipped updater and thought it was odd. I mean I was expecting:
- `createUpdaterArtifacts: true` to generate only for .msi and .setup.exe files
- `createUpdaterArtifacts: v1Compatible` to generate for .msi and .setup.exe files and also .zip legacy updater

amrbashir: > Just note that we pick the updater based on the signature file in tauri-actions and if we generate it for both (it used to be only a single signature file), we might need to make some tweaks there

Can't we check the value of `createUpdaterArtifacts` and pick signature file based on it? I mean if, `createUpdaterArtifacts: v1Compatible` then look for `.zip.sig`, otherwise just use normal `.sig` files

amrbashir: cc @FabianLars for this as well

Legend-Master: > `createUpdaterArtifacts: v1Compatible` to generate for .msi and .setup.exe files and also .zip legacy updater

Is there any reason to generate the signatures for the .msi and .setup.exe files in this mode? Old updaters wouldn't be able to upgrade with these anyways, and if they do after the PR that introduces it, why not to just use `createUpdaterArtifacts: true`?

> Can't we check the value of `createUpdaterArtifacts` and pick signature file based on it? I mean if, `createUpdaterArtifacts: v1Compatible` then look for `.zip.sig`, otherwise just use normal `.sig` files

Yes of course, just a note that we need to do it

amrbashir: > Is there any reason to generate the signatures for the .msi and .setup.exe files in this mode? Old updaters wouldn't be able to upgrade with these anyways, and if they do after the PR that introduces it, why not to just use `createUpdaterArtifacts: true`?

I did think of removing `v1Compatible` after the other PR, and it is probably the logical thing to do, but let's wait a bit on this

FabianLars: > createUpdaterArtifacts: v1Compatible to generate for .msi and .setup.exe files and also .zip legacy updater

This makes no sense to me tbh but generating 1 or 2 additional sigs doesn't matter so i won't argue against it. The only thing (the same thing as always) is that people won't read the docs (yes, we have them for the updater) and chose the wrong sigs. 😩 

> Can't we check the value of createUpdaterArtifacts and pick signature file based on it? I mean if, createUpdaterArtifacts: v1Compatible then look for .zip.sig, otherwise just use normal .sig files

We currently search for the files in order. First we look for the zip, then fall back to non-zipped. If you don't want it to pick up the zipped one just don't create them 🤷

amrbashir: > > Can't we check the value of createUpdaterArtifacts and pick signature file based on it? I mean if, createUpdaterArtifacts: v1Compatible then look for .zip.sig, otherwise just use normal .sig files
> 
> We currently search for the files in order. First we look for the zip, then fall back to non-zipped. If you don't want it to pick up the zipped one just don't create them 🤷

This makes sense, and this PR shouldn't affect that, right?

FabianLars: no it shouldn't

FabianLars: Ah actually it kinda does https://github.com/tauri-apps/tauri-action/blob/dev/src/upload-version-json.ts#L122-L129 but don't let that stop you here, i'll double check this later and update if needed.

amrbashir: As this PR is not important, I'd rather wait for tauri-action to accommodate first so we don't potentially break users

Legend-Master: I still don't see why we want to do this, but I'm fine with it if you persist

I think we only did the zip first check for nsis and msi installers (forget why I did it with a sort 🤣), so the other ones will be randomly picked (maybe not but not something that reliable), some changes are probably needed

amrbashir: > I still don't see why we want to do this, but I'm fine with it if you persist

Not persisting, I am fine with closing the PR, I just thought it was weird that `createUpdaterArtifacts: v1Compatible` would disable generating .sig for normal installers.

In my mind, I intuitively think that v1Compatible should in fact be compatible with v1 but shouldn't disable the default behavior but that could just be me.

Legend-Master: The way I think about it is that it keeps the v1 behavior which only generates the signature for the zipped updater, maybe it could have a better name by I can't think of one right now

FabianLars: https://github.com/tauri-apps/tauri-action/pull/908 don't judge me

lucasfernog (approved):

PR #10946, opened 2024-09-10
fix(core): Mark `identifier` as a required field.

FabianLars:
Noticed in https://discord.com/channels/616186924390023171/1276353618093674518 where only the .msi build failed to build even though bundle.publisher was set. If both publisher and identifier are set it correctly uses publisher and since identifier is required with this pr i don't really care about that potential issue.


lucasfernog (approved):
ohhhhh i didn't even know about this one!

Legend-Master:
This change itself has nothing wrong, but now files like `tauri.windows.conf.json` that's supposed to be merged into the main config file will get a `Missing property "identifier".` warning which might cause confusions

FabianLars:
Shit forgot to mention that. 😳

but idk I'd rather have it required in the Schema for more tools to pick up the reqs than having some custom validation that only triggers on build.

Legend-Master:
Maybe generate another schema file without the requirement?

PR #10961, opened 2024-09-11
feat(cli): enhance iOS logs

lucasfernog:
enhance simulator and device run output by only displaying app logs by default, and printing full process logs when pedantic verbosity is requested

ref https://github.com/tauri-apps/cargo-mobile2/pull/388

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,3 +66,4 @@ opt-level = "s"
 # See https://github.com/GREsau/schemars/issues/120 for reference
 [patch.crates-io]
 schemars_derive = { git = 'https://github.com/tauri-apps/schemars.git', branch = 'feat/preserve-description-newlines' }
+tauri = { path = "./crates/tauri" }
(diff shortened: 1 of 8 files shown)


amrbashir (approved):

Cargo.toml
```diff
@@ -66,3 +66,4 @@ opt-level = "s"
 # See https://github.com/GREsau/schemars/issues/120 for reference
 [patch.crates-io]
 schemars_derive = { git = 'https://github.com/tauri-apps/schemars.git', branch = 'feat/preserve-description-newlines' }
+tauri = { path = "./crates/tauri" }
```
amrbashir, line 69: leftover ?

lucasfernog: i actually want to have the log plugin here in the workspace (otherwise we don't get any app logs on mobile) and it has to match the tauri dependency

amrbashir: I see, that's fine then

amrbashir (approved):

PR #10952, opened 2024-09-11
feat(bundler/nsis): store main binary name to be used in registry for future updates

amrbashir:

PR #10962, opened 2024-09-11
fix(bundler): store and remove old main binary on updates if main binary name has change

amrbashir:
ref: https://github.com/tauri-apps/tauri/pull/10952

### DIFF
--- a/crates/tauri-bundler/src/bundle/windows/msi/wix.rs
+++ b/crates/tauri-bundler/src/bundle/windows/msi/wix.rs
@@ -540,6 +540,7 @@ pub fn build_wix_app_installer(
   let shortcut_guid = generate_package_guid(settings).to_string();
   data.insert("shortcut_guid", to_json(shortcut_guid.as_str()));
 
+  // Note: `main_binary_name` is not used in our template but we keep it as it is potentially useful for custom temples
   let main_binary_name = settings.main_binary_name().to_string();
   data.insert("main_binary_name", to_json(&main_binary_name));
 
--- a/crates/tauri-bundler/src/bundle/windows/msi/wix.rs
+++ b/crates/tauri-bundler/src/bundle/windows/msi/wix.rs
@@ -540,8 +540,8 @@ pub fn build_wix_app_installer(
   let shortcut_guid = generate_package_guid(settings).to_string();
   data.insert("shortcut_guid", to_json(shortcut_guid.as_str()));
 
-  let app_exe_name = settings.main_binary_name().to_string();
-  data.insert("app_exe_name", to_json(app_exe_name));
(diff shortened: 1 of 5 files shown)


crates/tauri-bundler/src/bundle/windows/msi/wix.rs
```diff
@@ -540,8 +540,8 @@ pub fn build_wix_app_installer(
   let shortcut_guid = generate_package_guid(settings).to_string();
   data.insert("shortcut_guid", to_json(shortcut_guid.as_str()));
 
-  let app_exe_name = settings.main_binary_name().to_string();
-  data.insert("app_exe_name", to_json(app_exe_name));
+  let main_binary_name = settings.main_binary_name().to_string();
+  data.insert("main_binary_name", to_json(&main_binary_name));
```
lucasfernog, line 545: also unused, but i guess that's fine for custom templates? if so we should comment here

crates/tauri-bundler/src/bundle/windows/templates/installer.nsi
```diff
@@ -637,6 +637,15 @@ Section Install
     WriteRegStr SHCTX "${UNINSTKEY}" $MultiUser.InstallMode 1
   !endif
 
+  ; Remove old main binary if it doesn't match new main binary name
+  ReadRegStr $0 SHCTX "${UNINSTKEY}" "MainBinaryName"
```
lucasfernog, line 641: does this delete the binary on each install? since the v2 installer also uses the same key.. maybe we should use a different one for v2?

lucasfernog (dismissed):

crates/tauri-bundler/src/bundle/windows/msi/wix.rs
amrbashir: yeah it is unused but I think it should stay for custom templates, will add a comment

crates/tauri-bundler/src/bundle/windows/templates/installer.nsi
amrbashir: no this key will be removed when running the uninstaller, so it is only valid when doing updates. It has to be the same key for every installer (whether v1 or v2) so it can detect main binary name changes on updates

crates/tauri-bundler/src/bundle/windows/msi/wix.rs
amrbashir, line 545:
```suggestion
  // Note: `main_binary_name` is not used in our template but we keep it as it is potentially useful for custom temples
  let main_binary_name = settings.main_binary_name().to_string();
  data.insert("main_binary_name", to_json(&main_binary_name));
```

PR #10977, opened 2024-09-12
feat: add `mainBinaryName` option

amrbashir:


crates/tauri-bundler/src/bundle/linux/debian.rs
```diff
@@ -133,8 +133,8 @@ pub fn generate_data(
 fn generate_changelog_file(settings: &Settings, data_dir: &Path) -> crate::Result<()> {
   if let Some(changelog_src_path) = &settings.deb().changelog {
     let mut src_file = File::open(changelog_src_path)?;
-    let bin_name = settings.main_binary_name();
```
lucasfernog, line 136: let's also document that the changelog path changed

crates/tauri-bundler/src/bundle/linux/debian.rs
```diff
@@ -306,7 +306,7 @@ fn generate_md5sums(control_dir: &Path, data_dir: &Path) -> crate::Result<()> {
 /// Copy the bundle's resource files into an appropriate directory under the
 /// `data_dir`.
 fn copy_resource_files(settings: &Settings, data_dir: &Path) -> crate::Result<()> {
-  let resource_dir = data_dir.join("usr/lib").join(settings.main_binary_name());
+  let resource_dir = data_dir.join("usr/lib").join(settings.product_name());
```
lucasfernog, line 309: and also document the resource path changed - this will cause new apps to duplicate the resources in the system

crates/tauri-cli/src/interface/rust.rs
```diff
@@ -912,12 +911,9 @@ impl AppSettings for RustAppSettings {
       for entry in fs_bins {
         let path = entry?.path();
         if let Some(name) = path.file_stem() {
-          let bin_exists = binaries.iter().any(|bin| {
-            bin.name() == name || path.ends_with(bin.src_path().unwrap_or(&"".to_string()))
```
lucasfernog, line 916: this check is here to fix the following issue:
- you have a `src-tauri/src/bin/app.rs` file (automatically discovered as a binary by Cargo)
- you configure the binary manually in Cargo.toml (yeah it's silly, but it's possible and allows you to configure more stuff)
```toml
[[bin]]
name = "lucass"
path = "src/bin/app.rs"
```

then the Tauri CLI will try to load both `app` and `lucass`, while only `lucass` exists

lucasfernog:
I also think our migration script should fill this value with the productName value, it'll prevent some issues if you do not ship the final v1 update before migrating to v2.

crates/tauri-bundler/src/bundle/linux/debian.rs
amrbashir: As far as I tested, at least `.deb` removes all bins and resources directory as well when installing a new version and then adds the new binaries and resources.

crates/tauri-bundler/src/bundle/linux/debian.rs
amrbashir: for tauri v1, the path is unchanged, it is only a breaking change for those on v2. Added a change file nevertheless

crates/tauri-bundler/src/bundle/linux/debian.rs
amrbashir: > and also document the resource path changed

for tauri v1, the path is unchanged, it is only a breaking change for those on v2. Added a change file nevertheless

crates/tauri-cli/src/interface/rust.rs
amrbashir: reverted

lucasfernog (approved):

PR #10978, opened 2024-09-12
refactor(core)!: rename plugin::PermissionState::Unknown to plugin::PermissionState::Prompt

lucasfernog:


amrbashir (dismissed):

crates/tauri/src/plugin.rs
```diff
@@ -923,7 +923,7 @@ impl std::fmt::Display for PermissionState {
       Self::Granted => write!(f, "granted"),
       Self::Denied => write!(f, "denied"),
       Self::PromptWithRationale => write!(f, "prompt-with-rationale"),
-      Self::Unknown => write!(f, "Unknown"),
+      Self::Unknown => write!(f, "prompt"),
```
amrbashir: shouldn't the variant be named `Prompt` then?

PR #10991, opened 2024-09-13
fix(core): `restart` cannot handle binary name change on macOS

lucasfernog:
Tauri v2 changed the default binary name of the bundled apps (no longer follows the productName, just uses the default name from Cargo instead). This breaks the `restart` function, which expects the current binary path to match the new one.

Due to this change, the restart() function for macOS is broken - the .app is correctly replaced even if the productName changed, but the restart() function cannot handle the new binary path. This change adds a simple check on macOS to read the `Contents/MacOS` folder and if it only contains a single binary, we use it to restart instead. This inference cannot be used if there's sidecars, so in this case we just let the existing implementation run and if it fails to restart, we do not panic but only warn and exit instead.

AppImage updates are not affected by this, and the Windows installer is responsible for restarting, so this change is only applied to macOS binaries.

### DIFF
--- a/core/tauri/src/api/process.rs
+++ b/core/tauri/src/api/process.rs
@@ -82,10 +82,49 @@ pub fn restart(env: &Env) {
   use std::process::{exit, Command};
 
   if let Ok(path) = current_binary(env) {
-    Command::new(path)
-      .args(&env.args)
-      .spawn()
-      .expect("application failed to start");
+    // on macOS on updates the binary name might have changed
+    // so we'll read the Contents/MacOS folder instead to infer the actual binary path
+    #[cfg(target_os = "macos")]
+    if let Some(parent) = path.parent() {
+      if parent.components().last()
+        == Some(std::path::Component::Normal(std::ffi::OsStr::new("MacOS")))
+      {
+        let macos_binaries = std::fs::read_dir(parent)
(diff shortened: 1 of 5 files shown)


lucasfernog:
Closes https://github.com/tauri-apps/plugins-workspace/issues/1746

core/tauri/src/api/process.rs
```diff
@@ -82,10 +82,49 @@ pub fn restart(env: &Env) {
   use std::process::{exit, Command};
 
   if let Ok(path) = current_binary(env) {
-    Command::new(path)
-      .args(&env.args)
-      .spawn()
-      .expect("application failed to start");
+    // on macOS on updates the binary name might have changed
+    // so we'll read the Contents/MacOS folder instead to infer the actual binary path
```
amrbashir: Can we read `Contents/Info.plist` to determine new binary name?

lucasfernog: ohhhhhh we can.. let's do that

lucasfernog: Amr the no-Apple dev that knows more about Apple than any other dev

amrbashir (dismissed):

amrbashir (approved):

PR #11005, opened 2024-09-15
fix(bundler): Use appimage files instead of debian files when building appimage

goenning:
I noticed the appimage bundler is using debian files instead of appimage


lucasfernog (approved):
oh nice catch! thank you :heart:

PR #11032, opened 2024-09-16
feat(core): add `app > windows > create` option to disable window creation at startup

### LINKED ISSUE #10950: [feat] Allow for windows to spawn manually
### Describe the problem

I'm frustrated that windows configured in `tauri.conf.json` spawn automatically during *setup* phase. The [`WindowConfig`](https://v2.tauri.app/reference/config/#windowconfig) allows to hide the window by setting `visible: false`, but that still spawns a hidden window and runs any code that is defined in the page loaded by the webview. In my use-case that is not acceptable, because the main window loads its content from the database and should only run if the database exists and has no pending migrations. Instead, the user should be shown a specific page (e.g., `/error`) that holds the error message.

This may seemingly be worked around by manually spawning a window during the *setup* phase, but this requires to pass any of the over 40 `WindowConfig` properties in rust code which worsens readability in my opinion, instead of keeping them in the configuration.

I would like to have a way of manually spawning windows from the [`WindowConfig`](https://v2.tauri.app/reference/config/#windowconfig). This would also require a new `WindowConfig` properties to prevent auto-spawning during the *setup* phase.

### Alternatives considered

- I considered spawning a window using `tauri::WebviewWindowBuilder::new(app, "foo", tauri::WebviewUrl::App("/foo".into())).build()?;`  
  This is not the best workaround as described above.
- I considered changing the url of the "main" window to the specific error page (e.g., `/error`), but was advised not to do that because it is possible that Tauri loads `/` and runs its code, and shows it to the user before I could then navigate to `/error` in the *setup* phase.

### Additional context

_No response_

amrbashir:
closes #10950

### DIFF
--- a/.changes/window-config-create.md
+++ b/.changes/window-config-create.md
@@ -1,0 +1,6 @@
+---
+"tauri": "minor:feat"
+"tauri-utils": "minor:feat"
+---
+
+Add `app > windows > create` option to choose whether to create this window at app startup or not.
--- a/.changes/window-config-create.md
+++ b/.changes/window-config-create.md
@@ -1,6 +1,6 @@
 ---
-"tauri": "minor:feat"
-"tauri-utils": "minor:feat"
+"tauri": "patch:feat"
+"tauri-utils": "patch:feat"
 ---
 
 Add `app > windows > create` option to choose whether to create this window at app startup or not.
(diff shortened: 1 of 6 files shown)


.changes/window-config-create.md
```diff
@@ -0,0 +1,6 @@
+---
+"tauri": "minor:feat"
+"tauri-utils": "minor:feat"
```
lucasfernog:
```suggestion
"tauri": "patch:feat"
"tauri-utils": "patch:feat"
```

lucasfernog (approved):

PR #11035, opened 2024-09-17
fix(cli): test timeout

lucasfernog:


FabianLars (approved):

lucasfernog:
15 min might still be tight if there's no cache, let's see...

PR #11037, opened 2024-09-17
feat(cli): Add transparent padding to macos icon

FabianLars:
opening as a draft cause idk whether it should be enabled by default and/or how to name the cli flag

edit: also the code may be a bit ugly still, i pushed once it was working to test it on my macbook

### DIFF
--- a/crates/tauri-cli/src/icon.rs
+++ b/crates/tauri-cli/src/icon.rs
@@ -58,6 +58,9 @@ pub struct Options {
   /// The background color of the iOS icon - string as defined in the W3C's CSS Color Module Level 4 <https://www.w3.org/TR/css-color-4/>.
   #[clap(long, default_value = "#fff")]
   ios_color: String,
+
+  #[clap(long)]
+  no_macos_padding: bool,
 }
 
 enum Source {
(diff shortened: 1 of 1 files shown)


thewh1teagle:
Hey
What's the status of the feature?
Can we add it? Or at least a flag to enable?
Thanks

FabianLars:
still the same as in my initial comment. idk if it should be default and how the cli flag should look like 🤷

gaomd:
Can we get this merged?

macOS icon requires special treatment because as per Apple's [guideline](https://developer.apple.com/design/human-interface-guidelines/app-icons/#macOS) and [icon templates](https://developer.apple.com/design/resources/#macos-apps), the actual content area of the macOS icon is the 824px * 824px out of the 1024px * 1024px. So when we only have 1 single icon, the [`tauri icon`](https://v2.tauri.app/reference/cli/#icon) cli should automatically add the transparent padding around the icon to ensure consistency with other macOS icons.

EDIT: however the fix in the PR seems only adding the transparent padding so the icon size looks consistent with other macOS apps, but does not add shadow which is an essential part of the macOS icons, hence making the solution less optimal. Another approach might be to add an CLI option to allow specifying the macOS-only icon and only generate `src-tauri/icons/icon.icns`, and document macOS icon specs in the Tauri docs.

Screenshot 2025-02-03 at 00 26 54

PR #11039, opened 2024-09-17
feat: add `upgradeCode` config option

amrbashir:


crates/tauri-bundler/src/bundle/settings.rs
```diff
@@ -351,6 +351,15 @@ impl Default for WixLanguage {
 /// Settings specific to the WiX implementation.
 #[derive(Clone, Debug, Default)]
 pub struct WixSettings {
+  /// A GUID upgrade code for MSI installer. This code **_must stay the same across all of your updates_**,
+  /// otherwise, Windows will treat your update as a different app and your users will have duplicate versions of your app.
+  ///
+  /// By default, tauri generates this code by generating a Uuid v5 using the string `<productName>.exe.app.x64` in the DNS namespace.
+  /// You can use Tauri's CLI to generate and print this code for you, run `tauri generate-wix-upgrade-code`.
```
lucasfernog:
```suggestion
  /// You can use Tauri's CLI to generate and print this code for you by running `tauri generate-wix-upgrade-code`.
```

lucasfernog (approved):
i moved the `generate-wix-upgrade-code` to `inspect wix-upgrade-code` since we're not really generating a new guid, but only rendering the derived value; it also now uses the same logic as the bundler (fallback to package > name from Cargo.toml as product name)

PR #11041, opened 2024-09-18
refactor(bundler/nsis): replace `nsis_process.dll` usage with `nsProcess.dll`

### LINKED ISSUE #37: Machines with Webroot seem to be detecting nsis_tauri_utils.dll as a virus as of very recently
We have had a Tauri-based application in production with a few hundred users for a while, but today started to get reports of Webroot picking up the nsis_tauri_utils.dll as malware. It is preventing installation of our app. 

Here is a VirusTotal scan of the file reported to use by one of our users: https://www.virustotal.com/gui/file/c0f2b928bc4c81cc5ca30a8932a6dc8cd617dd016679c057e23355fe732b2333?nocache=1

I was informed that the issue is fixed via allowlisting in Webroot, but obviously that is not ideal. 

Just wanted to add this issue here in case anyone has any thoughts, feedback, or has experienced the same thing. 

See screenshot of error during install:
Screenshot 2024-07-08 at 16 54 46

### LINKED ISSUE #39: When using KillProcess or KillProcessCurrentUser to close the tauri application during uninstallation, the tauri application cannot perform pre exit processing
KillProcess or KillProcessCurrentUser will call TerminateProcess, which will force the process to terminate and the process will not receive any signals.
My Tauri application needs to perform some cleaning and closing child processes before terminate, which can result in orphan processes and occupy files in the application directory, leading to uninstallation and reinstallation exceptions.
I tried to change TerminateProcess to GenerateConsoleCtrlEvent and replace nsis-tauri-utils.dll, but Tauri detects the hash value of the file and corrects it during build, so I cannot confirm if GenerateConsoleCtrlEvent will work.

amrbashir:
`nsis_process` dll is written by us in https://github.com/tauri-apps/nsis-tauri-utils but anti virus tools picks it up as virus. Its code does as much as `nsProcess` but I guss we just don't have enough popularity as `nsProcess` has with anti-virus tools.

closes https://github.com/tauri-apps/nsis-tauri-utils/issues/39
closes: https://github.com/tauri-apps/nsis-tauri-utils/issues/37

### DIFF
--- a/crates/tauri-bundler/src/bundle/windows/templates/GetProcessInfo.nsh
+++ b/crates/tauri-bundler/src/bundle/windows/templates/GetProcessInfo.nsh
@@ -1,3 +1,5 @@
+; https://nsis.sourceforge.io/Get_process_info
+;
 ; NSIS PROCESS INFO LIBRARY - GetProcessInfo.nsh
 ; Version 1.1 - Mar 28th, 2011
 ;
--- a/crates/tauri-bundler/src/bundle/windows/templates/GetProcessInfo.nsh
+++ b/crates/tauri-bundler/src/bundle/windows/templates/GetProcessInfo.nsh
@@ -1,0 +1,153 @@
+; NSIS PROCESS INFO LIBRARY - GetProcessInfo.nsh
+; Version 1.1 - Mar 28th, 2011
+;
+; Description:
+;   Gets process information.
+;
+; Usage example:
+;   ${GetProcessInfo} 0 $0 $1 $2 $3 $4
+;   DetailPrint "pid=$0 parent_pid=$1 priority=$2 process_name=$3 exe=$4"
+;
+; History:
+;   1.1 - 28/03/2011 - Added uninstall function, include guards, file header. Fixed getting full exe path on pre vista systems. (Sergius)
+;
+
+!ifndef GETPROCESSINFO_INCLUDED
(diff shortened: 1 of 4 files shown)


crates/tauri-bundler/src/bundle/windows/templates/GetProcessInfo.nsh
```diff
@@ -0,0 +1,153 @@
+; NSIS PROCESS INFO LIBRARY - GetProcessInfo.nsh
```
amrbashir, line 3:
```suggestion
; https://nsis.sourceforge.io/Get_process_info
;
; NSIS PROCESS INFO LIBRARY - GetProcessInfo.nsh
```

Legend-Master:
I don't like this to be honest, if this is a webroot only problem, can we contact them to white list us instead?

amrbashir:
I don't like it either but I am not sure how many more anti virus would have false positives. 

Also weirdly enough, nsis-tauri-utils@0.4.1 passes all anti viruses https://www.virustotal.com/gui/url/e814a313d4ba77fd51c3d1408c3ba49329a52656a767df50bdcbb867bcf837f6?nocache=1

amrbashir:
Converting to a draft for now

amrbashir:
should be fixed with https://github.com/tauri-apps/tauri/pull/11676 as well

PR #11056, opened 2024-09-18
feat: Expose ability to enable browser extensions in WebView2

### LINKED ISSUE #10909: [feat] Expose the ability to enable `are_browser_extensions_enabled` from `CoreWebView2EnvironmentOptions`
### Describe the problem

While I understand this would have to be Windows-only from the get-go (and a somewhat niche use-case), I'd like to not only load a custom extension (for use in overwriting request headers, something that can otherwise not be done in Tauri/Wry using external URLs), but also give my users the ability to load browser extensions as well.

To do so requires setting `CoreWebView2EnvironmentOptions.set_are_browser_extensions_enabled(true)` as well as calling `AddBrowserExtension` on the `ICoreWebview2Profile7`. While I can access the `ICoreWebView2Profile7` and call that function (via `WebviewWindow.with_webview` and some casting), after some digging, I couldn't find a way to set this environment option to begin with.

I haven't totally thought of the best way to go about this honestly, since it's as platform-specific as you can get, and I don't have a perfect understanding of the Wry & Tauri codebases.

I recently looked at [this Wry PR (#1344)](https://github.com/tauri-apps/wry/pull/1344/files) that implemented the ability to change the scrollbar style for Windows in Wry, but as far as I'm aware, this functionality isn't exposed in Tauri either, so I wouldn't know what the plan is/would be to do so (if there even is any). Assuming that does get exposed, the same methodology would probably apply to this as well.

### Alternatives considered

Like most problems, I could probably solve it by just forking and maintaining a separate Wry and Tauri, but that sounds like a huge pain lol.

### Additional context

_No response_

SpikeHD:
Addresses tauri-apps/tauri#10909.

* Updates `wry` to `0.44.0`, which introduces `browser_extensions_enabled` for Windows (from https://github.com/tauri-apps/wry/pull/1356)
* Updates `WebviewWindowBuilder` and `WindowConfig` structs to include that same configuration
* Adds `browser_extensions_enabled(bool)` to `WebviewWindowBuilder`

### DIFF
--- a/.changes/change-pr-11056.md
+++ b/.changes/change-pr-11056.md
@@ -1,6 +1,5 @@
 ---
 "tauri": patch:feat
-"tauri-cli": patch:feat
 "tauri-runtime-wry": patch:feat
 "tauri-runtime": patch:feat
 "tauri-utils": patch:feat
--- a/.changes/change-pr-11056.md
+++ b/.changes/change-pr-11056.md
@@ -1,0 +1,9 @@
+---
+"tauri": patch:feat
+"tauri-cli": patch:feat
+"tauri-runtime-wry": patch:feat
+"tauri-runtime": patch:feat
+"tauri-utils": patch:feat
+---
+
+Expose the ability to enabled browser extensions in WebView2 on Windows.
--- a/crates/tauri/src/webview/plugin.rs
+++ b/crates/tauri/src/webview/plugin.rs
@@ -42,8 +42,6 @@ mod desktop_commands {
     incognito: bool,
     #[serde(default)]
     zoom_hotkeys_enabled: bool,
-    #[serde(default)]
-    browser_extensions_enabled: bool,
   }
 
   #[derive(Serialize)]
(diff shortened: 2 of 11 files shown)


crates/tauri/src/webview/plugin.rs
```diff
@@ -42,6 +42,8 @@ mod desktop_commands {
     incognito: bool,
     #[serde(default)]
     zoom_hotkeys_enabled: bool,
+    #[serde(default)]
+    browser_extensions_enabled: bool,
```
amrbashir: Since JS is not a secure context, this value shouldn't be set from JS. cc @tweidinger

.changes/change-pr-11056.md
```diff
@@ -0,0 +1,9 @@
+---
+"tauri": patch:feat
+"tauri-cli": patch:feat
```
amrbashir:
```suggestion
```

amrbashir (approved):

amrbashir:
Thank you

PR #11066, opened 2024-09-20
feat: add `webview.clear_all_browsing_data`

### LINKED ISSUE #6567: [feat] API to clear webview-managed data
### Describe the problem

My application uses cookies/localStorage on the javascript side for various purposes.

I don't see a clear way to clear all the stuff stored by the Webview engine. I could have the javascript side clear these, but there are things that this side can't clear like disk cache, etc

I want to put a native button like 'Logout' that clears all this data. 

An API on app/app_handle that clears this data might make sense. Maybe it needs to be part of exit or relaunch?

### Alternatives considered

I considered manually deleting some of the path_resolver() directories, but directories like `/Users/pttamkin/Library/HTTPStorage/{pkg}` and `/Users/pttamkin/Library/WebKit/{pkg}` aren't captured here. Also not sure I can delete these while my app is running?

### Additional context

_No response_

amrbashir:
closes #6567


lucasfernog (approved):
lgtm but needs https://github.com/tauri-apps/wry/pull/1362

PR #11071, opened 2024-09-20
feat(core): add `Manager::unmanage`

### LINKED ISSUE #10897: [feature] Un-manage state
Hello. I'd like to have an operation opposite to managing state. It is said that state storage in Tauri behaves like a hashmap, could it be possible to have something like removing entries from this hashmap? Something in the likes of `fn unmanage<T>(&self) -> bool`? So that `app.try_state::<T>()` would return None? Thanks.

amrbashir:
closes #10897

### DIFF
--- a/crates/tauri/src/lib.rs
+++ b/crates/tauri/src/lib.rs
@@ -699,7 +699,7 @@ pub trait Manager<R: Runtime>: sealed::ManagerBase<R> {
     self.manager().state().set(state)
   }
 
-  /// Removes the state managed by the application for T.  Returns `true` if it actuall removed
+  /// Removes the state managed by the application for T.  Returns `true` if it was actually removed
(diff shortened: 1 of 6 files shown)


amrbashir:
will port over the tests later

crates/tauri/src/lib.rs
```diff
@@ -699,6 +699,14 @@ pub trait Manager<R: Runtime>: sealed::ManagerBase<R> {
     self.manager().state().set(state)
   }
 
+  /// Removes the state managed by the application for T.  Returns `true` if it actuall removed
```
lucasfernog:
```suggestion
  /// Removes the state managed by the application for T.  Returns `true` if it was actually removed
```

lucasfernog (approved):

PR #11096, opened 2024-09-23
feat(bundler): improve GitHub mirror URL generation with custom templates

### LINKED ISSUE #10497: [feat] provide mirror option for download nsis and wix3
### Describe the problem

github.com is blocked on our build servers. We managed to work around the nsis and wix download by providing our own mirror but there is no mirror option for them. It will be helpful if a similar option can be exposed for the same.

just like [ELECTRON_BUILDER_BINARIES_MIRROR](https://github.com/electron-userland/electron-builder/blob/f3404d32f029dfb63e83067012f15ef67838f326/packages/app-builder-lib/src/binDownload.ts#L21-L25)

```powershell
$env:TAURI_BUNDLER_NSIS_MIRROR="https://mirror.example.com/nsis.zip"
$env:TAURI_BUNDLER_NSIS_UTILS_MIRROR="https://mirror.example.com/nsis-utils.dll"
$env:TAURI_BUNDLER_WIX_MIRROR="https://mirror.example.com/wixxxx.zip"
```

### Alternatives considered

_No response_

### Additional context

_No response_

thep0y:
- Add support for custom GitHub mirror URL templates via `TAURI_BUNDLER_TOOLS_GITHUB_MIRROR_TEMPLATE` env var
- Refactor URL generation logic into separate functions for better modularity
- Implement regex-based parsing of GitHub URLs to extract components
- Add unit tests for new URL generation functionality
- Remove direct dependency on `ureq::AgentBuilder` in favor of more flexible approach
- Update `create_agent_and_url` function to use new URL generation methods

Closes: https://github.com/tauri-apps/tauri/issues/10497


crates/tauri-bundler/src/bundle/windows/util.rs
```diff
@@ -69,18 +69,58 @@ pub fn download_webview2_offline_installer(base_path: &Path, arch: &str) -> crat
...
   }
+
+  let mirror_url = generate_mirror_url_from_template(url);
+  let cdn_url = generate_mirror_url_from_base(url)?;
+
+  Ok(
+    mirror_url
+      .or(cdn_url)
+      .map(|alternative_url| (ureq::AgentBuilder::new().build(), alternative_url)),
+  )
```
amrbashir: This can be simplified into

```suggestion
  generate_mirror_url_from_template(url)
    .or_else(|_| generate_mirror_url_from_base(url))
    .map(|alt_url| {
      alt_url.map(|alt_url| (ureq::AgentBuilder::new().build(), alt_url))
    })
```

also notice usage of `or_else` instead of `or` which will lazily evaluate `generate_mirror_url_from_base` only if needed.

crates/tauri-bundler/src/bundle/windows/util.rs
```diff
@@ -69,18 +69,58 @@ pub fn download_webview2_offline_installer(base_path: &Path, arch: &str) -> crat
...
+fn create_agent_and_url(url: &str) -> crate::Result<(ureq::Agent, String)> {
+  generate_alternative_url(url)?.map_or_else(
+    || {
+      Ok((
+        ureq::AgentBuilder::new().try_proxy_from_env(true).build(),
+        url.to_owned(),
+      ))
+    },
+    Ok,
+  )
```
amrbashir: I generally dislike `map_or_else` since it is a badly named function and can be confusing at first glance, I would re-write this function without it
```suggestion
  generate_alternative_url(url).map(|alt_url| {
    alt_url.unwrap_or_else(|| (
        ureq::AgentBuilder::new().try_proxy_from_env(true).build(),
        url.to_owned(),
      ))
  })
```

crates/tauri-bundler/src/bundle/windows/util.rs
```diff
@@ -196,3 +236,59 @@ pub fn os_bitness<'a>() -> Option<&'a str> {
     _ => None,
   }
 }
+
+#[cfg(test)]
+mod tests {
+  use super::generate_mirror_url_from_template;
+  use std::env;
+
+  const URL: &str =
```
amrbashir: Let's call this `GOOD_URL`
```suggestion
  const GOOD_URL: &str =
```
and add another `BAD_URL` const next to it

amrbashir (changes requested):

crates/tauri-bundler/src/bundle/windows/util.rs
thep0y: `URL` is just a link to the source file that needs to be downloaded. I don't understand the point of changing it to `GOOD_URL`; this variable name has no relation to its function.

crates/tauri-bundler/src/bundle/windows/util.rs
thep0y: This change is just a matter of personal coding preference and does not affect its functionality; I have no objection to this modification.

crates/tauri-bundler/src/bundle/windows/util.rs
amrbashir: it does in the context of test, from a first glance, it is not clear why this const is defined at all but in contrast, `GOOD_URL` and `BAD_URL` clearly states that first is for the good test case where we produce a new URL based on the template and the second for the bad case where we return None.

crates/tauri-bundler/src/bundle/windows/util.rs
amrbashir: correct, it is just a coding preference, think of it as a nit-pick but for consistency of the codebase this is a slightly important nit-pick.

crates/tauri-bundler/src/bundle/windows/util.rs
thep0y: I still can't understand the purpose of naming it this way. The `URL` variable is just a resource address. If clarity in naming is needed, I think `GITHUB_ASSET_URL` would be more appropriate. I don't know where `GOOD_URL` and `BAD_URL` should be used.

amrbashir: `URL` is only used in tests right? there is `test_generate_mirror_url_correctly1` and `test_generate_mirror_url_correctly2` that test whether we successfully got a URL from the template, so I'd consider these the "Good" cases. 

There is also `test_generate_mirror_url_no_env_var` and `test_generate_mirror_url_non_github_url` that cover cases when there is no env var or we are not downloading from github, in which I'd consider these the bad cases.

Maybe good and bad are not good names here so indeed `GITHUB_ASSET_URL` and `NON_GITHUB_ASSET_URL` (which is defined inline in `test_generate_mirror_url_non_github_url` test case) are better names.

crates/tauri-bundler/src/bundle/windows/util.rs
```diff
@@ -69,18 +69,52 @@ pub fn download_webview2_offline_installer(base_path: &Path, arch: &str) -> crat
...
+        template
+          .replace("<owner>", &caps[1])
+          .replace("<repo>", &caps[2])
+          .replace("<version>", &caps[3])
+          .replace("<asset>", &caps[4])
+      })
+    })
+}
+
+fn generate_mirror_url_from_base(github_url: &str) -> crate::Result<Option<String>> {
```
amrbashir: I think it is fine to just return an `Option<String>` here

crates/tauri-bundler/src/bundle/windows/util.rs
```diff
@@ -69,18 +69,52 @@ pub fn download_webview2_offline_installer(base_path: &Path, arch: &str) -> crat
...
+}
+
+fn generate_alternative_url(url: &str) -> crate::Result<Option<(ureq::Agent, String)>> {
+  if !url.starts_with("https://github.com/") {
+    return Ok(None);
   }
+
+  generate_mirror_url_from_template(url)
+    .or_else(|_| generate_mirror_url_from_base(url))
+    .map(|alt_url| alt_url.map(|alt_url| (ureq::AgentBuilder::new().build(), alt_url)))
```
amrbashir: looks like I provided you with wrong code, I thought `generate_mirror_url_from_template` returned a Result.

With the change requested above, the correct code should be 
```suggestion
  generate_mirror_url_from_template(url)
    .or_else(|| generate_mirror_url_from_base(url))
    .map(|alt_url| (ureq::AgentBuilder::new().build(), alt_url))
```

amrbashir (changes requested):

crates/tauri-bundler/src/bundle/windows/util.rs
```diff
@@ -69,18 +69,52 @@ pub fn download_webview2_offline_installer(base_path: &Path, arch: &str) -> crat
...
-      AgentBuilder::new().try_proxy_from_env(true).build(),
-      url.to_owned(),
-    )),
+      parsed_cdn.set_path(github_url);
+      Ok(parsed_cdn.into())
+    })
+    .transpose()
+}
+
+fn generate_alternative_url(url: &str) -> crate::Result<Option<(ureq::Agent, String)>> {
```
amrbashir: With the change requested above, this can also be changed to just return an `Option<(ureq::Agent, Stirng)>`

crates/tauri-bundler/src/bundle/windows/util.rs
```diff
@@ -69,24 +69,53 @@ pub fn download_webview2_offline_installer(base_path: &Path, arch: &str) -> crat
...
-    Ok(cdn) if url.starts_with("https://github.com/") => {
-      let mut parsed_cdn = Url::parse(&cdn)?;
-      parsed_cdn.set_path(url);
-      Ok((AgentBuilder::new().build(), parsed_cdn.into()))
-    }
-    _ => Ok((
-      AgentBuilder::new().try_proxy_from_env(true).build(),
-      url.to_owned(),
-    )),
+fn generate_mirror_url_from_template(github_url: &str) -> Option<String> {
```
lucasfernog: I would name this to `generate_github_mirror_url_from_template` to be explicit that the URL must be a github URL

crates/tauri-cli/ENVIRONMENT_VARIABLES.md
```diff
@@ -15,6 +15,7 @@ These environment variables are inputs to the CLI which may have an equivalent C
 - `TAURI_CLI_NO_DEV_SERVER_WAIT` — Skip waiting for the frontend dev server to start before building the tauri application.
 - `TAURI_LINUX_AYATANA_APPINDICATOR` — Set this var to `true` or `1` to force usage of `libayatana-appindicator` for system tray on Linux.
 - `TAURI_BUNDLER_WIX_FIPS_COMPLIANT` — Specify the bundler's WiX `FipsCompliant` option.
+- `TAURI_BUNDLER_TOOLS_GITHUB_MIRROR_TEMPLATE` - Specify a GitHub mirror template to download files and tools used by tauri bundler.
 - `TAURI_BUNDLER_TOOLS_GITHUB_MIRROR` - Specify a GitHub mirror to download files and tools used by tauri bundler.
```
amrbashir:
```suggestion
- `TAURI_BUNDLER_TOOLS_GITHUB_MIRROR` - Specify a GitHub mirror to download files and tools used by tauri bundler.
- `TAURI_BUNDLER_TOOLS_GITHUB_MIRROR_TEMPLATE` - Specify a GitHub mirror template to download files and tools used by tauri bundler, for example: `https://mirror.example.com/<owner>/<repo>/releases/download/<version>/<asset>`.
```

amrbashir (approved):

PR #11098, opened 2024-09-23
refactor(core&cli): activate default features in build script instead of CLI

amrbashir:


crates/tauri/build.rs
```diff
@@ -236,6 +236,27 @@ fn main() {
...
+  {
+    if mobile {
+      println!("cargo:rustc-cfg=feature=\"rustls-tls\"");
+    } else {
+      println!("cargo:rustc-cfg=feature=\"native-tls\"");
+    }
+  }
+
+  // a workaround for default features activation based on profile
+  if has_default && env::var("PROFILE").as_deref() == Ok("release") {
```
lucasfernog: this won't work if you use `tauri build --debug`

amrbashir: we can keep this here and inject `tauri/custom-protocol` in debug from CLI? this way `cargo build` and `cargo build --release` work out of the box and match `tauri dev` and `tauri build` respectively. `tauri build --debug` should be a special case that is fine to keep in CLI.

lucasfernog: yeah that would be fine

crates/tauri-cli/src/interface/rust/desktop.rs
```diff
@@ -304,16 +305,26 @@ fn build_command(
...
   let mut features = config_features;
   if let Some(f) = options.features {
     features.extend(f);
   }
+  // custom-protocol is handled in tauri's build.rs script
+  // which makes `cargo build` same as `tauri dev`
+  // and `cargo build --release` same as `tauri build`
+  // but for `tauri build --debug` we need to build the code
+  // in `debug` profile and activate `custom-protocol` feature.
+  if !dev && options.debug && !has_custom_profile {
```
lucasfernog:
```suggestion
  if !dev {
```
if the custom profile is a debug one we still want to force the feature flag

amrbashir: [`3de0343` (#11098)](https://github.com/tauri-apps/tauri/pull/11098/commits/3de0343d5770fea0ecf61d3dc985eb36e6d63503)

lucasfernog (commented):
i'm not sure if we can actually use this..
from the docs (https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-cfg):
> Note that this does not affect Cargo’s dependency resolution. This cannot be used to enable an optional dependency, or enable other Cargo features.

and the TLS features all must enable optional dependencies from reqwest...

FabianLars:
ngl this feels super over engineered to me anyway. I don't think the old way was that bad, we should just mention the feature flag somewhere

amrbashir:
:sadge: I guess we will have to use custom logic in tauri-cli for now

PR #11104, opened 2024-09-24
fix(api): convert `Position` to `PhsyicalPosition` for `TrayIconEvent`

amrbashir:
ref: https://github.com/tauri-apps/plugins-workspace/pull/1822#issuecomment-2365442240


jbolda:
This seems to be considered a bug fix, but I imagine it is likely many dependent use cases have "updated" to this instead. Something to be mindful of, as we have noted in the ref.

.changes/api-tray-icon-event-value-mismatch-type.md
```diff
@@ -0,0 +1,6 @@
+---
+"@tauri-apps/api": "patch:bug"
+---
+
+Fix inavlid value passed to the callback `action` function that doesn't match the `TrayIconEvent` type.
```
lucasfernog:
```suggestion
Fix invalid value passed to the callback `action` function that doesn't match the `TrayIconEvent` type.
```

lucasfernog (approved):

PR #11105, opened 2024-09-24
refactor(core): return state in `unmanage`

amrbashir:


crates/tauri/src/lib.rs
```diff
@@ -700,11 +700,11 @@ pub trait Manager<R: Runtime>: sealed::ManagerBase<R> {
   }
 
   /// Removes the state managed by the application for T.  Returns `true` if it was actually removed
```
FabianLars:
```suggestion
  /// Removes the state managed by the application for T. Returns the state if it was actually removed.
```

lucasfernog (approved):

PR #11121, opened 2024-09-25
refactor!: simplify `TrayIconEvent` in JS by tagging it with `type` field

amrbashir:
ref: https://discord.com/channels/616186924390023171/986184094050316358/1287961960146669659 and https://github.com/tauri-apps/plugins-workspace/pull/1822#issuecomment-2365442240


packages/api/src/tray.ts
```diff
@@ -99,12 +24,30 @@ export interface TrayIconLeaveEvent {
...
+      /** Mouse button that triggered this event. */
+      button: MouseButton
+    }
+  : never) &
+  (T extends 'Click'
+    ? {
+        /** Mouse button state when this event was triggered. */
+        buttonState: MouseButtonState
+      }
+    : never)
```
amrbashir: still not sure how to make these show error in IDE when narrowing type, basically 
```ts
if (event.type === "Leave") {
  e.id; // should work
  e.button; // should error
  e.buttonState; // shuold error
}
```

Legend-Master, line 50:
```suggestion
export type BaseTrayIconEvent<T extends TrayIconEventType> = {
  /** The tray icon event type */
  type: T
  /** Id of the tray icon which triggered this event. */
  id: string
  /** Physical position of the click the triggered this event. */
  position: PhysicalPosition
  /** Position and size of the tray icon. */
  rect: {
    position: PhysicalPosition
    size: PhysicalSize
  }
}

export type TrayIconClickEvent = {
  /** Mouse button that triggered this event. */
  button: MouseButton
}

/**
 * Describes a tray icon event.
 *
 * #### Platform-specific:
 *
 * - **Linux**: Unsupported. The event is not emitted even though the icon is shown,
 * the icon will still show a context menu on right click.
 */
export type TrayIconEvent =
  | (BaseTrayIconEvent<'Click'> &
      TrayIconClickEvent & {
        /** Mouse button state when this event was triggered. */
        buttonState: MouseButtonState
      })
  | (BaseTrayIconEvent<'DoubleClick'> & TrayIconClickEvent)
  | BaseTrayIconEvent<'Enter'>
  | BaseTrayIconEvent<'Move'>
  | BaseTrayIconEvent<'Leave'>
```

Legend-Master: Try this

```ts
export type BaseTrayIconEvent<T extends TrayIconEventType> = {
  /** The tray icon event type */
  type: T
  /** Id of the tray icon which triggered this event. */
  id: string
  /** Physical position of the click the triggered this event. */
  position: PhysicalPosition
  /** Position and size of the tray icon. */
  rect: {
    position: PhysicalPosition
    size: PhysicalSize
  }
}

export type TrayIconClickEvent = {
  /** Mouse button that triggered this event. */
  button: MouseButton
}

/**
 * Describes a tray icon event.
 *
 * #### Platform-specific:
 *
 * - **Linux**: Unsupported. The event is not emitted even though the icon is shown,
 * the icon will still show a context menu on right click.
 */
export type TrayIconEvent =
  | (BaseTrayIconEvent<'Click'> &
      TrayIconClickEvent & {
        /** Mouse button state when this event was triggered. */
        buttonState: MouseButtonState
      })
  | (BaseTrayIconEvent<'DoubleClick'> & TrayIconClickEvent)
  | BaseTrayIconEvent<'Enter'>
  | BaseTrayIconEvent<'Move'>
  | BaseTrayIconEvent<'Leave'>
```

Legend-Master:
```diff
diff --git a/packages/api/src/tray.ts b/packages/api/src/tray.ts
index 45820fe4d..b662fe8ab 100644
--- a/packages/api/src/tray.ts
+++ b/packages/api/src/tray.ts
@@ -16,15 +16,7 @@ export type TrayIconEventType =
   | 'Move'
   | 'Leave'
 
-/**
- * Describes a tray icon event.
- *
- * #### Platform-specific:
- *
- * - **Linux**: Unsupported. The event is not emitted even though the icon is shown,
- * the icon will still show a context menu on right click.
- */
-export type TrayIconEvent<T extends TrayIconEventType = TrayIconEventType> = {
+export type BaseTrayIconEvent<T extends TrayIconEventType> = {
   /** The tray icon event type */
   type: T
   /** Id of the tray icon which triggered this event. */
@@ -36,18 +28,52 @@ export type TrayIconEvent<T extends TrayIconEventType = TrayIconEventType> = {
     position: PhysicalPosition
     size: PhysicalSize
   }
-} & (T extends 'Click' | 'DoubleClick'
-  ? {
-      /** Mouse button that triggered this event. */
-      button: MouseButton
-    }
-  : never) &
-  (T extends 'Click'
-    ? {
+}
+
+export type TrayIconClickEvent = {
+  /** Mouse button that triggered this event. */
+  button: MouseButton
+}
+
+/**
+ * Describes a tray icon event.
+ *
+ * #### Platform-specific:
+ *
+ * - **Linux**: Unsupported. The event is not emitted even though the icon is shown,
+ * the icon will still show a context menu on right click.
+ */
+export type TrayIconEvent =
+  | (BaseTrayIconEvent<'Click'> &
+      TrayIconClickEvent & {
         /** Mouse button state when this event was triggered. */
         buttonState: MouseButtonState
+      })
+  | (BaseTrayIconEvent<'DoubleClick'> & TrayIconClickEvent)
+  | BaseTrayIconEvent<'Enter'>
+  | BaseTrayIconEvent<'Move'>
+  | BaseTrayIconEvent<'Leave'>
+
+type TrayIconEventInternal = Omit<TrayIconEvent, 'rect' | 'position'> & {
+  position: {
+    x: number
+    y: number
+  }
+  rect: {
+    position: {
+      Physical: {
+        x: number
+        y: number
       }
-    : never)
+    }
+    size: {
+      Physical: {
+        width: number
+        height: number
+      }
+    }
+  }
+}
 
 /**
  * Tray icon types and utilities.
@@ -166,37 +192,24 @@ export class TrayIcon extends Resource {
       options.icon = transformImage(options.icon)
     }
 
-    const handler = new Channel<TrayIconEvent>()
+    const handler = new Channel<TrayIconEventInternal>()
     if (options?.action) {
       const action = options.action
       handler.onmessage = (e) => {
-        if (
-          'Physical' in e.rect.position &&
-          e.rect.position.Physical &&
-          typeof e.rect.position.Physical === 'object' &&
-          'x' in e.rect.position.Physical &&
-          'y' in e.rect.position.Physical
-        ) {
-          e.rect.position = new PhysicalPosition(
-            e.rect.position.Physical.x as number,
-            e.rect.position.Physical.y as number
-          )
-        }
-        if (
-          'Physical' in e.rect.size &&
-          e.rect.size.Physical &&
-          typeof e.rect.size.Physical === 'object' &&
-          'width' in e.rect.size.Physical &&
-          'height' in e.rect.size.Physical
-        ) {
-          e.rect.size = new PhysicalSize(
-            e.rect.size.Physical.width as number,
-            e.rect.size.Physical.height as number
+        // e.position = new PhysicalPosition(e.position.x, e.position.y)
+        const out = e as unknown as TrayIconEvent
+        out.position = new PhysicalPosition(e.position.x, e.position.y)
+        out.rect = {
+          position: new PhysicalPosition(
+            e.rect.position.Physical.x,
+            e.rect.position.Physical.y
+          ),
+          size: new PhysicalSize(
+            e.rect.size.Physical.width,
+            e.rect.size.Physical.height
           )
         }
-        e.position = new PhysicalPosition(e.position.x, e.position.y)
-
-        action(e)
+        action(out)
       }
       delete options.action
     }
```

@amrbashir I think something like this would be better

packages/api/src/tray.ts
amrbashir: yeah I've tried that and it is still not narrowing correctly

amrbashir: oh nvm, I had a typo, thanks for help

Legend-Master: hmm, it works on my end

image
image

amrbashir: yeah it works now, I had `TrayIconEventBase<'Click'>` twice instead of `TrayIconEventBase<'Click'>` and another `TrayIconEventBase<'DoubleClick'>`

Legend-Master:
If you like the idea of typing the internal event as well, here's the updated diff

```diff
diff --git a/packages/api/src/tray.ts b/packages/api/src/tray.ts
index 040117a09..118928dd2 100644
--- a/packages/api/src/tray.ts
+++ b/packages/api/src/tray.ts
@@ -52,6 +52,27 @@ export type TrayIconEvent =
   | TrayIconEventBase<'Move'>
   | TrayIconEventBase<'Leave'>
 
+type TrayIconEventInternal = Omit<TrayIconEvent, 'rect' | 'position'> & {
+  position: {
+    x: number
+    y: number
+  }
+  rect: {
+    position: {
+      Physical: {
+        x: number
+        y: number
+      }
+    }
+    size: {
+      Physical: {
+        width: number
+        height: number
+      }
+    }
+  }
+}
+
 /**
  * Tray icon types and utilities.
  *
@@ -169,37 +190,23 @@ export class TrayIcon extends Resource {
       options.icon = transformImage(options.icon)
     }
 
-    const handler = new Channel<TrayIconEvent>()
+    const handler = new Channel<TrayIconEventInternal>()
     if (options?.action) {
       const action = options.action
       handler.onmessage = (e) => {
-        if (
-          'Physical' in e.rect.position &&
-          e.rect.position.Physical &&
-          typeof e.rect.position.Physical === 'object' &&
-          'x' in e.rect.position.Physical &&
-          'y' in e.rect.position.Physical
-        ) {
-          e.rect.position = new PhysicalPosition(
-            e.rect.position.Physical.x as number,
-            e.rect.position.Physical.y as number
+        const out = e as unknown as TrayIconEvent
+        out.position = new PhysicalPosition(e.position.x, e.position.y)
+        out.rect = {
+          position: new PhysicalPosition(
+            e.rect.position.Physical.x,
+            e.rect.position.Physical.y
+          ),
+          size: new PhysicalSize(
+            e.rect.size.Physical.width,
+            e.rect.size.Physical.height
           )
         }
-        if (
-          'Physical' in e.rect.size &&
-          e.rect.size.Physical &&
-          typeof e.rect.size.Physical === 'object' &&
-          'width' in e.rect.size.Physical &&
-          'height' in e.rect.size.Physical
-        ) {
-          e.rect.size = new PhysicalSize(
-            e.rect.size.Physical.width as number,
-            e.rect.size.Physical.height as number
-          )
-        }
-        e.position = new PhysicalPosition(e.position.x, e.position.y)
-
-        action(e)
+        action(out)
       }
       delete options.action
     }
```

amrbashir:
narrowing using JS checks is better here IMO since it ensures we never construct invalid `PhysicalPosition` values when we don't receive any from Rust. I think I might need to add checks for `Logical` as well

Legend-Master:
Currently it's guaranteed to be `PhysicalPosition` but it might change depending on the tray-icon side, maybe we type two variants, and then do the conversion, the current way of checking is quite hard to understand and error prone to be honest

Legend-Master:
I honestly feel like it's probably easier to use a custom `Rect` or tray-icon's `Rect` in the rust side than doing it in the js side

amrbashir:
> Currently it's guaranteed to be `PhysicalPosition` but it might change depending on the tray-icon side, maybe we type two variants, and then do the conversion, the current way of checking is quite hard to understand and error prone to be honest

I will add checks for `Logical` variants which should cover the case if the position type changes

As for the confusing part, I don't think there is a way around that, we still need to check whether `Physical` or `Logical` property exist or not, I will add some comments to explain why we doing this 

> I honestly feel like it's probably easier to use a custom `Rect` or tray-icon's `Rect` in the rust side than doing it in the js side

We need to create an instance of `PhysicalPosition` or `LogicalPosition` so we still need to do checks and create the instances on the JS side

amrbashir:
I have added checks for `Logical` and incorporated your suggestions for internal type

Legend-Master:
> We need to create an instance of PhysicalPosition or LogicalPosition so we still need to do checks and create the instances on the JS side

We will still need to create the instance but we won't need to check `LogicalPosition`

I don't quite like this union type to be honest, this forces the user to check whether this is a physical one or a logical one, and doesn't really match the fact that we return `PhysicalPosition` and `PhysicalSize` on everything else

This is also hard to convert in the js side, getting the dpi requires an async ipc call, and we don't even have a getter for getting the dpi at position yet, we only have one for the windows

amrbashir:
good point, changed that and enforced the deserialization always matches on the Rust side

packages/api/src/tray.ts
```diff
@@ -5,90 +5,41 @@
...
-  /** Physical Y Position of the click the triggered this event. */
-  y: number
+  /** Physical position of the click the triggered this event. */
+  position: PhysicalPosition
   /** Position and size of the tray icon. */
   rect: {
-    position: PhysicalPosition
-    size: PhysicalSize
+    position: PhysicalPosition | LogicalPosition
+    size: PhysicalSize | LogicalSize
```
Legend-Master:
```suggestion
    position: PhysicalPosition
    size: PhysicalSize
```

crates/tauri/src/tray/mod.rs
```diff
@@ -563,3 +563,52 @@ impl<R: Runtime> Resource for TrayIcon<R> {
     self.app_handle.remove_tray_by_id(&self.id);
   }
 }
+
+#[cfg(test)]
+mod tests {
+  #[test]
+  fn tray_event_json_serialization() {
+    // NOTE: if this is test is ever changed, you probably need to change `TrayIconEvent` in JS as well
```
Legend-Master:
```suggestion
    // NOTE: if this test is ever changed, you probably need to change `TrayIconEvent` in JS as well
```

Legend-Master (changes requested):

Legend-Master (approved):

PR #11132, opened 2024-09-25
feat: add `ScopeObjectMatch` trait for easy scope validation

chippers:
This is a DX improvement for developers that are adding scopes to their commands (likely in plugins).

If the validation of a scope can be represented as a `bool`, the type can implement `ScopeObjectMatch` to enable a helper method on `CommandScope` that automatically checks that all deny scopes are not matched and enforcing allow scopes if they exist.

turns code like:
```rust
 #[command]
 pub(crate) async fn read_text<R: Runtime>(
     scope: CommandScope<Scope>,
     clipboard: State<'_, Clipboard<R>>,
 ) -> Result<String> {
    let text = clipboard.read_text()?;

     // first make sure the text doesn't match any existing deny scope
     if Scope::any(scope.denies(), &text) {
         return Err(crate::error::Error::Scope);
     }

     // ensure the text matches at least 1 allow scope, only if they exist
     let allow = scope.allows();
     if !allow.is_empty() && !Scope::any(allow, &text) {
         return Err(crate::error::Error::Scope);
     }

     Ok(text)
 }
```
into
```rust
#[command]
pub(crate) async fn read_text<R: Runtime>(
    scope: CommandScope<Scope>,
    clipboard: State<'_, Clipboard<R>>,
) -> Result<String> {
    let text = clipboard.read_text()?;
    if scope.matches(&text) {
        Ok(text)
    } else {
        Err(crate::error::Error::Scope)
    }
}
```

which removes the chances for developers to make simple logic errors when validating the deny and allow scopes - especially if they are using it multiple times in many commands.

### DIFF
--- a/crates/tauri/src/ipc/authority.rs
+++ b/crates/tauri/src/ipc/authority.rs
@@ -660,16 +660,16 @@ impl<T: ScopeObjectMatch> CommandScope<T> {
   /// #   }
   /// # }
   /// #
-  /// # fn do_work(_: String) -> Result<String, ()> {
+  /// # fn do_work(_: String) -> Result<String, &'static str> {
   /// #   Ok("Output".into())
   /// # }
   /// #
   /// #[command]
-  /// fn my_command(scope: CommandScope<Scope>, input: String) -> Result<String, ()> {
+  /// fn my_command(scope: CommandScope<Scope>, input: String) -> Result<String, &'static str> {
   ///   if scope.matches(&input) {
   ///     do_work(input)
   ///   } else {
-  ///     Err(())
+  ///     Err("Scope didn't match input")
   ///   }
   /// }
   /// ```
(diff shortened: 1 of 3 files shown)


crates/tauri/src/ipc/authority.rs
```diff
@@ -625,6 +625,69 @@ impl<T: ScopeObject> CommandScope<T> {
...
+  /// # }
+  /// #
+  /// #[command]
+  /// fn my_command(scope: CommandScope<Scope>, input: String) -> Result<String, ()> {
+  ///   if scope.matches(&input) {
+  ///     do_work(input)
+  ///   } else {
+  ///     Err(())
+  ///   }
+  /// }
```
amrbashir:
```suggestion
  /// # fn do_work(_: String) -> Result<String, &'static str> {
  /// #   Ok("Output".into())
  /// # }
  /// #
  /// #[command]
  /// fn my_command(scope: CommandScope<Scope>, input: String) -> Result<String, &'static str> {
  ///   if scope.matches(&input) {
  ///     do_work(input)
  ///   } else {
  ///     Err("Scope didn't match input")
  ///   }
  /// }
```

amrbashir (approved):

PR #11140, opened 2024-09-26
feat: add `Webview::show` and `Webview::hide`

### LINKED ISSUE #11126: [feat] Regarding the problem of hiding and displaying multiple webviews
### Describe the problem

In order to prevent the program from frequently creating and closing webviewWindow, should we add corresponding hiding and display functions?
for example, I need to open many webviews and switch back and forth. Here, only one webview needs to be displayed at all times, and the rest should be hidden to avoid creating a webview every time I open it. (I don't know if my example is correct, but I think it should be like this)

webview.hide(); hide the webviewWindow
webview.show(); display the webviewWindow

### Alternatives considered

_No response_

### Additional context

_No response_

amrbashir:
closes #11126


Legend-Master (approved):
Didn't test it, but the code looks good to me

PR #11154, opened 2024-09-26
feat: add `Window::set_enabled` and `Window::is_enabled`

### LINKED ISSUE #6660: [feat] Support disable specific window
### Describe the problem

I wish I could disable a specific window by using an API like `tauri::Window.get_window("target").unwrap.set_enabled(false)`, just like `BrowserWindow.setEnabled` in `Electron`.

This would be helpful to prevent users from interacting with a certain window while opening a window to configure some options.

I'd like to use an API like 'set_enabled' to enable or disable a certain window.

### Alternatives considered

_No response_

### Additional context

_No response_

amrbashir:
closes #6660


amrbashir:
the macOS implementation is NOT tested

lucasfernog (approved):
tested and fixed the is_enabled check for macOS, gonna trust you on linux and windows

PR #11163, opened 2024-09-27
refactor(core)!: pass webview label to uri scheme protocol handlers

### LINKED ISSUE #10691: [feat] labeld uri scheme protocol
### Describe the problem

when a custom uri scheme protocol is called, there is no way to know the request is send by which window and webview, so cannot do special logic between different windows and webviews.

微信图片_20240820100809

makes the protocol handler(~~`Fn(app, request) -> Response`~~) become something like this:

`Fn(app, window, webview, request) -> Response`

or

`Fn(app, window_label, webview_label, request) -> Response`

for both `register_uri_scheme_protocol` and `register_asynchronous_uri_scheme_protocol`

### Alternatives considered

_No response_

### Additional context

_No response_

amrbashir:
close #10691


crates/tauri/src/app.rs
```diff
@@ -1696,7 +1697,7 @@ tauri::Builder::default()
   pub fn register_uri_scheme_protocol<
     N: Into<String>,
     T: Into<Cow<'static, [u8]>>,
-    H: Fn(&AppHandle<R>, http::Request<Vec<u8>>) -> http::Response<T> + Send + Sync + 'static,
+    H: Fn(&AppHandle<R>, http::Request<Vec<u8>>, &str) -> http::Response<T> + Send + Sync + 'static,
```
lucasfernog: to make this more flexible, I'd change the signature to maybe contain a single `struct SchemeRequest { request: http::Request, app: AppHandle<R>, webview_label: String }`
otherwise we need to break the API again when we want to add any other information

amrbashir: I don't think the request should be hidden inside a struct and be kept as an argument so instead I did this `Fn(UriSchemeContext<'_, R>, http::Request<Vec<u8>>) -> http::Response<T>`

lucasfernog (approved):

Issue #11137, opened 2024-09-25
[bug] tauri android dev doesn't work OOB on real devices (can't connect to localhost server)


FabianLars:
### Describe the bug

https://discord.com/channels/616186924390023171/986184094050316358/1285544267854647296
https://discord.com/channels/616186924390023171/1282867292216234006

`tauri android dev` doesn't work in a plain create-tauri-app out of the box on a real device (did not test emulator) because the app won't be able to connect to the localhost server

### Reproduction

0) connect real device (didn't test emulator)
1) follow prereqs guide(s)
2) Create tauri app
3) tauri android init
4) tauri android dev
5) be sad :(
6) update vite's server.host setting to 0.0.0.0 and restart
7) be happy :)

### Expected behavior

The templates should work out of the box, or the docs should mention the prereqs i may be missing.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.22631 x86_64 (X64)
    ✔ WebView2: 129.0.2792.52
    ✔ MSVC: Visual Studio Build Tools 2022
    ✔ rustc: 1.80.1 (3f5fd8dd4 2024-08-06)
    ✔ cargo: 1.80.1 (376290515 2024-07-16)
    ✔ rustup: 1.27.1 (54dd3d00f 2024-04-24)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (default)
    - node: 20.17.0
    - pnpm: 9.10.0
    - npm: 10.8.2

[-] Packages
    - tauri 🦀: 2.0.0-rc.15
    - tauri-build 🦀: 2.0.0-rc.12
    - wry 🦀: 0.44.1
    - tao 🦀: 0.30.2
    - tauri-cli 🦀: 2.0.0-rc.16
    - @tauri-apps/api : 2.0.0-rc.4
    - @tauri-apps/cli : 2.0.0-rc.15

[-] Plugins
    - tauri-plugin-fs 🦀: 2.0.0-rc.5
    - @tauri-apps/plugin-fs : 2.0.0-rc.0 (outdated, latest: 2.0.0-rc.2)
    - tauri-plugin-shell 🦀: 2.0.0-rc.3
    - @tauri-apps/plugin-shell : 2.0.0-rc.1

[-] App
    - build-type: bundle
    - CSP: default-src 'self'; image-src data:
    - frontendDist: ../dist
    - devUrl: http://localhost:1420/
    - bundler: Vite
```

### Additional context

[...]

PR #11181, opened 2024-09-30
feat(cli): support Trunk for iOS dev out of the box

lucasfernog:
Trunk only allows us to have a dynamic development server address via the TRUNK_SERVE_ADDRESS environment variable, to we must set it from the Tauri CLI in order to support it

### DIFF
--- a/.changes/trunk-ios-dev.md
+++ b/.changes/trunk-ios-dev.md
@@ -1,0 +1,6 @@
+---
+"tauri-cli": patch:enhance
+"@tauri-apps/cli": patch:enhance
+---
+
+Set the `TRUNK_SERVE_ADDRESS` environment variable when running on iOS physical devices to support Trunk.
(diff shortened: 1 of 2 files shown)


.changes/trunk-ios-dev.md
```diff
@@ -0,0 +1,6 @@
+---
+"tauri-cli": patch:enhance
+"@tauri-apps/cli": patch:enhance
+---
+
+Set the `TRUNK_SERVE_ADDRESS` environment variable when running on iOS physical devices to support Trunk.
```
amrbashir, line 6: I think we need to do the same for android, no? see #11137

lucasfernog: I'll investigate that one separately (need to get an android phone from someone here first)

amrbashir (approved):

PR #11186, opened 2024-10-01
feat(cli): retain log verbosity when running a mobile IDE script

lucasfernog:
the mobile IDE scripts (android-studio-script and xcode-script) are executed in a separate process and should retain the verbosity provided by the user

### DIFF
--- a/crates/tauri-cli/src/lib.rs
+++ b/crates/tauri-cli/src/lib.rs
@@ -216,10 +216,18 @@ where
     Err(e) => e.exit(),
   };
 
+  let verbosity_number = std::env::var("TAURI_CLI_VERBOSITY")
+    .ok()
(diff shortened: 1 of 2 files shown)


crates/tauri-cli/src/lib.rs
```diff
@@ -216,10 +216,18 @@ where
     Err(e) => e.exit(),
   };
 
+  let verbosity_number = std::env::var("TAURI_CLI_VERBOSITY")
+    .ok()
+    .and_then(|v| v.parse().ok())
+    .unwrap_or(cli.verbose);
+  println!("{verbosity_number}");
```
amrbashir: left over?

amrbashir (approved):

PR #11191, opened 2024-10-02
feat(api): add `SERIALIZE_TO_IPC_FN` const and implement it for dpi types, add more constructors

amrbashir:


packages/api/src/dpi.ts
```diff
@@ -8,20 +8,40 @@
  * @since 2.0.0
  */
 class LogicalSize {
-  type = 'Logical'
+  type = 'Logical' as const
```
Legend-Master:
```suggestion
  readonly type = 'Logical'
```

packages/api/src/dpi.ts
```diff
@@ -41,13 +102,31 @@ class LogicalSize {
  * @since 2.0.0
  */
 class PhysicalSize {
-  type = 'Physical'
+  type = 'Physical' as const
```
Legend-Master:
```suggestion
  readonly type = 'Physical'
```

packages/api/src/dpi.ts
```diff
@@ -72,20 +192,40 @@ class PhysicalSize {
  * @since 2.0.0
  */
 class LogicalPosition {
-  type = 'Logical'
+  type = 'Logical' as const
```
Legend-Master:
```suggestion
  readonly type = 'Logical'
```

packages/api/src/dpi.ts
```diff
@@ -105,28 +286,72 @@ class LogicalPosition {
  * @since 2.0.0
  */
 class PhysicalPosition {
-  type = 'Physical'
+  type = 'Physical' as const
```
Legend-Master:
```suggestion
  readonly type = 'Physical'
```

Legend-Master (changes requested):

packages/api/src/webview.ts
```diff
@@ -359,9 +359,9 @@ class Webview {
    * @returns The webview's position.
    */
   async position(): Promise<PhysicalPosition> {
-    return invoke<{ x: number; y: number }>('plugin:webview|webview_position', {
+    return invoke<PhysicalPosition>('plugin:webview|webview_position', {
```
Legend-Master: This is not the actual type here, we should keep it as `{ x: number; y: number }`, `then((p) => new PhysicalPosition(p))` is fine though

And same for the other ones

crates/tauri/scripts/ipc.js
```diff
@@ -88,6 +88,40 @@
...
     }
 
+    // `postMessage` uses `structuredClone` to serialize the data before sending it
+    // unlike `JSON.stringify`, we can't extend a type with a method similar to `toJSON`
+    // so that `structuredClone` would use, so untl https://github.com/whatwg/html/issues/7428
+    // we manually call `toIPC`
+    function recursiveCallToIPC(data) {
+      if (
+        typeof data === 'object' &&
+        'constrcutor' in data &&
```
lucasfernog:
```suggestion
        'constructor' in data &&
```

crates/tauri/scripts/ipc.js
```diff
@@ -88,6 +88,40 @@
       return
     }
 
+    // `postMessage` uses `structuredClone` to serialize the data before sending it
+    // unlike `JSON.stringify`, we can't extend a type with a method similar to `toJSON`
+    // so that `structuredClone` would use, so untl https://github.com/whatwg/html/issues/7428
```
lucasfernog:
```suggestion
    // so that `structuredClone` would use, so until https://github.com/whatwg/html/issues/7428
```

crates/tauri/scripts/process-ipc-message-fn.js
```diff
@@ -17,19 +17,19 @@
...
       } else if (val instanceof ArrayBuffer) {
         return Array.from(new Uint8Array(val))
       }  else if (
-        val instanceof Object &&
+        typeof val === "object" &&
         '__TAURI_CHANNEL_MARKER__' in val &&
         typeof val.id === 'number'
       ) {
         return `__CHANNEL__:${val.id}`
+      } else if (typeof val === "object" &&'toIPC' in val){
```
lucasfernog:
```suggestion
      } else if (typeof val === "object" && 'toIPC' in val){
```

lucasfernog:
this one can wait to be merged after the v2 stable release right?

.changes/api-dpi-toJSON-toIPC.md
```diff
@@ -0,0 +1,8 @@
+---
+"@tauri-apps/cli": "patch:feat"
```
amrbashir:
```suggestion
"@tauri-apps/api": "minor:enhanced"
```

.changes/api-dpi-toJSON-toIPC.md
```diff
@@ -0,0 +1,8 @@
+---
+"@tauri-apps/api": "minor:enhanced"
```
amrbashir:
```suggestion
"@tauri-apps/api": "minor:enhance"
```

packages/api/src/window.ts
```diff
@@ -581,12 +581,9 @@ class Window {
    * @returns The window's outer size.
    */
   async outerSize(): Promise<PhysicalSize> {
-    return invoke<{ width: number; height: number }>(
-      'plugin:window|outer_size',
-      {
-        label: this.label
-      }
-    ).then(({ width, height }) => new PhysicalSize(width, height))
+    return invoke<PhysicalSize>('plugin:window|outer_size', {
```
Legend-Master:
```suggestion
    return invoke<{ width: number; height: number }>('plugin:window|outer_size', {
```

also this one 🙃

Legend-Master (changes requested):

packages/api/src/window.ts
```diff
@@ -2391,7 +2311,7 @@ async function availableMonitors(): Promise<Monitor[]> {
  */
 async function cursorPosition(): Promise<PhysicalPosition> {
   return invoke<PhysicalPosition>('plugin:window|cursor_position').then(
```
Legend-Master, line 2326:
```suggestion
  return invoke<{ x: number; y: number }>('plugin:window|cursor_position').then(
```

🙃

crates/tauri/scripts/process-ipc-message-fn.js
```diff
@@ -17,19 +17,19 @@
...
       } else if (val instanceof ArrayBuffer) {
         return Array.from(new Uint8Array(val))
       }  else if (
-        val instanceof Object &&
+        typeof val === "object" &&
         '__TAURI_CHANNEL_MARKER__' in val &&
         typeof val.id === 'number'
       ) {
         return `__CHANNEL__:${val.id}`
+      } else if (typeof val === "object" && 'toIPC' in val){
```
Legend-Master:
```suggestion
      } else if (typeof val === "object" && 'toIPC' in val) {
```

Legend-Master (changes requested):
Do you think we should use `toTauriIpc` instead of just `toIPC`?

amrbashir:
> Do you think we should use `toTauriIpc` instead of just `toIPC`?

yes, i am not to keen on `toIPC` and don't want to make it collide with user types, so I was thinking maybe `IpcValue` class to extend like `Resource` class.

amrbashir:
Made a `ToIpcSymbol` const and used that instead.

Legend-Master:
Sounds good to me, but shouldn't the symbol be a [`Symbol`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#browser_compatibility) not a string?

amrbashir:
> Sounds good to me, but shouldn't the symbol be a [`Symbol`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#browser_compatibility) not a string?

Normally it should be, but we access this value in multiple places and we can't keep re-defining it because it will not be the same value (that's how Symbol works, it is always unique). We can define it on the global `window` object and access from different places but so I don't want to expose this symbol in `core` module like this
```ts
const ToIPCSymbol = window.ToIPCSymbol
```
because it will break usage with next.js and SSR framworks and I don't want to deal with their issues again.

Legend-Master:
Shouldn't this work?

image

Legend-Master:
Never mind, we can't use import here

Legend-Master:
What about [`Symbol.for()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/for)?

amrbashir:
That works I guess but is it really that different? the main problem here IMO is that we can't share the value by using imports

Legend-Master:
Right, that doesn't really make a difference, let's just go with a string then, in that case maybe changing the name to `TO_IPC_KEY` instead since it's not really a symbol?

amrbashir:
renamed!

Legend-Master (dismissed):
Didn't test it, but the code looks good to me

packages/api/src/dpi.ts
```diff
@@ -33,6 +55,48 @@ class LogicalSize {
...
+   * import { LogicalSize } from '@tauri-apps/api/dpi';
+   * import { invoke } from '@tauri-apps/api/core';
+   *
+   * const size = new LogicalSize(400, 500);
+   * await invoke("do_something_with_size", { size: size.toIPC() })
+   * ```
+   *
+   * @since 2.0.0
+   */
+  [SERIALIZE_TO_IPC_FN]() {
```
lucasfernog, line 59: I like the whole toIPC approach (makes the `__CHANNEL__` thingy available to anyone) but this is actually a breaking change - existing commands relying on PhysicalPosition / PhysicalSize / LogicalPosition / LogicalSize cannot deserialize this new format, so we should revert these - if you explictly want a generic Position or Size type we need to create a wrapper type just like Rust has one

lucasfernog (changes requested):

packages/api/src/dpi.ts
amrbashir: that makes sense, I have added `Position` and `Size` classes and reverted breaking changes.

packages/api/src/core.ts
```diff
@@ -26,8 +76,6 @@ function transformCallback<T = unknown>(
 
 class Channel<T = unknown> {
   id: number
-  // @ts-expect-error field used by the IPC serializer
-  private readonly __TAURI_CHANNEL_MARKER__ = true
```
amrbashir: @lucasfernog isn't this a breaking change? removing the field will make new `@tauri-apps/api` packages not work with `tauri@2.0.0`

crates/tauri/scripts/process-ipc-message-fn.js
```diff
@@ -16,20 +16,19 @@
...
       } else if (val instanceof Uint8Array) {
         return Array.from(val)
       } else if (val instanceof ArrayBuffer) {
         return Array.from(new Uint8Array(val))
-      }  else if (
-        val instanceof Object &&
-        '__TAURI_CHANNEL_MARKER__' in val &&
-        typeof val.id === 'number'
-      ) {
-        return `__CHANNEL__:${val.id}`
```
amrbashir: same here

lucasfernog (approved):

PR #11205, opened 2024-10-03
fix: downgrade MSRV to 1.77.2 to support Windows 7

lucasfernog:


tweidinger (approved):

FabianLars:
I don't have access to discord right now, what was the issue prompting for this change?

1.75 was the last version with proper Win7 support so I'd like to know what 1.77 brings us here.

FabianLars:
ah nevermind, the change was later than initially announced (in 1.72 blog they said 1.75 is the last one but later changed that to 1.77 i guess).
Still, we shouldn't hold on to that for ever, this already caused so much pain in v1. Of course, it's totally fair for the first versions so everyone can use v2.
Also, let's not forget that Rust still supports windows 7 just under a new target triple which imo is good enough for the future since tauri's win7 support itself is more like best effort itself.

PR #11213, opened 2024-10-03
fix(core): deadlock when creating new window with tracing feature

lucasfernog:
when the tracing feature is enabled, all eval() calls are synchronous (so we can track how much time they take) so the window creation hooks deadlocks since we're also reacting to a main thread window event at the same time. running the hooks in a separate thread fixes this.

closes https://github.com/crabnebula-dev/devtools/issues/352

### DIFF
--- a/core/tauri/src/window.rs
+++ b/core/tauri/src/window.rs
@@ -351,19 +351,36 @@ impl<'a, R: Runtime> WindowBuilder<'a, R> {
     }
     .map(|window| self.manager.attach_window(self.app_handle.clone(), window))?;
 
-    self.manager.eval_script_all(format!(
-      "window.__TAURI_METADATA__.__windows = {window_labels_array}.map(function (label) {{ return {{ label: label }} }})",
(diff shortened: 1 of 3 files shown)


core/tauri/src/window.rs
```diff
@@ -351,19 +351,36 @@ impl<'a, R: Runtime> WindowBuilder<'a, R> {
...
-    )?;
+    if cfg!(feature = "tracing") {
+      std::thread::spawn(move || {
+        if let Err(e) = window_created_hook() {
+          log::error!("failed to trigger window creation hooks: {e}");
+        }
+      });
+    } else {
+      window_created_hook()?;
+    }
```
```suggestion
    #[cfg(not(feature = "tracing")]
    window_created_hook()?;
    #[cfg(feature = "tracing")]
    std::thread::spawn(move || {
      if let Err(e) = window_created_hook() {
        log::error!("failed to trigger window creation hooks: {e}");
      }
    });
```

amrbashir (approved):

PR #11234, opened 2024-10-05
make rpm name consistent with others

goenning:
Would you be open to make rpm name consistent with other platforms.

Because RPM currently users `-` instead of `_`, it does not follow the same order as the other files, so it gets confusing when looking at it on S3/Local

image


lucasfernog:
unfortunately that would now be a breaking change, so we'd need an alternative solution to this - maybe a filename format specifier in the config?

goenning:
Yeah that’s a good solution. I’d also use that to remove the “-1” from rpm releases

PR #11241, opened 2024-10-06
feat: allow overriding resource directory at compile-time

### LINKED ISSUE #8679: [bug] tauri resource directory is hard-coded
### Describe the bug

Tauri resource directory location is hard-coded making Tauri fail when it can't read or write from it, this is especially a problem in immutable distros like NixOS.

### Reproduction

https://github.com/NixOS/nixpkgs/pull/265771 - build this without the tauri patch in it.

### Expected behavior

For it to succeed in building.

### Full `tauri info` output

```text
N/A
```

### Stack trace

_No response_

### Additional context

There is a fix in the Github pr I linked in the files it's called [tauri-env-resource-dir.patch](https://github.com/NixOS/nixpkgs/blob/728e957b03d9194b7d277907cb1bc154081cf35a/pkgs/by-name/do/dorion/tauri-env-resource-dir.patch)

griffi-gh:
closes #8679

### DIFF
--- a/crates/tauri-utils/src/platform.rs
+++ b/crates/tauri-utils/src/platform.rs
@@ -264,7 +264,12 @@ fn is_cargo_output_directory(path: &std::path::Path) -> bool {
 /// On iOS, it's `${exe_dir}/assets`.
 ///
 /// Android uses a special URI prefix that is resolved by the Tauri file system plugin `asset://localhost/`
+///
+/// Additionally, the resource directory can be overriden by setting the `TAURI_RESOURCE_DIR` environment variable at compile time.
 pub fn resource_dir(package_info: &PackageInfo, env: &Env) -> crate::Result<PathBuf> {
+  if let Some(dir) = option_env!("TAURI_RESOURCE_DIR") {
+    return Ok(PathBuf::from(format!("{}/{}", dir, package_info.name)));
+  }
   #[cfg(target_os = "android")]
   return resource_dir_android(package_info, env);
   #[cfg(not(target_os = "android"))]


griffi-gh:
(formatted with rustfmt)

crates/tauri-utils/src/platform.rs
```diff
@@ -264,7 +264,12 @@ fn is_cargo_output_directory(path: &std::path::Path) -> bool {
 /// On iOS, it's `${exe_dir}/assets`.
 ///
 /// Android uses a special URI prefix that is resolved by the Tauri file system plugin `asset://localhost/`
+///
+/// Additionally, the resource directory can be overriden by setting the `TAURI_RESOURCE_DIR` environment variable at compile time.
 pub fn resource_dir(package_info: &PackageInfo, env: &Env) -> crate::Result<PathBuf> {
+  if let Some(dir) = option_env!("TAURI_RESOURCE_DIR") {
```
lucasfernog, line 270: this is not a really portable solution since you'd need separate builds per target platform; following the explanation from https://github.com/tauri-apps/tauri/issues/8679#issuecomment-1911288087 I think we should instead always check if `exe_dir.join(format!("../lib/{}", package_info.package_name()))` exists and if it does, use it as resource path

griffi-gh: > we should instead always check if `exe_dir.join(format!("../lib/{}", package_info.package_name()))` exists

#11429

griffi-gh:
Superseded by #11429

PR #11258, opened 2024-10-07
feat(cli): make tauri/cli fully support projects with non-standard structure

regexident:
Usually I'd first open a feature suggestion issue before embarking on the pull request itself for something like this, but I already have the implementation (as it was necessary to unblock myself), so I'm gonna skip the issue for this one.

The **tl;dr** is that we would like to propose adding explicit support for more flexible project structures (i.e. moving `src-tauri` out of the frontend app), which surprisingly took only minimal and non-invasive changes for the tauri cli tool.

(FYI: we'll be using `web-app` and `tauri-app` as dummy-names for the top-most `app` and `app/src-tauri` directories here throughout, they don't matter!)

With the patched `cargo-tauri` binary one is able to run `pnpm tauri …`/`cargo tauri …` from both, the app, as well as src-tauri directory (and even —very convenient!— from the top-most directory!), regardless of how the project structured, like so:

Given a project structured like this:

```plain
my-app
├── ...
├── tauri-app
│   └── ...
└── web-app
    └── ...
```

… one would be able to run this:

```terminal
cd my-app

# probably something you'd put in an `.env` file:
export TAURI_APP_PATH="./web-app"
export TAURI_FRONTEND_PATH="./tauri-app"

pnpm tauri dev
# or
cargo tauri dev
```

<details>
<summary>Currently the CLI tool crashes when running from projects with non-standard structure</summary>

Currently running `pnpm tauri …`, such as `pnpm tauri migrate` from the web app directory of a project with non-standard structure leads to the following crash:

```terminal
$ cd ./web-app
$ pnpm tauri migrate

thread '<unnamed>' panicked at crates/tauri-cli/src/helpers/app_paths.rs:107:5:
Couldn't recognize the current folder as a Tauri project. It must contain a `tauri.conf.json`, `tauri.conf.json5` or `Tauri.toml` file in any subfolder.
 ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  Command was killed with SIGABRT (Aborted): tauri migrate
```

Running the equivalent `cargo` command from the tauri src directory (in again, a non-standard structure) works, for what it's worth:

```terminal
$ cd ./tauri-app
$ cargo tauri migrate

Info Installing Cargo dependency "tauri-plugin-dialog"...
Updating crates.io index
  Adding tauri-plugin-dialog v2 to dependencies
...
Added permission `dialog:default` to `migrated` at <SNIP>/tauri-app/capabilities/migrated.json
Info Adding plugin to <SNIP>/tauri-app/src/main.rs
Info Running `cargo fmt`...
```

But since the `pnpm` route seems to be the far more common (and is usually the one that's dominantly recommended in the official guides) one I'd expect very few people to realize that while one crashes, the other works, even in non-standardly structured projects.

</details>

## Motivation

The standard structure of a tauri project looks something like this:

```plain
my-app
├── package.json
├── src
├── src-tauri
│   ├── Cargo.toml
│   ├── src
│   ├── tauri.config.json
│   └── ...
└── ...
```

The tauri project gets added as a sub-directory of the existing web project. This is certainly convenient for some things (like being able to have a single place for scripts: `package.json`, etc.) but also rather inconvenient for others.

### Flexible project structure

This rather rigid structure seems both, unnecessarily limiting and inconvenient, compared to say a structure like this:

```plain
my-app
├── tauri-app
│   ├── Cargo.toml
│   ├── src
│   ├── tauri.config.json
│   └── ...
└── web-app
    ├── package.json
    ├── src
    └── ...
```

The main source of inconvenience of the standard structure comes from the fact that **web dev projects tend to be littered with top-level configuration junk.**

### Junk, junk everywhere

<details>
<summary>What we mean by "top-level configuration junk"</summary>

As such the directory diagrams above where lying. Rather than being neat and clean:

```plain
web-app
├── package.json
├── src
└── ...
```

… the `web-app` directory's actual contents look like this garbage fire:

```plain
web-app
├── .gitignore
├── .npmrc
├── .prettierignore
├── .prettierrc
├── .svelte-kit
│   └── ...
├── build
│   └── ...
├── bundlemeta.json
├── components.json
├── eslint.config.mjs
├── node_modules
│   └── ...
├── package.json
├── packages
│   └── ...
├── playwright.config.ts
├── pnpm-lock.yaml
├── pnpm-workspace.yaml
├── postcss.config.ts
├── setupTest.ts
├── src
│   └── ...
├── static
│   └── ...
├── svelte.config.js
├── tailwind.config.ts
├── tests
│   └── ...
├── tsconfig.json
├── vite.config.ts
└── ...
```

</details>

### Poor developer experience & productivity

Other than `src`/`tests`/`package.json` hardly any of the many top-level files within a run-of-the-mill `web-app` project tend to be of actual regular(!) importance in one's day-to-day work, yet there is no way to hide them, making it at times hard to see the forest for the trees.

Unfortunately there really isn't much one can do about that (afaik), it's just the mess we find ourselves in when doing web dev work.
But forcing the same nonsense upon everybody working on the Rust bits of a Tauri project is really just adding insult to injury.

### Poor CI experience & performance, and significantly increased costs

Speaking of which, another major issue with the classic structure is that there simply is no straight-forward and reliable way to be smart about which jobs to run for a PR and on which CI runners, when everything is stuffed in the same directory.

This is particularly problematic in the context of a multi-platform Tauri project (which probably 90% of Tauri projects are) and the reality of macOS-based CI runners commonly being [10x as expensive as their Linux-based counterparts](https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/about-billing-for-github-actions#minute-multipliers).

From a cost-perspective one would thus probably at least want to run the platform-agnostic frontend tests on the cheap CI runner and possibly even try to carve out any macOS-specific backend tests so only those get frequently run on the expensive macOSs (10x) runner, and everything else on cheap Linux (1x) or Windows (2x) runners.

### Frontend and backend tests competing for CI time

One might decide to split frontend and backend test jobs into separate workflows and have them run in parallel, in order to reduce the time spent by devs waiting for their pull requests to finish running their actions.

And while this might help reduce the average time to wait it ends up increasing the average money spent as unlike a single serial workflow that can abort on the first error with parallel workflows your "other" workflow will keep running and thus keep blocking the queue when your workflow encounters an error. (There probably are arcane Github action spells to solve this, but shouldn't this stuff just do the right thing by default?)

One could certainly split up the workflows and add extensive `on.pull_request.paths`/`on.pull_request.exclude_paths` lists to each of them and achieve a more fine-grained CI workflow selection, but such lists tend to be frickle at best and at worst could lead to CI not running a workflow where it absolutely should have.

By turning `web-app` and `tauri-app` into sibling directories selecting the right CI jobs and the right CI runner for a given PR becomes almost trivial: 

<details>
<summary>How to effortlessly run separate workflows for the app and tauri</summary>

```yaml
# .github/workflows/tauri-app.yml

on:
    pull_request:
        branches: [main, "**"]
        paths:
            - "tauri-app/**"
            - ".github/workflows/tauri-app.yml"
# ...
```

```yaml
# .github/workflows/web-app.yml

on:
    pull_request:
        branches: [main, "**"]
        paths:
            - "web-app/**"
            - ".github/workflows/web-app.yml"
# ...
```

</details>

As a result we found our CI times to greatly improve as said sibling structure allowed us to limit the expensive macOS tests to just the `tauri-app` directory. (We're still running full tests on a fixed schedule as a safety net to catch any regressions introduced at the integration layer of `web-app` and `tauri-app`, but so far those have't caught anything slipping through.)

## Benefits of a non-standard project structure

Anecdotally we found the developer experience and productivity to greatly improve upon switching the app/src-tauri directories from the standard parent-child to the alternative sibling structure.

While focussing on one side of a tauri app other's directory can be folded away in the IDE, allowing for working uninterruptedly and with focus.

For larger teams where the web-frontend and tauri-backend are developed by different people having a clear separation has clear benefits that  probably don't need to be illustrated here on their own.

Additionally having separate directories helps with VCS features like [Github's code owners](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners), which afaik doesn't even support exclude paths, so rather than directory globs for `./web-app/**` and `./tauri-app/**` every top-level file would need to be listed separately, leading to additional maintenance overhead.

When it comes to CI ergonomics and economics we —again anecdotally—found that having straight-forward control (yet fine-grained where necessary) greatly helped with streamlining the project's CI, significantly reducing the mean waiting time per CI action, as well as greatly reduced cost overall.

(From a pure architectural and usability point of view one could go as far as argue that the sibling structure should probably be the default, not the other way round. But that's not what this PR is about!)

That said we would really like to see Tauri support a more flexible project structure. Especially since it doesn't seem to take that much to make its tools work with them.

We are fully aware that a change like this (i.e. introducing env vars) has to be taken carefully and will thus probably take more than just "it works, let's ship it!". As such this PR is sort of a proof-of-concept for how things could be done.

(Disclaimer: we have so far only run the patched CLI with a few commands on our project with a sibling structure, but all of them successfully so, which is promising! That said every non-trivial tool inevitably will have accrued implicit assumptions about the currently rather rigid project structure and thus it is expected that more changes may be necessary elsewhere.)

### DIFF
--- a/.changes/support-custom-repo-structure.md
+++ b/.changes/support-custom-repo-structure.md
@@ -1,0 +1,7 @@
+---
+"@tauri-apps/cli": patch:enhance
+"tauri-cli": patch:enhance
+---
+
+Support custom project directory structure where the Tauri app folder is not a subfolder of the frontend project.
+The frontend and Tauri app project paths can be set with the `TAURI_FRONTEND_PATH` and the `TAURI_APP_PATH` environment variables respectively.
--- a/crates/tauri-cli/src/helpers/app_paths.rs
+++ b/crates/tauri-cli/src/helpers/app_paths.rs
@@ -88,8 +88,7 @@ fn env_tauri_src_dir() -> Option<PathBuf> {
 }
 
 pub fn resolve_tauri_dir() -> Option<PathBuf> {
-  let src_dir =
-    env_tauri_src_dir().or_else(|| current_dir().map(|cwd| cwd.join("src-tauri")).ok())?;
+  let src_dir = env_tauri_src_dir().or_else(|| current_dir().ok())?;
 
   if src_dir.join(ConfigFormat::Json.into_file_name()).exists()
     || src_dir.join(ConfigFormat::Json5.into_file_name()).exists()
--- a/crates/tauri-cli/src/helpers/app_paths.rs
+++ b/crates/tauri-cli/src/helpers/app_paths.rs
@@ -18,8 +18,10 @@ use tauri_utils::{
 };
 
 const TAURI_GITIGNORE: &[u8] = include_bytes!("../../tauri.gitignore");
-const ENV_TAURI_APP_DIR: &str = "TAURI_APP_DIR";
-const ENV_TAURI_SRC_DIR: &str = "TAURI_SRC_DIR";
+// path to the Tauri app (Rust crate) directory, usually <cwd>/src-tauri
+const ENV_TAURI_APP_PATH: &str = "TAURI_APP_PATH";
+// path to the frontend app directory
+const ENV_TAURI_FRONTEND_PATH: &str = "TAURI_FRONTEND_PATH";
 
 static APP_DIR: OnceLock<PathBuf> = OnceLock::new();
 static TAURI_DIR: OnceLock<PathBuf> = OnceLock::new();
@@ -71,16 +73,16 @@ fn lookup<F: Fn(&PathBuf) -> bool>(dir: &Path, checker: F) -> Option<PathBuf> {
   None
 }
 
-fn env_tauri_app_dir() -> Option<PathBuf> {
-  std::env::var(ENV_TAURI_APP_DIR)
+fn env_tauri_app_path() -> Option<PathBuf> {
+  std::env::var(ENV_TAURI_APP_PATH)
     .map(PathBuf::from)
     .ok()?
     .canonicalize()
     .ok()
 }
 
-fn env_tauri_src_dir() -> Option<PathBuf> {
-  std::env::var(ENV_TAURI_SRC_DIR)
+fn env_tauri_frontend_path() -> Option<PathBuf> {
+  std::env::var(ENV_TAURI_FRONTEND_PATH)
     .map(PathBuf::from)
     .ok()?
     .canonicalize()
@@ -88,7 +90,7 @@ fn env_tauri_src_dir() -> Option<PathBuf> {
 }
 
 pub fn resolve_tauri_dir() -> Option<PathBuf> {
-  let src_dir = env_tauri_src_dir().or_else(|| current_dir().ok())?;
+  let src_dir = env_tauri_frontend_path().or_else(|| current_dir().ok())?;
 
   if src_dir.join(ConfigFormat::Json.into_file_name()).exists()
     || src_dir.join(ConfigFormat::Json5.into_file_name()).exists()
@@ -111,7 +113,7 @@ pub fn resolve_tauri_dir() -> Option<PathBuf> {
 
 pub fn resolve() {
   TAURI_DIR.set(resolve_tauri_dir().unwrap_or_else(|| {
-    let env_var_name = env_tauri_src_dir().is_some().then(|| format!("`{ENV_TAURI_SRC_DIR}`"));
+    let env_var_name = env_tauri_frontend_path().is_some().then(|| format!("`{ENV_TAURI_FRONTEND_PATH}`"));
     panic!("Couldn't recognize the {} folder as a Tauri project. It must contain a `{}`, `{}` or `{}` file in any subfolder.",
       env_var_name.as_deref().unwrap_or("current"),
       ConfigFormat::Json.into_file_name(),
@@ -131,7 +133,7 @@ pub fn tauri_dir() -> &'static PathBuf {
 }
 
 pub fn resolve_app_dir() -> Option<PathBuf> {
-  let app_dir = env_tauri_app_dir().unwrap_or_else(|| current_dir().expect("failed to read cwd"));
+  let app_dir = env_tauri_app_path().unwrap_or_else(|| current_dir().expect("failed to read cwd"));
 
   if app_dir.join("package.json").exists() {
     return Some(app_dir);
--- a/crates/tauri-cli/src/helpers/app_paths.rs
+++ b/crates/tauri-cli/src/helpers/app_paths.rs
@@ -18,6 +18,9 @@ use tauri_utils::{
 };
 
 const TAURI_GITIGNORE: &[u8] = include_bytes!("../../tauri.gitignore");
+const ENV_TAURI_APP_DIR: &str = "TAURI_APP_DIR";
+const ENV_TAURI_SRC_DIR: &str = "TAURI_SRC_DIR";
+
 static APP_DIR: OnceLock<PathBuf> = OnceLock::new();
 static TAURI_DIR: OnceLock<PathBuf> = OnceLock::new();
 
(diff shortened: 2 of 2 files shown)


regexident:
The tests were already failing for the upstream commit prior to this PR's changes. Not sure what to do about them. 🤔

lucasfernog:
I've renamed `TAURI_SRC_DIR` to `TAURI_FRONTEND_PATH` and `TAURI_APP_DIR` to `TAURI_APP_PATH` to be more clear on what path does the env var refer to, what do you think? app and src are too obscure IMO, as both the frontend and the tauri app directories contain the "app" and the "src" 😂

lucasfernog:
tests were failing because you changed the tauri.conf lookup function to start from the `<cwd>/src-tauri` folder, where we actually lookup from `<cwd>` instead to partially support your use case - having the Tauri app in a subfolder that is not named `src-tauri`

lucasfernog (dismissed):

regexident:
> I've renamed `TAURI_SRC_DIR` to `TAURI_FRONTEND_PATH` and `TAURI_APP_DIR` to `TAURI_APP_PATH` to be more clear on what path does the env var refer to, what do you think? app and src are too obscure IMO, as both the frontend and the tauri app directories contain the "app" and the "src" 😂

Works for me! 👍  I wasn't exactly sure about the somewhat overloaded names to begin with.

> tests were failing because you changed the tauri.conf lookup function to start from the `<cwd>/src-tauri` folder, where we actually lookup from `<cwd>` instead to partially support your use case - having the Tauri app in a subfolder that is not named `src-tauri`

Oh I see. Interesting. 🤔  When I noticed the tests failing I checked out the PR's parent commit to check if I broke something and it failed the same way. I guess I somehow didn't clear the caches properly.

lucasfernog (approved):

StreetStrider:
Hi. Great work. What changes must be done to existing project to migrate? I feel like I've been using this approach already (even without the patch): My `build.frontendDist = '../web'`. If someone's wanting to use this, what actions must be done? Is it enougt to just move directories, update `build.frontendDist` and go on? Or is something else required?

regexident:
@lucasfernog just updated the tauri CLI and got this:

```plain
$ cd ./web-app
$ pnpm tauri dev

thread '<unnamed>' panicked at crates/tauri-cli/src/helpers/app_paths.rs:117:5:
Couldn't recognize the `TAURI_FRONTEND_PATH` folder as a Tauri project. It must contain a `tauri.conf.json`, `tauri.conf.json5` or `Tauri.toml` file in any subfolder.
 ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  Command was killed with SIGABRT (Aborted): tauri dev
```

It looks like the env vars got mixed up in [8af2d51](https://github.com/tauri-apps/tauri/pull/11258/commits/8af2d51aa69c96ac0d1aebf87ad3ac81bb4b3ae3). 😦

regexident:
@StreetStrider what made the above structure work for us was adding `"cwd": "../web-app"` to all the npm commands in `tauri.config.json`:

```json
{
  "build": {
    "beforeBuildCommand": {
      "cwd": "../web-app",
      "script": "pnpm build"
    },
    "beforeDevCommand": {
      "cwd": "../web-app",
      "script": "pnpm dev"
    },
  }
}
```

Update: no longer necessary, see comment below!

regexident:
@StreetStrider I just checked and the above is actually no longer necessary (and might even break things) with the env vars introduced with this PR (and fixed in #11492).

See https://github.com/tauri-apps/tauri/pull/11492#issuecomment-2437688362 for a complete step-by-step of the minimal steps necessary to switch from tauri's standard structure to a custom one.

PR #11264, opened 2024-10-07
fix(core): return empty body for HEAD requests to `asset` protocol, closes #11263

### LINKED ISSUE #11263: [bug] the asset protocol should respect HEAD method and not return body
### Describe the bug

In the HTTP standard, a HEAD request is similar to a GET request but [does not return a message body in the response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD). 
> The server sends back a [200 OK](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200) response comprised only of headers. The response is effectively metadata that describes the resource instead of the resource itself.

In the current implementation of the asset protocol, the response body includes the entire file content regardless of whether a HEAD method is used, which can result in unnecessary memory consumption in certain cases.
```
let buf = if let Some(b) = read_bytes {
      b
    } else {
      crate::async_runtime::safe_block_on(async move {
        let mut local_buf = Vec::with_capacity(len as usize);
        file.read_to_end(&mut local_buf).await?;
        Ok::<Vec<u8>, anyhow::Error>(local_buf)
      })?
    };
    resp = resp.header(CONTENT_LENGTH, len);
    resp.body(buf.into())
```

The screenshot shows the entire file content is downloaded which is not intended for the HEAD request.
screenshot-20241008-005340

### Reproduction

_No response_

### Expected behavior

It should not return the file content in the response body with HEAD request.
screenshot-20241008-005547

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 14.3.1 arm64 (X64)
    - node: 22.5.1
    - pnpm: 9.12.0
    - yarn: 1.22.22
    - npm: 10.8.2

[-] Packages
    - tauri 🦀: 2.0.2
    - tauri-build 🦀: 2.0.1
    - wry 🦀: 0.44.1
    - tao 🦀: 0.30.3
    - @tauri-apps/api : 2.0.2
    - @tauri-apps/cli : 2.0.2

[-] Plugins
    - tauri-plugin-log 🦀: 2.0.1
    - @tauri-apps/plugin-log : 2.0.0
    - tauri-plugin-http 🦀: 2.0.1
    - @tauri-apps/plugin-http : 2.0.0
    - tauri-plugin-dialog 🦀: 2.0.1
    - @tauri-apps/plugin-dialog : 2.0.0
    - tauri-plugin-os 🦀: 2.0.1
    - @tauri-apps/plugin-os : 2.0.0
    - tauri-plugin-fs 🦀: 2.0.1
    - @tauri-apps/plugin-fs : 2.0.0

[-] App
    - build-type: bundle
    - CSP: default-src 'self' ipc: http://ipc.localhost; img-src 'self' data: asset: http://asset.localhost
    - frontendDist: ../out
    - devUrl: http://localhost:3000/
    - framework: React (Next.js)
    - bundler: Webpack
```

### Stack trace

_No response_

### Additional context

_No response_

chrox:


amrbashir (changes requested):
Thank you, could you add a change file in `.changes` directory?

chrox:
> Thank you, could you add a change file in `.changes` directory?

OK. Done.

.changes/change-pr-11264.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri": patch:bug
+---
+
+fix: successful response should have no body to a HEAD request, closes #11263
```
amrbashir:
```suggestion
Respond with empty body for `HEAD` requests to `asset` protocol
```

amrbashir (approved):
Thank you

PR #11290, opened 2024-10-09
fix(core): expose ipc::Invoke

lucasfernog:
this type is actually important for custom invoke_handler and plugin implementations, so at least making it visible on docs is important


suxiaoshao (approved):
Thanks!

amrbashir (approved):

PR #11322, opened 2024-10-11
feat(tauri-cli): add plugin remove command

### LINKED ISSUE #10384: [v2] No way to remove plugins
I couldn't find anything in docs or with `cargo tauri --help` that is related to removing plugins from your project. 

I know that removing plugins from project is rare, but I still think it's a solid "nice to have".

ShaunSHamilton:
Closes #10384 

TODO: I would like help understanding the failing snapshot test. Is my change invalid? Or, does the snapshot just need to be updated?

### DIFF
--- a/.changes/cli-remove-command.md
+++ b/.changes/cli-remove-command.md
@@ -1,0 +1,5 @@
+---
+"tauri-cli": minor:feat
+---
+
+Add a cli command to `remove` plugins from projects.
--- a/.changes/cli-remove-command.md
+++ b/.changes/cli-remove-command.md
@@ -1,5 +1,6 @@
 ---
 "tauri-cli": minor:feat
+"@tauri-apps/cli": minor:feat
 ---
 
-Add a cli command to `remove` plugins from projects.
+Add `tauri remove` to remove plugins from projects.
--- a/crates/tauri-cli/src/remove.rs
+++ b/crates/tauri-cli/src/remove.rs
@@ -22,9 +22,6 @@ use std::process::Command;
 pub struct Options {
   /// The plugin to remove.
   pub plugin: String,
-  /// Don't format code with rustfmt
-  #[clap(long)]
-  pub no_fmt: bool,
 }
 
 pub fn command(options: Options) -> Result<()> {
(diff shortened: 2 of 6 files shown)


crates/tauri-cli/src/remove.rs
```diff
@@ -0,0 +1,118 @@
...
+    let _ = acl::permission::rm::command(acl::permission::rm::Options {
+      identifier: format!("{plugin}:default"),
+    });
+  }
+
+  // remove plugin init code from main.rs or lib.rs
+  let plugin_init_fn = if plugin == "stronghold" {
+    "Builder::new(|pass| todo!()).build()"
+  } else if plugin == "localhost" {
+    "Builder::new(todo!()).build()"
```
amrbashir: `tauri add` command, adds `todo!()` because user are requested to add code here, so this code likely changes and so our check here won't work.

crates/tauri-cli/src/remove.rs
```diff
@@ -0,0 +1,118 @@
...
+    "Builder::new().build()"
+  } else {
+    "init()"
+  };
+  let plugin_init = format!(".plugin(tauri_plugin_{plugin_snake_case}::{plugin_init_fn})");
+
+  for file in [tauri_dir.join("src/main.rs"), tauri_dir.join("src/lib.rs")] {
+    let contents = std::fs::read_to_string(&file)?;
+
+    if contents.contains(&plugin_init) {
```
amrbashir: this is a naive way to check if the plugin code exists or not, and it doesn't take into account that the user may have formatted their code or not.

amrbashir: the naive way is fine for `tauri add` because it is an additive operation not destructive, unlike this `rm` command so we need to be more careful, either by enhancing the detection by 100% or take user input on if the detected line should be removed or not.

amrbashir (changes requested):

crates/tauri-cli/src/remove.rs
ShaunSHamilton: The only robust way I can think of is to use `syn` to turn the code into an AST that can have the node(s) removed. Would adding that be acceptable?

amrbashir: yeah sure, `syn` is fine as well. We use a similar approach to modify js/ts files in `tauri migrate`.

lucasfernog: it's pretty much impossible to remove all plugin references from the code - I'd just leave that to the user
because there's trait imports, api usage etc

or are we just talking about removing the plugin init() call?

ShaunSHamilton: @lucasfernog I suppose that is a good point. If we are unlikely to be able to remove everything, should we automagically remove anything?

.changes/cli-remove-command.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri-cli": minor:feat
+---
+
+Add a cli command to `remove` plugins from projects.
```
amrbashir:
```suggestion
"tauri-cli": minor:feat
"@tauri-apps/cli": minor:feat
---

Add `tauri remove` to remove plugins from projects.
```

crates/tauri-cli/src/remove.rs
```diff
@@ -0,0 +1,72 @@
...
+      .map(PackageManager::from_project)
+      .and_then(|managers| managers.into_iter().next())
+    {
+      let npm_name = format!("@tauri-apps/plugin-{plugin}");
+      manager.remove(&[npm_name], tauri_dir)?;
+    }
+
+    let _ = acl::permission::rm::command(acl::permission::rm::Options {
+      identifier: format!("{plugin}:default"),
+    });
```
amrbashir: We shouldn't ignore the error here and instead propagate it.

Also there maybe other permissions used that is not necissarily `{plugin}:default`, we should remove these as well.

amrbashir (changes requested):

crates/tauri-cli/src/remove.rs
ShaunSHamilton: The only way I could think to keep the current CLI usage was to add the ability for `<plugin-name>:*` to be used with `tauri permission rm`

crates/tauri-cli/src/remove.rs
```diff
@@ -0,0 +1,72 @@
...
+      .map(PackageManager::from_project)
+      .and_then(|managers| managers.into_iter().next())
+    {
+      let npm_name = format!("@tauri-apps/plugin-{plugin}");
+      manager.remove(&[npm_name], tauri_dir)?;
+    }
+
+    let _ = acl::permission::rm::command(acl::permission::rm::Options {
+      identifier: format!("{plugin}:*"),
+    });
```
amrbashir: we should propagate the error not silence it

```suggestion
    acl::permission::rm::command(acl::permission::rm::Options {
      identifier: format!("{plugin}:*"),
    })?;
```

amrbashir (changes requested):

amrbashir (approved):

amrbashir:
Thank you

PR #11333, opened 2024-10-12
docs(android): warn users that passing raw bytes over ipc is not possible on android

kartonrad:
Due to platform limitations, passing raw bytes to tauri commands is not possible :(
(no support for custom ipc)


crates/tauri/src/ipc/mod.rs
```diff
@@ -132,6 +136,9 @@ impl InvokeResponseBody {
 }
 
 /// The IPC request.
+///
+/// Request can be the parameter of a tauri command.
```
lucasfernog:
```suggestion
/// Includes the `body` and `headers` parameters of a Tauri command invocation.
```

lucasfernog (approved):

PR #11406, opened 2024-10-18
docs: Add missing info regarding the drop position offset.

shi-yan:

### DIFF
--- a/packages/api/src/webview.ts
+++ b/packages/api/src/webview.ts
@@ -589,6 +589,11 @@ class Webview {
    *
    * @returns A promise resolving to a function to unlisten to the event.
    * Note that removing the listener is required if your listener goes out of scope e.g. the component is unmounted.
+   * When the debugger panel is open, the drop position of this event may be inaccurate due to a known limitation.
+   * To retrieve the correct drop position, please detach the debugger.
+   * Note: Even without the debugger open, there is still a vertical offset in the drop position.
+   * This occurs because the origin point is the top-left corner of the window (including the title bar), 
+   * not the top-left corner of the page, which may be expected.
    */
   async onDragDropEvent(
     handler: EventCallback<DragDropEvent>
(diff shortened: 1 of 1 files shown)


packages/api/src/webview.ts
```diff
@@ -589,6 +589,11 @@ class Webview {
    *
    * @returns A promise resolving to a function to unlisten to the event.
    * Note that removing the listener is required if your listener goes out of scope e.g. the component is unmounted.
+   * When the debugger panel is open, the drop position of this event may be inaccurate due to a known limitation.
+   * To retrieve the correct drop position, please detach the debugger.
+   * Note: Even without the debugger open, there is still a vertical offset in the drop position.
+   * This occurs because the origin point is the top-left corner of the window (including the title bar), 
+   * not the top-left corner of the page, which may be expected.
```
FabianLars: not sure if we should document this part. So far we didn't document bugs (that can be fixed) 🤔

shi-yan: I want to document because based on the discussion https://github.com/tauri-apps/tauri/discussions/11141

> but i'm not sure if we can really fix that.

I had the impression that this was a hard limitation of webview. I hope this is documented because people keep asking the same question again and again. 

And to be honest, I almost gave up using Tauri, had you never told me it was the debugger, because I thought the current drag-drop in Tauri was very buggy, and the offset is a deal breaker for what I want to build.

FabianLars: I am not talking about the debugger, that should be documented. I'm talking about the second half that's talking about the small offset that's there even when the debugger is closed. We can include your changes as they are and remove this part once it's fixed too, i don't care too much.

amrbashir: we shouldn't document that bug (non-debugger), and we should actively try to fix the problem

FabianLars (approved):

PR #11414, opened 2024-10-19
feat(cli) Make `tauri migrate` update $schema in tauri.conf.json

Czxck001:
This makes `tauri migrate` command update the `$schema` field in `tauri.conf.json`. For projects created under v1, it could be something like `../node_modules/@tauri-apps/cli/schema.json` (for projects with npm), but now it needs to be updated to v2's schema, which is unified under URL `https://schema.tauri.app/config/2`.


lucasfernog (approved):
nice catch, thanks!

Issue #9039, opened 2024-02-29
[bug] Tauri v2 has constrained compatibility on Linux


hjmallon:
### Describe the bug

(I guess this is just that the minimal version requirements for v2 are > RHEL9 and this is probably wontfix)?

With Tauri v1 we were building with CentOS7. I have just tried to update to Tauri v2 and have had to confront a couple of problems.

1. Tauri v2 requires node v18+ (which seems fair), installing this on CentOS7 is not easy since nodejs dropped support for CentOS 7 in v18 in their upstream builds. There are some unsupported builds, but only for x86_64, not aarch64.
2. Tauri v2 uses a more recent webkit (https://github.com/tauri-apps/tauri-docs/blob/425563db3849ceed10fcc5d1b6d73ea6bc7f1ec4/src/content/docs/blog/tauri-2-0-0-alpha-3.md?plain=1#L32)

In order to fix the first problem I decided to build on CentOS9 stream, which is quite similar to our target platform. However CentOS9 stream uses glib2 2.68.4. Unfortunately the rust crate soup3 requires gio 2.70 and so I can no longer build without moving to Fedora or Ubuntu. (Ubuntu 20.04 which is still in support has the same issue).

### Reproduction

Try to build an example app on CentOS 9 Stream, or Rocky 9, or Ubuntu 20.04.

### Expected behavior

I would expect Tauri v2 to be capable of building on modern enterprise Linux and/or this stuff to be listed in the `What other breaking changes we are going to expect?` section here https://github.com/tauri-apps/tauri-docs/blob/425563db3849ceed10fcc5d1b6d73ea6bc7f1ec4/src/content/docs/blog/tauri-2-0-0-alpha-3.md?plain=1#L32

### Full `tauri info` output

```text
$ yarn tauri info
yarn run v1.22.21
$ tauri info

[✘] Environment
    - OS: CentOS 9.0.0 X64
    ✘ webkit2gtk-4.1: not installed
      Visit https://tauri.app/v1/guides/getting-started/prerequisites to learn more about tauri prerequisites
    ✔ rsvg2: 2.50.7
    ✔ rustc: 1.75.0 (82e1608df 2023-12-21)
    ✔ cargo: 1.75.0 (1d8b05cdd 2023-11-20)
    ⚠ rustup: not installed!
      If you have rust installed some other way, we recommend uninstalling it
      then use rustup instead. Visit https://rustup.rs/
    ⚠ Rust toolchain: couldn't be detected!
[...]

PR #11429, opened 2024-10-20
enhance: always check parent directory (fix #8679)

### LINKED ISSUE #8679: [bug] tauri resource directory is hard-coded

griffi-gh:
closes #8679

### DIFF
--- a/crates/tauri-utils/src/platform.rs
+++ b/crates/tauri-utils/src/platform.rs
@@ -309,12 +309,13 @@ fn resource_dir_from<P: AsRef<std::path::Path>>(
 
   #[cfg(target_os = "linux")]
   {
-    res = if curr_dir.ends_with("/data/usr/bin") {
+    // (canonicalize checks for existence, so there's no need for an extra check)
+    res = if let Ok(bundle_dir) = exe_dir
+      .join(format!("../lib/{}", package_info.name))
+      .canonicalize()
+    {
       // running from the deb bundle dir
-      exe_dir
-        .join(format!("../lib/{}", package_info.name))
-        .canonicalize()
-        .map_err(Into::into)
+      Ok(bundle_dir)
     } else if let Some(appdir) = &env.appdir {
       let appdir: &std::path::Path = appdir.as_ref();
       Ok(PathBuf::from(format!(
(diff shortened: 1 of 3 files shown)


crates/tauri-utils/src/platform.rs
```diff
@@ -309,12 +309,13 @@ fn resource_dir_from<P: AsRef<std::path::Path>>(
 
   #[cfg(target_os = "linux")]
   {
-    res = if curr_dir.ends_with("/data/usr/bin") {
+    // (canonicalize checks for existence, so there's no need for an extra check)
+    res = if let Ok(bundle_dir) = exe_dir
+      .join(format!("../lib/{}", package_info.name))
+      .canonicalize()
+    {
       // running from the deb bundle dir
```
lucasfernog:
```suggestion
```

lucasfernog (dismissed):
this should be fine but i'd also like input from @tweidinger since we're no longer checking the path we're at this allows the binary/resources to be anywhere in the filesystem - as long as the resources are in ../lib/name

tweidinger (dismissed):
I think it is fine to not ensure the path ends with the `/data/usr/bin` suffix, since the initial path is derived from the executable path and I don't think we make security specific promises of the resource dir.

lucasfernog (approved):

absassi:
Thanks for the fix.
Would it be possible to backport this to V1 as well? At the moment, I can't upgrade to V2 due to #9039.

FabianLars:
Sure, it's possible and we still accept prs for v1 :)

MarielaSassi:
> Sure, it's possible and we still accept prs for v1 :)

Backport done, please have a look. @FabianLars 
https://github.com/tauri-apps/tauri/pull/11960

PR #11451, opened 2024-10-22
feat: add `WebviewWindow/Webview::devtools`

### LINKED ISSUE #10849: [feat] Enable devtools only for certain webviews
### Describe the problem

Currently the `devtools` features enables the webview devtools for all webviews.

https://github.com/tauri-apps/tauri/blob/5ec74456b9ae45eaa63e225856de1b7eb83abe6f/crates/tauri-runtime-wry/src/lib.rs#L4137-L4140

I want to enable it for only a certain webview window. My use case is, I have an app that has a canvas and a manager. I want to have devtools in the canvas window in the release build so that users can inspect the elements, but not for the manager window.

I noticed that `WebviewBuilder` in `wry` has the [`with_devtools`](https://docs.rs/wry/latest/wry/struct.WebViewBuilder.html#method.with_devtools) method. Is it possible to have the same on `WebviewWindowBuilder` in `tauri`?

### Alternatives considered

_No response_

### Additional context

_No response_

amrbashir:
closes #10849


FabianLars:
Can we include in the docs that it's enabled by default?

lucasfernog (dismissed):

lucasfernog (approved):

PR #11485, opened 2024-10-24
feat: add HeaderConfig to SecurityConfig

39zde:
# Feature
Adds a new configuration option for the tauri configuration file. This being headers in the app>security. Headers defined the are added to every http response from tauri to the web view. This doesn't include IPC messages and error responses. The header names are limited to:
  

  - `Access-Control-Allow-Credentials`
  - `Access-Control-Allow-Headers`
  - `Access-Control-Allow-Methods`
  - `Access-Control-Expose-Headers`
  - `Access-Control-Max-Age`
  - `Cross-Origin-Embedder-Policy`
  - `Cross-Origin-Opener-Policy`
  - `Cross-Origin-Resource-Policy`
  - `Permissions-Policy`
  - `Timing-Allow-Origin`
  - `X-Content-Type-Options`
  - `Tauri-Custom-Header`

I primarily wanted to use [SharedArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer),
which requires cross-origin isolation. Since there was no effort in adding more headers I looked for the ones, that would make the most sense.
The Content-Security-Policy(CSP) remains untouched. I tried to implement a unified way to define headers, including the CSP, but to no avail.
Since it's a very dynamic header, with grave implications for security, it's better to remain untouched.
## Example configuration
 ```javascript
 {
  //..
   app:{
     //..
     security: {
       headers: {
         "Cross-Origin-Opener-Policy": "same-origin",
         "Cross-Origin-Embedder-Policy": "require-corp",
         "Timing-Allow-Origin": [
           "https://developer.mozilla.org",
           "https://example.com",
         ],
         "Access-Control-Expose-Headers": "Tauri-Custom-Header",
         "Tauri-Custom-Header": {
           "key1": "'value1' 'value2'",
           "key2": "'value3'"
         }
       },
       csp: "default-src 'self'; connect-src ipc: http://ipc.localhost",
     }
     //..
   }
  //..
 }
 ```
 In this example `Cross-Origin-Opener-Policy` and `Cross-Origin-Embedder-Policy` are set to allow for the use of [`SharedArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer).
 The result is, that those headers are then set on every response sent via the `get_response` function in crates/tauri/src/protocol/tauri.rs.
 The Content-Security-Policy header is defined separately, because it is also handled separately.

 For the helloworld example, this config translates into those response headers:
 ```http
 access-control-allow-origin:  http://tauri.localhost
 access-control-expose-headers: Tauri-Custom-Header
 content-security-policy: default-src 'self'; connect-src ipc: http://ipc.localhost; script-src 'self' 'sha256-Wjjrs6qinmnr+tOry8x8PPwI77eGpUFR3EEGZktjJNs='
 content-type: text/html
 cross-origin-embedder-policy: require-corp
 cross-origin-opener-policy: same-origin
 tauri-custom-header: key1 'value1' 'value2'; key2 'value3'
 timing-allow-origin: https://developer.mozilla.org, https://example.com
 ```
 Since the resulting header values are always 'string-like'. So depending on the what data type the HeaderSource is, they need to be converted.
  - `String`(JS/Rust): stay the same for the resulting header value
  - `Array`(JS)/`Vec\<String\>`(Rust): Item are joined by ", " for the resulting header value
  - `Object`(JS)/ `Hashmap\<String,String\>`(Rust): Items are composed from: key + space + value. Item are then joined by "; " for the resulting header value


crates/tauri-utils/src/config/security.rs
```diff
@@ -0,0 +1,783 @@
...
+impl From<HashMap<String, String>> for HeaderSource {
+  fn from(value: HashMap<String, String>) -> Self {
+    Self::Map(value)
+  }
+}
+
+/// A trait which implements on the [`Builder`] of the http create
+///
+/// Must add headers defined in the tauri configuration file to http responses
+pub trait HeaderAddition {
```
39zde: That's, why the http 1.1.0 dependency was added.
I think there's a way around it by defining the trait more like this:
```rust
pub trait HeaderAddition<T> {
   fn add_configured_headers(self, headers: Option<&HeaderConfig>) -> T;
}
```
and implementing the HeaderAddition somewhere, where http is already a dependency.
But I don't know, where that would be.

crates/tauri-utils/src/config/security.rs
39zde, line 1: this whole file may?/can?/should? be incorporated into crates/tauri-utils/src/config.rs

crates/tauri-utils/src/config/security.rs
amrbashir: that wouldn't work because of orphan rule imposed by the compiler, either the author of `http` crate implements `HeaderAddition` trait or author of `HeaderAddition` trait implements for `http` crate, otherwise rust will error out.

This is fine to be kept here.

crates/tauri-utils/src/config/security.rs
```diff
@@ -0,0 +1,783 @@
...
+  ///
+  /// this is a utility function, which is used in the same way as the `.header(..)` of the rust http library
+  fn add_configured_headers(mut self, headers: Option<&HeaderConfig>) -> Builder {
+    self = match headers {
+      Some(headers) => {
+        // Add the header Access-Control-Allow-Credentials, if we find a value for it
+        self = match &headers.access_control_allow_credentials {
+          Some(value) => self.header("Access-Control-Allow-Credentials", value.to_string()),
+          None => self,
+        };
```
amrbashir: this can be simplified into
```suggestion
        if let Some(value) match &headers.access_control_allow_credentials {
          self = self.header("Access-Control-Allow-Credentials", value.to_string());
        };
```
and same should be done for other headers

crates/tauri-utils/src/config/security.rs
```diff
@@ -0,0 +1,783 @@
...
+  fn add_configured_headers(self, headers: Option<&HeaderConfig>) -> Builder;
+}
+
+impl HeaderAddition for Builder {
+  /// Add the headers defined in the tauri configuration file to http responses
+  ///
+  /// this is a utility function, which is used in the same way as the `.header(..)` of the rust http library
+  fn add_configured_headers(mut self, headers: Option<&HeaderConfig>) -> Builder {
+    self = match headers {
+      Some(headers) => {
```
amrbashir: this also can be simplified to 

```rs
if let Some(headers) = headers {
  // ... rest of logic
}
```

crates/tauri-utils/src/config/security.rs
```diff
@@ -0,0 +1,783 @@
...
+}
+
+impl Display for HeaderConfig {
+  fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+
+        write!(f,"Access-Control-Allow-Credentials: ")?;
+        match &self.access_control_allow_credentials {
+          Some(value) => writeln!(f, "{}", value)?,
+          None => writeln!(f, "null")?,
+        };
```
amrbashir: This also can be simplified into
```suggestion
        writeln!(f,"Access-Control-Allow-Credentials: {}", self.access_control_allow_credentials .unwrap_or("null"))?;
```

crates/tauri-utils/src/config/security.rs
```diff
@@ -0,0 +1,783 @@
...
+      cross_origin_resource_policy: None,
+      permissions_policy: None,
+      timing_allow_origin: None,
+      x_content_type_options: None,
+      tauri_custom_header: None,
+    }
+  }
+}
+
+impl Display for HeaderConfig {
```
amrbashir: is there a reason why we implementing the `Display` trait?

amrbashir (changes requested):
Could you please revert the changes where you moved types from `config.rs` to `config/security.rs` so the relevant PRs could be reviewed separately, after merge, your can open another PR to refactor things into different files.

.changes/add-headers-config-option.md
```diff
@@ -0,0 +1,78 @@
...
+"tauri-runtime": 'minor:feat'
+"tauri-runtime-wry": 'minor:feat'
+"tauri-codegen": 'minor:feat'
+"tauri-macros": 'minor:feat'
+"tauri-plugin": 'minor:feat'
+"tauri-build": 'minor:feat'
+"tauri": 'minor:feat'
+"@tauri-apps/cli": 'minor:feat'
+"tauri-cli": 'minor:feat'
+"tauri-driver": 'minor:feat'
```
amrbashir: This change only concerns `tauri-utils` and `tauri`
```suggestion
"tauri-utils": 'minor:feat'
"tauri": 'minor:feat'
```

.changes/add-headers-config-option.md
```diff
@@ -0,0 +1,78 @@
...
+content-type: text/html
+cross-origin-embedder-policy: require-corp
+cross-origin-opener-policy: same-origin
+tauri-custom-header: key1 'value1' 'value2'; key2 'value3'
+timing-allow-origin: https://developer.mozilla.org, https://example.com
+```
+Since the resulting header values are always 'string-like'. So depending on the what data type the HeaderSource is, they need to be converted.
+ - `String`(JS/Rust): stay the same for the resulting header value
+ - `Array`(JS)/`Vec\<String\>`(Rust): Item are joined by ", " for the resulting header value
+ - `Object`(JS)/ `Hashmap\<String,String\>`(Rust): Items are composed from: key + space + value. Item are then joined by "; " for the resulting header value
```
amrbashir: This is a bit verbose for a change log entry, let's reword into 2~5 lines at max, examples and extensive documentation should be in the inline documentation

crates/tauri-utils/src/config.rs
```diff
@@ -1773,6 +1773,404 @@ pub struct AssetProtocolConfig {
...
+    /// The dir containing the index.html file that contains the secure isolation application.
+    dir: PathBuf,
+  },
+}
+
+impl Default for PatternKind {
+  fn default() -> Self {
+    Self::Brownfield
+  }
+}
```
amrbashir: Let's move these back to where they were so the git diff of this PR doesn't include any changes to them.

crates/tauri-utils/src/config.rs
```diff
@@ -1773,6 +1773,404 @@ pub struct AssetProtocolConfig {
...
+pub trait HeaderAddition {
+  /// adds all headers defined on the config file, given the current HeaderConfig
+  fn add_configured_headers(self, headers: Option<&HeaderConfig>) -> Builder;
+}
+
+impl HeaderAddition for Builder {
+  /// Add the headers defined in the tauri configuration file to http responses
+  ///
+  /// this is a utility function, which is used in the same way as the `.header(..)` of the rust http library
+  fn add_configured_headers(mut self, headers: Option<&HeaderConfig>) -> Builder {
```
amrbashir: It is not clear from the first glance which `Builder` we referring to
```suggestion
  fn add_configured_headers(mut self, headers: Option<&HeaderConfig>) -> http::response::Builder {
```

crates/tauri-utils/src/config.rs
```diff
@@ -1773,6 +1773,404 @@ pub struct AssetProtocolConfig {
...
+    if let Some(headers) = headers {
+
+        // Add the header Access-Control-Allow-Credentials, if we find a value for it
+        if let Some(value) match &headers.access_control_allow_credentials {
+          self = self.header("Access-Control-Allow-Credentials", value.to_string());
```
amrbashir: passing `&str` should be fine here, no need to call `to_string`
```suggestion
          self = self.header("Access-Control-Allow-Credentials", value.as_str());
```

crates/tauri-utils/src/config.rs
```diff
@@ -1773,6 +1773,404 @@ pub struct AssetProtocolConfig {
...
+        if let Some(value) match &headers.tauri_custom_header {
+          // Keep in mind to correctly set the Access-Control-Expose-Headers
+          self = self.header("Tauri-Custom-Header", value.to_string());
+        };
+      }
+    self
+  }
+}
+
+/// ## Header Config
```
amrbashir:
```suggestion
```

amrbashir (changes requested):

crates/tauri-utils/src/config.rs
39zde: The `HeaderSource` value is not always a `String`, so we can't just slice the value with `as_str`. A new String gets created instead, following the rules defined in `enum HeaderSource`
```rust
pub enum HeaderSource {
  /// string version of the header Value
  Inline(String),
  /// list version of the header value. Item are joined by "," for the real header value
  List(Vec<String>),
  /// (Rust struct | Json | JavaScript Object) equivalent of the header value. Items are composed from: key + space + value. Item are then joined by ";" for the real header value
  Map(HashMap<String, String>),
}
```

amrbashir: tbh I am not sure why we would support all of these variants, shouldn't be the `Inline` variant enough? it is also simple to document and simple for users to reason about

39zde: Wouldn't the same be true for the Content-Security-Policy (csp)?
I used the csp as a guide to design this option.

.changes/add-headers-config-option.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri-utils": 'minor:feat'
+"tauri": 'minor:feat'
+---
+Adds a new configuration option `headers` under >app>security for the tauri configuration file. Headers defined there are added to every http response from tauri to the web view. This doesn't include IPC messages and error responses. The `Content-Security-Policy`(CSP) remains untouched and still is defined separately.
```
amrbashir:
```suggestion
Adds a new configuration option `app > security > headers` to define headers that will be added to every http response from tauri to the web view. This doesn't include IPC messages and error responses.
```

crates/tauri-utils/src/config.rs
```diff
@@ -1773,6 +1774,358 @@ pub struct AssetProtocolConfig {
...
+  fn from(value: Vec<String>) -> Self {
+    Self::List(value)
+  }
+}
+
+impl From<HashMap<String, String>> for HeaderSource {
+  fn from(value: HashMap<String, String>) -> Self {
+    Self::Map(value)
+  }
+}
```
amrbashir: I don't see this implementations used anywhere, right? we can remove them in that case

amrbashir (changes requested):

crates/tauri-utils/src/config.rs
amrbashir: oh, it is fine to keep then

amrbashir (dismissed):
LGTM, just need approval of @tweidinger 

Note you need to run `cargo b -p tauri-schema-generator` to fix CI

tweidinger (approved):
Looks good to me.

PR #11486, opened 2024-10-24
feat: background color APIs

### LINKED ISSUE #1564: Allow changing the webview background color
**Describe the bug**
Window flashes a white background before rendering app content on macOS Big Sur.

The css for the `index.html` is:

```css
html, body {
  width: 100%; height: 100%;
  background: rgb(231, 232, 232);
}

@media (prefers-color-scheme: dark) {
  html, body {
    background: rgb(37, 38, 40);
  }
}
```

**To Reproduce**
1. Create an index.html with `<style>html, body { background: black; }</style>`
2. Run the app (dev, or dist bundle)
3. Screen will flash a white background before rendering the content.

**Expected behavior**
The main window should wait for the content to load, or it should apply the page's background color on open.

**Screenshots**

https://user-images.githubusercontent.com/1578018/115579993-b0ccbf80-a294-11eb-8c68-07898105ab25.mp4

OS: Mac OS, version 11.2.3 X64
Node: 15.13.0
NPM: 7.7.6
Yarn: 1.22.10
Rustc: 1.51.0

**Additional context**
This is running the app with the default values (no transparent window mode). When using transparent window mode, it doesn't flash, but the window loses all shadows and borders until it is resized.

https://user-images.githubusercontent.com/1578018/115580068-bfb37200-a294-11eb-832d-9639839ae126.mp4

### LINKED ISSUE #10519: [feat] set window background color
### Describe the problem

In the multiwebview example, the areas that are not drawn are filled with white.

multiwebview

Add `Window::set_background_color(&self, color: Color)` method

### Alternatives considered

_No response_

### Additional context

_No response_

amrbashir:
closes #10519
closes #1564

needs: https://github.com/tauri-apps/tao/pull/995


lucasfernog (approved):

dionysuzx:
is it possible to set it different on dark vs light mode? @amrbashir  (on backend, so like when it starts up, fetching first from system)

amrbashir:
you'd have to get the theme first before creating the window and then create the window with the color you want through Rust not in the config.

dionysuzx:
> you'd have to get the theme first before creating the window and then create the window with the color you want through Rust not in the config.

this won't work on macOS or iOS right? out of curiosity what is the blocker or path to getting it implemented for those systems? cc @amrbashir (thank you for your work on this btw!)

amrbashir:
I don't have a macOS so I can't test anything but there is https://developer.apple.com/documentation/webkit/wkwebview/3850574-underpagebackgroundcolor?language=objc that we could use but only supported on macOS 12+ and iOS 15+, so we need an alternative for older systems versions. This just needs to be implemented in `wry` repo (and `tao` if needed) and should automatically be supported in tauri afterwards.

PR #11514, opened 2024-10-28
feat(cli): remove UNC path prefix in `TAURI_APP_PATH` and `TAURI_FRONTEND_PATH`

bizarre:
When these env vars are configured on windows, a lot breaks because the path starts with `\\?\`:
> "CMD.EXE was started with the above path as the current directory.
UNC paths are not supported. Defaulting to Windows directory."

This just applies the same changes made in https://github.com/tauri-apps/tauri/pull/9420 but to the new `env_tauri_app_path` and `env_tauri_frontend_path` pathbufs, using [`dunce::simplified`](https://docs.rs/dunce/latest/dunce/fn.simplified.html).

### DIFF
--- a/crates/tauri-cli/src/helpers/app_paths.rs
+++ b/crates/tauri-cli/src/helpers/app_paths.rs
@@ -17,8 +17,6 @@ use tauri_utils::{
   platform::Target,
 };
 
-use dunce;
-
 const TAURI_GITIGNORE: &[u8] = include_bytes!("../../tauri.gitignore");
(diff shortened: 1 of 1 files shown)


crates/tauri-cli/src/helpers/app_paths.rs
```diff
@@ -17,6 +17,8 @@ use tauri_utils::{
   platform::Target,
 };
 
+use dunce;
+
```
amrbashir:
```suggestion
```

amrbashir (approved):

PR #11521, opened 2024-10-28
feat(cli): process `bundle > windows > wix > fragmentPaths` with Handlebars

### LINKED ISSUE #11520: Run tauri project as a windows service
### Describe the problem

I aim to run **my Tauri project** as a **Windows service**. To achieve this, I need to use the WiX Util extension, which includes the ServiceInstall element necessary for setting up Windows services.

Here’s the code I am trying to use:
```
<DirectoryRef Id="INSTALLDIR">
...
 <Component Id="Path" Guid="{{path_component_guid}}" Win64="$(var.Win64)">
                <File Id="Path" Source="{{app_exe_source}}" KeyPath="yes" Checksum="yes"/>
                 <ServiceInstall Id="AgentUiServiceInstaller" Type="ownProcess" Vital="yes" Name="GetAppAgentUIService" DisplayName="GetAppAgentUIService" Description="GetApp's Agent UI is running here" Start="demand" Account="LocalSystem" ErrorControl="ignore" Interactive="no">
                    <util:ServiceConfig FirstFailureActionType='restart' SecondFailureActionType='restart' ThirdFailureActionType='restart' RestartServiceDelayInSeconds='60' ResetPeriodInDays='1'/>
                </ServiceInstall>

                <ServiceControl Id="AgentUiServiceControl" Name="GetAppAgentUIService" Stop="both" Remove="uninstall" Start="install" Wait="yes" />
            </Component>
...
</DirectoryRef>
```

## The problem
The current Tauri template does not support adding the WiX Util extension in the main template, so I am attempting to use a custom fragment to include it. However, this workaround has significant limitations:

### 1. Installing the Full Component via Custom Fragment:
The ```.exe``` file cannot be installed directly through the custom fragment because there is no way to reference the ```app_exe_source``` variable within the fragment. The custom fragment does not support accessing bundler variables like ```{{app_exe_source}}```, which limits its usability.

### 2. Installing the Service Without the File Element:
If I attempt to install the service without specifying the file element, there is no way to link the service to the executable file. As a result, the service configuration cannot specify which file to execute when starting the service.

Ideally, there would be two potential solutions, both of which seem to be quite simple to implement::

1. Enable support for the WiX Util extension directly in the main Tauri template, like you doing with the custom fragment.
2. Allow custom fragments to access bundler variables like ```{{app_exe_source}}```, enabling the fragment to correctly reference the ```.exe``` file and establish a link between the service and the executable.

### Alternatives considered

_No response_

### Additional context

_No response_

amrbashir:
closes #11520

This PR also includes a small refactor of using `register_template_string` followed by `render` to use `render_template` directly.

### DIFF
--- a/crates/tauri-bundler/src/bundle/windows/msi/mod.rs
+++ b/crates/tauri-bundler/src/bundle/windows/msi/mod.rs
@@ -726,38 +726,26 @@ pub fn build_wix_app_installer(
     );
 
     // Create the update task XML
-    let mut skip_uac_task = Handlebars::new();
+    let skip_uac_task = Handlebars::new();
     let xml = include_str!("./update-task.xml");
-    skip_uac_task
-      .register_template_string("update.xml", xml)
-      .map_err(|e| e.to_string())
-      .expect("Failed to setup Update Task handlebars");
+    let update_content = skip_uac_task.render_template(xml, &data)?;
     let temp_xml_path = output_path.join("update.xml");
-    let update_content = skip_uac_task.render("update.xml", &data)?;
     fs::write(temp_xml_path, update_content)?;
 
     // Create the Powershell script to install the task
     let mut skip_uac_task_installer = Handlebars::new();
     skip_uac_task_installer.register_escape_fn(handlebars::no_escape);
     let xml = include_str!("./install-task.ps1");
-    skip_uac_task_installer
-      .register_template_string("install-task.ps1", xml)
-      .map_err(|e| e.to_string())
-      .expect("Failed to setup Update Task Installer handlebars");
+    let install_script_content = skip_uac_task_installer.render_template(xml, &data)?;
     let temp_ps1_path = output_path.join("install-task.ps1");
-    let install_script_content = skip_uac_task_installer.render("install-task.ps1", &data)?;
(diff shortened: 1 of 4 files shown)


crates/tauri-bundler/src/bundle/windows/msi/mod.rs
```diff
@@ -772,7 +760,9 @@ pub fn build_wix_app_installer(
   let extension_regex = Regex::new("\"http://schemas.microsoft.com/wix/(\\w+)\"")?;
   for fragment_path in fragment_paths {
     let fragment_path = current_dir.join(fragment_path);
-    let fragment = fs::read_to_string(&fragment_path)?;
+    let fragment_content = fs::read_to_string(&fragment_path)?;
+    let fragment_handle_bars = Handlebars::new();
+    let fragment = fragment_handle_bars.render_template(&fragment_content, &data)?;
```
lucasfernog:
```suggestion
    let fragment_handlebars = Handlebars::new();
    let fragment = fragment_handlebars.render_template(&fragment_content, &data)?;
```

lucasfernog (approved):

PR #11522, opened 2024-10-28
enhance(core): use `diagnostic::on_unimplemented` on rustc 1.78 and newer for async commands with references

amrbashir:


lucasfernog (approved):
that's clever!

PR #11575, opened 2024-11-03
enhance: include the path in ACL I/O errors

kornelski:
Errors about files missing that don't include the file path are difficult to diagnose:

>  failed to read plugin permissions: failed to read file: No such file or directory (os error 2)

I've added paths to ACL's I/O errors.

### DIFF
--- a/.changes/utils-acl-path-errors.md
+++ b/.changes/utils-acl-path-errors.md
@@ -1,0 +1,6 @@
+---
+"tauri": minor:enhance
+"tauri-utils": minor:enhance
+---
+
+Include the path in ACL I/O errors.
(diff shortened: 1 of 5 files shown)


FabianLars:
Thanks! I agree that this would be a valuable change but since it's exposed here https://docs.rs/tauri-utils/latest/tauri_utils/acl/enum.Error.html it would technically be a breaking change 🤔 @amrbashir any idea how to handle this?

p.s. we also need a change file before we can merge this

amrbashir:
This shouldn't be a breaking change since the `acl` module is marked as instable, see https://docs.rs/tauri-utils/latest/tauri_utils/acl/index.html#stability

amrbashir (approved):
Thank you

FabianLars:
ah didn't see that, cool

PR #11581, opened 2024-11-04
fix(event): handle AnyLabel listeners in `emit_to`

### LINKED ISSUE #11561: [bug] v2 Event system: AnyLabel isn't handled correct
### Describe the bug

The new event system, which primarily relies on `target`s to work, has some major flaws when it comes to handling these targets. Specifically targets of the kind `AnyLabel`, which is the default when just adding a string as target in both `emitTo` and `listen`.

For example, if a listener is set with the target `"main"`, and `emitTo("main", "eventName", "payload")` is run, the listener will not be triggered.

I have tried to trace this behavior back to the source, and have stumbled upon this snippet from the implementation of the `emit_to` function in Rust.
```rust
// https://github.com/tauri-apps/tauri/blob/12ffc19ce08eeafdedc65207f312432aa2cf9c84/crates/tauri/src/manager/mod.rs#L590-L606
match target {
  // if targeting all, emit to all using emit without filter
  EventTarget::Any => self.emit(event, payload),

  // if targeting any label, emit using emit_filter and filter labels
  EventTarget::AnyLabel {
    label: target_label,
  } => self.emit_filter(event, payload, |t| match t {
    EventTarget::Window { label }
    | EventTarget::Webview { label }
    | EventTarget::WebviewWindow { label } => label == &target_label,
    _ => false,
  }),

  // otherwise match same target
  _ => self.emit_filter(event, payload, |t| t == &target),
}
```
Here, `target` is the target set in `emit_to`. When this is of the kind `AnyLabel`, the code will enter the second `match`, where it checks for `t` which is the listener target. It checks if `t` is of kind `Window`, `Webview` or `WebviewWindow`, none of which apply to `AnyLabel` by which it will return false when the listener is initialized without a specific kind.

If a kind has been specified for `emit_to`, it will go to the bottom line, and only match the same targets. Again, if the listener is initialized without a specific kind, and therefore gets the kind `AnyLabel`, it will not trigger the listener.

### Reproduction

_No response_

### Expected behavior

As far as I understand it, the expected behaviour would be to let `AnyLabel` serve as a "wildcard" for both `emitTo` and `listen`. So if any of the two are of the kind `AnyLabel`, any event with a label would be served between the two.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.22631 x86_64 (X64)
    ✔ WebView2: 130.0.2849.56
    - node: 20.15.1
    - yarn: 4.5.1
    - npm: 10.7.0
    - deno: deno 2.0.4

[-] Packages
    - tauri 🦀: 2.0.6
    - tauri-build 🦀: 2.0.2
    - wry 🦀: 0.46.3
    - tao 🦀: 0.30.5
    - @tauri-apps/api : 2.0.3
    - @tauri-apps/cli : 2.0.5

[-] Plugins
    - tauri-plugin-dialog 🦀: 2.0.3
    - @tauri-apps/plugin-dialog : 2.0.1
    - tauri-plugin-fs 🦀: 2.0.3
    - @tauri-apps/plugin-fs : 2.0.1
    - tauri-plugin-process 🦀: 2.0.1
    - @tauri-apps/plugin-process : 2.0.0
    - tauri-plugin-http 🦀: 2.0.3
    - @tauri-apps/plugin-http : 2.0.1

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:4321/
    - framework: Svelte
    - bundler: Rollup
```

### Stack trace

_No response_

### Additional context

I would be willing to work on a PR fixing this issue, but I would like to discuss the exact issue and expected behavior in depth before starting to change things in the event API.

Mikkel-T:
closes #11561 

This Pull Request changes how AnyLabel is interpreted when using the event system. For more info, see issue #11561

### DIFF
--- a/.changes/event-anylabel-fix.md
+++ b/.changes/event-anylabel-fix.md
@@ -2,4 +2,4 @@
 "tauri": "patch:bug"
 ---
 
-Handle AnyLabel targets correctly in the event system
+Fix listeners created with `EventTarget::AnyLabel` never receiving events.
--- a/.changes/event-anylabel-fix.md
+++ b/.changes/event-anylabel-fix.md
@@ -1,0 +1,5 @@
+---
+"tauri": "patch:bug"
+---
+
+Handle AnyLabel targets correctly in the event system
--- a/crates/tauri/src/manager/mod.rs
+++ b/crates/tauri/src/manager/mod.rs
@@ -597,7 +597,31 @@ impl<R: Runtime> AppManager<R> {
       } => self.emit_filter(event, payload, |t| match t {
         EventTarget::Window { label }
         | EventTarget::Webview { label }
-        | EventTarget::WebviewWindow { label } => label == &target_label,
+        | EventTarget::WebviewWindow { label }
+        | EventTarget::AnyLabel { label } => label == &target_label,
+        _ => false,
+      }),
+
+      EventTarget::Window {
+        label: target_label,
+      } => self.emit_filter(event, payload, |t| match t {
+        EventTarget::AnyLabel { label } | EventTarget::Window { label } => label == &target_label,
+        _ => false,
+      }),
+
+      EventTarget::Webview {
+        label: target_label,
+      } => self.emit_filter(event, payload, |t| match t {
+        EventTarget::AnyLabel { label } | EventTarget::Webview { label } => label == &target_label,
+        _ => false,
+      }),
+
+      EventTarget::WebviewWindow {
+        label: target_label,
+      } => self.emit_filter(event, payload, |t| match t {
+        EventTarget::AnyLabel { label } | EventTarget::WebviewWindow { label } => {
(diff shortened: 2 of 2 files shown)


amrbashir (changes requested):
Thank you, this LGTM, could you just add a change file in `.changes` directory?

.changes/event-anylabel-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri": "patch:bug"
+---
+
+Handle AnyLabel targets correctly in the event system
```
amrbashir:
```suggestion
Fix listeners created with `EventTarget::AnyLabel` never receiving events.
```

amrbashir (approved):
Thank you

PR #11583, opened 2024-11-04
fix(tray): build tray on main thread

### LINKED ISSUE #11462: [bug] TrayIcon mouse events don't work if the icon was created from async command
### Describe the bug

After upgrading my app to tauri v2 I could not make the tray icon behave as it did in v1 at all. It would not react to mouse at all, not even context menu would show up. Debugging showed that events were not firing.

After hours of trying things in a dummy create-tauri-app to isolate the issue I noticed that if the tray icon is created in a synchronous command it works as expected, but if the command is `async` then the icon will be created but it does not react to mouse at all.

This was not an issue in tauri v1.

### Reproduction

Use `npm create tauri-app@latest` and modify the `greet` command like this
```rust
// Learn more about Tauri commands at https://tauri.app/develop/calling-rust/
#[tauri::command]
async fn greet(name: &str, app: AppHandle) -> String {
    tauri::tray::TrayIconBuilder::with_id("TRAY")
        .icon(app.default_window_icon().unwrap().clone())
        .on_tray_icon_event(|_icon, event| {
            println!("Tray icon event received {:?}", event);
            match event {
                TrayIconEvent::Click {
                    button: MouseButton::Left,
                    button_state: MouseButtonState::Up,
                    ..
                } => {
                    println!("Tray icon clicked");
                }
                _ => {}
            }
        })
        .on_menu_event(|_app, _event| {})
        .build(&app)
        .ok();
    format!("Hello, {}! You've been greeted from Rust!", name)
}
```

Notice that the icon ignores mouse.
But if you drop the `async` then it works as expected.

### Expected behavior

Icon works properly with both sync and async commands.

### Full `tauri info` output

```text
[✘] Environment
    ✔ WebView2: 129.0.2792.89
    ✘ Couldn't detect any Visual Studio or VS Build Tools instance with MSVC and SDK components. Download from https://aka.ms/vs/17/release/vs_BuildTools.exe
    ✔ rustc: 1.80.0 (051478957 2024-07-21)
    ✔ cargo: 1.80.0 (376290515 2024-07-16)
    - node: 16.14.2
    - pnpm: 8.6.2
    - npm: 8.5.0

[-] Packages
    - tauri 🦀: 2.0.6      
    - tauri-build 🦀: 2.0.2
    - wry 🦀: 0.46.3
    - tao 🦀: 0.30.3
    - tauri-cli 🦀: 1.3.1
    - @tauri-apps/api : 2.0.0 (outdated, latest: 2.0.3)
    - @tauri-apps/cli : 2.0.3 (outdated, latest: 2.0.4)

[-] Plugins
    - tauri-plugin-notification 🦀: 2.0.1
    - @tauri-apps/plugin-notification : 2.0.0
    - tauri-plugin-dialog 🦀: 2.0.3
    - @tauri-apps/plugin-dialog : 2.0.1
    - tauri-plugin-fs 🦀: 2.0.3
    - @tauri-apps/plugin-fs : 2.0.1
    - tauri-plugin-cli 🦀: 2.0.1
    - @tauri-apps/plugin-cli : 2.0.0
    - tauri-plugin-shell 🦀: 2.0.2
    - @tauri-apps/plugin-shell : 2.0.1

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:8080/
    - framework: React
    - bundler: Webpack
```

### Stack trace

_No response_

### Additional context

This may be windows specific.

amrbashir:
closes #11462

### DIFF
--- a/crates/tauri/src/tray/mod.rs
+++ b/crates/tauri/src/tray/mod.rs
@@ -10,6 +10,7 @@ use crate::app::{GlobalMenuEventListener, GlobalTrayIconEventListener};
 use crate::menu::ContextMenu;
 use crate::menu::MenuEvent;
 use crate::resources::Resource;
+use crate::UnsafeSend;
 use crate::{
   image::Image, menu::run_item_main_thread, AppHandle, Manager, PhysicalPosition, Rect, Runtime,
 };
@@ -343,28 +344,23 @@ impl<R: Runtime> TrayIconBuilder<R> {
   pub fn build<M: Manager<R>>(self, manager: &M) -> crate::Result<TrayIcon<R>> {
     let id = self.id().clone();
 
-    struct SafeSend<T>(T);
-    unsafe impl<T> Send for SafeSend<T> {}
-
     // SAFETY:
     // the menu within this builder was created on main thread
     // and will be accessed on the main thread
-    let inner = &mut Some(self.inner) as *mut _ as isize;
+    let unsafe_builder = UnsafeSend(self.inner);
 
     let (tx, rx) = std::sync::mpsc::channel();
-    let inner = manager
+    let unsafe_tray = manager
       .app_handle()
       .run_on_main_thread(move || {
-        let inner = unsafe { &mut *(inner as *mut Option<tray_icon::TrayIconBuilder>) };
-
         // SAFETY: will only be accessed on main thread
-        let _ = tx.send(inner.take().unwrap().build().map(SafeSend));
+        let _ = tx.send(unsafe_builder.take().build().map(UnsafeSend));
       })
       .and_then(|_| rx.recv().map_err(|_| crate::Error::FailedToReceiveMessage))??;
 
     let icon = TrayIcon {
       id,
-      inner: inner.0,
+      inner: unsafe_tray.take(),
       app_handle: manager.app_handle().clone(),
     };
 
--- a/crates/tauri/src/tray/mod.rs
+++ b/crates/tauri/src/tray/mod.rs
@@ -346,6 +346,9 @@ impl<R: Runtime> TrayIconBuilder<R> {
     struct SafeSend<T>(T);
     unsafe impl<T> Send for SafeSend<T> {}
 
+    // SAFETY:
+    // the menu within this builder was created on main thread
+    // and will be accessed on the main thread
     let inner = &mut Some(self.inner) as *mut _ as isize;
 
     let (tx, rx) = std::sync::mpsc::channel();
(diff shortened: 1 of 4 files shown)


crates/tauri/src/tray/mod.rs
```diff
@@ -342,10 +342,29 @@ impl<R: Runtime> TrayIconBuilder<R> {
...
+    let (tx, rx) = std::sync::mpsc::channel();
+    let inner = manager
+      .app_handle()
+      .run_on_main_thread(move || {
+        let inner = unsafe { &mut *(inner as *mut Option<tray_icon::TrayIconBuilder>) };
+
+        // SAFETY: will only be accessed on main thread
+        let _ = tx.send(inner.take().unwrap().build().map(SafeSend));
+      })
+      .and_then(|_| rx.recv().map_err(|_| crate::Error::FailedToReceiveMessage))??;
```
Legend-Master: The cast just feels wrong to me, maybe something like this?

```suggestion
    struct SafeSendTrayIconBuilder(tray_icon::TrayIconBuilder);
    unsafe impl Send for SafeSendTrayIconBuilder {}
    impl SafeSendTrayIconBuilder {
      fn build(self) -> Result<tray_icon::TrayIcon, tray_icon::Error> {
        self.0.build()
      }
    }

    // SAFETY:
    // the menu within this builder was created on main thread
    // and will be accessed on the main thread
    let inner = SafeSendTrayIconBuilder(self.inner);

    let (tx, rx) = std::sync::mpsc::channel();
    let inner = manager
      .app_handle()
      .run_on_main_thread(move || {
        // SAFETY: will only be accessed on main thread
        let _ = tx.send(inner.build().map(SafeSend));
      })
      .and_then(|_| rx.recv().map_err(|_| crate::Error::FailedToReceiveMessage))??;
```

amrbashir: nice one, I tried similar thing but it didn't work, looks like the trick is moving `.build()` into a method on the new type, otherwise the closure captures the `dyn ContextMenu` and then rust complains

amrbashir: Found that we don't even need to add a new `build` method, we can just access the inner value through a method and all works
```rs
struct UnsafeSend<T>(T);
unsafe impl<T> Send for UnsafeSend<T> {}

impl<T> UnsafeSend<T> {
  fn take(self) -> T {
    self.0
  }
}

```

Legend-Master (approved):

Issue #11478, opened 2024-10-24
[bug] rpm bundling is stuck on Ubuntu 22.04


izderadicka:
### Describe the bug

rpm bundling stucks

running `cargo tauri build`

Build is OK,  deb bundling takes like 10-15 secs,  but rpm bundling then runs forever (I stopped it after 10 minutes or so) -  in processes I see `cargo tauri build` running and taking 100% of one CPU/core.

### Reproduction

This happens for rqbit project on Ubuntu 22.04,  node and npm installed via nvm,  tauri dependencies installed as per docs
- https://github.com/ikatson/rqbit.git
- cd rqbit/desktop
- cargo tauri build

### Full `tauri info` output

```text
$ cargo tauri info

[✔] Environment
    - OS: Ubuntu 22.4.0 x86_64 (X64)
    ✔ webkit2gtk-4.1: 2.44.3
    ✔ rsvg2: 2.52.5
    ✔ rustc: 1.81.0 (eeb90cda1 2024-09-04)
    ✔ cargo: 1.81.0 (2dbb1af80 2024-08-20)
    ✔ rustup: 1.27.1 (54dd3d00f 2024-04-24)
    ✔ Rust toolchain: stable-x86_64-unknown-linux-gnu (environment override by RUSTUP_TOOLCHAIN)
    - node: 20.18.0
    - yarn: 1.22.11
    - npm: 10.8.2

[-] Packages
    - tauri 🦀: 2.0.0
    - tauri-build 🦀: 2.0.0
    - wry 🦀: 0.44.1
    - tao 🦀: 0.30.2
    - tauri-cli 🦀: 2.0.3
    - @tauri-apps/api : 2.0.1 (outdated, latest: 2.0.3)
    - @tauri-apps/cli : 2.0.3 (outdated, latest: 2.0.4)

[-] Plugins
    - tauri-plugin-shell 🦀: 2.0.0
    - @tauri-apps/plugin-shell : 2.0.0 (outdated, latest: 2.0.1)

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:1420/
    - framework: React
    - bundler: Vite
```

livwvil:
I feced with the same issue.
Ubuntu 22.04
Build also stucks inside docker container
```
FROM rust:bookworm

ENV NODE_VERSION 18
ENV PNPM_VERSION 9.5.0
ENV DEBIAN_FRONTEND noninteractive

[...]

PR #11584, opened 2024-11-05
feat(bundler): add option to configure RPM compression

amrbashir:
ref: https://github.com/tauri-apps/tauri/issues/11478


FabianLars (approved):

lucasfernog (approved):
do we need a separate compression level for debug apps? or is this ok for build times?

FabianLars:
considering the size differences it may indeed make sense to lower the level or even leave it uncompressed for debug builds.

amrbashir:
Let's just ship it as is and see if more tweaking needs to be done.

PR #11632, opened 2024-11-09
docs: document vanilla JS import alternative

### LINKED ISSUE #2937: [docs] Unexplained '@tauri-apps/api' does not resolve to a valid URL.
All of the docs just casually assume that `import … from '@tauri-apps/api'` works — but it doesn't.

I've used the recommended `cargo create-tauri-app` (v4.5.6). I just pressed enter to pick all the default options (Rust frontend, Vanilla JS). And I can't run any example code from the documentation. If I paste any example code into the default `main.js`, it doesn't run, because:

> TypeError: Module name, '@tauri-apps/api' does not resolve to a valid URL.

It's frustrating that the most basic default app template, one that your tool recommends by default and generates, seems to be completely incompatible with *all* of the documentation examples. 

And as far as I can tell, this gotcha is not documented at all. There is a documentation page for `@tauri-apps/api`:

https://tauri.app/reference/javascript/api/

but it says absolutely nothing about the import. I would expect it to at least document what is required to make this import syntax work.

The auto-generated `main.js` for Vanilla option is not helpful either. It uses an alternative import syntax, but there are no comments about the import. There's no explanation why it uses a different import style, and how does this relate to the different import syntax used in the documentation.

kornelski:
Fixes https://github.com/tauri-apps/tauri-docs/issues/2937


packages/api/src/index.ts
```diff
@@ -10,6 +10,16 @@
  * ```typescript
  * import { event, window, path } from '@tauri-apps/api'
  * ```
+ *
+ * ### Vanilla JS API
+ *
+ * The above import syntax is for TypeScript. If you're using vanilla JavaScript, use `window.__TAURI__` instead. It requires `app.withGlobalTauri` configuration option enabled.
```
amrbashir: I think it would be more accurate to say JavaScript with bundler, what do you think about
```suggestion
 * The above import syntax is for JavaScript/TypeScript with a bundler. If you're using vanilla JavaScript, you can use the global `window.__TAURI__` object instead. It requires `app.withGlobalTauri` configuration option enabled.
```

amrbashir (approved):

PR #11684, opened 2024-11-14
fix(windows): The restart api does not work in single-instance mode

muwoo:
fix: https://github.com/tauri-apps/plugins-workspace/issues/1692.
Single instance mode should not cause restarted applications to exit.

We can add an additional `restart_from_tauri_api` parameter to `tauri::api::process::restart` to tell single-instance that this startup should not be blocked.

link single-instance plugin changes:[ #2050](https://github.com/tauri-apps/plugins-workspace/pull/2050)


lucasfernog:
fixed by #12313

PR #11716, opened 2024-11-17
refactor(bundler): remove unused fs utils, add http utils

FabianLars:
~~not done yet, just checking if it still compiles in ci~~


FabianLars:
the reason for this pr was that i needed to re-use the windows download functions (including the github mirror handling) in a small linux refactor (getting rid of build-appimage.sh) and while looking into it i noticed that we had 2 fs helper modules with duplicate functionality.

amrbashir (approved):

PR #11741, opened 2024-11-19
fix(api): transform icon when creating icon menu item and predefined about menu item with icon

amrbashir:


FabianLars (approved):
~~idk what i'm looking at 😂~~

PR #11758, opened 2024-11-20
refactor(bundler): Move appimage bundler script logic into rust

FabianLars:


FabianLars:
probably needs a bit more testing so not really ready to merge, but ready for review for sure :)

crates/tauri-bundler/src/bundle/linux/appimage.rs
```diff
@@ -0,0 +1,267 @@
...
+    let data = download("https://raw.githubusercontent.com/tauri-apps/linuxdeploy-plugin-gstreamer/master/linuxdeploy-plugin-gstreamer.sh")?;
+    save(&gstreamer, data)?;
+  }
+
+  // TODO: dd if=/dev/zero bs=1 count=3 seek=8 conv=notrunc of="{{tauri_tools_path}}/linuxdeploy-${linuxdeploy_arch}.AppImage"
+
+  Ok(linuxdeploy)
+}
+
+fn save(path: &Path, data: Vec<u8>) -> std::io::Result<()> {
```
amrbashir:
```suggestion
fn write_and_make_executable(path: &Path, data: Vec<u8>) -> std::io::Result<()> {
```

crates/tauri-bundler/src/bundle/linux/appimage.rs
```diff
@@ -0,0 +1,267 @@
...
+    save(&gtk, data)?;
+  }
+
+  let gstreamer = tools_path.join("linuxdeploy-plugin-gstreamer.sh");
+  if !gstreamer.exists() {
+    let data = download("https://raw.githubusercontent.com/tauri-apps/linuxdeploy-plugin-gstreamer/master/linuxdeploy-plugin-gstreamer.sh")?;
+    save(&gstreamer, data)?;
+  }
+
+  // TODO: dd if=/dev/zero bs=1 count=3 seek=8 conv=notrunc of="{{tauri_tools_path}}/linuxdeploy-${linuxdeploy_arch}.AppImage"
```
amrbashir: is this resolved?

crates/tauri-bundler/src/bundle/linux/appimage.rs
```diff
@@ -0,0 +1,267 @@
...
+    main_binary.set_name(main_binary_name_kebab);
+  }
+
+  // generate deb_folder structure
+  let (data_dir, icons) = debian::generate_data(&settings, &package_dir)
+    .with_context(|| "Failed to build data folders and files")?;
+  fs_utils::copy_custom_files(&settings.appimage().files, &data_dir)
+    .with_context(|| "Failed to copy custom files")?;
+
+  fs::create_dir_all(output_path.clone())?;
```
amrbashir:
```suggestion
  fs::create_dir_all(&output_path)?;
```

crates/tauri-bundler/src/bundle/linux/appimage.rs
```diff
@@ -0,0 +1,267 @@
...
+
+  // Using create_dir_all for a single dir so we don't get errors if the path already exists
+  fs::create_dir_all(&app_dir_usr_bin)?;
+  fs::create_dir_all(&app_dir_usr_lib)?;
+
+  // Copy bins and libs that linuxdeploy doesn't know about
+  if settings.deep_link_protocols().is_some()
+    && fs::copy("/usr/bin/xdg-mime", app_dir_usr_bin.join("xdg-mime")).is_err()
+  {
+    log::error!("xdg-mime binary not found!");
```
amrbashir: shouldn't we exist here with error?

crates/tauri-bundler/src/bundle/linux/appimage.rs
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  // xdg-open is only 50kb (in a 80mb+ appimage) and quite commonly used so we just always add it
+  // we do however check if the user may have provided their own copy already
+  if !app_dir_usr_bin.join("xdg-open").exists()
+    && fs::copy("/usr/bin/xdg-open", app_dir_usr_bin.join("xdg-open")).is_err()
+  {
+    log::error!("xdg-open binary not found!");
```
amrbashir: same

crates/tauri-bundler/src/bundle/linux/appimage.rs
```diff
@@ -0,0 +1,267 @@
...
+    "/usr/libexec",
+  ];
+
+  for file in [
+    "WebKitNetworkProcess",
+    "WebKitWebProcess",
+    "libwebkit2gtkinjectedbundle.so",
+  ] {
+    for source in search_dirs.map(PathBuf::from) {
+      // TODO: Check if it's the same dir name on all systems
```
amrbashir, line 146: is this also resolved?

crates/tauri-bundler/src/bundle/linux/appimage.rs
FabianLars: yes, didn't find the todo comment anymore 😂 i'll move the code to the comment though, makes more sense

crates/tauri-bundler/src/bundle/linux/appimage.rs
FabianLars: yes and no, it's correct on ubuntu and fedora, and i think on arch too. For other systems i guess we'll have to wait and see if we get complaints.

crates/tauri-bundler/src/bundle/linux/appimage.rs
FabianLars: hmm, only if i put it behind a plugin check again, which is fine 🤷

crates/tauri-bundler/src/bundle/linux/appimage.rs
FabianLars: probablyyy, bundling xdg-open and xdg-mime in appimages proved to be not so perfect so not bundling it at all may be desired but i think we should solve that issue differently (in the plugin)

crates/tauri-bundler/src/bundle/linux/appimage.rs
amrbashir: sure, let's add a check again

crates/tauri-bundler/src/bundle/linux/appimage.rs
amrbashir: if we can't be sure, let's change the logic to use `find -L` so we don't break things

crates/tauri-bundler/src/bundle/linux/appimage.rs
amrbashir: let's exit here with error, to keep the logic the same since this is just a refactor PR and iterator on a fix in another PR

crates/tauri-bundler/src/bundle/linux/appimage.rs
FabianLars: we good https://github.com/WebKit/WebKit/blob/e5f600069d461e87967b7be349867e9864185a9d/Source/cmake/OptionsGTK.cmake#L245 (also the gtk plugin would probably crumble on those (non-existent) systems too)

crates/tauri-bundler/src/bundle/linux/appimage.rs
FabianLars: the opener plugin is bit of a problem in this context as it doesn't have a tauri.conf.json entry i can check here.

amrbashir (approved):

PR #11767, opened 2024-11-21
feat(core): Add `Scope::is_forbidden`

FabianLars:
ref https://github.com/tauri-apps/plugins-workspace/pull/2070

i didn't call is_forbidden in is_allowed to not do the OS operations twice.

### DIFF
--- a/crates/tauri/src/scope/fs.rs
+++ b/crates/tauri/src/scope/fs.rs
@@ -340,12 +340,11 @@ impl Scope {
 
   /// Determines if the given path is allowed on this scope.
   ///
-  /// This calls `is_forbidden` internally and therefore returns `false`
-  /// if the path was explicitly forbidden or neither allowed nor forbidden.
+  /// Returns `false` if the path was explicitly forbidden or neither allowed nor forbidden.
   ///
-  /// May return `false` if the path to check points to a broken symlink.
+  /// May return `false` if the path points to a broken symlink.
   pub fn is_allowed<P: AsRef<Path>>(&self, path: P) -> bool {
-    let path = verify_path(path);
+    let path = try_resolve_symlink_and_canonicalize(path);
 
     if let Ok(path) = path {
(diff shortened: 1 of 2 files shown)


crates/tauri/src/scope/fs.rs
```diff
@@ -380,6 +372,39 @@ impl Scope {
...
+        .unwrap()
+        .iter()
+        .any(|p| p.matches_path_with(&path, self.match_options))
+    } else {
+      false
+    }
+  }
+}
+
+fn verify_path<P: AsRef<Path>>(path: P) -> crate::Result<PathBuf> {
```
amrbashir: This sounds more like `try_resolve_symlink_and_canonicalize`

crates/tauri/src/scope/fs.rs
```diff
@@ -380,6 +372,39 @@ impl Scope {
...
+    if let Ok(path) = path {
+      let path: PathBuf = path.components().collect();
+      self
+        .forbidden_patterns
+        .lock()
+        .unwrap()
+        .iter()
+        .any(|p| p.matches_path_with(&path, self.match_options))
+    } else {
+      false
```
amrbashir: this should be `true`, broken symlinks should be forbidden

crates/tauri/src/scope/fs.rs
```diff
@@ -339,21 +339,13 @@ impl Scope {
   }
 
   /// Determines if the given path is allowed on this scope.
+  ///
+  /// This calls `is_forbidden` internally and therefore returns `false`
+  /// if the path was explicitly forbidden or neither allowed nor forbidden.
```
amrbashir:
```suggestion
  /// Returns `false` if the path was explicitly forbidden or neither allowed nor forbidden.
```

amrbashir (changes requested):

amrbashir (approved):

PR #11729, opened 2024-11-18
fix(windows): TrayIcon allow show_menu_on_left_click

sandercox:
The `show_menu_on_left_click` was implemented for Windows in tauri-apps/tray-icon#199 but the tauri TrayBuilder didn't allow the usage yet when platform was not "macOS". Added the `any` `target_os` guard. And updated the comments inline with the changes made in the PR mentioned.

PR #11726, opened 2024-11-18
feat(api/tray): add `TrayIcon.setShowMenuOnLeftClick` method

amrbashir:
Followup to #11729:

- Deprecate `setMenuOnLeftClick` and Add `setShowMenuOnLeftClick` to match Rust API.
- Make docs consistent
- Format some `run_item_main_thread` macro usages

PR #11797, opened 2024-11-26
fix: platform check on set_show_menu_on_left_click

sandercox:
The platform check was broken after a merge I'm afraid. It needs to spell out `target_os` in the check.

This is regression caused by changes in #11726 compared to the changes merged from #11729.


FabianLars:
`windows` is a shortcut for `target_family = "windows"` and _should_ work. (there's another `unix` shortcut but that includes macos and linux so we don't use it often)

sandercox:
Right... I would swear I had issues that went away with changing this but changing it back now it's still all ok. Sorry for my lack of knowledge here.

PR #11835, opened 2024-11-30
fix(path): wrong command for isAbsolute plugin

ilittlebig:
The current invoke argument for `isAbsolute` is wrong, and should be `plugin:path|is_absolute`.
Using `isAbsolute` from "@tauri-apps/api/path" return this error message otherwise:
```
path.isAbsolute not allowed. Command not found
```


FabianLars (approved):
oh damn, thanks!

PR #11852, opened 2024-12-02
fix(bundler): create tauri tools dir first. make tools executable.

FabianLars:


amrbashir (approved):

FabianLars:
stuff like this makes me go mental. i _explicitly_ tested this and it worked, so what did i actually test 😭

lucasfernog (approved):

PR #11971, opened 2024-12-14
feat: add support for injecting script into subframes (from wry)

Geometrically:
This PR ports https://github.com/tauri-apps/wry/pull/1365 to Tauri/tauri-runtime-wry

### DIFF
--- a/crates/tauri/src/webview/mod.rs
+++ b/crates/tauri/src/webview/mod.rs
@@ -671,7 +671,17 @@ fn main() {
     self
       .webview_attributes
       .initialization_scripts
-      .push(script.to_string());
+      .push((script.to_string(), true));
+    self
+  }
+
+  /// Adds an initialization script with the option to inject into sub-frames
+  #[must_use]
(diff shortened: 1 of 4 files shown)


Norbiros (approved):

crates/tauri/src/webview/mod.rs
```diff
@@ -671,7 +671,17 @@ fn main() {
...
       .webview_attributes
       .initialization_scripts
-      .push(script.to_string());
+      .push((script.to_string(), true));
+    self
+  }
+
+  /// Adds an initialization script with the option to inject into sub-frames
+  #[must_use]
+  pub fn initialization_script_for_main_only(mut self, script: &str, main_only: bool) -> Self {
```
amrbashir, line 680: This same method needs to be added on the `WebviewWindow` type as well.

amrbashir (changes requested):
apologies for the late review, could you also add a change file in .changes directory?

lucasfernog:
oh looks like we already merged this on a separate PR: #/13076
thank you for your contribution!

PR #12089, opened 2025-01-01
chore(deps): Update `jsonschema` to `0.28`

Stranger6667:
Hey!

This PR updates the `jsonschema` crate and applies its newer APIs.

### DIFF
--- a/crates/tauri-cli/src/helpers/config.rs
+++ b/crates/tauri-cli/src/helpers/config.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-License-Identifier: MIT
 
+use itertools::Itertools;
 use json_patch::merge;
 use serde_json::Value as JsonValue;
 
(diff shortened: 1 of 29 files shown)


crates/tauri-cli/src/helpers/config.rs
```diff
@@ -172,11 +173,11 @@ fn get_internal(
     || config_path.extension() == Some(OsStr::new("json5"))
   {
     let schema: JsonValue = serde_json::from_str(include_str!("../../config.schema.json"))?;
-    let schema = jsonschema::JSONSchema::compile(&schema).unwrap();
-    let result = schema.validate(&config);
-    if let Err(errors) = result {
+    let validator = jsonschema::validator_for(&schema).expect("Invalid schema");
+    let mut errors = validator.iter_errors(&config).peekable();
+    if errors.peek().is_some() {
```
amrbashir, line 178: do we really need to peek? I think we can just loop through the iterator normally and if no errors found, it will skip the iteration and nothing will happen

amrbashir (dismissed):

crates/tauri-cli/src/helpers/config.rs
Stranger6667: Good point!

Stranger6667: The only thing needed is to skip `exit` in the `if !reload` block which happens only after all errors are logged if there is at least one error. Right now peek is used exactly for this, but a boolean would work too

Stranger6667: Sorry for the force push, I updated this block.

amrbashir: > The only thing needed is to skip `exit` in the `if !reload` block which happens only after all errors are logged if there is at least one error. Right now peek is used exactly for this, but a boolean would work too

Oh didn't notice that one, apologies but you had it correct, can you revert it?

amrbashir: you also need to fix Cargo.lock conflicts with `dev` branch

Stranger6667: Sure! reverted + updated `Cargo.lock`

FabianLars (approved):

PR #12123, opened 2025-01-02
ci: Switch to upstream create-pull-request

FabianLars:
Tired of the warnings and i'd rather not update a 5 year old fork (even though it would be clean). I also don't see a reason for the fork to exist in the first place as locking upstream to a specific commit instead of using tags will provide the same safety for us.


tweidinger (approved):

tweidinger:
We should probably also update to v3 (or the pinned commit) here https://github.com/tauri-apps/tauri/blob/a52a34de83ccc6d0b3243a9bc3c07d0c35917c2e/.github/workflows/covector-version-or-publish.yml#L127

PR #12136, opened 2025-01-02
fix(bundler): change build metadata verification into a warning

### LINKED ISSUE #8038: [bug] NSIS bundler: semver version metadata is required to be numeric, but shouldn't be
### Describe the bug

When bundling an app with the tauri version e.g. "2.5.0-beta+1a2b3c4d", I get an error that "optional build metadata in app version must be numeric-only".

My understanding is that this limitation only should apply to the Wix bundler (#5286), but I am using NSIS, which uses windows' ProductVersion instead of FileVersion.

Since ProductVersion can accept arbitrary strings, there appears to be no technical reason that this limitation should exist for NSIS bundlers.

### Reproduction

1. Create a tauri project with `project > version` set to e.g. `0.1.0-beta+123` and `tauri > bundle > targets` set to `"nsis"`.
2. `tauri build`
3. Observe the correct bundle name, and ProductVersion in the file's properties:
```
    Finished 1 bundle at:
        .\src-tauri\target\release\bundle\nsis\issue_0.1.0-beta+123_x64-setup.exe
```
image

4. Change `project > version` to `0.1.0-beta+abc`
5. `tauri build`
6. Observe the error:
```
       Error failed to bundle project: optional build metadata in app version must be numeric-only
```

### Expected behavior

Step 6 should function exactly the same as step 3.

### Platform and versions

```text
[✔] Environment
    - OS: Windows 10.0.22621 X64
    ✔ WebView2: 118.0.2088.46
    ✔ rustc: 1.66.0 (69f9c33d7 2022-12-12)
    ✔ Cargo: 1.66.0 (d65d197ad 2022-11-15)
    - node: 16.20.2
    - yarn: 1.22.19
    - npm: 8.19.4

[-] Packages
    - tauri [RUST]: 1.5.2
    - tauri-build [RUST]: 1.5.0
    - wry [RUST]: 0.24.4
    - tao [RUST]: 0.16.4
    - tauri-cli [RUST]: not installed!
    - @tauri-apps/cli [NPM]: 1.5.2 (outdated, latest: 1.5.4)

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../dist
    - devPath: localhost
```

### Stack trace

_No response_

### Additional context

_No response_

unknovvn:
Closes #8038


FabianLars:
your first change file was correct. the bot sends the link to create a new one unconditionally at the moment :/

unknovvn:
@FabianLars haha, it got me. Gonna remove the bot generated file then. Thanks 🙇

FabianLars:
Okay so i just checked again and this won't work. `VIProductVersion` and `VIFileVersion` must be numeric only in a X.X.X.X format. This is a requirement by NSIS which if i got that right is because those values are used to set https://learn.microsoft.com/de-de/windows/win32/api/verrsrc/ns-verrsrc-vs_fixedfileinfo?redirectedfrom=MSDN which also requires numeric-only values.

I'm not sure what `VS_FixedFileInfo` is actually used for though because for example the versions in the file properties' Details tab are set via `VIAddVersionKey`.

We could modify the check to replace the build/pre data with `.0` if it's non-numeric, keeping the "normal" version for `VIAddVersionKey` and to then wait and see if anything breaks.

unknovvn:
I have additionally researched this topic a bit more and it looks like @FabianLars is correct - NSIS VIProductVersion requires all 4 parts to be numbers.
Personally I think that replacing build version with .0 if it's non-numeric is not correct as it's better to throw a build error and indicate that this format is not supported rather than introducing unexpected behavior. 

One thing I noticed, that msi convert_version method has additional validation regarding correctness of each version part. Which might be added to nsis in order to unify this behavior.

FabianLars:
> Personally I think that replacing build version with .0 if it's non-numeric is not correct as it's better to throw a build error and indicate that this format is not supported rather than introducing unexpected behavior.

Yeah, fair. I'm just thinking about how to expose VIAddVersionKey to devs, which supports non-numeric versions, without adding yet another version config.

unknovvn:
If we would decide to expose build version with non-numeric values as a VIAddVersionKey to devs - it would require interpreting VIProductVersion in some way. Either to add .0 at the end (which would replicate default NSIS behavior) or to add stripped out version of non-numeric build version (remove non-numeric values from the build version). In both cases this would introduce unexpected behavior

unknovvn:
Although after checking initial [Bug](https://github.com/tauri-apps/tauri/issues/8038) it looks like author of that ticket was expecting that file version would have default .0 revision value instead of non-numeric build version value (probably expecting same behavior for VIProductVersion). In that case it looks like for non-numeric build numbers we could set VIProductVersion and FileVersion with values where non-numeric build version would be replaced by .0 and for product version it would return original build version

unknovvn:
@FabianLars please verify if my assumptions are correct and this PR can be merged. 🙇

crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
```diff
@@ -77,12 +77,12 @@ OutFile "${OUTFILE}"
...
 InstallDir "${PLACEHOLDER_INSTALL_DIR}"
 
-VIProductVersion "${VERSIONWITHBUILD}"
+VIProductVersion "${VERSION}"
 VIAddVersionKey "ProductName" "${PRODUCTNAME}"
 VIAddVersionKey "FileDescription" "${PRODUCTNAME}"
 VIAddVersionKey "LegalCopyright" "${COPYRIGHT}"
 VIAddVersionKey "FileVersion" "${VERSION}"
-VIAddVersionKey "ProductVersion" "${VERSION}"
+VIAddVersionKey "ProductVersion" "${VERSIONWITHBUILD}"
```
FabianLars: This needs to be reverted.
1) FileVersion and ProductVersion should show the same value and both should support semver.
    - Talking about the VIAddVersionKey variant, _not_ `VIProductVersion`/`VIFileVersion` !
2) We use `VERSION` a few more times in this file where it expects the full semver.

VIProductVersion should be the only thing that gets a modified version.

So i think the approach in general (for the other files as well) would be:
1) Set VERSION with the user specified version without modifications
2) Try to use the user specified version for VERSIONWITHBUILD, print a warning if the prerelease/build data is non-numeric and replace it with `.0`
Example:
- User sets `1.0.0+123`
  - VERSION == `1.0.0+123`
  - VERSIONWITHBUILD == `1.0.0.123`
- User sets `1.0.0+build123`
  - VERSION == `1.0.0+build123`
  - VERSIONWITHBUILD == `1.0.0.0` **but with a warning log**

One thing that i'm unsure about is the prerelease data so `1.0.0-beta` instead of `1.0.0+beta`. This is currently ignored in the versionwithbuild logic. I guess that's fine though and good in context of my other comment because it means that updating should already work fine.

FabianLars (commented):
> In both cases this would introduce unexpected behavior

The question would be where? Where is VIProductVersion used that this would cause issues. The only thing i can think of that we have to check is updating == installing 1.0.0-builld.20 when 1.0.0-beta.1 is already installed on the system.

Tauri's update can handle that fine but the installers, or at least the wix installer have built-in checks for this as well, not sure how our nsis installer behaves.

crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
unknovvn: I have reverted changes, added warning log. Now it should work as you have described it

crates/tauri-bundler/src/bundle/windows/nsis/mod.rs
```diff
@@ -140,16 +140,20 @@ fn add_build_number_if_needed(version_str: &str) -> anyhow::Result<String> {
         "{}.{}.{}.{}",
         version.major, version.minor, version.patch, version.build
       ));
-    } else {
-      anyhow::bail!("optional build metadata in app version must be numeric-only");
     }
   }
 
+  log::warn!(
+    "Unable to parse version build metadata. Numeric value expected, received: {}",
```
FabianLars:
```suggestion
    "Unable to parse version build metadata. Numeric value expected, received: `{}`. This will be replaced with `0` in `VIProductVersion` because Windows requires this field to be numeric.",
```

FabianLars (dismissed):

FabianLars (dismissed):

FabianLars (approved):

Issue #12032, opened 2024-12-23
Window and webview commands are all async functions


Legend-Master:
Most of if not all of window and webview commands are sending the command to main thread and handled there, adding an async layer on top of it feels wrong to me, also this increases the binary size by 300-400 KB (tested with release build of helloworld example on Windows 3.60MB -> 3.96MB)

https://github.com/tauri-apps/tauri/blob/fdaf48fc4a706b968e92b2b44e3830c51a5be1fe/crates/tauri/src/window/plugin.rs#L49-L79

I wonder why we did this, since pretty much everything else uses normal functions

I'm aware of the dead lock I mentioned in https://github.com/tauri-apps/tauri/pull/10210#discussion_r1768238241, if that's the reason we did this, I feel like probably should fix that instead

Edit: did a bit more digging, seems like the dead lock is caused by us locking the `plugins` for invoking commands, and during the invoke, the command sends an event to the event loop synchronously and the event handler also needs to lock `plugins` for the event listeners, not sure why `on_event` and `extend_api` take `mut self` though (probably for the callbacks to be able to be `FnMut`?)

PR #12162, opened 2025-01-03
enhance(core): use sync commands for window and webview getters

Legend-Master:
Reference: #12032

I don't quite know the reason behined why all window and webview commands are async, seems like for the setters, they would trigger events during the call and causing dead locks, but I can't see any reason for the getters, let me know if this was actually intentional or not


amrbashir:
I believe we used to have them sync but it caused dead locks somewhere, I think @lucasfernog has more info about this

lucasfernog:
I'm pretty sure we'll have deadlocks if we do this :/

Legend-Master:
> I'm pretty sure we'll have deadlocks if we do this :/

Do you know any particular places where the deadlock will happen?

If we can't fix them, maybe it's a good idea to document the problem, I have encountered this before trying to call `window.center` in a sync command and it's quite frustrating trying to debug it

Issue #5250, opened 2022-09-21
[feat] Allow disable background throttling


mantou132:
### Describe the problem

When the APP minimizes, the webview timer will be throttling.

### Describe the solution you'd like

Electorn has `webPreferences.backgroundThrottling` options

https://www.electronjs.org/docs/latest/api/browser-window#new-browserwindowoptions

### Additional context

Related: https://github.com/electron/electron/issues/1786

FabianLars:
Sooo, i researched this topic 2 times already. Once for #5147 and a few weeks before that in context of a discord thread.
And it really doesn't look good. The webviews don't have APIs for this at all and of course we don't have as much control over the engines as electron does -> their solution is irrelevant for us :(
Here's the most relevant upstream issue i could find: <https://github.com/MicrosoftEdge/WebView2Feedback/issues/1172>

The only thing left that i couldn't figure out (though it didn't look good either) is to try to make it think it's visible and focused by some weird window manipulation, but that's beyond me - cc @amrbashir 

tldr: Don't wait for this feature :/

amrbashir:
This only affects Windows or other platforms as well?

> try to make it think it's visible and focused by some weird window manipulation

On Windows, we don't hide the webview when the window is minimized (although we probably should) and yet the timers are being throttled. So I don't think we can do much here without explicit support in the webviews.

FabianLars:
This question was always asked in the context of Windows only. No idea how the other webiews actually behave tho

chen-gaoyuan:
> Sooo, i researched this topic 2 times already. Once for #5147 and a few weeks before that in context of a discord thread. And it really doesn't look good. The webviews don't have APIs for this at all and of course we don't have as much control over the engines as electron does -> their solution is irrelevant for us :( Here's the most relevant upstream issue i could find: [MicrosoftEdge/WebView2Feedback#1172](https://github.com/MicrosoftEdge/WebView2Feedback/issues/1172)
> 
> The only thing left that i couldn't figure out (though it didn't look good either) is to try to make it think it's visible and focused by some weird window manipulation, but that's beyond me - cc @amrbashir
> 
> tldr: Don't wait for this feature :/
查了几天资料, 原因就是浏览器将不在前台的页面休眠了, Chromium最长会一分钟才唤醒一次, 导致setInterval不管设置多小的值也要等1分钟执行一次.
现在有几种方案:
[...]

PR #12181, opened 2025-01-03
feat: disable background throttling

bastiankistner:
This PR relates to https://github.com/tauri-apps/tauri/issues/5250

It requires a change of `wry` for which I've opened another PR: https://github.com/tauri-apps/wry/pull/1445

I've added an option `disableBackgroundThrottling` to allow developers to choose if their view should automatically be throttled / suspended by the browser after a certain time of being hidden or minimised. In the worst case, this default behaviour of browsers can lead to broken applications running in the background. 

For instance if a websocket connection is the only reason not to suspend a "tab" and the machine goes into sleep mode and wakes up again, the connection is often lost and the browser won't have any reason to keep the background tab active and therefore unload it, which will not allow the hidden view to "heal" itself. 

This default behaviour effectively renders any browser capabilities unusable for long running background processes. One alternative might be moving logic from the frontend to the backend, but browsers have meanwhile become a great choice for local first apps and if the browser already offers the required features easily accessible, why force developers to leave a territory they feel comfortable with?

As of now, this is only available for WebKit based browsers on iOS 17.0+,iPadOS 17.0+,Mac Catalyst 17.0+, macOS 14.0+, visionOS 1.0+ ( see https://developer.apple.com/documentation/webkit/wkpreferences/inactiveschedulingpolicy-swift.enum?language=objc ).

It can be assumed that it will also be implemented for WebView2 (Windows). And it might be that it's already usable for linux and android, although I couldn't find any documentation about it and also cannot test it locally due to a lack of devices. 

As an alternative for windows and non-webkit browsers, one can setup a pending WebLock transaction, which, according to the docs, can also keep the browser from suspending it.

----

I needed to update a snapshot test and am unsure if this is at all related to this change. Please let me know if I did something wrong here.

----
### Pending Tasks

- [x] upgrade `wry` package version when https://github.com/tauri-apps/wry/pull/1445 is merged, which will have `with_disable_background_throttling` on `WebViewBuilder`


.changes/disable-background-throttling.md
```diff
@@ -0,0 +1,11 @@
+---
+'tauri-cli': 'minor'
+'tauri-runtime': 'minor'
+'tauri-runtime-wry': 'minor'
+'tauri-schema-generator': 'minor'
+'tauri-utils': 'minor'
+'tauri': 'minor'
+'@tauri-apps/api': 'minor'
```
FabianLars:
```suggestion
'tauri-cli': 'minor:enhance'
'tauri-runtime': 'minor:enhance'
'tauri-runtime-wry': 'minor:enhance'
'tauri-utils': 'minor:enhance'
'tauri': 'minor:enhance'
'@tauri-apps/api': 'minor:enhance'
'@tauri-apps/cli': 'minor:enhance'
```
schema-generator doesn't get updated/released.

crates/tauri-runtime/src/webview.rs
```diff
@@ -444,6 +448,23 @@ impl WebviewAttributes {
...
+
+  /// Set whether background throttling should be disabled.
+  ///
+  /// By default, browsers throttle timers and even unload the whole tab (view) to free resources after roughly 5 minutes when
+  /// a view became minimized or hidden. This will permanently suspend all tasks until the documents visibility state
+  /// changes back from hidden to visible by bringing the view back to the foreground.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux / Windows / Android**: Unsupported yet. But workarounds like a pending WebLock transaction might suffice.
```
FabianLars:
```suggestion
  /// - **Linux / Windows / Android**: Unsupported. Workarounds like a pending WebLock transaction might suffice.
```

I think there should also be a line for ios / macos mentioning the minimum OS version since they are quite recent.

Needs to be applied to the docs in the other files as well of course.

FabianLars (changes requested):
thanks so much. If nothing major changes in the Wry PR i only have 2 very small comments:

crates/tauri-runtime/src/webview.rs
bastiankistner: I'll update it tomorrow! Thanks a lot!

Btw: I'm still running my app on this PR and I haven't had a single issue since! I'm certain this will benefit a lot of devs.

Thank you for taking the time to review it @FabianLars

bastiankistner:
@FabianLars I changed the comments according to your suggestions and also updated the wry PR with a version check. Let me know if there is anything else I should do before this PR can be merged.

bastiankistner:
I changed the wry version to `0.48.1`

bastiankistner:
@FabianLars is there something still missing in this PR or can we merge this with the release of wry 0.48.1 ?

FabianLars:
sorry, i was waiting with minor PRs intentionally to not block patch releases until the webview2-com and windows updates, but that ended up with me completely ignoring those prs.

Just skipping over the code one major issue i see is that tauri-utils cannot depend on wry, so that has to be reverted. That may also fix some of the build issues.

Api wise i don't think it makes sense to have Option<Enum> in the WebviewBuilder and WebviewWindowBuilder setter methods. We can use the enum directly here.

Lastly 2 small doc things: 1) the enum and maybe the methods should document clearly what the default variant is. 2) The `since` tags should be 2.3.0 - currently they are 2.2.0 (and one is 2.0.0)

bastiankistner:
@FabianLars thank you! Implemented the requested changes.

I have one question left: Since we have the `invoke('plugin:window|set_background|throttling')` here https://github.com/bastiankistner/tauri/blob/b82b6efc96c9d62b75f1086ae0435e1630541e03/packages/api/src/window.ts#L1553, don't we also need the corresponding methods in `tauri-runtime/src/lib.rs` in `WindowDispatch` and in `tauri/src/webview/mod.rs` for `impl Webview` and probably in `crates/tauri/src/webview/plugins.rs` ?

I'm uncertain if/how this works. I doubt that I can change the value once the webview is loaded. But when I open a new window through JS, it should be possible to specify it.

FabianLars:
ah didn't even notice that. since wry has no runtime method for this, only one to set it on the builder you can remove that ts function again. Only the WebviewOptions property will work here.

therefore this is a bit offtopic but the setter function would belong in webview.ts not window.ts because well, it's a webview method and not related to the window / windowing library.

bastiankistner:
@FabianLars removed it. Also had to update snapshot tests. I don't really know why. Would be great if you could let me know if I should undo those. Maybe it's due to my environment?

FabianLars:
never touched nor looked at the snapshot stuff so no idea 😂

bastiankistner:
@FabianLars is there anything missing or can we merge it soon?

FabianLars:
i don't see anything missing, no. Just wanted to release a patch release before merging the minor prs. I currently have monday in mind to start merging those because i'm busy over the weekend.

apologies for the long wait times.

bastiankistner:
@FabianLars just wanted to kindly follow up to check if this PR might be able to get merged this week? As mentioned before, if there is anything coming up, please let me know and I'll do my best to fix it.

FabianLars:
yep, still up for this week.

FabianLars (approved):
the day has finally come 😂

bastiankistner:
> the day has finally come 😂

No worries, I was already preparing to remind you on Monday 😂

thanks a lot! Now my builds will finally pass again 🙏

ScarAnd:
Hey @bastiankistner 
I experience the same problem as you when using a websocket connection in tauri v1. Is this feature available in tauri v1 or do I have to switch to v2 for that? It would also be really cool to see how you implemented that feature into your project if possible.

in the comment section of the `setBackgroundThrottling` function I see, that its available since the version 2.3.0 of tauri/api

PR #12313, opened 2025-01-08
fix: ensure RunEvent::Exit is triggered on restart

### LINKED ISSUE #1692: Unstable relaunch with single instance plugin
With `tauri_plugin_single_instance`, relaunch from `tauri_plugin_process` is not guaranteed. Sometimes the application just closes without restarting.

### LINKED ISSUE #2256: [bug] Store cannot save with `app_handle.restart()`
### Describe the bug

I have an App with Store and Process plugin activated. The store cannot be saved if I restart the App with `app_handle.restart()`, but it can be saved via `app_handle.exit(0)`.

### Reproduction

The code is from my project. I save the app settings with store, and then call the `app_handle.restart()`.

```
#[derive(Serialize, Deserialize)]
pub struct AppSettings {
    language: String,
    direct_generation: bool,
}

#[tauri::command]
pub fn get_app_settings(app_handle: AppHandle) -> AppSettings {
    let store = app_handle.store("store.json").unwrap();

    let language = store
        .get("language")
        .expect("No language setting is found.");
    let direct_generation = store
        .get("direct_generation")
        .expect("No direct generation setting is found.");

    AppSettings {
        language: language.get("value").unwrap().as_str().unwrap().to_string(),
        direct_generation: direct_generation.get("value").unwrap().as_bool().unwrap(),
    }
}

#[tauri::command]
pub fn set_app_settings(app_handle: AppHandle, language: &str, direct_generation: bool) {
    let store = app_handle.store("store.json").unwrap();

    store.set("language", json!({"value": language}));
    store.set("direct_generation", json!({"value": direct_generation}));

    // This won't save the store, after restart it still shows the old setting values
    app_handle.restart();

    // This works, it will save the store and after I restart the app manually it shows the new setting values
    // app_handle.exit(0);
}
```

I have those plugins activated.

```
pub fn run() {
    tauri::Builder::default()
        .plugin(tauri_plugin_dialog::init())
        .plugin(tauri_plugin_fs::init())
        .plugin(tauri_plugin_process::init())
        .plugin(tauri_plugin_shell::init())
        .plugin(tauri_plugin_store::Builder::default().build())
        .setup(|app| {
            StoreBuilder::new(app, "store.json")
                .default("language", json!({"value": "en"}))
                .default("direct_generation", json!({"value": false}))
                .build()?;

            Ok(())
        })
        .invoke_handler(tauri::generate_handler![
            get_app_settings,
            set_app_settings,
        ])
}
```

### Expected behavior

Store should be saved during restart, just like it exits.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Manjaro 24.2.1 x86_64 (X64)
    ✔ webkit2gtk-4.1: 2.46.5
    ✔ rsvg2: 2.59.2
    ✔ rustup: 1.27.1 (2024-05-07)
    - node: 23.4.0
    - pnpm: 9.12.3
    - yarn: 1.22.22
    - npm: 11.0.0

[-] Packages
    - tauri 🦀: 2.2.0
    - tauri-build 🦀: 2.0.4
    - wry 🦀: 0.48.0
    - tao 🦀: 0.31.0
    - tauri-cli 🦀: 2.0.4

[-] Plugins
    - tauri-plugin-fs 🦀: 2.2.0
    - tauri-plugin-store 🦀: 2.2.0
    - tauri-plugin-process 🦀: 2.2.0
    - tauri-plugin-shell 🦀: 2.2.0

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
```

### LINKED ISSUE #11392: [bug] App::restart does not restart after update.download_and_install in rust
### Describe the bug

When a new version on macOS has been downloaded and installed the app closes down but does not start again.

It is a desktop app for macOS and Windows. It has a main window that the user can close/hide while the icon tray remains. So on WindowEvent::CloseRequested we prevent closing the app. Could this cause the issue?

We just migrated to Tauri 2 so had to provide our own dialog (see updater.rs below) and doing version check and triggering download, install and restart (which fails on macOS).

The app is built with our self hosted GitLab CI/CD pipeline producing a static updater.json uploaded to S3 together with the artefacts:

```json
{
    "version": "#VERSION",
    "notes": "",
    "pub_date": "#PUB_TIME",
    "platforms": {
        "universal-apple-darwin": {
            "signature": "#UNIVERSAL_APPLE_DARWIN_SIGNATURE",
            "url": "https://autoremind-app.xxx/release/autoremind-app.app.tar.gz"
        },
    }
}
```

Important part of my main.rs:
```rs
use tauri::{ Builder, Manager, Window, WindowEvent, WebviewWindowBuilder, WebviewUrl};

#[path = "updater.rs"]
mod updater;

fn main() {
    let context = tauri::generate_context!();
    let url = format!("http://localhost:{}", BUILD_PORT).parse().unwrap();
    let window_url = WebviewUrl::External(url);

    //let app2 = tauri::Builder::default()
    Builder::default()
        .plugin(tauri_plugin_single_instance::init(|app, _args, _cwd| {
            println!("Trying to open another instance");
            let window = app.get_webview_window("main")
                .expect("no main window");
            window.show().map_err(|err| println!("{:?}", err)).ok();
            let _ = window.set_focus();
        }))
        .plugin(tauri_plugin_localhost::Builder::new(BUILD_PORT).build())
        .plugin(tauri_plugin_updater::Builder::new().build())
        .plugin(tauri_plugin_dialog::init())
        .plugin(tauri_plugin_process::init())
        .setup(|app| {
            WebviewWindowBuilder::new(
                app,
                "main".to_string(),
                if cfg!(dev) {
                    Default::default()
                } else {
                    window_url
                },
            )
                .title(MAIN_WINDOW_TITLE)
                .position(30.0, 30.0)
                .inner_size(800.0, 650.0)
                .initialization_script(&pass_app_environment_to_window())
                .build()?;

            menubar::build_tray_and_menus(app);

            badge::set_badge(app.handle(), 5);

            updater::init(&app);
            Ok(())
        })
        .on_window_event(handle_window_event)
        //.enable_macos_default_menu(false)
        .invoke_handler(tauri::generate_handler![
            open_two_way,
            close_two_way,
            set_badge
        ])
        //.build(context)  // in this case app.run() below is needed
        .run(context)
}

fn handle_window_event(window: &Window, event: &WindowEvent) {
    match event {
        WindowEvent::CloseRequested { api, .. } => {
            api.prevent_close(); // We want the icon tray to remain after closing main window
            window.hide().unwrap();
        }
        _ => {}
    }
}

```

I have created updater.rs:
```rs
use tauri::App;
use tauri_plugin_dialog::{DialogExt, MessageDialogButtons};
use tauri_plugin_updater::UpdaterExt;

pub fn init(app: &App) {
    let handle = app.handle().clone();
        let _response = update_setup(handle).await;
    });
}

async fn update_setup(app: tauri::AppHandle) -> tauri::Result<()> {
  if let Some(update) = app.updater().unwrap().check().await.unwrap() {

    let current_version = app.package_info().version.to_string();
    let new_version = update.version.to_string();

    app.dialog()
        .message(&format!("Version {new_version} is now available -- you have {current_version}. \nWould you like to install it now?"))
        .title("A new version of AutoRemind App is available")
        .buttons(MessageDialogButtons::OkCancelCustom("Yes".to_string(), "No".to_string()))
        .show(|result| match result {
            false => {
                println!("Ignored update");
                },
            true => {
                    let mut downloaded = 0;
                    update.download_and_install(|chunk_length, content_length| {
                        downloaded += chunk_length;
                        println!("downloaded {downloaded} / {content_length:?}");
                    }, || {
                        println!("download finished");
                    }).await.unwrap();

                    println!("update installed. Restarting.");
                    app.restart(); // This shuts down but does not not relaunch new version
            });
        }
    });
  }

  Ok(())
}
```

capabilities/main.json
```json
{
    "$schema": "./schemas/desktop-schema.json",
    "identifier": "main-capability",
    "description": "Capability for the main window",
    "windows": ["main"],
    "local": false,
    "remote": {
        "urls": ["http://localhost:4000/", "http://localhost:4200/"]
    },
    "permissions": [
        "core:event:allow-emit",
        "core:event:allow-listen",
        "core:window:allow-set-resizable",
        "core:window:allow-set-size",
        "core:webview:allow-internal-toggle-devtools",
        "updater:allow-check",
        "updater:allow-download",
        "updater:allow-download-and-install",
        "updater:allow-install",
        "process:default",
        "process:allow-exit",
        "process:allow-restart"
    ]
}
```
Important parts of tauri.config.json:
```json
{
    "productName": "AutoRemind App",
    "mainBinaryName": "AutoRemind App",
    "version": "0.1.10000", //replaced by the CI/CD script
    "identifier": "com.autoremind.app",
    "bundle": {
        "active": true,
        "externalBin": [],
        "targets": [
            "nsis",
            "dmg",
            "app"
        ],
        "macOS": {
            "entitlements": null,
            "exceptionDomain": "",
            "frameworks": [],
            "providerShortName": null,
            "signingIdentity": "Developer ID Application: AutoRemind Inc (XXXXXXXX)"
        },
        "resources": [],
        "createUpdaterArtifacts": true
    },
    "app": {
        "windows": [],
        "security": {
            "csp": null
        }
    },
    "plugins": {
        "updater": {
            "endpoints": [
                "https://autoremind-app.xxx/updater.json"
            ],
            "windows": {
                "installMode": "passive"
            },
            "pubkey": "XXXXXXXXXXXX"
        }
    }
}
```

zx

### Reproduction

I have not tried building a simple app since this would require a full CI pipeline and new S3 bucket to test. But it's the same every time I do this:
1. Let our GitLab CI pipeline build a new set of artefacts (DMG and .app.tar.gz file with .app.tar.sig file), create new updater.json file and upload all to S3 - version 1
2. Build a newer version 2
3. Download version 1 DMG and install (move app to Applications)
4. Launch the application clicking on the app in Applications
5. Accept dialog asking me to upgrade to version 2
6. Observe the app closing (but not starting again)
7. Manually launch the app from Applications and observe (in an About dialog) that it's on version 2

### Expected behavior

After installing the update the app should start again.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 14.6.1 arm64 (X64)
    - node: 22.3.0
    - pnpm: 9.7.0
    - npm: 10.8.1

[-] Packages
    - tauri 🦀: 2.0.2
    - tauri-build 🦀: 2.0.1
    - wry 🦀: 0.44.1
    - tao 🦀: 0.30.3

[-] Plugins
    - tauri-plugin-fs 🦀: 2.0.1
    - tauri-plugin-process 🦀: 2.0.1
    - tauri-plugin-updater 🦀: 2.0.2
    - tauri-plugin-localhost 🦀: 2.0.1
    - tauri-plugin-dialog 🦀: 2.0.1
    - tauri-plugin-single-instance 🦀: 2.0.1

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist/apps/portal-desktop/.next
    - devUrl: http://localhost:4200/
```

### Stack trace

```text
I do not know how to produce a stack trace. I've built a version with --debug and launched from Termial with RUST_LOG=debug (also tested with "trace") and got no errors or exceptions - only the app printing that it was restarting. This is the last I see:

downloaded 30171747 / Some(30203582)
downloaded 30188131 / Some(30203582)
downloaded 30203582 / Some(30203582)
download finished
update installed. Restarting.
```

### Additional context

_No response_

### LINKED ISSUE #12310: [bug] `AppHandle::restart` may exit process without waiting for `RunEvent::Exit` event emit to plugins / app.
### Describe the bug

`AppHandle::restart` may exit process without waiting for `RunEvent::Exit` event emit to plugins / app.

This is upstream / super issue for:
- #11392
- tauri-apps/plugins-workspace#1692
- tauri-apps/plugins-workspace#2256

### Reproduction

Add log on `RunEvent::Exit`, make do something a little slow (like file system operation, or just sleep), and button to call `restart()`.

mini app based on npx tauri init. (removing lib crate)

main.rs

```rust
// Prevents additional console window on Windows in release, DO NOT REMOVE!!
#![cfg_attr(not(debug_assertions), windows_subsystem = "windows")]

fn main() {
    tauri::Builder::default()
        .invoke_handler(tauri::generate_handler![restart])
        .expect("error while running tauri application")
        .run(|_, event| match event {
            tauri::RunEvent::ExitRequested { .. } => {
                println!("Exit requested...");
                std::thread::sleep(std::time::Duration::from_millis(1));
            }
            tauri::RunEvent::Exit => {
                println!("Exiting...");
            }
            _ => {}
        });
}

#[tauri::command]
async fn restart(app: tauri::AppHandle) -> () {
    app.restart();
}
```

index.html

```html
<!doctype html>

<script>
  var restart = async () => {
    try {
      await __TAURI_INTERNALS__.invoke("restart")
    } catch (e) {
      console.error(e);
    }
  }
</script>

<button onclick="restart()">restart</button>
```

tauri.conf.json
```json
{
  "$schema": "https://schema.tauri.app/config/2.0.0-rc",
  "productName": "Tauri App",
  "version": "0.1.0",
  "identifier": "com.tauri.dev",
  "build": {
    "frontendDist": "."
  },
  "app": {
    "windows": [
      {
        "title": "Tauri",
        "width": 800,
        "height": 600,
        "resizable": true,
        "fullscreen": false
      }
    ]
  }
}
```

### Expected behavior

`tauri::RunEvent::Exit` should always be called

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 15.0.1 arm64 (X64)
    - node: 22.6.0
    - pnpm: 8.15.4
    - npm: 10.8.2
    - deno: deno 1.45.5

[-] Packages
    - tauri 🦀: 2.1.1
    - tauri-build 🦀: 2.0.3
    - wry 🦀: 0.47.2
    - tao 🦀: 0.30.8
    - tauri-cli 🦀: 2.0.0-rc.3
    - @tauri-apps/api : 2.1.1 (outdated, latest: 2.2.0)
    - @tauri-apps/cli : 2.1.0 (outdated, latest: 2.2.2)

[-] Plugins
    - tauri-plugin-single-instance 🦀: 2.2.0
    - tauri-plugin-updater 🦀: 2.3.0
    - tauri-plugin-fs 🦀: 2.2.0

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: out
    - devUrl: http://localhost:3030/
    - framework: React (Next.js)
    - bundler: Webpack
```

### Stack trace

_No response_

### Additional context

I'm looking for fix this issue with waiting for exit event in restart.

anatawa12:
Fixes #12310
Fixes #11392
Fixes tauri-apps/plugins-workspace#1692
Fixes tauri-apps/plugins-workspace#2256


anatawa12:
I've fixed format error. sorry

SergeFan:
Any updates to this PR?

anatawa12:
I'm waiting for review.

anatawa12:
The CI failure is corepack issue (https://github.com/nodejs/corepack/issues/625) so it's not my fault and nothing I can do.

Legend-Master:
Maybe we can just store a `bool` in `AppManager` to indicate that we need to restart and we can restart when we receive the next `RuntimeRunEvent::Exit` event in `App::run`?

Edit: actually, no since we have marked it as diverging function, maybe we can do https://github.com/tauri-apps/tauri/pull/10171#issuecomment-2202047402, by introducing `request_exit` and `request_restart`? Or maybe delay and break it in v3

anatawa12:
Yes I think we can add `request_restart` (and possibly `force_request_restart` which will not possible to cancel exit with `prevent_exit`) and deprecate current one (and possibly remove in v3)
With new approach, we have to fix every use cases of restart in each plugins.

I may make separated PR for this new approach.

Legend-Master:
> With new approach, we have to fix every use cases of restart in each plugins.

It should be fine, I think only the process plugin is affected in all the official plugins? (the restart command)

anatawa12:
Iseached for use cases and restart command in process plugin and all their use cases (even in js) would also needs migration.

lucasfernog:
this is now becoming a really complicated solution :joy: I had to push a change to make it not deadlock when restart() is called on the main thread

crates/tauri/src/manager/mod.rs
```diff
@@ -221,6 +221,13 @@ pub struct AppManager<R: Runtime> {
   pub(crate) invoke_key: String,
 
   pub(crate) channel_interceptor: Option<ChannelInterceptor<R>>,
+
+  // the value is set to true when the event loop is already exited
+  event_loop_exit_mutex: Mutex<bool>,
+  event_loop_exit_condvar: std::sync::Condvar,
```
lucasfernog, line 227: i'm not super familiar with condvar, is there an advantage in using this over e.g. a multi producer multi consumer channel?

anatawa12: AFAIK, mcmc channel can only one thread at once but condvar can wake multiple threads at once so I used condvar

anatawa12:
I don't enough time to implement request_restart so I'm keeping this PR as-is but should we push forward this complex solution instead of reimplement with request_restart?

lucasfernog:
I'm considering going forward with both - fix the restart() behavior but also deprecate it and implement request_restart

Legend-Master:
I'm thinking if something like this would just work for `request_restart`

```diff
diff --git a/crates/tauri/src/app.rs b/crates/tauri/src/app.rs
index 728183d69..7813097ca 100644
--- a/crates/tauri/src/app.rs
+++ b/crates/tauri/src/app.rs
@@ -10,6 +10,7 @@ use crate::{
   },
   manager::{webview::UriSchemeProtocol, AppManager, Asset},
   plugin::{Plugin, PluginStore},
+  process::restart,
   resources::ResourceTable,
   runtime::{
     window::{WebviewEvent as RuntimeWebviewEvent, WindowEvent as RuntimeWindowEvent},
@@ -42,7 +43,7 @@ use std::{
   borrow::Cow,
   collections::HashMap,
   fmt,
-  sync::{mpsc::Sender, Arc, MutexGuard},
+  sync::{atomic, mpsc::Sender, Arc, MutexGuard},
 };
 
 use crate::{event::EventId, runtime::RuntimeHandle, Event, EventTarget};
@@ -477,6 +478,18 @@ impl<R: Runtime> AppHandle<R> {
     crate::process::restart(&self.env());
   }
 
+  /// Restarts the app by triggering [`RunEvent::ExitRequested`] with code [`RESTART_EXIT_CODE`] and [`RunEvent::Exit`]..
+  pub fn request_restart(&self) {
+    self
+      .manager
+      .restart_on_exit
+      .store(true, atomic::Ordering::Relaxed);
+    if self.runtime_handle.request_exit(RESTART_EXIT_CODE).is_err() {
+      self.cleanup_before_exit();
+      crate::process::restart(&self.env());
+    }
+  }
+
   /// Sets the activation policy for the application. It is set to `NSApplicationActivationPolicyRegular` by default.
   ///
   /// # Examples
@@ -1084,6 +1097,9 @@ impl<R: Runtime> App<R> {
         let event = on_event_loop_event(&app_handle, RuntimeRunEvent::Exit, &manager);
         callback(&app_handle, event);
         app_handle.cleanup_before_exit();
+        if self.manager.restart_on_exit.load(atomic::Ordering::Relaxed) {
+          restart(&self.env());
+        }
       }
       _ => {
         let event = on_event_loop_event(&app_handle, event, &manager);
diff --git a/crates/tauri/src/manager/mod.rs b/crates/tauri/src/manager/mod.rs
index 882b141ca..ec6610864 100644
--- a/crates/tauri/src/manager/mod.rs
+++ b/crates/tauri/src/manager/mod.rs
@@ -6,7 +6,7 @@ use std::{
   borrow::Cow,
   collections::HashMap,
   fmt,
-  sync::{Arc, Mutex, MutexGuard},
+  sync::{atomic::AtomicBool, Arc, Mutex, MutexGuard},
 };
 
 use serde::Serialize;
@@ -221,6 +221,8 @@ pub struct AppManager<R: Runtime> {
   pub(crate) invoke_key: String,
 
   pub(crate) channel_interceptor: Option<ChannelInterceptor<R>>,
+
+  pub(crate) restart_on_exit: AtomicBool,
 }
 
 impl<R: Runtime> fmt::Debug for AppManager<R> {
@@ -322,6 +324,7 @@ impl<R: Runtime> AppManager<R> {
       resources_table: Arc::default(),
       invoke_key,
       channel_interceptor,
+      restart_on_exit: AtomicBool::new(false),
     }
   }

```

lucasfernog:
> I'm thinking if something like this would just work for `request_restart`
> 
> ```diff
> diff --git a/crates/tauri/src/app.rs b/crates/tauri/src/app.rs
> index 728183d69..7813097ca 100644
> --- a/crates/tauri/src/app.rs
> +++ b/crates/tauri/src/app.rs
> @@ -10,6 +10,7 @@ use crate::{
>    },
>    manager::{webview::UriSchemeProtocol, AppManager, Asset},
>    plugin::{Plugin, PluginStore},
> +  process::restart,
>    resources::ResourceTable,
>    runtime::{
>      window::{WebviewEvent as RuntimeWebviewEvent, WindowEvent as RuntimeWindowEvent},
> @@ -42,7 +43,7 @@ use std::{
>    borrow::Cow,
>    collections::HashMap,
>    fmt,
> -  sync::{mpsc::Sender, Arc, MutexGuard},
> +  sync::{atomic, mpsc::Sender, Arc, MutexGuard},
>  };
>  
>  use crate::{event::EventId, runtime::RuntimeHandle, Event, EventTarget};
> @@ -477,6 +478,18 @@ impl<R: Runtime> AppHandle<R> {
>      crate::process::restart(&self.env());
>    }
>  
> +  /// Restarts the app by triggering [`RunEvent::ExitRequested`] with code [`RESTART_EXIT_CODE`] and [`RunEvent::Exit`]..
> +  pub fn request_restart(&self) {
> +    self
> +      .manager
> +      .restart_on_exit
> +      .store(true, atomic::Ordering::Relaxed);
> +    if self.runtime_handle.request_exit(RESTART_EXIT_CODE).is_err() {
> +      self.cleanup_before_exit();
> +      crate::process::restart(&self.env());
> +    }
> +  }
> +
>    /// Sets the activation policy for the application. It is set to `NSApplicationActivationPolicyRegular` by default.
>    ///
>    /// # Examples
> @@ -1084,6 +1097,9 @@ impl<R: Runtime> App<R> {
>          let event = on_event_loop_event(&app_handle, RuntimeRunEvent::Exit, &manager);
>          callback(&app_handle, event);
>          app_handle.cleanup_before_exit();
> +        if self.manager.restart_on_exit.load(atomic::Ordering::Relaxed) {
> +          restart(&self.env());
> +        }
>        }
>        _ => {
>          let event = on_event_loop_event(&app_handle, event, &manager);
> diff --git a/crates/tauri/src/manager/mod.rs b/crates/tauri/src/manager/mod.rs
> index 882b141ca..ec6610864 100644
> --- a/crates/tauri/src/manager/mod.rs
> +++ b/crates/tauri/src/manager/mod.rs
> @@ -6,7 +6,7 @@ use std::{
>    borrow::Cow,
>    collections::HashMap,
>    fmt,
> -  sync::{Arc, Mutex, MutexGuard},
> +  sync::{atomic::AtomicBool, Arc, Mutex, MutexGuard},
>  };
>  
>  use serde::Serialize;
> @@ -221,6 +221,8 @@ pub struct AppManager<R: Runtime> {
>    pub(crate) invoke_key: String,
>  
>    pub(crate) channel_interceptor: Option<ChannelInterceptor<R>>,
> +
> +  pub(crate) restart_on_exit: AtomicBool,
>  }
>  
>  impl<R: Runtime> fmt::Debug for AppManager<R> {
> @@ -322,6 +324,7 @@ impl<R: Runtime> AppManager<R> {
>        resources_table: Arc::default(),
>        invoke_key,
>        channel_interceptor,
> +      restart_on_exit: AtomicBool::new(false),
>      }
>    }
> ```

that should work - i'm fixing some other bugs but i can test this later

.changes/restart-may-not-publish-exit-event.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri": patch:bug
+---
+
+`AppHandle::restart()` may not send `RunEvent::Exit` event before exiting the application.
```
lucasfernog:
```suggestion
`AppHandle::restart()` now waits for `RunEvent::Exit` to be delivered before restarting the application.
```

lucasfernog (dismissed):

Legend-Master:
@lucasfernog the deadlock around the event loop seems to be a recurring theme now, could you tell where the deadlock happens? I only know the plugin lock which can be solved by not having the callbacks being MutFn so we don't need a lock to call them (breaking change so needs to wait for v3)

Also I think this main thread exit implementation should probably be moved to tauri-runtime-wry? Using UnsafeCell here feels so wrong to me

lucasfernog:
In this case we have a deadlock because we're running `wait_for_event_loop_exit` in the main thread, meaning `notify_event_loop_exit` does not have a chance to be triggered. The problem is that request_exit uses the event loop proxy, so the Exit callback is only triggered in the next iteration https://github.com/tauri-apps/tauri/blob/3fb8d7ca6b53ffdd6d4e5051e48d6fe78ea2c5aa/crates/tauri-runtime-wry/src/lib.rs#L2377

I'll see if there's a way to rewrite it in a better way, but I'm not super confident.

lucasfernog:
I would probably need to use the same mechanism to store the event loop callback and use it in a thread-safe struct (but ensuring it only gets called in the main thread)

Legend-Master:
> In this case we have a deadlock because we're running `wait_for_event_loop_exit` in the main thread, meaning `notify_event_loop_exit` does not have a chance to be triggered. The problem is that request_exit uses the event loop proxy, so the Exit callback is only triggered in the next iteration

Got it now, so the main problem is that the function is blocking a the event loop while waiting for the event loop to move and receive the next message

> // NOTE: request_exit cannot use the `send_user_message` function because it accesses the event loop callback 

I wonder why is this though, if we're not blocking on the main thread, this should be fine?

lucasfernog:
> > In this case we have a deadlock because we're running `wait_for_event_loop_exit` in the main thread, meaning `notify_event_loop_exit` does not have a chance to be triggered. The problem is that request_exit uses the event loop proxy, so the Exit callback is only triggered in the next iteration
> 
> 
> 
> Got it now, so the main problem is that the function is blocking a the event loop while waiting for the event loop to move and receive the next message
> 
> 
> 
> > // NOTE: request_exit cannot use the `send_user_message` function because it accesses the event loop callback 
> 
> 
> 
> I wonder why is this though, if we're not blocking on the main thread, this should be fine?

yes you're right
"it should be fine" yes, but that's why we're using UnsafeCell, we're calling from possibly another thread but checking if we're on the main one so we do not deadlock

Legend-Master:
> "it should be fine" yes

I see, so maybe we should update the comments to include where it would dead lock if we do it on the main thread? I was quite confused by it (still somewhat confused right now 😂)

And just one more thing, since the `callback` is a `MutFn`, if we call `restart` inside that callback, wouldn't we break the only a single mutable reference at a time through `UnsafeCell`?

lucasfernog:
> > "it should be fine" yes
> 
> I see, so maybe we should update the comments to include where it would dead lock if we do it on the main thread? I was quite confused by it (still somewhat confused right now 😂)
> 
> And just one more thing, since the `callback` is a `MutFn`, if we call `restart` inside that callback, wouldn't we break the only a single mutable reference at a time through `UnsafeCell`?

yeah that's something i was thinking about too.. maybe we should crash instead (so it's neither a deadlock nor a crazy unsafecell usage)
EDIT: or maybe just not trigger RunEvent::Exit in this case

Legend-Master:
In this case I'm leaning towards just leave this function as is and deprecate it, then introduce `request_restart` to be honest

Since this function has never triggered the `Exit` event before, I think it's fine to just change the docs to say it doesn't trigger the event and restart immediately, if you want the event, use `request_restart` instead

lucasfernog:
> In this case I'm leaning towards just leave this function as is and deprecate it, then introduce `request_restart` to be honest
> 
> Since this function has never triggered the `Exit` event before, I think it's fine to just change the docs to say it doesn't trigger the event and restart immediately, if you want the event, use `request_restart` instead

but that's not the only problem this PR tries to solve.. it also fixes the race condition for restart/exit

Legend-Master:
> but that's not the only problem this PR tries to solve.. it also fixes the race condition for restart/exit

Wait, what race condition?

lucasfernog:
> > but that's not the only problem this PR tries to solve.. it also fixes the race condition for restart/exit
> 
> Wait, what race condition?

@Legend-Master https://github.com/tauri-apps/tauri/issues/11392#issuecomment-2444618724

lucasfernog (approved):

Legend-Master:
> > > but that's not the only problem this PR tries to solve.. it also fixes the race condition for restart/exit
> > 
> > 
> > Wait, what race condition?
> 
> @Legend-Master [#11392 (comment)](https://github.com/tauri-apps/tauri/issues/11392#issuecomment-2444618724)

I see, but I still feel like we could just remove `self.runtime_handle.request_exit(RESTART_EXIT_CODE)` and call `crate::process::restart(&self.env())` directly, anyways, I'm not opposed to this either

PR #12324, opened 2025-01-09
fix(bundler/nsis): Include WebView2Loader.dll if found to match msi

FabianLars:


crates/tauri-bundler/src/bundle/windows/nsis/mod.rs
```diff
@@ -599,6 +599,17 @@ fn generate_resource_data(settings: &Settings) -> crate::Result<ResourcesMap> {
 
   let mut added_resources = Vec::new();
 
+  // Adding WebViewer2Loader.dll in case windows-gnu toolchain is used
```
amrbashir, line 603: should we make it behind a check for `windows-gnu` then?

FabianLars: that was the idea at first but none of the Settings properties contained useful info and this want meant to be a hotfix. i'll take a closer look and check if there's something we can use here.

FabianLars: okay idk what i did wrong but now settings.target() reports gnu again. i use that then and see if we still get reports about missing dll

amrbashir (approved):

PR #12328, opened 2025-01-09
chore: Fix clippy 1.84 warnings

FabianLars:
also had to make use of the new msrv aware cargo feature to update the lockfile because some dependency of napi caused unknown cfg issues but i couldn't figure out which one 🙃

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -583,9 +583,9 @@ dependencies = [
 
 [[package]]
 name = "aws-config"
-version = "1.5.13"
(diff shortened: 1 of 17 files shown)


FabianLars:
> also had to make use of the new msrv aware cargo feature to update the lockfile because some dependency of napi caused unknown cfg issues but i couldn't figure out which one 🙃

build fails because of msrv :))

Legend-Master (approved):

PR #12365, opened 2025-01-12
fix(windows): Resolve broken installation directory handling in MSI & NSIS, preventing duplicate installations during updates

don41382:
This PR addresses an issue where the MSI updater installs the application to the default directory (`C:\Program Files\<App Name>`), even if the application was initially installed in a custom directory by NSIS, such as the user-specific path (`C:\Users\<User>\AppData\Local\<App Name>`).

### Root Cause
The problem arises from a mismatch between NSIS and MSI in handling the installation directory path:
- The NSIS installer sets the installation path entry as a "no-key" (standard) value.
- The MSI updater, however, looks for a key named `InstallDir` to locate the existing installation directory.

### Consequences
This mismatch causes MSI to fail in detecting the original installation directory during updates, resulting in two separate installations:
- **Machine-wide directory**: `C:\Program Files\<App Name>`
- **User-specific directory**: `C:\Users\<User>\AppData\Local\<App Name>`

### Fix
This PR resolves the issue by ensuring proper handover of the installation directory between NSIS and MSI. Additionally, it ensures backwards compatibility to avoid breaking existing setups.

### DIFF
--- a/crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
+++ b/crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
@@ -656,7 +656,8 @@ Section Install
   WriteUninstaller "$INSTDIR\uninstall.exe"
 
   ; Save $INSTDIR in registry for future installations
-  WriteRegStr SHCTX "${MANUPRODUCTKEY}" "" $INSTDIR
+  WriteRegStr SHCTX "${MANUPRODUCTKEY}" "InstallDir" $INSTDIR
+
 
   !if "${INSTALLMODE}" == "both"
     ; Save install mode to be selected by default for the next installation such as updating
--- a/crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
+++ b/crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
@@ -656,8 +656,7 @@ Section Install
   WriteUninstaller "$INSTDIR\uninstall.exe"
 
   ; Save $INSTDIR in registry for future installations
-  WriteRegStr SHCTX "${MANUPRODUCTKEY}" "InstallDir" $INSTDIR
-
+  WriteRegStr SHCTX "${MANUPRODUCTKEY}" "" $INSTDIR
 
   !if "${INSTALLMODE}" == "both"
     ; Save install mode to be selected by default for the next installation such as updating
--- a/.changes/fix-bundler-msi-install-dir-lookup.md
+++ b/.changes/fix-bundler-msi-install-dir-lookup.md
@@ -1,0 +1,5 @@
+---
+tauri-bundler: 'patch:bug'
+---
+
+Fixed an issue that caused the `.msi` installer not to lookup the `INSTALLDIR` set in the `nsis` installer.
--- a/crates/tauri-bundler/src/bundle/windows/msi/main.wxs
+++ b/crates/tauri-bundler/src/bundle/windows/msi/main.wxs
@@ -70,9 +70,12 @@
         <Property Id="ARPURLUPDATEINFO" Value="{{homepage}}"/>
         {{/if}}
 
-        <!-- initialize with previous InstallDir -->
         <Property Id="INSTALLDIR">
-            <RegistrySearch Id="PrevInstallDirReg" Root="HKCU" Key="Software\\{{manufacturer}}\\{{product_name}}" Name="InstallDir" Type="raw"/>
+          <!-- First attempt: Search for "InstallDir" -->
+          <RegistrySearch Id="PrevInstallDirWithName" Root="HKCU" Key="Software\\{{manufacturer}}\\{{product_name}}" Name="InstallDir" Type="raw" />
+
+          <!-- Second attempt: If the first fails, search for the default value (which is how the old nsis installer stored the value) -->
+          <RegistrySearch Id="PrevInstallDirNoName" Root="HKCU" Key="Software\\{{manufacturer}}\\{{product_name}}" Type="raw" />
         </Property>
 
         <!-- launch app checkbox -->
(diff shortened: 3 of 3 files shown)


crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
```diff
@@ -656,7 +656,8 @@ Section Install
   WriteUninstaller "$INSTDIR\uninstall.exe"
 
   ; Save $INSTDIR in registry for future installations
-  WriteRegStr SHCTX "${MANUPRODUCTKEY}" "" $INSTDIR
+  WriteRegStr SHCTX "${MANUPRODUCTKEY}" "InstallDir" $INSTDIR
```
Legend-Master: Do we need to change this? Also I think this will break the NSIS updaters for custom install locations

Legend-Master (changes requested):

FabianLars:
btw just a bit of context. nsis -> msi updates were never supposed to work. officially only msi -> nsis updates are/were supported (intentionally).

don41382:
> btw just a bit of context. nsis -> msi updates were never supposed to work. officially only msi -> nsis updates are/were supported (intentionally).

Thanks for the quick response. Maybe I got this wrong. You mean, that the enduser does not initially installs the app with the `.exe` (nsis), but rather with the `.msi`?

But how to I enable `"installMode": "currentUser"` with a `.msi`? I am a little bit confused.

crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
don41382: another solution is, to keep the non-value key and have the InstallDir in addition to it. This would help to make the `.msi` updater not to choose the wrong location and keep custom installations working - right?

FabianLars:
> Thanks for the quick response. Maybe I got this wrong. You mean, that the enduser does not initially installs the app with the .exe (nsis), but rather with the .msi?

Nope. First of all, 99% of apps should not ship _both_ the msi and nsis exe but decide on one format before distributing their app.

The `msi -> nsis` flow is supported because the nsis installer is newer and is/was the recommended installer but was added when there were already many apps using msi installers. We just wanted to give devs a way to switch to the new format if they wish too. That of course doesn't mean that we shouldn't support `nsis -> msi` as well, we just never saw much of a reason to.

And generally, even though cross installer updates are supported, apps shouldn't make use of that feature without good reason.

> But how to I enable "installMode": "currentUser" with a .msi? I am a little bit confused.

The other way around. You would enable the perMachine mode for the nsis installer.

don41382:
> Nope. First of all, 99% of apps should not ship _both_ the msi and nsis exe but decide on one format before distributing their app.

I am shipping (user download from homepage) currently only the nsis version with `"installMode": "currentUser"`, which I understand from your comment is the default. I think, it's the better choice, because it allows to install without requiring admin privileges. 

On updating the windows version, I started to see duplicated installation on the same machine. The `.msi` didn't recall the correct installation directory and created a second installation in the "machine path". I guess, if you would have first started with `.msi`, than the "InstallIDir" would have set correctly and updating would not be a problem.

So two version existed and the Autostart ended up starting the old version. Which of course is a mess.

Maybe I am one of the few who started with distributing the nsis. Installing the `.msi` first, will set are correct `InstallDir` in the registry.

IMO tauri should support both ways and it should not matter which was installed fist - otherwise this creates just confusion. Wdyt?

FabianLars:
> Maybe I am one of the few who started with distributing the nsis. 

Nah, nowadays that should be more common. You just made the same mistake everyone does, also shipping the msi at the same time, or at least using it in the updater - which tbf is our fault especially if you use tauri-action which prios the msi for legacy reasons.

> IMO tauri should support both ways and it should not matter which was installed fist - otherwise this creates just confusion. Wdyt?

Already said that i'm fine with that. 🤷

don41382:
> Already said that i'm fine with that. 🤷

Perfect. If you are okay with the changes, than we just have to wait for @Legend-Master for his remarks regarding backwards compatibility to custom installation locations.

crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
Legend-Master: If we're already detecting the install location from both places in the `.msi` installer, I don't think we really need to change this for the NSIS installer?

don41382: That is true. We can leave the `installer.nsi` changes out. It felt just more streamline having a the `InstallDir` stored in the same location in both - installer and updater. But to be fair, I don't understand how currently custom installation make use of the regex entry.

Legend-Master: > We can leave the `installer.nsi` changes out.

Let's do this then

> It felt just more streamline having a the `InstallDir` stored in the same location in both - installer and updater.

Yeah, I would probably want the NSIS and Wix installer saving the location to the same place, but since we have already done it differently, I don't really want to maintain 2 different locations personally, especially when it's done in a very error prone language like NSIS

don41382: I reverted the `installer.nsi` commit and changed the comment slightly.

Let me know if it looks good for you.

Legend-Master: Yeah, looks good to me! Since I'm not the most familiar with the Wix installer code, I'll leave the review to other members (looks good on my part)

don41382: Thanks - I've also tried it and it seems to work. But please somebody else check it too!

FabianLars:
The wix change is fine, but we need a changefile for tauri-bundler before we can merge this and _all_ commits must be signed (only the first one is)

don41382:
@FabianLars All commits are verified

should I squash all commits and create a new `CHANGELOG.md` entry?

FabianLars:
> should I squash all commits

not necessary we sqaush merge anyway

> create a new CHANGELOG.md entry?

No, our release pipeline does that. A changefile in https://github.com/tauri-apps/tauri/tree/dev/.changes similar to https://github.com/tauri-apps/tauri/blob/fix/bundler-dlls/.changes/fix-bundler-wix-dlls.md would be required for that though

Legend-Master (approved):

Issue #12381, opened 2025-01-13
[bug] `tauri-driver` serializes `webviewOptions` as `null` if not present, but an object is required


bicarlsen:
### Describe the bug

`tauri-driver` serializes `webviewOptions` as `null` if not present, but an object is required. This results in the error 
```shell
WebDriverError: invalid argument: cannot parse capability: ms:edgeOptions from invalid argument: 
cannot parse webviewOptions from invalid argument: must be a dictionary when running "http://localhost:4444/session" with method "POST" and args 
"{"capabilities":{"alwaysMatch":{"tauri:options":{"application":"<path/to/binary>"}},"firstMatch":[{}]}}"
```

### Reproduction

Follow the [testing guide for WebDriverIO](https://v2.tauri.app/develop/tests/webdriver/example/webdriverio/).

### Expected behavior

Options should be parsed with correct types.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.26100 x86_64 (X64)
    ✔ WebView2: 131.0.2903.112
    ✔ MSVC: Visual Studio Build Tools 2022
    ✔ rustc: 1.83.0 (90b35a623 2024-11-26)
    ✔ cargo: 1.83.0 (5ffbef321 2024-10-29)
    ✔ rustup: 1.27.1 (54dd3d00f 2024-04-24)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (environment override by RUSTUP_TOOLCHAIN)
    - node: 23.6.0
    - pnpm: 9.12.3
    - npm: 10.9.2

[-] Packages
    - tauri 🦀: 2.2.2
    - tauri-build 🦀: 2.0.5
    - wry 🦀: 0.48.0
    - tao 🦀: 0.31.1
    - tauri-cli 🦀: 2.2.2
    - @tauri-apps/api : not installed!
    - @tauri-apps/cli : 2.2.2 (outdated, latest: 2.2.4)

[-] Plugins
    - tauri-plugin-log 🦀: 2.2.0
    - @tauri-apps/plugin-log : not installed!

[-] App
    - build-type: bundle
    - CSP: connect-src ipc: http://ipc.localhost; font-src https://fonts.gstatic.com; style-src 'unsafe-inline' 'self' https://fonts.googleapis.com; default-src 'self' customprotocol: asset:; img-src 'self' asset: http://asset.localhost blob: data:
    - frontendDist: ../dist
    - devUrl: http://localhost:1420/
```

[...]

PR #12383, opened 2025-01-13
fix(tauri-driver): Parse ms:edgeOptions separately

bicarlsen:
Resolves Issue #12381.


crates/tauri-driver/src/server.rs
```diff
@@ -50,13 +50,19 @@ impl TauriOptions {
 
   #[cfg(target_os = "windows")]
   fn into_native_object(self) -> Map<String, Value> {
+    let mut ms_edge_options = Map::new();
+    ms_edge_options.insert("binary".into(), json!(self.application));
+    ms_edge_options.insert("args".into(), self.args.into());
+
+    #[cfg(target_os = "windows")]
```
FabianLars:
```suggestion
```
whole function is behind that flag already

FabianLars (approved):

PR #12385, opened 2025-01-13
feat: allow overriding .rpm and .deb package names

AlexProgrammerDE:
My app SoulFire gets renamed to the package name `soul-fire`, which I don't want. I want it to be called `soulfire`. This is why I need these changes. There was also a discord conversation about this: https://discord.com/channels/616186924390023171/986184094050316358/1328375059919671336

### DIFF
--- a/crates/tauri-bundler/src/bundle/linux/debian.rs
+++ b/crates/tauri-bundler/src/bundle/linux/debian.rs
@@ -50,7 +50,12 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
   };
   let package_base_name = format!(
     "{}_{}_{}",
-    settings.product_name(),
+    &settings
+      .deb()
+      .package_name
+      .as_ref()
+      .cloned()
+      .unwrap_or_else(|| settings.product_name().to_string()),
     settings.version_string(),
     arch
   );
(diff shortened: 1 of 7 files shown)


pymete (approved):

lucasfernog:
what about adding a `src-tauri/tauri.linux.conf.json` with the following contents instead:
```json
{
  "productName": "soufire"
}
```
wouldn't that be enough? that configuration would automatically only apply to Linux builds

zakhenry:
@FabianLars can this be bumped to the next milestone since there is open questions now about there being an alternative approach? It's the last item in 2.4 and that's holding back some useful features

lucasfernog:
we just released 2.4 fyi :)

AlexProgrammerDE:
> what about adding a `src-tauri/tauri.linux.conf.json` with the following contents instead:
> 
> ```json
> {
>   "productName": "soufire"
> }
> ```
> 
> wouldn't that be enough? that configuration would automatically only apply to Linux builds

Hi, that would be a valid alternative.

lucasfernog:
I'll close this since there's already a platform-specific configuration feature that lets you change the product name based on target platform. Thanks for your contribution!

FabianLars:
A bit of additional context that may require re-opening this PR: https://github.com/tauri-apps/tauri/issues/14366

PR #12401, opened 2025-01-14
ci: Change MSRV check to use 1.84 resolver

FabianLars:
@amrbashir pinging you because back when i changed the msrv stuff and committed the lockfile you wanted to keep this check around, are you okay with this change?


amrbashir (approved):
LGTM, hopefully soon we won't need this MSRV check anymore

FabianLars:
one can dream :D

PR #12406, opened 2025-01-15
fix(cli): let xcode handle building for `ios build --open`

bradleat:
This pull request includes a small change so that `ios build --open` does not build before launching xcode.
This is done because xcode will bulid the entire project and some users who are using the `ios build --open` flow may be using some specific xcode setup that `call_for_targets_with_fallback` will fail to build. For instance, building a project with a Apple Watch Companion app will not work otherwise.

Changes in `fn run_build` in `crates/tauri-cli/src/mobile/ios/build.rs`:

* Addedd a conditional return statement to handle the `options.open` case early in the `run_build` function. This ensures that the function exits immediately if `options.open` is true, returning the `handle` object.

### DIFF
--- a/.changes/change-pr-12406.md
+++ b/.changes/change-pr-12406.md
@@ -1,0 +1,6 @@
+---
+"@tauri-apps/cli": "patch:enhance"
+"tauri-cli": "patch:enhance"
+---
+
+`ios build --open` will now let xcode start the rust build process.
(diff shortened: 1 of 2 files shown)


FabianLars (approved):

FabianLars:
kinda weirded out by slowly becomming our macos maintainer ngl 😂 I'm not too familiar with building directly in xcode but from what i understand this is working as desired.

FabianLars:
btw, for your next contributions (if any) we'll require all commits to be signed. will make a one time exception here :)

PR #12423, opened 2025-01-16
fix(macos): frameworks being signed with entitlements unnecessarily

tr3ysmith:
limited signing with entitlements for executables only and not frameworks as certain entitlements like multicast support will break frameworks that aren't registered for it.


FabianLars (approved):
i hope there aren't use-cases for the old behavior x) Thanks!

PR #12427, opened 2025-01-17
fix(nsis): uninstaller installdir reg clean up

Legend-Master:
Fix NSIS uninstaller not cleaning up `Software\${MANUFACTURER}\${PRODUCTNAME}` registry key used for the install location

### DIFF
--- a/.changes/nsis-cleanup-installdir-reg.md
+++ b/.changes/nsis-cleanup-installdir-reg.md
@@ -1,0 +1,5 @@
+---
+'tauri-bundler': 'patch:bug'
+---
+
+Fix NSIS uninstaller not cleaning up `Software\${MANUFACTURER}\${PRODUCTNAME}` registry key used for the install location
--- a/.changes/nsis-cleanup-installdir-reg.md
+++ b/.changes/nsis-cleanup-installdir-reg.md
@@ -2,4 +2,4 @@
 'tauri-bundler': 'patch:bug'
 ---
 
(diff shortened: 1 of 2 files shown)


amrbashir (commented):
This is actually not a bug but rather a desired behavior, which saves the location and install language for future installations. I have seen this behavior countless times in other installers.

Legend-Master:
Since I saw installer language was removed on uninstall through `DeleteRegValue HKCU "${MANUPRODUCTKEY}" "Installer Language"` before, I thought this was a bug instead of an intentional thing

What about we only do it if the "Delete the application data" gets checked? Probably makes sense to have everything cleaned up in that case (also in this case maybe we also move the delete language registry key?)

amrbashir:
That sounds fine

.changes/nsis-cleanup-installdir-reg.md
```diff
@@ -0,0 +1,5 @@
+---
+'tauri-bundler': 'patch:bug'
+---
+
+Fix NSIS uninstaller not cleaning up `Software\${MANUFACTURER}\${PRODUCTNAME}` registry key used for the install location
```
amrbashir:
```suggestion
Clean up `Software\${MANUFACTURER}\${PRODUCTNAME}` registry key in the NSIS uninstaller if "Delete application data" option is checked when uninstalling. 
```

amrbashir (approved):

PR #12438, opened 2025-01-18
Add log message to inform user of DevCommand being used in tauri-cli

3lpsy:
This PR adds a simple log message to output the `cargo` command that is being used when running `cargo tauri dev`. There is only a single line of code added to `desktop::run_dev`:

```
  log::info!(action = "Running"; "DevCommand (`{} {}`)", &dev_cmd.get_program().to_string_lossy(), dev_cmd.get_args().map(|arg| arg.to_string_lossy()).fold(String::new(), |acc, arg| format!("{acc} {arg}")));
```

Reasoning:
I was having trouble determining what was happening at runtime and what "magic" the `cargo tauri dev` command was doing. Explicitly stating what's being run could be useful to others who are unfamiliar with the `tauri-cli` (like myself) to more easily determine what the CLI is actually doing. If you disagree and think it'd add too much clutter to the output, feel free to close the PR.

### DIFF
--- a/.changes/log-dev-command.md
+++ b/.changes/log-dev-command.md
@@ -1,0 +1,6 @@
+---
+"tauri-cli": enhance
+"@tauri-apps/cli": enhance
+---
+
+Log the command used to start the rust app in development.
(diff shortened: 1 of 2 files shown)


amrbashir (dismissed):
Thank you, this is awesome

amrbashir (dismissed):

amrbashir (approved):

PR #12444, opened 2025-01-19
Document `Emitter` trait method panics

sftse:
Almost all methods of the `Listener` and `Emitter` traits take an event name and will panic if that name does not satisfy some preconditions. This is not documented and this PR fixes this. This PR also moves the panic locations closer to the user-facing code, making it immediately clear what has gone wrong and what methods were responsible for the panic. Prior to this change, the panic pointed to a location of little use to a user of `tauri` unless he reverse engineers how that code was called within `tauri`.

As a drive-by improvement, it was a bit unclear that the `Emitter` trait is purely defined based on methods on the `AppManager`. This PR makes all methods on the `Emitter` trait default methods and saves boilerplate. This change should not have any downsides because `Emitter` is a sealed trait that can only be implemented inside `tauri`.


amrbashir (commented):
Thank you, this LGTM, could you add a change file in `.changes` directory?

> This PR makes all methods on the `Emitter` trait default methods and saves boilerplate

This was done intentionally to customize the documentation example based on the used type, but I guess it is fine to remove.

FabianLars:
Should we even panic here? It may be good for development for most cases but what if devs use dynamic event names that are derived from some http request for example?

idk panicking in a gui library just always sounds awaful to me

amrbashir:
yeah maybe returning an Error here is better, the methods already return Result type

sftse:
I've added some changes that inch closer in that direction, replacing the assert with a checked `EventName` constructor along with some internal refactors, no longer passing `&str` and `String` but `EventName<&str>` and `EventName<String>`. I think the code structure benefits from these changes regardless of whether the panic should be removed or not. If the changes are not palatable (they're a bit larger than I originally set out to do with this PR) I can revert.

While `Emitter` does return an error, `Listener` does not, and will also panic when given bad event names. So I think it is not possible to get rid of the panics without breaking stable api.

crates/tauri/src/lib.rs
```diff
@@ -929,7 +938,12 @@ pub trait Emitter<R: Runtime>: sealed::ManagerBase<R> {
   ///   app.emit("synchronized", ());
   /// }
   /// ```
-  fn emit<S: Serialize + Clone>(&self, event: &str, payload: S) -> Result<()>;
+  /// # Panics
+  /// Will panic if `event` contains characters other than alphanumeric, `-`, `/`, `:` and `_`
+  fn emit<S: Serialize + Clone>(&self, event: &str, payload: S) -> Result<()> {
+    let event = EventName::new(event).unwrap();
```
amrbashir: instead of panicking here, just propagate the result  from `EventName::new`

crates/tauri/src/event/plugin.rs
```diff
@@ -12,15 +12,69 @@ use crate::plugin::{Builder, TauriPlugin};
...
-pub struct EventName(String);
+/// Checks if an event name is valid.
+fn is_event_name_valid(event: &str) -> bool {
+  event
+    .chars()
+    .all(|c| c.is_alphanumeric() || c == '-' || c == '/' || c == ':' || c == '_')
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct EventName<S = String>(S);
```
amrbashir: Let's move this type to `tauri/src/event/mod.rs`

crates/tauri/src/event/plugin.rs
```diff
@@ -12,15 +12,69 @@ use crate::plugin::{Builder, TauriPlugin};
...
   type Target = str;
 
   fn deref(&self) -> &Self::Target {
-    &self.0
+    self.0.as_ref()
+  }
+}
+
+#[derive(Clone)]
+pub(crate) struct InvalidEventName;
```
amrbashir: This could be just a variant on `tauri::Error` type

crates/tauri/src/event/plugin.rs
```diff
@@ -12,15 +12,69 @@ use crate::plugin::{Builder, TauriPlugin};
...
+      return Err(InvalidEventName);
+    }
+    Ok(EventName(s))
+  }
+  pub(crate) fn into_inner(self) -> S {
+    self.0
+  }
+  pub(crate) fn as_str_event(&self) -> EventName<&str> {
+    EventName(self.0.as_ref())
+  }
```
amrbashir:
```suggestion
  }

  pub(crate) fn into_inner(self) -> S {
    self.0
  }

  pub(crate) fn as_str_event(&self) -> EventName<&str> {
    EventName(self.0.as_ref())
  }
```

crates/tauri/src/manager/mod.rs
```diff
@@ -535,8 +537,10 @@ impl<R: Runtime> AppManager<R> {
     feature = "tracing",
     tracing::instrument("app::emit", skip(self, payload))
   )]
+  /// # Panics
+  /// Will panic if `event` contains characters other than alphanumeric, `-`, `/`, `:` and `_`
   pub fn emit<S: Serialize + Clone>(&self, event: &str, payload: S) -> crate::Result<()> {
-    assert_event_name_is_valid(event);
+    let event = EventName::new(event).unwrap();
```
amrbashir: this seems unnecessary since `Emitter::emit` does the same check

crates/tauri/src/manager/mod.rs
```diff
@@ -560,12 +564,14 @@ impl<R: Runtime> AppManager<R> {
...
   )]
+  /// # Panics
+  /// Will panic if `event` contains characters other than alphanumeric, `-`, `/`, `:` and `_`
   pub fn emit_filter<S, F>(&self, event: &str, payload: S, filter: F) -> crate::Result<()>
   where
     S: Serialize + Clone,
     F: Fn(&EventTarget) -> bool,
   {
-    assert_event_name_is_valid(event);
+    let event = EventName::new(event).unwrap();
```
amrbashir: same

crates/tauri/src/manager/webview.rs
```diff
@@ -680,14 +680,14 @@ fn on_webview_event<R: Runtime>(webview: &Webview<R>, event: &WebviewEvent) -> c
           paths: Some(paths),
           position,
         };
-        webview.emit_to_webview(DRAG_ENTER_EVENT, payload)?
+        webview.emit_to_webview(&DRAG_ENTER_EVENT, payload)?
```
amrbashir: isn't `DRAG_ENTER_EVENT` a copy type? can't we remove the `&`?

crates/tauri/src/event/mod.rs
```diff
@@ -124,20 +111,20 @@ impl FromStr for EventTarget {
 #[derive(Clone)]
 pub struct EmitArgs {
   /// Raw event name.
-  pub event_name: String,
+  pub event_name: EventName<String>,
```
amrbashir: let's remove this field and use the `event` field, we don't need both, see https://github.com/tauri-apps/tauri/pull/12460#discussion_r1923889703

amrbashir (changes requested):

crates/tauri/src/manager/webview.rs
sftse: Was indicative of forgetting to make `emit_to_webview` also take `EventName`. The `&event` is sort of magic to get Rust to deref coerce `&EventName<&str>` to `&str`. Since this is an [unidiomatic use](https://rust-lang.github.io/api-guidelines/predictability.html#c-deref) of `Deref`, I've removed the trait impl in the newest version.

amrbashir: I didn't mean to deref, I meant that `DRAG_ENTER_EVENT` is of type `EventName<&str>` so it should be copy and so we can have `emit_to_webview` method take `EventName<&str>` and use this const directly without `&`.

crates/tauri/src/event/plugin.rs
```diff
@@ -108,3 +80,15 @@ pub(crate) fn init<R: Runtime>() -> TauriPlugin<R> {
     .invoke_handler(crate::generate_handler![listen, unlisten, emit, emit_to])
     .build()
 }
+
+#[cfg(test)]
```
amrbashir: looks like these tests should be in `event/mod.rs` instead

crates/tauri/src/event/mod.rs
```diff
@@ -123,21 +178,18 @@ impl FromStr for EventTarget {
...
 
 impl EmitArgs {
-  pub fn new<S: Serialize>(event: &str, payload: S) -> crate::Result<Self> {
+  pub fn new<S: Serialize>(event: EventName<&str>, payload: S) -> crate::Result<Self> {
     #[cfg(feature = "tracing")]
     let _span = tracing::debug_span!("window::emit::serialize").entered();
     Ok(EmitArgs {
-      event_name: event.into(),
-      event: serde_json::to_string(event)?,
+      event: serde_json::to_string(event.as_str())?,
```
amrbashir: Let's remove this serialization, it is not needed
```suggestion
      event,
```

don't forget to also update `emit_js_script` function down below to add wrapping quotes in the JS script.

sftse:
This is tangentially related, but I really think the `Emitter` trait methods should take `&S` instead of `S`, a breaking change for the future. Using `S` prevents the compiler from being as helpful as it can be, and will suggest to clone if a non-copy `S` is passed, not realizing that `&S` also implements the trait bounds.

amrbashir (approved):
Thank you

PR #12445, opened 2025-01-19
fix(cli): don't force native-tls feature on desktop

### LINKED ISSUE #12448: [bug] undeclared hyper-tls dependency of tauri causes failure with offline builds
### Describe the bug

I recently tried to package an application that was build with tauri, but in my build sandbox I encountered the following error:

```
    Info `tauri-build` dependency has workspace inheritance enabled. The features array won't be automatically rewritten. Expected features: []
    Info `tauri` dependency has workspace inheritance enabled. The features array won't be automatically rewritten. Expected features: []
error: no matching package found
searched package name: `hyper-tls`
perhaps you meant:      hyper-rustls or hyper-util
location searched: registry `crates-io`
required by package `reqwest v0.12.8`
    ... which satisfies dependency `reqwest = "^0.12.5"` (locked to 0.12.8) of package `firezone-gui-client-common v1.4.0 (/build/source/rust/gui-client/src-common)`
    ... which satisfies path dependency `firezone-gui-client-common` (locked to 1.4.0) of package `firezone-gui-client v1.4.0 (/build/source/rust/gui-client/src-tauri)`
As a reminder, you're using offline mode (--offline) which can sometimes cause surprising resolution failures, if this error is too confusing you may wish to retry without the offline flag.
    Error failed to build app:
```

For context, I pre-download all dependencies using `cargo vendor` and later build without network access, but for some reason that is unknown to me, `tauri build` seems to execute a cargo build that requires more than what I have in my lock file. Specifically, it will try to fetch `hyper-tls` which then fails. The obvious and simplest solution for me was to add `hyper-tls` to the project dependencies (either directly or by enabling tauri's `native-tls` feature), even though the project doesn't actually depend on it. This just makes sure the dependency is available in the vendored directory so `tauri build` can succeed.

I also cannot figure out how to remove this dependency, say if I wanted to use `rustls` instead hyper-tls which seems to pull in openssl. Unfortunately, I don't know enough about tauri to find the root cause for this. It does however seem like something that shouldn't be happening.

### Reproduction

```bash
$ git clone https://github.com/firezone/firezone
$ git checkout 8c9427b7b133e5050be34c2ac0e831c12c08f02c
$ cd firezone/rust
$ cargo vendor          # follow instructions by cargo vendor and copy stuff to .cargo/config.toml
$ mkdir gui-client/dist # just a dummy, the error will happen way before we get to frontend stuff
$ cd gui-client
$ cargo tauri build -- --offline

    Info `tauri-build` dependency has workspace inheritance enabled. The features array won't be automatically rewritten. Expected features: []
    Info `tauri` dependency has workspace inheritance enabled. The features array won't be automatically rewritten. Expected features: []
error: no matching package found
searched package name: `hyper-tls`
perhaps you meant:      hyper-rustls or hyper-util
location searched: registry `crates-io`
required by package `reqwest v0.12.9`
    ... which satisfies dependency `reqwest = "^0.12.9"` (locked to 0.12.9) of package `firezone-gui-client-common v1.4.1 (/home/malte/firezone/rust/gui-client/src-common)`
    ... which satisfies path dependency `firezone-gui-client-common` (locked to 1.4.1) of package `firezone-gui-client v1.4.1 (/home/malte/firezone/rust/gui-client/src-tauri)`
As a reminder, you're using offline mode (--offline) which can sometimes cause surprising resolution failures, if this error is too confusing you may wish to retry without the offline flag.
    Error failed to build app:
```

### Expected behavior

All dependencies should be declared.

### Full `tauri info` output

```text
[-] Packages
    - tauri 🦀: 2.2.2
    - tauri-build 🦀: 2.0.5
    - wry 🦀: 0.48.0
    - tao 🦀: 0.31.0
    - tauri-cli 🦀: 2.1.0
    - @tauri-apps/cli : 2.1.0

[-] Plugins
    - tauri-plugin-notification 🦀: 2.2.0
    - tauri-plugin-shell 🦀: 2.2.0
    - @tauri-apps/plugin-shell : not installed!
    - tauri-plugin-fs 🦀: 2.2.0

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - bundler: Vite
```

### Stack trace

```text
N/A
```

### Additional context

_No response_

FabianLars:
this caused issues 2 or 3 times on discord now and since we use reqwest only on mobile it really doesn't make sense, especially considering that the http plugin etc all set rustls as default.

i also changed the deps in tauri to declare reqwest as a mobile only dependency.

Edit: fixes #12448

### DIFF
--- a/crates/tauri/Cargo.toml
+++ b/crates/tauri/Cargo.toml
@@ -68,11 +68,6 @@ serde_repr = "0.1"
 http = "1"
 dirs = "5"
 percent-encoding = "2"
-reqwest = { version = "0.12", default-features = false, features = [
-  "json",
-  "stream",
-] }
-bytes = { version = "1", features = ["serde"] }
 raw-window-handle = { version = "0.6", features = ["std"] }
 glob = "0.3"
 urlpattern = "0.3"
@@ -89,13 +84,16 @@ specta = { version = "^2.0.0-rc.16", optional = true, default-features = false, 
   "function",
   "derive",
 ] }
-[target."cfg(any(target_os = \"linux\", target_os = \"dragonfly\", target_os = \"freebsd\", target_os = \"openbsd\", target_os = \"netbsd\", target_os = \"windows\", target_os = \"macos\"))".dependencies]
+
+# desktop
+[target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd", target_os = "windows", target_os = "macos"))'.dependencies]
 muda = { version = "0.15", default-features = false, features = ["serde"] }
 tray-icon = { version = "0.19", default-features = false, features = [
   "serde",
 ], optional = true }
 
-[target."cfg(any(target_os = \"linux\", target_os = \"dragonfly\", target_os = \"freebsd\", target_os = \"openbsd\", target_os = \"netbsd\"))".dependencies]
+# linux
+[target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd"))'.dependencies]
 gtk = { version = "0.18", features = ["v3_24"] }
 webkit2gtk = { version = "=2.0.1", features = ["v2_40"] }
 
@@ -115,15 +113,22 @@ objc2-app-kit = { version = "0.2", features = [
 ] }
 window-vibrancy = "0.5"
 
+# windows
 [target."cfg(windows)".dependencies]
 webview2-com = "0.34"
 window-vibrancy = "0.5"
+windows = { version = "0.58", features = ["Win32_Foundation"] }
 
-[target."cfg(windows)".dependencies.windows]
-version = "0.58"
-features = ["Win32_Foundation"]
+# mobile
+[target.'cfg(any(target_os = "android", all(target_vendor = "apple", not(target_os = "macos"))))'.dependencies]
+bytes = { version = "1", features = ["serde"] }
+reqwest = { version = "0.12", default-features = false, features = [
+  "json",
+  "stream",
+] }
 
-[target."cfg(target_os = \"android\")".dependencies]
+# android
+[target.'cfg(target_os = "android")'.dependencies]
 jni = "0.21"
 
 # UIKit, i.e. iOS/tvOS/watchOS/visionOS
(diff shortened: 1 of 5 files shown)


FabianLars:
actually since we forcefully enable rustls-tls instead of checking it li

FabianLars:
actually, since we forcefully enable rustls-tls on mobile instead of checking it first like on desktop we could just as well "hardcode" it in the reqwest dependency here, no?

FabianLars:
after a tiny bit of research it seems like native-tls on mobile is nothing but a headache and rarely works so i just so i just enabled the feature permanently which effectively doesn't even change anything except the timing of when the feature is enabled.

i also have a few devs on this branch already with no issues reported so far. I'd really like to get this into the patch release i wanted to do this week tbh 😅

thomaseizinger:
> except the timing of when the feature is enabled.

But it is still removed on Desktop, right?

FabianLars:
yes and no. It should show up in your lockfile, but not like on build time as it did before(?) but when you add tauri to cargo.toml.
It will only actually be used when building for mobile though.

Is that a problem for your use case?

FabianLars:
Ahh i just noticed that i phrased my earlier comment very poorly. The native-tls flag is still completely gone, it's the rustls-tls flag that's always enabled now (but only used on mobile)

thomaseizinger:
The issue we have is that in order to fully support offline builds, we need to add the `native-tls` feature flag which brings in OpenSSL, despite nothing in our stack using OpenSSL (on purpose).

FabianLars:
okay so if nothing changed from what i've heard in https://github.com/firezone/firezone/issues/7799 & https://github.com/tauri-apps/tauri/issues/12448 the last changes will still work for you without needing the native-tls flag, but it will now behave as if you had added the rustls-tls feature flag if tauri-cli's feature detection wasn't broken. Compared to no feature flag like it kinda behaved in my first commit.

thomaseizinger:
Okay that makes sense, thank you! We already depend on `rustls-tls` so this is fine.

crates/tauri/Cargo.toml
```diff
@@ -179,9 +185,11 @@ objc-exception = ["tauri-runtime-wry/objc-exception"]
 linux-libxdo = ["tray-icon/libxdo", "muda/libxdo"]
 isolation = ["tauri-utils/isolation", "tauri-macros/isolation", "uuid"]
 custom-protocol = ["tauri-macros/custom-protocol"]
+# TODO: Remove these flags in v3 and/or enable them by default behind a mobile flag https://github.com/tauri-apps/tauri/issues/12384
```
amrbashir, line 188: let's add these to v3 project board

amrbashir (approved):

crates/tauri/Cargo.toml
FabianLars: TIL that we have a v3 project board

PR #12460, opened 2025-01-21
feat(tauri/Emitter): add `emit_str*` method to emit serialized data

### LINKED ISSUE #12459: [feat] Emit json string
### Describe the problem

I am implementing `pytauri` Python bindings for the [Emitter](https://docs.rs/tauri/latest/tauri/trait.Emitter.html) trait, but currently, the `Emitter` API is blocking me. Specifically:

In pytauri, the way we handle IPC between Python and the frontend is:

1. On the Python side, we serialize the PyObject to JSON bytes using [pydantic](https://github.com/pydantic/pydantic).
2. Use the IPC API provided by Tauri to send the bytes to the frontend.
3. [On the frontend, use `TextEncoder` and `JSON.parse` to parse the bytes into a JSON object](https://github.com/WSH032/pytauri/blob/d19b13e6f5f71eef1a7692b7ebd27e2daefc6e38/crates/tauri-plugin-pytauri/guest-js/src/index.ts#L37-L54).

I have implemented `Channel` and `Command` IPC using this method.

However, for `Emitter`, its `emit*` methods require `payload: Serialize`, and I cannot implement `Serialize` for a [PyObject](https://docs.rs/pyo3/latest/pyo3/type.PyObject.html) without affecting performance.

> Implementing `Serialize` for `PyObject` for serialization is far slower than using `pydantic` for serialization.

---

I noticed that `Emitter` just serializes the `payload` to JSON behind the scenes: <https://github.com/tauri-apps/tauri/blob/b8eb28877fe822dbe17999fc8af98ed7d0983679/crates/tauri/src/event/mod.rs#L134-L144>.

So, would Tauri accept adding the following method to `Emitter`:

```rust
fn emit_json(&self, event: &str, payload: String) -> Result<()>;
```

This way, I can use pydantic for serialization on the Python side without incurring the [high checking cost](https://github.com/davidhewitt/pythonize/blob/4491cdb46f35ab3fc1b419beef92171fbef510da/src/de.rs#L111-L150).

Add `emit_json_*` series methods

### Alternatives considered

None

### Additional context

This issue is blocking the release of pytauri version 0.2. If Tauri can accept this feature request, I am willing to submit a PR.

WSH032:
close #12459

### DIFF
--- a/crates/tauri/src/event/mod.rs
+++ b/crates/tauri/src/event/mod.rs
@@ -128,6 +128,15 @@ impl EmitArgs {
       payload: serde_json::to_string(payload)?,
     })
   }
+
+  pub fn new_str(event: EventName<&str>, payload: String) -> crate::Result<Self> {
+    #[cfg(feature = "tracing")]
+    let _span = tracing::debug_span!("window::emit::json").entered();
+    Ok(EmitArgs {
+      event: event.into_owned(),
+      payload,
+    })
+  }
 }
 
 /// An event that was emitted.
--- a/.changes/emit-str.md
+++ b/.changes/emit-str.md
@@ -1,0 +1,5 @@
+---
+tauri: 'minor:feat'
+---
+
+Add `emit_str*` methods to `Emitter` trait to allow emitting JSON serialized data directly.
--- a/crates/tauri/src/lib.rs
+++ b/crates/tauri/src/lib.rs
@@ -210,6 +210,7 @@ pub use self::utils::TitleBarStyle;
 
 use self::event::EventName;
 pub use self::event::{Event, EventId, EventTarget};
+use self::manager::EmitPayload;
 pub use {
   self::app::{
     App, AppHandle, AssetResolver, Builder, CloseRequestApi, RunEvent, UriSchemeContext,
@@ -940,7 +941,15 @@ pub trait Emitter<R: Runtime>: sealed::ManagerBase<R> {
   /// ```
   fn emit<S: Serialize + Clone>(&self, event: &str, payload: S) -> Result<()> {
     let event = EventName::new(event)?;
-    self.manager().emit(event, &payload)
+    let payload = EmitPayload::Serialize(&payload);
+    self.manager().emit(event, payload)
+  }
+
+  /// Similar to [`Emitter::emit`] but the payload is json serialized.
+  fn emit_str(&self, event: &str, payload: String) -> Result<()> {
+    let event = EventName::new(event)?;
+    let payload = EmitPayload::<()>::Str(payload);
+    self.manager().emit(event, payload)
   }
 
   /// Emits an event to all [targets](EventTarget) matching the given target.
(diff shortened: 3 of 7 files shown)


amrbashir (commented):
this feels more like `emit_str` than `emit_json`, let's rename

crates/tauri/src/event/mod.rs
```diff
@@ -141,6 +141,16 @@ impl EmitArgs {
       payload: serde_json::to_string(&payload)?,
     })
   }
+
+  pub fn new_json(event: &str, payload: String) -> crate::Result<Self> {
+    #[cfg(feature = "tracing")]
+    let _span = tracing::debug_span!("window::emit::json").entered();
+    Ok(EmitArgs {
+      event_name: event.into(),
+      event: serde_json::to_string(event)?,
```
amrbashir: let's remove this serde call as well, it is unnecessary, also do the same for `EmitArgs::new`.

Don't forget to update `emit_js_script` function later down in this file to add wrapping single quotes around event name.

AzharHussain366:
Bb. Bb

crates/tauri/src/event/mod.rs
WSH032: > let's remove this serde call as well, it is unnecessary, also do the same for EmitArgs::new.

I don't quite understand, do you mean `event: event.into()`? If so, wouldn't `event_name` and `event` be the same?

> Don't forget to update emit_js_script function later down in this file to add wrapping single quotes around event name.

Wouldn't this cause issues with escaping `"` and `'`?

amrbashir: it shouldn't be an issue because event names are validated to contain alphanumeric, `-`, `/`, `:` and `_`

amrbashir: we can remove the `event_name` as well, seems uneeded.

WSH032: Done. I call it a day. If there are any other changes needed, I will handle them tomorrow; or feel free to push to this PR.

amrbashir: Thank you, I will wait on https://github.com/tauri-apps/tauri/pull/12444 first before we can move forward with this PR as it touches the same code

amrbashir:
#12444 is now merged, we can move on with this PR once you fix the conflicts

WSH032:
will do it now

WSH032:
@amrbashir. Sorry for pinging you, could you please review this PR? I noticed that `tauri v2.3` #12518 is about to be released, and if this PR can make it into `v2.3`, `pytauri v0.2` will also benefit.

It's okay if you don't have time right now. I can postpone this feature to `pytauri v0.3`.

FabianLars:
No worries, 2.3 still needs a few days/weeks. I just started merging minor PRs for it. I also asked Amr to not merge yours until now (because I wanted to release patches first) so this is on me :)

amrbashir (approved):

PR #12461, opened 2025-01-21
fix(tauri): `Webview::navigate` unnecessarily borrows `&mut self`

### LINKED ISSUE #12430: [feat] `Webview::navigate` borrows `&self` instead of `&mut self`
### Describe the problem

<https://github.com/tauri-apps/tauri/blob/72748cc45cf670dd03c86c8deceb5942598f5ad9/crates/tauri/src/webview/mod.rs#L1266-L1269>

Currently, `Webview::navigate` borrows `&mut self`, but its underlying implementation [tauri_runtime::WebviewDispatch::navigate](https://docs.rs/tauri-runtime/2.3.0/tauri_runtime/trait.WebviewDispatch.html#tymethod.navigate) only requires an immutable borrow `&self`.

Is there any specific reason why `tauri::Webview` needs to do this? If not, can we change it to only require an immutable borrow `&self`?

---

## Why don't I want `&mut self`?

I am implementing Python bindings for `Webview::navigate` in pytauri: WSH032/pytauri#46.

**Python is multithreaded, and when implementing pyo3 bindings for a Rust struct, if at least one method requires `&mut self`, we have to add a thread lock, which degrades performance.**

As of now, except for `navigate`, other methods of `Webview` only require `&self`, and I don't want to add a thread lock just for only one method.

just change [WebviewWindow::navigate](https://docs.rs/tauri/latest/x86_64-pc-windows-msvc/tauri/webview/struct.WebviewWindow.html#method.navigate) and [Webview::navigate](https://docs.rs/tauri/latest/x86_64-pc-windows-msvc/tauri/webview/struct.Webview.html#method.navigate) to `&self`.

### Alternatives considered

None

### Additional context

[pyo3/Bound and interior mutability](https://pyo3.rs/v0.23.4/class.html#bound-and-interior-mutability)

WSH032:
close #12430

---

Sorry for opening this PR without permission.

#12430 is blocking me from implementing the [Python bindings for `Webview::navigate`](https://github.com/WSH032/pytauri/blob/d19b13e6f5f71eef1a7692b7ebd27e2daefc6e38/crates/pytauri-core/src/ext_mod_impl/webview.rs#L214-L218). I plan to release `pytauri v0.2` soon, so I hope `pytauri` can catch up with this PR.

> Is there any specific reason why tauri::Webview needs to do this? If not, can we change it to only require an immutable borrow &self?

Not sure if my thoughts are correct, feel free to close this PR if I am wrong.

### DIFF
--- a/.changes/webview-navigate-borrow-self.md
+++ b/.changes/webview-navigate-borrow-self.md
@@ -1,5 +1,5 @@
 ---
-'tauri': 'patch:bug'
+'tauri': 'minor:bug'
 ---
 
 `Webview::navigate` and `WebviewWindow::navigate` borrows `&self` instead of unnecessarily borrowing `&mut self`.
--- a/.changes/webview-navigate-borrow-self.md
+++ b/.changes/webview-navigate-borrow-self.md
@@ -1,0 +1,5 @@
+---
+'tauri': 'patch:bug'
+---
+
+`Webview::navigate` and `WebviewWindow::navigate` borrows `&self` instead of unnecessarily borrowing `&mut self`.
--- a/crates/tauri/src/webview/mod.rs
+++ b/crates/tauri/src/webview/mod.rs
@@ -1264,7 +1264,7 @@ fn main() {
   }
 
   /// Navigates the webview to the defined url.
-  pub fn navigate(&mut self, url: Url) -> crate::Result<()> {
+  pub fn navigate(&self, url: Url) -> crate::Result<()> {
     self.webview.dispatcher.navigate(url).map_err(Into::into)
   }
 
--- a/crates/tauri/src/webview/webview_window.rs
+++ b/crates/tauri/src/webview/webview_window.rs
@@ -1864,7 +1864,7 @@ impl<R: Runtime> WebviewWindow<R> {
   }
 
   /// Navigates the webview to the defined url.
-  pub fn navigate(&mut self, url: Url) -> crate::Result<()> {
+  pub fn navigate(&self, url: Url) -> crate::Result<()> {
     self.webview.navigate(url)
   }
 


sftse:
~~This would break the public API and need a major version update to respect semver.~~

.changes/webview-navigate-borrow-self.md
```diff
@@ -0,0 +1,5 @@
+---
+'tauri': 'patch:bug'
```
sftse:
```suggestion
'tauri': 'major:enhance'
```

WSH032: Why is this a breaking change? Downgrading from `&mut self` to `&self` should only trigger a clippy warning, right?

sftse: It seems you're right. Thanks.

FabianLars:
Does this really need a major version? What would a user's code look like that works with &mut self but not &self ? I can't think of anything 🤔 From my testing this doesn't even emit check/clippy warnings which would be totally fine.

Edit: In comparison switching from `&self` to `&mut self` would indeed be a clear breaking change.

In context of https://github.com/tauri-apps/tauri/issues/12465 and this:
{1D2E50D4-C451-4717-9215-27E28A7571F8}
i wouldn't mind a minor instead of patch though.

WSH032:
> i wouldn't mind a minor instead of patch though.

TBH, if strictly following semver, I agree more with releasing it as a `minor`.

I chose `patch` just for selfish reasons, as I worry that the `tauri v2.3` release cycle might be too long, causing `pytauri v0.2` to miss this PR.

Anyway, whether `tauri` chooses `patch` or `minor`, I am okay with it. Feel free to push to this PR.

FabianLars:
> I chose patch just for selfish reasons, as I worry that the tauri v2.3 release cycle might be too long, causing pytauri v0.2 to miss this PR.

I plan to release 2.3 for the new webview2-com and windows version relatively soon - (and yes, that's actually a breaking change but we can't have it locked for the whole lifetime of v2). I just want to get a few last patches in before that.

sftse:
I withdrew my objection because I tested the cases I had in mind and they turned out to be bogus, do you have an example how this is indeed a breaking change?

```rust
fn foo<T>(_: &T) {}
fn bar<T>(t: &mut T) {
    foo(t)
}
```
Edit: Seems to compile, which I've been reading as proof that this is backwards compatible.

FabianLars:
> do you have an example how this is indeed a breaking change?

as i said, no i don't :D I just can't think of any breakage for `&mut self -> &self` because you can call &self methods on mutable objects.

`&mut self -> &self` however is pretty obvious as you need `let mut` instead of `let` for `&mut self` to work.

FabianLars (approved):

PR #12485, opened 2025-01-23
Clippy

sftse:


FabianLars:
Thanks for your PR! Before we can merge it we need _all_ commits to be signed https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

Issue #11994, opened 2024-12-17
[bug] Tauri fails to build on Fedora 41 (Failed to get GBM device)


SpeeQz1:
### Describe the bug

When trying to run a default template app made with the `npm create tauri-app@latest` command using React on Fedora 41, the window appears to have an empty screen with the errors `Failed to get GBM device`.

`npm run tauri dev` log:

![image](https://github.com/user-attachments/assets/d8653a2e-c57b-4431-a035-72bbd134c711)

```sh
> tauri-app@0.1.0 tauri
> tauri dev

    Running BeforeDevCommand (`npm run dev`)

> tauri-app@0.1.0 dev
> vite


  VITE v6.0.3  ready in 176 ms

  ➜  Local:   http://localhost:1420/
    Info Watching /home/user/Downloads/tauri-app/src-tauri for changes...
    Updating crates.io index
  Downloaded foreign-types-shared v0.1.1
  Downloaded foreign-types v0.3.2
  Downloaded openssl-macros v0.1.1
  Downloaded openssl-probe v0.1.5
  Downloaded hyper-tls v0.6.0
  Downloaded tokio-native-tls v0.3.1
  Downloaded rustls-pemfile v2.2.0
  Downloaded native-tls v0.2.12
  Downloaded openssl-sys v0.9.104
  Downloaded rustls-pki-types v1.10.1
  Downloaded openssl v0.10.68
  Downloaded vcpkg v0.2.15
  Downloaded 12 crates (753.6 KB) in 0.34s
   Compiling equivalent v1.0.1
   Compiling hashbrown v0.15.2
   Compiling winnow v0.5.40
   Compiling indexmap v2.7.0
   Compiling toml_datetime v0.6.3
   Compiling serde_spanned v0.6.8
   Compiling cfg-if v1.0.0
   Compiling heck v0.5.0
   Compiling toml_edit v0.20.2
   Compiling serde v1.0.216
   Compiling toml v0.8.2
   Compiling libc v0.2.168
   Compiling system-deps v6.2.2
   Compiling once_cell v1.20.2
   Compiling memchr v2.7.4
   Compiling byteorder v1.5.0
   Compiling zerocopy v0.7.35
   Compiling ppv-lite86 v0.2.20
   Compiling siphasher v0.3.11
   Compiling pin-project-lite v0.2.15
   Compiling stable_deref_trait v1.2.0
   Compiling futures-core v0.3.31
   Compiling zerofrom v0.1.5
[...]

PR #12491, opened 2025-01-24
feat(bundler): Truly portable appimage (experimental)

### LINKED ISSUE #6172: [bug] Links with target=_blank and shell::open do not work from AppImage on Linux
### Describe the bug

When running a Tauri app on Linux using an AppImage, target=_blank links do not work, and shell::open does not work either.

### Reproduction

Minimal reproduction repo: https://github.com/0rvar/tauri-appimage-repro

Create a link with target=_blank, or use shell::open with an URL. Then run the tauri app as an AppImage. It will silently fail to open the link when clicked, and silently fail to open the browser with the URL from shell::open.

### Expected behavior

The browser should open with the URL

### Platform and versions

Linux, AppImage. See reproduction repo above

### Stack trace

_No response_

### Additional context

_No response_

### LINKED ISSUE #10078: [bug] tauri.shell.open does not working on differtent linux distribution than building platform with appimage build
### Describe the bug

One of user told me that buttons opens brower does not works on Arch Linux with AppImage built on GitHub Actions.

After small investigation, open function become not working if running on other distributions than the AppImage was built.

### Reproduction

Downloading https://github.com/anatawa12/tauri-link-appimage-test/releases/tag/app-v0.1.0 and run on Arch Linux will reproduce this problem and downloading https://github.com/vrc-get/vrc-get/issues/934#issuecomment-2164885456 and run on Ubuntu 24.04 will also reproduce this problem

### Expected behavior

It should works 

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 14.4.1 X64
    ✔ rustc: 1.79.0 (129f3b996 2024-06-10)
    ✔ cargo: 1.79.0 (ffa9cf99a 2024-06-03)
    - node: 20.12.2
    - pnpm: 8.15.4
    - yarn: 1.22.19
    - npm: 10.8.1

[-] Packages
    - tauri [RUST]: 1.6.8
    - tauri-build [RUST]: 1.5.2
    - wry [RUST]: 0.24.10
    - tao [RUST]: 0.16.9
    - tauri-cli [RUST]: 1.5.14
    - @tauri-apps/api [NPM]: 1.5.6
    - @tauri-apps/cli [NPM]: 1.5.14

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: out
    - framework: React (Next.js)
    - bundler: Webpack
```

### Stack trace

_No response_

### Additional context

Downstream issue: vrc-get/vrc-get#934

### LINKED ISSUE #10617: [bug] xdg-open does not work in release artifacts produced by tauri-action
Hey @FabianLars

3 different, but related, functions of my application do not work when using the AppImage of my application produced by tauri-action in GitHub Actions, but work in AppImages produced locally for myself and 3 other users (as well as other methods of running the application, such as `cargo tauri dev` or other packaging formats). On my system, at least, the .RPM produced by tauri-action seems to work, and it is only the .AppImage that does not work. This issue is possibly related to tauri-apps/tauri#6172.

1. My application [OpenDeck](https://github.com/ninjadev64/OpenDeck) loads plugins, which then communicate by sending JSON events back and forth over a websocket. One such event is `openUrl`, which a plugin can send to open a URL in the user's browser, which OpenDeck handles using Tauri's shell::open (see [here](https://github.com/ninjadev64/OpenDeck/blob/4bbd9f3fc17e396d56728af4571dfdb2dc6d57e5/src-tauri/src/events/inbound/misc.rs#L19)).

2. OpenDeck also has a button to open its config directory in its settings tab, which, on Linux, [uses xdg-open using std's `Command`](https://github.com/ninjadev64/OpenDeck/blob/4bbd9f3fc17e396d56728af4571dfdb2dc6d57e5/src-tauri/src/events/frontend.rs#L469) inside a Tauri command.

3. Additionally, OpenDeck's "starter pack" plugin has a "Run Command" action, in which users can input a command to run on key down and key up. The starter pack plugin is also written in Rust, and [uses std's `Command`](https://github.com/ninjadev64/opendeck-starterpack/blob/b9bed2922f30d511266ee5fa38c3e8c6b8e829a0/src/run_command.rs#L9) to implement this action.

For example, when pressing "open config directory" when using the AppImage from GitHub Actions, run with `./opendeck_2.0.0_amd64.AppImage`, this is printed to the terminal that the AppImage was run from:
```
xdg-mime: no method available for querying MIME type of '/home/aman/.config/com.amansprojects.opendeck'
xdg-mime: mimetype argument missing
Try 'xdg-mime --help' for more information.
/tmp/.mount_opendegOqUUe/usr/bin//xdg-open: line 882: x-www-browser: command not found
XPCOMGlueLoad error for file /usr/lib64/firefox/libxul.so:
/usr/lib64/libproxy/libpxbackend-1.0.so: undefined symbol: g_once_init_leave_pointer
Couldn't load XPCOM.
/tmp/.mount_opendegOqUUe/usr/bin//xdg-open: line 882: iceweasel: command not found
/tmp/.mount_opendegOqUUe/usr/bin//xdg-open: line 882: seamonkey: command not found
/tmp/.mount_opendegOqUUe/usr/bin//xdg-open: line 882: mozilla: command not found
/tmp/.mount_opendegOqUUe/usr/bin//xdg-open: line 882: epiphany: command not found
/tmp/.mount_opendegOqUUe/usr/bin//xdg-open: line 882: konqueror: command not found
/tmp/.mount_opendegOqUUe/usr/bin//xdg-open: line 882: chromium: command not found
/tmp/.mount_opendegOqUUe/usr/bin//xdg-open: line 882: chromium-browser: command not found
/tmp/.mount_opendegOqUUe/usr/bin//xdg-open: line 882: google-chrome: command not found
/tmp/.mount_opendegOqUUe/usr/bin//xdg-open: line 882: www-browser: command not found
/tmp/.mount_opendegOqUUe/usr/bin//xdg-open: line 882: links2: command not found
/tmp/.mount_opendegOqUUe/usr/bin//xdg-open: line 882: elinks: command not found
/tmp/.mount_opendegOqUUe/usr/bin//xdg-open: line 882: links: command not found
/tmp/.mount_opendegOqUUe/usr/bin//xdg-open: line 882: lynx: command not found
```
and then it seems to finally find and open w3m on my system in the terminal in the correct directory, whereas it should open in Nautilus for me, as it does when running OpenDeck from different methods.

Or, when using the Run Command action with `xdg-open https://example.com/`, this is the stdout of the executed command:
```
XPCOMGlueLoad error for file /usr/lib64/firefox/libxul.so:
/usr/lib64/libproxy/libpxbackend-1.0.so: undefined symbol: g_once_init_leave_pointer
Couldn't load XPCOM.
```

OpenDeck is using the latest Tauri v1.7 and `uses: tauri-apps/tauri-action@v0`. I would appreciate some steps to follow to resolve this issue.

### LINKED ISSUE #14208: [bug] [linuxdeploy] Linux AppImage script fails [podman] ['std::logic_error']
### Describe the bug

Looking to offer a local build option for our tauri app for our Linux users, we ran into a number of user related issues like package versions mis-matching issues, none of these problems with the tauri-app, just a reason why I was looking at doing a docker build.

Tried to make this build process easier by documenting and testing how to use docker as a clean starting point, I ran into problems, because I had switched to using podman for my container loads.

Doing an AppImage build I got to the following failed bundle position, which I was not able to find any info on or for
```
    Bundling [tauri_bundler::bundle::linux::appimage] Tari Universe (Alpha)_1.6.1_amd64.AppImage (/work/src-tauri/target/release/bundle/appimage/Tari Universe (Alpha)_1.6.1_amd64.AppImage)
     Running [tauri_bundler::utils] Command `/root/.cache/tauri/linuxdeploy-x86_64.AppImage  --appimage-extract-and-run --verbosity 1 --appdir /work/src-tauri/target/release/bundle/appimage/Tari Universe (Alpha).AppDir --plugin gtk --plugin gstreamer --output appimage`
linuxdeploy version 1-alpha (git commit ID 659c9db), GitHub actions build 10 built on 2024-07-26 15:40:22 UTC
terminate called after throwing an instance of 'std::logic_error'
  what():  subprocess failed (exit code 127)
       Error [tauri_cli] failed to bundle project:
    - `failed to run /root/.cache/tauri/linuxdeploy-x86_64.AppImage`
```
As the basic build and the rpm/deb bundles were completing without any noticeable issues, I did not think I had anything wrong with our process. Plus I was able to build using GitHub action ```tauri-apps/tauri-action``` in our workflow.

Seems that ```FUSE``` storage options with ```podman``` need a little more effort, as ```podman``` is a little stricter. Some info can be found reading the ReadHat Blog - https://www.redhat.com/en/blog/podman-inside-container, thou you would not find the above error ```'std::logic_error'```
  
Running a container to build tauri-app in
```bash
podman run -it --rm  \
  --cap-add=sys_admin --cap-add mknod --device=/dev/fuse \
  -v ${PWD}/../temp/root:/root \
  -v ${PWD}/../universe:/work \
  -w /work \
  -u root \
  --platform linux/amd64 \
  ubuntu:22.04 bash
```
with ```--cap-add=sys_admin --cap-add mknod --device=/dev/fuse``` been the part needed to fix the above error.

This does not include all the package install and setups. This was for x86_64 host and guest container.

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
root@f7e7c7bfc16c:/work# npm run tauri info

> tari-universe@1.6.1 tauri
> node ./scripts/tauri-wrapper.js info

Loading environment file: ./src-tauri/env.esme
[dotenv@17.2.2] injecting env (8) from src-tauri/env.esme -- tip: 📡 auto-backup env with Radar: https://dotenvx.com/radar

[✔] Environment
    - OS: Ubuntu 22.4.0 x86_64 (X64) (Unknown DE on Unknown Session)
    ✔ webkit2gtk-4.1: 2.48.5
    ✔ rsvg2: 2.52.5
    - node: 22.19.0
    - npm: 10.9.3

[-] Packages
    - tauri 🦀: 2.8.5
    - tauri-build 🦀: 2.4.1
    - wry 🦀: 0.53.3
    - tao 🦀: 0.34.3
    - tauri-cli 🦀: 2.8.4
    - @tauri-apps/api : 2.8.0
    - @tauri-apps/cli : 2.8.4

[-] Plugins
    - tauri-plugin-process 🦀: 2.3.0
    - @tauri-apps/plugin-process : not installed!
    - tauri-plugin-fs 🦀: 2.4.2
    - tauri-plugin-shell 🦀: 2.3.1
    - tauri-plugin-single-instance 🦀: 2.3.4
    - tauri-plugin-http 🦀: 2.5.2
    - @tauri-apps/plugin-http : not installed!
    - tauri-plugin-clipboard-manager 🦀: 2.3.0
    - @tauri-apps/plugin-clipboard-manager : 2.3.0
    - tauri-plugin-os 🦀: 2.3.1
    - tauri-plugin-cli 🦀: 2.4.0
    - @tauri-apps/plugin-cli : not installed!

[-] App
    - build-type: bundle
    - CSP: style-src 'self' 'unsafe-inline'; object-src 'self' data: blob:; img-src 'self' data: blob: base64: https://*.twimg.com https://cdn.npmjs.com/@tari-project/tari-tower http://127.0.0.1 http://127.0.0.1/ http://127.0.0.1:* http://localhost http://localhost/ http://localhost:* https://static.tari.com https://static.tari.com/* https://lh3.googleusercontent.com; worker-src 'self' blob: https://cdn.npmjs.com/@tari-project/tari-tower; default-src 'self' tauri: https: http://localhost:* http://127.0.0.1:*; script-src-elem 'self' 'unsafe-inline' 'unsafe-eval'; connect-src 'self' https: wss://ut.tari.com wss://relay.walletconnect.org https://rpc.ankr.com/eth https://ethereum-sepolia-rpc.publicnode.com tauri: tauri://localhost:* http://localhost:* http://127.0.0.1:* ipc: http://ipc.localhost data: blob: application/octet-stream base64; script-src 'self' 'unsafe-eval'; media-src 'self' data: blob: tauri: https:
    - frontendDist: ../dist
    - framework: React
    - bundler: Vite
```

### Stack trace

```text

```

### Additional context

_No response_

FabianLars:
will add description later

~~Edit1: issue with bun's single exe output https://github.com/oven-sh/bun/issues/28281~~ fixed with a hack: https://github.com/tauri-apps/tauri/pull/12491#issuecomment-4785971505

fixes https://github.com/tauri-apps/tauri/issues/10617 (xdg-open)
fixes https://github.com/tauri-apps/tauri/issues/10078 (xdg-open)
fixes https://github.com/tauri-apps/tauri/issues/6172 (xdg-open)
fixes https://github.com/tauri-apps/tauri/issues/14208 (NO_STRIP)


crates/tauri-bundler/src/bundle/linux/appimage/experimental.rs
```diff
@@ -0,0 +1,216 @@
...
+  // TODO(--NOW--): make it bundle only 4.0 (v1) / 4.1 (v2) / 6.0 (v3)
+  // TODO(--NOW--): Check if we can make parts of the opengl deps optional
+  // TODO(later): rustify this (finding the paths in rust instead of using bash glob patterns)
+  // TODO(later): Consider putting gst and pulse behind bundleMediaFramwork for ~150mb -> ~130mb
+  Command::new("/bin/sh")
+    .current_dir(&app_dir_path)
+    .args([
+      "-c",
+      &format!(
+        "{} -p {} -r {} -k {} /usr/lib/x86_64-linux-gnu/libGL* /usr/lib/x86_64-linux-gnu/libEGL* /usr/lib/x86_64-linux-gnu/libvulkan* /usr/lib/x86_64-linux-gnu/dri/* /usr/lib/x86_64-linux-gnu/libpulsecommon* /usr/lib/x86_64-linux-gnu/libnss_mdns* /usr/lib/x86_64-linux-gnu/gstreamer-1.0/* /usr/lib/x86_64-linux-gnu/libwebkit2gtk*",
```
Samueru-sama: Note that I'm not sure if all tauri apps end up using `libvulkan` and `gstreamer`.

[Also another user told me that they needed to include gio](https://github.com/gitbutlerapp/gitbutler/issues/5282#issuecomment-2605465483)

FabianLars: Yeah, gstreamer will be optional like in our old implementation.

The libvulkan thing is a good point though iirc i just counted it as part of the opengl todo. Either way, will keep it in mind, thanks.

> https://github.com/gitbutlerapp/gitbutler/issues/5282#issuecomment-2605465483

Ah shit, right. Totally forgot about this even though we still have some issues with it in our appimage impl. 🙃

crates/tauri-bundler/src/bundle/linux/appimage/experimental.rs
```diff
@@ -0,0 +1,257 @@
...
+
+  // TODO(--NOW--): Check if we can make parts of the opengl (incl. libvulkan) deps optional
+  // TODO(later): rustify this (finding the paths in rust instead of using bash glob patterns)
+  // TODO(later): Consider putting gst and pulse behind bundleMediaFramwork for ~150mb -> ~130mb
+  Command::new("/bin/sh")
+    .current_dir(&app_dir_path)
+    .args([
+      "-c",
+      &format!(
+        r#"{} -p {} -r {} -k {} \
```
Samueru-sama: Btw the `-r` is not needed anymore with `lib4bin` and should not be used as we found it breaks libraries.

FabianLars: coolio will remove it, thx

crates/tauri-bundler/src/bundle/linux/appimage/experimental.rs
```diff
@@ -151,7 +151,7 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
     .args([
       "-c",
       &format!(
-        r#"{} -p {} -r {} -k {} \
+        r#"{} -p {} {} -k {} \
```
Samueru-sama: Doesn't the `{}` also need to be removed? 👀 I'm don't know rust though.

FabianLars: that's for the strip flag (wanted to support a `NO_STRIP` env var in our bundler). i was confused at first as well 😅

crates/tauri-bundler/src/bundle/linux/appimage/experimental.rs
```diff
@@ -0,0 +1,257 @@
...
+      "-f",
+      "--set-owner",
+      "0",
+      "--set-group",
+      "0",
+      "--no-history",
+      "--no-create-timestamp",
+      "--compression",
+      "zstd:level=22",
+      "-S24",
```
Samueru-sama: What we were doing caught the attention of upstream dwarfs and now there has been several improvements and the recommended parameters are `-S26 -B8` instead.

FabianLars: cool, will update later.

do you think uruntime and dwarfs is something we can default to? would we even need to offer an option to switch to appimagetool?

Samueru-sama: > do you think uruntime and dwarfs is something we can default to? would we even need to offer an option to switch to appimagetool?

TLDR: Yesn't. 


The uruntime is 100% compatible witth the type2-runtime, and it is even [bug free](https://github.com/zen-browser/desktop/pull/6130) xd

Now _dwarfs_ uruntime is not 100% compatible, that is while it works with: 

* AM

* dbin

* Gearlever

* Soar

* AppImageUpdate


It does **not** work with appimagelauncher or the appimaged daemon, because these only support squashfs. 

Note that these two also have some compatibility issues between each other 😫, appimages made with go-appimage (which is where appimaged is from) are not compatible with appimagelauncher because appimagelauncher needs the runtime to be static-pie, which go-appimage does not do while the uruntime is btw. 

appimagelauncher also had an issue for a long time that it was not compatible with static runtimes (like the type2-runtime) for years until a few months ago, this was fixed by appimagelauncher but that fix has not made it to the stable releases yet.

There is also a PR opened at go-appimage to give support for dwarfs for appimaged but I don't think it will be merged anytime soon 👀

So if you ask me I wouldn't add support for appimagetool, specially considering this would introduce a bug of a crash on sleep that the type2-runtime currently has. At most I would add an option to use the squashfs-uruntime instead.

-----------------------------------------------------

OFF-TOPIC: Something that I would really love to see being added is support for zsync updates, which is the official way for appimages to update. This method provides delta updates, meaning it doesn't download the entire appimage but only the diff, we actually have a fork of appimageupdatetool that bypasses the github API limits because we abuse the thing to the point that the API is an issue for us lol

https://github.com/pkgforge-dev/AppImageUpdate-Enhanced-Edition

You can see some benchmark in the readme where it only uses 2.65 MIB to update CPU-X 👀

FabianLars: Okay thanks for the extensive answer. I think we'll default to uruntime+dwarfs with an option for uruntime+squashfs for now then while being open to switch defaults on user feedback.

i've wanted to add zsync support (for external updates) for ages but basically ceased work on appimages until your work and sharun etc came around.

imo self updates via zsync would be a nice addition as well and should be something we (again, in my opinion) could be added to the existing updater plugin. I have to admit that it's not exactly high prio for me though so i wouldn't work on this myself in the foreseeable future unless my brain's funny again and randomly decides to hyperfocus on that.

crates/tauri-cli/config.schema.json
```diff
@@ -3034,7 +3034,7 @@
           "type": "boolean"
         },
         "updateInformation": {
-          "description": "Update information to forward to the AppImage tooling according to <https://github.com/AppImage/AppImageSpec/blob/master/draft.md#update-information>.\n Can also be provided via the `UPINFO` env var.\n\n Requires `zsyncmake` to be installed on the build system.",
+          "description": "Update information to forward to the AppImage tooling according to <https://github.com/AppImage/AppImageSpec/blob/master/draft.md#update-information>.\n Can also be provided via the `UPINFO` env var.\n\n When enabled, the generated AppImage will not include the app version in its file name.\n\n Requires `zsyncmake` to be installed on the build system.",
```
Samueru-sama: Why this change? the [appimage spec](https://github.com/AppImage/AppImageSpec/blob/master/draft.md#type-2-image-format) recommends: `ApplicationName-$VERSION-$ARCH.AppImage`

AppImageUpdate already supports simple globbing in the `.zsync` filename.


`"gh-releases-zsync|$(echo "$GITHUB_REPOSITORY" | tr '/' '|')|latest|*x86_64.AppImage.zsync"` will match any file that ends wtih `x86_64.AppImage.zsync` and is on the latest stable release.

crates/tauri-bundler/src/bundle/linux/appimage/sharun.rs
```diff
@@ -18,6 +18,7 @@ use crate::{
 
 use super::write_and_make_executable;
 
+// TODO: xdg-open exec script
```
Samueru-sama: This is actually a bug on our end and I haven't implemented the `gio-desktop-launch` hook yet because I want max to refactor `lib4bin` a little first xd

crates/tauri-cli/config.schema.json
FabianLars: I misunderstood your earlier comment about how the file name isn't allowed to change.

crates/tauri-bundler/src/bundle/linux/appimage/sharun.rs
FabianLars: How refreshing lol. The other tooling always wanted a user side fix. Anyway, i'll probably still do something very simple here until it's fixed on your side, we'll see how long it takes for this or here to get ready&merged 😂

crates/tauri-cli/config.schema.json
Samueru-sama: > I misunderstood your earlier comment about how the file name isn't allowed to change.

What I meant is that right after the info is added to the runtime and the `.zsync` file is made, the file has to be released as as and no further changes to the name can be made. 

This is a problem I've seen quite a few other projects run into, specially those that make `VERSION` a commit sha and then change it at the last minute before making a stable release.

crates/tauri-bundler/src/bundle/linux/appimage/sharun.rs
Samueru-sama: > How refreshing lol. The other tooling always wanted a user side fix. Anyway, i'll probably still do something very simple here until it's fixed on your side, we'll see how long it takes for this or here to get ready&merged 😂

yeah in the meantime making [a simple wrapper](https://github.com/pkgforge-dev/GIMP-AppImage/blob/main/gimp-appimage.sh#L64-L67) script that calls `xdg-open` works.

crates/tauri-cli/config.schema.json
FabianLars: Yeah got it now. Will try to stress that in the docs.

crates/tauri-bundler/src/bundle/linux/appimage/sharun.rs
FabianLars: are there any requirements for that wrapper to work? I added it to this pr locally and it doesn't seem to work for me locally, same for the gimp appimage.

My setup is a bit scuffed though so maybe it's me.

Samueru-sama: > are there any requirements for that wrapper to work? I added it to this pr locally and it doesn't seem to work for me locally, same for the gimp appimage.
> 
> My setup is a bit scuffed though so maybe it's me.

you have `xdg-open` right? 

When you say gimp doesn't work that is when you try to open links from gimp the web browser doesn't open?

FabianLars: Yes to both. I have xdg-open and I'm trying to open the links in the gimp welcome screen.

fwiw xdg-open works in my test tauri app outside of the appimage

Samueru-sama: > Yes to both. I have xdg-open and I'm trying to open the links in the gimp welcome screen.
> 
> fwiw xdg-open works in my test tauri app outside of the appimage

What do you get on the terminal? I just tested it and it works. 

Something I did notice is a lot of gdk errors on the terminal when opening links, this is because of the gdk env variables propagating to the web browser, I need to unset them in the wrapper script. but it still works. 

This was actually added after this [bug](https://github.com/pkgforge-dev/GIMP-AppImage/issues/18) report as well

FabianLars: > What do you get on the terminal? I just tested it and it works.

Nothing when i click on a link. But yeah, could be my system as i'm using wsl  🤷

Samueru-sama: > Nothing when i click on a link. But yeah, could be my system as i'm using wsl 🤷

[Try again](https://github.com/pkgforge-dev/GIMP-AppImage/blob/main/gimp-appimage.sh#L65-L104)

FabianLars: still nothing, just so we're on the same page, i'm testing these links here 
grafik

Samueru-sama: > still nothing, just so we're on the same page, i'm testing these links here !

Yeah that's working here without issue on artix, ubuntu 14.04 (yes 14.04) and alpine linux. 

Nothing shows on the terminal? weird. 

You might wanna check extracting the AppImage and then editing `bin/xdg-open` and add something like `echo test` to make sure that gimp is indeed calling `xdg-open` when you go to open links. if it does run it then the issue is with the `xdg-open` on the system, likely a missing mimetype.

FabianLars: yeah idk, as long as it works outside of wsl on real systems/vms i don't really care and will postpone that to later. Just felt a bit weird that it's only an issue inside the appimages 🤷

FabianLars:
I think this is ready enough for reviews and actual testing

crates/tauri-bundler/src/bundle/linux/appimage/sharun.rs
```diff
@@ -0,0 +1,296 @@
...
+    let data = download(&format!("https://github.com/VHSgunzo/uruntime/releases/latest/download/uruntime-appimage-{fstype}-lite-{arch}"))?;
+    write_and_make_executable(&uruntime_lite, data)?;
+  }
+
+  let lib4bin = tools_path.join(format!("lib4bin-{arch}"));
+  if !lib4bin.exists() {
+    let data =
+      download("https://raw.githubusercontent.com/VHSgunzo/sharun/refs/heads/main/lib4bin")?;
+    write_and_make_executable(&lib4bin, data)?;
+  }
```
zer0x64: Quick note: lib4bin requires `patchelf`, which is not on default ubuntu(at least the base incus image I'm using to build), might be worth documenting. Also, the error message I got was a simple "/bin/sh command failed" which doesn't help much identifying the issue.

zer0x64 (commented):
I don't know much about the old and new format, but I tested it and I got some weird behavior.

I've got 2 different PC on Arch linux, one with discrete Nvidia and one with a AMD iGPU. Both builds the appimage in an incus ubuntu 24 container.

Old format:
- Nvidia works perfectly
- AMD doesn't work at all with the invalid EGL parameters error.

New format:
- Nvidia gets a bunch of "libEGL warning: egl: failed to create dri2 screen" warnings and bad performance, but it's working.
- AMD works perfectly.

I don't know if the Nvidia/AMD divide is the real culprit, but I've got very simillar setups otherwise so that's my guess.

I cannot open source my codebase yet (I can do it in about two weeks though) for reproduction, but if you have any tests to make me run, I'd be happy to help!

EDIT: Some more warning thrown with the new format with my NVidia PC. A quick google search tells me we might need to pack libnvidia-egl-wayland1 too
```
libEGL warning: egl: failed to create dri2 screen
libEGL warning: egl: failed to create dri2 screen
libEGL warning: egl: failed to create dri2 screen
Gtk-Message: 22:08:03.683: Failed to load module "window-decorations-gtk-module"
Gtk-Message: 22:08:03.683: Failed to load module "colorreload-gtk-module"
Gtk-Message: 22:08:03.788: Failed to load module "window-decorations-gtk-module"
Gtk-Message: 22:08:03.788: Failed to load module "colorreload-gtk-module"
libEGL warning: egl: failed to create dri2 screen
libEGL warning: egl: failed to create dri2 screen
```

Samueru-sama:
> EDIT: Some more warning thrown with the new format with my NVidia PC. A quick google search tells me we might need to pack libnvidia-egl-wayland1 too
> 
> ```
> libEGL warning: egl: failed to create dri2 screen
> libEGL warning: egl: failed to create dri2 screen
> libEGL warning: egl: failed to create dri2 screen
> Gtk-Message: 22:08:03.683: Failed to load module "window-decorations-gtk-module"
> Gtk-Message: 22:08:03.683: Failed to load module "colorreload-gtk-module"
> Gtk-Message: 22:08:03.788: Failed to load module "window-decorations-gtk-module"
> Gtk-Message: 22:08:03.788: Failed to load module "colorreload-gtk-module"
> libEGL warning: egl: failed to create dri2 screen
> libEGL warning: egl: failed to create dri2 screen
> ```

Do you have those errors with this [AppImage](https://github.com/probonopd/PrusaSlicer.AppImage/releases/tag/2.9.1)?

Also check if hardware acceleration is working [here](https://github.com/pkgforge-dev/PPSSPP-AppImage/releases/tag/1.18.1-3)

You can check that quickly by checking if the app shows up in nvtop

crates/tauri-bundler/src/bundle/linux/appimage/sharun.rs
Samueru-sama: `lib4bin` should have printed that patchelf was missing 🧐

image

zer0x64: That's how I figured it out, by running lib4bin manually. However, tauri-cli masks the error when running `cargo tauri build`

zer0x64:
> Do you have those errors with this [AppImage](https://github.com/probonopd/PrusaSlicer.AppImage/releases/tag/2.9.1)?

The Failed to load modules are still there, but not the `failed to create dri2 screen` ones, and performance looks good

> Also check if hardware acceleration is working [here](https://github.com/pkgforge-dev/PPSSPP-AppImage/releases/tag/1.18.1-3)
> 
> You can check that quickly by checking if the app shows up in nvtop

PPSSPP seems to work perfectly, and it does use hardware acceleration

However, I checked and my tauri app do not appear in nvtop

FabianLars:
i removed the libegl line as that's the only difference in the lib4bin call compared to the prusaslicer appimage.

@zer0x64 can you re-install the cli (may require the --force flag) and try again?

Really appreciate you testing this :)

FabianLars:
If that doesn't help the only thing left i see is the xvfb-run use which i hoped we could avoid (either way could be added as a config option i guess)

zer0x64:
> @zer0x64 can you re-install the cli (may require the --force flag) and try again?

Got even worse result, the app simply crash on launch now with error "Could not create GBM EGL display: EGL_SUCCESS. Aborting..."
 
> Really appreciate you testing this :)

That's the very least I could do :sweat_smile: I wish I could have helped more than that

Samueru-sama:
> > Do you have those errors with this [AppImage](https://github.com/probonopd/PrusaSlicer.AppImage/releases/tag/2.9.1)?
> 
> The Failed to load modules are still there, but not the `failed to create dri2 screen` ones, and performance looks good
> 
> > Also check if hardware acceleration is working [here](https://github.com/pkgforge-dev/PPSSPP-AppImage/releases/tag/1.18.1-3)
> > You can check that quickly by checking if the app shows up in nvtop
> 
> PPSSPP seems to work perfectly, and it does use hardware acceleration
> 
> However, I checked and my tauri app do not appear in nvtop

Now run the tauri app with `LD_DEBUG=libs` and send the output here.

zer0x64:
This is the full output without the bundled libegl (result: Doesn't launch):
[test.txt](https://github.com/user-attachments/files/20104331/test.txt)

This is the output with the bundled libegl (result: No hardware acceleration on Nvidia)
[test.txt](https://github.com/user-attachments/files/20104421/test.txt)

Samueru-sama:
> This is the output with the bundled libegl (result: No hardware acceleration on Nvidia) [test.txt](https://github.com/user-attachments/files/20104421/test.txt)

huh, the output seems fine, try this instead: 

run the appimage with `--appimage-extract`

And run the `AppRun` with strace: `strace -f ./AppRun 2>~/debug` and send that file here (will likely need to be compressed). 

Also linking the AppImage would be nice, since the sharun dev himself uses nvidia and can debug better 👀

Samueru-sama:
Try including `/usr/lib/x86_64-linux-gnu/gbm/*` to `lib4bin` since I noticed `libgbm.so.1` was bundled in the AppImage. 

This is likely a hook that we need to include at sharun btw 👀

FabianLars:
> Try including /usr/lib/x86_64-linux-gnu/gbm/* to lib4bin since I noticed libgbm.so.1 was bundled in the AppImage.

pushed 👍

zer0x64:
@Samueru-sama  Do you know which package the gbm/ dir comes from? It's simply not there on my container

zer0x64:
As for linking the AppImage, I unfortunately cannot do that publicly yet. I could send it via a private channel, but as a security guy I understand how that would not be optimal

Samueru-sama:
> @Samueru-sama Do you know which package the gbm/ dir comes from? It's simply not there on my container

 I really don't know sorry. 

We stopped building on debian/ubuntu all together because pain...

On archlinux you install mesa and you get that, on debian it is split into a zillion pieces, same for all other packages, often times they also conflict with each other for some reason and not mention everyone eventually has to use pip to workaround some nonsense xd

>As for linking the AppImage, I unfortunately cannot do that publicly yet. I could send it via a private channel, but as a security guy I understand how that would not be optimal

https://discord.gg/djJUs48Zbu best I can do xd

zer0x64:
@FabianLars been diagnosing this on discord with @Samueru-sama . Would it be possible to add a commit to remove the -s option from lib4bin? Or make it configurable, we think that might be the issue

FabianLars:
Sure, it was configurable earlier but I changed it cause the Linux deploy issue that was the reason I made it configurable doesn't apply here.

I'm only on my phone so I'll simply remove it without a config for now.

zer0x64:
Progress update: We narrowed it down to wayland, forcing it to use XWayland by setting an empty WAYLAND_DISPLAY seems to enable hardware accel (although the warnings are still there)

zer0x64:
> @Samueru-sama Do you know which package the gbm/ dir comes from? It's simply not there on my container

That still seems to be the leading theory, but I can't find the right package :melting_face:

FabianLars:
> Progress update: We narrowed it down to wayland, forcing it to use XWayland by setting an empty WAYLAND_DISPLAY seems to enable hardware accel (although the warnings are still there)

gives me linuxdeploy ptsd

> That still seems to be the leading theory, but I can't find the right package 🫠

How about those for a start:
```
libgm1 (or libgm-dev)
mesa-common-dev
mesa-vdpau-drivers
mesa-vulkan-drivers
mesa-va-drivers
```

FabianLars:
i mean, if we have to use arch as well then so be it, i'm so done with fucking desktop linux and wanted to drop support as a whole for ages now so we may as well add another build restriction. 🤷

Samueru-sama:
> i mean, if we have to use arch as well then so be it, i'm so done with fucking desktop linux and wanted to drop support as a whole for ages now so we may as well add another build restriction. 🤷

Yeah the news are not great, @zer0x64 is gonna be the messenger though 👀

FabianLars:
ah for the deps, did you check https://github.com/probonopd/PrusaSlicer.AppImage/blob/2.9.1/.github/workflows/build.yml#L33-L34 since that appimage seems to work fine?

zer0x64:
So yeah, the TL;DR:

Ubuntu 24.04 (the latest LTS) uses an older mesa version that handle dri library loading differently. Here's an article about the switch:
https://www.phoronix.com/news/Mesa-24.2-Modern-Interfaces

Sharun does not handle both ways at the same time. That means that to get hardware accel working on nvidia, you'd need to build on at least Ubuntu 25.04 (will try to test that tonight, that will require a bit of time to test)

If built on Ubuntu 24, nvidia hardware accel DOES work by forcing XWayland(by setting an empty WAYLAND_DISPLAY env variable)

I think it is already a major improvement over the linuxdeploy-based build. For my case, I went from "not working at all on AMD" to "works on both GPUs, one is laggy unless you add an env variable"

p-dugre-pvotal:
And yeah, small rant, debian being always outdated as fuck and being the standard for everything is so annoying when you want to pre-build anything. I totally get why you're tired of supporting it :upside_down_face:

zer0x64:
And yeah, small rant, debian being always outdated as fuck and being the standard for everything is so annoying when you want to pre-build anything. I totally get why you're tired of supporting it 🙃

Samueru-sama:
>Sharun does not handle both ways at the same time. That means that to get hardware accel working on nvidia, you'd need to build on at least Ubuntu 25.04 (will try to test that tonight, that will require a bit of time to test)

We actually still handle it both ways, sharun still sets `LIBGL_DRIVERS_PATH`.

But yeah everything points to `gbm` being needed here, which ubuntu 24.04 does not have: 

image

I'm not very sure why wayland with hardware accel does work with PrusaSlicer, I think they actually coded some workarounds for webkit2gtk. I'm assuming you checked that wayland was working with it btw 👀

I also haven't consulted with VHSgunzo, maybe he has an idea xd


**EDIT:** Prusa was not using wayland nvm

FabianLars:
@Samueru-sama since you mentioned Arch, is there anything i should look out for or would adding arch support be straight forward (in theory)? except the different /usr/lib path

Ubuntu is not a hard requirement for us/tauri. I just assumed it was the best supported coming from other appimage tooling. I wanted to support arch (and ideally fedora) in the future anyway.

Since users need 2 workflow jobs (one for appimage with latest ubuntu, one for deb/rpm for oldest) they may as well use an arch container for the appimage job.

Samueru-sama:
> @Samueru-sama since you mentioned Arch, is there anything i should look out for or would adding arch support be straight forward (in theory)? except the different /usr/lib path
> 
> Ubuntu is not a hard requirement for us/tauri. I just assumed it was the best supported coming from other appimage tooling. I wanted to support arch (and ideally fedora) in the future anyway.
> 
> Since users need 2 workflow jobs (one for appimage with latest ubuntu, one for deb/rpm for oldest) they may as well use an arch container for the appimage job.

Arch is fine and perfect for building appimages, because arch doesn't split packages into a million pieces and puts all the libs in `/usr/lib`

We use [archlinux to build most of our appimages](https://github.com/search?q=ghcr.io%2Fpkgforge-dev%2Farchlinux%3Alatest&type=code) in pkgforge-dev, and the ones that don't use alpine linux instead. 

Since tauri apps are hardware accelerated alpine cannot be used since that would break compatibility with the proprietary nvidia driver which only works with glibc. (remember in these appimages the libc is bundled). 

We also use some [modified archlinux packages](https://github.com/pkgforge-dev/llvm-libs-debloated) which greatly reduce the size of the appimage, they usually end up being 20% smaller than the linuxdeploy equivalent while being able to work everywhere.

However archlinux has a problem, and it is that it doesn't officially support `aarch64`, this means we have to use a fork called ArchlinuxARM and sometimes we have issues [like this](https://github.com/interstellar-app/interstellar/issues/129#issuecomment-2845877741), so until archlinux starts officially supporting arm I can't guarantee 24/7 stability on the repo 🥲

Samueru-sama:
Alright [sharun now has a gbm hook](https://github.com/VHSgunzo/sharun/commit/3860033a659121b935b4f6740eb9489639ab2a0c), so it doesn't have to be explicitly passed to lib4bin anymore.

ByteAtATime:
When building with this, the AppImage doesn't seem to include any sidecar binaries. I get an os error 2 when running the app, and the sidecar isn't present in the AppDir.

Samueru-sama:
@FabianLars sharun now has something called sharun-aio. It has a bundled `lib4bin` along with all the dependencies.

This should simplify the CI since you no longer have to be installing patchelf, strip, strace, etc

Usage:

```
xvfb-run -a ./sharun-aio l -p -v -e -s -k /path/to/libs_and_bins
rm ./sharun-aio
ln ./sharun ./AppRun
./sharun -g
```

FabianLars:
oh that's nice. will take a look. thx

tonymushah:
For context,
I use Fedora 42 on a daily basis.
I have a [tauri project](https://github.com/tonymushah/special-eureka)  that is ""mature"" enough for testing this.
I have no experience about package bundling.
_But I got time to kill so, I thought it was maybe worth it to test it?_

First, I made a [CI](https://github.com/tonymushah/special-eureka/pull/1037) then [a branch](https://github.com/tonymushah/special-eureka/tree/experimental-app-image-branch) to experiment with this feature. 

And after spending an entire day figuring out on how to compile my project with this,
the [builds](https://github.com/tonymushah/special-eureka/releases/tag/v0.2.1-1.0) **work** _fine_.
I mean, all of the tauri plugins*, my custom plugins and the CSS** works but it was a little bit wonky and less performant.

Then, I proceed on using `ubuntu-latest` instead of `ubuntu-24.04` on the CI and [it](https://github.com/tonymushah/special-eureka/releases/tag/v0.2.1-1.1) is less wonky than the previous one... I can't tell.

All I can tell is that, instead of having a blank screen that show me nothing (You know that `Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...` error), this feature fixed it.

*: I haven't tested it with the `updater` feature yet.
**: The method of using [`libwebkit2-gtk-4.1-0=2.44.0-2`](https://github.com/tonymushah/special-eureka/blob/19d69defd883057153ca7c4e239a338425cccc56/.github/workflows/build.yaml#L47) on CI broke my app's CSS on release builds.

oscar370:
This fixed the white window problem (`Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...`), but the backend (a PyInstaller-packed Python binary) now fails to start.

The runtime log shows:

```
Backend error: Failed to set ELF interpreter:
/tmp/.mount_pywalremp8008463124221033243/shared/bin/main: Read-only file system (os error 30)
Backend terminated with code: Some(1)
```

However, the exact same AppImage works fine when executed under `strace`, for example:

```bash
strace -f ./pywalify_0.1.0_amd64.AppImage 2>&1 | grep -i "read-only"
```

Under `strace`, there are no read-only errors, and both the backend and frontend work as expected.

Samueru-sama:
> This fixed the white window problem (`Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...`), but the backend (a PyInstaller-packed Python binary) now fails to start.
> 
> The runtime log shows:
> 
> ```
> Backend error: Failed to set ELF interpreter:
> /tmp/.mount_pywalremp8008463124221033243/shared/bin/main: Read-only file system (os error 30)
> Backend terminated with code: Some(1)
> ```
> 
> However, the exact same AppImage works fine when executed under `strace`, for example:
> 
> ```shell
> strace -f ./pywalify_0.1.0_amd64.AppImage 2>&1 | grep -i "read-only"
> ```
> 
> Under `strace`, there are no read-only errors, and both the backend and frontend work as expected.

The problem here is that the pyinstaller binary is trying to write inside the AppImage, which is a read only filesystem. 

The reason it work with strace is because when strace is used calls to suid binaries like `fusermount*` are blocked and uruntime automatically falls back to extract and run, which bypasses FUSE and just extracts the entire filesystem to `/tmp` which makes it no longer read only. 

image

This behaviour with strace will should also no longer happen with newer versions of the uruntime, since [now it can also use namespaces to mount instead of fusermount.](https://github.com/VHSgunzo/uruntime/issues/18)

TLDR: You will have to do [something like this](https://github.com/pkgforge-dev/gpu-screen-recorder-AppImage/blob/main/AppDir/bin/set-caps.src.hook#L46-L84) if you need to write to the filesystem

Samueru-sama:
btw @oscar370 sharun can be used to deploy python as well and it is much better, it doesn't use hacks like setting interpreter at runtime. 

[yt-dlp started to have issues with pyinstaller](https://github.com/yt-dlp/yt-dlp/issues/14694#event-20618982834) and I suggested them to use sharun which we already use for yt-dlp without problem but looks like they are not interested...

xTVaser:
Finally got around to giving this branch a shot as well - https://github.com/open-goal/launcher/pull/853
- Built with `ubuntu-24.04` (minimal change using the provided gha runner images)
- Pulled in the branch via `cargo install tauri-cli --git ...`

Initial impressions are pretty good:
- Users that reported it breaking on Arch now report it working
  - AppImages work again on other Arch systems like the SteamDeck
- Haven't heard from a Fedora user yet but I'm optimistic considering [this comment](https://github.com/tauri-apps/tauri/pull/12491#issuecomment-3386576810)
- Noticed zero issues or changes about the application itself
- The updater also seems to work fine (it doesn't relaunch the AppImage after updating but I can't remember if that was ever the case)
  - Was also able to update between new format versions
  - ~~Was able to update from an old format AppImage to the new format~~ Scratch this, realized that of course I couldn't test this since the old formats wouldn't have launched -- this might be something worth still testing.

FabianLars:
> (it doesn't relaunch the AppImage after updating but I can't remember if that was ever the case) 

It was supposed to work for the old appimage and when updating from old to new appimage 🤔 Doesn't look like a code issue in your app at first glance. I'll test that before the merge, so thanks for the report!

FTCHD:
Fingers crossed this gets merged before Jan 24 this year, it's a substantial upgrade for everyone with users on Linux.

FabianLars:
what happens on jan 24?

FTCHD:
> what happens on jan 24?

Celebrating the first birthday of this PR 🎂

FabianLars:
oh damn 🤣

andersmmg:
Does this PR not support resources yet? They don't seem to be bundled in the AppImage like they are in the deb when I use this branch

FabianLars:
that could be possible. sidecars weren't working for a long time either. thanks for bringing that up :)

DoctorDerek:
I'm bumping this excellent PR, which should be merged promptly, as I can confirm all of the following:
1. Tauri @latest builds break with "White Screen on Linux" (built on WSL Ubuntu, target Steam Deck).
2. "Widely-recommended ENV variables" do NOT fix the "White Screen on Linux" issue, at all, period.
3. The current PR / branch fixes that issue (no white screen, Tauri app works correctly on Steam Deck)
4. This PR has been open (meaning Tauri has been broken on Linux) for almost a full year, apparently.
5. This is a known issue in Tauri #11994.
6. People are still reporting this Linux issue across many other Tauri-based repositories across GitHub.

## How To Fix White Screen on Linux Tauri Builds (Using a Working Branch)
Here's the command I used (in WSL) to build a working Linux build (no white screen on Steam Deck):
```
cargo install tauri-cli --git https://github.com/tauri-apps/tauri --branch feat/truly-portable-appimage
export TAURI_BUNDLER_NEW_APPIMAGE_FORMAT="true"
cargo tauri build
```

I'm tagging @lucasfernog as it may not be widely known that Tauri is broken on Linux / Steam Deck.

ScottMorris:
Hey @FabianLars, it looks like there was an update with the latest `dev` branch! 😻 Thanks :)

I just checked the CLI version and it looks like the work in this PR is now using CLI `"version": "2.9.6"`.  I'm guessing this means if we're currently using this branch for improved AppImage support we have access to the latest CLI and tooling updates?

FabianLars:
yes, including some unreleased changes that will be part of 2.10 (important for the next updater plugin release).

when i'm less busy at work (where i'm also working on appimages atm) i may make heavy changes to this PR to use @Samueru-sama quick-sharun tool instead. I'll have to check how it works on non-arch distros first but just so you're prepared as it may break things momentarily.

ScottMorris:
That's really cool, I'll keep an eye on things.  I use an Ubuntu 24.04 image for my dev container and in my CI.  It sounds like this won't be permanently impacted?

Looking forward to what the next updates have in store!

Samueru-sama:
> @Samueru-sama quick-sharun tool instead. I'll have to check how it works on non-arch distros first but just so you're prepared as it may break things momentarily.

I've been quiet about `quick-sharun` here because it is meant to be used on archlinux only. It can be used on ubuntu but I only know that because one person used it there lol

There is also a bigger issue that will likely impact the next ubuntu release and it is [this nosense](https://github.com/VHSgunzo/sharun/issues/68) which is fixed in `quick-sharun` with a hack though.

FabianLars:
yeah i used it a bit on ubuntu as well today because my main wsl instance runs ubuntu and at least for a webkitgtk-free appimage it worked fine.

i still lean towards semi-forcing devs to use arch for good appimages though but we'll see when we get there. honestly i'd trade ubuntu for fedora support lol but i think i'm very biased here.

ScottMorris:
The challenge with Arch only is it adds more layers to working with GitHub runners. They appear to only have Ubuntu Linux runners [out of the box](https://github.com/actions/runner-images).  I aligned my dev environment for Tauri to mirror their runner so it was easier to debug issues. It's been pretty smooth, but having AppImage support working is more important.

Getting an Arch container working is probably trivial, so if that's the path then some retooling will be in order :)

FabianLars:
Changing the runner container is indeed very easy (a single line) and will be recommended and displayed in the docs no matter whether it'll be arch only.
At least in the beginning, maybe we can get the tooling/packages to support more distros just as well. That is, if Ubuntu can stop messing up, which I doubt.

hadim:
I am using this PR and it works fine for me. I am wondering what is blocking this from being merged and released?

debba:
I tried to apply the patch on my project, but didn't work as expected.

With appimage still continue to have blank screen and  I can see this error:

Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...

Do you have some suggestions?
https://github.com/debba/tabularis/blob/main/.github/workflows/release.yml

Samueru-sama:
> I tried to apply the patch on my project, but didn't work as expected.
> 
> With appimage still continue to have blank screen and I can see this error:
> 
> Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
> 
> Do you have some suggestions? https://github.com/debba/tabularis/blob/main/.github/workflows/release.yml

You are most likely missing some lib.

I can setup some repo to build the appimage using `quick-sharun` instead, it takes into account a lot of cases that would result in missing libs. 

edit: gonna do it now lol, it is super simple.

FabianLars:
ah whoops, i thought i pushed the quick-sharun change to this branch as well, mb

debba:
@FabianLars  oh so maybe depends on it?

Samueru-sama:
@debba let me know if you get OpenGL issues with this: 

<https://github.com/Samueru-sama/tabularis-appimage-demo/releases/tag/0.9.6%402026-03-08_1772964975>

debba:
@Samueru-sama Will do it for sure, thanks .
Unfortunately I am afk nor

debba:
It woks @Samueru-sama .
How can I migrate my workflow?

Drsheppard01:
> I tried to apply the patch on my project, but didn't work as expected.
> 
> With appimage still continue to have blank screen and I can see this error:
> 
> Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
> 
> Do you have some suggestions? https://github.com/debba/tabularis/blob/main/.github/workflows/release.yml

Usually needed next deps: 

```Shell
libappindicator3-dev \
librsvg2-dev \
patchelf \
libwebkit2gtk-4.1-0=2.44.0-2 \
libwebkit2gtk-4.1-dev=2.44.0-2 \
libjavascriptcoregtk-4.1-0=2.44.0-2 \
libjavascriptcoregtk-4.1-dev=2.44.0-2 \
gir1.2-javascriptcoregtk-4.1=2.44.0-2 \
gir1.2-webkit2-4.1=2.44.0-2
```

Follow: https://github.com/appimage-sig/Errors?tab=readme-ov-file#tauri-related-issues

FabianLars:
@debba If you re-install tauri-cli from this branch's last commit it should be more or less the same though you may want to copy the arch container stuff into your CI (docs on this will follow before PR is getting merged).

@Drsheppard01 This PR is supposed to fix issues like the one that made those specific webkitgtk versions required for some. So i'd prefer if nobody uses these versions in this PR so we can actually test it :P

debba:
Great very good work, thanks @FabianLars

ScottMorris:
> you may want to copy the arch container stuff into your CI

Thanks for the updates, it looks like there is a nice example of the CI in @Samueru-sama's demo `appimage.yml` workflow.

https://github.com/Samueru-sama/tabularis-appimage-demo/blob/a7b3dc8bd8389ba6d644546413b153ca79ba192c/.github/workflows/appimage.yml#L24

debba:
@Samueru-sama can i ask you what I am wrong in https://github.com/debba/tabularis/blob/main/.github/workflows/release.yml ?

Because I was installing tauri-cli started to this branch, but does not work.
From your repo works well!

Samueru-sama:
> @Samueru-sama can i ask you what I am wrong in https://github.com/debba/tabularis/blob/main/.github/workflows/release.yml ?
> 
> Because I was installing tauri-cli started to this branch, but does not work. From your repo works well!

Link the artifact that has the issues. 

But be aware if you are deploying on ubuntu it is likely not going to work, see https://github.com/tauri-apps/tauri/pull/12491#issuecomment-2863900055

In my repo what I did was just take the deb releases of the app and deploy that directly with `quick-sharun` [on archlinux](https://github.com/Samueru-sama/tabularis-appimage-demo/blob/a7b3dc8bd8389ba6d644546413b153ca79ba192c/.github/workflows/appimage.yml#L24)

debba:
@Samueru-sama thanks, this is the artifact:

https://github.com/debba/tabularis/releases/download/v0.9.7/tabularis_0.9.7_amd64.AppImage

So I need to change the machine name for building tauri app?
I used ubuntu 24.04

Samueru-sama:
> @Samueru-sama thanks, this is the artifact:
> 
> https://github.com/debba/tabularis/releases/download/v0.9.7/tabularis_0.9.7_amd64.AppImage

That appimage was not made with `quick-sharun`, it was made with linuxdeploy kekw

debba:
@Samueru-sama but in workflow I added installation of tauri-cli starting from this branch.

https://github.com/debba/tabularis/blob/main/.github/workflows/release.yml

FabianLars:
you have to change tauriScript to `tauriScript: "cargo tauri"` so that it uses the tauri-cli cargo bin you just installed. `pnpm tauri` keeps using the published npm package.

debba:
@FabianLars  ouch sorry, I feel like an idiot, bm

FTCHD:
@FabianLars can we rebase this with the latest tauri updates, please?

Samueru-sama:
> Edit1: issue with bun's single exe output [oven-sh/bun#28281](https://github.com/oven-sh/bun/issues/28281)

Fixed with [a hack](https://github.com/pkgforge-dev/Anylinux-sharun/blob/3b1a868e14cb1f79392be30205e4cb108e4470ca/src/main.rs#L397-L411) in my fork of sharun.

This means we are forced to set `LD_LIBRARY_PATH` and `LD_PRELOAD`. To prevent those vars from leaking into child processes, [anylinux.so](https://github.com/pkgforge-dev/Anylinux-AppImages/blob/main/useful-tools/lib/anylinux.c) makes sure to unset them via intercepting execve calls and similar.

In practice this means any bun binary will correctly work regardless of the distro. (As long as the kernel is newer than 5.1 since bun has this hard requirement lol).

FabianLars:
@Samueru-sama sorry to ping you here but since yesterday i got a few reports about the bundler hanging. I see the same in my simple test workflow here: https://github.com/FabianLars/mw-toolbox/actions/runs/28452723625/job/84319084321

It _seems_ like it gets stuck on stracing MiniBrowser but i have no idea what's even happening there so i don't think i can fix that 😅 Speaking of MiniBrowser - most apps won't need that, can we maybe not bundle it (optionally perhaps)?

Note that this doesn't seem to happen on Arch - only tested ubuntu so far (i know i know, building on ubuntu is stupid)

Samueru-sama:
> @Samueru-sama sorry to ping you here but since yesterday i got a few reports about the bundler hanging. I see the same in my simple test workflow here: https://github.com/FabianLars/mw-toolbox/actions/runs/28452723625/job/84319084321
> 
> It _seems_ like it gets stuck on stracing MiniBrowser but i have no idea what's even happening there so i don't think i can fix that 😅 Speaking of MiniBrowser - most apps won't need that, can we maybe not bundle it (optionally perhaps)?
> 
> Note that this doesn't seem to happen on Arch - only tested ubuntu so far (i know i know, building on ubuntu is stupid)

Well the problem is because of this error: 

`/home/runner/.cache/tauri/quick-sharun.sh: 1367: set: can't access tty; job control turned off`

it means `set -m` is failing, as result I cannot guarantee that the `MiniBrowser` process is killed during strace mode.

`MiniBrowser` might not be needed, but this problem can manifest with other processes including the app being deployed.

Btw we are no longer using `strace`, I keep calling it strace mode, but now it uses `LD_DEBUG=libs` to get the list of dlopened libraries.

https://github.com/pkgforge-dev/Anylinux-AppImages/blob/f89065c8266d1bfed94f405c41efc8f1489134f8/useful-tools/quick-sharun.sh#L1367-L1374

**I'm not sure how to fix this other than directly failing if `set -m` isn't possible.**

FabianLars:
Okay, scratch ubuntu 2204 then, GitHub deprecates it soon anyway - just wanted to check how bad 2204 really is/was. I remember we talked about mesa stuff a while ago that required at least 2504 or something.

Speaking of which, i tried Ubuntu 2604 (because an arch-only bundler would really suck for us here tbh) and that fails with good old pixbuf issues. Not sure if it's the same as https://github.com/VHSgunzo/sharun/issues/68

Anyway, i don't want to keep this PR open for another year so i'll rename things a bit and add more docs/warnings to have it arch only and marked as experimental 🤷

Samueru-sama:
> Speaking of which, i tried Ubuntu 2604 (because an arch-only bundler would really suck for us here tbh) and that fails with good old pixbuf issues. Not sure if it's the same as [VHSgunzo/sharun#68](https://github.com/VHSgunzo/sharun/issues/68)

That's odd, `quick-sharun` has code to handle glycin correctly. (basically disable the glycin sandbox since it relies on bwrap with a ton of hardcoded paths).

[Do you have issues with this?](https://github.com/Samueru-sama/Ristretto-AppImage-test/releases) (this is using glycin).

WIth that said glycin has a lot of other issues besides the sandbox being totally broken, another is the library is super bloated, we are now working on a ABI compatible replacement: https://github.com/QaidVoid/glycin-ng

[The same Ristretto appimage is 18MiB instead thanks to it](https://github.com/pkgforge-dev/Ristretto-AppImage/releases/tag/0.14.0-1%402026-07-01_1782913770)

---

You should really use archlinux, we now have smaller builds of webkit2gtk4 with `WEBKIT_EXEC_PATH` support and a demo [here](https://github.com/pkgforge-dev/Anylinux-AppImages/releases/tag/demo)

FabianLars:
> [Do you have issues with this?](https://github.com/Samueru-sama/Ristretto-AppImage-test/releases) (this is using glycin).

Nope, starts fine. But then again, building the tauri app on arch also starts fine. I've only seen this issue when building on Ubuntu 2604 and opening it on that same machine. (did not test many distros though)

> You should really use archlinux, we now have smaller builds of webkit2gtk4 with WEBKIT_EXEC_PATH support and a demo [here](https://github.com/pkgforge-dev/Anylinux-AppImages/releases/tag/demo)

Yeah... If it were that simple - we couldn't even force people to use ubuntu (or an older version of it) for the other bundles - also doesn't help that you now have to build the app twice once for deb/rpm once for appimage.
Just trying to have a low barrier of entry so more devs build for linux ~~without losing their minds~~

but whether it was working or not i want to push arch if not for your debloated packages alone.

Samueru-sama:
> > Edit1: issue with bun's single exe output [oven-sh/bun#28281](https://github.com/oven-sh/bun/issues/28281)
> 
> Fixed with [a hack](https://github.com/pkgforge-dev/Anylinux-sharun/blob/3b1a868e14cb1f79392be30205e4cb108e4470ca/src/main.rs#L397-L411) in my fork of sharun.

[kek](https://github.com/oven-sh/bun/issues/28281#issuecomment-5076233390)

Well it will still take a while for projects to update so it was worth it anyway.

crates/tauri-bundler/src/bundle/linux/appimage/sharun.rs
```diff
@@ -130,6 +130,22 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
...
+
+  // make sure to check whether the user has provided their own copy already
+  if settings.appimage().bundle_xdg_open
+    && !app_dir_bin.join("xdg-open").exists()
+    && Path::new("/usr/bin/xdg-open").exists()
+  {
+    bins.push("/usr/bin/xdg-open".into());
+  }
+
```
Samueru-sama: bundling `xdg-open` and `xdg-mime` is a bad idea.

These needs to come from the host always, specially `xdg-open` since people often replace it for their own solutions, I for example use [this](https://github.com/Anomalocaridid/handlr-regex)

FabianLars: yeah that's just leftover from trying to figure out why it doesn't work but fell asleep before reverting (don't ask why i'm not writing the code on the machine i test on). could be another consequence of my messed up test system (broken arch install + it's wsl lol) - edit: note to self: works outside the appimage so we need to double check if xdg-open/xdg-mime works before merging

but good call anyway, i forgot that i meant to remove that from the linuxdeploy plugin for ages...

FabianLars:
anyone else seeing this lately or just me?
```
Write lib.path: /home/fabianlars/_target/debug/bundle/appimage/ceftest.AppDir/lib/lib.path
ln: failed to create hard link '/home/fabianlars/_target/debug/bundle/appimage/ceftest.AppDir/lib/gstreamer-1.0/gst-completion-helper': File exists
       Error [tauri_cli] failed to bundle project: quick-sharun command failed to run.: `failed to run /bin/sh`
```

lucasfernog:
I tried to run some LLM-based tests on VMs and pushed a few fixes, let me know if you spot any issue
here's the report too: https://claude.ai/code/artifact/bd23c67a-8133-415a-be32-e5b600b66558

(i also have a folder with a bunch of logs and screenshots if needed, but appimage ran on a bunch of different systems

Samueru-sama:
> anyone else seeing this lately or just me?
> 
> ```
> Write lib.path: /home/fabianlars/_target/debug/bundle/appimage/ceftest.AppDir/lib/lib.path
> ln: failed to create hard link '/home/fabianlars/_target/debug/bundle/appimage/ceftest.AppDir/lib/gstreamer-1.0/gst-completion-helper': File exists
>        Error [tauri_cli] failed to bundle project: quick-sharun command failed to run.: `failed to run /bin/sh`
> ```

Ubuntu 25 by chance? 

likely [this bug](https://github.com/pkgforge-dev/xenia-canary-AppImage/issues/14#issuecomment-4304874030)

crates/tauri-bundler/src/bundle/linux/appimage/sharun.rs
```diff
@@ -221,9 +221,21 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
...
+  // every terminal-less build - which is every CI run - hangs forever on the
+  // first traced process that does not exit by itself. bash creates the group
+  // either way, so prefer it and fall back to sh where it is missing.
+  let shell = which::which("bash")
+    .map(|p| p.to_string_lossy().into_owned())
+    .unwrap_or_else(|_| "/bin/sh".into());
+
+  // Passing the script to the shell as an argument rather than building a
+  // `-c` string keeps paths containing spaces intact.
+  let mut cmd = Command::new(shell);
```
Samueru-sama, line 238: @lucasfernog This shouldn't be needed anymore: https://github.com/pkgforge-dev/Anylinux-AppImages/pull/853

FabianLars:
> Ubuntu 25 by chance?
likely this bug

Nope that happened on my arch build machine

Edit: happened when building tauri's webkitgtk version, cef version worked fine. Just mentioned because the project in the error is named ceftest

PR #12586, opened 2025-01-31
feat: add `tauriReady` function to core

### LINKED ISSUE #12348: [feat] Add way to check if Tauri's initialization scripts have run
### Describe the problem

Tauri's initialization scripts may not have executed when the webview starts executing user scripts. This can cause Tauri API calls to fail unexpectedly. For example calling `await listen(...)` at a top level will fail if the `__TAURI_INTERNALS__` object has not yet been populated.

I would like to be able to check if Tauri has finished initializing and wait if it has not. I would suggest a new variable like `window.__TAURI_READY__` or `window.__TAURI_INTERNALS__.ready` which would be set to true after all initialization scripts have finished executing. That could be wrapped in an function in the core library like:

```js
function isReady(): bool {
  return window.__TAURI_INTERNALS__?.ready ?? false;
}
```

Additionally, it would be nice to dispatch an event after the ready flag is set like `window.dispatchEvent("tauriReady", new Event())` so that a script can wait for the ready condition without having to poll.

### Alternatives considered

_No response_

### Additional context

See https://discord.com/channels/616186924390023171/1327342177608794113 for discussion.

alexmoon:
Closes #12348.

Adds a `tauriReady()` function to `@tauri-apps/api/core` which returns a promise that resolves when all of Tauri's initialization scripts have finished.

This allows users to work around an issue observed on Windows where the webview is running the page's scripts before the initialization scripts have finished running, leading to errors when attempting to access Tauri APIs before they have been initialized.

I have observed cases where my page's scripts appear to be run between initialization scripts (i.e. some initialization scripts have run but others have not), so this adds a new initialization script to the end of the sequence to set the ready flag and dispatch the event to ensure that all other initialization scripts have finished running before the ready flag is set.


lucasfernog:
> This allows users to work around an issue observed on Windows where the webview is running the page's scripts before the initialization scripts have finished running

FYI we're fixing that bug in tauri 2.5

lucasfernog:
I'll close this since we've fixed the underlying bug https://github.com/tauri-apps/wry/pull/1531

PR #12592, opened 2025-02-01
docs: update devUrl on config schema

Smankusors:
On the tauri.conf.json example from [Configuration Structure](https://v2.tauri.app/reference/config/#configuration-structure), the `devUrl` is still using deprecated value
```
    "devUrl": "../dist",
```

I was confused why it didn't work on my project, it turns out it's not supported on v2 😅 

so I updated the doc to this
```
    "devUrl": "http://localhost:3000",
```

wdyt?


amrbashir (approved):
Thank you

PR #12613, opened 2025-02-04
fix(ci): downgrade string_cache, string_cache_codegen and openssl crates

lucasfernog:


FabianLars:
If you name the branch `v1/msrv-check` the msrv check will run in the PR not just after merge

FabianLars (approved):

PR #12632, opened 2025-02-06
fix: don't eagerly set `ControlFlow::Exit` (fix #8631)

### LINKED ISSUE #8631: [bug] `run_iteration` example busy-loops on Windows
### Describe the bug

I'm getting this output indicating a busy loop:

image

```
Ran Tauri iteration at 4.7094867s
Ran Tauri iteration at 4.7097679s
Ran Tauri iteration at 4.7099947s
Ran Tauri iteration at 4.710248s
Ran Tauri iteration at 4.7104644s
Ran Tauri iteration at 4.7106774s
Ran Tauri iteration at 4.7108921s
Ran Tauri iteration at 4.7111063s
Ran Tauri iteration at 4.7113324s
Ran Tauri iteration at 4.7115689s
Ran Tauri iteration at 4.7118283s
Ran Tauri iteration at 4.7120621s
Ran Tauri iteration at 4.7122877s
Ran Tauri iteration at 4.712554s
Ran Tauri iteration at 4.8543264s
```

### Reproduction

https://github.com/ReactorScram/tauri_repro

Run `cargo tauri dev` or `cd src-tauri; cargo run`. stdout will be filled with constant print statements.

```rust
fn main() {
    let mut app = tauri::Builder::default()
        .invoke_handler(tauri::generate_handler![greet])
        .expect("should be able to build the Tauri app");

    let start_time = std::time::Instant::now();

    loop {
        let iteration = app.run_iteration();
        println!("Ran Tauri iteration at {:?}", start_time.elapsed());
        if iteration.window_count == 0 {
            break;
        }
    }
}
```

### Expected behavior

I expected from #5532 that `run_iteration` would block until the main event loop receives an event, then process it, then return.
Instead it doesn't seem to block, at least on Windows.

This is part of a larger yak shave for me. I'm trying to join a worker thread after Tauri exits, but all these methods have failed:
- `app.exit` doesn't trigger `RunEvent::Exit` (intentional behavior per #4662), so I can't clean up in there
- `run_iteration` busy loops
- No way to inject my own `RunEvent` into the main loop to wake up `run` and check whether we need to clean up
- No known way to just stop the Tauri event loop and let things `Drop` naturally, per #7358
- No way to register something like `on_exit` with `Builder` https://docs.rs/tauri/latest/tauri/struct.Builder.html

My fallback plan is to write my own cleanup function and call that, and have it call `app.exit` at the end. It'll work, but it'll require some comments explaining it.

### Full `tauri info` output

```text
(I'm on Windows 11 Home. I'm not sure why it says Windows 10.)

[✔] Environment
    - OS: Windows 10.0.22621 X64
    ✔ WebView2: 120.0.2210.133
    ✔ MSVC:
    ✔ rustc: 1.75.0 (82e1608df 2023-12-21)
    ✔ cargo: 1.75.0 (1d8b05cdd 2023-11-20)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (environment override by RUSTUP_TOOLCHAIN)
    - node: 20.10.0
    - npm: 10.2.3

[-] Packages
    - tauri [RUST]: 1.5.4
    - tauri-build [RUST]: 1.5.1
    - wry [RUST]: 0.24.7
    - tao [RUST]: 0.16.6
    - tauri-cli [RUST]: 1.5.6
    - @tauri-apps/cli [NPM]: 1.5.6 (outdated, latest: 1.5.9)

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../src
    - devPath: ../src
```

### Stack trace

```text
N/A
```

### Additional context

~~If this is a platform limitation, like maybe it's just impossible to gracefully stop a GUI event loop on Windows, I would understand.~~

~~But if it's not... can I just have an `app.stop` that stops the event loop, so everything can `Drop`?~~

Seems to be a `tao` thing, see later comment

thomaseizinger:
As discussed at length in #8631, `run_iteration` is basically unusable currently because it busy-loops the current thread. This _shouldn't_ be the case though because the underlying `tao::Runtime` allows suspending until an actual event from the OS is generated. However, once `ControlFlow::Exit` has been set, this remains sticky and cannot be unset.

Local testing on Windows confirms that removing this particular snippet fixes the busy-looping behaviour whilst retaining all other window-related functionality.

### DIFF
--- a/.changes/run-iteration-no-busy-loop.md
+++ b/.changes/run-iteration-no-busy-loop.md
@@ -1,0 +1,5 @@
+---
+'tauri-runtime-wry': 'patch:bug'
+---
+
+`Runtime::run_iteration` no longer sets `ControlFlow::Exit` upon `tao::Event::MainEventsCleared`. This allows `run_iteration` to actually suspend and not busy-loop when called in a loop.
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -2791,9 +2791,6 @@ impl<T: UserEvent> Runtime<T> for Wry<T> {
       .event_loop
       .run_return(|event, event_loop, control_flow| {
         *control_flow = ControlFlow::Wait;
-        if let Event::MainEventsCleared = &event {
-          *control_flow = ControlFlow::Exit;
-        }
 
         for p in plugins.lock().unwrap().iter_mut() {
           let prevent_default = p.on_event(


thomaseizinger:
Hmm, this might actually not fix things. I just realised that `run_iteration` now _never_ returns.

thomaseizinger:
Closing in favor of https://github.com/tauri-apps/tauri/pull/12668.

PR #12643, opened 2025-02-07
feat(uninstaller): remove autostart entry during uninstallation

### LINKED ISSUE #12422: [feat] Remove invalid autostart during the uninstallation process
### Describe the problem

在 Windows 11 通过 `tauri-plugin-autostart` 开启自启动后，执行 uninstall.exe 将程序卸载，此时仍然会有一个无效启动项被保留，是否可以将其在卸载过程中一并删除呢？

[AI Translation] When a program is set to autostart on Windows 11 using `tauri-plugin-autostart` and then uninstalled via `uninstall.exe`, an invalid startup entry often persists in the registry. Can this entry be removed during the uninstallation process?

当自启动启用后，注册表 `HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Run` 下会新建一个项，卸载后这个项仍旧会被保留，或许卸载程序可以考虑将此项一并删除？

[AI Translation] When autostart is enabled for a program, a new registry key is created at `HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Run`. Upon uninstallation, this registry key may persist. It would be beneficial if uninstallers were modified to remove this registry key during the uninstallation process.

### Alternatives considered

_No response_

### Additional context

_No response_

animeshchaudhri:
## Summary  
This PR fixes issue [#12422](https://github.com/tauri-apps/tauri/issues/12422) by ensuring that the autostart registry entry at  
`HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Run`  
is removed when the application is uninstalled.  

Previously, uninstalling an application that was set to autostart via `tauri-plugin-autostart` would leave behind an invalid registry entry, causing unintended behavior. This PR modifies the uninstallation process to clean up these entries properly.

## Related PRs  
- This is similar to the fix implemented in [PR #12427](https://github.com/tauri-apps/tauri/pull/12427) for another registry cleanup issue.

## Changes  
- Modify the uninstallation script to check and remove the autostart registry entry.
- Ensure cleanup happens consistently during uninstallation.

## Testing  
- **Reproduced the issue** by enabling autostart and uninstalling the app.
- **Verified the fix** by checking the registry after uninstallation to confirm the entry is removed.

## References  
Closes [#12422](https://github.com/tauri-apps/tauri/issues/12422).


FabianLars:
thanks for the pr. do you think you can do the same for wix/msi ?

ProjectNyaser:
> thanks for the pr. do you think you can do the same for wix/msi ?感谢提交的 PR。你认为你能为 wix/msi 做同样的事情吗？

image

I'm not sure if this would work.. how can I test? @FabianLars

animeshchaudhri:
> > thanks for the pr. do you think you can do the same for wix/msi ?感谢提交的 PR。你认为你能为 wix/msi 做同样的事情吗？
> 
> image
> 
> I'm not sure if this would work.. how can I test? @FabianLars 
> 

I have tested this and it works and I am working on wix/msi now

lucasfernog:
from the WiX docs:
> There is no standard way to remove a single registry value during uninstall 

https://docs.firegiant.com/wix3/xsd/wix/removeregistryvalue/

so i think we're stuck with fixing for NSIS only

FabianLars:
This PR is somewhat wrong as well since we only remove it from the current user but removing hkcu keys of other users is hard and afaik frowned upon.

With how often i have orphaned autostart entries it seems like most programs just go with the "fuck it, not our problem" and i think we should too (this pr is fine though we may want to consider guarding it behind the installMode as well)

lucasfernog (dismissed):

lucasfernog:
> This PR is somewhat wrong as well since we only remove it from the current user but removing hkcu keys of other users is hard and afaik frowned upon.
> 
> With how often i have orphaned autostart entries it seems like most programs just go with the "fuck it, not our problem" and i think we should too (this pr is fine though we may want to consider guarding it behind the installMode as well)

so this only makes sense if the installmode is peruser right? auto-launch seems to try to enable it on local machine too: https://github.com/zzzgydi/auto-launch/blob/6542a3c4033fe1f3a8757dd65daf7e0c872a0756/src/windows.rs#L55

FabianLars:
> so this only makes sense if the installmode is peruser right?

imo yes

> auto-launch seems to try to enable it on local machine too: [zzzgydi/auto-launch@6542a3c/src/windows.rs#L55](https://github.com/zzzgydi/auto-launch/blob/6542a3c4033fe1f3a8757dd65daf7e0c872a0756/src/windows.rs#L55)

which is a bit problematic because (in tauri) whether the app runs with admin or not is unrelated to the install mode.

animeshchaudhri:
@FabianLars  @lucasfernog  Thanks, guys, for fixing the issue! This was my first PR. I’ve been assigned more issues, but nobody answers my questions related to Tauri architecture, even when I ping them on Discord. So, I had to YOLO it.  lol

FabianLars:
we're yolo-ing it as well so this was a good start!

lucasfernog:
> > auto-launch seems to try to enable it on local machine too: [zzzgydi/auto-launch@6542a3c/src/windows.rs#L55](https://github.com/zzzgydi/auto-launch/blob/6542a3c4033fe1f3a8757dd65daf7e0c872a0756/src/windows.rs#L55)
> 
> which is a bit problematic because (in tauri) whether the app runs with admin or not is unrelated to the install mode.

I opened a PR to auto-launch to help us here https://github.com/zzzgydi/auto-launch/pull/21

lucasfernog (approved):

Legend-Master:
> which is a bit problematic because (in tauri) whether the app runs with admin or not is unrelated to the install mode.

That is added in only in the git version, so it's not used in our auto start plugin yet

> I opened a PR to auto-launch to help us here https://github.com/zzzgydi/auto-launch/pull/21

@lucasfernog There's one already https://github.com/zzzgydi/auto-launch/pull/20

Issue #8788, opened 2024-02-05
[bug] New app fails on android version < 8


loikki:
### Describe the bug

Hi,

Thanks for this amazing project!

I tried to create a new empty project with `create-tauri-app` and it works fine on my recent phone (FP4) but crashes on my old table (android 7; Galaxy Tab S2).

If you need anything else, I will be happy to help but I don't have a lot of experience with Android.

### Reproduction

❯ cargo create-tauri-app --alpha
✔ Project name · second
✔ Choose which language to use for your frontend · TypeScript / JavaScript - (pnpm, yarn, npm)
✔ Choose your package manager · npm
✔ Choose your UI template · React - (https://reactjs.org/)
✔ Choose your UI flavor · TypeScript
✔ Would you like to setup the project for mobile as well? · yes

cd second
npm i
// change bundle identifer to com.second.dev
npm run tauri android init
npm run tauri android dev

### Expected behavior

The app should not crash

### Full `tauri info` output

```text
root@ubuntu:~$ npm run tauri info

> tauri-app@0.0.0 tauri
> tauri info


[✔] Environment
    - OS: Manjaro 23.1.3 X64
    ✔ webkit2gtk-4.1: 2.42.4
    ✔ rsvg2: 2.57.1
    ✔ rustc: 1.75.0 (82e1608df 2023-12-21)
    ✔ cargo: 1.75.0 (1d8b05cdd 2023-11-20)
    ✔ rustup: 1.26.0 (2023-11-14)
    ✔ Rust toolchain: stable-x86_64-unknown-linux-gnu (default)
    - node: 21.5.0
    - npm: 10.3.0

[-] Packages
    - tauri [RUST]: 2.0.0-beta.2
    - tauri-build [RUST]: 2.0.0-beta.1
    - wry [RUST]: 0.35.2
    - tao [RUST]: 0.25.0
    - tauri-cli [RUST]: 2.0.0-beta.0
    - @tauri-apps/api [NPM]: 2.0.0-beta.0
    - @tauri-apps/cli [NPM]: 2.0.0-beta.1

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:1420/
    - framework: React
    - bundler: Vite
```

### Stack trace

```text
root@ubuntu:~$ npm run tauri android dev

> tauri-app@0.0.0 tauri
> tauri android dev

[...]

PR #12646, opened 2025-02-07
update(readme): Remove beta tag from android and iOS; Alert that Android 7 is unsupported

ahqsoftwares:
Just a small update to the README to reflect the current status of the project


lucasfernog:
why is android 7 not supported again? i think i missed that discussion

FabianLars:
https://github.com/tauri-apps/tauri/issues/8788

ahqsoftwares:
> why is android 7 not supported again? i think i missed that discussion

Yep, currently its not supported due to #8788, and to add salt to the injury, Google Chrome too dropped Android 7 Nougat

lucasfernog (approved):

Issue #12078, opened 2024-12-31
[feat] cookie


1937650094:
### Describe the problem

When webview is embedded in the instagram website, how do I get all its cookies and set cookies to block all the requests and responses of the webview, just like capturing packets?

### Describe the solution you'd like

When webview is embedded in the instagram website, how do I get all its cookies and set cookies to block all the requests and responses of the webview, just like capturing packets?

### Alternatives considered

When webview is embedded in the instagram website, how do I get all its cookies and set cookies to block all the requests and responses of the webview, just like capturing packets?

### Additional context

When webview is embedded in the instagram website, how do I get all its cookies and set cookies to block all the requests and responses of the webview, just like capturing packets?

FabianLars:
there's no api for this yet. Closing as a duplicate of #5823, #11691 and maybe #4899

Issue #11691, opened 2024-11-14
[feat] Add Set Cookie API to Webview


murnifine:
### Describe the problem

I need an API that can set cookies like the API made by electron, or like the chrome extension API

### Describe the solution you'd like

```rust
    let webview = tauri::webview::WebviewBuilder::new(
        "tauri_main",
        WebviewUrl::External("mywebsite.com".parse().unwrap())
    );

    let res = data;

    webview.set_cookie({
        name: res.name,
        value: res.value,
        domain: res.domain,
        path: res.path,
        expirationDate: res.expirationDate,
        httpOnly: res.httpOnly,
        sameSite: res.sameSite,
        secure: res.secure,
        storeId: e.storeId: res.storeId,
        url: res.url
    });
```

### Additional context

[https://www.electronjs.org/docs/latest/api/cookies#cookiessetdetails](https://www.electronjs.org/docs/latest/api/cookies#cookiessetdetails)
[https://developer.chrome.com/docs/extensions/reference/api/cookies?hl=id#method-set](https://developer.chrome.com/docs/extensions/reference/api/cookies?hl=id#method-set)

Legend-Master:
Added in #13661

Issue #11330, opened 2024-10-12
[feat] Get secure and http only cookies


thewh1teagle:
### Describe the problem

Hi,

I would like to extract all cookies from the Tauri webview. Currently, using `document.cookie` doesn't include HTTP-only or secure cookies. Could you add an option to extract those as well on Windows/macOS/Linux?

Thanks!

### Describe the solution you'd like

For instance when using playwright with `page.cookies()` I can get all the cookies. 
Provide similar function on the webview instance.

### Additional context

You can try checking this by opening https://x.com/, logging in, and then attempting to extract all the cookies. You'll notice that some of them are missing.

charrondev:
I see that https://github.com/tauri-apps/wry/pull/1394 has been merged sometime ago and the latest version of tauri use this version of wry.

Is there a way to get the underlying wry instance from a Webview?

Legend-Master:
Added in #12665

Issue #5823, opened 2022-12-13
[feat] Add a CookieManager to the WebView


chen-gaoyuan:
### Describe the problem

I need an API to get cookies from an external WebView

### Describe the solution you'd like

```ts
const webview = new WebviewWindow('theUniqueLabel', {
  url: 'https://tauri.app/',
});
const allCookie = await webview.cookieManager.getAllCookie();
// ...
```

AlloryDante:
Get & Set

Legend-Master:
#12665 added getters and #13661 added setters, keeping this open for now since they're currently both rust side only

AlloryDante:
> [#12665](https://github.com/tauri-apps/tauri/pull/12665) added getters and [#13661](https://github.com/tauri-apps/tauri/pull/13661) added setters, keeping this open for now since they're currently both rust side only

You're the best

PR #12665, opened 2025-02-09
feat: add `Webview::cookies` and `Webview::cookies_for_url()`

charrondev:
**Closes**

- https://github.com/tauri-apps/tauri/issues/11330

**Related**

- Relevant discussion https://github.com/tauri-apps/tauri/discussions/11655
- Relevant merge into Wry https://github.com/tauri-apps/wry/pull/1394

Related issues not fully handled by this PR:

- https://github.com/tauri-apps/tauri/issues/5823
- https://github.com/tauri-apps/tauri/issues/12078
- https://github.com/tauri-apps/tauri/issues/11691
- https://github.com/tauri-apps/tauri/issues/4899

## Changes

- Added `Webview::cookies_for_url()` and `tauri::Cookie`

## Implementation notes

I'm open to any requested changes here, but here's how I've gone about this:

- Expose `cookie::Cookie` as a re-export from `tauri_runtime::Cookie` and `tauri::Cookie`. Notably I did not re-export the `wry::Cookie` which is a re-export of `cookie::Cookie` since `tauri_runtime` does not have a dependency on `wry`.
- Add `cookies_for_url()` on the webview dispatcher and add an implementation for wry. I used  `WebviewDispatcher::reparent` as my template for a dispatch the takes a parameter and returns a result.
- I chose not to expose `wry::webview::cookies()` as it seemed to have an implementation note that it didn't work on Android at the moment.

## Notes from my testing

Notably the wry support seems to only work when a site served over http (which is what I need it for). In my manual testing there doesn't seem to be a clear story for cookies on the `tauri://` protocol (and there probably shouldn't be?). In any case by linking this PR into my own project I am able to extract the cookies from content served over http/https.

### DIFF
--- a/crates/tauri-runtime/Cargo.toml
+++ b/crates/tauri-runtime/Cargo.toml
@@ -34,6 +34,7 @@ http = "1"
 raw-window-handle = "0.6"
 url = { version = "2" }
 dpi = { version = "0.1", features = ["serde"] }
+cookie = { version = "0.18" }
 
 [target."cfg(windows)".dependencies.windows]
 version = "0.58"
--- a/crates/tauri-runtime/Cargo.toml
+++ b/crates/tauri-runtime/Cargo.toml
@@ -34,7 +34,7 @@ http = "1"
 raw-window-handle = "0.6"
 url = { version = "2" }
 dpi = { version = "0.1", features = ["serde"] }
-# WARNING: cookie::Cookie is re-exported so bumping this is a breaking change
+# WARNING: cookie::Cookie is re-exported so bumping this is a breaking change, documented to be done as a minor bump
 cookie = "0.18"
 
 [target."cfg(windows)".dependencies.windows]
--- a/crates/tauri-runtime/Cargo.toml
+++ b/crates/tauri-runtime/Cargo.toml
@@ -34,7 +34,8 @@ http = "1"
 raw-window-handle = "0.6"
 url = { version = "2" }
 dpi = { version = "0.1", features = ["serde"] }
-cookie = { version = "0.18" }
+# WARNING: cookie::Cookie is re-exported so bumping this is a breaking change
+cookie = "0.18"
 
 [target."cfg(windows)".dependencies.windows]
 version = "0.58"
--- a/crates/tauri/src/webview/mod.rs
+++ b/crates/tauri/src/webview/mod.rs
@@ -13,6 +13,7 @@ use http::HeaderMap;
 use serde::Serialize;
 use tauri_macros::default_runtime;
 pub use tauri_runtime::webview::PageLoadEvent;
+pub use tauri_runtime::Cookie;
 #[cfg(desktop)]
 use tauri_runtime::{
   dpi::{PhysicalPosition, PhysicalSize, Position, Size},
@@ -25,7 +26,6 @@ use tauri_runtime::{
 pub use tauri_utils::config::Color;
 use tauri_utils::config::{BackgroundThrottlingPolicy, WebviewUrl, WindowConfig};
 pub use url::Url;
-pub use tauri_runtime::Cookie;
 
 use crate::{
   app::{UriSchemeResponder, WebviewEvent},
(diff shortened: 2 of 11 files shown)


crates/tauri/src/webview/mod.rs
```diff
@@ -1157,6 +1158,11 @@ impl<R: Runtime> Webview<R> {
     Ok(())
   }
 
+  /// Returns all cookies for a specified URL including HTTP-only and secure cookies.
```
FabianLars: Can you add a note about the scheme requirements here? imo quite important

FabianLars (commented):
thank you!!

> I chose not to expose wry::webview::cookies() as it seemed to have an implementation note that it didn't work on Android at the moment.

It would be fine to do the same in tauri as well (having this function return an empty vec with a platform-specific note)

crates/tauri/src/webview/mod.rs
charrondev: I've added a note. This actually seems more of a bug/runtime limitation than anything. I would imagine this is because the `tauri://` protocol behaves like the the `file://` protocol in browsers and doesn't actually expose a functional cookie store.

charrondev:
I've added the relevant note and the `cookies()` function.

I've also gone in and fixed up my fustfmt and clippy issues.

charrondev:
As I'm testing this more deeply, I'm finding some concerning going. Fetching cookies is occasionally hanging the entire process. I will try to dig in some more.

charrondev:
So after testing a bit more I've narrowed down the source of the hangs. It seems to happen in the MacOS implemenation specifically getting caught up in wry's `wait_for_blocking_operation()`.

https://github.com/tauri-apps/wry/blob/cf181943444bca172d969615471997e48e10b611/src/wkwebview/mod.rs#L1057-L1077

I've gone and made a PR over on wry to resolve it https://github.com/tauri-apps/wry/pull/1486

whispersilk:
Would love to see this feature!

Bubbet:
Where are you meant to use this? I've tried using the pr in a `on_page_load` event and inside a command and in both circumstances it just gets hung or something and never actually returns the cookies. The process doesn't lock up but its like it waiting forever. Edit: I'm also on windows.

AlloryDante:
I cant agree more. We need a system for reading/setting cookies for different URL's....

AlloryDante:
@FabianLars Can you approve this PR?

AlloryDante:
> Where are you meant to use this? I've tried using the pr in a `on_page_load` event and inside a command and in both circumstances it just gets hung or something and never actually returns the cookies. The process doesn't lock up but its like it waiting forever.

I get the same hunging when I do it on windows....

lucasfernog:
the issue on Windows is similar to window creation, which must be done on a separate thread because we need to wait for an async operation on the webview side. see https://github.com/tauri-apps/wry/issues/583
i've updated the docs to mention this behavior

AlloryDante:
> the issue on Windows is similar to window creation, which must be done on a separate thread because we need to wait for an async operation on the webview side. see [tauri-apps/wry#583](https://github.com/tauri-apps/wry/issues/583) i've updated the docs to mention this behavior

Got it. Thank you for the docs and for the example file.

lucasfernog:
the only gotcha here is that we're re-exporting the cookie crate which is on 0.18 meaning we can't update it to 0.19 for instance without it being flagged as a breaking change, i'll at least update our toml for this

crates/tauri-runtime/Cargo.toml
```diff
@@ -34,6 +34,7 @@ http = "1"
 raw-window-handle = "0.6"
 url = { version = "2" }
 dpi = { version = "0.1", features = ["serde"] }
+cookie = { version = "0.18" }
```
lucasfernog:
```suggestion
# WARNING: cookie::Cookie is re-exported so bumping this is a breaking change
cookie = "0.18"
```

lucasfernog (dismissed):

FabianLars:
alternatively we could do the same we do in with_webview and document that this may change between minor versions. Seeing the crate's activity we probably don't have to worry about new 0.x versions in the near future.

AlloryDante:
Finally we will have a way to get the cookies for the websites. If wry allows maybe in the future it would be nice to also have a set cookie method. I know for a fact that I am gonna use it...

Thank you so much for adding support for this @lucasfernog . I know it doesnt seem much but for my use case I am heavly dependant on cookie access and I was actually deciding if it would be better to switch to electron and you added this nice PR.

lucasfernog:
thanks to @charrondev for bringing the bindings here and @amrbashir for working on the wry implementations

PR #12668, opened 2025-02-10
feat: introduce `App::run_return`

thomaseizinger:
The current `App::run_iteration` function is buggy because it busy-loops. To address the use-case of cleaning up resources in the host program, we introduce `App::run_return` which builds on top of `tao`'s `Eventloop::run_return`.

Related: #8631.


crates/tauri/src/app.rs
```diff
@@ -1092,6 +1092,43 @@ impl<R: Runtime> App<R> {
...
+  ///
+  /// std::process::exit(exit_code);
+  /// ```
+  pub fn run_return<F: FnMut(&AppHandle<R>, RunEvent) + 'static>(mut self, mut callback: F) -> i32 {
+    let manager = self.manager.clone();
+    let app_handle = self.handle().clone();
+
+    if !self.ran_setup {
+      if let Err(e) = setup(&mut self) {
+        panic!("Failed to setup app: {e}");
```
thomaseizinger: We could avoid this panic if we would change the method signature to something like `Result<ExitCode, Error>`.

FabianLars: Let's do it then. winit does that as well if we need an argument in favor.

crates/tauri-runtime/src/lib.rs
```diff
@@ -437,6 +437,10 @@ pub trait Runtime<T: UserEvent>: Debug + Sized + 'static {
   #[cfg(desktop)]
   fn run_iteration<F: FnMut(RunEvent<T>) + 'static>(&mut self, callback: F);
 
+  /// Equivalent to [`Runtime::run`] but returns the exit code instead of exiting the process.
+  #[cfg(desktop)]
+  fn run_return<F: FnMut(RunEvent<T>) + 'static>(&mut self, callback: F) -> i32;
```
thomaseizinger: Adding a new, non-default function to a public trait is not semver compatible. Is it okay to bump this to `3.0` or should we try and find a semver-compatible way?

thomaseizinger: Unfortunately, `tauri_runtime`'s version leaks into the public API of `tauri` through e.g. this trait impl: https://github.com/thomaseizinger/tauri/blob/9be493e2cf56e951e5ba1e5956ba46a18854bf7f/crates/tauri/src/app.rs#L144

Therefore, bumping `tauri-runtime` to major would also mean bumping `tauri` to major if we are being strict about semver.

thomaseizinger: I decided to give it a default impl that panics.

thomaseizinger:
Do we want to deprecate `run_iteration` as part of this?

crates/tauri-runtime/src/lib.rs
thomaseizinger: Only just read that we are not following semver with this crate.

.changes/introduce-run-return.md
```diff
@@ -0,0 +1,7 @@
+---
+tauri: 'minor:feat'
+tauri-runtime: 'minor:feat'
+tauri-runtime-wry: 'minor:feat'
+---
+
+Add `App::run_return` function. Contrary to `App::run`, this will **not** exit the thread but instead return the requested exit-code. This allows the host app to perform further cleanup after Tauri has exited.
```
thomaseizinger:
```suggestion
Add `App::run_return` function. Contrary to `App::run`, this will **not** exit the process but instead return the requested exit-code. This allows the host app to perform further cleanup after Tauri has exited.
```

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -2840,13 +2843,13 @@ impl<T: UserEvent> Runtime<T> for Wry<T> {
     let active_tracing_spans = self.context.main_thread.active_tracing_spans.clone();
     let proxy = self.event_loop.create_proxy();
 
-    self.event_loop.run(move |event, event_loop, control_flow| {
+    self.event_loop.run_return(move |e, event_loop, cf| {
```
thomaseizinger: I renamed these variables here so that this stays on one line. That way, the git diff shows nicely that `run_return` is effectively what `run` used to be.

crates/tauri/src/app.rs
```diff
@@ -1092,6 +1092,47 @@ impl<R: Runtime> App<R> {
...
+  ///   })
+  ///   .unwrap();
+  ///
+  /// std::process::exit(exit_code);
+  /// ```
+  #[cfg(desktop)]
+  pub fn run_return<F: FnMut(&AppHandle<R>, RunEvent) + 'static>(
+    mut self,
+    mut callback: F,
+  ) -> std::result::Result<i32, Box<dyn std::error::Error>> {
```
thomaseizinger: I've just used the same error as from the setup callback. It is missing `Send + Sync + 'static` bounds unfortunately but that is already the case and can't be changed now I think.

FabianLars:
> Do we want to deprecate run_iteration as part of this?

imo yes

FabianLars:
iirc run_return also works on android (though probably good to test that first), only iOS should be a problem. I'm wondering whether we should really go for the cfg flag you used or just try to do what winit for example does and just document that it doesn't return ever on iOS (using tao's run instead of run_return for iOS internally). didn't look too much into the code yet so idk how feasible that is.

thomaseizinger:
> iirc run_return also works on android (though probably good to test that first), only iOS should be a problem. I'm wondering whether we should really go for the cfg flag you used or just try to do what winit for example does and just document that it doesn't return ever on iOS (using tao's run instead of run_return for iOS internally). didn't look too much into the code yet so idk how feasible that is.

Removing the cfg is a semver-compatible change so we can always do that later? I'd prefer an incremental approach if possible! :)

`run_iteration` is only exposed on desktop hence why I copied that. (To me, adding `run_return` is a bugfix for `run_iteration`).

Deciding on what the mobile story is here seems like a different problem to me that I'd rather not tackle, also because I don't know the internals of tao and Tauri well enough :)

thomaseizinger:
@FabianLars 

- `App::run_iteration` has been deprecated.
- I had to re-introduce an actual implementation of `Wry::run` because `run_return` in tao is not available on iOS and Android. To deduplicate the code, I extracted a factory fn for the event handler.

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -2830,39 +2830,43 @@ impl<T: UserEvent> Runtime<T> for Wry<T> {
...
+  F: FnMut(RunEvent<T>) + 'static,
+{
+  let windows = self.context.main_thread.windows.clone();
+  let window_id_map = self.context.window_id_map.clone();
+  let web_context = self.context.main_thread.web_context.clone();
+  let plugins = self.context.plugins.clone();
+
+  #[cfg(feature = "tracing")]
+  let active_tracing_spans = self.context.main_thread.active_tracing_spans.clone();
+  let proxy = self.event_loop.create_proxy();
```
WSH032:
```suggestion
  let windows = runtime.context.main_thread.windows.clone();
  let window_id_map = runtime.context.window_id_map.clone();
  let web_context = runtime.context.main_thread.web_context.clone();
  let plugins = runtime.context.plugins.clone();

  #[cfg(feature = "tracing")]
  let active_tracing_spans = runtime.context.main_thread.active_tracing_spans.clone();
  let proxy = runtime.event_loop.create_proxy();
```

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -2830,39 +2830,43 @@ impl<T: UserEvent> Runtime<T> for Wry<T> {
...
+    self
+      .event_loop
+      .run_return(make_event_handler(&self, callback))
+  }
+}
+
+fn make_event_handler<T, F>(
+  runtime: Wry<T>,
+  mut callback: F,
+) -> impl FnMut(Event<'_, T>, &EventLoopWindowTarget<T>, &mut ControlFlow)
```
WSH032:
```suggestion
) -> impl FnMut(Event<'_, Message<T>>, &EventLoopWindowTarget<Message<T>>, &mut ControlFlow)
```

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -2830,39 +2830,43 @@ impl<T: UserEvent> Runtime<T> for Wry<T> {
       });
   }
 
-  fn run<F: FnMut(RunEvent<T>) + 'static>(self, mut callback: F) {
-    let windows = self.context.main_thread.windows.clone();
-    let window_id_map = self.context.window_id_map.clone();
-    let web_context = self.context.main_thread.web_context;
-    let plugins = self.context.plugins.clone();
+  fn run<F: FnMut(RunEvent<T>) + 'static>(self, callback: F) {
+    self.event_loop.run(make_event_handler(&self, callback))
```
WSH032: `event_loop.run` take the partial ownship, so you can't borrow `&self`.

*I don't know why Tao took ownership; maybe it was for the sake of cross-platform API consistency?*

WSH032:
What can be done to push this forward? I really want to see this feature in `tauri 2.3` so that I can finalize the Python interpreter in `pytauri` (the current `run` implementation kills 🐍, and we can't clean up resources at all 😱).

thomaseizinger:
> What can be done to push this forward? I really want to see this feature in `tauri 2.3` so that I can finalize the Python interpreter in `pytauri` (the current `run` implementation kills 🐍, and we can't clean up resources at all 😱).

Unfortunately, the repository settings prevent me from effectively iterating on this because CI has to be approved constantly and running `cargo check` locally doesn't work because some native dependencies aren't satisfied and I couldn't be bothered yet to write a `shell.nix` that sets up fully sets up all dependencies.

crates/tauri-runtime-wry/src/lib.rs
thomaseizinger: The returned closure should not capture the passed reference so it should be fine?

thomaseizinger:
> and I couldn't be bothered yet to write a `shell.nix` that sets up fully sets up all dependencies.

Wasn't as much work as I thought it would be (wuhu AI!) so iterating locally now.

crates/tauri-runtime-wry/src/lib.rs
WSH032: Yep. I'm just curious why `tao`'s `run` takes ownership, as it is just a wrapper around `run_return`, which only requires a reference 🤓

thomaseizinger:
Should be compiling now. Some of the tests also fail on `dev` for me so I can't produce a clean `cargo test` locally.

crates/tauri-runtime-wry/src/lib.rs
thomaseizinger: > Yep. I'm just curious why `tao`'s `run` takes ownership, as it is just a wrapper around `run_return`, which only requires a reference 🤓

iOS doesn't have `run_return` and `run` blocks until the app is quit so I guess it is to have some kind of consistency across platforms.

FabianLars: maybe for legacy reasons. `run` is older than `run_return` (or at least on some platforms, can't remember the timeline exactly) and may have required more ownership. Or it was done without giving it much of a thought since run never returns anyway.

FabianLars: Or maybe it used to be like this in winit for reasons nobody can remember and we just left it as it was when we forked it 🤷

thomaseizinger:
Yey all green!

crates/tauri/src/app.rs
WSH032: I just tried it locally, and `crate::Result<i32>` works for me.

examples/run-return/README.md
WSH032, line 1: forgot update `cargo run --example run-return`

crates/tauri/src/app.rs
```diff
@@ -1092,6 +1092,47 @@ impl<R: Runtime> App<R> {
...
+    let manager = self.manager.clone();
+    let app_handle = self.handle().clone();
+
+    if !self.ran_setup {
+      setup(&mut self)?;
+    }
```
WSH032: Just curious, don't we need to `match event` `Ready/Exit` like in `App::run`?

thomaseizinger: I am not too familiar with this bit and copied what was done for `run_iteration` before.  When is the ready event emitted?

For exit, given that we now return control flow, I don't think we need to match on the Exit event.

crates/tauri-runtime-wry/src/lib.rs
thomaseizinger: I think `App::run_return` taking ownership makes sense. You shouldn't be able to modify `App` after it exited.

crates/tauri/src/app.rs
thomaseizinger: I've updated it to copy what `run` does.

thomaseizinger:
@FabianLars @WSH032 Ready for another review round. I've reverted the change of returning `Result` from `run_return` and I am now copying exactly what `run` does: Calling `setup` when `RuntimeRunEvent::Ready` is emitted. I don't know the details of the lifecycle but sticking to what `run` does seems more correct here.

Refactoring this can be left to a future PR.

WSH032 (approved):
> I've reverted the change of returning Result from run_return

I have no opinion on this. I just think that if `run_return` does return a `Result`, I would prefer it to be `tauri::Result<T>` (`tauri::Error::Setup`) rather than `std::result::Result<T, Box<dyn Error>>`.

---

Regarding the lifecycle, we should look at @FabianLars's opinion.

thomaseizinger:
> > I've reverted the change of returning Result from run_return
> 
> I have no opinion on this. I just think that if `run_return` does return a `Result`, I would prefer it to be `tauri::Result<T>` (`tauri::Error::Setup`) rather than `std::result::Result<T, Box<dyn Error>>`.

I agree. Unfortunately, returning a `Result` is not compatible with running `setup` inside the event-loop unless we refactor `callback` to be able to fail and abort the event-loop.

thomaseizinger:
> unless we refactor `callback` to be able to fail and abort the event-loop.

I think this is possible but haven't looked at the code of all platforms. It is more involved than what I'd like to do at the moment.

FabianLars:
> I don't know the details of the lifecycle but sticking to what run does seems more correct here.

It is more correct* and i'd consider it a bug in run_iteration that it didn't do that as well.
\* "correct" is debatable but there are a few things (mostly on macos) depending on that.

> I agree. Unfortunately, returning a Result is not compatible with running setup inside the event-loop unless we refactor callback to be able to fail and abort the event-loop.

(take this with a grain of salt. i'm on my phone so basically blind) What has the callback to do with that? As far as i can see you'd just return from run_return where rn there's a panic completely ignoring the callback. The event loop should get stopped simply by dropping it (which we do by returning)

I don't really care about the error type we return, i'd just like to get rid of the panic here tbh. If this is not possible, then the function's documentation should mention the panic just in case.

thomaseizinger:
> What has the callback to do with that?

The callback offers no way to cancel the eventloop it is running in so the only control flow we have available is panicking. With some refactoring, I think it would be possible to return a `ControlFlow` or `Result` from the callback which would allow us to thread the value through the event-loop (and cancel it the same way we would when we return the exitcode).

I also agree that the panic here is not pretty and I am keen to get rid of it. I'd like to keep this for a follow-up PR though so this feature here can be shipped.

I'll update the docs.

thomaseizinger:
> I don't really care about the error type we return, i'd just like to get rid of the panic here tbh. If this is not possible, then the function's documentation should mention the panic just in case.

I opened https://github.com/tauri-apps/tauri/issues/12815 because I think there are a few more things to discuss here in trying to get rid of that panic.

thomaseizinger:
@FabianLars I rebased on `dev` to fix the failing workflow. Let me know if there is anything else we need to ship this.

FabianLars:
i just deleted the draft i wrote 2 hours ago that i forgot to send 👏 

tldr: looks good, any improvements in follow up prs will probably result in breaking changes but that's fine 🤷

crates/tauri/src/app.rs
```diff
@@ -1092,6 +1096,57 @@ impl<R: Runtime> App<R> {
     });
   }
 
+  /// Runs the application, returning the exit code to use.
```
FabianLars: can we add a Platform-specific note (see set_device_event_filter above fe) for no mobile support here? A bit more obvious than the cfg.

thomaseizinger:
> tldr: looks good, any improvements in follow up prs will probably result in breaking changes but that's fine 🤷

I thought about this. I think we could still make a backwards-compatible change by returning an exit-code of -1 in case there is an error in the setup and just log it on ERROR or WARN level.

thomaseizinger:
I've been testing this integration in our application and it works quite well: https://github.com/firezone/firezone/pull/7959.

I think it might be worth showing this in a more complex example. The basic idea is:

- Use the app builder to build your app instance
- Grab an `AppHandle`, pass it to whatever your controller thingy is, i.e. your "app" outside of Tauri
- Call `AppHandle::exit` whenever you want to shutdown. In our case, we do this when the controller task exits
- Block the main-thread with `run_return`
- Wait for the controller task to finish _after_ `run_return`
- Return from `main`

In our case, we don't actually use the return value from `run_return` but pass a `Result` back from the controller task and let that determine if we exited successfully or not.

FabianLars:
just a fyi, i'll have to keep this pr open/unmerged for a few more days. I wanted to include it in 2.3.0 but that was just released (i only was told once ci was running 🤷 ) and we'll need to delay minor changes a bit to be able to quickly release new patches if something went wrong in 2.3.0

thomaseizinger:
> just a fyi, i'll have to keep this pr open/unmerged for a few more days. I wanted to include it in 2.3.0 but that was just released (i only was told once ci was running 🤷 ) and we'll need to delay minor changes a bit to be able to quickly release new patches if something went wrong in 2.3.0

Okay no worries :)

Thanks for the reviews! Looking forward to it being released!

cuchaz:
Just wanted to chime in and say this PR fixes the busy main loop problem for me in Linux. Thanks! I'm looking forward to seeing it (maybe) in the 2.4 release.

Though there's a side-effect of not being able to use `features = ["config-toml"]` when pointing the cargo dependency to a git commit. Weird.

cuchaz:
Ok, after using the new `run_return` function for a bit, I'm starting to find some cracks. It looks like when the function exits, Tauri hasn't quite finished cleanup yet? The windows are still visible, for instance. The windows don't disappear from my desktop (Linux, btw) until the process exits. Ideally, after `run_return` exits, the windows would be gone.

After trying a bunch of things to poke Tauri into closing the windows before exiting the event loop, I may have found why the cleanup isn't happening. My best guess is that cleaning up the windows requires sending events, and if we exit the event loop too early, those events never get processed?

Here's some code I've figured out that successfully manages to close the windows before the event loop exits, but it requires giving the event loop some extra time to do ... something before allowing it to stop.

```Rust

fn try_main(args: Args) -> Result<Code,Error> {

  // start the tokio runtime
  tokio::runtime::Builder::new_multi_thread()
    .enable_time()
    .build()
    .wrap_err()
    .context(ctx!("Failed to start tokio runtime"))?
    .block_on(async {

      tauri::async_runtime::set(tokio::runtime::Handle::current());

      let state = AppStateContainer::new(args);

      debug!("Starting tauri ...");

      // run the tauri app
      let exit = tauri::Builder::default()
        .manage(state.clone())
        .invoke_handler(tauri::generate_handler![
          status
        ])
        .setup(|app| setup(app).map_err(Into::into))
        .wrap_err()
        .context(ctx!("Failed to build tauri app"))?
        .run_return({
          let mut exiting = false;
          move |app, event| match event {

            tauri::RunEvent::ExitRequested { api, .. } => {
              
              // if we've already been here before, just let the event loop exit
              if exiting {
                return;
              }
              exiting = true;

              // otherwise, prevent the event loop from exiting
              api.prevent_exit();
              
              // then hide all the windows
              for win in app.webview_windows().values() {
                match win.hide() {
                  Ok(()) => info!("Hid window {}!", win.label()),
                  Err(e) => warn!("Failed to hide window {}: {}", win.label(), e)
                }
              }
              
              // and set a task to call exit again later,
              // to give the event loop time to actually hide the windows
              tokio::task::spawn({
                let app = app.clone();
                async move {
                  debug!("Waiting for event loop to hide things ...");
                  tokio::time::sleep(std::time::Duration::from_millis(200))
                    .await;
                  debug!("Wait complete: burn it down!");
                  app.exit(0)
                  // DANGER WILL ROBINSON: if app.exit() fails internally, it will call std::process::exit() !!!
                }
              });
            }
            
            _ => () // ignore other events
          }
        });

      debug!("Tauri finished, cleaning up state ...");

      // clean up the state
      state.close()
        .await?;

      debug!("Finished");

      Ok(Code::new(exit))
    })
}
```

This code seems to function as a workaround, but it's not ideal, because `app.exit` looks like it could actually call `std::process::exit`. Exiting early would skip any post-tauri cleanup and defeat the whole purpose of `run_return`.

I'm not sure how to use this info to propose a change to the way Tauri works, but the solution here seems to be allowing the event loop to process the event queue before actually exiting. Maybe that can help someone who knows Tauri better to come up with a fix?

thomaseizinger:
> Ideally, after `run_return` exits, the windows would be gone.

In my usecase, I am calling `exit` when the last window got closed so that might be the issue? I am not sure we can solve this in Tauri? Once you call `exit`, the eventloop will stop. It sounds like you are triggering "exit" too early?

thomaseizinger:
I guess we could extend the internal "cleanup before exit" handler perhaps?

cuchaz:
In my case, I'm triggering "exit" by clicking the close icon on the last window. I'm not sure how to make that happen earlier or later. But after the window closes, and before the process can exit, I need to make sure other resources get cleaned up. Like the `state` variable in my example above.

cuchaz:
I tried to poke around in the Tauri code a bit to see if I could find a solution there. Maybe I have something?

This re-implementation of [`Runtime::run_return` for `Wry`](https://github.com/tauri-apps/tauri/pull/12668/files#diff-dd84e40463ad6c84b6c7485e577d8df25832f4064b0e2a0ea2843371f7eab815R2840) seems to work in my case. Is this an ok thing to be doing?
```Rust
#[cfg(desktop)]
fn run_return<F: FnMut(RunEvent<T>) + 'static>(mut self, callback: F) -> i32 {
  use tao::platform::run_return::EventLoopExtRunReturn;

  let event_handler = make_event_handler(&self, callback);

  let exit = self.event_loop.run_return(event_handler);
    
  // after the event loop is ready to exit, run one more iteration to dispatch any cleanup messages
  // otherwise, the windows will become orphaned from the Rust code and remain visible until the process exits
  self.event_loop.run_iteration(callback);
    
  exit
}
```

cuchaz:
I tried poking around the Tauri code a bit to see if I could find a solution. Maybe I have something?

This re-implementation of [`Runtime::run_return` for `Wry`](https://github.com/tauri-apps/tauri/pull/12668/files#diff-dd84e40463ad6c84b6c7485e577d8df25832f4064b0e2a0ea2843371f7eab815R2840) seems to get all the way through cleanup before exiting the fn. Running one more iteration of the event loop seems to do the trick.

```Rust
#[cfg(desktop)]
fn run_return<F: FnMut(RunEvent<T>) + 'static>(mut self, callback: F) -> i32 {
  use tao::platform::run_return::EventLoopExtRunReturn;

  let event_handler = make_event_handler(&self, callback);

  let exit = self.event_loop.run_return(event_handler);

  // after the event loop is ready to exit, run one more iteration to dispatch any cleanup messages
  // otherwise, the windows will become orphaned from the Rust code and remain visible until the process exits
  self.run_iteration(|_| {});

  exit
}
```

This seems to work in my case. Is this an ok thing to do in general? If not, I don't really know why it works, but maybe someone else knows a better way to get the cleanup to finish out.

thomaseizinger:
> This re-implementation of [`Runtime::run_return` for `Wry`](https://github.com/tauri-apps/tauri/pull/12668/files#diff-dd84e40463ad6c84b6c7485e577d8df25832f4064b0e2a0ea2843371f7eab815R2840) seems to get all the way through cleanup before exiting the fn. Running one more iteration of the event loop seems to do the trick.

We want to deprecate `run_iteration` actually because it IMO misuses the underlying `run_return` abstraction.

I am happy for other people to jump in here, I am not gonna get to work on this for a few days.

cuchaz:
I finally found a way to do async state cleanup while Tauri's event loop is still running, so it's no longer necessary to rely on Tauri's internal cleanup to finish successfully. I'll share the workaround here for posterity, in case it might help anyone else:

```Rust

fn main() {

  // start the tokio runtime
  tokio::runtime::Builder::new_multi_thread()
    .build()
    .unwrap()
    .block_on(async {

      tauri::async_runtime::set(tokio::runtime::Handle::current());

      // create our state
      let state = MyState::new();
      
      // run the tauri app
      tauri::Builder::default()
        .manage(state)
        .invoke_handler(...)
        .setup(...)
        .build(...)
        .unwrap()
        .run({
          let mut exiting = false;
          move |app, event| match event {

            tauri::RunEvent::ExitRequested { api, .. } => {

              // if we've already been here before, just let the event loop exit
              if exiting {
                return;
              }
              exiting = true;
              
              // don't exit just yet: still need to cleanup
              api.prevent_exit();

              // start a task to cleanup the state, since we need async
              tokio::task::spawn({
                let app = app.clone();
                async move {
                  
                  // get a reference to our state from Tauri
                  let state = app.state::<MyState>()
                    .deref()
                    .clone();
                  
                  // clean up the state
                  state.close()
                    .await
                    .unwrap();

                  // ask tauri event loop to exit again
                  app.exit(0)
                }
              });
            }
            
            _ => ()
          }
        });
    })
}
```

Error handling and Tauri configuration were omitted, for brevity. This pattern should work on the release version of Tauri too.

thomaseizinger:
In our case, the equivalent of your `close` returns an error for us that I'd like to use to inform the exit code. This is possible but duplicates all error handling because the app can also fail prior to Tauri starting and it would be nice to unify all of that into one fallible function and hence the need for `run_return`.

I am almost certain we can achieve the same thing with some modifications to the internals of `event_loop.run_return`. It should be possible to run one or a few more ticks of the event-loop after receiving the exit event.

.changes/run-return-mobile.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri-runtime": minor:feat
+---
+
+`Runtime::run_return` now must also be implemented on mobile targets.
```
thomaseizinger: I guess this can be omitted because the function is only added in this PR (see other changelog entry?)

thomaseizinger:
So I've tested this current implementation against our app again and I cannot reproduce the problem you are describing @cuchaz. Here is how I am using this: https://github.com/firezone/firezone/blob/8e3558653efdff5c9e31bc6b19d2bc5af7d06848/rust/gui-client/src-tauri/src/client/gui.rs#L116-L342

I think the important part here is:

- Within the `setup` hook, spawn a "Controller"-like task. Your controller is what defines the lifecycle of your application. What we want is to exit the application when the controller exits. (In our case, exiting is triggered by a tray-menu button but that is an implementation detail).
- When the controller task exits, we call `AppHandle::exit`. This is the trigger that will actually shutdown the application.
- Once the controller task is spawned, we send it into a oneshot channel.
- `run_return` calls `prevent_exit()` if (and only if) we don't have an exit-code. Whether or not we have an exit-code differentiates between a manually triggered exit via `AppHandle::exit` and the last window being closed.
- After run-return finishes, we read from our one-shot channel to receive the controller-task handle.
- We wait for the controller task handle to finish to read the return status.

I think the above could also be amended to close the application once the last window is closed **if** you still call `prevent_exit` from within your loop and manually call `AppHandle::exit` with a code. This will allow Tauri to perform enough ticks of the event-loop to always close all windows and clean-up resources and only exit once you are actually sending the `AppHandle::exit` event.

From my perspective, this PR doesn't need any further implementation patches. I think the way it works makes sense. What we might want is a more elaborate example / tutorial to show people, how to implement a graceful exit with Tauri.

@cuchaz Could you try if the following fixes your problem?

```rust
.run_return(move |handle, event| match event {
    tauri::RunEvent::ExitRequested {
        api, code: None, ..
    } => {
        api.prevent_exit();
		handle.exit(0);
    }
    tauri::RunEvent::MenuEvent(event) 
    | tauri::RunEvent::Exit
    | tauri::RunEvent::WindowEvent { .. }
    | tauri::RunEvent::WebviewEvent { .. }
    | tauri::RunEvent::Ready
    | tauri::RunEvent::Resumed
    | tauri::RunEvent::MainEventsCleared
    | tauri::RunEvent::TrayIconEvent(_)
    | _ => {}
});
```

I know it looks a bit silly but effectively, this should allow the event-loop to process the final "close" click of the window but also immediately schedule another trigger of exit (this time with an exit code) which will not be prevented and will exit the app.

.changes/introduce-run-return.md
```diff
@@ -0,0 +1,9 @@
+---
+tauri: 'minor:feat'
+tauri-runtime: 'minor:feat'
+tauri-runtime-wry: 'minor:feat'
+---
+
+Add `App::run_return` function. Contrary to `App::run`, this will **not** exit the process but instead return the requested exit-code. This allows the host app to perform further cleanup after Tauri has exited. `App::run_return` is only available on desktop platforms.
```
lucasfernog:
```suggestion
Add `App::run_return` function. Contrary to `App::run`, this will **not** exit the process but instead return the requested exit-code. This allows the host app to perform further cleanup after Tauri has exited. `App::run_return` is not available on iOS and fallbacks to the regular `App::run` functionality.
```

lucasfernog (approved):

thomaseizinger:
Yey! Thanks everyone for collaborating on this :)

cuchaz:
Wow, I am somehow unable to reproduce the lingering window problem anymore. `run_return` works perfectly now. Exact same code I was using before and everything.

The only difference on my end between then and now is I applied some OS updates and rebooted my computer. Sometimes, my terminal (gnome-terminal) hangs instead of opening a window, but rebooting causes the problem to go away. xterm doesn't suffer from the same problem though. I just assumed that gnome-terminal hanging was unrelated to this Tauri issue, but now I'm not so sure. Maybe there was a bug in my OS's windowing system that was fixed in a recent update? Seems unlikely, but maybe it's possible. This is super weird. I'm not crazy, I promise. =)

Anyway, thanks for all your hard work. I'm looking forward to the next release.

cuchaz:
I was excited to try the new 2.4 release and tragically, the lingering window issue has returned! I figured out why I wasn't able to reproduce the issue in my last message. It's because I forgot my local clone of Tauri had [the fix I described earlier](https://github.com/tauri-apps/tauri/pull/12668#issuecomment-2712543517) applied to it, so the lingering issue didn't happen in that case.

I am able to reliably reproduce this lingering window issue using the latest 2.4 release. [Here's a complete minimal working example](https://codeberg.org/cuchazinteractive/tauri-issue-linger), if you'd like to use it to help with debugging. Since the issue is with windowing systems, it's very likely the issue is specific to the operating system. Here's my OS info:

```
me@machine:~$ uname -a
Linux machine 5.15.0-134-generic #145-Ubuntu SMP Wed Feb 12 20:08:39 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
me@machine:~$ cat /etc/lsb-release 
DISTRIB_ID=LinuxMint
DISTRIB_RELEASE=21
DISTRIB_CODENAME=vanessa
DISTRIB_DESCRIPTION="Linux Mint 21 Vanessa"
```
Since Linux Mint is based on Ubuntu (which is based on Debian), I'd be pretty surprised if the issue wasn't reproducible in Ubuntu (or Debian) too.

The `main.rs` file that reliably reproduces this issue is very simple:
```rust
fn main() {

  println!("Start");

  tauri::Builder::default()
    .expect("Failed to build Tauri app")
    .run_return(|_, _| {});

  println!("Tauri exited, but window is still visible during sleep!");
  std::thread::sleep(std::time::Duration::from_secs(3));

  println!("Finished");
}
```
There are no async tasks in setup, like you mentioned earlier, or nested futures or anything like that. This code is completely sync.

To reproduce the issue, build and run the complete example project from the link above. After the Tauri window appears, click the close button. While the code is `sleep`ing, the window should still be visible, meaning, Tauri hasn't quite finished its cleanup even though `run_return` has exited.

Let me know if you have any other questions. I'll do what I can do help out.

thomaseizinger:
Have you tried the suggestion I made?

Match on `RunEvent::ExitRequested { code: None }`, surpress the exit but call `app_handle.exit(0)` immediately

cuchaz:
Last time I looked at the source code, `app_handle.exit()` has a code path in it that calls the system `exit` function, so it's unsuitable for cleanup handling, since the cleanup might never actually happen. A proper solution to this issue should never rely on an `exit` call.

thomaseizinger:
> Last time I looked at the source code, `app_handle.exit()` has a code path in it that calls the system `exit` function, so it's unsuitable for cleanup handling, since the cleanup might never actually happen. A proper solution to this issue should never rely on an `exit` call.

That is only when it fails! Under normal circumstances, this performs an ordenary shutdown.

thomaseizinger:
See https://github.com/tauri-apps/tauri/blob/dd137283341ce0e6aabfd158d07d77c6feeb920e/crates/tauri/src/app.rs#L530-L537

cuchaz:
And what happens if there's an error? It's not acceptable in my app to skip cleanup even if there's an error in Tauri. Rust gives us the tools to handle these cases correctly. We should use them.

thomaseizinger:
I haven't look at the details but I'd assume this only happens when a channel is closed which I guess never happens in practice?

You'd have to trace the code-path. We should probably remove that exit as well so we can panic in the event handler and force a regular unwind. I didn't try that though.

Issue #11671, opened 2024-11-13
[feat] Expose `WindowEvent::KeyboardInput` in Rust


stijnfrishert:
### Describe the problem

I want to be able to react to key events coming from the application event loop.

I am building a shortcut system for my Desktop app, on macOS. I would like to be able to trigger hotkeys, even if the application doesn't have any windows open. This rules out any js key listener solutions, because they are tied to windows. Plus, this feels like something that should be done over in Rust-land, anyway.

The current recommended solution is to use [rdev](https://crates.io/crates/rdev). This works up to a point (see https://github.com/tauri-apps/tauri/issues/11670), but requires my app to ask for full access to all OS-level keyboard inputs (Accessibility), which is a permission I would rather not depend on.

In reality, this is something that should be solvable by propagating events from the event loop. I scoured through `Tao` a bit and found key event types, so this seems to be within the realm of possibility?

### Describe the solution you'd like

I want to be able to register a key event handler on my app. I envision this would look something like this, but that's of course up for debate:

```rust
tauri::Builder::default()
    .on_key_event(|app, event|{ }))
    .run()
```

I can imagine plug-ins might want to do this too, so maybe we should have a system to register (and unregister) multiple handlers.

### Alternatives considered

- Binding to key event listeners in js, but this can only be done on windows. Besides, imho this is something to be solved on the Rust side, using the application event loop.

- Using `rdev`, but this requires my app to ask for full keyboard monitoring on an OS-level, which is not a permission I want to ask for, really.

  In addition, people have to twiddle with [device_event_filter](https://docs.rs/tauri/latest/tauri/struct.Builder.html#method.device_event_filter), which is neither intuitive nor easy to find.

### Additional context

I've searched through the Discord and issue list, and it seems more people have been requesting this.

amrbashir:
similar to the other issue, your problem may just be solved by our global shortcut plugin, https://tauri.app/plugin/global-shortcut/

FabianLars:
i disagree, app-local shortcuts and/or an equivalent to js key events in rust are a regularly requested features (i was sure there was one on github already before this but i can't find it...)

amrbashir:
> app-local shortcuts and/or an equivalent to js key events in rust

not sure what you mean by this, please explain more

FabianLars:
basically key events that only work when the app is in focus.

[...]

PR #12677, opened 2025-02-10
feat: Expose window keyboard events to Tauri

stijnfrishert:
This PR implements #11671, which exposes keyboard input events for Tao in Tauri on the window-level.

To safeguard against breaking changes in Tao, I've opted to use types from [`keyboard-types`](https://docs.rs/keyboard-types/latest/keyboard_types/) as the public facing API in Tauri, and written converter functions in `tauri-runtime-wry` to move from one domain the the other.

Currently the PR exposes the following attributes:

 - Key state (pressed vs releases)
 - Logical key
 - Physical key
 - Location (Left/Right/etc.)
 - Is repeat?

Especially the logical and physical keys were a hassle, because both enums are alike, but there are subtle differences. I hope I've done the super -> meta conversion correct, and space needed special handling as well.

I have not done any testing, because I'm not sure how to do local Tauri testing.

## Draft Questions

 - How should I handle converting non-exhaustive enums, such as `ElementState` and `KeyLocation`? Right now the converter function [`.expects()`](https://github.com/tauri-apps/tauri/pull/12677/files#diff-d8df86823d3b080899ae980ded7e13c13205c1d4ad7076583ab44a3d4dd13492R8), but I'm not sure that's the most future -proof way of handling things?


amrbashir:
> * How should I handle converting non-exhaustive enums, such as `ElementState` and `KeyLocation`? Right now the converter function [`.expects()`](https://github.com/tauri-apps/tauri/pull/12677/files#diff-d8df86823d3b080899ae980ded7e13c13205c1d4ad7076583ab44a3d4dd13492R8), but I'm not sure that's the most future -proof way of handling things?

instead of returning None in these conversion functions, just remove the Option and add `unimplemented!()` macro call with a helpful message.

stijnfrishert:
@amrbashir done!

lucasfernog:
there's a big problem here: wry replaces the window content view: https://github.com/tauri-apps/wry/blob/0352b7aa8b9d65e42d1bdd38ab2142a78b013304/src/wkwebview/mod.rs#L581
the content view is first set by tao, where it reads the keyboard events: https://github.com/tauri-apps/tao/blob/986f07aa7e031315a1bf314bcfbc0343234be4c7/src/platform_impl/macos/view.rs#L670
what it means is that we lose all keyboard input events on macOS.. and we don't have an easy fix for that, that's why the keyboard events were never exposed

PR #12688, opened 2025-02-12
Update run_benchmark.rs

focustest210:


lucasfernog:
If I understood correctly you mostly made some code cleanup here right?

PR #12701, opened 2025-02-14
feat(tauri): export struct `tauri::ExitRequestApi`

WSH032:
Sorry for opening this PR before creating an issue (because the changes in this PR are very small).

I encountered a similar issue to tauri-apps/plugins-workspace#2161, `ExitRequestApi` seems to be unintentionally [sealed](https://docs.rs/tauri/2.2.5/tauri/enum.RunEvent.html#variant.ExitRequested.field.api).

`pytauri` needs to wrap `struct ExitRequestApi` to implement pyo3 bindings. Can we get this feature into `tauri v2.3`?

---

BTW, if you don't mind, can I derive `Clone` for `ExitRequestApi`?

### DIFF
--- a/.changes/export-ExitRequestApi.md
+++ b/.changes/export-ExitRequestApi.md
@@ -1,0 +1,5 @@
+---
+tauri: 'minor:feat'
+---
+
+Export `struct tauri::ExitRequestApi`.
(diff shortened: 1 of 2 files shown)


FabianLars:
> BTW, if you don't mind, can I derive Clone for ExitRequestApi?

i don't think anything speaks against it but i'd prefer a seperate PR for it just in case.

FabianLars (approved):

WSH032:
> i don't think anything speaks against it but i'd prefer a seperate PR for it just in case.

Thanks, i will open a new one later.

PR #12710, opened 2025-02-15
feat(tauri): derive `Clone` for `tauri::ExitRequestApi`

WSH032:
follow <https://github.com/tauri-apps/tauri/pull/12701#issuecomment-2660966231>

### DIFF
--- a/.changes/clone-ExitRequestApi.md
+++ b/.changes/clone-ExitRequestApi.md
@@ -1,0 +1,5 @@
+---
+tauri: 'minor:feat'
+---
+
+derive `Clone` for `tauri::ExitRequestApi`.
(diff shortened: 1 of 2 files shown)


FabianLars:
Context for reviewers: I asked for this to be a seperate PR because it was weird to me that CloseRequestedApi implemented Clone but ExitRequestedApi did not. Is there a reason i'm missing or was it just an oversight?

lucasfernog:
both CloseRequestedApi and ExitRequestApi must be used immediately in the event loop runner because the decision handler checks if the action was prevented immediately - meaning you can't move it to a separate thread and prevent later. 
i'm not sure why you would need to clone it, but it doesn't hurt to have it.

lucasfernog (approved):

PR #12723, opened 2025-02-17
fix(tauri): deprecate `Manager::unmanage` to fix `use-after-free`

### LINKED ISSUE #12721: [bug] `StateManager`  is unsound
### Describe the bug

```rust
fn foo(app: tauri::App) {
    use tauri::Manager;

    #[derive(Debug)]
    struct Foo(i32);

    app.manage(Foo(0));
    let state = app.state::<Foo>().inner();
    let mut returned_state = app.unmanage::<Foo>().unwrap();
    returned_state.0 = 42;

    println!("{}", state.0); // 👈 encountered a dangling reference (use-after-free)
}
```

While retaining a reference to the state, removing the managed state from the manager will result in a dangling pointer.

### Reproduction

Here is an equivalent implementation, run it with [miri](https://github.com/rust-lang/miri) (miri cannot directly run tauri):

```rust
use std::{
    any::{Any, TypeId},
    cell::UnsafeCell,
    collections::HashMap,
    hash::BuildHasherDefault,
    sync::{Arc, Mutex},
};

///
pub struct State<'r, T: Send + Sync + 'static>(&'r T);

    #[inline(always)]
        self.0
    }
}

impl<T: Send + Sync + 'static> std::ops::Deref for State<'_, T> {
    type Target = T;

    #[inline(always)]
    fn deref(&self) -> &T {
        self.0
    }
}

impl<T: Send + Sync + 'static> Clone for State<'_, T> {
    fn clone(&self) -> Self {
        State(self.0)
    }
}

        self.0 == other.0
    }
}

        f.debug_tuple("State").field(&self.0).finish()
    }
}

#[derive(Default)]
struct IdentHash(u64);

impl std::hash::Hasher for IdentHash {
    fn finish(&self) -> u64 {
        self.0
    }

    fn write(&mut self, bytes: &[u8]) {
        for byte in bytes {
            self.write_u8(*byte);
        }
    }

    fn write_u8(&mut self, i: u8) {
        self.0 = (self.0 << 8) | (i as u64);
    }

    fn write_u64(&mut self, i: u64) {
        self.0 = i;
    }
}

type TypeIdMap = HashMap<TypeId, Box<dyn Any>, BuildHasherDefault<IdentHash>>;

/// The Tauri state manager.
#[derive(Debug)]
pub struct StateManager {
    map: Mutex<UnsafeCell<TypeIdMap>>,
}

// SAFETY: data is accessed behind a lock
unsafe impl Sync for StateManager {}
unsafe impl Send for StateManager {}

impl StateManager {
    pub(crate) fn new() -> Self {
        Self {
            map: Default::default(),
        }
    }

    fn with_map_ref<'a, F: FnOnce(&'a TypeIdMap) -> R, R>(&'a self, f: F) -> R {
        let map = map.get();
        // SAFETY: safe to access since we are holding a lock
        f(unsafe { &*map })
    }

    fn with_map_mut<F: FnOnce(&mut TypeIdMap) -> R, R>(&self, f: F) -> R {
        let mut map = self.map.lock().unwrap();
        let map = map.get_mut();
        f(map)
    }

    pub(crate) fn set<T: Send + Sync + 'static>(&self, state: T) -> bool {
        self.with_map_mut(|map| {
            let already_set = map.contains_key(&type_id);
            if !already_set {
                map.insert(type_id, Box::new(state) as Box<dyn Any>);
            }
            !already_set
        })
    }

    pub(crate) fn unmanage<T: Send + Sync + 'static>(&self) -> Option<T> {
        self.with_map_mut(|map| {
            map.remove(&type_id)
                .and_then(|ptr| ptr.downcast().ok().map(|b| *b))
        })
    }

    /// Gets the state associated with the specified type.
    pub fn get<T: Send + Sync + 'static>(&self) -> State<'_, T> {
        self.try_get()
            .expect("state: get() when given type is not managed")
    }

    /// Gets the state associated with the specified type.
    pub fn try_get<T: Send + Sync + 'static>(&self) -> Option<State<'_, T>> {
        self.with_map_ref(|map| {
            map.get(&TypeId::of::<T>())
                .and_then(|ptr| ptr.downcast_ref::<T>())
                .map(State)
        })
    }
}

fn main() {
    #[derive(Debug)]
    struct Foo(i32);

    let state_manager = Arc::new(StateManager::new());

    state_manager.set(Foo(0)); // manage the state
    let state = state_manager.try_get::<Foo>().unwrap().inner(); // then get the state
    let mut returned_state = state_manager.unmanage::<Foo>().unwrap(); // unmanage the state
    returned_state.0 = 42;

    println!("{}", state.0); // 👈 encountered a dangling reference (use-after-free)
}
```

### Expected behavior

The unsoundness comes from this line: <https://github.com/tauri-apps/tauri/blob/3f680588cd5ada3a44245fd77b5b87d7ee4c478e/crates/tauri/src/state.rs#L123>.

It allows retaining a reference to the Hashmap value after the `MutexGuard` is dropped.

### Full `tauri info` output

```text
rust 1.82 stable
tauri 2.2.5
```

### Stack trace

`cargo +nightly miri run`

```text
error: Undefined Behavior: out-of-bounds pointer use: alloc1105 has been freed, so this pointer is dangling
   --> src/main.rs:152:20
    |
152 |     println!("{}", state.0); // 👈 encountered a dangling reference (use-after-free)
    |                    ^^^^^^^ out-of-bounds pointer use: alloc1105 has been freed, so this pointer is dangling
    |
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
    = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
help: alloc1105 was allocated here:
   --> src/main.rs:111:37
    |
    |                                     ^^^^^^^^^^^^^^^
help: alloc1105 was deallocated here:
   --> src/main.rs:121:62
    |
121 |                 .and_then(|ptr| ptr.downcast().ok().map(|b| *b))
    |                                                              ^
    = note: BACKTRACE (of the first span):
    = note: this error originates in the macro `$crate::format_args_nl` which comes from the expansion of the macro `println` (in Nightly builds, run with -Z macro-backtrace for more info)

note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace

error: aborting due to 1 previous error
```

### Additional context

_No response_

WSH032:
close #12721

As a reference, here is another [implementation that retains `Manager::unmanage`](https://github.com/WSH032/tauri/commit/1ec2554c8e85563ca11fe98e5e1a5b8f529de3f3).
For semantic compatibility, I don't think it's ideal in terms of performance and ergonomics.

I prefer the solution of deprecating `Manager::unmanage`, as the need to remove state seems to be a rare requirement (I personally can't think of a use case, and you can always achieve it with `Mutex` and `Option`).

### DIFF
--- a/.changes/fix-state-manager-unsoundness.md
+++ b/.changes/fix-state-manager-unsoundness.md
@@ -1,0 +1,5 @@
+---
+tauri: 'minor:bug'
+---
+
+Remove and deprecate `Manager::unmanage` to fix `use-after-free` unsoundness, see tauri-apps/tauri#12721 for details.
--- a/.changes/fix-state-manager-unsoundness.md
+++ b/.changes/fix-state-manager-unsoundness.md
@@ -2,4 +2,4 @@
 tauri: 'minor:bug'
 ---
 
-Remove and deprecate `Manager::unmanage` to fix `use-after-free` unsoundness, see tauri-apps/tauri#12721 for details.
+Deprecate `Manager::unmanage` to fix `use-after-free` unsoundness, see tauri-apps/tauri#12721 for details.
--- a/crates/tauri/src/lib.rs
+++ b/crates/tauri/src/lib.rs
@@ -714,23 +714,28 @@ pub trait Manager<R: Runtime>: sealed::ManagerBase<R> {
   ///
   /// <div class="warning">
   ///
-  /// To fix [tauri-apps/tauri#12721] `use-after-free`` unsoundness,
-  /// this method is removed and becomes a no-op (Always returns `None`).
+  /// This method is *UNSAFE* and calling it will cause previously obtained references through
+  /// [Manager::state] and [State::inner] to become dangling references.
   ///
-  /// If you really want to unmanage a state, use [std::sync::Mutex] and [Option::take] to wrap the state.
+  /// It is currently deprecated and may be removed in the future.
+  ///
+  /// If you really want to unmanage a state, use [std::sync::Mutex] and [Option::take] to wrap the state instead.
+  ///
+  /// See [tauri-apps/tauri#12721] for more information.
   ///
   /// [tauri-apps/tauri#12721]: https://github.com/tauri-apps/tauri/issues/12721
   ///
   /// </div>
   #[deprecated(
     since = "2.3.0",
-    note = "This method is removed and becomes a no-op (Always returns `None`)."
+    note = "This method is unsafe, since it can cause dangling references."
   )]
   fn unmanage<T>(&self) -> Option<T>
   where
     T: Send + Sync + 'static,
   {
-    None
+    // The caller decides to break the safety here, then OK, just let it go.
+    unsafe { self.manager().state().unmanage() }
   }
 
   /// Retrieves the managed state for the type `T`.
--- a/crates/tauri/src/lib.rs
+++ b/crates/tauri/src/lib.rs
@@ -711,11 +711,26 @@ pub trait Manager<R: Runtime>: sealed::ManagerBase<R> {
   }
 
   /// Removes the state managed by the application for T. Returns the state if it was actually removed.
+  ///
+  /// <div class="warning">
+  ///
+  /// To fix [tauri-apps/tauri#12721] `use-after-free`` unsoundness,
+  /// this method is removed and becomes a no-op (Always returns `None`).
+  ///
+  /// If you really want to unmanage a state, use [std::sync::Mutex] and [Option::take] to wrap the state.
+  ///
+  /// [tauri-apps/tauri#12721]: https://github.com/tauri-apps/tauri/issues/12721
+  ///
+  /// </div>
+  #[deprecated(
+    since = "2.3.0",
+    note = "This method is removed and becomes a no-op (Always returns `None`)."
+  )]
   fn unmanage<T>(&self) -> Option<T>
   where
     T: Send + Sync + 'static,
   {
-    self.manager().state().unmanage()
+    None
   }
 
   /// Retrieves the managed state for the type `T`.
--- a/crates/tauri/src/state.rs
+++ b/crates/tauri/src/state.rs
@@ -4,7 +4,6 @@
 
 use std::{
   any::{Any, TypeId},
-  cell::UnsafeCell,
   collections::HashMap,
   hash::BuildHasherDefault,
   sync::Mutex,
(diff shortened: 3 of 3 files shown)


FabianLars:
I totally get the sentiment here but completely removing it like this is a bit awkward considering that this api was added on user request. I've never dealt with such issues though so idk. would love to hear some more opinions here.

p.s. to be clear, i'm 100% with you on the deprecation itself, just _very_ unsure about making it noop

WSH032:
> just very unsure about making it noop

maybe add a new `unsafe fn unsafe_unmanage` ? and let `unmanage` as alias of `unsafe_unmanage` ?

amrbashir:
how about using an Arc internally instead of a box?

WSH032:
> how about using an Arc internally instead of a box?

Do you mean replacing `Box` with `Arc` based on this PR, or providing another implementation using `Arc` to mitigate the `use-after-free` issue?

amrbashir:
same implementation as now, but instead of using a Box, we use Arc and on each `.get()` we clone the arc and increase its strong counter, this means even if the user `unmanage` a state, they would still have valid values until all Arcs are dropped.

WSH032:
Then the implementation of `State::inner` would be broken. We can choose only one between `unmanager` and `State::inner`.

[ref impl](https://github.com/WSH032/tauri/commit/1ec2554c8e85563ca11fe98e5e1a5b8f529de3f3)

amrbashir:
I left a few comments on that implementation but it would be better to just add it here so we can discuss it better

WSH032:
follow <https://github.com/tauri-apps/tauri/pull/12745#discussion_r1962852299>

---

> at this point, I would go with #12723 for now, but instead of completely removing, we just document its pitfall and maybe print a warning in debug builds.

@amrbashir Do you mean to keep the original unsafe behavior of `unmanage`? And what about `#[deprecated]`?

amrbashir:
> > at this point, I would go with #12723 for now, but instead of completely removing, we just document its pitfall and maybe print a warning in debug builds.
> 
> @amrbashir Do you mean to keep the original unsafe behavior of `unmanage`? And what about `#[deprecated]`?

yeah keep the original unsafe behavior for now, keep `#[deprecated]` that explains why this is unsafe, and note that it maybe removed in the future.

WSH032:
ping @amrbashir

WSH032:
`rust 1.85` seems add more clippy lint ruels

amrbashir (approved):

sftse:
Soundness is excluded from the stability guarantees of Rust, as was the breaking change in the stdlib of making `fn set_var` unsafe because there was no other way. Misstating soundness is not optional.

FabianLars:
> Soundness is excluded from the stability guarantees of Rust

got a source for that? i do remember something like that but can't find anything rn.

> as was the breaking change in the stdlib of making fn set_var unsafe because there was no other way

except that this change is "opt-in" by being tied to edition 2024 (meaning in edition 2021 it's not an unsafe fn).

> Misstating soundness is not optional.

I don't understand what that means, including the phrasing before your edit.

sftse:
This isn't tied to edition, but to stdlib version. Upgrading the version will result in breakage, it was just that the edition boundary was as good as any for such an exceptional change. Retaining the 2021 edition in the `Cargo.toml` will not prevent breakage.

I decided to phrase it more diplomatically, but a function that in its public api is declared as safe is misstating its soundness if it can be used to obtain UB.

FabianLars:
> This isn't tied to edition, but to stdlib version. Upgrading the version will result in breakage, it was just that the edition boundary was as good as any for such an exceptional change. Retaining the 2021 edition in the Cargo.toml will not prevent breakage.

assuming we're still talking about just set_var then that doesn't seem to be true. I use 1.86 in a edition 2021 project and set_var is not an unsafe fn. If i change the edition it starts complaining. Literally just tested it. Apologies if i misunderstood the stdlib version part.

> I decided to phrase it more diplomatically, but a function that in its public api is declared as safe is misstating its soundness if it can be used to obtain UB.

Ah got it. Yeah, that's fair and i'd agree

sftse:
Excuse my mistake, I never checked and was going by the documentation that states it is unsafe regardless of version. That is a bit surprising, thanks for surfacing that.

The current API seems like it could remain safe, I'm working on a PR.

PR #12745, opened 2025-02-20
fix(tauri): fix `use-after-free` unsoundness for state manager

### LINKED ISSUE #12721: [bug] `StateManager`  is unsound

WSH032:
close #12721

another impl to #12723

### DIFF
--- a/crates/tauri/src/state.rs
+++ b/crates/tauri/src/state.rs
@@ -4,10 +4,11 @@
 
 use std::{
   any::{Any, TypeId},
-  cell::UnsafeCell,
   collections::HashMap,
   hash::BuildHasherDefault,
-  sync::Mutex,
+  marker::PhantomData,
+  ops::Deref,
+  sync::{Arc, Mutex},
 };
 
 use crate::{
@@ -15,33 +16,62 @@ use crate::{
   Runtime,
 };
 
+type Wrapper<T> = Arc<T>;
+
 /// A guard for a state value.
 ///
 /// See [`Manager::manage`](`crate::Manager::manage`) for usage examples.
-pub struct State<'r, T: Send + Sync + 'static>(&'r T);
+pub struct State<'r, T: Send + Sync + 'static>(
+  Wrapper<T>,
+  // 👇 TODO: just for 2.x semantic compatibility, we can't remove `'r` for now
+  PhantomData<&'r ()>,
+);
 
 impl<'r, T: Send + Sync + 'static> State<'r, T> {
   /// Retrieve a borrow to the underlying value with a lifetime of `'r`.
   /// Using this method is typically unnecessary as `State` implements
   /// [`std::ops::Deref`] with a [`std::ops::Deref::Target`] of `T`.
+  ///
+  /// <div class="warning">
+  ///
+  /// To fix [tauri-apps/tauri#12721], the internal structure of [State] changed in version `2.3`.
+  /// So This method is equivalent to [Self::into_inner] and [Arc::into_raw] now, which means it may cause memory leaks.
+  /// Please refer to [[Arc::into_raw]] for related limitations.
+  ///
+  /// [tauri-apps/tauri#12721]: https://github.com/tauri-apps/tauri/issues/12721
+  ///
+  /// </div>
   #[inline(always)]
+  #[deprecated(
+    since = "2.3.0",
+    note = "will cause memory leak, use `into_inner` or `Deref::deref` instead"
+  )]
   pub fn inner(&self) -> &'r T {
+    let ptr = Arc::into_raw(self.0.clone());
+    // SAFETY: this ptr is valid, because we just created it;
+    // and the lifetime is 'static (it's leaked), so it's safe to return it.
+    unsafe { &*ptr }
+  }
+
+  /// Retrieve the inner value of the [State].
+  /// For `2.0` semantic compatibility, we cannot directly return [Arc<T>] in [crate::Manager::state].
+  pub fn into_inner(self) -> Arc<T> {
     self.0
   }
 }
 
-impl<T: Send + Sync + 'static> std::ops::Deref for State<'_, T> {
+impl<T: Send + Sync + 'static> Deref for State<'_, T> {
   type Target = T;
 
   #[inline(always)]
   fn deref(&self) -> &T {
-    self.0
+    Deref::deref(&self.0)
   }
 }
 
 impl<T: Send + Sync + 'static> Clone for State<'_, T> {
   fn clone(&self) -> Self {
-    State(self.0)
+    State(Wrapper::clone(&self.0), PhantomData)
   }
 }
 
@@ -97,12 +127,13 @@ impl std::hash::Hasher for IdentHash {
   }
 }
 
-type TypeIdMap = HashMap<TypeId, Box<dyn Any>, BuildHasherDefault<IdentHash>>;
+/// The `key` must equal to `value.type_id()`, see the safety doc in methods of [StateManager] for details.
+type TypeIdMap = HashMap<TypeId, Box<dyn Any + Sync + Send>, BuildHasherDefault<IdentHash>>;
 
 /// The Tauri state manager.
 #[derive(Debug)]
 pub struct StateManager {
-  map: Mutex<UnsafeCell<TypeIdMap>>,
+  map: Mutex<TypeIdMap>,
 }
 
 // SAFETY: data is accessed behind a lock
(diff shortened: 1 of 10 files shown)


WSH032:
ping @amrbashir

crates/tauri/src/state.rs
```diff
@@ -4,44 +4,74 @@
...
+  #[deprecated(
+    since = "2.3.0",
+    note = "will cause memory leak, use `into_inner` or `Deref::deref` instead"
+  )]
   pub fn inner(&self) -> &'r T {
+    let ptr = Arc::into_raw(self.0.clone());
+    // SAFETY: this ptr is valid, because we just created it;
+    // and the lifetime is 'static (it's leaked), so it's safe to return it.
+    unsafe { &*ptr }
+  }
```
amrbashir, line 54: a simple fix for this would be 
```rs
 pub fn inner(&'r self) -> &'r T {
   self.0.as_ref()
 }
```

In normal cases, this should be a breaking change but since this fixes a memory leak, it should be fine

crates/tauri/src/state.rs
```diff
@@ -4,44 +4,74 @@
...
+  ops::Deref,
+  sync::{Arc, Mutex},
 };
 
 use crate::{
   ipc::{CommandArg, CommandItem, InvokeError},
   Runtime,
 };
 
+type Wrapper<T> = Arc<T>;
```
amrbashir, line 19: This just makes it confusing, let's remove and use `Arc` directly

crates/tauri/src/state.rs
```diff
@@ -116,37 +147,32 @@ impl StateManager {
...
-    })
+    let mut map = self.map.lock().unwrap();
+    let type_id = TypeId::of::<Wrapper<T>>();
+    let already_set = map.contains_key(&type_id);
+    if !already_set {
+      map.insert(
+        type_id,
+        // SAFETY: keep the type of the key is the same as the type of the value，
+        // see following methods for details.
+        Box::new(Wrapper::new(state)) as Box<dyn Any + Sync + Send>,
```
amrbashir, line 159: we could remove `Box` and use `Arc` directly, should avoid a level of indirection

amrbashir (changes requested):

crates/tauri/src/state.rs
WSH032: I personally believe that most people do not need the `unmanage` feature (it seems that tauri itself does not use it also); if users really need `unmanage`, they can implement it themselves using `Mutex` and `Option::take`. This way, users who do not need mutability (`unmanage`) do not have to pay the performance cost (use `Arc` instead of `Box`, and `Arc::clone` for every `get`).

So, does tauri really want to introduce a BREAKING CHANGE in `2.x` for this? Moreover, this BREAKING CHANGE would make `pub fn inner(&'r self) -> &'r T` quite useless (almost equivalent to removing it).

amrbashir: I assume you mean `app.state().inner()` use-case won't work? in that case I guess your solution is correct, and `deref` should be used instead.

WSH032: Here is my use case for `fn inner`: <https://github.com/WSH032/pytauri/blob/47fe27d6f67c6d490ce753ac7976eb4b8a43cbdd/crates/tauri-plugin-pytauri/src/lib.rs#L103-L109>.

With `fn inner<'a>(&'a self) -> &'r T`, I can return a reference of another type without needing to continue holding `State`.

If we really want to proceed as mentioned above, we might need to provide a method similar to [`MutexGuard::map`](https://doc.rust-lang.org/std/sync/struct.MutexGuard.html#method.map) to help convert `State<'_, T>` to `State<'_, U>`.

amrbashir: at this point, I would go with https://github.com/tauri-apps/tauri/pull/12723 for now, but instead of completely removing, we just document its pitfall and maybe print a warning in debug builds.

then in v3, we rework the API and introduce `.map` and remove `inner` in favor of `deref`

PR #12759, opened 2025-02-21
fix(bundler): bundle resources to correct path during RPM bundling when resources are specified as a map

nekename:
My app's resources were specified as `"resources": { "target/plugins/": "plugins/" }`. However, a user on Fedora reported some functionality working incorrectly, and I was able to reproduce it immediately on openSUSE Tumbleweed, which left me confused as this had been working for many other users, so why could I reproduce it with no steps? I realised it was an issue with the bundled .rpm file, which a quick look at the archives confirmed.

RPM left, DEB right
image

Looking at the code that produces these bundles, the debian bundler called Settings::copy_resources, which iterated over Resource structs, instead of Strings like the RPM bundler was doing. I modified the RPM bundler to also iterate over Resources in the same fashion. It seems safe to remove the call to resource_relpath as it appears the appropriate logic is already being done in tauri-utils [here](https://github.com/tauri-apps/tauri/blob/d7b998fe71eca4d5471d73900f7694c043a17256/crates/tauri-utils/src/resources.rs#L239).

Some tests in other parts of the project were failing, but `cargo test --package tauri-bundler` seems to be OK (there aren't any tests for this part afaict though). Speaking of tests, I haven't tested this to be working as I really don't want to swap out all my installed tooling for a locally built one, but I thought submitting a PR might get this in faster than an issue, so hopefully you'll be able to test it.


nekename:
Thanks for forcing me to sign my commits, I would never have gotten around to setting that up otherwise

nekename:
The failing lint check is not in my code.

nekename:
@FabianLars?

FabianLars:
that's me

FabianLars (approved):

nekename:
Yay

PR #12760, opened 2025-02-21
chore: align `.gitignore` of android template with docs

typed-sigterm:
Add my description here...

### DIFF
--- a/crates/tauri-cli/templates/mobile/android/.gitignore
+++ b/crates/tauri-cli/templates/mobile/android/.gitignore
@@ -17,3 +17,4 @@ key.properties
 
 /.tauri
 /tauri.settings.gradle
+/keystore.properties


crates/tauri-cli/templates/mobile/android/.gitignore
```diff
@@ -16,4 +16,5 @@ local.properties
 key.properties
 
 /.tauri
-/tauri.settings.gradle
\ No newline at end of file
+/tauri.settings.gradle
+/keystore.properties
```
lucasfernog, line 20: this filename isn't part of the template; it's just part of the documentation (meaning you could also use a different name if you do not follow the docs 100%)

typed-sigterm: Got it. My original thought is just to reduce the probability of newbies making mistakes :)

PR #12786, opened 2025-02-23
fix(linux): Add missing RPM signature

kandrelczyk:
Current version of CLI does not generate signatures for RPM files. I don't think there's any reason to exclude RPM so I guess it was simply missed. I've tested it locally and signature is generated correctly.

### DIFF
--- a/.changes/rpm-signature.md
+++ b/.changes/rpm-signature.md
@@ -1,0 +1,5 @@
+---
+'tauri-cli': 'patch:enhance'
+---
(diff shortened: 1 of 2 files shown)


FabianLars:
thanks for the pr! could you also add a small changefile similar to https://github.com/tauri-apps/tauri/blob/dev/.changes/nsis-ukranian.md ?

FabianLars (approved):

PR #12802, opened 2025-02-24
chore(deps): avoid duplicate crates

rex4539:


FabianLars:
i appreciate the PR but i'm really not a fan of this change :/

i think for a few dependencies that we actually use in many places (like serde) would make sense but not for every single one of them (you missed a few btw)

PR #12806, opened 2025-02-25
docs: improve documentation around incognito and data store

Simon-Laux:
basically https://github.com/tauri-apps/wry/pull/1499 but for tauri.

---

Small improvements to the documentation:
- document that icognito uses the nonPersistant data store on MacOS.
- on `with_data_store_identifier` documentation point to the incognito mode to use the non persistant store.

Why?
Because I was a bit confused about this at fist and want to make things more clear for others.

I don't think this needs a changes file or version bump.

### DIFF
--- a/crates/tauri/src/webview/mod.rs
+++ b/crates/tauri/src/webview/mod.rs
@@ -728,7 +728,10 @@ fn main() {
   ///
   ///  ## Platform-specific:
   ///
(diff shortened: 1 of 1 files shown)


amrbashir (approved):

amrbashir:
Thank you

PR #12810, opened 2025-02-25
build: bump Android `minSdk` to API level 24

daniel-mader:
When building a Tauri plugin for mobile, the following lint error occurs:

```
/android/.tauri/tauri-api/src/main/java/app/tauri/PathPlugin.kt:37: Error: Call requires API level 24 (current min is 21): android.content.ContextWrapper#getDataDir [NewApi]
          resolvePath(invoke, activity.dataDir.absolutePath)
                                       ~~~~~~~
```

Since a call to `dataDir` would most likely panic on Android API level <24, bumping `minSdk` seems like the only viable option.

### DIFF
--- a/.changes/bump-android-minsdk.md
+++ b/.changes/bump-android-minsdk.md
@@ -1,0 +1,5 @@
+---
+tauri: 'patch:deps'
+---
+
+Bump the minimum supported Android SDK version to 24.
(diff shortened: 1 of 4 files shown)


FabianLars:
> Since a call to dataDir would most likely panic on Android API level <24, bumping minSdk seems like the only viable option.

Not necessarily. Typically in these situations there are similar apis that work in lower api levels as well. After a quick google search it seems like there's also something like `context.getApplicationInfo().dataDir` though we'd have to check if that really matches the path.

lucasfernog:
looks like we have a way to keep the sdk 21 requirement (we don't want to break that part yet) as Fabian-Lars suggested.

see #12870

PR #12817, opened 2025-02-26
enhance(windows): disable our in-client resizing for undecorated window with shadows

amrbashir:
ref: https://github.com/tauri-apps/tao/pull/1052

### DIFF
--- a/.changes/undecorated-resizing-windows-outside.md
+++ b/.changes/undecorated-resizing-windows-outside.md
@@ -1,0 +1,5 @@
+---
+"tauri": "patch:enhance"
+---
+
+On Windows, undecorated window with shadows, now have native resize handles outside of the window client area.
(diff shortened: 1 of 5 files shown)


Legend-Master (dismissed):

Legend-Master:
Note that I did find a very small issue about the resize handle direction on the top right corner, which I don't really know what is causing it, probably some went off by one or so

https://github.com/user-attachments/assets/e5ad7ccc-2908-41c2-b1fd-2ce146b08f06

amrbashir:
@Legend-Master this should be fixed now, it happened because we were still doing full hit testing which meant the few pixels at top-left and top-right are considered a resize handle inside the client area.

The fix is to skip doing hit testing altogether for undecorated window with shadows and consider it as top only border, because we only care for the top resize handle in that case, and is also great for performance

Legend-Master (approved):

PR #12830, opened 2025-02-26
docs: copy over important missing docs from wry that the custom uri schemes are implemented differently on windows and android

Simon-Laux:
Took me quite a while to figure this out, this should help others to understand it faster than me.


amrbashir (approved):
thank you

PR #12833, opened 2025-02-27
fix(deps): os webview not gated in wry feature

Legend-Master:
Those 2 dependencies were only used in `PlatformWebview` which is gated with `wry` feature

### DIFF
--- a/crates/tauri/Cargo.toml
+++ b/crates/tauri/Cargo.toml
@@ -95,7 +95,7 @@ tray-icon = { version = "0.20", default-features = false, features = [
 # linux
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd"))'.dependencies]
 gtk = { version = "0.18", features = ["v3_24"] }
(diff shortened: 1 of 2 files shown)


crates/tauri/Cargo.toml
```diff
@@ -120,9 +120,13 @@ window-vibrancy = "0.6"
...
 # windows
 [target."cfg(windows)".dependencies]
-webview2-com = "0.36"
+webview2-com = { version = "0.36", optional = true }
 window-vibrancy = "0.6"
-windows = { version = "0.60", features = ["Win32_Foundation"] }
+windows = { version = "0.60", features = [
+  "Win32_Foundation",
+  "Win32_UI",
+  "Win32_UI_WindowsAndMessaging",
```
FabianLars, line 128: what's that for?

Legend-Master: We used it in https://github.com/tauri-apps/tauri/blob/4bffc326ea57c1f9301699d5237a29bee8a5c58d/crates/tauri/src/app.rs#L1910

`webview2-com` used to enable those 2 features so we didn't have those feature enabled on `windows` which we should

FabianLars (approved):

PR #12839, opened 2025-02-27
docs: Update wording from #12830

FabianLars:
This primarily addresses the incorrect urls (missing localhost).


FabianLars:
actually, i think i'm gaslighting myself here. are we actually enforcing it in tauri? we "expect" it at least in convertFileSrc 🤔

amrbashir (approved):

PR #12884, opened 2025-03-03
fix: allow double `--` in permissions as command can have double `__`

oscartbeaumont:
A command in Tauri can technically be called `Something__Else` (as it's a valid Rust identifier) which would map to `allow-Something--Else` in the `default.toml` permissions file, however the permissions file rejects the double `--` so in practice it's impossible to grant the permission to run a command with a `__` in it's name.


.changes/change-pr-12884.md
```diff
@@ -0,0 +1,18 @@
...
+"tauri-runtime": patch
+"tauri-runtime-wry": patch
+"tauri-codegen": patch
+"tauri-macros": patch
+"tauri-plugin": patch
+"tauri-build": patch
+"tauri": patch
+"@tauri-apps/cli": patch
+"tauri-cli": patch
+"tauri-driver": patch
```
FabianLars:
```suggestion
"tauri-utils": "patch:bug"
```

lucasfernog (approved):

Issue #12820, opened 2025-02-26
[bug] Huge increase in app size by upgrading from Tauri V1 to V2


Squitch1:
### Describe the bug

Migrating a project from Tauri V1, especially a pretty old version like the 1.2.1 to v2 cause a huge increase in the final binary size, passing roughly from 3MiB to 6.2Mib when all optimization flags are set.

Here is the result of running `cargo-bloat` on a standard release build:

* on V1:
```
File  .text     Size Crate
13.4%  33.3% 854.5KiB std
 5.4%  13.5% 346.5KiB tauri
 3.5%   8.7% 224.6KiB tokio
 2.5%   6.3% 161.2KiB tess
 2.1%   5.1% 131.2KiB serde_json
 1.6%   4.1% 105.0KiB tao
 1.5%   3.7%  94.1KiB tauri_runtime_wry
 1.0%   2.5%  65.4KiB brotli_decompressor
 0.7%   1.8%  47.1KiB serde
 0.7%   1.7%  44.3KiB gtk
 0.7%   1.7%  42.6KiB regex_lite
 0.6%   1.6%  40.5KiB cfb
 0.6%   1.5%  39.6KiB url
 0.5%   1.3%  32.6KiB crossbeam_channel
 0.4%   1.1%  27.7KiB portable_pty
 0.4%   1.0%  25.9KiB infer
 0.4%   1.0%  24.6KiB wry
 0.4%   0.9%  23.3KiB http
 0.3%   0.9%  22.0KiB idna
 0.3%   0.8%  20.5KiB glib
 2.7%   6.7% 172.8KiB And 51 more crates. Use -n N to show more.
40.3% 100.0%   2.5MiB .text section size, the file size is 6.2MiB
```
* on V2:
```
File  .text     Size Crate
10.6%  23.0%   1.0MiB tauri
10.6%  22.8%   1.0MiB std
 7.0%  15.2% 706.7KiB tokio
 5.4%  11.7% 545.0KiB tess
 1.6%   3.5% 165.0KiB tauri_runtime_wry
 1.5%   3.2% 149.4KiB serde_json
 0.7%   1.6%  73.1KiB x11_dl
 0.7%   1.4%  65.5KiB muda
 0.6%   1.3%  58.5KiB tauri_plugin_fs
 0.6%   1.2%  56.5KiB tao
 0.5%   1.1%  50.4KiB serde
 0.5%   1.0%  47.9KiB url
[...]

Issue #4670, opened 2022-07-14
[feat] bindings for different languages/runtimes


InfernalAzazel:
### Describe the problem

Python is used in almost every industry, easy to use

When will python be supported as a backend language, I am very much looking forward to that day.

### Describe the solution you'd like

python as backend language

Arkitect-z:
Can’t wait to use python machine learning on Tauri !

link89:
I am wondering if `nanomsg` can be used as a message bus between `wry` and `Python` process or thread. Theoretically this method could be language agnostic.

InfernalAzazel:
Expect the feat to bind Python for 3 months, day 74

ghost:
yes python coders literally need this kind of stuff

InfernalAzazel:
There was a problem with the translation last time,
I've been following for 104 days
I'm waiting to support python
Come in every once in a while to see the progress
I don't know if it can be used in 2023?

FabianLars:
> I don't know if it can be used in 2023?

2023 at the earliest, i wouldn't be surprised if it takes even longer. There was no work done for it yet.

InfernalAzazel:
Artificial intelligence is very popular now, and I have been looking forward to the arrival of binding Python

mchao123:
Hey there! I was wondering if it would be possible to add TypeScript/JavaScript as supported backend development languages. A lot of us Electron users would really appreciate this feature. It would be awesome if we could choose to use either Bun.js, Deno, or Node.js as the runtime. It can be quite challenging for frontend developers to switch to other languages, so having this option would be a game-changer. I understand that it's a big ask, but I think it would make a lot of us really happy. Thank you for considering this request.

mchao123:
Here's a suggestion to share with Tauri developers: Consider leveraging Rust to provide APIs for other backend languages. For example, you can bind Node.js and provide IPC communication APIs and window objects. The remaining tasks such as file I/O and network requests can be handled by Node.js. This way, future development teams only need to update and maintain the IPC and window APIs.

JonasKruckenberg:
This is something we're incredibly excited and eager to implement and it's something we have had on our roadmap (https://tauri.app underneath the features section) for ever.
However, it is a monstrous, herculean effort to actually pull of. Each language has its own concepts, primitives and idiomatic ways of doing things and we're already stretched thin as-is.
[...]

PR #12890, opened 2025-03-04
feat: add a new option to remove unused commands

Legend-Master:
Reference: https://github.com/tauri-apps/tauri/issues/12820#issuecomment-2694819634

Introducing a new config option `build > removeUnusedCommands`  to let the tauri-cli pass in an environment variable to let the build script of `tauri-plugin` and `tauri` to generate a list of allowed commands from the ACL, this will then be used by the `generate_handler` macro to remove unused commands

Note: this won't be accounting for dynamically added ACLs so make sure to check it when using this

Some results:

- Release build of the `helloworld` example in the repo gone from 3.92 MB to 1.91 MB on Windows (it doesn't use any plugin commands, so the improvement is quite big, also the basic executable contains the `image` dependency, it's around 2.67 MB if that's removed, but still a great improvement!)


WSH032:
Hi @Legend-Master (*I stumbled upon this PR* 😇).

Can we implement this feature in `build.rs` or as cargo features instead of `tauri-cli`?

Sometimes I only use `cargo build --features tauri/custom-protocol` instead of `tauri build`.

ref: <https://github.com/tauri-apps/tauri/issues/4670#issuecomment-2417589615>

> Additionally, the regular Tauri development process heavily relies on Tauri-Cli to generate executables. However, for Pyo3 bindings, dynamic libraries (.so/.dll/.dylib) are required, so we cannot use Tauri-Cli and must use cargo build instead.

Legend-Master:
I have tried it first before this and unfortunately since we can't run the build script before all the macros in plugins, it's not really viable

I could expose the command for generating the used commands though so that you can run it first and then exposing it to the environment manually

WSH032:
> I could expose the command for generating the used commands though so that you can run it first and then exposing it to the environment manually.

In some cases, we don't even have `tauri-cli` 😥, and `cargo` is the only tool available.

> since we can't run the build script before all the macros in plugins

Can we set an environment variable pointing to the `capabilities` files and then parse the allowed commands in the `generate_handler!` macro within the plugins?

This way, we wouldn't need to rely on `tauri-cli` to generate `allowed-commands.json`.

When I want to enable this feature, I will manually set this environment variable pointing to the `capabilities`.

(*I'm not very familiar with this part of Tauri, so my idea might be wrong*)

Legend-Master:
> In some cases, we don't even have tauri-cli 😥, and cargo is the only tool available.

I would also like to make those features not relying on the tauri cli if possible

> Can we set an environment variable pointing to the capabilities files and then parse the allowed commands in the generate_handler! macro within the plugins?

Actually good idea, I'll give it a try when I got time, but instead of doing it in the macro, I think in plugins's build script should be better? (I'll see when I get there)

My current approach is not that great as it relies on the build scripts to generate the acl manifest first, so definitely something I would like to improve as well

> (I'm not very familiar with this part of Tauri, so my idea might be wrong)

To be honest, I'm not very familiar with acl and macro related things either, so any suggestion would be much appreciated

crates/tauri-cli/src/build.rs
```diff
@@ -55,6 +55,10 @@ pub struct Options {
   /// Skip prompting for values
   #[clap(long, env = "CI")]
   pub ci: bool,
+  /// Try to remove unused commands registered from plugins base on the ACL list,
+  /// this feature requires tauri-plugins 2.1 and tauri 2.4
```
WSH032:
```suggestion
  /// Try to remove unused commands registered from plugins base on the ACL list,
  /// this feature requires tauri-plugins 2.1 and tauri 2.4.
  /// Alternatively, you can set the environment variable `REMOVE_UNUSED_COMMANDS` to the absolute path of the `src-tauri` directory for the Cargo build,
  /// which is equivalent to enabling this feature for the tauri-cli build.
```

Can we mention the usage of the `REMOVE_UNUSED_COMMANDS` environment variable in the documentation?

Legend-Master: I'm not sure if we should do it here as this is for the `tauri-cli build --help` output, but I think we can mention it in the docs https://tauri.app/concept/size/ once this feature lands

WSH032: > but I think we can mention it in the docs <https://tauri.app/concept/size/> once this feature lands

I agree with you.

I didn't find other documentation sections in the PR, which led me to mistakenly believe that `tauri-cli build --help` would serve as the documentation. 😂

Legend-Master: Honestly I feel like this is an implementation detail we should hide from normal users, or at least we don't want people to rely on it as this might change in the future, so probably a collapsed section in the docs that documents how the internal works would be most fitting

WSH032: I think I can accept that. Or, mark it as unstable in the documentation to inform users that "enabling this feature through environment variables" might be removed at any time?

Legend-Master: I think we just document this in the website as an internal thing, and state something like we did for the other unstable things

> /// # Stability
> /// This is the output of the [`generate_context`] macro, and is not considered part of the stable API.
> /// Unless you know what you are doing and are prepared for this type to have breaking changes, do not create it yourself.

WSH032: LGTM 👍

lucasfernog:
@Legend-Master you deserve the contributor of the year award for being able to add a feature to the ACL :joy:

lucasfernog:
looks like this option is something you want enabled on every dev/build command run, why not move it to the configuration instead?

Legend-Master:
@lucasfernog The way we currently implementing this for the app's manifest will remove all the app commands if the app has no application ACLs

I'll try to fix it later

lucasfernog:
@Legend-Master i tried to fix it - let me know what you think of this approach. kinda complicated since we also need to generate handlers for inline plugins.

Legend-Master:
> @Legend-Master i tried to fix it - let me know what you think of this approach. kinda complicated since we also need to generate handlers for inline plugins.

Well, it still removes the app commands if there're inline plugins but no app manifest right now (`app-menu:default` seems to be also included through `default` in the `__app-acl__` entry)

Also I would like to push these thing to the json file as we're already reading and writing from there

Legend-Master:
> Well, it still removes the app commands if there're inline plugins but no app manifest right now (`app-menu:default` seems to be also included through `default` in the `__app-acl__` entry)

I get it now, it's because of UNC paths...

> Also I would like to push these thing to the json file as we're already reading and writing from there

I have gone ahead and did that

>  // when there's no app manifest / inlined plugins we should skip generating the allowed command list
>  // which then skips the removal of unused commands

I have also moved this to inside of `generate_allowed_commands` which will help removing the file to reflect potential edits

@lucasfernog This should be working now, tested locally with the api example with some configurations and all seems to work fine

crates/tauri-macros/src/command/handler.rs
```diff
@@ -31,33 +31,108 @@ pub struct Handler {
...
+  let mut unused_commands = Vec::new();
+
+  let command_prefix = if let Some(plugin_name) = &plugin_name {
+    format!("plugin:{plugin_name}|")
+  } else {
+    // All application commands are allowed if we don't have an application ACL
+    //
+    // note that inline plugins without the #![plugin()] attribute would also get to this check
+    // which means inline plugins must have an app manifest to get proper unused command removal
+    if !allowed_commands.has_app_acl {
```
lucasfernog: i would move this if out of this block as it's a side effect while we're just getting the command prefix

Legend-Master:
Actually, we should not skip this when `permissions_map` is empty in `generate_allowed_commands`, that could just mean we didn't specify any commands so we need to remove everything

Legend-Master:
> Actually, we should not skip this when `permissions_map` is empty in `generate_allowed_commands`, that could just mean we didn't specify any commands so we need to remove everything

Sorry, it was right, it's about the permissions, I thought it was capabilities somehow 😂

lucasfernog (dismissed):
looks good! :fire:

crates/tauri-utils/src/config.rs
```diff
@@ -2670,6 +2670,15 @@ pub struct BuildConfig {
   pub before_bundle_command: Option<HookCommand>,
   /// Features passed to `cargo` commands.
   pub features: Option<Vec<String>>,
+  /// Try to remove unused commands registered from plugins base on the ACL list during `tauri build`,
+  /// the way it works is that tauri-cli will read this and set the environment variables for the build script and macros,
+  /// and they'll try to get all the allowed commands and remove the rest
+  ///
+  /// Note:
+  ///   - This won't be accounting for dynamically added ACLs so make sure to check it when using this
+  ///   - This feature requires tauri-plugins 2.1 and tauri 2.4
```
FabianLars:
```suggestion
  ///   - This feature requires tauri-plugin 2.1 and tauri 2.4
```
(regen schema)

lucasfernog: noooooo my beautiful green CI!!!!!

Legend-Master:
I feel like we shouldn't  apply it for dev as that would likely cause rebuilds due to different environment variables from rust analyzer, and I don't think this is useful in dev anyways

1111mp:
@Legend-Master Your work is great, thank you. ❤️ 

I have a question I hope you can answer: Is this determined by whether the user explicitly provides the required permissions for the command when using the plugin to determine whether it needs to be added to `allowed-commands.json`. And is it possible to remove unused `command` in the project itself without using plugins?

Thank you again for your contribution.

Legend-Master:
> @Legend-Master Your work is great, thank you. ❤️

❤️

> Is this determined by whether the user explicitly provides the required permissions for the command

Yes, but it should already be provided by your capability files that you don't need any additional configs

> using the plugin to determine whether it needs to be added to `allowed-commands.json`

I'm not sure I get what `plugin` do you mean by that, but this feature would work with or without the plugins (by plugin I mean for example `tauri-plugin-updater`), the build script (both `tauri`, `tauri-plugin`, `tauri-build` crates) will generate the `allowed-commands.json` file based on your capability files separately for each crate

crates/tauri-utils/src/config/parse.rs
```diff
@@ -174,28 +174,25 @@ pub fn is_configuration_file(target: Target, path: &Path) -> bool {
...
-/// Returns the raw configuration and the platform config path, if any.
+/// Returns the raw configuration and used config paths.
 ///
 /// [JSON Merge Patch (RFC 7396)]: https://datatracker.ietf.org/doc/html/rfc7396.
-pub fn read_from(
-  target: Target,
-  root_dir: PathBuf,
-) -> Result<(Value, Option<PathBuf>), ConfigError> {
-  let mut config: Value = parse_value(target, root_dir.join("tauri.conf.json"))?.0;
+pub fn read_from(target: Target, root_dir: &Path) -> Result<(Value, Vec<PathBuf>), ConfigError> {
```
WSH032, line 180: @Legend-Master Does this crate not follow semantic versioning? Changing `root_dir: PathBuf` to `root_dir: &Path` is a breaking change. My library depends on both `tauri` and `tauri-utils`, and this change caused a compilation failure when upgrading to `tauri v2.4`.

Legend-Master: > Does this crate not follow semantic versioning?

No, But I think we should still note that in the release note (honestly we rarely did it for `tauri-utils` and `tauri-runtime`, but maybe we should as people started using them)

> https://docs.rs/tauri-utils/latest/tauri_utils/config/index.html#stability
> This is a core functionality that is not considered part of the stable API. If you use it, note that it may include breaking changes in the future.

WSH032: > These items are intended to be non-breaking from a de/serialization standpoint only. Using and modifying existing config values will try to avoid breaking changes, but they are free to add fields in the future - causing breaking changes for creating and full destructuring.

This documentation misled me into thinking that only [`Config`](https://docs.rs/tauri-utils/latest/tauri_utils/config/struct.Config.html) is unstable (I mistakenly thought other APIs were stable 😂)

---

I am currently using these APIs to [dynamically load `tauri.conf.json` at runtime](https://github.com/pytauri/pytauri/blob/1102329a8b02ab7a871e44915efa9d8e60202347/python/pytauri-wheel/src/lib.rs#L180-L195). I want to know if it is necessary to [pin the version of `tauri-utils`](https://github.com/pytauri/pytauri/blob/1102329a8b02ab7a871e44915efa9d8e60202347/python/pytauri-wheel/Cargo.toml#L33-L37)?

Legend-Master: > This documentation misled me into thinking that only [`Config`](https://docs.rs/tauri-utils/latest/tauri_utils/config/struct.Config.html) is unstable (I mistakenly thought other APIs were stable 😂)

I'm not familiar with this to be honest, not sure if other things are stable or we just didn't put the stability notice at the right place

> I am currently using these APIs to [dynamically load `tauri.conf.json` at runtime](https://github.com/pytauri/pytauri/blob/1102329a8b02ab7a871e44915efa9d8e60202347/python/pytauri-wheel/src/lib.rs#L180-L195). I want to know if it is necessary to [pin the version of `tauri-utils`](https://github.com/pytauri/pytauri/blob/1102329a8b02ab7a871e44915efa9d8e60202347/python/pytauri-wheel/Cargo.toml#L33-L37)?

I would say yes

WSH032: > if it is necessary to pin the version of tauri-utils

> I would say yes

Thanks! It's okay to me.

PR #12893, opened 2025-03-05
feat(core): add `Manager::run_on_main_thread/run_on_main_thread_return`, remove old explicit methods

amrbashir:

### DIFF
--- a/crates/tauri/src/menu/mod.rs
+++ b/crates/tauri/src/menu/mod.rs
@@ -95,6 +95,8 @@ macro_rules! gen_wrappers {
 
       impl<R: Runtime> Drop for $inner<R> {
         fn drop(&mut self) {
+          use $crate::Manager;
+
           let inner = self.inner.take();
           // SAFETY: inner was created on main thread and is being dropped on main thread
           let inner = $crate::UnsafeSend(inner);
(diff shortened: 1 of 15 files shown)


amrbashir:
Just opening so I don't forget, this should wait for v3

Legend-Master:
As we're here, perhaps we can also get rid of `shared_app_impl!`? Rust analyzer basically don't work when writing things inside of it

Also maybe `WebviewWindow` and `WebviewWindowBuilder`? We can make a trait `Webview` and `Window` and implement things there so we don't have to duplicate things every time we add something in it

If you guys think it's a good idea, I'll put it on https://github.com/orgs/tauri-apps/projects/32 so at least we don't permanently forget about it

amrbashir:
Yeah, I have been wanting to do that for quite some time but don't have the time just yet. This would need to have `tauri::prelude` module with it.

lucasfernog:
`run_on_main_thread_blocking` would be a name that matches with our dialog plugin APIs

crates/tauri/src/menu/mod.rs
```diff
@@ -120,6 +102,45 @@ macro_rules! gen_wrappers {
         }
       }
 
+      impl<R: $crate::Runtime> $type<R> {
+        /// Returns a unique identifier associated with this menu item.
+        pub fn id(&self) -> &MenuId {
+        self.0.inner.id()
```
Legend-Master, line 108:
```suggestion
          self.0.inner.id()
```

Legend-Master (approved):

Issue #12909, opened 2025-03-06
[bug] zoom plugin doesn't work as expected on linux with swedish keyboard


Daedaluz:
### Describe the bug

`Ctrl -` zooms out
`Ctrl 0` resets zoom
`Ctrl +` does nothing

Also, is the ctrl + mousewheel also supposed to work?
I found [this issue](https://github.com/tauri-apps/wry/issues/388) but it looks like it was closed as completed.

Mousewheel zoom does not work either.

### Reproduction

Quite hard to give any instructions on how to reproduce anywhere, but:

Enable `"zoomHotkeysEnabled": true` in tauri.conf.json
Add "core:webview:allow-set-webview-zoom" to capabilities

Test zoom with `ctrl +`

### Expected behavior

`Ctrl +` should zoom in

### Full `tauri info` output

```text
Info GET https://docs.rs/crate/tauri/
    Info Request { method: GET, uri: https://docs.rs/crate/tauri/, version: HTTP/1.1, headers: {"accept-encoding": "gzip", "user-agent": "ureq/3.0.2", "accept": "*/*", "host": "docs.rs"} }
    Info Response { status: 308, version: HTTP/1.1, headers: {"content-length": "0", "connection": "keep-alive", "server": "nginx/1.14.0 (Ubuntu)", "date": "Thu, 06 Mar 2025 14:16:52 GMT", "location": "/crate/tauri", "<NOTICE>": "7 HEADERS ARE REDACTED"} }
    Info Redirect (308 Permanent Redirect): GET https://docs.rs/crate/tauri
    Info GET https://docs.rs/crate/tauri
    Info Request { method: GET, uri: https://docs.rs/crate/tauri, version: HTTP/1.1, headers: {"accept-encoding": "gzip", "user-agent": "ureq/3.0.2", "accept": "*/*", "host": "docs.rs"} }
    Info Response { status: 302, version: HTTP/1.1, headers: {"content-length": "0", "connection": "keep-alive", "server": "nginx/1.14.0 (Ubuntu)", "date": "Thu, 06 Mar 2025 03:41:24 GMT", "location": "/crate/tauri/latest", "<NOTICE>": "7 HEADERS ARE REDACTED"} }
    Info Redirect (302 Found): GET https://docs.rs/crate/tauri/latest
    Info GET https://docs.rs/crate/tauri/latest
[...]

PR #12911, opened 2025-03-06
fix(core): add ctrl '+' key as zoom in

Daedaluz:
Fix for https://github.com/tauri-apps/tauri/issues/12909

On scandinavian keyboards, the + key is usually used for zooming.

This adds `ctrl +` as an alternative to zoom in.


lucasfernog (approved):

lucasfernog:
nice catch, thanks!

PR #12958, opened 2025-03-12
feat(tauri-build): set CompanyName for Windows

Enigmatrix:
Set `CompanyName` for the produced .exe files for Windows. Follows the [documentation](https://v2.tauri.app/reference/config/#publisher) and uses the second element on `bundleId` if `publisher` is not set.

### DIFF
--- a/crates/tauri-build/src/lib.rs
+++ b/crates/tauri-build/src/lib.rs
@@ -623,6 +623,17 @@ pub fn try_build(attributes: Attributes) -> Result<()> {
       res.set("ProductName", product_name);
     }
 
+    let company_name = config.bundle.publisher.unwrap_or_else(|| {
+      config
+        .identifier
+        .split('.')
+        .nth(1)
+        .unwrap_or(&config.identifier)
+        .to_string()
+    });
+
+    res.set("CompanyName", &company_name);
+
     let file_description = config
       .product_name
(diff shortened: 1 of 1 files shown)


amrbashir:
Why not set the Publisher field instead? Why should we fallback, is sitting it necessary?

Enigmatrix:
CompanyName (File Version Info, see [winresource](https://github.com/BenjaminRi/winresource/blob/5c4fce39b369600b89c37e14a5101d1d82268417/lib.rs#L277)) and Publisher (for MSI) seem to be different. Seems like its not set during Tauri builds, which this PR fixes. Letting it be set makes it appear different in Startup Apps, for example:
image

compared to without this change:
image

amrbashir:
So sitting it is not necessary and should be left to the user to decide. If they want, they should set the `publisher` field, otherwise we don't fallback and keep it empty.

amrbashir:
This PR is good, I am saying we just use the `publisher` field without fallback.

Enigmatrix:
Oh, I see. I was following the documentation here:

image

and I saw that other usages of the publisher did the same thing.

amrbashir:
I see, I don't know why we decided on that but let's go with it for the sake of consistency.

amrbashir (approved):

amrbashir:
Thank you

PR #12969, opened 2025-03-13
feat(tauri): add `fn Context::add_capability`

### LINKED ISSUE #12968: [feat] Dynamically Add `Capability` to `Context`
### Describe the problem

I want to compile Tauri into a dynamic library and dynamically set `Context::config`, `Assets`, and `RuntimeCapability` at runtime, so that pytauri can get rid of the dependency on the Rust compiler. (pytauri/pytauri#84)

Here is my current implementation: pytauri/pytauri#108 (*Tauri team doesn't need to know these details, just providing more context*)

I have referred to the Tauri source code and written a [custom `Context` implementation to set `config` and `Assets` dynamically](https://github.com/pytauri/pytauri/blob/bf61cd18ffa03fb46febb48589d9255c82dcf3dc/python/pytauri-wheel/src/lib.rs#L46-L106).

But currently, I have encountered a problem. Tauri only provides `Manager::add_capability` to dynamically add `capability` at runtime, while I hope to achieve this in `Context` as well, so that pytauri can automatically [load `capability` from `tauri.conf.json` and `capabilities/`](https://github.com/tauri-apps/tauri/blob/339a075e33292dab67766d56a8b988e46640f490/crates/tauri-codegen/src/context.rs#L388-L416).

Add a `Context::add_capability` method, which will be similar to `Manager::add_capability`.

### Alternatives considered

Stabilize [`Context::runtime_authority_mut`](https://github.com/tauri-apps/tauri/blob/339a075e33292dab67766d56a8b988e46640f490/crates/tauri/src/lib.rs#L482-L491)

### Additional context

_No response_

WSH032:
close #12968

### DIFF
--- a/.changes/feat-context-add-capability.md
+++ b/.changes/feat-context-add-capability.md
@@ -1,0 +1,5 @@
+---
+"tauri": "minor:feat"
+---
+
+Added `fn Context::add_capability` to allow adding capabilities in the context.
--- a/crates/tauri/src/lib.rs
+++ b/crates/tauri/src/lib.rs
@@ -490,6 +490,16 @@ impl<R: Runtime> Context<R> {
     &mut self.runtime_authority
   }
 
+  /// Adds a capability to the context.
+  ///
+  /// See [Manager::add_capability] for more information.
+  //
+  // TODO: Once we stabilize [Context::runtime_authority_mut], we can deprecate this method.
+  #[inline(always)]
+  pub fn add_capability(&mut self, capability: impl RuntimeCapability) -> crate::Result<()> {
+    self.runtime_authority.add_capability(capability)
+  }
+
   /// Create a new [`Context`] from the minimal required items.
   #[inline(always)]
   #[allow(clippy::too_many_arguments)]


Legend-Master:
I'm not familiar with this part of the code, but I feel like adding this function doesn't really make sense, like if `runtime_authority_mut` is not stable, why would `add_capability` be stable 😂, honestly, the entire `Context` struct is not stable anyways, it doesn't really matter if you just use `runtime_authority_mut` in my opinion

WSH032:
> it doesn't really matter if you just use runtime_authority_mut in my opinion

Fortunately, in `pytauri-wheel`, we can control the version of Tauri, so I'm not too worried about `Context` API breaking changes at the moment.

That said, `runtime_authority_mut` is currently sufficient for my (pytauri) use case. However, before proceeding, I wanted to see the Tauri team's thoughts on #12968.

Because if `runtime_authority_mut` is silently removed without a similar replacement API like `add_capability`, I will almost certainly create a feature request similar to #12968 again. 😂

Legend-Master:
Personally, I don't think anything inside `Context` are considered stable, and if you're concerned about `runtime_authority_mut` getting removed one day, it's the same with `add_capability`, we'll try not to do that though

WSH032:
> and if you're concerned about runtime_authority_mut getting removed one day, it's the same with add_capability, we'll try not to do that though

My main concern is that `runtime_authority_mut` is an even more unstable API within the unstable `Context` (it doesn't even appear in the documentation), so its removal might be entirely unconscious (as the Tauri team might not be aware of its use cases).

Alternatively, I hope to at least remove the `#[doc(hidden)]` attribute from `runtime_authority_mut`, even if keeping:

```rust
/// # Stability
///
/// This API is unstable.
```

lucasfernog:
The `runtime_authority_mut` was mostly added for testing environments and it was marked as doc(hidden) before add_capability was introduced. I agree we could improve its documentation instead of adding this helper function.

WSH032:
`#[doc(hidden)]` feels like a private API used for internal purposes, which could even be removed in a `patch` version; whereas an documented `unstable` API should at least maintain stability within a `minor` version (*at least that's what I would do*).

> I agree we could improve its documentation instead of adding this helper function

TBH, I'm not a fan of `Context::add_capability` either. I would prefer `runtime_authority_mut` to stabilize (or a similar API) rather than adding and then deprecating `Context::add_capability`.

Anyway, it seems we have reached a consensus on rejecting this PR, so I will close it now. In `pytauri-wheel`, I plan to temporarily use `runtime_authority_mut` until #12968 brings a better solution.

Issue #3198, opened 2022-01-12
[bug] `__cmd__perform_request` is defined multiple times


Borber:
### Describe the bug

the name `__cmd__perform_request` is defined multiple times `__cmd__perform_request` must be defined only once in the macro namespace of this  with `#[tauri::command]`

### Reproduction

![image](https://user-images.githubusercontent.com/30563826/149076878-11e37809-c9fd-4033-9940-554da9b23ad9.png)

```rust
#![cfg_attr(
  all(not(debug_assertions), target_os = "windows"),
  windows_subsystem = "windows"
)]

use serde::Deserialize;

#[derive(Debug, Deserialize)]
pub struct RequestBody {
  id: i32,
  name: String,
}

#[tauri::command]
pub fn perform_request(endpoint: String, body: RequestBody) -> String {
  println!("{} {:?}", endpoint, body);
  "message response".into()
}

fn main() {
  tauri::Builder::default()
    .invoke_handler(tauri::generate_handler![
      perform_request
    ])
    .run(tauri::generate_context!())
    .expect("error while running tauri application");
}

```

### Platform and versions

```shell
Operating System - Windows, version 10.0.19044 X64
Webview2 - 97.0.1072.55
Visual Studio Build Tools - Not installed

Node.js environment
  Node.js - 17.3.1
  @tauri-apps/cli - 1.0.0-beta.10
  @tauri-apps/api - 1.0.0-beta.8

Global packages
  npm - 8.3.0
  pnpm - 6.11.0
  yarn - 1.22.15

Rust environment
  rustc - 1.57.0
  cargo - 1.57.0

Rust environment
  rustup - 1.24.3
  rustc - 1.57.0
  cargo - 1.57.0
  toolchain - stable-x86_64-pc-windows-msvc (environment override by RUSTUP_TOOLCHAIN)

App directory structure
/.vscode
/dist
/node_modules
/public
/src
/src-tauri
```

FabianLars:
I guess you figured it out because you closed it. But if someone else has a similar issue i'll answer anyway.
The problem here is the "pub" keyword on the command. Removing that will fix it. This is only a problem if the command and the `tauri::Builder` are in the same file.
For example, commands in a seperate file can be `pub` (with `pub(crate)` probably being a better choice)

Borber:
[...]

PR #12979, opened 2025-03-14
fix(macros): allow pub commands in crate root, closes #9362

### LINKED ISSUE #9362: [feat] Allow using #[command] across workspaces
### Describe the problem

It seems that right now I can only call the `#[tauri::command]` on the root `src/` folder of my project. Due to the size and complexity of the project that I am working on, I need to separate parts of the codebase into Cargo workspaces. I want to be able to call certain functions in these workspaces from the frontend so I have added these workspace as a dependency of the root package which is responsible for initialising my main app. However when I declare a function inside a workspace as a command throws an error like this:

```
error[E0255]: the name `__cmd__get_details` is defined multiple times
  --> config/src/lib.rs:62:8
   |
61 | #[tauri::command]
   | ----------------- previous definition of the macro `__cmd__get_details` here
62 | pub fn get_details() -> Result<UserConfig> {
   |        ^^^^^^^^^^^ `__cmd__get_details` reimported here
   |
   = note: `__cmd__get_details` must be defined only once in the macro namespace of this module
```

I would like to be able to declare functions in workspaces as tauri commands just as easily as I can do it in my root package.

### Alternatives considered

My current approach is to make wrappers for all the functions in my root package and declare then declare these wrappers as tauri commands. This is terrible in my eyes as it is very error-prone, adds a useless function call and requires me to use weird gimmicks to properly bubble up the errors in the wrapper function.

### Additional context

Tauri version: v1.6.1
OS: Fedora Linux 39
webkitgtk6.0 version: 2.42.1

lucasfernog:
moves the command macro export to a separate module, so it can be defined as pub in the crate root

ref #3198

### DIFF
--- a/.changes/command-crate-root-pub.md
+++ b/.changes/command-crate-root-pub.md
@@ -1,0 +1,5 @@
+---
+"tauri-macros": patch:bug
+---
+
+Allow defining a command as public in the crate root.
--- a/crates/tauri-macros/src/command/handler.rs
+++ b/crates/tauri-macros/src/command/handler.rs
@@ -43,8 +43,13 @@ impl Parse for Handler {
         // the name of the actual command function
         let command = last.ident.clone();
 
-        // set the path to the command function wrapper
-        last.ident = super::format_command_wrapper(&command);
+      // set the path to the command function wrapper
+      last.ident = super::format_command_wrapper(&command);
+      // the wrapper path is actually `${format_command_wrapper}::${format_command_wrapper}`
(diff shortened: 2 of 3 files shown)


lucasfernog:
there's a reason why chip gave up trying this :) simply can't make it work on all possible cases. this PR fixes the root crate pub command problem, but it doesn't work when you import the command function and refer to it in generate_handler![] instead of using the absolute path to the command.

PR #12983, opened 2025-03-15
chore: Fix example readme instructions. Add `pnpm build:debug`.

### LINKED ISSUE #12980: [docs] readme.md under example/api
cannot find the scripts because  ./script is not there. please check the script file name also for linux/mac. scripts for windows are missing, too.

https://github.com/tauri-apps/tauri/tree/dev/examples/api

FabianLars:
closes #12980


lucasfernog (approved):

FabianLars:
@lucasfernog that's what the --plugin cli flag in package.json was supposed to do. Did that not work? (And how did you notice?)

lucasfernog:
in prod your script is correct but i didn't want debug builds to generate a different bundle file making a possible incorrect git addition

FabianLars:
Then we can remove the flag again and make build:debug an alias for build (I do believe that unminified debug builds can sometimes be helpful though)

lucasfernog:
they will still be unminified, I made that change only for bundle.global.js

FabianLars:
oh shit i didn't see that we always only modified the bundle. Thinking about it we should probably stick to that (== removing the --plugin flag again) and let the users' bundlers do the minification (quick google search suggests that this is best practice). Will open a new pr for that in a sec

Issue #12990, opened 2025-03-15
[bug] Windows: `withGlobalTauri: true`: `window.__TAURI__` is undefined while executing top-level scripts: the examples are broken


WofWca:
### Describe the bug

The examples do not work, at least for me on Windows.

### Reproduction

1. `git clone https://github.com/tauri-apps/tauri.git && cd tauri`
2. `cargo run --example helloworld`
3. Type something and submit the form.

https://github.com/user-attachments/assets/fe1ae998-7ef9-43ce-abfd-c79132f32adc

### Expected behavior

The "Hello from Rust" message appears

### Full `tauri info` output

```text
Tauri repo b05f82d35ba068bfeb44193b204fbfe365415a25
```

### Stack trace

```text

```

### Additional context

If I open dev tools and refresh the window, then it starts working.

Looks like `with_global_tauri` utilizes `with_initialization_script`

https://github.com/tauri-apps/tauri/blob/95fc3cd424a1d05b818be435abf1510c06dd7a87/crates/tauri/src/manager/webview.rs#L377-L405

https://github.com/tauri-apps/tauri/blob/cedb24d494b84111daa3206c05196c8b89f1e994/crates/tauri-runtime-wry/src/lib.rs#L4532

But its docs state

> It is guaranteed that code is executed before `window.onload`.

I considered just adding a docstring to `withGlobalTauri` mentioning the fact that you have to wait for `onload` to finish before using it, but then I saw that we have docs on this https://v2.tauri.app/develop/calling-rust/#basic-example, and the docs don't say anything about `onload`.

Legend-Master:
I believe this is the cause: https://github.com/tauri-apps/wry/pull/1418#discussion_r2005617414

PR #12992, opened 2025-03-15
chore: fix helloworld example hang on Windows

WofWca:
The issue was introduced in cedb24d494b84111daa3206c05196c8b89f1e994 (https://github.com/tauri-apps/tauri/pull/12665).

As the `.cookies()` docs state, the command should be async.

I discovered the issue while I was preparing the report of https://github.com/tauri-apps/tauri/issues/12990.

On a side note: I don't think that the cookies thing needs to be in the "helloworld" example.

### DIFF
--- a/examples/helloworld/main.rs
+++ b/examples/helloworld/main.rs
@@ -5,7 +5,7 @@
 #![cfg_attr(not(debug_assertions), windows_subsystem = "windows")]
 
 #[tauri::command]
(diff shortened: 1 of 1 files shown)


amrbashir (commented):
Looks like the cookies thing slipped here accidentally, so lets remove that instead

WofWca:
Closed in favor of https://github.com/tauri-apps/tauri/pull/12995

PR #12998, opened 2025-03-16
fix: use bun.lock to identify bun's lockfile (fix #12914)

### LINKED ISSUE #12914: [bug] Detect bun as package manager by checking bun.lock
Bun moved away from using its binary lock file format, so the scripts (like`tauri add`) that depends on this fail.

https://bun.sh/docs/install/lockfile

The new `bun.lock` should probably be added to this list

https://github.com/tauri-apps/tauri/blob/2d029a9f53117f756f9d6783fab74b06a7ab3e8e/crates/tauri-cli/src/helpers/npm.rs#L84

jason89521:
fix #12914

### DIFF
--- a/.changes/fix-bun-lockfile.md
+++ b/.changes/fix-bun-lockfile.md
@@ -1,0 +1,5 @@
+---
+'tauri-cli': 'patch:bug'
+---
+
+For bun's lockfile, check both `bun.lock` and `bun.lockb`.
--- a/crates/tauri-cli/src/helpers/npm.rs
+++ b/crates/tauri-cli/src/helpers/npm.rs
@@ -81,7 +81,7 @@ impl PackageManager {
             PackageManager::Yarn
           };
           found.push(yarn);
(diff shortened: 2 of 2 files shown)


FabianLars:
thanks for the pr!

.lockb is still supported so we should support both instead of only .lock

we also need a small changefile in `.changes/` :)

jason89521:
> thanks for the pr!
> 
> .lockb is still supported so we should support both instead of only .lock
> 
> we also need a small changefile in `.changes/` :)

I pushed [62d7814](https://github.com/tauri-apps/tauri/pull/12998/commits/62d7814db4889e835154e7afd8e61831897af5ad) and [2480e6a](https://github.com/tauri-apps/tauri/pull/12998/commits/2480e6af314cdc2fb87a7eb0ceb78c0107dadca6). Please check whether I did it in the right way. Thanks!

FabianLars (approved):

PR #13000, opened 2025-03-16
docs: clarify Capability::windows/webviews OR nature

WofWca:
It is intuitive to expect that for the capability to be enabled, _both_ the window label and the webview label have to match. However, this is not the case: the capability is enabled if _either_ the window label matches a pattern
in `Capability::windows` _or_ the webview label matches a pattern in `Capability::webviews`.

This commit should somewhat clarify this oddity
and protect developers from adding excessive permissions.

For reference, `Capability::webviews` was added in 0cb0a15ce22af3d649cf219ac04188c14c5f4905
(https://github.com/tauri-apps/tauri/pull/8789).

### DIFF
--- a/crates/tauri-utils/src/acl/capability.rs
+++ b/crates/tauri-utils/src/acl/capability.rs
@@ -148,7 +148,11 @@ pub struct Capability {
   pub local: bool,
   /// List of windows that are affected by this capability. Can be a glob pattern.
   ///
-  /// On multiwebview windows, prefer [`Self::webviews`] for a fine grained access control.
+  /// If a window label mathes any of the patterns in this list,
+  /// the capability will be enabled on all the webviews of that window,
+  /// regardless of the value of [`Self::webviews`].
+  /// On multiwebview windows, prefer specifying [`Self::webviews`] and omitting [`Self::windows`]
+  /// for a fine grained access control.
   ///
   /// ## Example
   ///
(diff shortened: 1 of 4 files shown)


crates/tauri-utils/src/acl/capability.rs
```diff
@@ -148,7 +148,11 @@ pub struct Capability {
   pub local: bool,
   /// List of windows that are affected by this capability. Can be a glob pattern.
   ///
-  /// On multiwebview windows, prefer [`Self::webviews`] for a fine grained access control.
+  /// If a window label mathes any of the patterns in this list,
```
amrbashir:
```suggestion
  /// If a window label matches any of the patterns in this list,
```

crates/tauri-utils/src/acl/capability.rs
```diff
@@ -148,7 +148,11 @@ pub struct Capability {
...
+  /// the capability will be enabled on all the webviews of that window,
+  /// regardless of the value of [`Self::webviews`].
```
amrbashir, line 153:
```suggestion
  /// regardless of the value of [`Self::webviews`].
  ///
```

crates/tauri-utils/src/acl/capability.rs
```diff
@@ -157,8 +161,9 @@ pub struct Capability {
   pub windows: Vec<String>,
   /// List of webviews that are affected by this capability. Can be a glob pattern.
   ///
-  /// This is only required when using on multiwebview contexts, by default
-  /// all child webviews of a window that matches [`Self::windows`] are linked.
+  /// The capability will be enabled on all the webviews
+  /// whose label matches any of the patterns in this list,
+  /// regardless of whether the webview's window label mathes a pattern in [`Self::windows`].
```
amrbashir:
```suggestion
  /// regardless of whether the webview's window label matches a pattern in [`Self::windows`].
```

WofWca:
Thanks for the suggestions! Do I still need to sign the commit that applies them?

lucasfernog (dismissed):

lucasfernog:
thanks!

PR #13001, opened 2025-03-16
docs: improve `tauri_build::build` & `try_build`

WofWca:
The docstring has been unchanged ever since
40ac52971eb350e839f41401043c5cde31fa8974 (2021).
The `build()` function has grown massively since then,
now it's not only about Windows.

`try_build` is not only about not panicking, but it also takes the
`attributes: Attributes` argument,
which is very important for controlling command permissions.

### DIFF
--- a/crates/tauri-build/src/lib.rs
+++ b/crates/tauri-build/src/lib.rs
@@ -419,18 +419,14 @@ pub fn is_dev() -> bool {
 
 /// Run all build time helpers for your Tauri Application.
 ///
-/// The current helpers include the following:
-/// * Generates a Windows Resource file when targeting Windows.
-///
 /// # Platforms
 ///
-/// [`build()`] should be called inside of `build.rs` regardless of the platform:
-/// * New helpers may target more platforms in the future.
-/// * Platform specific code is handled by the helpers automatically.
-/// * A build script is required in order to activate some cargo environmental variables that are
-///   used when generating code and embedding assets - so [`build()`] may as well be called.
+/// [`build()`] should be called inside of `build.rs` regardless of the platform.
 ///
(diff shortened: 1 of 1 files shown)


crates/tauri-build/src/lib.rs
```diff
@@ -419,18 +419,17 @@ pub fn is_dev() -> bool {
...
 /// # Platforms
 ///
-/// [`build()`] should be called inside of `build.rs` regardless of the platform:
-/// * New helpers may target more platforms in the future.
-/// * Platform specific code is handled by the helpers automatically.
-/// * A build script is required in order to activate some cargo environmental variables that are
-///   used when generating code and embedding assets - so [`build()`] may as well be called.
+/// [`build()`] should be called inside of `build.rs` regardless of the platform.
 ///
-/// In short, this is saying don't put the call to [`build()`] behind a `#[cfg(windows)]`.
```
amrbashir, line 433: I think we should keep this line

crates/tauri-build/src/lib.rs
```diff
@@ -450,7 +449,7 @@ pub fn build() {
   }
 }
 
-/// Non-panicking [`build()`].
+/// Like [`build()`], but takes an extra configuration argument, and does not panic.
```
amrbashir:
```suggestion
/// Same as [`build()`], but takes an extra configuration argument, and does not panic.
```

crates/tauri-build/src/lib.rs
WofWca: We have "[`build()`] should be called inside of `build.rs` regardless of the platform." just above, isn't it enough?

lucasfernog (dismissed):

PR #13002, opened 2025-03-16
fix: align log like Cargo

nanoqsh:
I noticed that Tauri's log prints crookedly compared to Cargo
image

This PR fixes it, now everything prints aligned
image


amrbashir (approved):
Thank you

PR #13013, opened 2025-03-17
docs: update `.disable_javascript()` docs

WofWca:
The initial version of the MR that introduced this to Wry
(https://github.com/tauri-apps/wry/pull/1496)
had no Android implementation, but it has been added
in a later version of the MR, so the comment is no longer valid.


amrbashir (approved):

amrbashir:
Thank you

PR #13032, opened 2025-03-20
fix: channel no longer being always allowed

Legend-Master:
We have introduced a small breaking change in #12890 that `Channel` will now follow ACL instead of always allowed, even though this is allowed in the default permission group, it's still a breaking change, let's delay this change to v3

~~Also changed the plugin name to `channel` to prepare for this change~~
Also commented a TODO about changing the plugin name to `channel` in v3


amrbashir:
> Also changed the plugin name to `channel` to prepare for this change

That in itself is a breaking change

Legend-Master:
> That in itself is a breaking change

This is never exposed before and only used by our init scripts internally so I don't think it's considered a breaking change?

Legend-Master:
Anyways, this can be done in v3 as well, reverted the plugin name change

lucasfernog:
we need to check if our optimization won't remove the channel fetch command though

amrbashir:
> > That in itself is a breaking change
> 
> This is never exposed before and only used by our init scripts internally so I don't think it's considered a breaking change?

Right, I was thinking the js package was using that but looks like it is not.

Legend-Master:
> we need to check if our optimization won't remove the channel fetch command though

I was thinking I removed the plugin mark so it will always get preserved but then I realized that it will just be considered as app's command, should be good to go now with 9fd5f391f26eb336050a7d306404ded5c31cb4e5, added back the special check

lucasfernog (approved):

PR #13033, opened 2025-03-20
fix: don't ship global api bundle if withGlobalTauri is false

Legend-Master:
Another effort to reduce bloat, we're currently shipping `bundle.global.js` (~40KB currently) no matter if it's used or not, this PR changes it to use our existing code `plugin_global_api_scripts` for the core plugins as well

Reference: #12820

### DIFF
--- a/.changes/config.json
+++ b/.changes/config.json
@@ -4,7 +4,7 @@
     "feat": "New Features",
     "enhance": "Enhancements",
     "bug": "Bug Fixes",
-    "pref": "Performance Improvements",
+    "perf": "Performance Improvements",
     "changes": "What's Changed",
     "sec": "Security fixes",
     "deps": "Dependencies",
--- a/.changes/dont-ship-global-bundle-if-no-global-tauri.md
+++ b/.changes/dont-ship-global-bundle-if-no-global-tauri.md
@@ -1,0 +1,8 @@
+---
+tauri: 'patch:perf'
+tauri-build: 'patch:perf'
+tauri-plugin: 'patch:perf'
+tauri-utils: 'patch:perf'
+---
+
+Don't ship global `bundle.global.js` if `app > withGlobalTauri` is set to false
--- a/README.md
+++ b/README.md
@@ -43,13 +43,13 @@ The list of Tauri's features includes, but is not limited to:
 
 Tauri currently supports development and distribution on the following platforms:
 
-| Platform          | Versions                                                                                                        |
-| :---------------- | :-------------------------------------------------------------------------------------------------------------- |
-| Windows           | 7 and above                                                                                                     |
-| macOS             | 10.15 and above                                                                                                 |
-| Linux             | webkit2gtk 4.0 for Tauri v1 (for example Ubuntu 18.04). webkit2gtk 4.1 for Tauri v2 (for example Ubuntu 22.04). |
-| iOS/iPadOS        | 9 and above                                                                                                     |
-| Android           | 7 and above (currently 8 and above)                                                                             |
+| Platform   | Versions                                                                                                        |
+| :--------- | :-------------------------------------------------------------------------------------------------------------- |
+| Windows    | 7 and above                                                                                                     |
+| macOS      | 10.15 and above                                                                                                 |
+| Linux      | webkit2gtk 4.0 for Tauri v1 (for example Ubuntu 18.04). webkit2gtk 4.1 for Tauri v2 (for example Ubuntu 22.04). |
+| iOS/iPadOS | 9 and above                                                                                                     |
+| Android    | 7 and above (currently 8 and above)                                                                             |
 
 ## Contributing
 
--- a/crates/tauri-build/src/acl.rs
+++ b/crates/tauri-build/src/acl.rs
@@ -432,8 +432,6 @@ pub fn build(out_dir: &Path, target: Target, attributes: &Attributes) -> super::
   let capabilities_path = save_capabilities(&capabilities)?;
   fs::copy(capabilities_path, out_dir.join(CAPABILITIES_FILE_NAME))?;
 
-  tauri_utils::plugin::save_global_api_scripts_paths(out_dir);
-
   let mut permissions_map = inline_plugins_acl.permission_files;
   if has_app_manifest {
     permissions_map.insert(APP_ACL_KEY.to_string(), app_acl.permission_files);
--- a/crates/tauri-build/src/lib.rs
+++ b/crates/tauri-build/src/lib.rs
@@ -511,6 +511,8 @@ pub fn try_build(attributes: Attributes) -> Result<()> {
 
   acl::build(&out_dir, target, &attributes)?;
 
+  tauri_utils::plugin::save_global_api_scripts_paths(&out_dir, None);
+
   println!("cargo:rustc-env=TAURI_ENV_TARGET_TRIPLE={target_triple}");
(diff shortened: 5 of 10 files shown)


lucasfernog (approved):

WSH032:
I remember that in `2.3`, setting `withGlobalTauri=True` at build time and then setting `withGlobalTauri=False` at runtime could still disable the global JS API.  
However, in `2.5`, the runtime configuration no longer works. I think this is due to changes in `crates/tauri/src/manager/webview.rs`?

I want to upstream the [runtime configuration `context`](https://github.com/tauri/tauri/blob/9916a02159369818f9ff930eee040f977a4b48da/python/pytauri-wheel/src/lib.rs#L170-L302) feature to Tauri, so I've started to pay attention to which configurations are compile-time and which are runtime.

Legend-Master:
> I remember that in 2.3, setting withGlobalTauri=True at build time and then setting withGlobalTauri=False at runtime could still disable the global JS API.

Yeah, I think so, but only for the ones in tauri not the plugins, this PR just combined the logic together so that they are now bundled together

WSH032:
So in reality, was I just taking advantage of a "bug"?  
Or, is Tauri willing to make `withGlobalTauri` configurable at runtime? If it affects the build and cannot be configured at runtime, I think it would make more sense to put it in `BuildConfig`.

Legend-Master:
> So in reality, was I just taking advantage of a "bug"?

I can't say if that was a bug or not to be honest since it can really be augured both ways

> Or, is Tauri willing to make withGlobalTauri configurable at runtime?

I don't think that's a good idea, that means we'll have to ship the scripts no matter what

> If it affects the build and cannot be configured at runtime, I think it would make more sense to put it in BuildConfig.

Since it affects what we inject to the app's code directly, I think it makes more sense to stay in `app`

`Context::config_mut` and all the `Context` related methods should ideally really only be used by the `generate_context` macro, and if you use it directly, you'll need to mimic the behavior of that macro

If all you want is to set it to false through `Context::config_mut` at runtime though, we can add a check to not injecting it

WSH032:
My idea is to add a feature similar to `dynamic-acl` or a configuration like `BuildConfig::remove_unused_commands` to decide at compile time whether to ship the scripts. `AppConfig::with_global_tauri` would remain for deciding at runtime whether to actually inject the scripts.

My use case is to precompile Tauri as a dynamic library or executable, so that users don't need a Rust compiler.
**Only** `BuildConfig` and `BundleConfig` determine compile-time configuration and cannot be changed at runtime (i.e., modifying them at runtime has no effect).
`identifier`, `AppConfig`, and `PluginConfig` can be modified dynamically at runtime.

I'll create a PR to try implementing this feature. It's somewhat similar to:

```rust
impl Context {
    /// tauri_dir: absolute path to `src-tauri/`
    fn load_runtime_context(&mut self, tauri_dir: PathBuf);
}
```

`fn load_runtime_config` will load `{tauri_dir}/tauri.conf.json` and mimic the behavior of the `generate_context` macro. For example, [loading the web page from disk according to `BuildConfig::FrontendDist`](https://github.com/pytauri/pytauri/blob/9916a02159369818f9ff930eee040f977a4b48da/python/pytauri-wheel/src/lib.rs#L37-L60), [loading and setting the global icon from disk according to `BundleConfig::icon`](https://github.com/pytauri/pytauri/blob/9916a02159369818f9ff930eee040f977a4b48da/python/pytauri-wheel/src/lib.rs#L108-L142), and dynamically adding `acl` from `{tauri_dir}/capabilities`, etc.

This is somewhat similar to [Python's initialization](https://docs.python.org/3/c-api/init_config.html#), where compile-time configuration is used by default, but some parts can be configured at runtime.

Legend-Master:
Look at what we have in `AppConfig`, I changed my mind, maybe it makes more sense to be in `BuildConfig` 😂

I don't fully understand what exactly you want to do with generating contexts at runtime though, like do you want to ship the app with all the config and capability files? (or it's just easier for the pytauri to work without the help of `tauri-build` and `generate_context` macro in dev?)

WSH032:
What I want is: I only ship a precompiled executable (or dynamic library), and users provide their own `tauri.conf.json` and `capabilities` files.
The executable will generate the corresponding `Context` based on the user-provided `tauri.conf.json` and `capabilities`, and then use this `Context` to initialize the Tauri App.

*Not sure if I explained this clearly.*

WSH032:
Here is a practical usage example: https://github.com/pytauri/pytauri/tree/main/examples/tauri-app-wheel/python/src/tauri_app_wheel

The prebuilt binary is [downloaded from PyPI](https://pypi.org/project/pytauri-wheel/#files), while the `capabilities` and `Tauri.toml` files are provided by the user at runtime. The [script](https://github.com/pytauri/pytauri/blob/9916a02159369818f9ff930eee040f977a4b48da/examples/tauri-app-wheel/python/src/tauri_app_wheel/__init__.py#L126-L130) can dynamically generate the `Context` to drive the Tauri App, such as [creating windows](https://github.com/pytauri/pytauri/blob/9916a02159369818f9ff930eee040f977a4b48da/examples/tauri-app-wheel/python/src/tauri_app_wheel/Tauri.toml#L12-L15), [setting permissions](https://github.com/pytauri/pytauri/blob/9916a02159369818f9ff930eee040f977a4b48da/examples/tauri-app-wheel/python/src/tauri_app_wheel/capabilities/default.toml#L5-L10), and [loading assets from disk (instead of using embedded assets)](https://github.com/pytauri/pytauri/blob/9916a02159369818f9ff930eee040f977a4b48da/examples/tauri-app-wheel/python/src/tauri_app_wheel/Tauri.toml#L6-L7).

Currently, I have [hacked the `Context`](https://github.com/pytauri/pytauri/blob/9916a02159369818f9ff930eee040f977a4b48da/python/pytauri-wheel/src/lib.rs#L183-L302) to achieve this, which is quite fragile. So I'm wondering if this functionality could be upstreamed to Tauri (even as an unstable API).

Legend-Master:
I see now, after skimming through the code, correctly me if I'm wrong though

So you're trying to mimic the `generate_context`, but because it's python, this would happen at runtime, and the `generate_context` you used in your code would be generating a barebone `Context` with the info in your repo's environment, and then you would modify that based on the user's code at runtime, and what you want is more or less that we provide a way to generate the `Context` struct in rust (which can be used directly by a rust function), instead of that macro that generates rust code (text)

WSH032:
> So you're trying to mimic the generate_context, but because it's python, this would happen at runtime, and the generate_context you used in your code would be generating a barebone Context with the info in your repo's environment, and then you would modify that based on the user's code at runtime

Exactly.

> what you want is more or less that we provide a way to generate the Context struct in rust (which can be used directly by a rust function), instead of that macro that generates rust code (text)

There might be a slight difference from what I had in mind. If I understand correctly, it's impossible to avoid using macros entirely, because something always needs to be embedded (such as `plugin_global_api_scripts` and `plist`).  
What I want right now is just a **runtime** api to patch the existing `Context` (which is generated at build time by `generate_context`). This api needs to mimic the behavior of [`generate_context`](https://github.com/tauri-apps/tauri/blob/6a39f49991e613e8f3befe0e8dff288482ccdd89/crates/tauri-codegen/src/context.rs#L134) (for example, setting the icon, providing assets, etc.).

I'll submit a PR later, as long as someone is available to review it.

PR #13040, opened 2025-03-21
fix: `run_return` not responding to `restart`

Legend-Master:
So the `run_return` PR and `request_restart`/`restart` change PRs happened at the same time, this didn't get into that function

Since `run` and `run_return` share the exactly same callback, moved it to a function


lucasfernog:
i think we should handle the restart on our end anyway, otherwise the API would look weird.
(let's document that too)

lucasfernog (approved):

PR #13043, opened 2025-03-21
docs: Improve BaseDirectory documentation

### LINKED ISSUE #7555: [docs] Improve `BaseDirectory` documentation
Is it possible to improve the documentation for `BaseDirectory` here:

https://github.com/tauri-apps/tauri/blob/4db363a03c182349f8491f46ced258d84723b11f/tooling/api/src/path.ts#L19

It might be enough to just add a reference to each enum value to the respective `path.appFooBar` documentation (so add a note to `BaseDirectory.AppCache` that points to `path.appCacheDir()`)

Other enum that's documented:

https://github.com/tauri-apps/tauri/blob/b7ae725/tooling/api/src/window.ts#L228

Discord thread: <https://discord.com/channels/616186924390023171/1137125879571558480> and <https://discord.com/channels/616186924390023171/731495028677148753/1137122018744074391> that have a bit more context for someone using the enums without knowing what they resolve to.

Would be nice if we didn't have to deal with inline headers `###` downstream in docs, but I can handle that for the sake of more documentation <3

Levminer:
Improves the docs so that you can see where folders point on different operating systems easily.

fixes #7555

### DIFF
--- a/crates/tauri/src/path/mod.rs
+++ b/crates/tauri/src/path/mod.rs
@@ -115,7 +115,7 @@ pub enum BaseDirectory {
   /// The Resource directory.
   /// Resolves to the resource directory of this app.
   Resource = 11,
-  /// A temporary directory. 
+  /// A temporary directory.
   /// Resolves to [`std::env::temp_dir`].
   Temp = 12,
   /// The default app config directory.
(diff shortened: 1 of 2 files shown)


amrbashir (approved):

amrbashir:
Thank you

PR #13057, opened 2025-03-23
enhance: include permission group permissions in the generated schema

Legend-Master:
- Add in a `#### This default permission set includes the following:` for the docs before default permissions
- Include permission group permissions in the generated schema
- Also add `markdownDescription` field to enhance the schema display in vscode

image

### DIFF
--- a/.changes/include-permissions-in-schema.md
+++ b/.changes/include-permissions-in-schema.md
@@ -1,0 +1,5 @@
+---
+tauri: patch:enhance
+---
+
+Include permission group permissions in the generated schema
--- a/.changes/include-permissions-in-schema.md
+++ b/.changes/include-permissions-in-schema.md
@@ -1,5 +1,6 @@
 ---
 tauri: patch:enhance
+tauri-utils: patch:enhance
 ---
 
 Include permission group permissions in the generated schema
--- a/.changes/include-permissions-in-schema.md
+++ b/.changes/include-permissions-in-schema.md
@@ -3,4 +3,4 @@ tauri: minor:enhance
 tauri-utils: minor:enhance
 ---
 
-Include permission group permissions in the generated schema
+Enhanced the description of generated docs and schema for permission sets to include list of permissions within.
--- a/.changes/include-permissions-in-schema.md
+++ b/.changes/include-permissions-in-schema.md
@@ -1,6 +1,6 @@
 ---
-tauri: patch:enhance
-tauri-utils: patch:enhance
+tauri: minor:enhance
+tauri-utils: minor:enhance
 ---
 
 Include permission group permissions in the generated schema
--- a/.changes/includes-permissions-in-docs.md
+++ b/.changes/includes-permissions-in-docs.md
@@ -1,5 +1,0 @@
----
-tauri-utils: minor:enhance
----
-
-Add in a `#### This default permission set includes the following:` for the docs before default permissions
--- a/.changes/includes-permissions-in-docs.md
+++ b/.changes/includes-permissions-in-docs.md
@@ -1,0 +1,5 @@
+---
+tauri-utils: patch:enhance
+---
+
+Add in a `#### This default permission set includes the following:` for the docs before default permissions
--- a/.changes/includes-permissions-in-docs.md
+++ b/.changes/includes-permissions-in-docs.md
@@ -1,5 +1,5 @@
 ---
-tauri-utils: patch:enhance
+tauri-utils: minor:enhance
 ---
 
 Add in a `#### This default permission set includes the following:` for the docs before default permissions
--- a/crates/tauri-utils/src/acl/schema.rs
+++ b/crates/tauri-utils/src/acl/schema.rs
@@ -88,13 +88,17 @@ pub trait PermissionSchemaGenerator<
     let mut permission_schemas = Vec::new();
 
     // schema for default set
-    if let (Some(description), Some(permissions)) = (
-      self.default_set_description(),
-      self.default_set_permissions(),
-    ) {
-      let description = add_permissions_to_description(description, permissions);
-      let default = Self::perm_id_schema(name, "default", Some(&description));
-      permission_schemas.push(default);
+    if self.has_default_permission_set() {
+      let description = self.default_set_description().unwrap_or_default();
+      let description = if let Some(permissions) = self.default_set_permissions() {
+        add_permissions_to_description(description, permissions)
+      } else {
+        description.to_string()
+      };
+      if !description.is_empty() {
+        let default = Self::perm_id_schema(name, "default", Some(&description));
+        permission_schemas.push(default);
+      }
     }
 
     // schema for each permission set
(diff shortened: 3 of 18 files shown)


crates/tauri/build.rs
```diff
@@ -437,22 +444,15 @@ fn define_default_permission_set(
...
   let default_toml = permissions_out_dir.join("default.toml");
   let toml_content = format!(
-    r#"# {LICENSE_HEADER}
+    r#"{LICENSE_HEADER}
 
 [default]
-description = """Default core plugins set which includes:
-{}
-"""
+description = "Default core plugins set which includes:"
```
amrbashir: Why remove permissions list here?

crates/tauri-utils/src/acl/schema.rs
```diff
@@ -72,14 +88,28 @@ pub trait PermissionSchemaGenerator<
     let mut permission_schemas = Vec::new();
 
     // schema for default set
-    if self.has_default_permission_set() {
-      let default = Self::perm_id_schema(name, "default", self.default_set_description());
+    if let (Some(description), Some(permissions)) = (
```
amrbashir: what if a permission set doesn't have a description? we should still push that to list of schemas

crates/tauri-utils/src/acl/schema.rs
```diff
@@ -72,14 +88,28 @@ pub trait PermissionSchemaGenerator<
...
+      self.default_set_description(),
+      self.default_set_permissions(),
+    ) {
+      let default = Self::perm_id_schema(
+        name,
+        "default",
+        Some(&add_permissions_to_description(description, permissions)),
+      );
```
amrbashir: for the sake of readability, let's break this a bit
```suggestion
      let description = add_permissions_to_description(description, permissions);
      let default = Self::perm_id_schema(name, "default", Some(&description));
```

crates/tauri-utils/src/acl/schema.rs
```diff
@@ -72,14 +88,28 @@ pub trait PermissionSchemaGenerator<
...
     for set in self.permission_sets() {
-      let schema = Self::perm_id_schema(name, &set.identifier, Some(&set.description));
+      let schema = Self::perm_id_schema(
+        name,
+        &set.identifier,
+        Some(&add_permissions_to_description(
+          &set.description,
+          &set.permissions,
+        )),
+      );
```
amrbashir: same

crates/tauri-utils/src/acl/schema.rs
```diff
@@ -93,6 +123,18 @@ pub trait PermissionSchemaGenerator<
   }
 }
 
+fn add_permissions_to_description(description: &str, permissions: &[String]) -> String {
```
amrbashir: ideally this could be a default method on the trait `PermissionSchemaGenerator`

```rs
trait PermissionSchemaGenerator {
    fn default_set_description_with_permissions(&self) -> String;
        // then here you can use `self.default_set_description()`
        // and `self.default_set_permissions()`
    }
}
```

amrbashir (changes requested):

crates/tauri/build.rs
Legend-Master: Since this will be displayed from the schema/docs gen, including it here would duplicate things

This is how it looks right now

image

crates/tauri-utils/src/acl/schema.rs
Legend-Master: Currently permission sets always have a description

crates/tauri-utils/src/acl/schema.rs
Legend-Master: This is also used for the permission sets, so I don't think it makes sense to have a function purely for the default description with permissions

crates/tauri-utils/src/acl/schema.rs
Legend-Master: Never mind, `DefaultPermission` has an optional description

crates/tauri-utils/src/acl/schema.rs
amrbashir: you already added the function, I am just saying it should be moved here and avoid passing arguments.

crates/tauri/build.rs
amrbashir: what about other editors that don't respect `markdownDescription`?

crates/tauri-utils/src/acl/schema.rs
Legend-Master: Sorry I don't quite follow, this is used by both default permission set and normal permission sets, do you mean extract the default permission code into a function?

crates/tauri/build.rs
Legend-Master: Supported editors will show it in rendered markdown format, others will just show the plain text like we always did, the schema are applied to both `description` and `markdownDescription`

crates/tauri-utils/src/acl/schema.rs
amrbashir: Ah now I understand what you mean. we could make another trait but not worth it. this is fine.

crates/tauri/build.rs
amrbashir: Ah and that is handled by the `add_permissions_to_description` function.

amrbashir: Should `add_permissions_to_description` function also add the wording, i.e `which includes`, because right now this is how it looks
image

Legend-Master: I originally did that, but then I realized that because our doc gen adds in the commands at the end of the description, many of our plugin docs are made like this to make the doc gen fill in the commands

> ```toml
> [default]
> description = """
> This set of permissions describes the what kind of
> file system access the `fs` plugin has enabled or denied by default.
> 
> #### Granted Permissions
> 
> ...
> 
> #### Included permissions within this default permission set:
> """
> ```

I'm not sure if we should change the docs gen or just align the behavior with it, I think we might want to do it eventually in the docs gen, but right now it feels like a breaking change

amrbashir: I would prefer to add it here and align plugins docs to this change. 

Also this is not a breaking change, it is just some changes to how we present some docs so it is fine.

Legend-Master: Did that, we can tweak the plugin docs once this lands

Also feel free to tweak the wording

.changes/include-permissions-in-schema.md
```diff
@@ -0,0 +1,6 @@
+---
+tauri: minor:enhance
+tauri-utils: minor:enhance
+---
+
+Include permission group permissions in the generated schema
```
amrbashir:
```suggestion
Enhanced the description of generated docs and schema for permission sets to include list of permissions within.
```

.changes/includes-permissions-in-docs.md
```diff
@@ -0,0 +1,5 @@
+---
+tauri-utils: minor:enhance
+---
+
+Add in a `#### This default permission set includes the following:` for the docs before default permissions
```
amrbashir: then we can remove this change file

amrbashir (approved):

PR #13058, opened 2025-03-23
docs: improve `Builder::run` docs

WofWca:
Otherwise people might use it and not realize
that `App::run` also takes the callback argument,
which could be useful for some.


amrbashir (approved):
Thank you

PR #13075, opened 2025-03-25
feat: add x11 feature flag to allow disabling x11 dependencies

aurelj:
This allows me to reduce the number of dependencies and final binary size of my tauri app which is running on a wayland only kiosk.

Note that this is a draft for now as it depends on the following tao and wry PR to be released:
- https://github.com/tauri-apps/tao/pull/1103
- https://github.com/tauri-apps/wry/pull/1528


aurelj:
Those changes were merged in #13629

PR #13076, opened 2025-03-25
feat: expose api to run initialization scripts on all frames.

Simon-Laux:
I need this to disable webrtc in all frames.
```js
try {
window.RTCPeerConnection = ()=>{};
RTCPeerConnection = ()=>{};
} catch (e){console.error("failed to overwrite RTCPeerConnection apis",e)}
try {
    window.webkitRTCPeerConnection = ()=>{};
    webkitRTCPeerConnection = ()=>{};
} catch (e){}
```

Currently the init script only runs on the main script despite the documentation telling otherwise:
```rs
/// Since it runs on all top-level document navigastions (and also child frame page navigations
/// ...
pub fn initialization_script(mut self, script: &str) -> Self {
```
~~This pr exposes the argument (`run_only_on_main_frame`) whether to run in all frames or only on the main frame.~~

- add API to run initialisation scripts on all frames
  - `WebviewBuilder::initialization_script_on_all_frames`
  - `WebviewWindowBuilder::initialization_script_on_all_frames`
  - `WebviewAttributes::initialization_script_on_all_frames`

### Progress

- [x] If this change requires a new version, then add a change file in `.changes` directory with the appropriate bump, see https://github.com/tauri-apps/tauri/blob/dev/.changes/README.md
- [x] Ensure that all your commits are signed https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits
- [x] Ensure `cargo test` and `cargo clippy` passes.

### DIFF
--- a/.changes/init-script-on-all-frames.md
+++ b/.changes/init-script-on-all-frames.md
@@ -1,0 +1,9 @@
+---
+tauri: minor
+tauri-runtime: minor
+---
+
+- Breaking API Additions:
+  - `WebviewBuilder::initialization_script` add `run_only_on_main_frame` parameter
+  - `WebviewWindowBuilder::initialization_script` add `run_only_on_main_frame` parameter
+- add API `WebviewAttributes::initialization_script_on_all_frames`
--- a/.changes/init-script-on-all-frames.md
+++ b/.changes/init-script-on-all-frames.md
@@ -3,7 +3,7 @@ tauri: minor:feat
 tauri-runtime: minor:feat
 ---
 
-- add API to run initialisation scripts on all frames
+- add API to run initialization scripts on all frames
   - `WebviewBuilder::initialization_script_on_all_frames`
   - `WebviewWindowBuilder::initialization_script_on_all_frames`
   - `WebviewAttributes::initialization_script_on_all_frames`
--- a/.changes/init-script-on-all-frames.md
+++ b/.changes/init-script-on-all-frames.md
@@ -1,6 +1,6 @@
 ---
-tauri: minor
-tauri-runtime: minor
+tauri: minor:feat
+tauri-runtime: minor:feat
 ---
 
 - add API to run initialisation scripts on all frames
--- a/.changes/init-script-on-all-frames.md
+++ b/.changes/init-script-on-all-frames.md
@@ -3,7 +3,7 @@ tauri: minor
 tauri-runtime: minor
 ---
 
-- Breaking API Additions:
-  - `WebviewBuilder::initialization_script` add `run_only_on_main_frame` parameter
-  - `WebviewWindowBuilder::initialization_script` add `run_only_on_main_frame` parameter
-- add API `WebviewAttributes::initialization_script_on_all_frames`
+- add API to run initialisation scripts on all frames
+  - `WebviewBuilder::initialization_script_on_all_frames`
+  - `WebviewWindowBuilder::initialization_script_on_all_frames`
+  - `WebviewAttributes::initialization_script_on_all_frames`
--- a/crates/tauri-runtime/src/webview.rs
+++ b/crates/tauri-runtime/src/webview.rs
@@ -201,9 +201,6 @@ pub struct WebviewAttributes {
   /// When webview load a new page, this initialization code will be executed.
   /// It is guaranteed that code is executed before `window.onload`.
   ///
-  /// Second parameter represents if script should be added to main frame only or sub frames also.
-  /// `true` for main frame only, `false` for all frames including sub frames.
-  ///
   /// ## Platform-specific
   ///
   /// - **Android on Wry:** The Android WebView does not provide an API for initialization scripts,
--- a/crates/tauri-runtime/src/webview.rs
+++ b/crates/tauri-runtime/src/webview.rs
@@ -208,7 +208,7 @@ pub struct WebviewAttributes {
   ///
   /// - **Android on Wry:** The Android WebView does not provide an API for initialization scripts,
   ///   so we prepend them to each HTML head. They are only implemented on custom protocol URLs.
-  pub initialization_scripts: Vec<(String, bool)>,
+  pub initialization_scripts: Vec<InitializationScript>,
   pub data_directory: Option<PathBuf>,
   pub drag_drop_handler_enabled: bool,
   pub clipboard: bool,
@@ -333,7 +333,10 @@ impl WebviewAttributes {
   ///   so we prepend them to each HTML head. They are only implemented on custom protocol URLs.
   #[must_use]
   pub fn initialization_script(mut self, script: &str) -> Self {
-    self.initialization_scripts.push((script.to_string(), true));
+    self.initialization_scripts.push(InitializationScript {
+      script: script.to_string(),
+      for_main_frame_only: true,
+    });
     self
   }
 
@@ -354,9 +357,10 @@ impl WebviewAttributes {
   ///   You can disable WebRTC this way, for example.
   #[must_use]
   pub fn initialization_script_on_all_frames(mut self, script: &str) -> Self {
-    self
-      .initialization_scripts
-      .push((script.to_string(), false));
+    self.initialization_scripts.push(InitializationScript {
+      script: script.to_string(),
+      for_main_frame_only: false,
+    });
     self
   }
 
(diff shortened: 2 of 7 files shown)


crates/tauri/src/webview/webview_window.rs
```diff
@@ -785,15 +785,17 @@ impl<R: Runtime, M: Manager<R>> WebviewWindowBuilder<'_, R, M> {
...
+  ///         .initialization_script(INIT_SCRIPT, true)
   ///         .build()?;
   ///       Ok(())
   ///     });
   /// }
   /// ```
   #[must_use]
-  pub fn initialization_script(mut self, script: &str) -> Self {
-    self.webview_builder = self.webview_builder.initialization_script(script);
+  pub fn initialization_script(mut self, script: &str, run_only_on_main_frame: bool) -> Self {
```
Legend-Master: Note that we can not do breaking change in v2 stable

Legend-Master (changes requested):

amrbashir:
there is also #11971 , I am fine to go with any just as long as it is not breaking change.

Simon-Laux:
I removed the breaking change making a new api instead. I named it `initialization_script_on_all_frames` because IMO the `initialization_script_for_main_only` name from the webkit api is very confusing.
But in the end I'm fine with either this pr or https://github.com/tauri-apps/tauri/pull/11971 getting merged, most important is that the feature is accessible.

crates/tauri-runtime/src/webview.rs
```diff
@@ -307,10 +318,45 @@ impl WebviewAttributes {
...
+  /// This is executed on all frames, main frame and also sub frames.
+  /// If you only want to run it in the main frame, use [Self::initialization_script] instead.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Android on Wry:** The Android WebView does not provide an API for initialization scripts,
+  ///   so we prepend them to each HTML head. They are only implemented on custom protocol URLs.
+  ///
+  /// - On **macOS** and **iOS** you can use this to overwrite browser apis to make them inacessible even in iframes.
+  ///   You can disable WebRTC this way, for example.
```
Simon-Laux: In our tests this also worked on linux and possibly even windows https://github.com/deltachat/deltachat-desktop/pull/4851 (though windows is chromium and not webkit, so we are still a bit surprised it worked there as well)

crates/tauri-runtime/src/webview.rs
```diff
@@ -197,7 +197,18 @@ impl<T: UserEvent, R: Runtime<T>> PartialEq for DetachedWebview<T, R> {
...
+  /// It is guaranteed that code is executed before `window.onload`.
+  ///
+  /// Second parameter represents if script should be added to main frame only or sub frames also.
+  /// `true` for main frame only, `false` for all frames including sub frames.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Android on Wry:** The Android WebView does not provide an API for initialization scripts,
+  ///   so we prepend them to each HTML head. They are only implemented on custom protocol URLs.
+  pub initialization_scripts: Vec<(String, bool)>,
```
Legend-Master: `Vec<(String, bool)>` is not very easy to understand, can we use a struct for this instead similar to https://github.com/tauri-apps/wry/pull/1531/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R2262-R2269

.changes/init-script-on-all-frames.md
```diff
@@ -0,0 +1,9 @@
+---
+tauri: minor
+tauri-runtime: minor
```
Legend-Master: This should fix the `check change tags` check

```suggestion
tauri: minor:feat
tauri-runtime: minor:feat
```

crates/tauri-runtime/src/webview.rs
Simon-Laux: Copied your struct from https://github.com/tauri-apps/wry/pull/1531, I also added a `new` method because building the structure itself might not be that easy to use for tauri api users.

Simon-Laux: your wry pr is not merged yet, that's why I didn't use it yet. but now it only needs to be changed in one place

crates/tauri-runtime/src/webview.rs
```diff
@@ -197,7 +197,18 @@ impl<T: UserEvent, R: Runtime<T>> PartialEq for DetachedWebview<T, R> {
...
+  /// It is guaranteed that code is executed before `window.onload`.
+  ///
+  /// Second parameter represents if script should be added to main frame only or sub frames also.
+  /// `true` for main frame only, `false` for all frames including sub frames.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Android on Wry:** The Android WebView does not provide an API for initialization scripts,
+  ///   so we prepend them to each HTML head. They are only implemented on custom protocol URLs.
+  pub initialization_scripts: Vec<InitializationScript>,
```
Legend-Master, line 208: @amrbashir I just realized that `WebviewAttributes` is re-exported in tauri, this will be a breaking change, but I think we did something like that before by adding a new field `javascript_disabled`, is this ok to do? (to be honest I don't really know why it's re-exported)

crates/tauri-runtime/src/webview.rs
```diff
@@ -307,10 +318,49 @@ impl WebviewAttributes {
...
+  /// This is executed on all frames, main frame and also sub frames.
+  /// If you only want to run it in the main frame, use [Self::initialization_script] instead.
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Android on Wry:** The Android WebView does not provide an API for initialization scripts,
+  ///   so we prepend them to each HTML head. They are only implemented on custom protocol URLs.
+  ///
+  /// - On **macOS** and **iOS** you can use this to overwrite browser apis to make them inacessible even in iframes.
+  ///   You can disable WebRTC this way, for example.
```
Legend-Master:
```suggestion
```

Let's remove these, they're not specific to macos and ios and I think it's obvious enough what an initialization script can do

crates/tauri-runtime/src/webview.rs
Simon-Laux: isn't the direct access to the webview api experimental and behind the unstable feature flag?

amrbashir: this type should not be exposed, it is internal

amrbashir: it may have been exposed by accident, we should fix that.

amrbashir: anyways it is fine to modify I guess, since there is no public API that uses it. cc @lucasfernog

.changes/init-script-on-all-frames.md
```diff
@@ -0,0 +1,9 @@
+---
+tauri: minor:feat
+tauri-runtime: minor:feat
+---
+
+- add API to run initialisation scripts on all frames
```
Legend-Master:
```suggestion
- add API to run initialization scripts on all frames
```

crates/tauri-runtime/src/webview.rs
```diff
@@ -197,7 +197,18 @@ impl<T: UserEvent, R: Runtime<T>> PartialEq for DetachedWebview<T, R> {
 pub struct WebviewAttributes {
   pub url: WebviewUrl,
   pub user_agent: Option<String>,
-  pub initialization_scripts: Vec<String>,
+  /// A list of initialization javascript scripts to run when loading new pages.
+  /// When webview load a new page, this initialization code will be executed.
+  /// It is guaranteed that code is executed before `window.onload`.
+  ///
+  /// Second parameter represents if script should be added to main frame only or sub frames also.
+  /// `true` for main frame only, `false` for all frames including sub frames.
```
Legend-Master:
```suggestion
```

Legend-Master (changes requested):

Legend-Master (approved):
Look good to me, just wait for lucas to confirm if it's ok to modify `WebviewAttributes`

crates/tauri-runtime/src/webview.rs
lucasfernog: agreed

crates/tauri/src/webview/webview_window.rs
```diff
@@ -765,9 +765,12 @@ impl<R: Runtime, M: Manager<R>> WebviewWindowBuilder<'_, R, M> {
   /// Adds the provided JavaScript to a list of scripts that should be run after the global object has been created,
   /// but before the HTML document has been parsed and before any other script included by the HTML document is run.
   ///
-  /// Since it runs on all top-level document and child frame page navigations,
+  /// Since it runs on all top-level document navigastions (and also child frame page navigations, if you set `run_only_on_main_frame` to false),
```
Legend-Master:
```suggestion
  /// Since it runs on all top-level document navigations (and also child frame page navigations, if you set `run_only_on_main_frame` to false),
```

Legend-Master (changes requested):

Legend-Master (approved):
Thanks for the patience 🙃

PR #13079, opened 2025-03-26
feat(cli): allow for toml files in --config arg, closes #7859

### LINKED ISSUE #7859: [feat] Formats other than JSON for additional configuration files
### Describe the problem

Currently, Tauri also supports JSON5 and TOML file formats for the configuration file. However, the only format supported for additional configuration files to merge by using the `--config` CLI option seems to be JSON. That's what the CLI help says:

```
-c, --config <CONFIG>           JSON string or path to JSON file to merge with tauri.conf.json
```

And it's also consistent with the experiments that I did.

From a CLI user standpoint, it would make sense to accept also JSON5 and TOML file formats through the `--config` CLI option.

### Alternatives considered

_No response_

### Additional context

I wanted to use one of these formats, mainly to have comments in the configuration files, but stuck with JSON because I have additional configuration files and didn't want to mix formats.

Pietagorh:
closes #7859 
## Why
1. can already use toml for the base config, just bringing parity to the --config feat
2. pretty small change

## Tests
- [x] `cargo test`
- [x] `cargo clippy`

### DIFF
--- a/crates/tauri-cli/Cargo.toml
+++ b/crates/tauri-cli/Cargo.toml
@@ -51,6 +51,7 @@ tauri-bundler = { version = "2.3.0", default-features = false, path = "../tauri-
 colored = "2"
 serde = { version = "1", features = ["derive"] }
 serde_json = { version = "1", features = ["preserve_order"] }
+json5 = "0.4"
 notify = "8"
 notify-debouncer-full = "0.5"
 shared_child = "1"
--- a/crates/tauri-cli/src/lib.rs
+++ b/crates/tauri-cli/src/lib.rs
@@ -63,10 +63,22 @@ impl FromStr for ConfigValue {
     } else {
       let path = PathBuf::from(config);
       if path.exists() {
-        Ok(Self(serde_json::from_str(
-          &read_to_string(&path)
-            .with_context(|| format!("invalid configuration at file {config}"))?,
-        )?))
+        let raw = &read_to_string(&path)
+          .with_context(|| format!("invalid configuration at file {config}"))?;
+        match path.extension() {
+          Some(ext) if ext == "toml" => Ok(Self(::toml::from_str(raw)?)),
+          Some(ext) if ext == "json5" => Ok(Self(::json5::from_str(raw)?)),
+          // treat all other extensions as json
+          _ => Ok(Self(
+            // from tauri-utils/src/config/parse.rs:
+            // we also want to support **valid** json5 in the .json extension
+            // if the json5 is not valid the serde_json error for regular json will be returned.
+            match ::json5::from_str(raw) {
+              Ok(json5) => json5,
+              Err(_) => serde_json::from_str(raw)?,
+            },
+          )),
+        }
       } else {
         anyhow::bail!("provided configuration path does not exist")
       }
--- a/.changes/toml-json5-configs-in-cli.md
+++ b/.changes/toml-json5-configs-in-cli.md
@@ -1,0 +1,6 @@
+---
+tauri-cli: minor:enhance
+"@tauri-apps/cli": minor:enhance
+---
+
+Add support for passing TOML and JSON5 config files to `--config` arg
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8537,6 +8537,7 @@ dependencies = [
  "insta",
  "itertools 0.13.0",
  "json-patch 3.0.1",
+ "json5",
  "jsonrpsee",
  "jsonrpsee-client-transport",
  "jsonrpsee-core",
--- a/crates/tauri-cli/src/build.rs
+++ b/crates/tauri-cli/src/build.rs
@@ -45,7 +45,7 @@ pub struct Options {
   /// Skip the bundling step even if `bundle > active` is `true` in tauri config.
   #[clap(long)]
   pub no_bundle: bool,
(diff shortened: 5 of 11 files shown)


crates/tauri-cli/src/lib.rs
```diff
@@ -63,10 +63,17 @@ impl FromStr for ConfigValue {
...
+            .with_context(|| format!("invalid configuration at file {config}"))?;
+        if let Some(extension) = path.extension() {
+          match extension.to_str().expect("invalid unicode in file extension") {
+            "toml" => Ok(Self(::toml::from_str(raw)?)),
+            "json" | "json5" => Ok(Self(serde_json::from_str(raw)?)),
+            _ => anyhow::bail!("unsupported file type")
+          }
+        } else {
+          anyhow::bail!("invalid file path (requires file extension)")
+        }
```
amrbashir: we can always default to JSON, and for Json5, we need to use `json5` crate
```suggestion
        match path.extension() {
          Some(ext) if ext == "toml" => Ok(Self(::toml::from_str(raw)?)),
          Some(ext) if ext == "json5" => Ok(Self(json5::from_str(raw)?)),
          // default to JSON
          _ => Ok(Self(serde_json::from_str(raw)?))
        }
```

.changes/toml-config-patch-in-cli.md
```diff
@@ -0,0 +1,5 @@
+---
+tauri-cli: minor:enhance
+---
+
+File paths to TOML files can now be used in the cli --config arg
```
amrbashir:
```suggestion
Add support for passing TOML and JSON5 config files to `--config` arg
```

.changes/toml-config-patch-in-cli.md
```diff
@@ -0,0 +1,5 @@
+---
+tauri-cli: minor:enhance
```
amrbashir:
```suggestion
tauri-cli: minor:enhance
"@tauri-apps/cli": minor:enhance
```

amrbashir (changes requested):

crates/tauri-cli/src/lib.rs
Pietagorh: right!
isn't the usual behavior to try parsing .json's with json5 as well?

amrbashir: if a json5 file has comments in it, serde_json will not be able to parse it.

amrbashir: I might have misunderstood your question, in that case yeah we could just use `json5` for `.json` files as well.

amrbashir: you also need to update the documentation option to mention JSON5 support as well.

Pietagorh: > I might have misunderstood your question, in that case yeah we could just use `json5` for `.json` files as well.

From what i've seen in tauri-utils, we parse .json with serde, if we get an error we try with json5 and if json5 gives an error we return the serde error
I'll check again later tho

amrbashir: yeah, I guess we can do that as well

Pietagorh:
Updated:
- changed what's been requested
- added the feature flag stuff (`config-toml` and `config-json5`)
**i couldn't confirm that tauri-macros actually correctly dispatches the config features cause i don't know how to build the full package** (passing the individual features works tho)

crates/tauri-cli/Cargo.toml
```diff
@@ -149,3 +150,5 @@ native-tls = [
 ]
 native-tls-vendored = ["native-tls", "tauri-bundler/native-tls-vendored"]
 rustls = ["tauri-bundler/rustls", "cargo-mobile2/rustls", "ureq/rustls"]
+config-json5 = ["json5"]
+config-toml = []
```
amrbashir: no need to add these features. This is the CLI and not shipped with the final user app so we can always include them.

crates/tauri-macros/Cargo.toml
```diff
@@ -22,11 +22,12 @@ syn = { version = "2", features = ["full"] }
...
+tauri-cli = { version = "2.4.0", path = "../tauri-cli" }
 
 [features]
 custom-protocol = []
 compression = ["tauri-codegen/compression"]
 isolation = ["tauri-codegen/isolation"]
-config-json5 = ["tauri-codegen/config-json5", "tauri-utils/config-json5"]
-config-toml = ["tauri-codegen/config-toml", "tauri-utils/config-toml"]
+config-json5 = ["tauri-codegen/config-json5", "tauri-cli/config-json5", "tauri-utils/config-json5"]
+config-toml = ["tauri-codegen/config-toml", "tauri-cli/config-toml", "tauri-utils/config-toml"]
```
amrbashir: these need to be reverted, tauri-macros doesn't interact with tauri-cli at all.

tauri-cli is just a binary on its own.

amrbashir (changes requested):

crates/tauri-cli/Cargo.toml
Pietagorh: oh ok thanks

amrbashir (approved):

amrbashir:
Thank you

Pietagorh:
> Thank you

No, thank you lol you did most of it (thx for showing me how to do it cleanly as well x))

PR #13136, opened 2025-04-04
fix: channel callback never cleaned up from `window`

### LINKED ISSUE #13133: [bug] Listening to Tauri channel events creates memory leak
### Describe the bug

Using a [Channel](https://v2.tauri.app/develop/calling-frontend/#channels) to listen to events sent from the Rust backend creates a memory leak in the web view.

### Reproduction

For example:

```vue title="ExampleComponent.vue"
<script lang="ts" setup>
// imports are omitted

const onProgress = new Channel<unknown>();

onProgress.onmessage = (progress: unknown) => {
   // Handle progress events...
};

const result = await invoke('some_tauri_command', { onProgress });

// ... do stuff with the result ...

// At this point, if this Vue component is destroyed/unmounted, for example when you
// navigate to another view, you might expect the `onProgress` reference to be garbage
// collected along with the `onmessage` callback/closure.
//
// However, Tauri keeps a reference to the `onmessage` closure attached to the `window`
// object indefinitely. This means that any references in scope of the `onmessage` closer
// when it was created are also kept in memory. This would include the Vue component
// itself.
</script>
```

Here's the code that creates the leak:

   1. [Tauri calls an internal function called `transformCallback`, passing in the `onmessage`
      callback function.][1]
   2. [`transformCallback` creates a random UUID and then saves the callback to the `window` object, never to be deleted and providing no way to delete it yourself.][2]
      * _Callback could be deleted if [`once` parameter passed to `transformCallback`][3] however it [is not][4]. This would be undesirable though since `onmessage` handlers can be called multiple times while an IPC function is running._

[1]: https://github.com/tauri-apps/tauri/blob/f805061d1152bc4790dbdb9475a506afcdd1de75/packages/api/src/core.ts#L89
[2]: https://github.com/tauri-apps/tauri/blob/f805061d1152bc4790dbdb9475a506afcdd1de75/crates/tauri/scripts/core.js#L27
[3]: https://github.com/tauri-apps/tauri/blob/f805061d1152bc4790dbdb9475a506afcdd1de75/crates/tauri/scripts/core.js#L23
[4]: https://github.com/tauri-apps/tauri/blob/f805061d1152bc4790dbdb9475a506afcdd1de75/packages/api/src/core.ts#L111

### Full `tauri info` output

```text
[-] Packages
    - tauri 🦀: 2.2.5
    - tauri-build 🦀: 2.0.5
    - wry 🦀: 0.48.1
    - tao 🦀: 0.31.1
    - tauri-cli 🦀: 2.4.0
    - @tauri-apps/api : 2.2.0 (outdated, latest: 2.4.1)
    - @tauri-apps/cli : 2.3.1 (outdated, latest: 2.4.1)

[-] Plugins
    - tauri-plugin-upload 🦀: 2.2.1
    - @tauri-apps/plugin-upload : 2.2.1
    - tauri-plugin-opener 🦀: 2.2.5
    - @tauri-apps/plugin-opener : 2.2.5 (outdated, latest: 2.2.6)
    - tauri-plugin-log 🦀: 2.2.1
    - @tauri-apps/plugin-log : 2.2.3 (outdated, latest: 2.3.1)
    - tauri-plugin-http 🦀: 2.3.0
    - @tauri-apps/plugin-http : 2.3.0 (outdated, latest: 2.4.3)
    - tauri-plugin-sql 🦀: 2.2.0
    - @tauri-apps/plugin-sql : 2.2.0
    - tauri-plugin-os 🦀: 2.2.0
    - @tauri-apps/plugin-os : 2.2.0 (outdated, latest: 2.2.1)
    - tauri-plugin-store 🦀: 2.2.0
    - @tauri-apps/plugin-store : 2.2.0
    - tauri-plugin-fs 🦀: 2.2.0
    - @tauri-apps/plugin-fs : 2.2.0 (outdated, latest: 2.2.1)
```

### Additional context

**Workaround: Delete the `onmessage` property from the `channel` object when the Vue component is
destroyed, or when the channel is no longer used.**

```ts{11-13,15,19-21}
import { onBeforeUnmount } from 'vue';
// Other imports omitted

const onProgress = new Channel<unknown>();

onProgress.onmessage = (progress: unknown) => {
   // Handle progress events...
};

// TODO: This is only temporary until Tauri fixes the Channel memory leak issue
function preventMemoryLeak() {
   delete onProgress.onmessage;
}

onBeforeUnmounted(preventMemoryLeak);

try {
   const result = await invoke('some_tauri_command', { onProgress });
} finally {
   preventMemoryLeak();
}
```

Legend-Master:
Fix #13133

Also:

- added a feature that allows passing the callback as the parameter of constructor of `Channel` so you can use it like this `new Channel((message) => console.log(message))`
- only go through the fetch process if the data is large enough for it to be faster


Legend-Master:
This change will require syncing the version of rust crate `tauri` and js package `@tauri-apps/api`, since this is marked as a minor I think we can do this right?

FabianLars:
hmm, yeah. we've communicated that minor versions are synced a few times (even if we don't fully follow it) so it should be fine. not the first time in 2.x either 🤷

crates/tauri/src/ipc/channel.rs
```diff
@@ -97,44 +117,68 @@ impl FromStr for JavaScriptChannelId {
...
-        .0
-        .lock()
-        .unwrap()
-        .insert(data_id, body);
 
-      webview.eval(format!(
-        "window.__TAURI_INTERNALS__.invoke('{FETCH_CHANNEL_DATA_COMMAND}', null, {{ headers: {{ '{CHANNEL_ID_HEADER_NAME}': '{data_id}' }} }}).then((response) => window['_' + {}]({{ message: response, id: {i} }})).catch(console.error)",
-        callback_id.0
-      ))?;
+        // Don't go through the fetch process if the payload is small
```
FabianLars: finally someone who agrees with me on that approach 😂

Legend-Master: This should significantly reduce the load in places like sending updater's download progress

lucasfernog:
sample benchmark script:
```rust
pub fn spam(channel: Channel<tauri::ipc::InvokeResponseBody>) -> tauri::Result<()> {
  for i in 1..=255 {
    channel.send(tauri::ipc::InvokeResponseBody::Raw(
      std::iter::repeat(i).take(1024).collect(),
    ))?;
  }
  Ok(())
}
```
(change 1024 to whatever you want to test)
```js
channel = new __TAURI__.core.Channel();
i = 0;
channel.onmessage = (m) => {console.log(m); i++; if (i === 255) { console.timeEnd('spam') } };
console.time('spam');
__TAURI__.core.invoke('spam', { channel });
```

lucasfernog (approved):

crates/tauri/src/ipc/channel.rs
```diff
@@ -97,44 +118,73 @@ impl FromStr for JavaScriptChannelId {
...
+          // Don't go through the fetch process if the payload is small
+          InvokeResponseBody::Json(string) if string.len() < MAX_JSON_DIRECT_EXECUTE_THRESHOLD => {
+            webview.eval(format!(
+              "window['_{callback_id}']({{ message: {string}, index: {current_index} }})"
+            ))?;
+          }
+          InvokeResponseBody::Raw(bytes) if bytes.len() < MAX_RAW_DIRECT_EXECUTE_THRESHOLD => {
+            webview.eval(format!(
+              "window['_{callback_id}']({{ message: {}, index: {current_index} }})",
+              serde_json::to_string(&bytes)?,
```
WSH032, line 159: @Legend-Master This is a BREAKING CHANGE.

I need to use channels to send binary data:

- <https://github.com/pytauri/pytauri/blob/c738a5949859455e89e8095df1d0726831c29f1a/crates/tauri-plugin-pytauri/guest-js/src/index.ts#L94-L97>
- <https://github.com/pytauri/pytauri/blob/c738a5949859455e89e8095df1d0726831c29f1a/crates/pytauri-core/src/ext_mod_impl/ipc.rs#L301>

Previously, the received data was an `ArrayBuffer`, but now it has become an `Array`.

Legend-Master: That's pretty much the only part not written by me in this PR (ebb919b) 😅, anyways, I'll take a look tomorrow

WSH032: Oh, my bad, I didn't notice that 😂

PR #13139, opened 2025-04-04
fix: Correct target_os cfg on drag_and_drop method

p4rtridge:
### Problem

Currently, the `target_os` cfg on the `drag_and_drop` method is incorrect, causing compile-time errors when building on platforms other than Windows.

### Solution
Change from `#[cfg(windows)]` to `#[cfg(target_os = "windows")]` to resolve the issue.

### DIFF
--- a/crates/tauri/src/webview/webview_window.rs
+++ b/crates/tauri/src/webview/webview_window.rs
@@ -690,7 +690,7 @@ impl<'a, R: Runtime, M: Manager<R>> WebviewWindowBuilder<'a, R, M> {
   }
 
   /// Enables or disables drag and drop support.
-  #[cfg(windows)]
+  #[cfg(target_os = "windows")]
   #[must_use]
   pub fn drag_and_drop(mut self, enabled: bool) -> Self {
     self.window_builder = self.window_builder.drag_and_drop(enabled);


amrbashir:
`#[cfg(windows)]` is just a sugar for `#[cfg(target_family = "windows")]` which includes `#[cfg(target_os = "windows")]` so they are basically the same, what compile errors do you have?

p4rtridge:
> `#[cfg(windows)]` is just a sugar for `#[cfg(target_family = "windows")]` which includes `#[cfg(target_os = "windows")]` so they are basically the same, what compile errors do you have?

I got `method not found` error while building application on platforms other than Windows. If you saying that `#[cfg(windows)]` is just a sugar syntax then maybe `#[cfg(windows]` has a problem.

lucasfernog:
To use `drag_and_drop` on other platforms you'd actually need to remove the `#[cfg]` attribute instead.

lucasfernog:
so i actually recommend you do this:
```rust
let mut builder = tauri::WebviewWindowBuilder::new(/* insert your args here */);
#[cfg(windows)]
{
  builder = builder.drag_and_drop(false);
}
let window = builder.build()?;
```
note that currently you can't disable drag and drop support on macOS and Linux. That's why this function is Windows-only.

PR #13145, opened 2025-04-05
fix: make `isTauri` runtime-unrelated

typed-sigterm:
This is useful when running test without mocking DOM.


typed-sigterm:
Lint should be fixed now

lucasfernog (approved):
tbh you should probably mock the isTauri function on your end, but i'll take this (probably there's some use case i'm missing where you can't really do that?)

PR #13146, opened 2025-04-05
chore(example): migrate API example to svelte 5

Legend-Master:
I got a whole bunch of warnings everytime I do dev/build here, so I eventually decided to fix it


lucasfernog (approved):
<3

PR #13150, opened 2025-04-05
fix(core): missing `core:` in referenced commands

Legend-Master:
Fix missing `core:` in referenced commands in ACL error messages


lucasfernog (approved):

lucasfernog:
clever

PR #13163, opened 2025-04-07
chore: update tao, wry, windows, webview2-com

lucasfernog:
tao 0.33 closes #13164
wry 0.51 closes #13165
windows 0.61 closes #13036
webview2-com 0.37 closes #13035

### DIFF
--- a/.changes/update-tao.md
+++ b/.changes/update-tao.md
@@ -1,0 +1,5 @@
+---
+"tauri-runtime-wry": minor:deps
+---
+
+Update tao to 0.33.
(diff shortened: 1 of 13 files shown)


Legend-Master (dismissed):

Legend-Master:
Maybe we can also migrate `InitializationScript` related functions and docs in this repo to use the wry one before we publish them in the new minor?

Legend-Master (approved):

lucasfernog:
@Legend-Master i think changing `&str` to `impl Into<String>` can be a breaking change in some weird scenarios - like using `into()` on your own since we're changing from a known type to an impl

Legend-Master:
Right but I think the docs suggests this should be a minor?

> https://doc.rust-lang.org/cargo/reference/semver.html#fn-generalize-compatible

PR #13182, opened 2025-04-08
Feature/binary patch bundle type

kandrelczyk:


lucasfernog:
maybe we should use https://github.com/denoland/sui instead

lucasfernog:
btw what's exactly the goal of this PR? to support WiX/NSIS at the same time? changing the updater target is a breaking change ([cargo-packager](https://github.com/crabnebula-dev/cargo-packager/) fixes this by adding a package `format` field in the update JSON)

kandrelczyk:
@lucasfernog sorry, this PR is a mess so I turned it into a draft. I have a cleaner one coming, will try to create it today. 

I will link the original issue there. Closing this one for now.

amrbashir:
> btw what's exactly the goal of this PR? to support WiX/NSIS at the same time? changing the updater target is a breaking change ([cargo-packager](https://github.com/crabnebula-dev/cargo-packager/) fixes this by adding a package `format` field in the update JSON)


This doesn't break anything, it merely embeds metadata that we can read in the updater plugin and perform certain actions/cleanups when updating.

This could also fix some guessing work we do regarding updates, like when updating .deb or .rpm, we need to know which package manager to use, instead of guessing and falling back if our guess was incorrect

lucasfernog:
yeah i meant just the target change would be breaking, adding the format would be ok

FabianLars:
The target change is also not breaking if we add the fallback logic. At least that was the plan when we last talked about this.

PR #13203, opened 2025-04-11
ci: remove redundant corepack enable

trivikr:
The command `corepack enable` is not required to be run after globally installing corepack

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -34,9 +34,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - run: |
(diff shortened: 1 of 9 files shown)


Legend-Master (approved):

Legend-Master:
https://github.com/tauri-apps/tauri/pull/12604#issuecomment-2640239679 seems to be fixed right?

lucasfernog:
that's weird, just by installing it we're already enabling it?? ok i guess :D

FabianLars:
tbh I'd rather remove the use of corepack altogether in favor of the pnpm setup action (iirc their recommendation as well) and the addition of npm i was just meant as a quick hotfix.

PR #13209, opened 2025-04-11
feat: bundle type detection at runtime via binary patching

kandrelczyk:
This is first PR to address https://github.com/tauri-apps/plugins-workspace/issues/2277

Thanks to this the updater plugin will be able to determine which installer to download. 

There is a breaking change here requiring updater plugin and tauri-cli to be compatible since we need __TAURI_BUNDLE_TYPE variable to be present in the binary. In case the variable is not found error is reported.  Not sure if it's ok or if we need a better solution here.


crates/tauri-bundler/src/bundle.rs
```diff
@@ -92,11 +114,21 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
...
+    .iter()
+    .find(|b| b.main())
+    .expect("Main binary missing in settings");
+
   let mut bundles = Vec::<Bundle>::new();
   for package_type in &package_types {
     // bundle was already built! e.g. DMG already built .app
     if bundles.iter().any(|b| b.package_type == *package_type) {
       continue;
+    } // REMOVE THIS?
```
kandrelczyk: Not sure about this. This only applies to DMG, right? We don't want to skip anything on linux or windows  as the binary needs to be patched for each type anyway.

.gitignore
```diff
@@ -1,54 +1,55 @@
...
+
+# miscellaneous
+/.vs
+.DS_Store
+.Thumbs.db
+*.sublime*
+.idea
+debug.log
+TODO.md
+.aider*
```
kandrelczyk, line 55: Sorry for those weird patches. I think this is line end issue because I worked on both windows and linux. I only added '.aider*' here.

crates/tauri-bundler/src/error.rs
```diff
@@ -1,124 +1,139 @@
...
+  Glob(#[from] glob::GlobError),
+  /// Failed to parse the URL
+  #[error("`{0}`")]
+  UrlParse(#[from] url::ParseError),
+  /// Failed to validate downloaded file hash.
+  #[error("hash mismatch of downloaded file")]
+  HashError,
+  /// Failed to parse binary
+  #[error("Binary parse error: `{0}`")]
+  BinaryParseError(String),
```
kandrelczyk: Same line end issue. New errors are here.

.gitignore
Legend-Master: You probably want to set `autocrlf` to `true` or `input`

> `git config --global core.autocrlf true`/`git config --global core.autocrlf input`
> https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#:~:text=with%20these%20issues.-,core.autocrlf,-If%20you%E2%80%99re%20programming

crates/tauri-bundler/src/bundle.rs
```diff
@@ -13,6 +13,28 @@ mod settings;
 mod updater_bundle;
 mod windows;
 
+impl From<goblin::error::Error> for crate::error::Error {
```
lucasfernog: this can be done inside crate::error::Error::BinaryParseError by leverating thiserror `#[from]` attribute

lucasfernog:
i think we should consider using https://github.com/denoland/sui it has an easy to use API

kandrelczyk:
> i think we should consider using https://github.com/denoland/sui it has an easy to use API

Definitely looks simpler than goblin. Wish I knew about it before :)

Should I just change the patching implementation?

kandrelczyk:
@lucasfernog I've pushed some of the suggested changes. I decided no to use sui as this would add dependency to the app, not just the builder. If there anything else missing here?

lucasfernog (approved):
awesome

Legend-Master:
Didn't look at it yet, but it seems like the ubuntu build is failing https://github.com/tauri-apps/tauri/actions/runs/16120739808/job/45485616619

kandrelczyk:
@Legend-Master I've just checked and the tests I added to `plugins-workspace/plugins/updater/tests/app-updater` still pass. I will check the build ASAP.

kandrelczyk:
@Legend-Master pushed the fix. Turns out my tests used `--debug` but the failing test don't. In release mode the bundle type variable was stripped even though it's referenced in public method. I've added `#[used]` to it and release builds work now. 

...Now I see this PR was closed already. Do I have to create a new PR with the fix?

Legend-Master:
> ...Now I see this PR was closed already. Do I have to create a new PR with the fix?

Yeah, that's how we usually do it

kandrelczyk:
@Legend-Master Submitted here: https://github.com/tauri-apps/tauri/pull/13808

kittuov:
Hey, Looks like the patcher is modifying the binary after code signing. This is rendering the signing to be invalid. maybe move the signing to after binary is patched?

kandrelczyk:
@kittuov Hey, can you point me to the place were we sign the binary? I'm only familiar wit generating the signatures files for the bundles.

Legend-Master:
> @kittuov Hey, can you point me to the place were we sign the binary? I'm only familiar wit generating the signatures files for the bundles.

It was signing the main executable on Windows (you sign both the binaries and the bundles on Windows instead of just the bundles on Linux, and we don't do binary patching on macOS yet), should be fixed in #13870 (we sign the binaries first before generating all the bundles, but now we need to sign them after the binary patching)

Thanks for taking a look anyways 🙃

Issue #12550, opened 2025-01-28
[RFC] Drop Windows 7 support


FabianLars:
We all know it's coming and at least as far as i am concerned it was planned for v3. Now that we're rushing v3 for gtk4 i'm not sure sure anymore. I'll still put this into the v3 tracking board for now but we can move it into v4 later if we prefer.

For community members, please share your opinions! If you're fine with dropping Windows 7 support a 👍 reaction is enough, if you want/need Windows 7 support, we'd appreciate hearing more about your use-case (plus a 👎 reaction) :)

exbit:
I am in need of windows 7 support. But at the moment I can't even run version 2 of Tauri on Windows SP1. The build runs successfully, but when I run it on Windows 7 I always get an error: 0xc0000005. Is there a detailed instruction on how to properly make a build for Windows 7?

```
[✔] Environment
    - OS: Windows 10.0.26100 x86_64 (X64)
    ✔ WebView2: 131.0.2903.146
    ✔ MSVC:
        - Visual Studio Community 2019
        - Visual Studio Enterprise 2022
    ✔ rustc: 1.86.0-nightly (bf1b174e7 2025-01-28)
    ✔ cargo: 1.86.0-nightly (cecde95c1 2025-01-24)
    ✔ rustup: 1.27.1 (54dd3d00f 2024-04-24)
    ✔ Rust toolchain: nightly-i686-pc-windows-msvc (default)
    - node: 23.6.0
    - pnpm: 9.15.4
    - yarn: 1.22.22
    - npm: 10.9.2
    - bun: 1.2.0

[-] Packages
    - tauri 🦀: 2.2.5
    - tauri-build 🦀: 2.0.5
    - wry 🦀: 0.48.1
    - tao 🦀: 0.31.1
    - tauri-cli 🦀: 2.2.5
    - @tauri-apps/api : 2.2.0
    - @tauri-apps/cli : 2.2.7

[-] Plugins
    - tauri-plugin-opener 🦀: 2.2.5
    - @tauri-apps/plugin-opener : 2.2.5

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:1420/
    - framework: Vue.js
    - bundler: Vite
```

FabianLars:
@exbit You need to build the app with rust 1.77.2 (may require a few `cargo update` commands to make all crates build). If this doesn't help, please open a seperate issue or discussion.

[...]

PR #13221, opened 2025-04-13
feat: bump MSRV to 1.90

lucasfernog:
previous MSRV change was introduced by #11205
to follow the ecosystem and keep our dependencies up to date we must update our MSRV

currently our agreed standard is to keep MSRV to "latest - 3", which currently is 1.83


Legend-Master:
Do we want to wait for v3? Like we sort of promised Windows 7 support for v2?

FabianLars:
> currently our agreed standard is to keep MSRV to "latest - 3", which currently is 1.83

When did we agree to that? I always felt like the wg disagreed with me when i talked about this x)

This indeed virtually drops windows 7/8 support because the win7 rust target doesn't work in tauri (at least not for me and the others who tried it with me).

Maybe we should follow through with what we've semi-promised and rush gtk4 migration and therefore tauri v3.

fwiw https://github.com/tauri-apps/tauri/issues/12550

lucasfernog:
let's be safe and leave this for v3+, I don't want to make people angry right now 😂

Tunglies:
Is more aggressive version etc. 1.84 and 1.85 would be consider?

FabianLars:
Yes, when we will merge this PR it will likely be higher than 1.83 - it was just the most recent & compatible when the PR was opened

Legend-Master:
The clippy error will be fixed in #15586, let's just focus on the version bumps in this PR.

Legend-Master (approved):

PR #13227, opened 2025-04-14
fix(core): use `Headers` in `sendIpcMessage`

### LINKED ISSUE #13223: [bug] `invoke` does not handle `Headers` correctly
### Describe the bug

<https://github.com/tauri-apps/tauri/blob/8a1d49082033cc0bae2258fed3c3aeb539665acf/crates/tauri/scripts/ipc-protocol.js#L38>

`...options.headers` works for `{ foo: "bar" }` but does not work for a `Headers` object.

### Reproduction

```typescript
import { invoke } from "@tauri-apps/api/core";

// work ⭐
await invoke("foo", undefined, {
  headers: {
    foo: "bar"
  }
});

// not work ❌
const headers = new Headers({ foo: "bar" });
await invoke("foo", undefined, {
  headers
});
```

Can't receive `foo: "bar"` header in backend.

### Expected behavior

_No response_

### Full `tauri info` output

```text
[✔] Environment
    ✔ WebView2: 135.0.3179.73
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (environment override by RUSTUP_TOOLCHAIN)
    - node: 20.11.1
    - pnpm: 9.10.0
    - npm: 10.2.4

[-] Packages
    - tauri 🦀: 2.4.0
    - tauri-build 🦀: 2.1.0
    - wry 🦀: 0.50.5
    - tao 🦀: 0.32.8
    - tauri-cli 🦀: 2.2.7
    - @tauri-apps/api : 2.1.1 (outdated, latest: 2.4.1)
    - @tauri-apps/cli : 2.2.7 (outdated, latest: 2.4.1)

[-] Plugins
    - @tauri-apps/plugin-opener : 2.2.2 (outdated, latest: 2.2.6)
    - tauri-plugin-clipboard-manager 🦀: 2.2.2
    - @tauri-apps/plugin-clipboard-manager : not installed!
    - tauri-plugin-fs 🦀: 2.2.0
    - tauri-plugin-notification 🦀: 2.2.2

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ./frontend
    - bundler: Vite
```

### Stack trace

```text

```

### Additional context

I have not checked the case for `isolation` mode because I don't quite understand how to use this mode (its documentation is not very clear 😵).

Legend-Master:
Fix #13223

This also makes it easier to debug cases like https://github.com/tauri-apps/plugins-workspace/pull/1965 where non-ascii characters get passed in to it and it silently replaces them instead of throwing an error

Also, this should make override header keys case insensitive


WSH032:
Thanks! I can confirm it fixed the issue.

crates/tauri/scripts/ipc-protocol.js
```diff
@@ -27,16 +27,25 @@
...
+      })
+      if (options?.headers) {
+        const overrideHeaders =
+          options.headers instanceof Headers
+            ? options.headers.entries()
+            : Object.entries(options.headers)
+        for (const [key, value] of overrideHeaders) {
+          headers.set(key, value)
+        }
+      }
```
WSH032:
```suggestion
      const headers = new Headers(options?.headers);
      headers.set("Content-Type", contentType);
      headers.set("Tauri-Callback", callback);
      headers.set("Tauri-Error", error);
      headers.set("Tauri-Invoke-Key", __TAURI_INVOKE_KEY__);
```

I'm just a bit curious, why not implement it this way? It seems more elegant.

I guess it's to preserve previous behavior (allowing users to override preset headers)? But overriding these headers seems like an unreasonable behavior.

Legend-Master: > I guess it's to preserve previous behavior (allowing users to override preset headers)?

Yeah, that's the reason why I did it like this

> But overriding these headers seems like an unreasonable behavior.

That's true though

I'm not entirely sure if it's supposed to be able to override these or not, cc @lucasfernog

lucasfernog (dismissed):
awesome

crates/tauri/scripts/ipc-protocol.js
lucasfernog: welllll nice catch.. that's used internally so we shouldn't allow them to be overwritten (it would break the IPC if you did so - never running or responding depending on which value you change)

lucasfernog: @Legend-Master can you take his suggestion and change `new Headers(options?.headers)` to `new Headers((options && options.headers) || {})` (null cannot be used in that constructor)

lucasfernog: pushed :)

PR #13240, opened 2025-04-15
fix(ci): ubuntu 20.04 no longer exists

lucasfernog:


FabianLars:
btw i have a PR lined up (which i wanted to push like 2 weeks ago...) that'll use zig as the linker so we can switch to ubuntu-latest for the cli and still support old glibc

FabianLars (approved):

PR #13261, opened 2025-04-18
feat(core): add capability `devOnly` field

lucasfernog:
a dev environment might need particular scope configurations (such as allowing a localhost API). This change allows marking a whole capability file as "dev only"

### DIFF
--- a/.changes/capability-dev-only.md
+++ b/.changes/capability-dev-only.md
@@ -1,0 +1,7 @@
+---
+"tauri-utils": minor:feat
+"tauri-codegen": minor:feat
+"tauri": minor:feat
+---
+
+Added `devOnly` option to the capability that lets you define a capability only for development - for instance to allow local APIs in scopes.
--- a/crates/tauri-cli/config.schema.json
+++ b/crates/tauri-cli/config.schema.json
@@ -1409,6 +1409,10 @@
           "items": {
             "$ref": "#/definitions/Target"
           }
+        },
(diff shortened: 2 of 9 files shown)


FabianLars:
What about buildOnly/prodOnly (whatever you wanna call it)?

lucasfernog:
> What about buildOnly/prodOnly (whatever you wanna call it)?

the main idea of devOnly is that you could enable for instance connecting to a localhost server in dev mode only, not affecting the security scope of a prod app. I don't see value in the opposite - if you allow connecting to the real API backend for instance, that doesn't impact the security of the dev mode.

amrbashir:
How about just allowing to pass a capability from CLI, this way you can cover the dev only situation and potentially other environment specific capabilities


The problem is all capabilities files are active by default and so we need to pair this with should a config option to select which capabilities file should be loaded

AlexanderBastowros:
Please correct me if I am wrong, can this not already be done by setting the devUrl in the `tauri.config.json`?

https://v2.tauri.app/reference/config/#devurl

FabianLars:
i'm not sure how that's related to capabilities so i can't correct you 😅 The closest thing related to that is Lucas' `such as allowing a localhost API` example but he's not talking about the url that's loaded in the webview (it's also just one example)

AlexanderBastowros:
I meant it as an alternative, but I see your point. 👍

PR #13268, opened 2025-04-21
fix(core): raw channel message type regression

Legend-Master:
Fix https://github.com/tauri-apps/tauri/pull/13136#discussion_r2051756480

Fix a regression introduced in #13138 that made the raw type messages received from `Channel.onmessage` became `number[]` instead of `ArrayBuffer` when that message is small

### DIFF
--- a/crates/tauri/src/ipc/channel.rs
+++ b/crates/tauri/src/ipc/channel.rs
@@ -154,9 +154,9 @@ impl JavaScriptChannelId {
             ))?;
           }
           InvokeResponseBody::Raw(bytes) if bytes.len() < MAX_RAW_DIRECT_EXECUTE_THRESHOLD => {
-            let formated_bytes = serde_json::to_string(&bytes)?;
+            let bytes_as_json_array = serde_json::to_string(&bytes)?;
             webview.eval(format!(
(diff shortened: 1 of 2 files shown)


Legend-Master:
@WSH032 Hey could you give this PR a try and see if it fixes the regression?

crates/tauri/src/ipc/channel.rs
```diff
@@ -154,9 +154,9 @@ impl JavaScriptChannelId {
             ))?;
           }
           InvokeResponseBody::Raw(bytes) if bytes.len() < MAX_RAW_DIRECT_EXECUTE_THRESHOLD => {
+            let formated_bytes = serde_json::to_string(&bytes)?;
```
amrbashir: nit: why formated_bytes? isn't `serialized_bytes` a better name, or `bytes_as_json`?

amrbashir (dismissed):

crates/tauri/src/ipc/channel.rs
Legend-Master: Changed name to `bytes_as_json_array`, didn't know what to name it, and `formated_bytes` was the first name popped in my head 🙃

amrbashir (approved):

PR #13269, opened 2025-04-21
enhance(core): respect `data-tauri-drag-region="false"`

amrbashir:

### DIFF
--- a/crates/tauri/src/window/scripts/drag.js
+++ b/crates/tauri/src/window/scripts/drag.js
@@ -11,12 +11,15 @@
   // moves after the double click, it should be cancelled (see https://github.com/tauri-apps/tauri/issues/8306)
   //-----------------------//
   const TAURI_DRAG_REGION_ATTR = 'data-tauri-drag-region'
-  let x = 0,
-    y = 0
+
+  // initial mousedown position for macOS
+  let initialX = 0
+  let initialY = 0
+
   document.addEventListener('mousedown', (e) => {
     if (
-      // element has the magic data attribute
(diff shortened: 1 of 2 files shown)


crates/tauri/src/window/scripts/drag.js
```diff
@@ -11,12 +11,18 @@
...
+  let initialY = 0
+
   document.addEventListener('mousedown', (e) => {
+    const attr = e.target.getAttribute(TAURI_DRAG_REGION_ATTR)
     if (
       // element has the magic data attribute
-      e.target.hasAttribute(TAURI_DRAG_REGION_ATTR)
+      attr
+      //  and not false
+      && attr !== 'false'
```
Legend-Master: This should probably be 

```suggestion
      e.target.hasAttribute(TAURI_DRAG_REGION_ATTR)
      //  and not false
      && e.target.getAttribute(TAURI_DRAG_REGION_ATTR) !== 'false'
```

`e.target.getAttribute(TAURI_DRAG_REGION_ATTR)` will return an empty string when used like this `<div data-tauri-drag-region></div>`

Legend-Master (changes requested):

crates/tauri/src/window/scripts/drag.js
amrbashir: ah right, we can just check for null then, since that's what we care about
```suggestion
      attr !== null
      //  and not false
      && attr !== 'false'
```

Legend-Master (approved):

PR #13272, opened 2025-04-21
docs: update example for webview ts api

situ2001:
As title shown, currently, the example in `Webview` is not correct and tsc raises error:

```
Argument of type '{ url: string; }' is not assignable to parameter of type 'WebviewOptions'.
  Type '{ url: string; }' is missing the following properties from type 'WebviewOptions': x, y, width, heightts(2345)
```


amrbashir (approved):

amrbashir:
Thank you

PR #13276, opened 2025-04-22
feat: add workarea getter for monitor

amrbashir:


crates/tauri-runtime-wry/src/monitor/mod.rs
```diff
@@ -17,16 +17,11 @@ mod macos;
...
-}
-
 pub trait MonitorExt {
   /// Get the work area of this monitor
   ///
   /// ## Platform-specific:
   ///
   /// - **Android / iOS**: Unsupported.
-  fn work_area(&self) -> PhysicalRect;
+  fn work_area(&self) -> Rect;
```
Legend-Master: I think it's time for a `PhysicalRect` 🙃

amrbashir: yeah you are right, I have added them and added `LogicalRect` as well

amrbashir: will see if I can upstream them later on.

Legend-Master (approved):
Code looks good, didn't test it though

KurtGokhan:
Although it returns the size correctly, the position is always 0,0 in MacOS. Any idea about that?

PR #13277, opened 2025-04-22
fix (android): keyboard/screen resizing (fix: #7868)

### LINKED ISSUE #7868: [bug] Android Keyboard overlay/push behavior changes between app launches
### Describe the bug

Basically, as the screenshots explain, sometimes when I open my app, the keyboard is overlaying over the app, and other times it pushes the content up (shortens the window height)?

I think probably the expected behavior is the latter?

I am still seeing this regression now in `2.0.0-alpha.14`
Don't remember the version I was running when I first started seeing this, but I think it was in august maybe around `2.0.0-alpha.8`?

Screenshot_2023-08-28_at_23 00 47

### Reproduction

I think it should happen on any Android app with a text input at the bottom of the screen.

- Open app
- tap text input

Notice that the keyboard covers the input (screenshot right)

- back out of app (just exit app, don't quit)
- open app again

Keyboard should now no longer be covering the input (screenshot left)

Kinda being lazy, but I can build a reproduction repo if there isn't an easily-editable one you area already working with

### Expected behavior

Input should move to be above keyboard (at least that was the behavior prior).

### Platform and versions

```text
[✔] Environment
    - OS: Mac OS 13.5.1 X64

[-] Packages
    - tauri [RUST]: git+https://github.com/tauri-apps/tauri/?branch=dev#af3268a4be69e2dee72ba867ee54bc36d12712e6 (2.0.0-alpha.11)
    - tauri-build [RUST]: no manifest (2.0.0-alpha.8, 2.0.0-alpha.8)
    - wry [RUST]: 0.31.1
    - tao [RUST]: 0.22.2
    - @tauri-apps/cli [NPM]: 2.0.0-alpha.14

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../src-www
    - devPath: http://0.0.0.0:3000/
```

### Stack trace

_No response_

### Additional context

_No response_

retorted:
Closes https://github.com/tauri-apps/tauri/issues/7868

Sets a default behavior for resizing the Webview window when the android soft-keyboard opens or closes.

### DIFF
--- a/crates/tauri-cli/templates/mobile/android/app/src/main/AndroidManifest.xml
+++ b/crates/tauri-cli/templates/mobile/android/app/src/main/AndroidManifest.xml
@@ -13,6 +13,7 @@
         <activity
             android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale|smallestScreenSize|screenLayout|uiMode"
             android:launchMode="singleTask"
+            android:windowSoftInputMode="adjustResize"
             android:label="@string/main_activity_title"
             android:name=".MainActivity"
             android:exported="true">


lucasfernog:
I think this is something that each user needs to decide on their own and adapt their manifests. Since not every app supports such a small webview height.

lucasfernog:
For instance capacitor doesn't give you that config by default: https://github.com/ionic-team/capacitor/blob/main/android-template/app/src/main/AndroidManifest.xml

vanvuongngo:
It fixed my problem and how about to document this case at least?

Legend-Master:
I don't think this makes any differences on Android 17 here. What should be the expected behavior here, the 100vh equals the screen space - virtual keyboard height?

PR #13288, opened 2025-04-23
Use `format_callback::format_raw` for channels

oscartbeaumont:
Right now when reloading the webview a channel from the previous page might fire. This will cause some JS code similar to `window[_some_id](...)` to be injected into the page. When `window[_some_id]` evaluates it will be `undefined` as the webview context was reset when the page was reloaded. This will cause the webview to crash, sometimes taking out other JS running on the page with it.

This fix makes use of the existing `format_callback::format_raw` for parsing data back to the webview which has graceful handling on this condition.


crates/tauri/src/ipc/channel.rs
```diff
@@ -149,15 +151,14 @@ impl JavaScriptChannelId {
...
+              CallbackFn(callback_id),
+              format!("{{ message: {string}, index: {current_index} }}"),
+            )?)?;
           }
           InvokeResponseBody::Raw(bytes) if bytes.len() < MAX_RAW_DIRECT_EXECUTE_THRESHOLD => {
             let bytes_as_json_array = serde_json::to_string(&bytes)?;
-            webview.eval(format!(
-              "window['_{callback_id}']({{ message: new Uint8Array({bytes_as_json_array}).buffer, index: {current_index} }})",
-            ))?;
+            webview.eval(format_callback::format_raw(CallbackFn(callback_id), format!("{{ message: new Uint8Array({bytes_as_json_array}).buffer, index: {current_index} }}"))?)?;
```
Legend-Master: I think `format_callback::format_raw` is not supposed to be dealing with js code but just plain json, and `new Uint8Array` might not work here?

Let's add a new function and use it instead:

```rust
pub fn format_raw_js(id: u32, js: &str) -> crate::Result<String> {
  format!(
  r#"
  if (window["_{id}"]) {{
    window["_{id}"]({arg})
  }} else {{
    console.warn("[TAURI] Couldn't find callback id {id} in window. This happens when the app is reloaded while Rust is running an asynchronous operation.")
  }}"#,
  )
}
```

Also we can change `format_raw`'s callback to use it as well

crates/tauri/src/ipc/channel.rs
```diff
@@ -180,9 +181,13 @@ impl JavaScriptChannelId {
       }),
       Some(Box::new(move || {
         let current_index = counter_clone.load(Ordering::Relaxed);
-        let _ = webview_clone.eval(format!(
-          "window['_{callback_id}']({{ end: true, index: {current_index} }})",
-        ));
+        let _ = webview_clone.eval(format!("window['_{callback_id}']()",));
```
Legend-Master: I don't quite understand this change, is this some left over code?

Legend-Master (changes requested):

Legend-Master:
Also a change file like this [one](https://github.com/tauri-apps/tauri/blob/00dfc32a2db55cb7330da5d714bd9979aab1e45f/.changes/fix-webview-proxy-url.md) would be nice 🙃

crates/tauri/src/ipc/channel.rs
oscartbeaumont: This was just a mistake.

oscartbeaumont:
I have applied the fixes and added a changelog.

I have been unable to test these changes as when I run `pnpm build` I get the following:
```
> tauri-workspace@0.0.0 build /Users/oscar/Desktop/tauri
> pnpm run -r build

Scope: 6 of 7 workspace projects
examples/api build$ vite build
│ vite v6.2.6 building for production...
│ transforming...
│ ✓ 6 modules transformed.
│ ✗ Build failed in 104ms
│ error during build:
│ [vite]: Rollup failed to resolve import "@tauri-apps/api/core" from "/Users/oscar/Desktop/tauri/examples/api/src/App.svelte".
│ This is most likely unintended because it can break your application at runtime.
│ If you do want to externalize this module explicitly add it to
│ `build.rollupOptions.external`
│     at viteLog (file:///Users/oscar/Desktop/tauri/node_modules/.pnpm/vite@6.2.6_@types+node@22.13.5_jiti@2.4.2_sass@1.77.8_terser@5.39.…
│     at onRollupLog (file:///Users/oscar/Desktop/tauri/node_modules/.pnpm/vite@6.2.6_@types+node@22.13.5_jiti@2.4.2_sass@1.77.8_terser@5…
│     at onLog (file:///Users/oscar/Desktop/tauri/node_modules/.pnpm/vite@6.2.6_@types+node@22.13.5_jiti@2.4.2_sass@1.77.8_terser@5.39.0_…
│     at file:///Users/oscar/Desktop/tauri/node_modules/.pnpm/rollup@4.40.0/node_modules/rollup/dist/es/shared/node-entry.js:20704:32
│     at Object.logger [as onLog] (file:///Users/oscar/Desktop/tauri/node_modules/.pnpm/rollup@4.40.0/node_modules/rollup/dist/es/shared/…
│     at ModuleLoader.handleInvalidResolvedId (file:///Users/oscar/Desktop/tauri/node_modules/.pnpm/rollup@4.40.0/node_modules/rollup/dis…
│     at file:///Users/oscar/Desktop/tauri/node_modules/.pnpm/rollup@4.40.0/node_modules/rollup/dist/es/shared/node-entry.js:21288:26
└─ Failed in 1.1s at /Users/oscar/Desktop/tauri/examples/api
packages/api build$ rollup -c --configPlugin typescript
└─ Running...
packages/cli build$ cross-env TARGET=node napi build --platform --profile release-size-optimized
└─ Running...
/Users/oscar/Desktop/tauri/examples/api:
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  api@1.0.0 build: `vite build`
Exit status 1
 ELIFECYCLE  Command failed with exit code 1.
```

Following the `CONTRIBUTING.md` this is required to run `examples/api` which would let me to test if these changes are working correctly.

Legend-Master:
> I have been unable to test these changes as when I run pnpm build I get the following:

What if you build the API and CLI and then go to the example/api directory and run `pnpm build`? I think it might be the API build task deleted the dist folder when regenerating out files while vite is trying to build from them

crates/tauri/src/ipc/format_callback.rs
```diff
@@ -92,18 +92,23 @@ pub fn format<T: Serialize>(function_name: CallbackFn, arg: &T) -> crate::Result
...
-      console.warn("[TAURI] Couldn't find callback id {fn} in window. This happens when the app is reloaded while Rust is running an asynchronous operation.")
-    }}"#,
-      fn = function_name.0
-    )
+    format_raw_js(function_name.0, Some(arg))
   })
 }
 
+/// Formats a callback function invocation, properly accounting for error handling.
+pub fn format_raw_js(id: u32, js: Option<&str>) -> String {
```
Legend-Master: Any reason using `js: Option<&str>` instead of `js: &str` here?

crates/tauri/src/ipc/channel.rs
Legend-Master: Then why don't we remove it? `let _ = webview_clone.eval(format!("",));` doesn't make sense either

oscartbeaumont-crabnebula: Sorry, misunderstood. Fixed.

crates/tauri/src/ipc/format_callback.rs
oscartbeaumont-crabnebula: Fixed!

oscartbeaumont-crabnebula:
> What if you build the API and CLI and then go to the example/api directory 

The following works:
```
# in the `/` directory of the repo
pnpm i
cd packages/api
pnpm build
cd ../cli
pnpm build
cd ../../examples/api
pnpm tauri dev
```

Thanks for your help!

Legend-Master (approved):

Legend-Master:
Nice catch, good work 🙃

PR #13294, opened 2025-04-24
fix: Allow the use of target dir inside frontendDist/distDir(#13287)

### LINKED ISSUE #13287: [bug] Allow the use of `target` dir inside frontendDist/distDir
### Describe the bug

If you define a path in your app like http://localhost:3000/target, you'll get a folder in your ./out directory called ./out/target that includes the resources for that path.  The tauri build sees this folder and generates a build error 
"The configured distDir includes the `["target"]` folder. Please isolate your web assets on a separate folder and update `tauri.conf.json > build > distDir`."  Renaming the path to something else fixes the problem but that sucks if you really want the name of the path to be target.

### Reproduction

Create a next-js app and add a src/pages/target/index.ts 
run npm tauri build
See failure.

### Expected behavior

The builder shouldn't care about the name of the paths in the out folder

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.26100 X64
    ✔ WebView2: 135.0.3179.85
    ✔ rustc: 1.86.0 (05f9846f8 2025-03-31)
    ✔ cargo: 1.86.0 (adf9b6ad1 2025-02-28)
    - node: 22.14.0
    - pnpm: 10.9.0
    - npm: 11.1.0

[-] Packages
    - tauri [RUST]: 1.8.2
    - tauri-build [RUST]: 1.5.6
    - wry [RUST]: 0.24.11
    - tao [RUST]: 0.16.10
    - @tauri-apps/cli [NPM]: 1.6.3 (outdated, latest: 2.5.0)

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../out
    - framework: React (Next.js)
    - bundler: Webpack
```

### Stack trace

```text

```

### Additional context

_No response_

kingsword09:
fix(#13287): allow the target directory to be inside frontendDir as long as it is not the Rust target directory inside frontendDir.


FabianLars:
didn't mean to merge the dev branch sry🙃

crates/tauri-cli/src/build.rs
```diff
@@ -172,19 +172,36 @@ pub fn setup(
         ));
     }
 
+    // Issue #13287 - Allow the use of target dir inside frontendDist/distDir
+    // https://github.com/tauri-apps/tauri/issues/13287
+    let mut target_path = get_cargo_target_dir(&options.args)?;
+    target_path = tauri_path.join(target_path);
+    let target_path = fs::canonicalize(target_path)?;
```
FabianLars: This section doesn't make much sense. get_cargo_target_dir will return an absolute path and so will overwrite tauri_path. joining it to the tauri_path may not be correct for relative paths either i think.

crates/tauri-cli/src/build.rs
```diff
@@ -172,19 +172,36 @@ pub fn setup(
...
+
     let mut out_folders = Vec::new();
-    for folder in &["node_modules", "src-tauri", "target"] {
-      if web_asset_path.join(folder).is_dir() {
+    if target_path.starts_with(web_asset_path) {
```
FabianLars: web_asset_path is not canonicalized yet (the canonicalize call in line 169 does not carry over) so this check may be false for the same directories.

crates/tauri-cli/src/build.rs
```diff
@@ -172,19 +172,36 @@ pub fn setup(
...
+        if !relative_str.is_empty() {
+          out_folders.push(relative_str.to_string());
+        }
+      }
+    }
+
+    for folder in &["node_modules", "src-tauri"] {
+      let sub_path = web_asset_path.join(folder);
+      // If the subfolder exists and its canonicalized path equals the target path, add to out_folders
+      if sub_path.is_dir() && fs::canonicalize(&sub_path)? == target_path {
```
FabianLars: this check is incorrect, we check if they are the node_modules or src-tauri folders, neither are the cargo target path. In other words, we just need the is_dir check here like it was before.

FabianLars (changes requested):
thanks for the PR, just a few relatively minor things :)

crates/tauri-cli/src/build.rs
```diff
@@ -174,28 +174,26 @@ pub fn setup(
...
+    let target_path = fs::canonicalize(get_cargo_target_dir(&options.args)?)?;
     let mut out_folders = Vec::new();
-    if target_path.starts_with(web_asset_path) {
-      // Get the path of target_path relative to web_asset_path
-      if let Ok(relative_path) = target_path.strip_prefix(web_asset_path) {
-        let relative_str = relative_path.to_string_lossy();
-        if !relative_str.is_empty() {
-          out_folders.push(relative_str.to_string());
+    if let Ok(web_asset_canonical) = web_asset_path.canonicalize() {
+      if target_path.starts_with(&web_asset_canonical) {
```
FabianLars: sorry, forgot that in the first review. the outer `if target_path.starts_with(&web_asset_canonical) {` can be removed since strip_prefix already includes that check as well.

FabianLars (approved):
thank you :)

PR #13306, opened 2025-04-25
fix(core): immediately unregister event listener on unlisten call

lucasfernog:
the unlisten function is currently async, but marked as `() => void` in the TypeScript definition. To avoid a breaking change, we're going to immediately unregister the listener function so it's not called.

this fixes a race condition where after calling unlisten() you would still receive events if you do not `await` it and there's a new event triggering while the unlisten command is running


Legend-Master (dismissed):
Didn't test it, but code looks good

PR #13307, opened 2025-04-25
fix(core): fix TrayIcon.getById returning new resource IDs

lucasfernog:
this prevents the close() from working properly if you somehow lose the new() resource ID (for instance when the app reloads) and need to pick it up again and close it.


Legend-Master:
Wouldn't this be a problem when 2 webviews trying to get the same tray icon? Since they're stored in a single webview's resource table, the other webview will not be able to use `TrayIcon.getById` anymore

PR #13310, opened 2025-04-26
fix: follow-up fix for monitor work area on macOS

lucasfernog:
follow-up for #13309


lucasfernog:
(I'm still confirming if this works 100% or not)

PR #13316, opened 2025-04-26
fix: use app's resource table for storing tray icons

Legend-Master:
Reference: https://github.com/tauri-apps/tauri/pull/13307#issuecomment-2831695180

The approach is similar to what we did for the store plugin in https://github.com/tauri-apps/plugins-workspace/pull/1860


lucasfernog (approved):
this means now we can call TrayIcon.getById for a tray that was created via config / Rust right?

Legend-Master:
> this means now we can call TrayIcon.getById for a tray that was created via config / Rust right?

We always could, this command I think was created exactly for that

The story was if I remembered correctly, I asked about this in an issue that it would be nice to be able to do that and amr then implemented it

merlinaudio:
@Legend-Master 

As far as I know, the command was also created so you don't accidentally create multiple trays with HMR for example

Otherwise, if you just did this in Vite:

```ts
function createTray() { return TrayIcon.new({ id: "foo" }) }

// main.ts
createTray();
```

Then you'd get many trays. So

```ts
function createTray() {
  const existingTray = await TrayIcon.getById("foo");
  return existingTray ?? TrayIcon.new(...)
}
```

Is the only way to make it work "properly" from JS for many dev scenarios

--- 

Thank you for working on this fix!!

PR #13322, opened 2025-04-27
refactor: use `WindowConfig` for `create_webview`

Legend-Master:
I think some of the properties inside `WebviewConfig` were missing and we don't really maintain them as much, let's just use the `WindowConfig` like we did for `window:create`

Maybe we could split them up in `tauri-utils` like this in the future

```rust
#[derive(Debug, PartialEq, Clone, Deserialize, Serialize)]
#[serde(flatten)]
pub struct WebviewWindowConfig {
  pub window_config: WindowConfig,
  pub webview_config: WebviewConfig,
}
```


lucasfernog:
unfortunately the proper change would need a breaking change :( i wanted it!!!

lucasfernog (approved):

PR #13325, opened 2025-04-27
refactor: rework `transformCallback`

Legend-Master:
This PR refactors `transformCallback` and the functions calls to it into a few js functions, and instead of binding them on `window`, we put them inside `window.__TAURI_INTERNALS__.callbacks` instead

Also introduced

- `window.__TAURI_INTERNALS__.runCallback` to replace the old `window['_{cb}']` which is quite hard to understand what it's doing
- `window.__TAURI_INTERNALS__.unregisterCallback` to replace `delete window['_{cb}']` 

Honestly, `window['_{cb}']` is super hard to understand and we also need to use `format_callback` to send a warning message to test if the callback is still there in case of webview reloads


Legend-Master:
Should we also expose `unregisterCallback` and `runCallback`?

lucasfernog (approved):
you have so many great ideas :heart:

Issue #13302, opened 2025-04-25
[feat] Support set fileAssociations icon


wyhaya:
### macOS

The ⁠Info.plist file allows adding the `CFBundleTypeIconFile` property to specify the icon for associated files. It would be ideal if this could be configured in the `⁠tauri.conf.json` file.

```xml
<dict>
  <key>CFBundleTypeExtensions</key>
  <array>
    <string>db</string>
  </array>
  <key>CFBundleTypeIconFile</key>
  <string>my-icon.icns</string>
  <key>CFBundleTypeName</key>
  <string>...</string>
</dict>
```

### Describe the solution you'd like

```diff json
 {
     "fileAssociations": [
         {
             "ext": ["db"],
+            "icon": "./my-icon.icns",
             "name": "SQLite file"
         }
     ]
 }
```

PR #13345, opened 2025-04-30
feat: add LSItemContentTypes and CFBundleTypeIconFile support for fileAssociations on macOS

kingsword09:
According to [apple docs](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-101685):
1. `LSItemContentTypes`: `CFBundleTypeExtensions` has been deprecated since `OS X v10.5`, so `LSItemContentTypes` has been added. In macOS 10.5 and later, this key (when present) takes precedence over these type-identifier keys: `CFBundleTypeExtensions`, `CFBundleTypeMIMETypes`, `CFBundleTypeOSTypes`.
2. `CFBundleTypeIconFile`: this key contains a string with the name of the icon file (.icns) to associate with this macOS document type. #13302
3. To avoid breaking changes, the attribute is set as optional.

### DIFF
--- a/.changes/feat-add-utis-support-for-fileassociation.md
+++ b/.changes/feat-add-utis-support-for-fileassociation.md
@@ -1,0 +1,8 @@
+---
+"tauri-bundler": "minor:feat"
+"tauri-utils": "minor:feat"
+"tauri-cli": "minor:feat"
+"@tauri-apps/cli": "minor:feat"
+---
+
+Feat: add `LSItemContentTypes` and `CFBundleTypeIconFile` support for `fileAssociations` on macOS.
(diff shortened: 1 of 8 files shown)


FabianLars:
Thanks for the PR! I think we need to think a tiny bit more about the icon support. We _may_ wanna think about letting users specify the file path to the icns file on the dev system and copy it ourselves to where it's needed (iirc that's what we do for other icons/images) or alternatively check if bundling the icns file as a resource works and document that.
We probably should keep in mind that iOS' `CFBundleTypeIconFiles` works differently (multiple pngs instead of one icns) for that decision even if we don't support file assocs on ios yet.

PR #13368, opened 2025-05-03
docs: Added more detailed comments to menu `popup` method and DPI-related classes

geeseofbeverlyroad:
The documentation for `popup` was missing a description of what an undefined value for `at` means.

The documentation for `Physical/LogicalSize` and `-Position` were missing what "physical" and "logical" pixels mean, as their names are not obvious it feels good to explain they are dealing with DPI.


lucasfernog (approved):
thank you!

PR #13374, opened 2025-05-04
docs: Fix description and add example for `WebviewWindowBuilder::from_config`

miguelpduarte:
The documentation for `WebviewWindowBuilder::from_config` mentions changing the label of the new `WebviewWindowBuilder`, which is not possible (there is no such API).

Instead, the label must be changed in the `WindowConfig` that is passed into `WebviewWindowBuilder::from_config`.

This change fixes that description and adds an example code snippet for this use-case.

Additionally, it updates the existing example, as the `WindowConfig` that was being retrieved from `app.config()` was being cloned and then used behind a reference. As such, we can remove the unnecessary clone to make the example cleaner.


lucasfernog (approved):
nice catch, thanks!

PR #13392, opened 2025-05-07
fix(cli): don't canonicalize cargo's target dir in frontendDist verification

FabianLars:
currently you'll get os error 2 if there's no target dir yet. cargo metadata already returns an absolute path so no need to canonicalize it. if we use dunce we can still canonicalize the frontendDist path and get a path that should match the cargo metadata output.


FabianLars:
no changefile cause it's an unreleased change from https://github.com/tauri-apps/tauri/pull/13294

amrbashir (approved):

PR #13399, opened 2025-05-08
fix(windows): Set drag and drop to false in window builder in `disable_drag_and_drop_handler()`

guillemcordoba:
Without this, `disable_drag_and_drop_handler()` does not change the bevahior of the tao windows.

For me, this ended up triggering [this panic](https://github.com/tauri-apps/tao/blob/67f00588b4927271cb8b89f1d0c10db3c9727093/src/platform_impl/windows/window.rs#L109). This PR fixes this and I don't have the panic anymore.


Legend-Master (approved):
I have been using this method for quite some now without hitting the panic, but anyways, this change makes sense to me, just not sure if we should do this on all platforms or just for Windows (doesn't really matter to be honest since other platforms don't have this restriction anyways)

@amrbashir sorry for the ping but I'm not entirely sure if this is fine or not, pinging you since you wrote most of the drag drop stuffs

PR #13406, opened 2025-05-09
feat: check if webview runtime is accessible when creating a webview

amrbashir:


crates/tauri-runtime-wry/src/dialog/windows.rs
```diff
@@ -0,0 +1,127 @@
...
+}
+
+pub fn error<S: AsRef<str>>(err: S) {
+  dialog_inner(err, Level::Error);
+}
+
+fn dialog_inner<S: AsRef<str>>(err: S, level: Level) {
+  let err = err.as_ref();
+
+  let title = match level {
```
Legend-Master:
```suggestion
pub fn error<S: AsRef<str>>(err: S) {
  dialog_inner(err.as_ref(), Level::Error);
}

fn dialog_inner(err: &str, level: Level) {
  let title = match level {
```

crates/tauri-runtime-wry/src/dialog/windows.rs
```diff
@@ -0,0 +1,127 @@
...
+    Level::Info => w!("Info"),
+  };
+
+  #[cfg(not(feature = "common-controls-v6"))]
+  {
+    use windows::Win32::UI::WindowsAndMessaging::*;
+
+    let err = remove_hyperlink(err);
+    let err = HSTRING::from(err);
+    let err = PCWSTR(err.as_ptr());
```
Legend-Master:
```suggestion
```

crates/tauri-runtime-wry/src/dialog/windows.rs
```diff
@@ -0,0 +1,127 @@
...
+      if msg == TDN_HYPERLINK_CLICKED {
+        let link = PCWSTR(lparam.0 as _);
+        let _ = unsafe { ShellExecuteW(None, None, link, None, None, SW_SHOWNORMAL) };
+      }
+
+      S_OK
+    }
+
+    let err = HSTRING::from(err);
+    let err = PCWSTR(err.as_ptr());
```
Legend-Master, line 71:
```suggestion
```

Legend-Master (dismissed):
Code looks good, just a few nit picks, didn't test it though

crates/tauri-runtime-wry/src/dialog/windows.rs
amrbashir: but this one can't because it is used in a struct with an owned value not a generic `Param<T>`

crates/tauri-runtime-wry/src/dialog/windows.rs
amrbashir: this one can be removed

crates/tauri-runtime-wry/src/dialog/windows.rs
Legend-Master: Right right, I'm so paranoid nowadays when I see `PCWSTR` 😂

Legend-Master (approved):

Issue #13382, opened 2025-05-06
[bug] Tauri 2.0 compiles much slower than version 1.0. How can I configure it to improve compilation speed?


zjfpcb:
### Describe the bug

[profile.dev]
incremental = true # 以较小的步骤编译您的二进制文件。

[profile.release]
codegen-units = 1 # 允许 LLVM 执行更好的优化。
lto = true # 启用链接时优化。
opt-level = "s" # 优先考虑小的二进制文件大小。如果您更喜欢速度，请使用 `3`。
panic = "abort" # 通过禁用 panic 处理程序来提高性能。
strip = true # 确保移除调试符号。
The compilation speed remains slow regardless of whether the above configurations are applied or not. Tauri 2.0 takes 1 minute and 10 seconds to compile, while version 1.0 completes in under 20 seconds. How can I improve the compilation time for version 2.0?

### Full `tauri info` output

```text
Tauri 2.0 takes 1 minute and 10 seconds to compile, while version 1.0 completes in under 20 seconds.
```

### Stack trace

```text

```

FabianLars:
Please include the output of the `tauri info` command as required by the issue template.

Are we talking about `tauri build`, `tauri build --debug` or `tauri dev` ? Do your v1 and v2 apps have similar features and dependencies?

zjfpcb:
> Please include the output of the `tauri info` command as required by the issue template.
> 
> Are we talking about `tauri build`, `tauri build --debug` or `tauri dev` ? Do your v1 and v2 apps have similar features and dependencies?

Yes, v1 and v2 are the same code, but the Rust versions are different. v1 is 1.76, while v2 is 1.77.2.

Legend-Master:
Not entirely sure how I much I could do, but I'll try to dig into this a little bit, the long compile time has been bothering me as well for quite some time now

samheuck:
I'm on MacOS (w/ a super fast Mac Studio M4) and using VS-Code w/ rust-analyzer. I just switched an app I'm working on to V2. I also experienced an order of magnitude slower compilation time when running `tauri dev`.

After some observation, I noticed issues with rust-analyzer blocking. I also noticed that certain dependencies were getting recompiled nu-necessarily.

These two VSCode settings:

```
"rust-analyzer.cargo.targetDir": true,
"rust-analyzer.cargo.extraEnv":  {
    "MACOSX_DEPLOYMENT_TARGET": "10.13"
}
```

and this in `tauri.conf.json`

```
[...]

PR #13410, opened 2025-05-10
chore: feature gate html manipulation code

Legend-Master:
Some initial effort on reducing the compile time, reference: #13382

This will be a breaking change in the `tauri-utils` crate, and we didn't mark this module as unstable before, so I'm not entirely sure if we can do this or should we wait for v3

### DIFF
--- a/.changes/feature-gate-html-manipulation.md
+++ b/.changes/feature-gate-html-manipulation.md
@@ -1,0 +1,6 @@
+---
+"tauri": 'patch:changes'
+"tauri-utils": 'patch:breaking'
+---
+
+Feature gated the HTML manipulation code in `tauri-utils` behined a flag to reduce compile time
--- a/crates/tauri-cli/Cargo.toml
+++ b/crates/tauri-cli/Cargo.toml
@@ -63,6 +63,7 @@ tauri-utils = { version = "2.4.0", path = "../tauri-utils", features = [
   "schema",
   "config-json5",
   "config-toml",
+  "html-manipulation",
 ] }
 tauri-utils-v1 = { version = "1", package = "tauri-utils", features = [
   "isolation",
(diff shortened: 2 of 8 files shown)


FabianLars:
we release breaking changes in tauri-utils all the time much to the dismay of users (and me cause i have to tell them to run cargo update lol) so why stop now

Legend-Master:
Feel like we probably should just mark the entire crate unstable 🙃

FabianLars:
well the problem is really how we consume it in tauri. if people update tauri they often do that in a way that doesn't update tauri-utils. We should really think about something like locking minor versions internally imho.

whether we mark it unstable or not, i don't care. most of the prior breaking changes could have been non-breaking as well if we'd change the api design a bit and/or use non_exhaustive more often.

FabianLars (approved):

Legend-Master:
> We should really think about something like locking minor versions internally imho.

Yeah, I'm also thinking about it, this is especially problematic for the plugins, we simplifies them to just 2 while we actually use some of features that requires a higher version

> whether we mark it unstable or not, i don't care. most of the prior breaking changes could have been non-breaking as well if we'd change the api design a bit and/or use non_exhaustive more often.

To be fair, this crate is mostly for sharing code between the build script and the runtime crates, and probably can't be stable anyways

FabianLars:
> Yeah, I'm also thinking about it, this is especially problematic for the plugins, we simplifies them to just 2 while we actually use some of features that requires a higher version

I mean, the plugins are an easy fix. We can just set the minimum tauri versions they need. i thought that's what we were doing but just didn't had a reason already x)

> To be fair, this crate is mostly for sharing code between the build script and the runtime crates, and probably can't be stable anyways

Maybe we shouldn't re-export it as-is from `tauri` then, but that's something for the next majors to worry about 🤷

PR #13418, opened 2025-05-12
refactor: put dynamic acl to a feature

Legend-Master:
Put dynamic ACL into a feature `dynamic-acl`, this is currently enabled by default to align with the previous behaviors, you can disable it through `default-features = false` to reduce the final binary size by not including the ACL references

This helps eliminate the need of keeping a reference of all the ACLs in the release build (this saves ~100KB without any extra plugins from testing), I have tried to just remove all the references to the `RuntimeAuthority::acl` field but that's not enough for the rust's dead code elimination to remove it

Not entirely sure how often this is used, if it's not commonly used, maybe we could disable it by default in v3

Reference: #12820


FabianLars:
> Not entirely sure how often this is used, if it's not commonly used, maybe we could disable it by default in v3

As far as i can tell it's a rare sight

crates/tauri/src/ipc/authority.rs
```diff
@@ -264,9 +143,15 @@ impl RuntimeAuthority {
...
 
   /// Adds the given capability to the runtime authority.
-  pub fn add_capability(&mut self, capability: impl RuntimeCapability) -> crate::Result<()> {
+  #[cfg(feature = "dynamic-acl")]
+  pub fn add_capability(&mut self, capability: impl super::RuntimeCapability) -> crate::Result<()> {
+    self.add_capability_inner(capability.build())
+  }
+
+  #[cfg(feature = "dynamic-acl")]
+  fn add_capability_inner(&mut self, capability: CapabilityFile) -> crate::Result<()> {
```
FabianLars, line 152: Why is this split into 2 functions when _inner isn't used anywhere else?

Legend-Master: This is honestly a pretty weird thing that the compiler can't optimize yet, so basically, if a function is generic, the compiler generates a new function for every generic parameter, and this will be a problem for bigger functions, and the idea here is to reduce the generic part to as small as possible, and create another function that's non-generic, a blog that talks about it more here https://kbknapp.dev/generically-bloated/

FabianLars: Ahh irght, that makes sense. Didn't think about that this would be worse for large functions but in handsight it's obvious x)

FabianLars (approved):

PR #13426, opened 2025-05-13
fix(cli): slow dev startup

Legend-Master:
Noticed the CLI commands were quite slow to start, and they seem to use quite a lot of CPU as well, and it turned out it's polling on the file watcher results on `Cargo.toml`, what a scam 🤦‍♂️

### DIFF
--- a/crates/tauri-cli/src/interface/rust.rs
+++ b/crates/tauri-cli/src/interface/rust.rs
@@ -134,7 +134,7 @@ impl Interface for Rust {
       watcher.watch(tauri_dir().join("Cargo.toml"), RecursiveMode::NonRecursive)?;
       let (manifest, modified) = rewrite_manifest(config)?;
       if modified {
-        // Wait for the modified event so we don't trigger a re-build laster on
+        // Wait for the modified event so we don't trigger a re-build later on
         let _ = rx.recv_timeout(Duration::from_secs(2));
       }
       manifest
--- a/crates/tauri-cli/src/interface/rust.rs
+++ b/crates/tauri-cli/src/interface/rust.rs
@@ -11,7 +11,7 @@ use std::{
   process::Command,
   str::FromStr,
   sync::{mpsc::sync_channel, Arc, Mutex},
-  time::{Duration, Instant},
+  time::Duration,
 };
 
 use anyhow::Context;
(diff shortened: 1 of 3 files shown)


crates/tauri-cli/src/interface/rust.rs
```diff
@@ -131,16 +131,18 @@ impl Interface for Rust {
...
+        let timeout = Duration::from_secs(2);
+        loop {
+          if now.elapsed() >= timeout {
+            break;
+          }
+          if rx.try_recv().is_ok() {
+            break;
+          }
         }
       }
```
Legend-Master: This is the changed part, other parts are just some small refactors that I did when looking through them

Also, I'm not entirely sure what does this watch even do, I would assume it's for waiting for the file modification to take effect from `rewrite_manifest`, is it?

FabianLars:
loooool that's the actual "add sleep 2 secs so we can improve perf later" meme 🤣🤣🤣

crates/tauri-cli/src/interface/rust.rs
FabianLars: looks like it yeah. just so that it doesn't trigger a rebuild ig

Legend-Master: Added some comments and changed the loop to `recv_timeout`

crates/tauri-cli/src/interface/rust.rs
```diff
@@ -126,22 +126,16 @@ impl Interface for Rust {
...
-        if now.elapsed() >= timeout {
-          break;
-        }
-        if rx.try_recv().is_ok() {
-          break;
-        }
+      watcher.watch(tauri_dir().join("Cargo.toml"), RecursiveMode::NonRecursive)?;
+      let (manifest, modified) = rewrite_manifest(config)?;
+      if modified {
+        // Wait for the modified event so we don't trigger a re-build laster on
```
FabianLars:
```suggestion
        // Wait for the modified event so we don't trigger a re-build later on
```

crates/tauri-cli/src/interface/rust.rs
FabianLars: Should we maybe reduce the timeout then? I find it hard to believe that it could take up to 2 seconds 🤔 Or maybe, better safe than sorry idk

Legend-Master: Whatever I guess, the timeout is just there to make sure we move on, if there's indeed an modification and no errors, the watcher should tell us before the timeout, I'm not sure what time to set for the denouncer though, 1 sec feels too long to me, but I don't really care since it's rare, better safe then sorry I guess 🙃

FabianLars: Your call 🤷

FabianLars (approved):

PR #13476, opened 2025-05-20
fix(core): capability filtering crashing allowed command generation

lucasfernog:
tauri-plugin and tauri build scripts cannot have access to the capabilities file (generated by tauri-build) and can only infer capabilities from the config path (from the env var)

### DIFF
--- a/crates/tauri-utils/src/acl/build.rs
+++ b/crates/tauri-utils/src/acl/build.rs
@@ -11,7 +11,7 @@ use std::{
 };
 
 use crate::{
-  acl::{has_app_manifest, AllowedCommands, Error},
+  acl::{has_app_manifest, AllowedCommands, Error, CAPABILITIES_FILE_NAME},
   config::Config,
   write_if_changed,
 };
@@ -419,11 +419,6 @@ pub fn generate_allowed_commands(
     println!("cargo:rerun-if-changed={}", capabilities_path.display());
   }
 
-  let mut capabilities = crate::acl::build::parse_capabilities(&format!(
-    "{}/**/*",
-    glob::Pattern::escape(&capabilities_path.to_string_lossy())
-  ))?;
-
   let target_triple = env::var("TARGET")?;
   let target = crate::platform::Target::from_triple(&target_triple);
   let (mut config, config_paths) = crate::config::parse::read_from(target, &config_directory)?;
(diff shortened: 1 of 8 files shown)


crates/tauri-utils/src/acl/build.rs
```diff
@@ -460,9 +455,23 @@ pub fn generate_allowed_commands(
     })
     .collect();
 
-  capabilities.extend(crate::acl::get_capabilities(&config, None, None)?);
+  let capabilities_file_path = out_dir.join(CAPABILITIES_FILE_NAME);
+  // the capabilities file only exist in the tauri-build context
+  if !capabilities_file_path.exists() {
```
Legend-Master: Would this call be a problem when the build script re-runs? I think maybe we could move back the 

```rust
  if let Some(capabilities_file_path) = pre_built_capabilities_file_path {
    if capabilities_file_path.exists() {
      let capabilities_file =
        std::fs::read_to_string(capabilities_file_path).context("failed to read capabilities")?;
      capabilities_from_files =
        serde_json::from_str(&capabilities_file).context("failed to parse capabilities")?;
    }
  }
```

part to `tauri-codegen`, and replace `pre_built_capabilities_file_path` param with `capabilities_from_files: BTreeMap<String, Capability>`, and we don't have to save the files here (this file should only be used for passing the data from `tauri-build` to `tauri-codegen` right? (maybe we should document this, I'm having a hard time revisiting this code 😂))

lucasfernog: same, it was a nightmare 🤣

crates/tauri-utils/src/acl/build.rs
```diff
@@ -460,9 +456,18 @@ pub fn generate_allowed_commands(
...
+  } else {
+    crate::acl::build::parse_capabilities(&format!(
+      "{}/**/*",
+      glob::Pattern::escape(&capabilities_path.to_string_lossy())
+    ))?
+  };
+  let capabilities = crate::acl::get_capabilities(&config, capabilities_from_files, None)?;
 
   let permission_entries = capabilities
+    .clone()
```
Legend-Master: No need to clone here?

```suggestion
```

Legend-Master (dismissed):

crates/tauri-utils/src/acl/build.rs
lucasfernog: oops testing leftover

Legend-Master (approved):

PR #13483, opened 2025-05-21
fix(cli): use platform name to detect iOS architecture to build

lucasfernog:
somehow some user's xcode is sending `--platform iOS Simulator ... arm64` though mine sends `--platform iOS ... Simulator arm64` so the arch detection is broken, i've tweaked it a bit here so it checks the platform argument aswell


lucasfernog:
fix confirmed by the user

FabianLars (approved):

PR #13512, opened 2025-05-26
fix(schema): update description for hardenedRuntime

Tunglies:
Relatived issues: https://github.com/tauri-apps/tauri-docs/issues/3303

The issue reports a broken or missing link for "hardened runtime" on the Tauri documentation page at https://v2.tauri.app/reference/config/#hardenedruntime. The documentation attempts to link to Apple’s official documentation for "hardened runtime," but the generated URL is incorrect and leads to a "Page not found" error.

### DIFF
--- a/crates/tauri-cli/config.schema.json
+++ b/crates/tauri-cli/config.schema.json
@@ -3436,7 +3436,7 @@
           ]
         },
         "hardenedRuntime": {
-          "description": "Whether the codesign should enable [hardened runtime] (for executables) or not.\n\n [hardened runtime]: <https://developer.apple.com/documentation/security/hardened_runtime>",
+          "description": "Whether the codesign should enable [hardened runtime](https://developer.apple.com/documentation/security/hardened_runtime) (for executables) or not.",
           "default": true,
           "type": "boolean"
         },
--- a/crates/tauri-schema-generator/schemas/config.schema.json
+++ b/crates/tauri-schema-generator/schemas/config.schema.json
@@ -3436,7 +3436,7 @@
           ]
         },
         "hardenedRuntime": {
-          "description": "Whether the codesign should enable [hardened runtime] (for executables) or not.\n\n [hardened runtime]: <https://developer.apple.com/documentation/security/hardened_runtime>",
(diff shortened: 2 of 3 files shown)


FabianLars:
Thanks for the pr!

The schema is auto-generated from this file https://github.com/tauri-apps/tauri/blob/dev/crates/tauri-utils/src/config.rs#L639-L641 so your PR changes would get lost the next time the project is compiled.

Also, if written like this https://github.com/tauri-apps/tauri/blob/dev/crates/tauri-utils/src/config.rs#L962 it should work okay, see https://v2.tauri.app/reference/config/#webviewinstallmode (if we want to keep the format instead of `See more`

Tunglies:
> Thanks for the pr!
> 
> The schema is auto-generated from this file https://github.com/tauri-apps/tauri/blob/dev/crates/tauri-utils/src/config.rs#L639-L641 so your PR changes would get lost the next time the project is compiled.
> 
> Also, if written like this https://github.com/tauri-apps/tauri/blob/dev/crates/tauri-utils/src/config.rs#L962 it should work okay, see https://v2.tauri.app/reference/config/#webviewinstallmode (if we want to keep the format instead of `See more`

I've made some changes, this should now follow the format and properly include the links in the comments.

Tunglies:
It seems the previous CI error was caused by submitting only the source code changes without updating the schema.json file. I've now added the updated schema.json file generated by running `cargo run -p tauri-schema-generator.` The CI should pass now.
Could we try rerunning the CI checks? @FabianLars

FabianLars (approved):

FabianLars:
thank you. do you think you can [sign your commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) as well? I need all of them signed to be able to merge a PR here even for doc changes :/

Tunglies:
> thank you. do you think you can [sign your commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) as well? I need all of them signed to be able to merge a PR here even for doc changes :/

I think I've sign with verified commit now. Let's merge :D

FabianLars:
yep looks good, thanks.

PR #13521, opened 2025-05-28
feat: add option to not wait on notarization to finish

FabianLars:
The first time you try to notarize an app can take multiple hours (apparently even days), very easily hitting the timeout of github actions (and consuming insane amounts of build minutes in private repos).

This PR offers a simple flag to opt out of the waiting mechanic with the drawback of not being able to staple the notarization ticket onto the app automatically.

In the future we probably should look into a timeout mechanic instead but this is more of a hotfix.

Note that i did not test this.


FabianLars:
With `Note that i did not test this.` i meant actual signing/notarizing not building the CLI 😭 ~~what did i doooo~~ ah it's the tests

lucasfernog:
since this is supposed to be a temporary measure i don't think it should be done in the config file; maybe a CLI option would be better? i feel like most people wouldn't use it anyway until they hit this particular problem, which is too late anyway...

FabianLars:
i wanted to add it as a cli flag at first but it also felt kinda weird idk (actually just couldn't think of an arg name...). will change it tomorrow-ish

also, i think we should even add this to our signing guide as a recommendation for the first notarization.

FabianLars:
still not tested (no notarization setup) and the changes also felt weird. was a bit more work than expected for a simple cli arg

lucasfernog (approved):

PR #13529, opened 2025-05-29
refactor: simplify `future` in command IPC structs

Legend-Master:
Reference: #13382

This reduced the incremental compilation time from 4.36s to 3.59s on my machine quite consistently (tested by adding and removing a new line inside an async command)

I don't fully understand the differences between an async function vs `impl Future` in rust though, but I think they're the same in this case

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8350,7 +8350,6 @@ dependencies = [
  "dirs 6.0.0",
  "dunce",
  "embed_plist",
- "futures-util",
  "getrandom 0.2.15",
  "glob",
  "gtk",
--- a/crates/tauri/Cargo.toml
+++ b/crates/tauri/Cargo.toml
@@ -52,7 +52,6 @@ tokio = { version = "1", features = [
   "fs",
   "io-util",
 ] }
-futures-util = "0.3"
 uuid = { version = "1", features = ["v4"], optional = true }
 url = "2"
 anyhow = "1"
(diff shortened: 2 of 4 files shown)


FabianLars:
> but I think they're the same in this case

I think so too

FabianLars:
from ~3.2 to ~2.9 on macos (battery saving mode, forgot to say that in the last pr) - only a single command in my test project tho

FabianLars (approved):

FabianLars:
you may want to wait for @lucasfernog to chime in cause i have no clue why there was a impl Future instead of an async fn in the first place.

FabianLars:
Or maybe not 😂

Legend-Master:
Let's just merge it and see anything breaks 🙃

I guess it was probably for supporting prior rust editions without `async` keyword (< 2018) but since that was quite a while back, I'm not entirely sure

chippers:
> Let's just merge it and see anything breaks 🙃
> 
> I guess it was probably for supporting prior rust editions without `async` keyword (< 2018) but since that was quite a while back, I'm not entirely sure

Was written back when the async ecosystem was still settling down even though future got stabilized, so at the time it was still pretty common to write async with iterator like code from the futures crate.

This adds an implicit lifetime to the functions but doesn't matter in this case (I think).

The new variable binding is after the await so I don't think there will be problems there but could be easy to introduce one in the future

PR #13558, opened 2025-06-03
fix(windows): allow web fullscreen APIs to work

### LINKED ISSUE #13557: [bug] Video player native fullscreen (element fullscreen) not working, only web fullscreen available (except on macOS with macOSPrivateApi)
### Describe the bug

When I try to make a video player go fullscreen (like, proper OS-level fullscreen where only the video shows), it just ends up filling the app window (web fullscreen). This happens on Windows.
I've tried a few different players (DPlayer, Artplayer, Video.js), and they all do the same thing. So, it feels like a Tauri thing.
The weird part is, on a Mac, if I set "macOSPrivateApi": true in the tauri.conf.json, then native fullscreen works perfectly! But that flag doesn't help on Windows, and I still get the web fullscreen behavior. I've double-checked my capabilities and permissions, and they seem fine.
I've also searched GitHub for other Tauri projects using video players. Unfortunately, there aren't many examples, but the few I found seem to exhibit the same issue on Windows – they can't achieve true OS-level fullscreen either.

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.18363 x86_64 (X64)
    ✔ WebView2: 136.0.3240.92
    ✔ rustc: 1.86.0 (05f9846f8 2025-03-31)
    ✔ cargo: 1.86.0 (adf9b6ad1 2025-02-28)
    - node: 20.19.2
    - npm: 10.8.2

[-] Packages
    - tauri �🦀: 2.5.1
    - tauri-build �🦀: 2.2.0
    - wry �🦀: 0.51.2
    - tao �🦀: 0.33.0
    - @tauri-apps/api : 2.5.0
    - @tauri-apps/cli : 2.5.0

[-] Plugins
    - tauri-plugin-opener �🦀: 2.2.7
    - @tauri-apps/plugin-opener : 2.2.7

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - framework: Vue.js
    - bundler: Vite
```

### Stack trace

```text

```

### Additional context

_No response_

Legend-Master:
Reference: https://github.com/MicrosoftEdge/WebView2Feedback/issues/1834

Fix #13557

I believe it works in this way on both Mac and Linux, so let's align this behavior on Windows as well

### DIFF
--- a/.changes/windows-fullscreen.md
+++ b/.changes/windows-fullscreen.md
@@ -1,5 +1,5 @@
 ---
-tauri-wry: "patch:bug"
+tauri-runtime-wry: "patch:bug"
 ---
 
 Allow web fullscreen APIs to work (e.g. `video.requestFullscreen` and `document.exitFullscreen`)
--- a/.changes/windows-fullscreen.md
+++ b/.changes/windows-fullscreen.md
@@ -1,0 +1,5 @@
+---
+tauri-wry: "patch:bug"
+---
+
+Allow web fullscreen APIs to work (e.g. `video.requestFullscreen` and `document.exitFullscreen`)
--- a/.changes/windows-fullscreen.md
+++ b/.changes/windows-fullscreen.md
@@ -2,4 +2,4 @@
 tauri-runtime-wry: "patch:bug"
 ---
 
-Allow web fullscreen APIs to work (e.g. `video.requestFullscreen` and `document.exitFullscreen`)
+Allow web fullscreen APIs to work on Windows (e.g. `video.requestFullscreen` and `document.exitFullscreen`)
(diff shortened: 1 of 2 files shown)


FabianLars:
I'll check macos in an hour or so. I don't think it behaves like this but didn't check for ages

Legend-Master:
Since that's what the author said in #13557, I assumed it is the case, for Linux side, it seems to work like that in WSL at least

FabianLars:
i was so sure it worked like this on more platforms than just windows but i guess i didn't look closely enough when people talked about it in #support . And yes, can confirm that it's a proper fullscreen on macos. Which tbh is kinda sad, i'd imagine there are use cases for the current windows behavior.

FabianLars (approved):
No access to Windows today, so untested approval.

Legend-Master:
I have tested a few times and it seems to work fine 🙃

PR #13564, opened 2025-06-04
feat(core): webview window focusable property, closes #11130

### LINKED ISSUE #11130: [feat] Webview Window Focusable Property
### Describe the problem

There isn't the ability to create a window that is not focusable, like a onscreen keyboard. Electron has a BaseWindow property that can be set of focusable which then causes the window never to gain focus.

This would be a property that is set on the WebviewWindow which will define if the window can take focus or not. The window should still be interactable with the user.

### Alternatives considered

I have attempted to create the window and change the window style with WinApi but have not been successful in changing the styles. As I don't always know where the focus has come from I cannot always just give it back with a event when the focus is gained in the window.

### Additional context

_No response_

robertrpf:
Adds the option to create non focusable windows.

Now you can use focusable property to tauri.conf.json.

Default behaviour is true.

Using focusable as false will create a non focusable window.

 "app": {
    "windows": [
      {
        "title": "demoapp",
        "width": 800,
        "height": 600,
        "focusable": false
      }
    ],
  },

The feature was requested in https://github.com/tauri-apps/tauri/issues/11130


lucasfernog (approved):
nice one!

PR #13596, opened 2025-06-09
fix: failing scope test on Windows

Legend-Master:
Remove the scope test that depends on the current cwd and that drive letter

I think what they're testing are basically the same as the above tests that we don't really need them anymore, but I don't mind the suggestion from @FabianLars to guard this behined a check (if not run from `C:`)


FabianLars (approved):
okay whatever, the failing checks get on my nerves

PR #13609, opened 2025-06-12
feat: allow all frame init scripts in plugin

### LINKED ISSUE #13577: [bug] `tauri::Plugin`'s `js_init_script` does not run in iframes
### Describe the bug

In the documentation, it says `it runs on all top-level document and child frame page navigations` but I think I misunderstood the following because when adding a way to tweak `window` and `navigator` properties to my internal plugin, it seemed like it only applied to the `top` window and none of the iframes' window.

## SETUP 
1. Navigate to a page that has a `data:` URL.
2. Scripts in this page creates an iframe to an external `https:` domain which contains an HTML document that is loaded.

## PLUGIN

As simple as...

```rust
pub const INIT_SCRIPT: &str = r#"
  console.log(-1, "Hello, from", window.location.origin);

  const SCREEN_WIDTH = 393;
  const SCREEN_HEIGHT = 852;

  const tweak_screen = (prop, val) => window.screen.__defineGetter__(prop, () => val);
  tweak_screen("height", SCREEN_HEIGHT);
  tweak_screen("width", SCREEN_WIDTH);
  tweak_screen("availHeight", SCREEN_HEIGHT);
  tweak_screen("availWidth", SCREEN_WIDTH);
  tweak_screen("pixelDepth", 24);
  tweak_screen("colorDepth", 24);

  window.__defineGetter__("innerHeight", () => SCREEN_HEIGHT);
  window.__defineGetter__("innerWidth", () => SCREEN_WIDTH);
  window.screen.orientation.__defineGetter__("type", () => "portrait-primary");
  navigator.__defineGetter__("platform", () => "iPhone");
"#;

// ... in the plugin's `init` function ...

Builder::new("internal-api")
  .js_init_script(INIT_SCRIPT.to_string())
  .build()
```

## DEMO

We can simply observe this by looking at the console and see that only a `null` domain is logged while another log is expected from iframe.

Also, we can check if properties are modified in the iframe.

Image

As we can see, they're only modified in the `top` window but not in the iframe's window.

### Expected behavior

Initialization script should also run in the iframe, as mentioned in the documentation.

### Full `tauri info` output

```text
[✔] Environment
    - node: 24.1.0
    - pnpm: 10.11.0
    - yarn: 1.22.22
    - npm: 11.4.1
    - bun: 1.2.15

[-] Packages
    - tauri 🦀: 2.5.1
    - tauri-build 🦀: 2.2.0
    - wry 🦀: 0.51.2
    - tao 🦀: 0.33.0
    - @tauri-apps/api : 2.5.0
    - @tauri-apps/cli : 2.5.0

[-] Plugins
    - tauri-plugin-opener 🦀: 2.2.6, (outdated, latest: 2.2.7)
    - @tauri-apps/plugin-opener : 2.2.6 (outdated, latest: 2.2.7)
    - tauri-plugin-http 🦀: 2.4.3, (outdated, latest: 2.4.4)
    - @tauri-apps/plugin-http : 2.4.3 (outdated, latest: 2.4.4)
    - tauri-plugin-dialog 🦀: 2.2.1, (outdated, latest: 2.2.2)
    - @tauri-apps/plugin-dialog : 2.2.1 (outdated, latest: 2.2.2)
    - tauri-plugin-os 🦀: 2.2.1
    - @tauri-apps/plugin-os : 2.2.1

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - framework: SolidJS
    - bundler: Vite

[-] iOS
    - Developer Teams: Mikkel ALMONTE--RINGAUD (ID: WLAPHT4WXV)
```

### Stack trace

```text

```

### Additional context

_No response_

Legend-Master:
Closes #13577


crates/tauri/src/plugin.rs
```diff
@@ -328,13 +356,45 @@ impl<R: Runtime, C: DeserializeOwned> Builder<R, C> {
...
   ///   Builder::new("example")
-  ///     .js_init_script(INIT_SCRIPT.to_string())
+  ///     .js_init_script(INIT_SCRIPT)
   ///     .build()
   /// }
   /// ```
   #[must_use]
-  pub fn js_init_script(mut self, js_init_script: String) -> Self {
-    self.js_init_script = Some(js_init_script);
+  pub fn js_init_script(mut self, js_init_script: impl Into<String>) -> Self {
```
FabianLars, line 370:
```suggestion
  // TODO: Rename to initialization_script in v3
  pub fn js_init_script(mut self, js_init_script: impl Into<String>) -> Self {
```

we'd may have to rename the trait method but what do you think about this? imo it's so weird that we have js_init_script (plugin) and initialization_script (webview), both of which doing the same thing.

Legend-Master: Yep, makes sense to me

crates/tauri/src/plugin.rs
```diff
@@ -50,15 +51,37 @@ pub trait Plugin<R: Runtime>: Send {
...
+  ///   For remote URLs, we use [onPageStarted] which is not guaranteed to run before other scripts.
+  ///
+  /// [addDocumentStartJavaScript]: https://developer.android.com/reference/androidx/webkit/WebViewCompat#addDocumentStartJavaScript(android.webkit.WebView,java.lang.String,java.util.Set%3Cjava.lang.String%3E)
+  /// [onPageStarted]: https://developer.android.com/reference/android/webkit/WebViewClient#onPageStarted(android.webkit.WebView,%20java.lang.String,%20android.graphics.Bitmap)
   fn initialization_script(&self) -> Option<String> {
     None
   }
 
+  // TODO: Change `initialization_script` to this in v3
+  /// Same as [`Plugin::initialization_script_2`] but returns an [`InitializationScript`] instead
```
lucasfernog:
```suggestion
  /// Same as [`Plugin::initialization_script`] but returns an [`InitializationScript`] instead
```

lucasfernog (approved):

PR #13618, opened 2025-06-15
fix(cli): prevent .app identifier (fix #12674)

### LINKED ISSUE #12674: [bug] Prevent `.app` in identifier
### Describe the bug

This will cause undesired behavior on macOS cause `.app` is a system thing for apps.

Image

### Reproduction

use any `identifier` that ends in `.app` and you'll see macOS thinking it's an app instead of a directory.

### Expected behavior

The tauri runtime should throw an error and if people still want this behavior have a flag to allow it.

### Full `tauri info` output

```text
N/A
```

### Stack trace

```text

```

### Additional context

_No response_

Sky-walkerX:
closes #12674

### DIFF
--- a/crates/tauri-cli/src/build.rs
+++ b/crates/tauri-cli/src/build.rs
@@ -160,7 +160,7 @@ pub fn setup(
       std::process::exit(1);
     } else {
       log::warn!(
-        "The bundle identifier \"{}\" set in `{} identifier` ends with `.app`. This is not recommended as it will cause issues on macOS.",
+        "The bundle identifier \"{}\" set in `{} identifier` ends with `.app`. This is not recommended because it conflicts with the application bundle extension on macOS.",
         config_.identifier,
         bundle_identifier_source
       );
--- a/crates/tauri-cli/src/build.rs
+++ b/crates/tauri-cli/src/build.rs
@@ -150,6 +150,23 @@ pub fn setup(
     std::process::exit(1);
   }
 
+  if config_.identifier.ends_with(".app") {
+    if cfg!(target_os = "macos") {
+      log::error!(
+        "The bundle identifier \"{}\" set in `{} identifier` must not end with `.app` on macOS because it conflicts with the application bundle extension.",
+        config_.identifier,
+        bundle_identifier_source
+      );
+      std::process::exit(1);
+    } else {
+      log::warn!(
+        "The bundle identifier \"{}\" set in `{} identifier` ends with `.app`. This is not recommended as it will cause issues on macOS.",
+        config_.identifier,
+        bundle_identifier_source
+      );
+    }
+  }
+
   if let Some(before_build) = config_.build.before_build_command.clone() {
     helpers::run_hook("beforeBuildCommand", before_build, interface, options.debug)?;
   }
--- a/.changes/warn-dot-app-id.md
+++ b/.changes/warn-dot-app-id.md
@@ -1,0 +1,6 @@
+---
+"tauri-cli": "patch:feat"
+"@tauri-apps/cli": "patch:feat"
+---
+
+Warn the user that the app id shouldn't end in `.app` because it conflicts with the application bundle extension on macOS


Hacksore (approved):

crates/tauri-cli/src/build.rs
```diff
@@ -150,6 +150,23 @@ pub fn setup(
...
+    if cfg!(target_os = "macos") {
+      log::error!(
+        "The bundle identifier \"{}\" set in `{} identifier` must not end with `.app` on macOS because it conflicts with the application bundle extension.",
+        config_.identifier,
+        bundle_identifier_source
+      );
+      std::process::exit(1);
+    } else {
+      log::warn!(
+        "The bundle identifier \"{}\" set in `{} identifier` ends with `.app`. This is not recommended as it will cause issues on macOS.",
```
Legend-Master: Maybe we could explain the reason on other platforms as well? (`because it conflicts with the application bundle extension`)

Sky-walkerX: Done, please check.

Legend-Master (approved):
Thanks!

Since I'm not familiar with how this really works on macOS, I'll ask @FabianLars to review as well

FabianLars (approved):

Legend-Master:
I wonder if we should also update the default app id in `create-tauri-app` as it defaults to `com.tauri-app.app` which could cause problems if the user just wants to experiment with things without changing that first

Sky-walkerX:
> I wonder if we should also update the default app id in `create-tauri-app` as it defaults to `com.tauri-app.app` which could cause problems if the user just wants to experiment with things without changing that first

Maybe we could prompt the user to set a custom appId, or just generate one from the project name

Legend-Master:
Technically, we are doing that, just thinking we might want to get rid of the `.app` part by default

image

lucasfernog:
this is actually a breaking change, there's a bunch of apps ending with `.app` that do not really care about the config paths (i'm actually working on one of those :D )

crates/tauri-cli/src/build.rs
```diff
@@ -150,6 +150,23 @@ pub fn setup(
...
+  if config_.identifier.ends_with(".app") {
+    if cfg!(target_os = "macos") {
+      log::error!(
+        "The bundle identifier \"{}\" set in `{} identifier` must not end with `.app` on macOS because it conflicts with the application bundle extension.",
+        config_.identifier,
+        bundle_identifier_source
+      );
+      std::process::exit(1);
+    } else {
+      log::warn!(
```
lucasfernog, line 162: I would always warn instead of exiting like above

Hacksore:
Yeah I suppose if you are not caring about macOS and you have a `.app` your builds will now fail which is not desired.

An opt-in CLI flag would not make sense to me but has tauri considered something like compatibility dates in configs so we can introduce semi-breaking changes like this?

[How cloudflare does it via Compatibility dates](https://developers.cloudflare.com/workers/configuration/compatibility-dates/).

FabianLars:
What about those that wkwebview creates?

True about the braking change, didn't really think about that but imo we should still really consider making this a hard error in a future major. (With an opt-out for migrators ig)

Legend-Master:
So I guess we just warn about it for now? I'm not sure if the directory is not usable or just can't be used in finder, if it's just in finder, I guess just a warning would be a better choice for now

Sky-walkerX:
So should I do log::warn and remove log::error and exit from the code block ?

Legend-Master:
Yeah, let's do that, we can always make it a hard fail in the future

Sky-walkerX:
> Yeah, let's do that, we can always make it a hard fail in the future

PR -> #13627, please check

lucasfernog:
> So I guess we just warn about it for now? I'm not sure if the directory is not usable or just can't be used in finder, if it's just in finder, I guess just a warning would be a better choice for now

it's still usable, but Finder shows it as an app and you'd need to actually rightclick -> show package contents to open it

PR #13621, opened 2025-06-16
fix(tauri-utils): Use write_if_changed more

Flakebi:
Replace `fs::write` with `write_if_changed` in two places. This can prevent unnecessary rebuilds. (I didn’t encounter any, but this should be ok nonetheless.)

### DIFF
--- a/crates/tauri-utils/src/acl/build.rs
+++ b/crates/tauri-utils/src/acl/build.rs
@@ -91,7 +91,8 @@ pub fn define_permissions<F: Fn(&Path) -> bool>(
   let pkg_name_valid_path = pkg_name.replace(':', "-");
   let permission_files_path = out_dir.join(format!("{}-permission-files", pkg_name_valid_path));
   let permission_files_json = serde_json::to_string(&permission_files)?;
-  fs::write(&permission_files_path, permission_files_json)
+
+  write_if_changed(&permission_files_path, permission_files_json)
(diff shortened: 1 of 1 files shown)


Legend-Master:
I believe these 2 places are used only in target (`OUT_DIR`) directory and we don't need to guard it here

Flakebi:
`write_if_changed` should be used when the generated file is a dependency for another build step or process (i.e. by `include!`ing the generated file).
If the written file changes, cargo will trigger the dependent build step. If the file doesn’t change, cargo can skip the dependent step.

I agree with you that write_if_changed shouldn’t matter here if `src` is read-only or not.
But for the benefit of unnecessary rebuilds, it shouldn’t really matter if the file is in OUT_DIR or somewhere else, `write_if_changed` can have benefits nonetheless.
For what it’s worth, there are other places that use write_if_changed for files in OUT_DIR: https://github.com/tauri-apps/tauri/blob/100dc94c4847ee9a684f38f73d882dc5ac174941/crates/tauri-utils/src/acl/build.rs#L306

(Sorry for the slow response)

Legend-Master (approved):
> But for the benefit of unnecessary rebuilds, it shouldn’t really matter if the file is in OUT_DIR or somewhere else, write_if_changed can have benefits nonetheless.

Sounds reasonable to me, thanks for explaining

> (Sorry for the slow response)

No problems 🙃

Issue #13284, opened 2025-04-23
[feat] Allow `--config` argument when running `tauri ios init`


voxelbee:
### Describe the problem

It would make sense to allow the `--config` parameter when using `tauri <ios|android> init` as this would allow us to setup different flavors correctly. Currently I created 2 extra config files `tauri.conf.dev.json` and   `tauri.conf.stg.json`. These have a custom `identifier` set inside and currently I need to regenerate the XCode project to build these custom identifiers but the init command doesn't allow specifiying a different config. It only uses the default one.

### Describe the solution you'd like

We should be able to use the `--config tauri.conf.stg.json` the same as when we run the `dev` or `build` commands.

### Alternatives considered

Maybe a better setup for different flavors could be discussed where we don't have to regenerate the project files to change which flavor we are building. There could be different targets in the XCode project for each flavor? But for now just allowing `--config` in the `init` command will allow this to work.

voxelbee:
Well when runing `npm tauri ios build --export-method debugging --config src-tauri/tauri.conf.dev.json` or `npm tauri ios dev --config src-tauri/tauri.conf.dev.json  ` i get the following error: [#10925](https://github.com/tauri-apps/tauri/issues/10925) it would be great to have proper flavors support.

```
> ponderosa@0.1.0 tauri /Users/peytonhammersley/src/ponderosa/app
> tauri ios xcode-script -v --platform iOS --sdk-root /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS18.0.sdk --framework-search-paths '/Users/peytonhammersley/Library/Developer/Xcode/DerivedData/Ponderosa-dodgonqdttjvblcqnpyegkjbvcuj/Build/Products/release-iphoneos  "."' --header-search-paths '/Users/peytonhammersley/Library/Developer/Xcode/DerivedData/Ponderosa-dodgonqdttjvblcqnpyegkjbvcuj/Build/Products/release-iphoneos/include ' --gcc-preprocessor-definitions '' --configuration release arm64


thread '<unnamed>' panicked at crates/tauri-cli/src/mobile/mod.rs:415:6:
failed to read CLI options: Error when opening the TCP socket: Connection refused (os error 61)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
 ELIFECYCLE  Command failed.
[...]

PR #13625, opened 2025-06-17
fix(cli): use original identifier to fix mobile options reading

lucasfernog:
the iOS and Android CLI commands leverage an android_studio_script/xcode_script that is executed by the native IDE or build tool. This script reads the Tauri configuration to find the app identifier used to communicate with the parent Tauri CLI process to read CLI options.

The communication is broken when the `--config` arg is used, since the IDE script does not have access to that value before reaching the parent process, which is impossible without knowing the actual identifier used.

To bypass this we'll agree on using the original identifier. This obviously won't work if the original tauri.conf.json do not have an identifier, so we error out in this case

### DIFF
--- a/.changes/fix-mobile-identifier-changes.md
+++ b/.changes/fix-mobile-identifier-changes.md
@@ -1,0 +1,6 @@
+---
+"tauri-cli": patch:bug
+"@tauri-apps/cli": patch:bug
+---
+
+Fixes Android and iOS dev/build commands not working when the app identifier is being modified by the `--config` option.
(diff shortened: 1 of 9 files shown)


FabianLars:
The code doesn't load on my phone, does this help with https://github.com/tauri-apps/tauri/issues/13284 ?

lucasfernog:
> The code doesn't load on my phone, does this help with #13284 ?

that's a different problem honestly.. it's related, but after initializing with the given identifier, the xcode_script/android_studio_script would still need to be able to know it.. we do have code to parse it from the Xcode project config, but for Android it's different (and would need to account for Kotlin vs Gradle syntax)

lucasfernog:
this closes the actual bug reported on this comment https://github.com/tauri-apps/tauri/issues/13284#issuecomment-2824555135

FabianLars (approved):
(untested)

PR #13629, opened 2025-06-17
Update kuchikiki, html5ever, tao, wry, webview2-com

sftse:
The subtree involving `kuchikiki` and `html5ever` holds back a number of dependencies from updating. This PR is a small step towards updating that subtree, which contains `syn 1.0` and `rand 0.7.3`.

This updates just `kuchikiki` and `html5ever` to the most recent version both are compatible with. The update was not as straightforward as initially assumed. Merging this PR removes the dependency on `tauri-utils 1.0`, which was leading to cargo dependency resolution errors, and inlines unstable functions from `schemars` that were relied on and no longer exist.

This depends on a private updated version of `wry` with which it will compile. The PR to update `wry` in the same way is [here](https://github.com/tauri-apps/wry/pull/1566).

### DIFF
--- a/.changes/kuchikiki-bump.md
+++ b/.changes/kuchikiki-bump.md
@@ -1,0 +1,7 @@
+---
+"tauri-cli": minor:deps
+"@tauri-apps/cli": minor:deps
+"tauri-utils": minor:deps
+---
+
+Update html5ever to 0.29 and kuchikiki to version 0.8.8-speedreader.
--- a/.changes/tao-wry-bump.md
+++ b/.changes/tao-wry-bump.md
@@ -1,0 +1,6 @@
+---
+"tauri": minor:deps
+"tauri-runtime-wry": minor:deps
+---
+
+Updated tao to 0.34, wry to 0.52 and webview2-com to 0.38.
--- a/.changes/x11-feature.md
+++ b/.changes/x11-feature.md
@@ -1,0 +1,7 @@
+---
+"tauri": minor:feat
+"tauri-runtime-wry": minor:feat
+---
+
+Added `x11` Cargo feature (enabled by default). Disabling it is useful for apps that only support Wayland, reducing its size.
+**NOTE**: When manually disabling tauri default features, you must enable the `x11` feature to support it.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7205,9 +7205,9 @@ dependencies = [
 
 [[package]]
 name = "schemars"
-version = "0.8.16"
+version = "0.8.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "45a28f4c49489add4ce10783f7911893516f15afe45d015608d41faca6bc4d29"
+checksum = "09c024468a378b7e36765cd36702b7a90cc3cba11654f6685c8f233408e89e92"
 dependencies = [
  "dyn-clone",
  "indexmap 1.9.3",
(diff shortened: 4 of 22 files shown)


FabianLars:
i really appreciate the work you've poured into this, but is this really the only way? What about some feature flags to control the used deps in the v1 crate perhaps?

having to duplicate the code and maintain it in multiple places is something i really really want to avoid (think about more major versions as well).

sftse:
I think that would depend on whether v1 is EOL and no longer supported or still receiving maintenance patches. As more major versions are released it becomes increasingly tedious to patch previous versions to prevent pulling along the old dependencies.

In this case, imho the trade-off is in favor of inlining the module. The majority of the inlined code is type definitions and the probability of a bug there is low.

It may make sense to remove v1 code entirely if v3 is released. This would mean v1 users cannot automatically migrate to v3, which seems reasonable.

sftse:
A possible alternative would be to feature gate `html5ever` and `kuchikiki` in `tauri-utils v1` and remove its use of private api functions from `schemars`. But that is a temporary fix, on a long enough time frame all `tauri-utils v1` dependencies are likely to become outdated.

FabianLars:
> I think that would depend on whether v1 is EOL and no longer supported or still receiving maintenance patches. As more major versions are released it becomes increasingly tedious to patch previous versions to prevent pulling along the old dependencies.

What i had in mind was that we've had a few v1 config updates even after v2 was released. But i guess there's a somewhat high chance we'll have to make changes in the migrator anyway.

> In this case, imho the trade-off is in favor of inlining the module. The majority of the inlined code is type definitions and the probability of a bug there is low.

wasn't thinking about straight bugs but rather us forgetting to update both branches (based on past mistakes)

> It may make sense to remove v1 code entirely if v3 is released. This would mean v1 users cannot automatically migrate to v3, which seems reasonable.

i'm in favor of that at least even ignoring this discussion here.

---

you know what, i changed my mind. since there isn't really any work being done on v1 anyway anymore. with better MSRV policies in the future we also shouldn't hit issues like this again so in the v3 cli we may not have to inline the v2 config (or with 1.77 we may have to wait for v4)

FabianLars:
@lucasfernog @Legend-Master any hard feelings?

sftse:
@FabianLars The PR to [wry](https://github.com/tauri-apps/wry/pull/1566) needs to be released first, then I'll update to include it.

FabianLars:
yepyep, idk why i started ci here tbh, probably muscle memory.

Legend-Master:
I'm not 100% sure but I believe the old `tauri-utils` was only pulled in for the config files, would it make sense to feature gate things? (considering these 2 dependencies are gated with `html-manipulation` in v2) (to be honest I feel like it might deserve its own crate 😂)

Don't mind inlining it either, I feel like we don't do much on v1 right now anyways and v3 is sorta coming

FabianLars:
> (to be honest I feel like it might deserve its own crate 😂)

if you mean something like `tauri-config` then that thought struck me a few times the last few weeks as well. I think the utils crate does too much.
i think the acl module would be a good target for another crate as well but didn't really look at it (nor did i work in it ever).

sftse:
> I'm not 100% sure but I believe the old `tauri-utils` was only pulled in for the config files, would it make sense to feature gate things? (to be honest I feel like it might deserve its own crate 😂)

Yes, that is right, only the config module had to be inlined.
I'm cautious about feature flags as exponentially many combinations need to be tested and when dependency resolution does not work as expected it can be quite time consuming to fix. Doing this upgrade with the inline code would have been trivial.

Legend-Master:
Spent some time playing with splitting the config to a new crate, and it is harder than I thought, the configs are not just configs, they're also types we use at runtime internally and contains `ToToken` implementations we use in codegen, also some `impl`s used in various places (especially around the ACL stuffs), since we can't `impl` for another crate, that means we can't separate them easily and still do the `impl` in `tauri-utils`

lucasfernog (approved):

Issue #8911, opened 2024-02-20
[bug] bundleresources is not working for Android


vangork:
### Describe the bug

It won't bundle the resources specified in the bundleresources config for Android (`yarn tauri android dev`). Does this option apply to Android?
```
"bundle": {
    "resources": [
        "resources/*"
    ]
}
```

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.22000 X64
    ✔ WebView2: 121.0.2277.128
    ✔ MSVC: Visual Studio Community 2022
    ✔ rustc: 1.71.0 (8ede3aae2 2023-07-12)
    ✔ cargo: 1.71.0 (cfd3bbd8f 2023-06-08)
    ✔ rustup: 1.26.0 (5af9b9484 2023-04-05)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (default)
    - node: 18.16.0
    - yarn: 1.22.19
    - npm: 9.5.1

[-] Packages
    - tauri [RUST]: 2.0.0-alpha.21
    - tauri-build [RUST]: 2.0.0-alpha.14
    - wry [RUST]: 0.35.2
    - tao [RUST]: 0.24.0
    - tauri-cli [RUST]: 1.3.1
    - @tauri-apps/api [NPM]: 2.0.0-alpha.13
    - @tauri-apps/cli [NPM]: 2.0.0-alpha.20

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../dist
    - devPath: http://localhost:1420/
    - framework: Vue.js
    - bundler: Vite
```

wyatt-herkamp:
I am having the same issue.

wyatt-herkamp:
I am using this for a school project. I found a workaround. By making a zip file in the build.r.s and using include_bytes. Then I extract the files if needed. I know it isn't the best option. However, this is a small school project.

vangork:
@wyatt-herkamp FYI. I just follow [android doc](https://developer.android.com/guide/topics/resources/providing-resources#OriginalFiles) and handle the files with an android plugin.

loikki:
I had the same issue and decided to move towards using this library https://github.com/pyrossh/rust-embed. It works very well and simplifies a lot the data management.

biaggii:
I had the same issue, i can used resource by typescript frontend, but rustend can't

oddpxl:
It works just fine on MacOS.

[...]

Issue #13490, opened 2025-05-22
How to add DLLs for different architectures?


jiangkaifree:
Okay, in my project development, I need to use the dll file provided by the c++ engineer. However, his dll depends on some system dlls, such as vcruntime140.dll. Now I have encountered some problems. The vcruntime140.dll is different on different Windows architectures. Now I have some definitions in my configuration file `tauri.window.config.json`：
```
{
  "tauri": {
    "bundle": {
      "resources": ["*.dll", "conf/*", "dependency/*/*", "log/*", "config.ini"]
     }
  }  
}
```
I want to know how to incorporate different DLL files into installation packages for different Windows architectures (x64, x86).
I hope you can help me and give me some solutions. Thank you very much.

Issue #8501, opened 2023-12-30
[feat] platform/arch-specific resources


zhangsan946:
### Describe the problem

In additional to executable files, static/dynamic library files are also platform & arch dependent. Need a way to embed external binaries for other extentions such as: .dll, .so,  beside executable files.

### Describe the solution you'd like

Support externalBin setup as following:
```
"externalBin": [
    "binaries/my-sidecar"
]
```
And in additional:
```
"externalBin": [
    {path: "binaries/my-sidecar-dll", ext: "dll"}
]
```

FabianLars:
I'd say the resources feature is what you're asking for <https://tauri.app/v1/guides/building/resources/>. They can be platform specific via https://tauri.app/v1/api/config#platform-specific-configuration.

Your pr also only works on windows because linux and macos typically don't have library files next to executables.

FabianLars:
whoops, didn't mean to close this before you have a chance to respond

zhangsan946:
Hi @FabianLars, thanks for the reponse. The resource feature won't help bescause the library file is also arch dependent like the excutable file. Please take the wintun.dll file as an example https://www.wintun.net/builds/wintun-0.14.1.zip. It has to keep the name of `wintun.dll` and it has different version for x86, amd64, arm.. unless resource feature could be not only platform specific  but also arch specific.
On linux, library search path is `LD_LIBRARY_PATH` specified path,  executable's rpath, /lib folder. User could set the LD_LIBRARY_PATH if the executable's rpath is not set to be the same folder to executables.

FabianLars:
> The resource feature won't help bescause the library file is also arch dependent like the excutable file

right, maybe we could expand the platform specific configs to also be arch specific (optionally) but i can't think of a different usecase than this issue so maybe that's a bit too much. Using the `--config`/`-c` flag works but having to use a manual step for this feels awkward too.

> On linux, library search path is LD_LIBRARY_PATH specified path, executable's rpath, /lib folder. User could set the LD_LIBRARY_PATH if the executable's rpath is not set to be the same folder to executables.

That's why i said your solution doesn't work. In the .deb bundle the sidecars will be installed into /usr/bin where .so files don't belong. In the .appimage bundle the rpath and iirc the ld_library_path are set but either way appimages also have an internal `/usr/bin/` folder where it wouldn't be as bad to have .so files in but still unusual.

----

[...]

Issue #6802, opened 2023-04-26
[feat] Add opportunity to include .dll or .so libraries along with the binary


Priler:
### Describe the problem

Hi!
I'm using [Vosk ](https://docs.rs/vosk/latest/vosk/)with Rust Tauri.
In order to work properly, Vosk libraries should be bundled with the Tauri binary or included in the PATH of a user.

With the Rust itself bundling Vosk libraries is easy and even can be automated with tools like [cargo make](https://github.com/sagiegurari/cargo-make).
But with Tauri, it does not work.
Because bundling is make with `cargo tauri build` command, and `cargo make` totally skipped in process.

Thanks!

### Describe the solution you'd like

Allow bundling libraries like `.dll` or `.so` alongside with sidecars and resources.
This libraries should be placed with the app `.exe` file.

amrbashir:
Have you tried using https://tauri.app/v1/api/config#bundleconfig.resources ? also you can change the runner using https://tauri.app/v1/api/config/#buildconfig.runner which you maybe able to set to `cargo-make` binary if it exists in `PATH`.

FabianLars:
Closing in favor of #11812 cause it has more traction

PR #13640, opened 2025-06-20
docs: where does resource dir resolve to

Legend-Master:
Reference: #8911


FabianLars:
do we have any stability guarantees or is this a "may change at anytime" kinda situation (even though i think a change is unlikely)

Legend-Master:
Yeah, nice catch actually, this path is unlike other paths, you would normally only read not write in this directory, so we could potentially change things, to be fair though, it would be quite unlikely except for maybe android

Legend-Master:
I can't think of a good word for saying this is unlikely but they might change 😂, any suggestions that doesn't sound scary?

FabianLars:
tbh i wouldn't even include that. i'd keep it simple and go with something like "these paths may change at any time". or maybe include something like "these paths are considered an implementation detail..." (not sure if i got the definition of that word right).

they'd also be likely different for new bundle types as well.

since people usually "abuse" this api for the exe location we should probably just cave and give them the api they want, documenting its limitations (read only blabla) ~~and laugh at them if they do the obvious mistakes anyway~~ (yes this topic annoys me, and i even brought it up myself lol)

Legend-Master:
Maybe something similar to this? 

>https://doc.rust-lang.org/std/io/index.html#platform-specific-behavior

Went with 

> Although we provide the exact path where this function resolves to, this is not a contract and things might change in the future

FabianLars (approved):

WSH032:
Another use case for `resource_dir` is to bundle dynamic libraries and then load them via rpath:

- <https://github.com/tauri-apps/tauri/issues/13490>
- <https://github.com/tauri-apps/tauri/issues/8501>
- <https://github.com/tauri-apps/tauri/issues/6802>

I'm not sure about the situation on mobile platforms, but on desktop, changing these paths would break these workflows.
I admit that using `resource_dir` to bundle dynamic libraries is not a good practice, but currently there is no better alternative in tauri to achieve this.

Legend-Master:
I mean, if we change this path, we would also change the bundler code to match that, so the resources will get placed to the new place, then you can still use it to read files or load dynamic libraries as before right?

WSH032:
> I mean, if we change this path, we would also change the bundler code to match that, so the resources will get placed to the new place, then you can still use it to read files or load dynamic libraries as before right?

[rpath is set at compile time](https://github.com/pytauri/create-pytauri-app/blob/380943a30eefe0e51b852d48c232cdeaea8fe478/templates/%7B%7B%20'.'%20%7D%7D/%7B%7B%20project_name%20%7D%7D/scripts/linux/build.sh.jinja#L12), and `resource_dir` should only be used at runtime, right?

Also, on Windows, DLLs usually have to be placed next to the executable.

*I admit this approach is quite fragile, and maintaining such cases is a burden for both tauri and downstream users. I haven't come up with a better solution for now.*

Legend-Master:
I kinda see, I was thinking about loading them dynamically at runtime but forcing that feels quite wrong as that's quite a bit harder  than dynamic linking

I think we might want a dedicated config option for this to be honest since this path could vary quite a bit on different platforms (also from the normal resource directories)

WSH032:
> I think we might want a dedicated config option for this to be honest since this path could vary quite a bit on different platforms (also from the normal resource directories)

Yeah, I agree. I think this is more or less the issue that #8501 aims to solve. If a new option is introduced in the future, I'd be happy to migrate from `resource_dir` to it.

PR #13661, opened 2025-06-23
feat: add `WebView::set_cookie` and `WebView::delete_cookie`

WSH032:
ref #12665
wry pr tauri-apps/wry#1569


crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -3649,6 +3683,18 @@ fn handle_user_message<T: UserEvent>(
...
+              log::error!("failed to set webview cookie: {e}");
+            }
+          }
+
+          WebviewMessage::DeleteCookie(cookie) => {
+            if let Err(e) = webview.delete_cookie(&cookie) {
+              log::error!("failed to delete webview cookie: {e}");
+            }
+          }
+
```
WSH032, line 3725: I have some questions. Why do some methods use a `channel` to wait for a `Result<()>` to return, while others simply ignore (log) errors?

WSH032 (commented):
This PR is ready, but it needs to wait for the Wry PR to be merged and for Tauri to upgrade the Wry version.

WSH032:
Oops, something went wrong on Linux. Should we enable `x11` for Wry?
Anyway, I'm done for today and will handle it tomorrow.

AlloryDante:
Waiting for this to be merged. Cant wait to test it out ❤️

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -1688,6 +1690,30 @@ impl<T: UserEvent> WebviewDispatch<T> for WryWebviewDispatcher<T> {
...
+    send_user_message(
+      &self.context,
+      Message::Webview(
+        *self.window_id.lock().unwrap(),
+        self.webview_id,
+        WebviewMessage::DeleteCookie(cookie.clone().into_owned()),
+      ),
+    )?;
+    Ok(())
+  }
```
WSH032: I've changed my mind. I think the signature here should use `cookie: Cookie<'static>` instead of implicitly cloning it.
If we change our minds in the future, we can still optimize it using `AsRef`.

But I'm going to sleep now.

WSH032 (commented):
Also, I think we should re-export `cookie` crate at `mod webview`, otherwise users won't be able to construct `Cookie` instances directly.

lucasfernog (approved):
nice one!

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -1688,6 +1690,30 @@ impl<T: UserEvent> WebviewDispatch<T> for WryWebviewDispatcher<T> {
...
+
+  fn delete_cookie(&self, cookie: Cookie<'_>) -> Result<()> {
+    send_user_message(
+      &self.context,
+      Message::Webview(
+        *self.window_id.lock().unwrap(),
+        self.webview_id,
+        WebviewMessage::DeleteCookie(cookie.clone().into_owned()),
```
WSH032, line 1711: The `cookie.clone()` here was forgotten to be removed.

crates/tauri/src/webview/mod.rs
```diff
@@ -7,13 +7,19 @@
...
+///
+/// # Stability
+///
+/// This re-exported crate is still on an alpha release and might receive updates in minor Tauri releases.
+pub use cookie;
 use http::HeaderMap;
 use serde::Serialize;
 use tauri_macros::default_runtime;
 pub use tauri_runtime::webview::{NewWindowFeatures, PageLoadEvent};
-pub use tauri_runtime::Cookie;
```
WSH032, line 16: We used to expose this as a [public API](https://docs.rs/tauri/latest/tauri/webview/struct.Cookie.html). Wouldn't removing it cause some issues?

Legend-Master: Yeah, we should probably bring it back

Edit: done in #14020

AlloryDante:
I have this function that I've made and I try to delete the session cookie. The use case is that I make an account manager and I need to handle sessions on multiple roblox accounts. 

The code successfully enters in the if statement but the delete_cookie doesnt seem to actually delete it. I refresh the page, I run the command multiple times and it simply does not work.

Does anyone has any ideeas why this might be the case?

```rs
#[tauri::command]
async fn delete_session(window: tauri::WebviewWindow) {
    
    let cookies = window.cookies().unwrap();
    for cookie in cookies {
        if cookie.name() == ".ROBLOSECURITY" {
            println!("{:?}", cookie);
            window.delete_cookie(cookie).unwrap();
            
            println!("Deleted .ROBLOSECURITY cookie");
        }
    }
    window.eval("window.location.replace('https://www.roblox.com/')").unwrap();
}
```

PR #13690, opened 2025-06-26
fix(macros): cache rustc -V output

montyc1999:
The `#[tauri::command]` macro runs a `rustc -V` version check for stateful async commands.

Each `rustc -V` takes 50-100ms on my machine, which can add up to several seconds for a Tauri app with many commands.

To demonstrate this, try adding 100 async commands to the `helloworld` example like so:
https://github.com/montyc1999/tauri/commit/9b0de6b13396c63202c6116f3e606fa19582c343

On my machine, an incremental build with `touch examples/helloworld/main.rs && cargo build --example helloworld` takes **4.6s** without caching, or **1.2s** with caching.


Legend-Master (dismissed):
Nice catch, thanks!

Legend-Master (approved):

Issue #13753, opened 2025-07-02
[feat] mock: add mock for listen, emit and channel


JosephBrooksbank:
### Describe the problem

In #13744, @Legend-Master mentions "we should probably implement something that handles (mocks) channels and event listeners"

### Describe the solution you'd like

I was already in the process of doing so in my local project so I decided to throw it into the base `mock.ts` instead. 
[Commit here](https://github.com/JosephBrooksbank/tauri/commit/19adf8abf0de5ee53b27970230e9849f7360397e)

This works for my use of `listen` already. I'm able to `emit` a message during the test and it handles just like one from the real Rust backend would:

```typescript
	test('testing emit', async () => {

		mockIPC((cmd, _: unknown) => {
			if (cmd === 'some-other-invoke') {
				return true; // current mocking works as expected
			}
		});

		render(<App />);
		act(() => {
			emit('get-stats', {
				total_cpu: 100,
				total_mem: '2000',
				name: 'pod 1'
			});
		})

		await waitFor(() => {
			expect(screen.geByText(/Machine Stats/i)).toBeInTheDocument();
			expect(screen.getByText(/"name": "\/pod 1"/i)).toBeInTheDocument();
			expect(screen.getByText(/"total_cpu": 100/i)).toBeInTheDocument();
			expect(screen.getByText(/"total_mem": "2000"/i)).toBeInTheDocument();
		});
	});
```

I'm sure there are other situations that don't work though, like `emit_to` or channels. 

Should I continue work on this? Is the current implementation I have sufficient to make a PR for?

Legend-Master:
Some support is better than no support at all 🙃 as long as we document the limitations, a PR would be much appreciated

Legend-Master:
Will keep this open since channels are not supported yet

PR #13783, opened 2025-07-08
feat(mock): add mock for listen and emit

JosephBrooksbank:
created from #13753 

Allow events emitted with `emit` to be handled correctly by `listen` callbacks when in a mocked environment.

Sibling PR [here](https://github.com/tauri-apps/tauri-docs/pull/3395) for a documentation update


packages/api/src/mocks.ts
```diff
@@ -59,19 +62,88 @@ function mockInternals() {
...
+
+  function handleEventPlugin(cmd: string, args?: InvokeArgs): unknown {
+    switch (cmd.split('|')[1]) {
+      case 'listen':
+        return handleListen(args as { event: EventName; handler: number })
+      case 'emit':
+        return handleEmit(args as { event: EventName; payload?: unknown })
+      case 'unlisten':
+        return handleRemoveListener(args as { event: EventName; id: number })
+    }
```
Legend-Master:
```suggestion
    switch (cmd) {
      case 'plugin:event|listen':
        return handleListen(args as { event: EventName; handler: number })
      case 'plugin:event|emit':
        return handleEmit(args as { event: EventName; payload?: unknown })
      case 'plugin:event|unlisten':
        return handleRemoveListener(args as { event: EventName; id: number })
    }
```

I know the intention was to reduce duplications, but I think this makes it easier to understand

packages/api/src/mocks.ts
```diff
@@ -107,11 +179,17 @@ export function mockIPC(
...
+  function unregisterListener(event: EventName, id: number) {
+    unregisterCallback(id)
+  }
+
   window.__TAURI_INTERNALS__.invoke = invoke
   window.__TAURI_INTERNALS__.transformCallback = registerCallback
   window.__TAURI_INTERNALS__.unregisterCallback = unregisterCallback
   window.__TAURI_INTERNALS__.runCallback = runCallback
   window.__TAURI_INTERNALS__.callbacks = callbacks
+  window.__TAURI_EVENT_PLUGIN_INTERNALS__.unregisterListener =
```
Legend-Master, line 205: I think we can skip mocking this, it's only used internally for the js side to immediately unregister the listeners, but since we're already mocking all things in js, it's no longer needed

Legend-Master (dismissed):
Thanks! Just some small nitpicks

packages/api/src/mocks.ts
```diff
@@ -59,19 +62,88 @@ function mockInternals() {
...
+
   // eslint-disable-next-line @typescript-eslint/require-await
   async function invoke<T>(
     cmd: string,
     args?: InvokeArgs,
     _options?: InvokeOptions
   ): Promise<T> {
+    if (isEventPluginInvoke(cmd)) {
+      return handleEventPlugin(cmd, args) as T
+    }
```
Legend-Master: Actually, I'm not entirely sure if this is fine to do, since this way you can't handle these events yourself even if you wanted to, this will break the code for people doing custom handlings, maybe we should make this opt-in until v3?

packages/api/src/mocks.ts
JosephBrooksbank: [The docs](https://tauri.app/develop/calling-frontend/#listening-to-events-on-the-frontend) make note that you should call `unlisten` whenever your component unmounts, so I have `useEffect`s like this:

```typescript
	useEffect(() => {
		invoke("start_monitoring", { containerName });
		const unlisten = listen<IContainerStats>(`${containerName}-stats`, (event) => {
			setStats(event.payload);
		});

		return () => {
			unlisten.then(f => f());
		};
	}, [containerName]);
```

The function returned by `listen` (`unlisten`) is real code in `event.ts` which does call `window...unregisterListener`. If I don't include this as a function the unit tests error. 

I could just stub `unregisterListener` if you think that's better, I agree that the behavior isn't really necessary for unit tests.

packages/api/src/mocks.ts
JosephBrooksbank: Good call on that.

packages/api/src/mocks.ts
Legend-Master: Right right, totally forgot that

Legend-Master (approved):
Thanks!

PR #13786, opened 2025-07-08
fix(nsis): per-machine installer not requesting elevation when run by non-admin users

catalinsh:
According to the [NSIS docs](https://nsis.sourceforge.io/Reference/RequestExecutionLevel) `RequestExecutionLevel highest` will request the highest execution level available for the current user.

This makes it so that, if the user launching the installer is not an Administrator, the installer will not ask to be elevated and the installation will fail when reaching a step that requires admin priviledges.

Switching to `RequestExecutionLevel admin` ensures the installer always asks for elevation.

### DIFF
--- a/.changes/fix-nsis-per-machine-installer-elevation.md
+++ b/.changes/fix-nsis-per-machine-installer-elevation.md
@@ -1,0 +1,5 @@
+---
+tauri-bundler: patch:bug
+---
+
+Fix NSIS per-machine installer not requesting elevation when run by non-admin users.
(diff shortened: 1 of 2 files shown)


FabianLars:
thanks for the pr :)

for the reviewers someone else(?) confirmed that this works as well: https://discord.com/channels/616186924390023171/1381811624624652331

Legend-Master (approved):
Makes sense to me, nice catch

They correspond to `requireAdministrator` and `highestAvailable` in case anyone also wants to take a look

> https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests#trustinfo

Issue #8493, opened 2023-12-29
[bug] Error Permission denied (os error 13)


guoapeng:
### Describe the bug

I've just upgraded my tauri project to 2.0 alpha version and encountered the issue "Error Permission denied (os error 13)".
I have a gitlab pipeline job  to create Android apk and aab production bundles  on a host using gitlab-runner account.  I sometime encounter the issue above.

the situation is that I have a development environment setup on the same host with gitlab runner using another account different from gitlab-runner account to build a testing application. it seems like there is  conflict between the two builds 
I know that I can setup different hosts for different purpose to solve the problem. 
but I would like to report the issue to tauri team that the error message is too general to have less useful infor for troubleshooting and not easy to turn on verbose log and even on same host the build process should not have conflict between each other.

### Reproduction

execute `npm run tauri android build ` on same host with different accounts and source code on different locations, the issue will be reproduced.

### Full `tauri info` output

```text
$npm run tauri -- info

> demo-web@1.0.0 tauri
> tauri info


[✘] Environment
    - OS: Fedora 39.0.0 X64
    ✘ webkit2gtk-4.1: not installed
      Visit https://tauri.app/v1/guides/getting-started/prerequisites to learn more about tauri prerequisites
    ✔ rsvg2: 2.57.1
    ✔ rustc: 1.74.1 (a28077b28 2023-12-04)
    ✔ cargo: 1.74.1 (ecb9851af 2023-10-18)
    ✔ rustup: 1.26.0 (5af9b9484 2023-04-05)
    ✔ Rust toolchain: stable-x86_64-unknown-linux-gnu (default)
    - node: 18.13.0
    - npm: 8.19.3

[-] Packages
    - tauri [RUST]: 2.0.0-alpha.20
    - tauri-build [RUST]: 2.0.0-alpha.13
    - wry [RUST]: 0.35.1
    - tao [RUST]: 0.24.0
    - tauri-cli [RUST]: 2.0.0-alpha.20
    - @tauri-apps/api [NPM]: 2.0.0-alpha.13
    - @tauri-apps/cli [NPM]: 2.0.0-alpha.20

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../dist/demo-web/browser
    - devPath: http://localhost:1420/
    - framework: Angular
    - bundler: Webpack
```

### Stack trace

```text
> demo-web@1.0.0 tauri
> tauri android build
        Info detected host target triple "x86_64-unknown-linux-gnu"
[...]

Issue #7460, opened 2023-07-19
[bug] OS Error 13: Permission Denied


RimuruOtter:
### Describe the bug

I am trying to use the sidecar for building in binaries to the app but get this error everytime
```bash
thread 'tokio-runtime-worker' panicked at 'Failed to spawn command: Io(Os { code: 13, kind: PermissionDenied, message: "Permission denied" })', src/main.rs:68:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
I saw a previous post about this same problem but it fixed when the person upgraded their dependencies but mine are already up to date as of "1.4".

If any more details are needed let me know please. I tried looking at other examples and nothing has worked. Also tauri permissions are fully allowed so that shouldn't be the issue.

### Reproduction

tauri.conf.json
```json
{
	"build": {
		"beforeDevCommand": "yarn dev",
		"beforeBuildCommand": "yarn build",
		"devPath": "http://localhost:1420",
		"distDir": "../dist",
		"withGlobalTauri": true
	},
	"package": {
		"productName": "Neko Launcher",
		"version": "0.0.0"
	},
	"tauri": {
		"allowlist": {
            "all": false,
			"shell": {
				"sidecar": true,
                "scope": [
                    { "name": "bin/auth", "sidecar": true }
                ]
			},
			"window": {
				"close": true,
				"startDragging": true
			},
			"http": {
				"scope": ["https://api.minecraftservices.com/*"],
				"all": true,
				"request": true
			},
            "fs": {
                "all": true
            },
            "path": {
                "all": true
            },
            "os": {
                "all":true
            },
            "notification": {
                "all": true
            }
		},
		"bundle": {
			"active": true,
			"targets": "all",
			"identifier": "me.media.dev",
			"icon": [
				"icons/32x32.png",
				"icons/128x128.png",
				"icons/128x128@2x.png",
				"icons/icon.icns",
				"icons/icon.ico"
			], 
            "externalBin": ["bin/auth"],
            "deb": {
                "depends": []
            }
		},
		"security": {
			"csp": "default-src 'self'"
		},
		"windows": [
			{
				"fullscreen": false,
				"resizable": true,
				"decorations": false,
				"transparent": true,
				"title": "Neko Launcher",
				"width": 1100,
				"height": 700
			}
		],
		"macOSPrivateApi": true
	}
}
```

Here the sidecar is being called in a tauri command:
```rust
#[tauri::command]
[...]

PR #13793, opened 2025-07-09
fix(macos): check if path is file or dir or neither (e.g. /dev/null) and improve error messages

ryanseys:
Trying to delete (rm) existing paths are assumed to be files during build, and fails due to not checking if it is a directory or a file first. Directories need to be removed recursively.

In trying to debug some weird permission denied errors (#7460, #8493), I came across this code and figured I would try to add some additional error messages to help with debugging.

Prior to this PR, we assumed that if `path.is_file()` is `false`, then `path.is_dir()` is `true`, but in some cases the path is valid, but is not a file or directory e.g. (`/dev/null`)

Also added some tests.


FabianLars (approved):
i guess this doesn't hurt, thanks for the pr :)

ryanseys:
thanks for the review + merge!

PR #13808, opened 2025-07-11
add #[used] to __TAURI_BUNDLE_TYPE so that it's not stripped 

kandrelczyk:
Fixing build issue caused by https://github.com/tauri-apps/tauri/pull/13209

### DIFF
--- a/crates/tauri-utils/src/platform.rs
+++ b/crates/tauri-utils/src/platform.rs
@@ -347,6 +347,7 @@ fn resource_dir_from<P: AsRef<std::path::Path>>(
 
 // Variable holding the type of bundle the executable is stored in. This is modified by binary
 // patching during build
+#[used]
 #[no_mangle]
 #[cfg_attr(not(target_vendor = "apple"), link_section = ".taubndl")]
 #[cfg_attr(target_vendor = "apple", link_section = "__DATA,taubndl")]


Legend-Master (approved):

Legend-Master:
Thanks for the quick fix ❤️

Legend-Master:
Hmm, it still seems to be failing 😂 https://github.com/tauri-apps/tauri/actions/runs/16214825076/job/45781764969

kandrelczyk:
That's really strange because it's just bundling the app here, right? And the app uses version of tauri from the repo? I guess I will have to set up javascript env on my computer  and run the test locally.

Legend-Master:
> That's really strange because it's just bundling the app here, right? And the app uses version of tauri from the repo?

Yeah, I think so, it runs the tauri cli to init the project (`tauri init`) and then `tauri build` in it

I don't quite understand why is the error either right now

kandrelczyk:
Ok, I figured out it's because of rust version. It's working on nightly but failing on stable. #[used] works the same between rust versions and I confirmed that __TAURI_BUNDLE_TYPE is in the binary so it's looks like the patching code is not finding it for some reason. I will have to review the patching code again.

kandrelczyk:
The variable was stripped after all. Looks like stable rust does something differently, no idea why. Making the static var mutable helped. The test now pass with rust 1.88.0. Hopefully it will finally work. 
PR here: https://github.com/tauri-apps/tauri/pull/13812

PR #13809, opened 2025-07-11
feat: reduce Debug format size for binary buffers

Beanow:
This makes the debug output of large Tauri structs more manageable.
I primarily looked at `AppHandle` and the binary buffer fields that it contains.

Try it out by logging the AppHandle in `examples/api/src-tauri/src/lib.rs`:
```rust
log::info!("Running app {app_handle:?}");
```

The icons, images and (compressed) assets can cause some massive debug outputs because their contents are arbitrarily large and the formatting of bytes as decimals (`[0, 0, 0, 0, ...]`) is not compact either.

Generally I'm formatting them here as `[u8; {len}]` similar to the type definition for a fixed length array.

### Alternatives

I used some temporary structs which are only used during the formatting itself. Instead we could also create permanent structs to override the Debug behavior.

Benefit would be that in some cases we can go back to deriving Debug for the parent struct. Downside is we'd need to implement more than just Debug to retain all the properties we care about, like Clone and accessing the contents.

As alternative format to the `[u8; {len}]` we could also use a more compact notation that preserves the contents like base64. Benefit would be you can still debug the contents of these buffers. Downside is it would still be arbitrarily large, just more dense.

### DIFF
--- a/crates/tauri/src/lib.rs
+++ b/crates/tauri/src/lib.rs
@@ -397,12 +397,32 @@ pub struct Context<R: Runtime> {
   pub(crate) plugin_global_api_scripts: Option<&'static [&'static str]>,
 }
 
+/// Temporary struct that overrides the Debug formatting for the `app_icon` field.
+///
+/// It reduces the output size compared to the default, as that would format the binary
+/// data as a slice of numbers `[65, 66, 67]`. This instead shows the length of the Vec.
+///
+/// For example: `Some([u8; 493])`
+pub(crate) struct DebugAppIcon<'a>(&'a Option<Vec<u8>>);
+
+impl std::fmt::Debug for DebugAppIcon<'_> {
+  fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+    match self.0 {
+      Option::None => f.write_str("None"),
+      Option::Some(icon) => f
+        .debug_tuple("Some")
+        .field(&format_args!("[u8; {}]", icon.len()))
+        .finish(),
+    }
+  }
+}
+
 impl<R: Runtime> fmt::Debug for Context<R> {
   fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
     let mut d = f.debug_struct("Context");
     d.field("config", &self.config)
       .field("default_window_icon", &self.default_window_icon)
-      .field("app_icon", &self.app_icon)
+      .field("app_icon", &DebugAppIcon(&self.app_icon))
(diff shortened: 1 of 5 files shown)


crates/tauri/src/lib.rs
```diff
@@ -397,12 +397,32 @@ pub struct Context<R: Runtime> {
...
 impl<R: Runtime> fmt::Debug for Context<R> {
   fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
     let mut d = f.debug_struct("Context");
     d.field("config", &self.config)
       .field("default_window_icon", &self.default_window_icon)
-      .field("app_icon", &self.app_icon)
+      .field("app_icon", &DebugAppIcon(&self.app_icon))
       .field("package_info", &self.package_info)
       .field("pattern", &self.pattern)
       .field("plugin_global_api_scripts", &self.plugin_global_api_scripts);
```
Beanow, line 428: Sidenote: these are missing several fields that exist on the struct.
I'm not sure whether that's intentional or whether the Debug implementation just is out of date.

crates/tauri/src/manager/mod.rs
```diff
@@ -233,7 +233,7 @@ impl<R: Runtime> fmt::Debug for AppManager<R> {
       .field("plugins", &self.plugins)
       .field("state", &self.state)
       .field("config", &self.config)
-      .field("app_icon", &self.app_icon)
+      .field("app_icon", &DebugAppIcon(&self.app_icon))
       .field("package_info", &self.package_info)
       .field("pattern", &self.pattern);
```
Beanow, line 238: Likewise missing fields.

crates/tauri/src/lib.rs
```diff
@@ -397,12 +397,32 @@ pub struct Context<R: Runtime> {
   pub(crate) plugin_global_api_scripts: Option<&'static [&'static str]>,
 }
 
+/// Temporary struct that overrides the Debug formatting for the `app_icon` field.
+///
+/// It reduces the output size compared to the default, as that would format the binary
+/// data as a slice of numbers `[65, 66, 67]`. This instead shows the length of the Vec.
+///
+/// For example: `Some([u8; 493])`
+pub(crate) struct DebugAppIcon<'a>(&'a Option<Vec<u8>>);
```
Beanow, line 406: Wasn't 100% sure on this one. Because it seems only macos uses this `app_icon`.
On Linux it was always None for me.

lucasfernog (approved):

PR #13811, opened 2025-07-11
feat(cli): allow runner configuration to be an object with cwd and args

### LINKED ISSUE #13779: [feat] Allow "runner" in tauri build config to be an object
### Describe the problem

It is possible to set the "runner" to replace default cargo in the tauri.conf.json

```
{
    [...]
    "build" : {
        [...],
        "runner" : "my_cargo"
    }
}
```

But this runner can only be a string which should be a binary to launch.
It would be nice if the runner configuration could offer more flexibility, like setting the working directory or options that could be passed to it.

Allow "runner" to be either a string or an object as follows:

```
"runner" : { "cwd" :  "<dir>" , "cmd" : "my_runner", "args" : "--quiet" }
```

### Alternatives considered

_No response_

### Additional context

_No response_

mhbagheri-99:
This PR implements the feature requested in #13779 to allow the "runner" configuration in `tauri.conf.json` to be either a string (existing behavior) or an object with additional properties.

## Changes

- Enhanced the runner configuration to support an object format with the following properties:
  - `cmd`: The command to execute (equivalent to the previous string format)
  - `cwd`: Optional working directory to run the command from
  - `args`: Optional array of arguments to pass to the runner

## Example Usage

The runner can now be configured as an object:

```json
{
  "build": {
    "runner": {
      "cmd": "my_custom_cargo",
      "cwd": "/path/to/project",
      "args": ["--quiet", "--offline"]
    }
  }
}
```

Or continue using the existing string format for backwards compatibility:

```json
{
  "build": {
    "runner": "cargo"
  }
}
```

## Backwards Compatibility

This change maintains full backwards compatibility - existing configurations using a string for the runner will continue to work as before.

Closes #13779

---

### DIFF
--- a/crates/tauri-utils/src/config.rs
+++ b/crates/tauri-utils/src/config.rs
@@ -2770,7 +2770,7 @@ pub enum HookCommand {
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 pub struct BuildConfig {
   /// The binary used to build and run the application.
-  pub runner: Option<String>,
+  pub runner: Option<RunnerConfig>,
   /// The URL to load in development.
   ///
   /// This is usually an URL to a dev server, which serves your application assets with hot-reload and HMR.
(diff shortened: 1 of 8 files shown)


crates/tauri-utils/src/config.rs
```diff
@@ -3824,4 +3847,283 @@ mod test {
...
+    assert!(object_json.contains("\"args\":null") || !object_json.contains("args"));
+  }
+}
+
+/// The runner configuration.
+#[skip_serializing_none]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[cfg_attr(feature = "schema", derive(JsonSchema))]
+#[serde(untagged)]
+pub enum RunnerConfig {
```
Legend-Master: Should we move these to above, like close to `HookCommand`?

Legend-Master (dismissed):
Thanks! Just a nitpick

crates/tauri-utils/src/config.rs
mhbagheri-99: @Legend-Master moved the implementation to make more sense according to your comment 🙂

Legend-Master (approved):

PR #13812, opened 2025-07-11
make static bundle type var mutable

kandrelczyk:
Related to https://github.com/tauri-apps/tauri/pull/13808#issuecomment-3061308622

Making the static variable mutable so that it's not stripped on stable rust. 

The no_mangle and link_section are made unsafe to comply with 2024 edition changes: https://doc.rust-lang.org/reference/abi.html

### DIFF
--- a/crates/tauri-utils/src/platform.rs
+++ b/crates/tauri-utils/src/platform.rs
@@ -348,25 +348,27 @@ fn resource_dir_from<P: AsRef<std::path::Path>>(
 // Variable holding the type of bundle the executable is stored in. This is modified by binary
 // patching during build
 #[used]
-#[no_mangle]
-#[cfg_attr(not(target_vendor = "apple"), link_section = ".taubndl")]
-#[cfg_attr(target_vendor = "apple", link_section = "__DATA,taubndl")]
-static __TAURI_BUNDLE_TYPE: &str = "UNK";
+#[unsafe(no_mangle)]
+#[cfg_attr(not(target_vendor = "apple"), unsafe(link_section = ".taubndl"))]
+#[cfg_attr(target_vendor = "apple", unsafe(link_section = "__DATA,taubndl"))]
+static mut __TAURI_BUNDLE_TYPE: &str = "UNK";
 
 /// Get the type of the bundle current binary is packaged in.
 /// If the bundle type is unknown, it returns [`Option::None`].
 pub fn bundle_type() -> Option<BundleType> {
-  match __TAURI_BUNDLE_TYPE {
-    "DEB" => Some(BundleType::Deb),
-    "RPM" => Some(BundleType::Rpm),
-    "APP" => Some(BundleType::AppImage),
-    "MSI" => Some(BundleType::Msi),
-    "NSS" => Some(BundleType::Nsis),
-    _ => {
-      if cfg!(target_os = "macos") {
-        Some(BundleType::App)
-      } else {
-        None
+  unsafe {
+    match __TAURI_BUNDLE_TYPE {
+      "DEB" => Some(BundleType::Deb),
+      "RPM" => Some(BundleType::Rpm),
+      "APP" => Some(BundleType::AppImage),
+      "MSI" => Some(BundleType::Msi),
+      "NSS" => Some(BundleType::Nsis),
+      _ => {
+        if cfg!(target_os = "macos") {
+          Some(BundleType::App)
+        } else {
+          None
+        }
       }
     }
   }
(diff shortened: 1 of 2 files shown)


Legend-Master:
I know it's a bit later, but I feel like the whole binary patching thing is a bit awkward to me, would it make sense for the bundler to bundle a file that contains the bundler type? (sorry about not mentioning this at the start)

The docs about `#[used]` only guarantees that it will not be tripped by rust compiler, but it doesn't ensure if it will stay after linking

> The no_mangle and link_section are made unsafe to comply with 2024 edition changes

It's added in 1.82 so we can't use it yet 😂

kandrelczyk:
I checked and with __TAURI_BUNDLE_TYPE made mutable it works even without #[used] so I wouldn't worry about the linker. I think the issue was some other optimization. Maybe it was just substituting the variable with a string directly? 

Obviously I would prefer no to start from scratch  with this but also don't see how bundling a file would work.  At runtime we would have to find the file and it's location would depend on OS/distro so we would have to know bundle type to figure out bundle type. Patching the binary doesn't introduce any new dependencies at runtime and works the same on all OSes/distros. The patching code is also pretty simple. The only issue was the optimization done by compiler  but if it's working now on stable and nightly I would continue with this solution.

Legend-Master:
> don't see how bundling a file would work. At runtime we would have to find the file and it's location would depend on OS/distro so we would have to know bundle type to figure out bundle type.

Technically, you can check for each location which will not be many, I do agree about the dependency thing though, not a fan of an extra file either

> The patching code is also pretty simple. The only issue was the optimization done by compiler

This is the thing, I don't know if the generated code / linking is stable enough for us to rely on (in cases like it did without `mut`)

cc @lucasfernog @FabianLars any thoughts?

crates/tauri-utils/src/platform.rs
```diff
@@ -351,22 +351,24 @@ fn resource_dir_from<P: AsRef<std::path::Path>>(
 #[no_mangle]
 #[cfg_attr(not(target_vendor = "apple"), link_section = ".taubndl")]
 #[cfg_attr(target_vendor = "apple", link_section = "__DATA,taubndl")]
-static __TAURI_BUNDLE_TYPE: &str = "UNK";
+static mut __TAURI_BUNDLE_TYPE: &str = "UNK";
```
Legend-Master, line 356: Maybe adding a comment about why it's a `mut`?

Legend-Master (dismissed):
Thanks for the effort

FabianLars:
> cc FabianLars any thoughts

I'm busy over the weekend so I'll either get back to you in the evening(s) or on Monday.

That said, I was never a big fan of the binary patching but reverting it now doesn't sound nice either. (I know the ping wasn't about this)

Legend-Master (approved):

kandrelczyk:
Another alternative mentioned was https://github.com/denoland/sui

I guess we could add it to the bundler and updater-plugin. It would be easy to switch in updater. I would need some time to figure out how to use it in the bundler but if we feel it's a better solution I can do it. 

That being said if we don't see any clear risks with binary patching I would stick with it.

Legend-Master:
> Another alternative mentioned was https://github.com/denoland/sui

Not a big fan of pull in `sui` since it pulls in `image` crate which is quite big, I think that's also the reason why you didn't go for it either in the initial PR

> That being said if we don't see any clear risks with binary patching I would stick with it.

Yeah, I'm just not sure since we hit that error and I'm a bit concerned if it's ok to rely on this behavior, since it might be different for different linkers

kandrelczyk:
> Not a big fan of pull in sui since it pulls in image crate which is quite big, I think that's also the reason why you didn't go for it either in the initial PR

Yes but with sui we could move it all into updater-plugin so minimal app that don't use wouldn't increase in size. Still better to keep the app small even with updater...

> Yeah, I'm just not sure since we hit that error and I'm a bit concerned if it's ok to rely on this behavior, since it might be different for different linkers

I see two things we could do:

1. turn the error about missing __TAURI_BUNDLE_TYPE into a warning. If the bundler fails to patch binary updater-plugin would simply fallback to previous behavior. The issue here is that if people miss the warning they will report issues about deb/rpm updates not working. 
2. Let people skip the binary patching. Add a flag like `cargo tauri build --no-bundle-type` so that people hitting eventual linker issues can still build the app. Again, it would simply fallback to old updater behavior.

Legend-Master:
> turn the error about missing __TAURI_BUNDLE_TYPE into a warning. If the bundler fails to patch binary updater-plugin would simply fallback to previous behavior. The issue here is that if people miss the warning they will report issues about deb/rpm updates not working.

Was thinking about something similar, I guess we could make it a warning and let the user report the problem if they ever hit it

Can't think of any better options right now (except for the bundling a file approach)

kandrelczyk:
Ok, let see what others think before I make any changes.

FabianLars:
option 1 sounds reasonable for the time being. wanted to keep the old behavior in v2 anyway (as a fallback).

PR #13822, opened 2025-07-14
feat: expose some window props from runtime-wry

clearlysid:
I'm working on [tauri-plugin-egui](https://github.com/clearlysid/tauri-plugin-egui), which needs access to a few props from `tauri-runtime-wry` — this PR just adds a `pub` keyword for them.

No change to the broader `tauri` crate's API surface and no breaking/regressive changes. Would love to get this merged in order to allow publishing my plugin on crates.io. Thank you! 🙌


FabianLars:
one of your commits is not signed which is a requirement for me to be able to merge this pr.

also a changefile would be appreciated https://github.com/tauri-apps/tauri/tree/dev/.changes (just for tauri-runtime-wry)

clearlysid:
Thank you! Signed and added 🫡

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -2388,6 +2388,12 @@ pub struct WindowWrapper {
   focused_webview: Arc<Mutex<Option<String>>>,
 }
 
+impl WindowWrapper {
+  pub fn label(&self) -> String {
+    self.label.clone()
+  }
```
Legend-Master:
```suggestion
  pub fn label(&self) -> &str {
    &self.label
  }
```

Don't really care much, but I think it should probably be something like this

Legend-Master (dismissed):

crates/tauri-runtime-wry/src/lib.rs
clearlysid: Agreed. Will update my plugin's downstream dependencies based on this and commit.

Legend-Master (approved):

PR #13358, opened 2025-05-01
feat: add `--root-certificate-path` option for mobile dev

lucasfernog:
lets you use a HTTPS development server

example usage:
```
cargo install tauri-cli --git https://github.com/tauri-apps/tauri --branch feat/mobile-dev-cert
cargo tauri android dev --open --root-certificate-path "/Users/lucas/Library/Application Support/mkcert/rootCA.pem" --features tauri/rustls-tls
```

PR #13824, opened 2025-07-14
feat(bundler/cli): Add feature flag to use system certificates

FabianLars:
opening this so i don't forget about it (i doubt that can happen since lately i get at least one report per week about this).

one question though, should we enable this by default or put it behind an env var or cli flag?


FabianLars:
https://github.com/tauri-apps/tauri/pull/13358 is also veryyy loosely related

Legend-Master:
I'm kinda new to this, what's the reason for us to use `webpki-roots` + `rustls` instead of `native-tls` + `rustls-platform-verifier` here originally?

FabianLars:
native-tls kinda sucks (iirc for us it was mostly about cross comp) - i think the ecosystem is also moving away from it (or at least considering it) eg https://github.com/seanmonstar/reqwest/issues/2723

and platform verifier just never was on our minds before / back then 🤷

FabianLars:
i also don't mind adding it directly but since i don't have any experience with it i thought that's a opt-out would be nice. On the other hand if it were to cause issues we'd release a hotfix that disables it for everyone so idk if the flag really makes sense.

Legend-Master:
I see, thanks for the explanation and the link 

So from my understanding, outside of the cross compilation staffs (we don't cross compile our cli right?), the OS implementation of TLS might be flawed from the discussions under 
https://github.com/seanmonstar/reqwest/issues/2025#issuecomment-2010674997 ?

I feel like the flag probably doesn't help since that only works if they install the cli through cargo, maybe we could just do it and see if it actually breaks anything

`rustls-platform-verifier` + `rustls` seems to be the best option here for an app, the only downside seems to be some Linux distros don't manage the system certificates well?

> https://github.com/rust-lang/rustup/issues/3400#issuecomment-1683762630

FabianLars:
> the OS implementation of TLS might be flawed from the discussions under
https://github.com/seanmonstar/reqwest/issues/2025#issuecomment-2010674997 ?

guess so? idk, i didn't look that much into it and it was quite some time ago. it just didn't work (well) for us 🤷 

> I feel like the flag probably doesn't help since that only works if they install the cli through cargo, maybe we could just do it and see if it actually breaks anything

i meant to enable it by default for the js cli as well. either way i guess i just drop the flag?

> the only downside seems to be some Linux distros don't manage the system certificates well?

that sucks but i can't say i care 🫠 since this is only concerning dev machines i don't think it matters, what would you like to do?

Legend-Master:
> i meant to enable it by default for the js cli as well. either way i guess i just drop the flag?

To be fair, since we already have a few other related flags, I don't mind having this feature flag

> that sucks but i can't say i care 🫠 since this is only concerning dev machines i don't think it matters, what would you like to do?

I would say yeah, go for `rustls-platform-verifier` + `rustls`, I believe rustup also uses it so if they would hit the problem installing rust first before hitting this one 🙃

crates/tauri-bundler/Cargo.toml
```diff
@@ -78,3 +78,4 @@ default = ["rustls"]
 native-tls = ["ureq/native-tls"]
 native-tls-vendored = ["native-tls", "native-tls/vendored"]
 rustls = ["ureq/rustls"]
+platform-certs = ["ureq/platform-verifier"]
```
Legend-Master, line 81: We could also sync `default = ["rustls", "platform-certs"]` here?

Legend-Master (dismissed):

crates/tauri-bundler/Cargo.toml
FabianLars: i was wary of breaking tauri-bundler users but according to crates.io that's just dioxus so i wouldn't mind doing it.

Legend-Master (approved):

PR #13825, opened 2025-07-14
failed binary patch warning

kandrelczyk:
As agree in https://github.com/tauri-apps/tauri/pull/13812#issuecomment-3065893529 changing error to warning here.

Keep in mind that tests I added to updater-plugin will fail if the binary can't be patched correctly so there's still some verification for this feature in the pipeline.

### DIFF
--- a/crates/tauri-bundler/src/bundle.rs
+++ b/crates/tauri-bundler/src/bundle.rs
@@ -127,7 +127,7 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
     }
 
     if let Err(e) = patch_binary(&settings.binary_path(main_binary), package_type) {
-      log::warn!("Failed to add package type information to the binary. Updater plugin may not be able to update this package: {}", e);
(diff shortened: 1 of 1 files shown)


crates/tauri-bundler/src/bundle.rs
```diff
@@ -126,7 +126,9 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
       continue;
     }
 
-    patch_binary(&settings.binary_path(main_binary), package_type)?;
+    if let Err(e) = patch_binary(&settings.binary_path(main_binary), package_type) {
+      log::warn!("Failed to add package type information to the binary. Updater plugin may not be able to update this package: {}", e);
```
```suggestion
      log::warn!("Failed to add bundle type to the binary: {e}. Updater plugin may not be able to update this package. This shouldn't normally happen, please report it to https://github.com/tauri-apps/tauri/issues");
```

We should let people know that this is probably not their fault if it failed and let them report the problem

Legend-Master (dismissed):
Thanks, just some suggestions around the wording

Legend-Master (approved):

PR #13837, opened 2025-07-16
feat(tauri): add `plugin_boxed` methods

WSH032:
## Describe the problem

*Because this PR is small enough, I didn’t open a feature request issue in advance.*

I am implementing the feature to add tauri plugins from Python for pytauri: pytauri/pytauri#220.
In pyo3/pytauri, we cannot use generics, so I [store these `TauriPlugin` as `Box<dyn Plugin>`](https://github.com/pytauri/pytauri/blob/aa995097a883d8a6bd67c2b2c50c180a5dafca8f/crates/pytauri-core/src/ext_mod_impl/plugin.rs#L7-L13).

However, `Box<dyn Plugin>` itself does not implement the `Plugin` trait, so I cannot use them in `Builder::plugin` and `AppHandle::plugin`.

## Describe the solution you'd like

Therefore, I would like to add `plugin_dyn` methods to accept `Box<dyn Plugin>` instead of `impl Plugin<R>`.

## Alternatives considered

Alternatively, we could `impl Plugin<R> for Box<dyn Plugin<R>>`. But I don't really like this approach, because it would result in [`Box<Box<dyn Plugin<R>>>`](https://github.com/tauri-apps/tauri/blob/bda8304107da7ca60caaba5674faa793491898c6/crates/tauri/src/app.rs#L489).

But if tauri doesn't want to add a new API, I can also accept this implementation.

### DIFF
--- a/crates/tauri/src/app.rs
+++ b/crates/tauri/src/app.rs
@@ -484,10 +484,16 @@ impl<R: Runtime> AppHandle<R> {
   ///     Ok(())
   ///   });
   /// ```
-  #[cfg_attr(feature = "tracing", tracing::instrument(name = "app::plugin::register", skip(plugin), fields(name = plugin.name())))]
   pub fn plugin<P: Plugin<R> + 'static>(&self, plugin: P) -> crate::Result<()> {
-    let mut plugin = Box::new(plugin) as Box<dyn Plugin<R>>;
+    self.plugin_dyn(Box::new(plugin))
+  }
 
+  /// Adds a Tauri application plugin.
+  ///
+  /// This method is similar to [`Self::plugin`],
+  /// but accepts a boxed trait object instead of a generic type.
+  #[cfg_attr(feature = "tracing", tracing::instrument(name = "app::plugin::register", skip(plugin), fields(name = plugin.name())))]
+  pub fn plugin_dyn(&self, mut plugin: Box<dyn Plugin<R>>) -> crate::Result<()> {
     let mut store = self.manager().plugins.lock().unwrap();
     store.initialize(&mut plugin, self, &self.config().plugins)?;
     store.register(plugin);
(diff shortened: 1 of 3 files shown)


crates/tauri/src/app.rs
```diff
@@ -484,10 +484,16 @@ impl<R: Runtime> AppHandle<R> {
...
-    let mut plugin = Box::new(plugin) as Box<dyn Plugin<R>>;
+    self.plugin_dyn(Box::new(plugin))
+  }
 
+  /// Adds a Tauri application plugin.
+  ///
+  /// This method is similar to [`Self::plugin`],
+  /// but accepts a boxed trait object instead of a generic type.
+  #[cfg_attr(feature = "tracing", tracing::instrument(name = "app::plugin::register", skip(plugin), fields(name = plugin.name())))]
+  pub fn plugin_dyn(&self, mut plugin: Box<dyn Plugin<R>>) -> crate::Result<()> {
```
Legend-Master: Maybe `plugin_boxed` or `plugin_boxed_dyn`? Since just `_dyn` is not clear about that part

Legend-Master (dismissed):
Don't mind adding it personally

crates/tauri/src/app.rs
WSH032: Done. I chose `plugin_boxed` because `plugin_boxed_dyn` is a bit too long for me. I also added `#[inline]`.

*My company computer does not have the key for signing. Can anyone help me sign it, or I can sign it after work.*

crates/tauri/src/app.rs
```diff
@@ -484,10 +484,17 @@ impl<R: Runtime> AppHandle<R> {
   ///     Ok(())
   ///   });
   /// ```
-  #[cfg_attr(feature = "tracing", tracing::instrument(name = "app::plugin::register", skip(plugin), fields(name = plugin.name())))]
+  #[inline]
```
Legend-Master:
```suggestion
```

Generic functions are implicitly inlinable, we don't need to add this

Legend-Master (dismissed):
Will wait for you to sign the commits when you get back 🙃

crates/tauri/src/app.rs
WSH032: Done.

Legend-Master (approved):

PR #13848, opened 2025-07-17
Add `WebviewBuilder::limit_navigations_to_app_bound_domains`

### LINKED ISSUE #13847: [feat] Add limit_navigations_to_app_bound_domains into WebviewBuilder and WebviewWindowBuilder
### Describe the problem

Extension of [wry#1587](https://github.com/tauri-apps/wry/issues/1587). On iOS Service Workers are disabled in WKWebView by default.

This can be fixed by setting App-Bound Domains for iOS in `Info.plist` in addition to running
`webViewConfiguration.limitsNavigationsToAppBoundDomains = YES;` on the WKWebViewConfiguration object. Then all WKWebViews used in the app are able to use Service Workers [[source](https://stackoverflow.com/questions/49673399/service-workers-unavailable-in-wkwebview-in-ios-11-3/64155509#64155509)].

It additionally further locks down the WKWebView to only allow requests to resources from domains which are explicitly whitelisted. That said, I mainly want the feature so I can use a Service Worker in my Tauri App.

I'd like to extend the API in Wry as of (https://github.com/tauri-apps/wry/pull/1588) into the WebviewBuilder and the WebviewWindowBuilder structs in Tauri.

### Alternatives considered

I considered using a Tauri plugin; unfortunately the reference to the WKWebviewConfiguration object is not made available to a Tauri plugin, at least as far as I could tell.

### Additional context

App-Bound Domains Blog Post: https://webkit.org/blog/10882/app-bound-domains/
StackOverflow issue about enabling Service Workers in WKWebViews: https://stackoverflow.com/questions/49673399/service-workers-unavailable-in-wkwebview-in-ios-11-3/64155509#64155509

zphrs:
closes #13847.

Draft because this can only be merged once Wry's version gets bumped to 0.52.1 or greater. Blocked by [Wry PR #1578](https://github.com/tauri-apps/wry/pull/1578#issue-3192964463).


zphrs:
@FabianLars sorry I just got to rebasing, this should be good to merge in. LMK if there's any changes I should make.

crates/tauri-runtime/src/webview.rs
```diff
@@ -760,6 +764,39 @@ impl WebviewAttributes {
...
+  /// Default is false.
+  ///
+  /// See https://webkit.org/blog/10882/app-bound-domains/ and
+  /// https://developer.apple.com/documentation/webkit/wkwebviewconfiguration/limitsnavigationstoappbounddomains
+  ///
+  /// ## Platform-specific
+  ///
+  /// - **Linux / Windows / Android / MacOS:** Unsupported.
+  #[must_use]
+  #[cfg(target_os = "ios")]
```
FabianLars: can we not put these setters behind a cfg flag like traffic_light_position for example?
Makes it easier to discover and prevents IDEs from tripping up.

If users want they can put it behind a cfg themselves.

I know our code base is a bit inconsistent about this but it's on our todo list i hope

.changes/app-bound-domains.md
```diff
@@ -0,0 +1,7 @@
+---
+"tauri": patch:feat
+"tauri-runtime": patch:feat
+"tauri-runtime-wry": patch:feat
```
FabianLars:
```suggestion
"tauri": minor:feat
"tauri-runtime": minor:feat
"tauri-runtime-wry": minor:feat
```

FabianLars:
no worries and thanks for the PR :)

crates/tauri-runtime/src/webview.rs
zphrs: Sure, I can make them a no-op when the target isn't iOS.

zphrs:
@FabianLars are there any further changes you'd like me to make before this gets merged in?

FabianLars:
do you think this should be configurable in tauri.conf.json as well?

zphrs:
> do you think this should be configurable in tauri.conf.json as well?

Sure! I wired this up.

zphrs:
@Legend-Master just checking in since I saw you added the 2.12 label recently. Should I rebase my changes again?

Legend-Master:
Yes, I personally prefer a merge so we don't lose the review history from a rebase + force push, but do whatever you prefer

zphrs:
> Yes, I personally prefer a merge so we don't lose the review history from a rebase + force push, but do whatever you prefer

done!

Legend-Master:
Apart from the missing `,`, looks like this is also available on macOS right?

> https://developer.apple.com/documentation/webkit/wkwebviewconfiguration/limitsnavigationstoappbounddomains

zphrs:
> Apart from the missing `,`, looks like this is also available on macOS right?
> 
> > https://developer.apple.com/documentation/webkit/wkwebviewconfiguration/limitsnavigationstoappbounddomains

It seems that limitAppBoundDomains can be set in MacOS as well but based on [searching through the WebKit source code](https://github.com/search?q=repo%3AWebKit%2FWebKit%20limitsNavigationsToAppBoundDomains&type=code), the sole WebKit issue mentioning limitAppBoundDomains and MacOS (https://bugs.webkit.org/show_bug.cgi?id=216628), and the [blog post on App-Bound Domains](https://webkit.org/blog/10882/app-bound-domains/) the flag doesn't change any behavior on MacOS; in fact the internal getter for whether it is set is gated behind a `#if PLATFORM(IOS_FAMILY)` flag [[source code](https://github.com/WebKit/WebKit/blob/main/Source/WebKit/UIProcess/API/Cocoa/WKWebViewConfiguration.mm#L606-L610)]. I also just messaged the WebKit #help slack channel and am waiting for a response.

If we do want to have it set the flag on MacOS as well then I'd have to go back and edit Wry as well and I think it's unnecessary to have it do anything on MacOS because the flag only seems to be checked on iOS.

Legend-Master:
Thanks for the detailed digging! We can merge this first and if it ever gets supported on macOS in the future we can always add that then.

.changes/app-bound-domains.md
```diff
@@ -0,0 +1,7 @@
+---
+"tauri": minor:feat
+"tauri-runtime": minor:feat
+"tauri-runtime-wry": minor:feat
```
Legend-Master, line 4:
```suggestion
"tauri-runtime-wry": minor:feat
"tauri-utils": minor:feat
```

crates/tauri-utils/src/config.rs
```diff
@@ -2276,6 +2276,36 @@ pub struct WindowConfig {
...
+  /// You should also add any additional domains which your app requests assets from.
+  /// Assets served through custom protocols like Tauri's IPC are added to the
+  /// list automatically. Available on iOS only.
+  ///
+  /// Default is false.
+  ///
+  /// See https://webkit.org/blog/10882/app-bound-domains/ and
+  /// https://developer.apple.com/documentation/webkit/wkwebviewconfiguration/limitsnavigationstoappbounddomains
+  ///
+  /// ## Platform-specific
```
Legend-Master, line 2331: Add an empty line between `##` and the following `-`, also the other ones

Legend-Master (dismissed):
Nice work, thanks!

Just a new nitpicks and we can get this merged.

crates/tauri-runtime/src/webview.rs
```diff
@@ -538,6 +543,35 @@ impl WebviewAttributes {
...
+      /// list automatically. Available on iOS only.
+      ///
+      /// Default is false.
+      ///
+      /// See https://webkit.org/blog/10882/app-bound-domains/ and
+      /// https://developer.apple.com/documentation/webkit/wkwebviewconfiguration/limitsnavigationstoappbounddomains
+      ///
+      /// ## Platform-specific
+      /// - **iOS**: Supported since version 14.0+.
+      /// - **Linux / Windows / Android / MacOS:** Unsupported.
```
Legend-Master: This can be removed and moved to `WebviewAttributes::limit_navigations_to_app_bound_domains`

```suggestion
```

zphrs:
When re-reading the doc comment while making the above fixes I realized I forgot to implement this:

> Assets served through custom protocols like Tauri's IPC are added to the list automatically.

Because we don't know what domains people use for their custom protocols at build time (e.g. custom://domain1, custom://domain2, etc.), I'll change the comment to be:
```text
Assets served through custom protocols configured by Tauri (tauri://localhost, ipc://localhost, and 
isolation-{uuid}://localhost) are added to the WKAppBoundDomains list automatically. Domains on 
protocols defined via 
[`register_asynchronous_uri_scheme_protocol`](tauri::Builder::register_asynchronous_uri_scheme_protocol) 
or [`register_uri_scheme_protocol`](tauri::Builder::register_uri_scheme_protocol) are not added 
automatically. If you use a custom protocol, be sure to list all domains you use on such protocols 
in the Info.plist file.
```
I'll try to get the automatic `Info.plist` additions implemented today. I also just pushed the nits from your review.

zphrs:
After inspecting the WebKit source code, I found that the ENABLE_APP_BOUND_DOMAINS is only set to true if PLATFORM(IOS_FAMILY):
https://github.com/WebKit/WebKit/blob/1029c99e92ffa6735a57c79f0d436702f27e0e38/Source/WTF/wtf/PlatformEnableCocoa.h#L83-L85

It also seems that simply adding `localhost` to the WKAppBoundDomains Info.plist field will automatically allow any protocol's `localhost` domain based on reading through the source code. See: 
https://github.com/WebKit/WebKit/blob/898be46c29af657b82dab5b773edffa86e31715e/Source/WebKit/UIProcess/WebsiteData/Cocoa/WebsiteDataStoreCocoa.mm#L685-L694

and:

https://github.com/WebKit/WebKit/blob/898be46c29af657b82dab5b773edffa86e31715e/Source/WebCore/platform/RegistrableDomain.h#L43-L46

that uses:

https://github.com/WebKit/WebKit/blob/898be46c29af657b82dab5b773edffa86e31715e/Source/WebCore/platform/RegistrableDomain.h#L118-L124

Note that "https" is added just to immediately be stripped for actually constructing the list of domains that are app-bound. Also see the following for where schemes and domains are checked for app-boundedness. Note especially that **either** the protocol must be app-bound **or** the domain must be app-bound.

https://github.com/WebKit/WebKit/blob/898be46c29af657b82dab5b773edffa86e31715e/Source/WebKit/UIProcess/WebsiteData/Cocoa/WebsiteDataStoreCocoa.mm#L732-L737

So I will just add localhost by default to the WKAppBoundDomains Info.plist field if limitNavigationsToAppBoundDomains is set in the tauri config.

zphrs:
@Legend-Master Sorry I forgot to add the automatic addition of localhost to the Info.plist file if limitNavigationsToAppBoundDomains is set in the tauri.config.json file. Now I believe this PR should be good to go.

crates/tauri-cli/src/helpers/plist.rs
```diff
@@ -33,6 +33,16 @@ pub fn merge_plist(src: Vec<PlistKind>) -> crate::Result<plist::Value> {
     };
     if let Some(dict) = src_plist.into_dictionary() {
       for (key, value) in dict {
+        if let Some(new) = value.as_array() {
```
Legend-Master: Why don't we do this in the caller side like we did for other things like `entitlements`?

crates/tauri-cli/src/mobile/ios/build.rs
```diff
@@ -259,6 +259,21 @@ pub fn run(options: Options, noise_level: NoiseLevel, dirs: &Dirs) -> Result<Bui
       }
     }
   }
+
+  // Check if any window has limit_navigations_to_app_bound_domains set to true
+  if tauri_config
+    .app
+    .windows
+    .iter()
+    .any(|w| w.limit_navigations_to_app_bound_domains)
```
Legend-Master: This does not account for webviews created dynamically, what about we add a new field in the config file instead (and document it in `limit_navigations_to_app_bound_domains`)?

crates/tauri-cli/src/helpers/plist.rs
zphrs: Because ideally the set of domains detected based on the "tauri.conf.json" file will be unioned with the set of domains manually specified in the "Info.plist" and "Info.ios.plist" files so that the user only needs to specify dynamically accessed domains in the user-provided "Info.plist" files and all domains used in the "windows" array with the "limitDomainsToAppBoundDomains" flag set will automatically get added.

For example, if there is a window config of:
```json
{
    "app": {
        "windows": [
            {
                "url": "index.html",
                "limitDomainsToAppBoundDomains": true,
            },
            {
                "url": "https://example.com",
                "limitDomainsToAppBoundDomains": true,
            }
        ],
    }
}
```

and an `Info.ios.plist` of:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
<key>WKAppBoundDomains</key>
<array>
    <string>other.example</string>
    <string>localhost</string>
</array>
</dict>
</plist>
```
then ideally the final output plist would be:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
<key>WKAppBoundDomains</key>
<array>
    <string>other.example</string>
    <string>localhost</string>
    <string>example.com</string>
</array>
</dict>
</plist>
```
The PR you reviewed always added the localhost domain and would not properly configure WKAppBoundDomains for the above case because the example.com domain would not be added to the array. Now I realize it makes more sense to extract the domain of any window(s) with `limitDomainsToAppBoundDomains: true` to add to the WKAppBoundDomains list. I just pushed a fix for this.

crates/tauri-cli/src/mobile/ios/build.rs
zphrs: I think that keeping any manual WKAppBoundDomains configuration for dynamic calls in the Info.ios.plist file makes more sense over adding another parameter to tauri.conf.json because it keeps iOS-specific configuration in an iOS-specific file. 

I also did document that dynamically created windows would need to add the domain of that window to the WKAppBoundDomains array in the documentation for the dynamic limit_app_bound_domains dynamic configuration calls.

Possibly it makes more sense to not automatically set up the WKAppBoundDomains array for windows configured in tauri.conf.json, just for consistency sake.

crates/tauri-cli/src/helpers/plist.rs
Legend-Master: I see, thanks for the super detailed explanations. I wonder if this could break other things where duplicates are intended?

crates/tauri-cli/src/mobile/ios/build.rs
Legend-Master: > I think that keeping any manual WKAppBoundDomains configuration for dynamic calls in the Info.ios.plist file makes more sense over adding another parameter to tauri.conf.json because it keeps iOS-specific configuration in an iOS-specific file.

That makes sense. I was thinking something like `bundle > ios > limit_navigations_to_app_bound_domains` but yeah, keeping it in `Info.ios.plist` sounds better.

> Possibly it makes more sense to not automatically set up the WKAppBoundDomains array for windows configured in tauri.conf.json, just for consistency sake.

I personally think the auto setup might be a bit unexpecting for some people, but as long as the documentation's there, I'm good with either decisions. Also asking if @FabianLars has any opinions?

FabianLars: I personally would leave the whole config to the user. But as a compromise we can add tauri://localhost automatically, but imho nothing more.

This is a very security sensitive topic so we should be overly cautious.
Also, what we add automatically can't be removed by the user.

fwiw, capacitor doesn't inject anything.

zphrs: Yeah I’m with @FabianLars, I’ll revert my automatic additions of any domains to WKAppBoundDomains and unify the documentation between the dynamic calls and the WindowConfig param.

zphrs:
@FabianLars @Legend-Master I've reverted the automatic addition of WKAppBoundDomains and updated the documentation accordingly. Hopefully this PR is finally good to go 😅

Legend-Master (approved):
Thanks! Sorry about the wait the and thanks for the patience!

~~_Just another day, we'll be able to celebrate the anniversary of this PR 😂_~~

PR #13881, opened 2025-07-23
feat: add config option for custom watch folders

pepperoni505:
This PR adds support to define custom watch folders for `tauri dev`. A new entry under `build` in config named `watchFolders` accepts an array of relative (or absolute) paths to watch


crates/tauri-utils/src/config.rs
```diff
@@ -2892,6 +2892,9 @@ pub struct BuildConfig {
   ///   - This feature requires tauri-plugin 2.1 and tauri 2.4
   #[serde(alias = "remove-unused-commands", default)]
   pub remove_unused_commands: bool,
+  /// List of relative paths to directories to watch for changes when running `tauri dev`.
```
Legend-Master: They could be absolute paths as well right?

crates/tauri-utils/src/config.rs
```diff
@@ -2892,6 +2892,9 @@ pub struct BuildConfig {
...
+  #[serde(alias = "watch-directories", default)]
+  pub watch_folders: Vec<PathBuf>,
```
Legend-Master: Is this additional folders to watch? If so could we rename it to something like `additional_watch_folders` so it's not ambiguate?

Legend-Master (commented):
Didn't look into it too much yet, just some early feedbacks:

crates/tauri-utils/src/config.rs
pepperoni505: Yep, good catch! Fixed

crates/tauri-utils/src/config.rs
pepperoni505: Resolved

crates/tauri-cli/src/interface/rust.rs
```diff
@@ -489,10 +503,11 @@ impl Rust {
     let (tx, rx) = sync_channel(1);
     let frontend_path = frontend_dir();
 
-    let watch_folders = get_watch_folders()?;
+    let additional_watch_folders = get_watch_folders(additional_watch_folders)?;
```
Legend-Master: So they should still be `watch_folders` right? (`additional_watch_folders` is the parameter in)

crates/tauri-cli/src/interface/rust.rs
```diff
@@ -411,13 +415,22 @@ fn expand_member_path(path: &Path) -> crate::Result<Vec<PathBuf>> {
...
 
   // We always want to watch the main tauri folder.
   let mut watch_folders = vec![tauri_path.to_path_buf()];
 
+  // Add the additional watch folders, resolving the path from the tauri path if it is relative
+  watch_folders.extend(additional_watch_folders.iter().cloned().map(|dir| {
+    if dir.is_absolute() {
+      dir
+    } else {
+      tauri_path.join(dir)
```
Legend-Master, line 431: We need to have the path converted to absolute paths here with `std::path::absolute` or the `common_path::common_path_all` above would probably fail on things like `/somefolder/to/tauri/../watch-folder`

crates/tauri-utils/src/config.rs
```diff
@@ -2892,6 +2892,9 @@ pub struct BuildConfig {
   ///   - This feature requires tauri-plugin 2.1 and tauri 2.4
   #[serde(alias = "remove-unused-commands", default)]
   pub remove_unused_commands: bool,
+  /// List of relative or absolute paths to directories to watch for changes when running `tauri dev`.
```
Legend-Master: We can omit `relative or absolute` if it takes any paths 🙃, for example:

```md
Additional paths to watch for changes when running `tauri dev`
```

crates/tauri-cli/src/interface/rust.rs
pepperoni505: Ah whoops, was the result of search and replace 😅

crates/tauri-cli/src/interface/rust.rs
```diff
@@ -411,13 +416,30 @@ fn expand_member_path(path: &Path) -> crate::Result<Vec<PathBuf>> {
...
+      .map(|dir| {
+        let path = if dir.is_absolute() {
+          dir
+        } else {
+          tauri_path.join(dir)
+        };
+
+        canonicalize(path)
+      })
+      .filter_map(Result::ok),
```
Legend-Master: Do we want to ignore all non-existing paths or do we want some sort of warnings?

Legend-Master (commented):
Looks good, do you mind also adding a change file?

crates/tauri-cli/src/interface/rust.rs
pepperoni505: Good point. Just added an error log, can you check to see if that wording is good?

crates/tauri-cli/src/interface/rust.rs
```diff
@@ -411,13 +416,36 @@ fn expand_member_path(path: &Path) -> crate::Result<Vec<PathBuf>> {
...
+        canonicalize(path)
+      })
+      .filter_map(|res| match res {
+        Ok(val) => Some(val),
+        Err(e) => {
+          log::error!("Failed to locate additional watch folder: {}", e);
+          None
+        }
+      }),
+  );
```
Legend-Master:
```suggestion
  watch_folders.extend(additional_watch_folders.iter().filter_map(|dir| {
    let path = if dir.is_absolute() {
      dir
    } else {
      &tauri_path.join(dir)
    };

    let canonicalized = canonicalize(path).ok();
    if canonicalized.is_none() {
      log::warn!(
        "Additional watch folder '{}' not found, ignoring",
        path.display()
      );
    }
    canonicalized
  }));
```

Legend-Master (dismissed):
Looks good to me! Just a nitpick

.changes/additional-watch-folders.md
```diff
@@ -0,0 +1,6 @@
+---
+"tauri-cli": minor:feat
```
Legend-Master:
```suggestion
"@tauri-apps/cli": minor:feat
```
```suggestion
"tauri-cli": minor:feat
"@tauri-apps/cli": minor:feat
```

Legend-Master: Oops, committed the wrong one 😂

Legend-Master (dismissed):
Thanks!

Legend-Master (approved):

PR #13921, opened 2025-07-30
fix(windows): bundler should not sign non-binaries

### LINKED ISSUE #13341: [bug] SignTool Error: This file format cannot be signed because it is not recognized
### Describe the bug

When `bundle > resources` contains a map where some of the resources are not binary files, attempting to sign on Windows results in the build failing:

```
SignTool Error: This file format cannot be signed because it is not recognized.
```

I believe this is a regression after I recently migrated to latest.

### Reproduction

`tauri.conf.json` should have `bundle > resources` with something like:
```
    "resources": {
      "packaging/README.md": "README.md"
    },
```

`bundle > windows` should also be configured for proper signing, and for reference, I also am using `msi` as the bundle target.

Then build the project: `npm run tauri build`

### Expected behavior

Previously, I don't think files listed under the `resources` map were signed at build time. I think it makes sense that Tauri now attempts to, but it should skip non-binary files.

### Full `tauri info` output

```text
[✔] Environment
    ✔ WebView2: 135.0.3179.98
    ✔ rustc: 1.85.1 (4eb161250 2025-03-15)
    ✔ cargo: 1.85.1 (d73d2caf9 2024-12-31)
    ✔ Rust toolchain: 1.81.0-x86_64-pc-windows-msvc (overridden by 'C:\j\workspace\desktop-app_dwosk_tauri\rust-toolchain.toml')
    - node: 23.11.0
    - npm: 10.9.2

[-] Packages
    - tauri 🦀: 2.5.1
    - tauri-build 🦀: 2.2.0
    - wry 🦀: 0.51.2
    - tao 🦀: 0.33.0
    - @tauri-apps/cli : 2.5.0

[-] Plugins
    - tauri-plugin-updater 🦀: 2.7.1
    - tauri-plugin-single-instance 🦀: 2.2.3
    - tauri-plugin-dialog 🦀: 2.2.1
    - tauri-plugin-notification 🦀: 2.2.2
    - tauri-plugin-process 🦀: 2.2.1
    - @tauri-apps/plugin-process : not installed!
    - tauri-plugin-deep-link 🦀: 2.2.1
    - @tauri-apps/plugin-deep-link : not installed!
    - tauri-plugin-os 🦀: 2.2.1
    - @tauri-apps/plugin-os : not installed!
    - tauri-plugin-fs 🦀: 2.2.1

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../../frontend/build
    - devUrl: http://localhost:3000/
```

### Stack trace

```text

```

### Additional context

_No response_

Legend-Master:
The bundler will no longer try to sign non-binary and already signed binary files on Windows

Fix #13341

Note: didn't test yet

### DIFF
--- a/crates/tauri-bundler/src/bundle/windows/sign.rs
+++ b/crates/tauri-bundler/src/bundle/windows/sign.rs
@@ -259,12 +259,24 @@ pub fn try_sign<P: AsRef<Path>>(file_path: P, settings: &Settings) -> crate::Res
 }
 
 /// If the file is signable (is a binary file) and not signed already
+/// (will skip the verification if not on Windows since we can't verify it)
 pub fn should_sign(file_path: &Path) -> crate::Result<bool> {
-  Ok(
-    file_path
-      .extension()
-      .and_then(|extension| extension.to_str())
-      .is_some_and(|extension| matches!(extension, "exe" | "dll"))
-      && !verify(file_path)?,
-  )
+  let is_binary = file_path
+    .extension()
+    .and_then(|extension| extension.to_str())
+    .is_some_and(|extension| matches!(extension, "exe" | "dll"));
+  if !is_binary {
+    return Ok(false);
+  }
+
+  #[cfg(windows)]
+  {
+    let already_signed = verify(file_path)?;
+    Ok(!already_signed)
+  }
+  // Skip verification if not on Windows since we can't verify it
+  #[cfg(not(windows))]
+  {
+    Ok(true)
+  }
 }
--- a/crates/tauri-bundler/src/bundle/windows/sign.rs
+++ b/crates/tauri-bundler/src/bundle/windows/sign.rs
@@ -257,3 +257,14 @@ pub fn try_sign<P: AsRef<Path>>(file_path: P, settings: &Settings) -> crate::Res
   }
   Ok(())
 }
+
+/// If the file is signable (is a binary file) and not signed already
+pub fn should_sign(file_path: &Path) -> crate::Result<bool> {
+  Ok(
+    file_path
+      .extension()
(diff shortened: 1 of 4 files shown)


crates/tauri-bundler/src/bundle/windows/sign.rs
```diff
@@ -257,3 +257,26 @@ pub fn try_sign<P: AsRef<Path>>(file_path: P, settings: &Settings) -> crate::Res
   }
   Ok(())
 }
+
+/// If the file is signable (is a binary file) and not signed already
+/// (will skip the verification if not on Windows since we can't verify it)
```
lucasfernog, line 262: well on macOS we could use `codesign --verify`

FabianLars: on macos we need to sign all files and iirc the signature has to be the same for all files, not sure rn

Legend-Master: > well on macOS we could use `codesign --verify`

Does it verify Windows binaries?

> on macos we need to sign all files and iirc the signature has to be the same for all files, not sure rn

Hmm, does that apply to Windows installers? Or do you mean `codesign --verify` checks a directory instead of a file?

I'm not familiar about how signing works on macOS, how do they store the signing info? ([docs](https://developer.apple.com/library/archive/documentation/Security/Conceptual/CodeSigningGuide/AboutCS/AboutCS.html#:~:text=These%20signatures%20are%20stored%20in%20a%20file%20called%20_CodeSignature/CodeResources%20within%20the%20bundle.) says it's stored in `_CodeSignature/CodeResources` which seems to be an xml file)

FabianLars (approved):

PR #13943, opened 2025-08-03
fix(windows): `patch_binary` causing codesigning verification failure

### LINKED ISSUE #13942: [bug] binary-patching update (`232265c`) causes codesigning verification to fail
### Describe the bug

Hi, we use tauri to build EV-codesigned binaries for https://sentinowl.com and noticed our production binaries were no longer showing our organization as being a verified publisher during elevated Windows UAC prompts triggered by our app:

Image

When inspecting the digital signatures for the binary I noticed that it was being signed twice after we updated our CLI version to `2.7.1`:

Image

Image

Codesign verification tools were also complaining with verification errors, whereas prior to this CLI version bump we'd never received errors:
```
PS C:\Users\samli\Desktop\sigcheck> .\sigcheck.exe .\sentinowl.exe

Sigcheck v2.90 - File version and signature viewer
Copyright (C) 2004-2022 Mark Russinovich
Sysinternals - www.sysinternals.com

C:\Users\samli\Desktop\sigcheck\doublesigned073\sentinowl.exe:
        Verified:       The digital signature of the object did not verify.
        Signing date:   9:29 PM 7/29/2025
        Publisher:      Everzoo Inc.
        Company:        sentinowl
        Description:    Sentinowl
        Product:        Sentinowl
        Prod version:   0.7.3
        File version:   0.7.3
        MachineType:    64-bit
```
```
[i] sam@hydrus:#### (br:main) (st:1) $ osslsigncode verify -in sentinowl.exe
PE checksum   : 00D7B433

Signature Index: 0  (Primary Signature)

Message digest algorithm  : SHA256
Current message digest    : CDFB87ECA0FC8EC9F397D943974DEBF720863353C522D8039BCE0410EF942ACD
Calculated message digest : E7B7992FDFAB1F9404D8B707B7F3F9DCFB8834AD7C6766F89E5594BDA79F5D29     MISMATCH!!!

Signature verification: failed

...
```

### Reproduction

Since reproduction requires having an EV codesigning certificate provisioned by a root CA, I've included screenshots of the before/after of the Windows UAC prompts and `Digital Signatures` tabs for our signed binary below.

### Expected behavior

I was able to mitigate the bug by downgrading our `@tauri-apps/cli` version from `2.7.1` to `2.6.2`

This is what the UAC prompt looks like when building the binary using CLI version `2.6.2`:

Image

Along with the binary only having a single codesigning signature attached:

Image

Output from codesign verification tools:

```
PS C:\Users\samli\Desktop\sigcheck> .\sigcheck.exe .\sentinowl.exe

Sigcheck v2.90 - File version and signature viewer
Copyright (C) 2004-2022 Mark Russinovich
Sysinternals - www.sysinternals.com

C:\Users\samli\Desktop\sigcheck\fixed\sentinowl.exe:
        Verified:       Signed
        Signing date:   11:38 PM 8/2/2025
        Publisher:      Everzoo Inc.
        Company:        sentinowl
        Description:    Sentinowl
        Product:        Sentinowl
        Prod version:   0.7.4
        File version:   0.7.4
        MachineType:    64-bit
```

```
[i] sam@hydrus:#### (br:main) (st:1) $ osslsigncode verify -in sentinowl.exe
PE checksum   : 00D82C55

Signature Index: 0  (Primary Signature)

Message digest algorithm  : SHA256
Current message digest    : 9F4B7DD87AABC12847A81D9FFF8AB2E85C58303CCED9EDF7B5F68A2CE6CC2FE1
Calculated message digest : 9F4B7DD87AABC12847A81D9FFF8AB2E85C58303CCED9EDF7B5F68A2CE6CC2FE1

...

Signature CRL verification: ok
Signature verification: ok

Number of verified signatures: 1
Succeeded
```

### Full `tauri info` output

```text
> sentinowl@ tauri /src/sentinowl-desktop
> tauri info


[✔] Environment
    - OS: Debian 12.0.0 x86_64 (X64) (Unknown DE on Unknown Session)
    ✔ webkit2gtk-4.1: 2.48.3
    ✔ rsvg2: 2.54.7
    ✔ rustc: 1.88.0 (6b00bc388 2025-06-23)
    ✔ cargo: 1.88.0 (873a06493 2025-05-10)
    ✔ Rust toolchain: 1.88.0-x86_64-unknown-linux-gnu (default)
    - node: 22.17.0
    - pnpm: 10.13.1
    - npm: 10.9.2

[-] Packages
    - tauri 🦀: 2.7.0
    - tauri-build 🦀: 2.3.1
    - wry 🦀: 0.52.1
    - tao 🦀: 0.34.0
    - @tauri-apps/api : 2.7.0
    - @tauri-apps/cli : 2.7.1

[-] Plugins
    - tauri-plugin-fs 🦀: 2.4.1
    - tauri-plugin-shell 🦀: 2.3.0
    - @tauri-apps/plugin-shell : 2.3.0
    - tauri-plugin-dialog 🦀: 2.3.2
    - @tauri-apps/plugin-dialog : 2.3.2

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - framework: React
    - bundler: Vite
```

### Stack trace

```text
<N/A>
```

### Additional context

I did a `git bisect` between `2.6.2` (https://github.com/tauri-apps/tauri/commit/02775963415f32ec3ac86b0a07b032b6e478860f) and `2.7.1` (https://github.com/tauri-apps/tauri/commit/fd63f229d562397a503c784f008fcc4b338746dc) to find the breaking change: https://github.com/tauri-apps/tauri/commit/232265c70e1c213bbb3f84b5541ddc07d330fce1 / https://github.com/tauri-apps/tauri/pull/13209

We're using the following tauri build command to trigger our builds from within linux hosts/containers:
```
$ pnpm tauri build --target x86_64-pc-windows-msvc --runner cargo-xwin -- --bin sentinowl --features gui
```

This results in both MSI and NSIS bundle flows being triggered since we're not able to pass `--bundles NSIS` as an arg to restrict the build to a single package type (I'm guessing `NSIS` is being rejected as a valid arg due to our build hosts being linux)

acx0:
fixes #13942

https://github.com/tauri-apps/tauri/pull/13209 introduced changes causing codesigned binaries to fail verification

Fix has been verified by building the tauri CLI locally and using that to publish EV-signed binaries hosted on https://sentinowl.com

### DIFF
--- a/.changes/fix-binary-patching-codesign-verification-failure.md
+++ b/.changes/fix-binary-patching-codesign-verification-failure.md
@@ -1,5 +1,6 @@
 ---
-'@tauri-apps/cli': 'patch:sec'
+'tauri-cli': 'patch:bug'
+'@tauri-apps/cli': 'patch:bug'
 ---
 
 Fix codesigning verification failures caused by binary-patching during bundling
--- a/.changes/fix-binary-patching-codesign-verification-failure.md
+++ b/.changes/fix-binary-patching-codesign-verification-failure.md
@@ -1,0 +1,5 @@
+---
+'@tauri-apps/cli': 'patch:sec'
+---
+
+Fix codesigning verification failures caused by binary-patching during bundling
--- a/crates/tauri-bundler/src/bundle.rs
+++ b/crates/tauri-bundler/src/bundle.rs
@@ -53,7 +53,7 @@ pub use self::{
 use anyhow::Context;
 pub use settings::{NsisSettings, WindowsSettings, WixLanguage, WixLanguageConfig, WixSettings};
 
-use std::{fmt::Write, path::PathBuf};
+use std::{fmt::Write, path::PathBuf, io::{Seek, SeekFrom}};
 
 /// Generated bundle metadata.
 #[derive(Debug)]
@@ -122,7 +122,23 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
     .iter()
     .find(|b| b.main())
     .expect("Main binary missing in settings");
+  let main_binary_path = settings.binary_path(main_binary);
 
+  // When packaging multiple binary types, we make a copy of the unsigned main_binary so that we can
+  // restore it after each package_type step. This avoids two issues:
+  //  - modifying a signed binary without updating its PE checksum can break signature verification
+  //    - codesigning tools should handle calculating+updating this, we just need to ensure
+  //      (re)signing is performed after every `patch_binary()` operation
+  //  - signing an already-signed binary can result in multiple signatures, causing verification errors
+  let main_binary_reset_required =
+    matches!(target_os, TargetPlatform::Windows) && settings.can_sign() && package_types.len() > 1;
+  let mut unsigned_main_binary_copy = tempfile::tempfile()?;
+  if main_binary_reset_required {
+    let mut unsigned_main_binary = std::fs::File::open(&main_binary_path)?;
+    std::io::copy(&mut unsigned_main_binary, &mut unsigned_main_binary_copy)?;
+  }
+
+  let mut main_binary_signed = false;
   let mut bundles = Vec::<Bundle>::new();
   for package_type in &package_types {
     // bundle was already built! e.g. DMG already built .app
@@ -130,14 +146,22 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
       continue;
     }
 
-    if let Err(e) = patch_binary(&settings.binary_path(main_binary), package_type) {
+    if let Err(e) = patch_binary(&main_binary_path, package_type) {
       log::warn!("Failed to add bundler type to the binary: {e}. Updater plugin may not be able to update this package. This shouldn't normally happen, please report it to https://github.com/tauri-apps/tauri/issues");
     }
 
     // sign main binary for every package type after patch
     if matches!(target_os, TargetPlatform::Windows) && settings.can_sign() {
-      let bin_path = settings.binary_path(main_binary);
-      windows::sign::try_sign(&bin_path, settings)?;
+      if main_binary_signed && main_binary_reset_required {
+        let mut signed_main_binary = std::fs::OpenOptions::new()
+          .write(true)
+          .truncate(true)
+          .open(&main_binary_path)?;
+        unsigned_main_binary_copy.seek(SeekFrom::Start(0))?;
+        std::io::copy(&mut unsigned_main_binary_copy, &mut signed_main_binary)?;
+      }
+      windows::sign::try_sign(&main_binary_path, settings)?;
+      main_binary_signed = true;
     }
 
     let bundle_paths = match package_type {
@@ -160,6 +184,7 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
 
       #[cfg(target_os = "windows")]
       PackageType::WindowsMsi => windows::msi::bundle_project(settings, false)?,
+      // note: don't restrict to windows as NSIS installers can be built in linux using cargo-xwin
       PackageType::Nsis => windows::nsis::bundle_project(settings, false)?,
 
       #[cfg(target_os = "linux")]
--- a/crates/tauri-bundler/src/bundle.rs
+++ b/crates/tauri-bundler/src/bundle.rs
@@ -53,7 +53,11 @@ pub use self::{
 use anyhow::Context;
 pub use settings::{NsisSettings, WindowsSettings, WixLanguage, WixLanguageConfig, WixSettings};
 
-use std::{fmt::Write, path::PathBuf, io::{Seek, SeekFrom}};
+use std::{
+  fmt::Write,
+  io::{Seek, SeekFrom},
+  path::PathBuf,
+};
 
 /// Generated bundle metadata.
 #[derive(Debug)]
--- a/crates/tauri-bundler/src/bundle/windows/util.rs
+++ b/crates/tauri-bundler/src/bundle/windows/util.rs
@@ -85,8 +85,7 @@ pub fn os_bitness<'a>() -> Option<&'a str> {
 }
 
 pub fn patch_binary(binary_path: &PathBuf, package_type: &crate::PackageType) -> crate::Result<()> {
-  let file_data = std::fs::read(binary_path)?;
-  let mut file_data = file_data; // make mutable
+  let mut file_data = std::fs::read(binary_path)?;
 
   let pe = match goblin::Object::parse(&file_data)? {
     goblin::Object::PE(pe) => pe,
@@ -132,6 +131,8 @@ pub fn patch_binary(binary_path: &PathBuf, package_type: &crate::PackageType) ->
     )
   })?;
 
+  // see "Relative virtual address (RVA)" for explanation of offset arithmetic here:
+  // https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#general-concepts
   let file_offset = rdata_section.pointer_to_raw_data as usize
     + (rva as usize).saturating_sub(rdata_section.virtual_address as usize);
 


Legend-Master (dismissed):
Thanks! We'll need a [change file](https://github.com/tauri-apps/tauri/blob/dev/.changes/README.md) and a `cargo fmt` and we can merge it

acx0:
Thanks for the quick review @Legend-Master, I'll update the PR to include those.

What are your thoughts on making this error fatal?
```rust
    if let Err(e) = patch_binary(&main_binary_path, package_type) {
      log::warn!("Failed to add bundler type to the binary: {e}. Updater plugin may not be able to update this package. This shouldn't normally happen, please report it to https://github.com/tauri-apps/tauri/issues");
    }
```
I haven't had time to trace through how patching the `package_type` string into the binary affects client updates but based on the `plugin may not be able to update this package` log str, this seems like something that should be bubbled up immediately rather than being swallowed and left to hoping someone notices the warning amongst all the other build output.

Is there an RFC for this binary-patching feature that covers its scope and/or documents its failure modes?

Legend-Master:
> I haven't had time to trace through how patching the package_type string into the binary affects client updates but based on the plugin may not be able to update this package log str, this seems like something that should be bubbled up immediately rather than being swallowed and left to hoping someone notices the warning amongst all the other build output.

So this was changed in #13825 to basically prioritizing that we don't break the build and incrementally adding this feature (detecting installer type) at least in v2 (by the way this function hasn't landed yet, see https://github.com/tauri-apps/plugins-workspace/pull/2624, everything related to updaters are kinda slow to make, they need to be backward compatible and they're hard to test, also the design will need to be made with no breaking changes in the future in mind)

> Is there an RFC for this binary-patching feature that covers its scope and/or documents its failure modes?

Not really... And I wasn't involved much by the time this was merged, and we don't have any docs around this yet, mainly because https://github.com/tauri-apps/plugins-workspace/pull/2624 is not merged yet

I would probably prefer we bundle a text file along side the executable but I said it a bit too late and the binary patching method didn't show much of a problem at the time (no body really tried the signing process during the time so that's why this leaked through)

acx0:
Appreciate the context.

> I would probably prefer we bundle a text file along side the executable but I said it a bit too late and the binary patching method didn't show much of a problem at the time (no body really tried the signing process during the time so that's why this leaked through)

Is it too late to back off the binary-patching approach? It seems like it's going to be prone to regressions/bugs if there's no straight forward way to unit-test/verify that future code changes don't result in signed binaries from failing verification.

We've already had users uninstall our app due to the "untrusted publisher" warnings that were showing up, before we realized what was going on, so I'm more than happy to spend any time/effort to contribute here in order to prevent something like that from happening again.

Legend-Master:
It's quite bad we didn't find it earlier, I didn't even know you can sign a file twice to be honest 

> so I'm more than happy to spend any time/effort to contribute here.

That's awesome, but I currently don't have a better idea then the binary patching to be honest, the extra text file could probably work, but it would be also quite a lot of work, we need to make sure we place them at a good location, remember to sign them, remove them on uninstall, etc

Let me know if you have something in mind, also probably want to talk with @kandrelczyk here since he contributed that solution 

And just by the way, I might get busy the next few weeks so I might be able to follow that closely

.changes/fix-binary-patching-codesign-verification-failure.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tauri-apps/cli': 'patch:sec'
```
acx0: @Legend-Master let me know if `patch:sec` makes sense for this change

kandrelczyk:
Hey,

Just to add what we currently have regarding unit-tests and see if anything can be added to cover issues like this one in the future.

In updater plugin we have `test/app-updater` tests that build, update and run the app on Linux and Windows. On Linux we use AppImage and run the app twice checking that on first run it's updated  and on second run it's up to date. On Windows we do both Nsis and MSI but only run the app once to see that it is updated correctly. 

Obviously the first issue here is that this functionality is split between tauri-cli and updater plugin. The tests are in updater plugin but we don't have any additional verification in tauri-cli. Maybe some test that runs 'sigcheck.exe` on each binary should be added? Would that detect this issue? 

Do I understand it correctly that the current issue is caused by double signing the binary and only the second bundle is affected? In the unit test for updater plugin the "to be updated" app (v 1.0.0) is actually only bundled once. The updated version (v 1.0.1) is bundled twice but we never run it after updating. Would this issue have been detected by simply running the version with invalid signature? What exactly happens when we do this? Maybe the test should be modified to also run the app after update? I don't remember why it's not done on Windows. 

As for changing the binary patching to some other method based on bundling files I could do the Linux part. If someone can take care of the Windows part I think this could be done in a manner transparent for the updater plugin so that we don't have to do more work there. If we are worried that binary patching might cause more issues in the future it could be worth doing.

acx0:
Thanks for the additional context @kandrelczyk 

1. > and see if anything can be added to cover issues like this one in the future.

I'll take a look at the components you mentioned and see if I can add any tests that'll fail if the behaviour ever regresses to the 2.7.1 CLI functionality. I can follow up with these changes in a separate PR once I've had time to investigate/experiment.

2. > some test that runs 'sigcheck.exe` on each binary

I think something along the lines of this will need to be done using some multiplatform signature verification tool (`osslsigncode`/`jsign`/etc) or rust library, I'm just not sure what the validation output will look like if the certificate file is self-signed - as I imagine it would be for certs created in unit tests. E.g. `osslsigncode` validates the entire certificate chain and outputs a "pass/fail" based on that. I imagine `sigcheck` does the same.

3. > Do I understand it correctly that the current issue is caused by double signing the binary and only the second bundle is affected?

Can't say definitively because our build hosts are linux machines so `bundle.rs::bundle_project()` attempts to process both MSI and NSIS package types but only `windows::nsis::bundle_project()` is actually invoked since `windows::msi::bundle_project()` is gated behind `#[cfg(target_os = "windows")]` - but I imagine the first bundle would be processed correctly due to `main_binary` being in a clean/freshly-compiled state.

I believe if we'd been allowed to pass `--bundles NSIS` to our `tauri build` command we wouldn't have discovered this issue as `settings.package_types()` would only contained `PackageType::Nsis`. I was planning on submitting another diff to allow NSIS being supplied on linux builds of the tauri CLI.

But I imagine if the build is being done on a windows host the same bug would manifest as the signing tool we're using (jsign) is very popular for setups where the codesigning private key is hosted in a cloud-based KMS, and would end up being invoked in the same way we observed in linux.

4. > Would this issue have been detected by simply running the version with invalid signature?

I don't think so- Windows SmartScreen only seems to pop up if the downloaded binary, i.e. the NSIS installer, is unsigned. Once you've run the installer and launch the extracted program binary, the "unknown publisher" SmartScreen warning never shows up again. The bug didn't affect the signing of the NSIS installer `.exe` otherwise we would have noticed during our CI/testing that the signature was invalid. SmartScreen would've also blocked the execution of the installer in this case.

We only noticed the invalid signature of the program binary because our program performs a one-time admin-escalated operation on new installations. This triggers a UAC prompt which verifies+renders the details of any codesign certificates attached to the binary, and in this case was showing as "unknown publisher" which is what drove us to look into the codesigning flow.

5. > Maybe the test should be modified to also run the app after update?

We actually had the invalid-signature binaries deployed for about two weeks before we noticed the issue, and had also published a couple of automatic updates to users during that period without any errors. Our desktop agent emits client events/crash reports in the event of any update failures, which never occurred because of this behaviour mentioned above:

> Once you've run the installer and launch the extracted program binary, the "unknown publisher" SmartScreen warning never shows up again

It was actually our website observability/analytics dashboard that showed us a spike in new users signing up and immediately deleting their accounts after downloading the binary, which prompted us to investigate.

kandrelczyk:
Thanks @acx0 , now I understand what happened. I think the updater plugin tests are fine than and just some additional signature checks should be added to tauri-cli tests. There is already test present that bundles the patched binary so adding some signature verification shouldn't be hard.

.changes/fix-binary-patching-codesign-verification-failure.md
Legend-Master: I think it should just be `patch:bug`, security means we have fixed a security vulnerability which is not the case here

Legend-Master:
```suggestion
'tauri-cli': 'patch:bug'
'@tauri-apps/cli': 'patch:bug'
```

Legend-Master (dismissed):
Thanks!

I couldn't test this since I don't have a Windows signing key (and not enough time to setup a self signed certificate), @FabianLars could you help testing it?

FabianLars:
> could you help testing it?

hmm idk how valuable i can be here. i only have it set up on windows and there the current live version works fine for me as well as far as i can tell. Edit: I assume this is specific to the cross-platform compilation or by the signtool you use but i can't test either rn.
This PR doesn't break that though so that's good enough for me.

Legend-Master (approved):

Legend-Master:
Thank you all

kandrelczyk:
Hey, the work on the updater plugin stalled again. I understand everyone is busy but the work on tauri-cli will only be useful when we also merge the plugin PR so let's not forget about it :)

acx0:
Thank you all for the quick feedback and reviews! @Legend-Master @kandrelczyk @FabianLars if you ever need me to test out any changes to the bundling/signing/autoupdate flows, please don't hesitate to `@acx0` or email me:

* We use the NSIS bundler for our windows builds (via `cargo-xwin` on linux), and DMG bundler for macos builds (with signing+notarization enabled using our Apple Developer Program issued cert) so those should be quick/easy for me to test
* I imagine It shouldn't be too much work to provision a windows build host so I could probably assist with MSI bundler testing too
* We also have plans to distribute linux binaries soon and leverage the updater plugin there as well

Is there any other discussion forum/channel that tauri devs use for coordinating long-term changes? I'm not sure how much time I'll have to help out with new feature implementation at the moment but I'm definitely interested in providing assistance with testing/code review for upcoming bundling/updater changes, whether it's for the currrent binary-patching approach or considering/investigating any alternatives.

I've made some personal TODOs to look into the other possible work items that were discussed above, namely the `--bundles` parsing validation, but I'll reach out if I come up with any other ideas for testing the codesigning as well once I have some time.

FabianLars:
> Is there any other discussion forum/channel that tauri devs use for coordinating long-term changes?

hmm maybe the #contributing channel on our discord server (aside from github of course). we don't really have or discuss long-term plans atm though.

PR #13946, opened 2025-08-04
docs: how `security > capabilities` works

### LINKED ISSUE #12325: [bug] Listen permission problem
### Describe the bug

Hi. I'm having trouble with permission to use listen.

When I use listen in my code and run it for the first time, I get `event.listen not allowed. Command not found`

So I thought I should add the permission explicitly.

Documentation seems to indicate that `core:event:allow-listen` can be added as a permission, so I added that code to `src-tauri/capabilities/default.json`.

```
src-tauri/capabilities/default.json
{
  "$schema": "../gen/schemas/desktop-schema.json",
  "identifier": "default",
  "description": "Capability for the main window",
  "windows": ["main"],
  "permissions": [
    "core:default",
    "core:event:allow-listen",
    "shell:allow-open",
    "fs:read-all",
    "fs:write-all",
    "fs:allow-read-file",
    "fs:allow-write-file",
    {
      "identifier": "fs:scope",
      "allow": ["**/*"]
    }
  ]
}
```

However, this makes a new issue in build time. `Permission core:event:allow-listen not found`.

How can I use listen function?

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 15.1.0 arm64 (X64)
    - node: 21.7.1
    - pnpm: 8.15.4
    - npm: 10.5.0
    - deno: deno 1.46.3

[-] Packages
    - tauri :crab:: 2.2.0
    - tauri-build :crab:: 2.0.4
    - wry :crab:: 0.48.0
    - tao :crab:: 0.31.1
    - @tauri-apps/api : 2.2.0
    - @tauri-apps/cli : 2.2.2

[-] Plugins
    - tauri-plugin-fs :crab:: 2.2.0
    - @tauri-apps/plugin-fs : 2.2.0
    - tauri-plugin-shell :crab:: 2.2.0
    - @tauri-apps/plugin-shell : 2.2.0

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - framework: React
    - bundler: Vite
```

### Stack trace

_No response_

### Additional context

[1]
1. npm create tauri-app@latest
2. added this code in FE React
```
useEffect(() => {
    const unlisten = listen<string>("global-mouse-event", (event) => {
      const { x, y } = JSON.parse(event.payload);
      console.log(`Mouse moved to: X=${x}, Y=${y}`);
    });

    return () => {
      unlisten.then((f) => f());
    };
  }, []);
```
3. It needed permission.
`Unhandled Promise Rejection: event. listen not allowed. Command not found`

+ so I added permission
```
{
  "$schema": "../gen/schemas/desktop-schema.json",
  "identifier": "default",
  "description": "Capability for the main window",
  "windows": ["main"],
  "permissions": ["core:default", "core:event:allow-listen", "opener:default"]
}
```
-> Same problem

[2]
It has same problem even if i rm -rf .cargo and remove src-tauri/target and reinstall rustup.

Legend-Master:
Closes #12325

### DIFF
--- a/crates/tauri-utils/src/config.rs
+++ b/crates/tauri-utils/src/config.rs
@@ -2468,7 +2468,7 @@ pub struct SecurityConfig {
   pub pattern: PatternKind,
   /// List of capabilities that are enabled on the application.
   ///
-  /// By default (not set or empty list), all capability files from `./capabilities/` (relative to your `Cargo.toml`) are included,
+  /// By default (not set or empty list), all capability files from `./capabilities/` are included,
   /// by setting values in this entry, you have fine grained control over which capabilities are included
   ///
   /// You can either reference a capability file defined in `./capabilities/` with its identifier or inline a [`Capability`]
--- a/crates/tauri-utils/src/config.rs
+++ b/crates/tauri-utils/src/config.rs
@@ -2470,6 +2470,24 @@ pub struct SecurityConfig {
   ///
   /// By default (leftout or empty list), all capability files from `./capabilities/` (relative to your `build.rs`) are included,
   /// by setting values in this entry, you fine grain the control over which one to include
+  ///
+  /// You can either reference a capability file defined in `./capabilities/` with its identifier or inline a [`Capability`]
+  ///
+  /// ### Example
+  ///
+  /// ```json
+  /// {
+  ///   "app": {
+  ///     "capabilities": [
+  ///       "main-window",
+  ///       {
+  ///         "identifier": "drag-window",
+  ///         "permissions": ["core:window:allow-start-dragging"]
+  ///       }
+  ///     ]
+  ///   }
+  /// }
+  /// ```
   #[serde(default)]
   pub capabilities: Vec<CapabilityEntry>,
   /// The headers, which are added to every http response from tauri to the web view
--- a/crates/tauri-utils/src/config.rs
+++ b/crates/tauri-utils/src/config.rs
@@ -2468,8 +2468,8 @@ pub struct SecurityConfig {
   pub pattern: PatternKind,
   /// List of capabilities that are enabled on the application.
   ///
-  /// By default (leftout or empty list), all capability files from `./capabilities/` (relative to your `build.rs`) are included,
-  /// by setting values in this entry, you fine grain the control over which one to include
+  /// By default (not set or empty list), all capability files from `./capabilities/` (relative to your `build.rs`) are included,
+  /// by setting values in this entry, you have fine grained control over which capabilities are included
   ///
   /// You can either reference a capability file defined in `./capabilities/` with its identifier or inline a [`Capability`]
   ///
--- a/crates/tauri-utils/src/config.rs
+++ b/crates/tauri-utils/src/config.rs
@@ -2468,7 +2468,7 @@ pub struct SecurityConfig {
   pub pattern: PatternKind,
   /// List of capabilities that are enabled on the application.
   ///
-  /// By default (not set or empty list), all capability files from `./capabilities/` (relative to your `build.rs`) are included,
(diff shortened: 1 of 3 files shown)


crates/tauri-utils/src/config.rs
```diff
@@ -2468,7 +2468,8 @@ pub struct SecurityConfig {
   pub pattern: PatternKind,
   /// List of capabilities that are enabled on the application.
   ///
-  /// If the list is empty, all capabilities are included.
+  /// By default (leftout or empty list), all capability files from `./capabilities/` (relative to your `build.rs`) are included,
```
FabianLars:
```suggestion
  /// By default (unset or empty list), all capability files from `./capabilities/` (relative to your `build.rs`) are included,
```

most nit-picky nitpick so far i think 😂 i think left-out is too "complicated" of a word here (#support hurt me), something like unset or not set or something may be better imo.

Also, is it really relative to build.rs ? What if the user moves build.rs into src (setting it in Cargo.toml as well) do we generate/read those files to/from src/capabilities/ then? genuine question, never thought about this before.

crates/tauri-utils/src/config.rs
```diff
@@ -2468,7 +2468,8 @@ pub struct SecurityConfig {
...
+  /// by setting values in this entry, you fine grain the control over which one to include
```
FabianLars:
```suggestion
  /// by setting values in this entry, you have fine grain control over which capabilities are included.
```
idk why i'm like this

crates/tauri-utils/src/config.rs
Legend-Master: > Also, is it really relative to build.rs ?

I have not even tried that before 😂, our code in `tauri-build` just uses `./capabilities/**` to glob the files, I need to check if that's relative to `build.rs` or `Cargo.toml`

Legend-Master: Went with `not set` 😂

Legend-Master:
Honestly the json example in the schema rendered kinda terribly, don't know if we can do much at the moment

image

crates/tauri-utils/src/config.rs
Legend-Master: It's relative to `Cargo.toml`, updated, good call 😂

> https://doc.rust-lang.org/cargo/reference/build-scripts.html#:~:text=In%20addition%20to%20environment%20variables%2C%20the%20build%20script%E2%80%99s%20current%20directory%20is%20the%20source%20directory%20of%20the%20build%20script%E2%80%99s%20package.

FabianLars:
yeah, nothing in the tauri.conf.json tooltips renders, even simple a \`main\` doesn't render as `main` - no idea why🤷

FabianLars:
> Relative to Cargo.toml

Technically that means relative to tauri.conf.json as well.

FabianLars (dismissed):

Legend-Master:
> Technically that means relative to tauri.conf.json as well.

Actually, in that case maybe we should not include the `(relative to your Cargo.toml)` part if this is shown in `tauri.conf.json`, every relative path is relative to the file itself right?

FabianLars:
yep

FabianLars (approved):

Legend-Master:
> Honestly the json example in the schema rendered kinda terribly, don't know if we can do much at the moment
> 
> image

image

This should be fixed now with https://github.com/microsoft/vscode-json-languageservice/pull/283, released with vscode 1.106

> yeah, nothing in the tauri.conf.json tooltips renders, even simple a `main` doesn't render as `main` - no idea why🤷

So vscode treats all the schemas as plain texts, and only treats it as markdown if we opt in through `markdownDescription` (compared to just `description` in the json schema standard), I did this for the generated permission schema in #13057, and I kind of want to do it for the rest in #13662

PR #13947, opened 2025-08-04
fix(bundler): log sign command stdout formatted

FabianLars:
before
grafik

----
after
grafik

### DIFF
--- a/crates/tauri-bundler/src/bundle/windows/sign.rs
+++ b/crates/tauri-bundler/src/bundle/windows/sign.rs
@@ -214,9 +214,7 @@ pub fn sign_custom<P: AsRef<Path>>(
   let output = cmd.output_ok()?;
 
   let stdout = String::from_utf8_lossy(output.stdout.as_slice()).into_owned();
-  for line in stdout.lines() {
-    log::info!("{line}");
-  }
+  log::info!("{stdout}");
 
   Ok(())
 }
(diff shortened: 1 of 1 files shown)


crates/tauri-bundler/src/bundle/windows/sign.rs
```diff
@@ -214,7 +214,9 @@ pub fn sign_custom<P: AsRef<Path>>(
   let output = cmd.output_ok()?;
 
   let stdout = String::from_utf8_lossy(output.stdout.as_slice()).into_owned();
-  log::info!("{:?}", stdout);
```
Legend-Master, line 217: I believe `log::info!("{stdout}");` would just work

FabianLars: that would look like this 
grafik

i wouldn't mind that either especially since idk what other tools may output

Legend-Master: Ah, I see, not sure if it will look better if we add a new line at the beginning so everything's aligned

Legend-Master (dismissed):

crates/tauri-bundler/src/bundle/windows/sign.rs
FabianLars: how bout this then
grafik

Legend-Master: Liked it, maybe trim the output so there's no extra line at the end?

FabianLars: stop it with your extra wishes

Legend-Master (approved):

PR #13955, opened 2025-08-05
fix(example): runtime crash when counter lt 0

Jadon-Jesse:
+ Changed counter variable type from signed to unsigned


FabianLars (approved):
thank you

PR #13959, opened 2025-08-07
feat: expose internal TrayIcon

petersamokhin:
Hello there! Thanks a lot for Tauri.

This PR exposes the internal TrayIcon to be able to access the platform implementations.
This is pretty important and my use case is to access the underlying `ns_status_item()` (and then NSWindow, and then NSScreen). TrayIcon already exposes that, but it's unavailable if using the Tauri plugin.

### DIFF
--- a/crates/tauri/src/tray/mod.rs
+++ b/crates/tauri/src/tray/mod.rs
@@ -603,6 +603,11 @@ impl<R: Runtime> TrayIcon<R> {
       })
     })
   }
+
+  /// Get the inner tray icon.
+  pub fn inner(&self) -> &tray_icon::TrayIcon {
+    &self.inner
+  }
 }
 
 impl<R: Runtime> Resource for TrayIcon<R> {
--- a/crates/tauri/src/tray/mod.rs
+++ b/crates/tauri/src/tray/mod.rs
@@ -604,16 +604,16 @@ impl<R: Runtime> TrayIcon<R> {
     })
   }
 
-  /// Get the inner tray icon.
+  /// Do something with the inner [`tray_icon::TrayIcon`] on main thread
   ///
   /// Note that `tray-icon` crate may be updated in minor releases of Tauri.
   /// Therefore, it’s recommended to pin Tauri to at least a minor version when you’re using `with_inner_tray_icon`.
   pub fn with_inner_tray_icon<F, T>(&self, f: F) -> crate::Result<T>
   where
-    F: FnOnce(tray_icon::TrayIcon) -> T + Send + 'static,
+    F: FnOnce(&tray_icon::TrayIcon) -> T + Send + 'static,
     T: Send + 'static,
   {
-    run_item_main_thread!(self, |self_: Self| { f(self_.inner.clone()) })
+    run_item_main_thread!(self, |self_: Self| { f(&self_.inner) })
   }
 }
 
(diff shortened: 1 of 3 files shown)


FabianLars:
Thanks for the PR! Would you mind adding a comment similar to the one we have here? https://docs.rs/tauri/latest/tauri/webview/struct.WebviewWindow.html#method.with_webview If we expose the TrayIcon instance like this updating tray-icon would be a breaking change strictly speaking.
A comment about the main thread requirement when interacting with the instance would probably be nice to have too.

Also, I'd appreciate a changefile (check the .changes dir for examples) :)

petersamokhin:
@FabianLars definitely makes sense, thanks for noting! I added the changes document, and also updated the newly introduced fn to follow the `with_webview` approach that also explicitly handles the main thread.

There's one little difference from `with_webview`: the new fn returns whatever the passed `f` returns, instead of always returning `()`, just for convenience. Consumers still can return `()` if they wish but do it explicitly (even with a `;`). Let me know if you want this changed, not insisting here.

Thanks for the prompt response!

petersamokhin:
Fixed the build — somehow an import was removed from tests, reverted 🤷

crates/tauri/src/tray/mod.rs
```diff
@@ -603,6 +603,18 @@ impl<R: Runtime> TrayIcon<R> {
       })
     })
   }
+
+  /// Get the inner tray icon.
```
Legend-Master:
```suggestion
  /// Do something with the inner [`tray_icon::TrayIcon`] on main thread
```

crates/tauri/src/tray/mod.rs
```diff
@@ -603,6 +603,18 @@ impl<R: Runtime> TrayIcon<R> {
...
+  ///
+  /// Note that `tray-icon` crate may be updated in minor releases of Tauri.
+  /// Therefore, it’s recommended to pin Tauri to at least a minor version when you’re using `with_inner_tray_icon`.
+  pub fn with_inner_tray_icon<F, T>(&self, f: F) -> crate::Result<T>
+  where
+    F: FnOnce(tray_icon::TrayIcon) -> T + Send + 'static,
```
Legend-Master: I think we can avoid the clone here right? `&tray_icon::TrayIcon` should be able to do everything needed, and if people want, they could in rare cases clone it themselves

petersamokhin: @Legend-Master fixed 🫡

crates/tauri/src/tray/mod.rs
petersamokhin: @Legend-Master fixed 🫡

Legend-Master (dismissed):

crates/tauri/src/tray/mod.rs
```diff
@@ -623,7 +635,6 @@ mod tests {
   fn tray_event_json_serialization() {
     // NOTE: if this test is ever changed, you probably need to change `TrayIconEvent` in JS as well
 
-    use super::*;
```
Legend-Master: 😂

petersamokhin: aaaaaaaaaaaaaaaaaa

petersamokhin: sorry for that 💀  
idk jetbrains pls stop

petersamokhin: fixed

Legend-Master (approved):

PR #13968, opened 2025-08-09
fix(windows): fix double free (STATUS_HEAP_CORRUPTION) of resizing handler's userdata

s5bug:
Using WM_NCDESTROY instead of WM_DESTROY is more correct for freeing userdata, as windows can receive multiple WM_DESTROY events if they're parented. WM_NCDESTROY is the last message a window can _ever_ receive, whereas we were experiencing behavior where our application would crash due to a Tauri window receiving multiple WM_DESTROY events.

We tracked this down with ASAN (which reports a double-free almost immediately in our application) and this change fixes the issue.

### DIFF
--- a/.changes/fix-resizing-double-free.md
+++ b/.changes/fix-resizing-double-free.md
@@ -1,0 +1,5 @@
+---
+"tauri-runtime-wry": "patch:bug"
+---
+
+Use WM_NCDESTROY instead of WM_DESTROY to free window userdata, fixing a double-free occurring in the Windows resizing handler for undecorated windows which caused STATUS_HEAP_CORRUPTION
--- a/crates/tauri-runtime-wry/src/undecorated_resizing.rs
+++ b/crates/tauri-runtime-wry/src/undecorated_resizing.rs
@@ -249,7 +249,7 @@ mod windows {
         data.has_undecorated_shadows = wparam.0 != 0;
       }
 
-      WM_DESTROY => {
+      WM_NCDESTROY => {
         let data = data as *mut UndecoratedResizingData;
         drop(Box::from_raw(data));
       }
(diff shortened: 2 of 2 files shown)


Legend-Master (dismissed):

> as windows can receive multiple WM_DESTROY events if they're parented

I still want to know how/why is it receiving multiple `WM_DESTROY` events though, I don't think that should normally happen, and we might have another bug to fix

s5bug:
I wouldn't be surprised if it's due to some weird behavior with cross-parenting and not Tauri's fault at all? Google didn't help at all in that regard (I started off writing this trying to just prevent the double-free rather than seeing if the logic being in WM_DESTROY was the root issue). I'll go ahead and add a changes file!

s5bug:
Reading online says WM_DESTROY can be received multiple times if one sends WM_DESTROY manually instead of calling `DestroyWindow`... haven't checked if this happens in Tauri at all (and don't know how I would, as the few minutes it took me to write this today is the extent of my familiarity with the Tauri codebase)

For the record, without ASAN, the behavior in production Tauri as of right now means it still takes quite a while to receive a STATUS_HEAP_CORRUPTION. We had to create and destroy a bunch of windows to turn the double-free into a tangible error. It wouldn't shock me if there's issues elsewhere that nobody's caught just because they're so probabilistic, and getting ASAN to work on Windows is a chore (´・ω・`)

Legend-Master:
> Reading online says WM_DESTROY can be received multiple times if one sends WM_DESTROY manually instead of calling DestroyWindow

Yeah, also calling `DestroyWindow` multiple times I believe

>  haven't checked if this happens in Tauri at all 

I don't think we have any code sending `WM_DESTROY`, but can't be certain about calling `DestroyWindow` multiple times (since we do call it but we don't send `WM_DESTROY` manually like ever)

> For the record, without ASAN, the behavior in production Tauri as of right now means it still takes quite a while to receive a STATUS_HEAP_CORRUPTION. We had to create and destroy a bunch of windows to turn the double-free into a tangible error. It wouldn't shock me if there's issues elsewhere that nobody's caught just because they're so probabilistic, and getting ASAN to work on Windows is a chore (´・ω・`)

If it doesn't cause any other problems after this change, I guess we just call it a day 🙃

Legend-Master (approved):
Thanks again 🙃

Issue #13969, opened 2025-08-09
[bug] Panic instead of `Error` when HWND used for `parent_raw` is no longer valid by `build` (race condition)


s5bug:
### Describe the bug

A panic occurs in `tauri-runtime-wry` when a dead HWND is used for `parent_raw`.

### Reproduction

1. grab an HWND for a Window you can close
2. stall the Tauri app (through debugging or otherwise) between `parent_raw` and `build`
3. close the Window with that HWND
4. resume the Tauri app and attempt to `build` the Tauri window

### Expected behavior

I (personally) expect the error to be returned as part of the `Result` of `WebviewWindowBuilder`'s `build`.

### Full `tauri info` output

```text
> cargo tauri info

[✔] Environment
    - OS: Windows 10.0.26100 aarch64 (X64)
    ✔ WebView2: 138.0.3351.121
    ✔ MSVC: Visual Studio Community 2022
    ✔ rustc: 1.88.0-nightly (191df20fc 2025-04-18)
    ✔ cargo: 1.88.0-nightly (d811228b1 2025-04-15)
    ✔ rustup: 1.27.1 (54dd3d00f 2024-04-24)
    ✔ Rust toolchain: nightly-2025-04-19-aarch64-pc-windows-msvc (environment override by RUSTUP_TOOLCHAIN)
    - node: 22.14.0
    - npm: 11.4.0

[-] Packages
    - tauri 🦀: 2.7.0
    - tauri-build 🦀: 2.3.1
    - wry 🦀: 0.52.1
    - tao 🦀: 0.34.0
    - tauri-cli 🦀: 2.7.1

[-] Plugins
    - tauri-plugin-opener 🦀: 2.4.0

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ./src-ui
```

### Stack trace

```text
thread 'main' panicked at [removed]\.cargo\git\checkouts\tauri-bf88ae093107f7d0\151d58a\crates\tauri-runtime-wry\src\lib.rs:4306:55:
called `Result::unwrap()` on an `Err` value: OsError { line: 110, file: "[removed]\\.cargo\\registry\\src\\index.crates.io-1949cf8c6b5b557f\\tao-0.34.0\\src\\platform_impl\\windows\\mod.rs", error: IoError(Os { code: -2147023496, kind: Uncategorized, message: "Invalid window handle." }) }
```

### Additional context

I attempted to fix this myself by just turning the `.unwrap()` into a `?` but
[...]

PR #13970, opened 2025-08-09
fix: return error on window creation failed

Legend-Master:
Reference: #13969

Not sure if we could change the `CreateWindow` error varient to something like what we have for `CreateWebview` to include the actual error

```rust
  #[error("failed to create webview: {0}")]
  CreateWebview(Box<dyn std::error::Error + Send + Sync>),
```

I'm not entirely sure if this panic is intentional or not (I would assume not) and the git blame shows that it's from ages ago

### DIFF
--- a/.changes/return-error-on-window-creation-failed.md
+++ b/.changes/return-error-on-window-creation-failed.md
@@ -1,0 +1,5 @@
+---
+"tauri-runtime-wry": "patch:bug"
+---
+
+Return `tauri_runtime::Error::CreateWindow` instead of panic on window creation failed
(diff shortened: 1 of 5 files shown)


FabianLars:
> Not sure if we could change the CreateWindow error varient to something like what we have for CreateWebview to include the actual error

Modifying the existing variant would count as a breaking change i think? We could add a new variant but maybe just add a todo for v3 on the existing variant idk 🤷 

> I'm not entirely sure if this panic is intentional or not (I would assume not) and the git blame shows that it's from ages ago

Doesn't really matter if it is/was intention (to me at least), we should strive towards never panicking in our libs imo

FabianLars (approved):

PR #13993, opened 2025-08-12
feat(cli): check plugin versions for incompatibilities

### LINKED ISSUE #13960: [feat] warn user about Rust and JavaScript version mismatch in CLI
### Describe the problem

It's quite common that the Rust and JavaScript versions get misaligned without realizing

Warn the user about it on `tarui dev` and `tauri build`

### Alternatives considered

_No response_

### Additional context

_No response_

lucasfernog:
check core plugin versions for incompatibilities between Cargo and NPM releases

a plugin NPM/cargo version is considered "incompatible" if their major or minor versions are not equal

on dev we show an warning
on build we error out (with a `--ignore-incompatible-plugins` flag to prevent that)

this is an idea from @oscartbeaumont
we've seen several plugin changes that require updates for both the cargo and the NPM releases of a plugin, and if they are not in sync, the functionality does not work e.g. https://github.com/tauri-apps/plugins-workspace/pull/2573 where the change actually breaks the app updater if you miss the NPM update


FabianLars:
We also talked about this briefly here https://github.com/tauri-apps/tauri/issues/13960 (and in like a hundred discord threads 😂 )

tldr: Tony is mostly concerned about start up time. Similar for me, after all we removed the cli update check for that reason.

FabianLars:
oh and technically we sync plugins on patch versions at the moment. if we want the tool to only check minor then we need to make sure we don't have ipc breaking changes in patch releases (even if we consider it a fix). I don't mind either, we just need to agree on something and i guess document it somewhere.

FabianLars:
> tldr: Tony is mostly concerned about start up time. Similar for me, after all we removed the cli update check for that reason.

Well, should have probably tested the PR first before that. At least from my testing this seems to be pretty much instant. Too fast to measure its impact x) Nice job! Let's wait for the critics ( @Legend-Master ) though.

Legend-Master:
Looks really good from a quick glance, offloading to a thread so we don't block startup and using rayon to parallelize the js package manager calls, I'll give it a try as well see if I notice the speed differences

Legend-Master:
After some testing, the bottle neck is at getting the javascript packages, `npm`, `pnpm` `yarn`, list all support querying multiple packages at once, so we can use it and remove rayon instead, current one takes around 9 seconds (100ms parsing `Cargo.toml` + `Cargo.lock`) when running inside the `plugin-workspace` (most extreme scenario with all official plugins)

I'll see if I could push a change here

Legend-Master:
Pushed [c6e63a0](https://github.com/tauri-apps/tauri/pull/13993/commits/c6e63a0f86a9c24996dba95f302e0a35987df87b)

I have tested this with `npm` and `pnpm`, and the time used reduced to <1s, I think this is acceptable now with it being on a separate thread

Edit: ~~`yarn` needs special handling, it doesn't work right now~~ added and tested

crates/tauri-cli/src/helpers/mod.rs
```diff
@@ -59,11 +59,7 @@ pub fn resolve_tauri_path<P: AsRef<Path>>(path: P, crate_name: &str) -> PathBuf
 
 pub fn cross_command(bin: &str) -> Command {
   #[cfg(target_os = "windows")]
-  let cmd = {
-    let mut cmd = Command::new("cmd");
-    cmd.arg("/c").arg(bin);
-    cmd
-  };
+  let cmd = Command::new(format!("{bin}.cmd"));
```
Legend-Master: I can't seem to make the `yarn` command work without changing this, `cmd /c` seems to have some escape problems with `|` that `yarn` pattern uses

Would love some help

lucasfernog: this is the only way i could get it to work: `cmd /c "yarn list --pattern @tauri-apps/api^^^|@tauri-apps/plugin-log --json --depth 0"` maybe it's easier to just leave it as is? :joy:

Legend-Master: Or maybe, although it says it's deprecated, we can still use multiple values and don't pass in --pattern

lucasfernog: well yarn is not going to change anything ever..

Legend-Master: Switched to that

crates/tauri-cli/src/info/plugins.rs
```diff
@@ -18,6 +19,92 @@ use crate::{
...
+    } else {
+      None
+    };
+
+  let lock: Option<CargoLock> = get_workspace_dir()
+    .ok()
+    .and_then(|p| fs::read_to_string(p.join("Cargo.lock")).ok())
+    .and_then(|s| toml::from_str(&s).ok());
+
+  let know_plugins = helpers::plugins::known_plugins();
```
Legend-Master, line 48: I think we want to also include `@tauri-app/api` here? And maybe also `@tauri-app/cli`

crates/tauri-cli/src/build.rs
```diff
@@ -131,6 +137,32 @@ pub fn setup(
...
+          "{} (v{}) : {} (v{})",
+          p.crate_name, p.crate_version, p.npm_name, p.npm_version
+        )
+      })
+      .collect::<Vec<_>>()
+      .join("\n");
+    if options.ignore_incompatible_plugins {
+      log::error!("Found incompatible Tauri plugins. Make sure the NPM and crate versions are on the same major/minor releases:\n{}", incompatible_text);
+    } else {
+      anyhow::bail!("Found incompatible Tauri plugins. Make sure the NPM and crate versions are on the same major/minor releases:\n{}", incompatible_text);
```
Legend-Master: Instead of `incompatible Tauri plugins`, maybe `version mismatched tauri packages`?

crates/tauri-cli/src/build.rs
```diff
@@ -70,6 +71,11 @@ pub struct Options {
   /// On subsequent runs, it's recommended to disable this setting again.
   #[clap(long)]
   pub skip_stapling: bool,
+  /// Do not error out if aversion incompatibility is detected on an installed plugin.
+  ///
+  /// Only use this when you are sure the mismatch is incorrectly detected as incompatible plugin versions can lead to unknown behavior.
+  #[clap(long)]
+  pub ignore_incompatible_plugins: bool,
```
Legend-Master: Same here, maybe `ignore_version_mismatched_plugins`

crates/tauri-cli/src/helpers/npm.rs
```diff
@@ -254,4 +255,157 @@ impl PackageManager {
       Ok(None)
     }
   }
+
+  pub fn current_package_versions(
```
Legend-Master, line 260: Probably want to migrate the other ones to use this one as much as possible to speed things up, could push that to the future though

crates/tauri-cli/src/info/plugins.rs
lucasfernog: well we do not have a good way to match those..

FabianLars: we'd need to agree on whether to match the tauri package versions and which packages and patch or minor and then actually do that 🙃 

sooner or later we'll probably have to tackle this.

crates/tauri-cli/src/build.rs
FabianLars: if we also want to match the tauri packages as in the other comment than we could strip the `_plugins` and go with something like `ignore_version_mismatches`

crates/tauri-cli/src/info/plugins.rs
Legend-Master: `@tauri-app/api` and `tauri` will need to match in minor versions at least

FabianLars: except that we failed with that multiple times (especially in v1 at the end). again, i'm not against it but if we add the check here we must start to take this serious.

lucasfernog: i think before we actually run the check, we should sync our versions for a while, otherwise we suddenly break most builds

let's sync tauri, api and cli
and maybe pin minors for inner dependencies too

crates/tauri-cli/src/build.rs
lucasfernog: good idea

FabianLars:
With the intention being that we want to include tauri versions in the future as well we're a bit in a weird situation here because not checking the patch versions for plugins can be problematic since plugins are synced on the patch level. Some patch releases contained IPC sensitive changes.

We could take a step back and require patches to not make any ipc breaking changes (probably means that we mostly get js-only or rust-only PRs) but still keep syncing on patch releases anyway.

Legend-Master:
Personally, I think we should sync the versions for the tauri packages as well, like we're currently doing for the plugins, it is easier to understand, that if the versions exactly matches, it will definitely work (if you see the versions being different, your first reaction would be am I doing something wrong that they don't fully match? instead of knowing it's fine as long as they're on the same minor version)

For the check and IPC breaking changes though, I think we should keep them in minors only (except for cases where the last minor broke the IPC in some ways and it's for fixing that), so if people want (like if they rely on the `window.__TAURI_INTERNALS__.invoke` directly [^1]), they could pin the version range to minors (`~` instead of `^`)

Technically, if we do IPC breaking changes in tauri, plugin's global init script (`api-iife.js`) will break, in that case we should pin plugins' tauri version to a minor range, and release a new version on every tauri minor, the downsides being we'll need to release plugins a lot more frequently even with no code change at all (and if we don't sync the version of plugins with tauri, it could also be a bit hard to understand, but if we do sync them, we'll need to delay all the minor changes to be released with tauri)

[^1]: A reference of people might be doing it: https://github.com/tauri-apps/plugins-workspace/pull/2897#discussion_r2255624382

crates/tauri-cli/src/info/plugins.rs
Legend-Master: I think we could at least add a check that checks the version of `tauri` and `@tauri-app/api` here?

lucasfernog:
I agree with what you said overall.
But syncing tauri->plugins is a bit more complicated.. we can pull if off but it is harder to manage indeed, having to delay a particular plugin feature because tauri is not getting a minor release for a while..

lucasfernog:
can I merge this? I kinda want to release tauri soon

Legend-Master:
> But syncing tauri->plugins is a bit more complicated.. we can pull if off but it is harder to manage indeed, having to delay a particular plugin feature because tauri is not getting a minor release for a while..

Yeah, same thoughts, we could leave this one for now

> can I merge this? I kinda want to release tauri soon

Yeah sure, I just want to also add the `tauri` <-> `@tauri-app/api` check here, or maybe that could be for another day

crates/tauri-cli/src/build.rs
```diff
@@ -131,6 +137,32 @@ pub fn setup(
   mobile: bool,
 ) -> Result<()> {
   let tauri_path = tauri_dir();
+
+  log::info!("Looking up installed plugins to check incompatible versions...");
```
Legend-Master: I'm still a bit concerned about this blocking the build process, is it possible that we do this in another thread as well?

This is the time I got in the `plugin-workspace`'s API example in debug build

```
        Info Looking up installed plugins to check incompatible versions...
[crates\tauri-cli\src\build.rs:166:3] start.elapsed() = 1.4956269s
```

FabianLars:
> having to delay a particular plugin feature because tauri is not getting a minor release for a while..

we pulled out the core stuff into plugins to get around exactly that lol

edit: i think i misunderstood what you said, just act like i'm not here :)

Legend-Master:
> Yeah sure, I just want to also add the tauri <-> @tauri-app/api check here, or maybe that could be for another day

Added in cac22e42d3f09357fa8feaaff79994365d1e0b6f, and LOL, found out we got a mismatch in `plugin-workspace` while trying it 😂

crates/tauri-cli/src/build.rs
```diff
@@ -131,6 +137,17 @@ pub fn setup(
   mobile: bool,
 ) -> Result<()> {
   let tauri_path = tauri_dir();
+
+  let ignore_version_mismatches = options.ignore_version_mismatches;
+  std::thread::spawn(move || {
+    if let Err(error) = check_mismatched_packages(frontend_dir(), tauri_path) {
+      log::error!("{error}");
+      if !ignore_version_mismatches {
+        std::process::exit(1);
```
Legend-Master: This doesn't seem to stop the spawned processes (e.g. `vite`, `cargo`) though, feels like a pain to deal with 😂

lucasfernog: ah right.. no easy way to deal with this unless we go with tokio tasks, easier to have graceful shutdown :/

lucasfernog: it's either that or get back to sync (which is ok imo, build isn't run as often - and let's not forget cargo takes 500s while our check takes just 1 :joy: )

Legend-Master: > it's either that or get back to sync (which is ok imo, build isn't run as often - and let's not forget cargo takes 500s while our check takes just 1 😂 )

Yeah, ~1 second is acceptable, let's use the sync version for now, sigh, I wish the cargo build doesn't take that long 😂

Just for the reference time with release build, it took around 500ms to check in `plugins-workspace` and ~700ms to check in my own project (with `npm list` taking up almost all the time) on my machine with an 11th gen i5 11400

I honestly don't know why those js package managers are so slow (maybe because the js file sizes are too big so parsing them could take a while?), `pnpm --version` takes 300-400ms, similar numbers for npm, with our cli, `tauri --version` takes only 20-50ms

Legend-Master (dismissed):

Legend-Master (approved):

oscartbeaumont:
Thank you everyone for getting this through! So happy to know it's going to be much less likely anyone is going to end up in the same unfortunate position as us with a broken autoupdater in production caused by an `cargo update`.

PR #13995, opened 2025-08-13
fix(ios): Tauri iOS build with binary XCFramework dependencies

### LINKED ISSUE #13332: [bug] Tauri2 iOS Integration with AdMob - GoogleMobileAds Module Not Found
### Describe the bug

I am working on an iOS app using Tauri2, and I need to integrate AdMob. However, I am encountering an issue related to the GoogleMobileAds module.

I followed the instructions from both Tauri and Google and constructed the Package.swift for the tauri-plugin-admob as follows:

```swift
import PackageDescription

let package = Package(
    name: "tauri-plugin-admob",
    platforms: [
        .iOS(.v13)
    ],
    products: [
        .library(
            name: "tauri-plugin-admob",
            type: .static,
            targets: ["tauri-plugin-admob"])
    ],
    dependencies: [
        .package(name: "Tauri", path: "../.tauri/tauri-api"),
        .package(
            url: "https://github.com/googleads/swift-package-manager-google-mobile-ads.git",
            .upToNextMajor(from: "12.3.0")),
    ],
    targets: [
        .target(
            name: "tauri-plugin-admob",
            dependencies: [
                .byName(name: "Tauri"),
                .product(name: "GoogleMobileAds", package: "swift-package-manager-google-mobile-ads"),
            ],
            path: "Sources"
        )
    ]
)
```

In the Sources directory, I added import **GoogleMobileAds** in the Swift file. However, during compilation, I encountered the following error:

```
error: no such module 'GoogleMobileAds'
```

I tried a few other approaches, such as adding the following to **swiftSettings**:

```swift
swiftSettings: [
    .unsafeFlags([
        "-F", getFrameworkPath(for: "GoogleMobileAds"),
        "-F", getFrameworkPath(for: "UserMessagingPlatform"),
    ])
],
```

This seemed to pass the compilation stage, but I ran into a linker error during the linking phase:

```
"_OBJC_CLASS_$_GADMobileAds", referenced from:
in libtauri_plugin_admob-960460d5dc2cb946.rlib[20](AdmobPlugin.swift.o)
```

I also explored modifying build.rs and attempted various `cargo:rustc-link-search=native= `configurations, but nothing worked as expected. Unfortunately, I am not very familiar with either Rust or Swift, so I’ve been struggling to resolve the issue.

Based on my research, it seems like the GoogleMobileAds SPM is essentially just a wrapper for GoogleMobileAds.xcframework, and that swift-rs does not handle xcframeworks well.

Any guidance or help on how to solve this issue would be greatly appreciated. Thank you!

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 15.5.0 x86_64 (X64)
    ✔ rustc: 1.84.1 (e71f9a9a9 2025-01-27)
    ✔ cargo: 1.84.1 (66221abde 2024-11-19)
    ✔ Rust toolchain: stable-x86_64-apple-darwin (default)
    - node: 23.11.0
    - pnpm: 10.9.0
    - yarn: 1.22.19
    - npm: 10.2.4
    - bun: 1.2.10
    - deno: deno 2.2.11
```

### Stack trace

```text

```

### Additional context

_No response_

will3942:
Fix Tauri iOS build with binary XCFramework dependencies, allows extracting binaryTargets that are zipped and also not including XCFrameworks when linking.

Closes https://github.com/tauri-apps/tauri/issues/13332

### DIFF
--- a/crates/tauri-utils/src/build.rs
+++ b/crates/tauri-utils/src/build.rs
@@ -75,6 +75,7 @@ fn link_xcode_library(name: &str, source: impl AsRef<std::path::Path>) {
     .arg("OTHER_SWIFT_FLAGS=-no-verify-emitted-module-interface")
     .current_dir(source)
     .env_clear()
+    .env("PATH", std::env::var_os("PATH").unwrap_or_default())
     .status()
     .unwrap();
 
@@ -85,6 +86,8 @@ fn link_xcode_library(name: &str, source: impl AsRef<std::path::Path>) {
     .join("Products")
     .join(format!("{configuration}-{sdk}"));
 
+  println!("cargo::rustc-link-search=framework={}", lib_out_dir.display());
+  println!("cargo:include={}", lib_out_dir.display());
   println!("cargo:rerun-if-changed={}", source.display());
   println!("cargo:rustc-link-search=native={}", lib_out_dir.display());
   println!("cargo:rustc-link-lib=static={name}");
--- a/crates/tauri-utils/src/build.rs
+++ b/crates/tauri-utils/src/build.rs
@@ -90,7 +90,6 @@ fn link_xcode_library(name: &str, source: impl AsRef<std::path::Path>) {
     "cargo::rustc-link-search=framework={}",
     lib_out_dir.display()
   );
-  println!("cargo:include={}", lib_out_dir.display());
   println!("cargo:rerun-if-changed={}", source.display());
   println!("cargo:rustc-link-search=native={}", lib_out_dir.display());
   println!("cargo:rustc-link-lib=static={name}");
--- a/crates/tauri-utils/src/build.rs
+++ b/crates/tauri-utils/src/build.rs
@@ -86,7 +86,10 @@ fn link_xcode_library(name: &str, source: impl AsRef<std::path::Path>) {
     .join("Products")
     .join(format!("{configuration}-{sdk}"));
 
-  println!("cargo::rustc-link-search=framework={}", lib_out_dir.display());
+  println!(
+    "cargo::rustc-link-search=framework={}",
+    lib_out_dir.display()
+  );
   println!("cargo:include={}", lib_out_dir.display());
   println!("cargo:rerun-if-changed={}", source.display());
   println!("cargo:rustc-link-search=native={}", lib_out_dir.display());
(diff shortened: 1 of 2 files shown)


crates/tauri-utils/src/build.rs
```diff
@@ -85,6 +86,11 @@ fn link_xcode_library(name: &str, source: impl AsRef<std::path::Path>) {
     .join("Products")
     .join(format!("{configuration}-{sdk}"));
 
+  println!(
+    "cargo::rustc-link-search=framework={}",
+    lib_out_dir.display()
+  );
+  println!("cargo:include={}", lib_out_dir.display());
```
lucasfernog:
```suggestion
```

doesn't seem like cargo:include actually exists as a cargo instruction

crates/tauri-utils/src/build.rs
```diff
@@ -85,6 +86,11 @@ fn link_xcode_library(name: &str, source: impl AsRef<std::path::Path>) {
     .join("Products")
     .join(format!("{configuration}-{sdk}"));
 
+  println!(
+    "cargo::rustc-link-search=framework={}",
```
lucasfernog, line 90:
```suggestion
    "cargo:rustc-link-search=framework={}",
```

lucasfernog (changes requested):

crates/tauri-utils/src/build.rs
will3942: Is this correct? Double `:` appears to be the correct convention from the docs - https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-link-search (and also later in the existing build.rs code)

FabianLars: Both are correct. `::` is the new syntax. Tauri's msrv is high enough to be able to use the new syntax.

lucasfernog: oh ok i thought we were still tied to the old syntax :)

lucasfernog (approved):

crates/tauri-utils/src/build.rs
FabianLars: afaik it was added in 1.77

PR #14006, opened 2025-08-14
feat: Add default download handler

### LINKED ISSUE #4633: [bug] Can not download in front end
### Describe the bug

in the browser, we can use <a> like this:
```js
 const eleLink = document.createElement("a");
                eleLink.href = url;
                eleLink.download = `${title} - ${artist}`;
                eleLink.style.display = "none";
                document.body.appendChild(eleLink);
                eleLink.click();
                document.body.removeChild(eleLink);
```
to download file in front; but in tauri,this function can not take effect.

### Reproduction

_No response_

### Expected behavior

_No response_

### Platform and versions

```shell
› OS: Mac OS 12.4.0 X64
  › Node.js: 16.14.0
  › npm: 8.3.1
  › pnpm: 6.32.3
  › yarn: 1.22.17
  › rustup: 1.24.3
  › rustc: 1.61.0
  › cargo: 1.61.0
```

### Stack trace

_No response_

### Additional context

_No response_

### LINKED ISSUE #13734: Saving files from vanilla js code does nothing (but is working in modern browsers).
On modern webbrowsers, I can use the following code snippet ([source](https://codeberg.org/MaerchenfeeimGarten/DOORS/src/branch/main/index.html#L1067)) to save, for example, game progress or other data to a location choosen by the user:

```
function download_txt() {
        var hiddenElement = document.createElement('a')

         hiddenElement.href = 'data:attachment/text,' + encodeURI(JSON.stringify(geschafft_array)); //geschafft_array containes the game progress
         hiddenElement.target = '_blank';
         hiddenElement.download = 'savegame.doors';
         hiddenElement.click();
}
```

If I put the whole code of the game ([source code](https://codeberg.org/MaerchenfeeimGarten/DOORS/)) in the `src` folder of a vanilla-js-rust based tauri based project managed with cargo, this code does nothing and I did not get any error messages I could try to read. How could I resolve this issue? Maybe some permissions are not set or I need to replace the function with one that is compatible to the tauri framework?

FabianLars:
I wonder if the is_local check is a bit too much? after all browsers download files from all websites as well 🤔 

Together with https://github.com/tauri-apps/wry/pull/1594 & https://github.com/tauri-apps/wry/pull/1599 this now:
closes #13734
closes #4633


FabianLars:
I forgot to add a note about this on the builder methods. Will do that tomorrow (Friday)

Legend-Master:
Should we enable this in wry by default instead? I feel like it doesn't quite make sense to block downloads by default anyways, and it is already the case for webview2

> I wonder if the is_local check is a bit too much? after all browsers download files from all websites as well 🤔

Don't feel like we need to restrict that either

FabianLars:
Ah now that you mention it i meant to disable it by default on windows but forgot about it. Anyway, i don't care about the default so if you prefer that i'll change it in Wry.

PR #14009, opened 2025-08-15
feat(tauri): re-export `PixelUnit`, `PhysicalUnit`, `LogicalUnit`

WSH032:
If we don't export these APIs, we won't be able to use [WindowSizeConstraints](https://docs.rs/tauri/latest/tauri/struct.WindowSizeConstraints.html) directly.

*I noticed that tauri already exports a lot of items from the `dpi` crate at the top-level namespace, which feels a bit like namespace pollution to me 😂*

### DIFF
--- a/.changes/export-PixelUnit.md
+++ b/.changes/export-PixelUnit.md
@@ -1,0 +1,5 @@
+---
+"tauri": minor
+---
+
+re-export `PixelUnit`, `PhysicalUnit`, `LogicalUnit`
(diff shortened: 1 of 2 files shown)


FabianLars:
> I noticed that tauri already exports a lot of items from the dpi crate at the top-level namespace, which feels a bit like namespace pollution to me 😂

Agreed, it's getting a bit crowded there at the top level 😂 

created https://github.com/tauri-apps/tauri/issues/14011 🤷

FabianLars (approved):

PR #14012, opened 2025-08-15
feat: impl `AsRef<Window>` and `on_webview_event` for `WebviewWindow`

WSH032:
Similar to the purpose of <https://github.com/tauri-apps/tauri/pull/14008#issue-3324704887>, I implemented these methods for symmetry.

Currently, `Webview` still has two separate APIs: `set_bounds` and `set_size`, but I think these APIs are not suitable for `WebviewWindow`, so I didn't add them.

In addition, I also fixed some typos.

PR #14026, opened 2025-08-18
Remove `AsRef<Window>` on `WebviewWindow`

Legend-Master:
This reverts the `AsRef<Window>` impl for `WebviewWindow` made in #14012

cc @WSH032, after some discussion with the team, we think we should avoid this type of breaking changes in minors

And we can instead, add explicit methods like `as_window`, `as_webview`

### DIFF
--- a/.changes/WebviewWindow-on_webview_event.md
+++ b/.changes/WebviewWindow-on_webview_event.md
@@ -2,10 +2,4 @@
 tauri: minor:enhance
 ---
 
-- Implemented `Webview::on_webview_event` for `WebviewWindow` as well
-- Implemented `AsRef<Window<R>>` for `WebviewWindow<R>`
-
-    This can be considered a *BREAKING CHANGE* in very specific cases:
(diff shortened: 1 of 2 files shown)


amrbashir (approved):

WSH032:
It's fine.

I just looked up some resources and found that whether this is considered a breaking change is somewhat controversial: <https://std-dev-guide.rust-lang.org/breaking-changes/new-trait-impls.html#inference-breaks-when-a-second-generic-impl-is-introduced>.

Personally, I wouldn't consider it a breaking change; otherwise, `AsRef` would be no different from `Deref` (both could only have one implementation).

But since it does break plugins before `2.8`, reverting it is reasonable. This isn't very elegant—it's really unfortunate for Rust 😂

Legend-Master:
Yeah, that's the problem, it is not a problem by itself but the way we have a single AsRef impl to begin with makes it a breaking change and the fact that it does affect the plugins usages right now

PR #14031, opened 2025-08-18
feat(core): add support to universal app links on macOS

lucasfernog:
follow-up for https://github.com/tauri-apps/tao/pull/1108

this also requires an embedded.provisioningprofile to be provided, just like distributing to app stores


FabianLars (approved):
I can't test applinks right now but the entitlements are applied to the app correctly :)

PR #14037, opened 2025-08-19
feat(tauri): make AppManager::get_url public

seanaye:
## Summary
- Changes the visibility of `AppManager::get_url` from `pub(crate)` to `pub` to allow external access to the base URL configuration


FabianLars:
First of all, thanks for the PR :)

I don't think this PR does what you want it to do. You won't have access to get_url in a tauri app with this change either because AppManger is not directly exposed.

I'd recommend to start with tauri::Manager (it should probably get a `base_url()` method for this) and then see how its implement https://docs.rs/tauri/latest/tauri/trait.Manager.html#implementors

seanaye:
Ah sorry I assumed app manager was public without checking

seanaye:
I don't think I actually need to expose this functionality anymore, will close PR

PR #13876, opened 2025-07-22
feat: add `WebviewBuilder::on_new_window` and `WebviewBuilder::on_document_title_changed`

thlstsul:
feat: add `WebviewBuilder::on_new_window` and `WebviewBuilder::on_document_title_changed`

PR #14039, opened 2025-08-19
fix(tauri): build without the wry feature flag

lucasfernog:


Legend-Master:
I think it's (`-p tauri`) probably still not enough, `with_environment` clearly needs `webview2_com` which is also behind `wry` feature flag

lucasfernog:
well on Linux it showed me the error locally.. damn

Legend-Master:
Seems like the feature is enabled through this

https://github.com/tauri-apps/tauri/blob/c3252f72f625b2a8c4f9d5c0f279706a77999d4c/crates/tauri/Cargo.toml#L178

from #3566

lucasfernog:
so `cargo build` will be able to do the check (like i tested locally) but not cargo test..

Legend-Master:
> https://github.com/tauri-apps/tauri/blob/c3252f72f625b2a8c4f9d5c0f279706a77999d4c/crates/tauri/Cargo.toml#L178

Do you still remember the reason for us to do this back then? Or maybe we could try removing this?

Also ideally we should remove the dependency on `webview2-com` and `webkit2gtk` in `tauri-runtime`

> https://github.com/tauri-apps/tauri/pull/13876#issuecomment-3198988086

lucasfernog:
> Do you still remember the reason for us to do this back then? Or maybe we could try removing this?

the command tests needs this IIRC

> Also ideally we should remove the dependency on webview2-com and webkit2gtk in tauri-runtime

we should refactor this for tauri 2.9

Legend-Master:
Also should we also put `objc2-web-kit` and related APIs to behind `wry` feature?

PR #14052, opened 2025-08-21
feat(bundle): add --no-sign flag to skip code signing in bundling pro…

### LINKED ISSUE #11626: [feat] build without signing cli flag
### Describe the problem

Myself, and especially contributors to a Tauri-based app, do not want/have the signing keys when testing. However, when the various environment variables are not set, the build just fails.

There is no quick way to build a non-signed bundle of an app.

```bash
tauri build --no-sign
```

### Alternatives considered

Currently, the workaround is to have something like a `tauri.conf.dev.json` file with values like:

```json
"bundle": { "windows": null }
```
Then, use `tauri build --config src-tauri/tauri.conf.dev.json`.

I just find it odd there is no flag in the cli to disable signing.

### Additional context

It is helpful to go through the whole workflow of installing, manually opening, checking the closing and re-opening behaviour, checking uninstalling (if all expected files are removed). Also, with things like the updater plugin, checking the automated update/restart workflow.

ShigrafS:
### Description
This PR addresses an issue where building a Tauri application locally fails if code signing environment variables are not configured. It adds a new **`--no-sign`** CLI flag to the `tauri build` command, allowing developers and contributors to easily create unsigned application bundles for testing and development purposes.

---

### Summary
This PR adds the `--no-sign` flag to the `tauri build` command, streamlining the development workflow by allowing unsigned bundles to be built without the need for code signing keys. This is particularly useful for local testing, where the full build and install process can be checked without the overhead of configuring signing certificates.

### 🧩 Related Issue
Closes #11626 
`[feat] build without signing cli flag`

### ✨ What's New
- **New CLI Flag**: Adds the `--no-sign` option to `tauri build` and `tauri bundle`.
- **Bundler Logic**: Updates the `tauri-bundler` to check for this flag and skip the code signing step for both macOS and Windows.
- **Codebase Integration**: A new `no_sign` field is added to the bundler's `Settings` struct to pass this flag's state through the build pipeline.

### 💡 Why This Matters
- **Improved Developer Experience**: Simplifies the build process for contributors and new developers who don't have access to signing keys.
- **Faster Iteration**: Allows for rapid testing of the installer, uninstaller, and updater functionality without the time-consuming signing process.
- **Flexibility**: Provides a native CLI solution, eliminating the need for workarounds like using separate configuration files for local builds.

### How has this PR been tested?
- **Local Builds**: Tested with and without the `--no-sign` flag on both macOS and Windows to ensure that the signing process is correctly skipped.
- **Logic Verification**: Confirmed that the `no_sign` flag is properly handled within `tauri-bundler`, preventing the call to the signing functions.

### Is this a breaking change?
- No. This change adds a new, optional flag and does not modify any existing functionality.

### Does this PR require an update to the documentation?
- Yes. The `tauri build` and `tauri bundle` CLI references should be updated to include the new `--no-sign` flag and its purpose.

### Checklist:
- [x] The code has been tested locally
- [ ] Tests have been added to cover all new functionality
- [ ] The documentation has been updated to reflect any changes


ShigrafS:
Hi @Legend-Master 
This PR ready. 
Kindly review it.

Legend-Master:
Thanks for the PR, but before a proper review, first I'll need ask you to cleanup the unrelated changes and things like `// <-- ADDED THIS LINE TO FIX THE COMPILER ERROR` 😂

ShigrafS:
Sure! 
I'll make changes, verify build and ping you.

ShigrafS:
@Legend-Master 
Build successfully: `Finished `dev` profile [unoptimized + debuginfo] target(s) in 28m 13s` 

The other changes are due to linter. 
Please let me know how should I proceed with it. 
Thanks!

Legend-Master:
> The other changes are due to linter.
> Please let me know how should I proceed with it.

You'll need to revert them as well

ShigrafS:
Thank you for your guidance @Legend-Master 
I've made the necessary changes.

crates/tauri-cli/src/bundle.rs
```diff
@@ -92,6 +92,9 @@ pub struct Options {
   /// On subsequent runs, it's recommended to disable this setting again.
   #[clap(long)]
   pub skip_stapling: bool,
+
```
Legend-Master: Need some docs here

crates/tauri-bundler/src/bundle/settings.rs
```diff
@@ -778,6 +778,7 @@ pub struct Settings {
   target_platform: TargetPlatform,
   /// The target triple.
   target: String,
+  pub no_sign: bool,
```
Legend-Master:
```suggestion
  no_sign: bool,
```

Legend-Master (commented):
We'll also a change file

@FabianLars Since I'm not familiar with signing on macOS (I think it would fail if you have `signingIdentity` set but not the signing env vars), do you think we should also add this for macOS? I know we don't force the signing on Linux and the mobile signing process is not covered by tauri cli so probably don't need it for those platforms

crates/tauri-bundler/src/bundle/settings.rs
ShigrafS: Reverting this back to `public`, since making this private is breaking the code.

Legend-Master: This should be put into a getter below instead of being `pub` here

lucasfernog (approved):

ShigrafS:
@Legend-Master @lucasfernog 
Should I rebase the branch?

FabianLars:
nah you're good, github shows no merge conflicts yet. we're just waiting for 2.9 to get a bit closer before we merge it.

PR #14054, opened 2025-08-21
fix: set webview2 path before initializing runtime

### LINKED ISSUE #13817: Creating webviews doesn't account for fixed webview2 runtime in Windows environments without any runtime installed
In Windows systems which don't come with Webview included (Windows 7 and Windows 10), building Webviews using a fixed version Webview runtime presents a dialog with the message:

> Could not find the WebView2 Runtime.
> Make sure it is installed or download it from (...)
> You may have it installed on another user account, but it is not available for this one.

I was able to replicate it in Windows 7 with the `create-tauri-app` project.
Some remarks:
- Setup Rust 1.77.2 toolchain 
- Downgraded cargo lock to version 3
- Downgraded project dependencies for the rust version
- Also set up the dynamic link for vcruntime (`STATIC_VCRUNTIME="false"` & `-C target-feature=+crt-static`)

This message is presented because the condition to do so is only dependent on Windows's own webview version detection `GetAvailableCoreWebView2BrowserVersionString`, which is percolated up through `wry`'s `webview_version`.

Now, I might be mistaken but it seems `GetAvailableBrowserVersionString` might not account, as far as I can tell from their docs, the `WEBVIEW2_BROWSER_EXECUTABLE_FOLDER` environment variable set by tauri to indicate a fixed Webview runtime and only detecting installed runtime versions.

So, I would have expected to be able to build webviews with only a fixed runtime present in the system.

I seem to recall in older issues, `wry` not being intended to be concerned with the webview runtime distribution details [#1378](https://github.com/tauri-apps/tauri/issues/1378), so I think this shouldn't be a `wry` issue.

Perhaps the solution might be somewhere around or before reaching `tauri-runtime-wry`'s context initialization. But this could imply greater awareness of the `WEBVIEW2_BROWSER_EXECUTABLE_FOLDER` environment variable.

I understand if this is not feasable to resolve, particularly if it was only for Windows 7 (which I am biased to, because i'm currently trying to work with it). But I think this could be a problem as well with early windows 10 systems.

FabianLars:
fixes #13817

i don't really like not re-using resource_dir() but i couldn't think of anything that wasn't even worse (eg initializing dummy PackageInfo / Env was really ugly - adding new methods to tauri-utils that don't take those args felt weird as well).

### DIFF
--- a/crates/tauri/src/app.rs
+++ b/crates/tauri/src/app.rs
@@ -2165,6 +2165,26 @@ tauri::Builder::default()
       },
     };
 
+    // The env var must be set before the Runtime is created so that GetAvailableBrowserVersionString picks it up.
+    #[cfg(windows)]
+    {
+      if let crate::utils::config::WebviewInstallMode::FixedRuntime { path } =
+        &manager.config.bundle.windows.webview_install_mode
+      {
+        if let Some(exe_dir) = std::env::current_exe()
+          .ok()
+          .and_then(|p| p.parent().map(|p| p.to_path_buf()))
+        {
+          std::env::set_var("WEBVIEW2_BROWSER_EXECUTABLE_FOLDER", exe_dir.join(path));
+        } else {
+          #[cfg(debug_assertions)]
+          eprintln!(
+            "failed to resolve resource directory; fallback to the installed Webview2 runtime."
+          );
+        }
+      }
+    }
+
     #[cfg(any(windows, target_os = "linux"))]
     let mut runtime = if self.runtime_any_thread {
       R::new_any_thread(runtime_args)?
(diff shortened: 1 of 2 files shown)


crates/tauri/src/app.rs
```diff
@@ -2165,6 +2165,26 @@ tauri::Builder::default()
       },
     };
 
+    // The env var must be set before the Runtime is created so that GetAvailableBrowserVersionString picks it up.
+    #[cfg(windows)]
+    {
+      if let crate::utils::config::WebviewInstallMode::FixedRuntime { path } =
+        &manager.config.bundle.windows.webview_install_mode
+      {
+        if let Some(exe_dir) = std::env::current_exe()
```
lucasfernog: can we use `utils::platform::current_exe` instead?

FabianLars: ah whoops

lucasfernog (approved):

PR #14083, opened 2025-08-25
fix: don't set macos deployment target in dev

### LINKED ISSUE #11577: [bug] Tauri unilaterally overrides MACOSX_DEPLOYMENT_TARGET, causes sys crates to spuriously rebuild
### Describe the bug

On macOS, there's a special env var `MACOSX_DEPLOYMENT_TARGET` that implicitly configures the system's C compiler, linker, and opts in to or out of breaking changes in macOS. Objects built for different deployment targets may be incompatible. The `cc` crate [knows this](https://github.com/rust-lang/cc-rs/blob/7a786c5e3dc149f32e565360dd8c3e43748592f5/src/lib.rs#L3766) and emits `cargo:rerun-if-env-changed=MACOSX_DEPLOYMENT_TARGET`, which invalidates cache of sys crates whenever `cargo build` sees a different value of this variable.

Unfortunately, something in Tauri is trying to override `MACOSX_DEPLOYMENT_TARGET` to `10.13`.

This causes two bugs:

1. Even if I set `MACOSX_DEPLOYMENT_TARGET` env var to a different value, the project is built for 10.13 anyway.

2. The override causes `cargo` to invalidate cache of sys crates, and rebuild `objc-sys` and all of `wry`, `tauri`, and its plugins, every time `cargo tauri dev` sees any file change!

   > Dirty objc-sys v0.3.5: the env variable MACOSX_DEPLOYMENT_TARGET changed

   I'm not sure where Cargo is getting another value from. I suspect that `rust-analyzer` is building the project in a way that doesn't support Tauri's way of overriding `MACOSX_DEPLOYMENT_TARGET`, or perhaps there's just a race condition in Cargo that compares an unset value from before the build start with a value set during the build.

   The end result is that `cargo tauri dev` is painfully slow.

### Reproduction

To reproduce, modify `src/main.rs` to add:

```rust
compile_error!(env!("MACOSX_DEPLOYMENT_TARGET"));
```

and run:

```bash
unset MACOSX_DEPLOYMENT_TARGET
cargo tauri dev
```

The expected result:

```text
error: environment variable `MACOSX_DEPLOYMENT_TARGET` not defined at compile time
```

The actual result:

```text
error: 10.13
```

----

Second test:

```bash
export MACOSX_DEPLOYMENT_TARGET=11.0
cargo tauri dev
```

The expected result:

```text
error: 11.0
```

The actual result:

```text
error: 10.13
```

### Expected behavior

_No response_

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 15.1.0 arm64 (X64)
    - node: 23.1.0
    - yarn: 1.22.22
    - npm: 10.9.0

[-] Packages
    - tauri 🦀: 2.0.6
    - tauri-build 🦀: 2.0.2
    - wry 🦀: 0.46.3
    - tao 🦀: 0.30.5
    - tauri-cli 🦀: 2.0.4

[-] Plugins
    - tauri-plugin-shell 🦀: 2.0.2

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../src
```

### Stack trace

_No response_

### Additional context

_No response_

FabianLars:
fixes #11577 Also interesting: https://github.com/rust-lang/cargo/issues/13115

@lucasfernog i'm really indecisive about this. Do we want to try to "fix" it for all current users (this PR)? Or would you prefer if we'd add a .cargo/config.toml file to all our templates that sets the env var?
Or do you perhaps have another idea?

### DIFF
--- a/.changes/skip-deployment-target-in-dev.md
+++ b/.changes/skip-deployment-target-in-dev.md
@@ -1,7 +1,7 @@
 ---
 tauri-build: patch:enhance
 tauri-cli: patch:enhance
-@tauri-apps/cli: patch:enhance
+'@tauri-apps/cli': patch:enhance
 ---
 
 Tauri now ignores `macOS.minimumSystemVersion` in `tauri dev` to prevent forced rebuilds of macOS specific dependencies when using something like `rust-analyzer` at the same time as `tauri dev`.
--- a/.changes/skip-deployment-target-in-dev.md
+++ b/.changes/skip-deployment-target-in-dev.md
@@ -1,7 +1,7 @@
 ---
 tauri-build: patch:enhance
 tauri-cli: patch:enhance
-@tauri-apps/clo: patch:enhance
+@tauri-apps/cli: patch:enhance
 ---
 
 Tauri now ignores `macOS.minimumSystemVersion` in `tauri dev` to prevent forced rebuilds of macOS specific dependencies when using something like `rust-analyzer` at the same time as `tauri dev`.
--- a/.changes/skip-deployment-target-in-dev.md
+++ b/.changes/skip-deployment-target-in-dev.md
@@ -1,0 +1,7 @@
+---
+tauri-build: patch:enhance
+tauri-cli: patch:enhance
+@tauri-apps/clo: patch:enhance
+---
+
+Tauri now ignores `macOS.minimumSystemVersion` in `tauri dev` to prevent forced rebuilds of macOS specific dependencies when using something like `rust-analyzer` at the same time as `tauri dev`.
--- a/crates/tauri-build/src/lib.rs
+++ b/crates/tauri-build/src/lib.rs
@@ -573,8 +573,10 @@ pub fn try_build(attributes: Attributes) -> Result<()> {
       }
     }
 
-    if let Some(version) = &config.bundle.macos.minimum_system_version {
-      println!("cargo:rustc-env=MACOSX_DEPLOYMENT_TARGET={version}");
+    if !is_dev() {
+      if let Some(version) = &config.bundle.macos.minimum_system_version {
+        println!("cargo:rustc-env=MACOSX_DEPLOYMENT_TARGET={version}");
+      }
     }
   }
 
--- a/crates/tauri-cli/config.schema.json
+++ b/crates/tauri-cli/config.schema.json
@@ -3483,7 +3483,7 @@
           ]
         },
         "minimumSystemVersion": {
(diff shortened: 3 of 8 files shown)


lucasfernog:
in the large projects i've built i always set the env var in the .cargo/config.toml file, but this is a very good idea
i think we should do both - so local builds also do not reset the cache

.changes/skip-deployment-target-in-dev.md
```diff
@@ -0,0 +1,7 @@
+---
+tauri-build: patch:enhance
+tauri-cli: patch:enhance
+@tauri-apps/clo: patch:enhance
```
lucasfernog:
```suggestion
@tauri-apps/cli: patch:enhance
```

FabianLars: lol

.changes/skip-deployment-target-in-dev.md
```diff
@@ -0,0 +1,7 @@
+---
+tauri-build: patch:enhance
+tauri-cli: patch:enhance
+@tauri-apps/cli: patch:enhance
```
lucasfernog:
```suggestion
'@tauri-apps/cli': patch:enhance
```

lucasfernog (approved):

FabianLars:
> i think we should do both - so local builds also do not reset the cache

we can't do both, that would overwrite this PR

FabianLars:
unless you mean, this PR for those who didn't set the env and the env for the new projects, that would indeed make sense. sorry i should go to bed x)

lucasfernog:
> unless you mean, this PR for those who didn't set the env and the env for the new projects

that's right

JeffTsang:
This PR breaks macos apps that use swift-rs. It seems that swift-rs needs the deployment target to be set.

FabianLars:
Hmm, in my opinion this isn't our responsibility. Other crates like cc-rs / cmake-rs whatever have fallback values. I would have expected it to work already since you need to set the version in `SwiftLinker::new`. I guess this is the relevant issue on their side: https://github.com/Brendonovich/swift-rs/issues/68

JeffTsang:
Given that swift-rs is the only way to run swift code on macos, it would be nice to maintain compatibility until the deployment target issues are fixed on their side or until tauri provides another way to run swift code on macos. For example, would you consider adding a setting to tauri.conf.json to set the macos deployment target in dev?

FabianLars:
Nah a setting doesn't make sense, at that point I'd rather add it back and tell people to add a cargo env if they have recompilation issues.

Is it enough to insert the cargo instruction in your app's build.rs file or does swift-rs need the env var set by the tauri cli? (This PR disabled both).

JeffTsang:
Inserting the cargo instruction in build.rs works.

PR #14089, opened 2025-08-26
feat: Expose ScrollBarStyle webview option to tauri.

Themayu:
This commit exposes the `scroll_bar_style` option from wry via tauri's `WebviewBuilder` and `WebviewWindowBuilder` APIs, as well as the configuration and JavaScript APIs.

A continuation of https://github.com/tauri-apps/wry/issues/1339, to expose this option to tauri.


Themayu:
I'm not going to bother with fixing the formatting until the PR itself is ready for review.

Themayu:
While testing this, I discovered that on Windows, attempting to create a webview from the JavaScript API will cause the hosting window to immediately hang if it attempts to use the same data directory as another webview without sharing the same environment options. The PR includes no mitigation for this behaviour, as I was told to continue as if I hadn't noticed it, and open a separate issue about it.

This extends to the configuration file and the Rust API as well, though the Rust API allows for changing the data directory. This behaviour is documented on Microsoft's website under [CoreWebView2Environment.CreateCoreWebView2ControllerAsync](<https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2environment?view=webview2-winrt-1.0.3296.44#createcorewebview2controllerasync-1>):
> WebView creation fails if a running instance using the same user data folder exists, and the Environment objects have different `CoreWebView2EnvironmentOptions`. For example, if a WebView was created with one `CoreWebView2EnvironmentOptions.Language`, an attempt to create a WebView with a different `CoreWebView2EnvironmentOptions.Language` using the same user data folder fails.

Themayu:
Oh. I forgot the change file. My bad, adding that now.

Themayu:
Sorry about that, I just realised that the multiple rebases I did during this deleted some of my commit signatures. I've had to redo all of them.

Themayu:
Removed `tauri-schema-generator` from the change files, as per covector error message during last run.

Themayu:
One final push to change the line endings in my change files to LF (I blame git for it being CRLF, as I have git configured to change all line endings to LF on commit.) Decided to amend it to the previous commit instead of creating a new one specifically for it, so I force pushed.

lucasfernog:
> While testing this, I discovered that on Windows, attempting to create a webview from the JavaScript API will cause the hosting window to immediately hang if it attempts to use the same data directory as another webview without sharing the same environment options. The PR includes no mitigation for this behaviour, as I was told to continue as if I hadn't noticed it, and open a separate issue about it.
> 
> This extends to the configuration file and the Rust API as well, though the Rust API allows for changing the data directory. This behaviour is documented on Microsoft's website under [CoreWebView2Environment.CreateCoreWebView2ControllerAsync](https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/winrt/microsoft_web_webview2_core/corewebview2environment?view=webview2-winrt-1.0.3296.44#createcorewebview2controllerasync-1):
> 
> > WebView creation fails if a running instance using the same user data folder exists, and the Environment objects have different `CoreWebView2EnvironmentOptions`. For example, if a WebView was created with one `CoreWebView2EnvironmentOptions.Language`, an attempt to create a WebView with a different `CoreWebView2EnvironmentOptions.Language` using the same user data folder fails.

nice catch.. we do have options to force a webview to use the same environment a previously created one - i think we just need to track all data directories and use it accordingly

lucasfernog (dismissed):
well done!

PR #14093, opened 2025-08-26
feat(cli): ensure mobile Rust targets are installed

lucasfernog:
currently we only install Rust targets for mobile when initializing the projects.

Users can commit the Android/iOS projects, so running the init command is not a requirement to develop mobile apps. This change ensures the Rust targets are installed before trying to compile them.

### DIFF
--- a/.changes/ensure-targets-mobile.md
+++ b/.changes/ensure-targets-mobile.md
@@ -1,0 +1,6 @@
+---
+"@tauri-apps/cli": patch:enhance
+"tauri-cli": patch:enhance
+---
+
+Ensure Rust targets for mobile are installed when running the dev and build commands (previously only checked on init).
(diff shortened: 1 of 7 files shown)


FabianLars (approved):
i feel like Tony will not be happy about this or maybe not because mobile dev is already really slow to start anyway

(untested)

Legend-Master:
Probably ok to me with just this amount of slow down 😂

```
   4        0.104 rustup target list --installed
   5        0.041 Get-History
   6        0.068 rustc --print sysroot
```

lucasfernog:
well we print the sysroot, then read it to list all installed targets :|

lucasfernog:
DX costs a few seconds 😂 there's still so much we can/must do to ensure a good experience and handle errors etc

PR #14094, opened 2025-08-26
feat(cli): detect Android env and install SDK and NDK if needed

lucasfernog:
changes the Android setup to be a bit more automated - looking up ANDROID_HOME and NDK_HOME from common system paths and installing the Android SDK and NDK if needed using the command line tools

### DIFF
--- a/.changes/ensure-android-env.md
+++ b/.changes/ensure-android-env.md
@@ -1,0 +1,6 @@
+---
+"tauri-cli": minor:feat
+"@tauri-apps/cli": minor:feat
+---
+
+Try to detect ANDROID_HOME and NDK_HOME environment variables from default system locations and install them if needed using the Android Studio command line tools.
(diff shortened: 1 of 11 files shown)


FabianLars:
I feel like this is something where sooner or later people will ask us for an opt-out mechanism but we'll see what kind of feedback we'll get, cool change either way

Legend-Master:
I feel like we should ask in the cli first before downloading, these packages are quite big and they are stored in the android studio paths (are not in tauri's usual temp directory)

amrbashir:
I agree with @Legend-Master, should be a prompt (defaults to No), and in CI always No. 

We should also output a list of what changes have we made to their system

lucasfernog:
good idea, i'll also link to the docs in case they want to do the process manually

Legend-Master (dismissed):

Legend-Master (approved):

Usmanslabiu:
Nice one

PR #14103, opened 2025-08-28
fix: add ECMAScript Explicit Resource Management to Resource

ddosakura:
Add [ECMAScript Explicit Resource Management](https://github.com/tc39/proposal-explicit-resource-management) to Resource.

This makes the [await using](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/await_using) work:

```ts
import { create, BaseDirectory } from "@tauri-apps/plugin-fs"
...
{
  await using file = await create("foo/bar.txt", { baseDir: BaseDirectory.AppConfig });
  await file.write(new TextEncoder().encode("Hello world"));
  // Before `file` goes out of scope, it is disposed by calling `file[Symbol.asyncDispose]()` and awaited.
}
```


Legend-Master:
Thanks for bring this up!

I have one question though, I believe `Symbol.asyncDispose` doesn't exist on older browsers, should we provide a polyfill here?

> `Symbol.asyncDispose ??= Symbol("Symbol.asyncDispose");`
> https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#using-declarations-and-explicit-resource-management:~:text=Symbol.asyncDispose%20%3F%3F%3D%20Symbol(%22Symbol.asyncDispose%22)%3B

cc @lucasfernog @FabianLars

FabianLars:
Not only older browsers but webkit as a whole as well which means this is a Windows only change.

Legend-Master:
> but webkit as a whole

Technically it's a coming soon https://webkit.org/blog/16993/news-from-wwdc25-web-technology-coming-this-fall-in-safari-26-beta/ but yeah, this is still useful for people using the transpiler/polyfill themselves though

FabianLars:
hmm true but having macos 26 as a minimum version sounds a bit problematic (and who knows when/if linux follows)

Legend-Master (approved):
Thanks and sorry for the long waits!

I think we should also looking into provide a guide on how to use this awesome feature! The symbol polyfill should not be needed in the tauri lib code, it stays `undefined` in unsupported browsers if you don't use `using` at all. If you do want to opt-in, use a compiler (e.g. tsc 5.2+)/bundler (e.g. rollup) that supports that syntax + defining

```js
Symbol.dispose ??= Symbol("Symbol.dispose");
Symbol.asyncDispose ??= Symbol("Symbol.asyncDispose");
```

to support old browsers.

PR #14108, opened 2025-08-28
feat(core): add config for Info.plist extensions, closes #13667

### LINKED ISSUE #13667: [feat] Different info.plist files for staging and production builds
### Describe the problem

I have a `tauri.conf.json` and a `tauri-staging.conf.json` which overwrites some of the values in the main config for staging builds.

The `Info.plist` file needs to have different activity values for staging and production.

At the moment it's not possible to set different `Info.plist` files for these different configs.

In the same way it's possible to set the path of the `entitements.plist` file, make it possible to set the path of the `info.plist` file in the Tauri config file.

### Alternatives considered

_No response_

### Additional context

_No response_

lucasfernog:


crates/tauri-utils/src/config.rs
```diff
@@ -664,6 +664,11 @@ pub struct MacConfig {
   pub provider_short_name: Option<String>,
   /// Path to the entitlements file.
   pub entitlements: Option<String>,
+  /// Path to a Info.plist file to merge with the default Info.plist.
+  ///
+  /// Note that Tauri also looks for a `Info.plist` and `Info.macos.plist` file in the same directory as the Tauri configuration file.
```
FabianLars: Afaik this is a lie, i don't think info.macos.plist is a thing yet (it should though). The PR itself works fine though.

PR #14109, opened 2025-08-28
feat(tauri-runtime-wry): ensure webview2 environment options matches

lucasfernog:
on Windows the Webview2 environment options must match when created webviews sharing the same data directory (unless specified, all webviews share the same data directory). This PR validates the options, warning if they do not match, and forces the webview to be created with the same previous environment to prevent a crash.

discovered as part of #14089

### DIFF
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -4580,7 +4580,6 @@ You may have it installed on another user account, but it is not available for t
   let web_context = match entry {
     Occupied(occupied) => {
       let occupied = occupied.into_mut();
-      occupied.referenced_by_webviews.insert(label.clone());
 
       #[cfg(windows)]
       if cfg!(debug_assertions) && occupied.environment_options != environment_options {
@@ -4591,6 +4590,8 @@ You may have it installed on another user account, but it is not available for t
         #[cfg(not(feature = "tracing"))]
         eprintln!("{message}");
       }
+
+      occupied.referenced_by_webviews.insert(label.clone());
 
       occupied
     }
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -4583,8 +4583,11 @@ You may have it installed on another user account, but it is not available for t
 
       #[cfg(windows)]
       if cfg!(debug_assertions) && occupied.environment_options != environment_options {
-        let message =
-          format!("environment options {environment_options:?} for {web_context_key:?} do not match previously defined options for the same data directory on webviews {:?}, expected {:?}. Since this results in a crash, we are going to force the previously defined environment options", occupied.referenced_by_webviews, occupied.environment_options);
+        let message = format!(
+          "environment options {environment_options:?} for {} do not match previously defined options for the same data directory on webviews {:?}, expected {:?}. Since this results in a crash, we are going to force the previously defined environment options",
+          web_context_key.as_ref().map(|p| p.to_string_lossy()).unwrap_or_else(|| "".into()), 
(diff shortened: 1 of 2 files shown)


crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -157,11 +160,24 @@ use window::WindowExt as _;
...
+  #[cfg(windows)]
+  environment_options: WebView2EnvironmentOptions,
   // on Linux the custom protocols are associated with the context
   // and you cannot register a URI scheme more than once
   pub registered_custom_protocols: HashSet<String>,
 }
 
+#[cfg(windows)]
+#[derive(Debug, PartialEq, Eq)]
+struct WebView2EnvironmentOptions {
```
Legend-Master, line 174: To be honest, this feels very error prune to me but I don't know if there's a better way

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -157,11 +160,24 @@ use window::WindowExt as _;
 pub struct WebContext {
   pub inner: WryWebContext,
   pub referenced_by_webviews: HashSet<String>,
+  #[cfg(windows)]
+  environment: Option<ICoreWebView2Environment>,
+  #[cfg(windows)]
+  environment_options: WebView2EnvironmentOptions,
```
Legend-Master, line 166: Maybe?

```suggestion
  #[cfg(all(debug_assertions, windows))]
  environment_options: WebView2EnvironmentOptions,
```

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -4584,6 +4626,11 @@ You may have it installed on another user account, but it is not available for t
     .with_clipboard(webview_attributes.clipboard)
     .with_hotkeys_zoom(webview_attributes.zoom_hotkeys_enabled);
 
+  #[cfg(windows)]
+  if let Some(environment) = &web_context.environment {
+    webview_builder = webview_builder.with_environment(environment.clone());
```
Legend-Master, line 4631: The whole environment thing is a bit confusing now, that we have it in `on_new_window` and tell people to pass it in with `with_environment`/`window_features`, but we also have it done automatically here

Some documentation around the related attributes would be nice

Legend-Master (commented):
I think with this, we should update the documentation `webview_attributes.environment` and related function, then mark `NewWindowOpener.environment` as deprecated since we now pass them automatically now and you can't use a different one within the same data directory anyways

crates/tauri-runtime-wry/src/lib.rs
lucasfernog: well this is just for the warning - we ALWAYS force the same environment for the same data directory

crates/tauri-runtime-wry/src/lib.rs
lucasfernog: well i'm not sure if we should be explicit or not. like only validate the environment options matches? though in this case we get to that "error prune" validation issue..

since the on_new_window environment is all handled by window_features i think it's ok to keep it as is, so it's explicitly matched instead of relying on internal details

but i agree documentation should be improved anyway

Legend-Master: > well i'm not sure if we should be explicit or not. like only validate the environment options matches? though in this case we get to that "error prune" validation issue..

I think forcing the same environment makes sense here, no matter the options, it's too error prune the other way around

> since the on_new_window environment is all handled by window_features i think it's ok to keep it as is, so it's explicitly matched instead of relying on internal details

If we want to force the environment, keeping it feels like some extra confusions, anyways, we can't remove it in v2 anyways, let's just update documentation

Issue #13180, opened 2025-04-08
[bug] "npx run tauri iOS dev" Couldn't recognize the current folder as a Tauri project


bagolomb:
### Describe the bug

Hello
I have run "npx tauri ios init" and "npc run tauri ios dev"
and I keep getting the error below.
 
I have setup the project recently, am on Tauri-cli 2.4 and have confirmed I have a tauri.conf.json in src-tauri and it's not gitignored.

@FabianLars Suggested I add this to a GitHub issue

All help appreciated. very lost as to how to solve this

```
" export variant\=normal
    /bin/sh -c /Users/bradygolomb/Library/Developer/Xcode/DerivedData/OrbitalNotes-cspiuqizwmtpyxdnwwprxuvekeqg/Build/Intermediates.noindex/OrbitalNotes.build/debug-iphonesimulator/OrbitalNotes_iOS.build/Script-619A52B0E784687FF827AF4D.sh

thread '<unnamed>' panicked at crates/tauri-cli/src/helpers/app_paths.rs:136:5:
Couldn't recognize the current folder as a Tauri project. It must contain a `tauri.conf.json`, `tauri.conf.json5` or `Tauri.toml` file in any subfolder.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
/Users/bradygolomb/Library/Developer/Xcode/DerivedData/OrbitalNotes-cspiuqizwmtpyxdnwwprxuvekeqg/Build/Intermediates.noindex/OrbitalNotes.build/debug-iphonesimulator/OrbitalNotes_iOS.build/Script-619A52B0E784687FF827AF4D.sh: line 2: 99071 Abort trap: 6           npx tauri ios xcode-script -v --platform ${PLATFORM_DISPLAY_NAME:?} --sdk-root ${SDKROOT:?} --framework-search-paths "${FRAMEWORK_SEARCH_PATHS:?}" --header-search-paths "${HEADER_SEARCH_PATHS:?}" --gcc-preprocessor-definitions "${GCC_PREPROCESSOR_DEFINITIONS:-}" --configuration ${CONFIGURATION:?} ${FORCE_COLOR} ${ARCHS:?}
Command PhaseScriptExecution failed with a nonzero exit code

note: Run script build phase 'Build Rust Code' will be run during every build because the option to run the script phase "Based on dependency analysis" is unchecked. (in target 'OrbitalNotes_iOS' from project 'OrbitalNotes')
** BUILD FAILED **


[...]

PR #14118, opened 2025-08-29
fix(cli): mobile init when using pnpm dlx

lucasfernog:
i noticed this when testing #13180 (though the original issue refers to npx, which I could not reproduce yet)


FabianLars:
`pnpx` still exists as a (deprecated?) alias for `pnpm dlx`. Does that matter here?

lucasfernog:
> `pnpx` still exists as a (deprecated?) alias for `pnpm dlx`. Does that matter here?

in that case the generated project will use `pnpm dlx @tauri-apps/cli ...` to run the CLI too (same as when running `pnpm dlx ..`

lucasfernog:
to test this, you need to run `pnpx file:/absolute/path/to/tauri/packages/cli android init` (or `pnpm dlx` instead of `pnpx`)
it'll fail to load the .node because it is gitignored, so you'll need to copy `packages/cli/cli.linux-x64-gnu.node` to `$HOME/.cache/pnpm/dlx/<hash>/pkg/node_modules/@tauri-apps/cli` where the hash is printed in the error message of the pnpm command run

FabianLars (approved):

Issue #14075, opened 2025-08-24
[bug] IOS log are too verbose


saurL:
### Describe the bug

On IOS when using `tauri_plugin_log`  logs are too verbose and it's not possible to see usefull logs because of it.
According to @FabianLars it is not the expected behaviour

<img width="1277" height="808" alt="Image" src="https://github.com/user-attachments/assets/c0734d56-3a71-4ab8-b01c-21445f981af1" />

One workaround is this add `.level(log::LevelFilter::Info)`

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 15.5.0 arm64 (X64)
    ✔ Xcode Command Line Tools: installed
    ✔ rustc: 1.89.0 (29483883e 2025-08-04)
    ✔ cargo: 1.89.0 (c24e10642 2025-06-23)
    ✔ rustup: 1.28.2 (e4f3ad6f8 2025-04-28)
    ✔ Rust toolchain: stable-aarch64-apple-darwin (default)
    - node: 22.18.0
    - pnpm: 10.14.0
    - npm: 10.9.3

[-] Packages
    - tauri 🦀: git+https://github.com/saurL/tauri?branch=feat%2Fpush-notifications#75edc968c75664ccaa9a9c5c29943704008af8c9 (2.8.2)
    - tauri-build 🦀: No version detected
    - wry 🦀: 0.53.1
    - tao 🦀: git+https://github.com/saurL/tao?branch=feat%2Fpush-notification#a62638b2af5b45dcca2af682ea469ebff90f8396 (0.34.2)
    - @tauri-apps/api : 2.8.0
    - @tauri-apps/cli : 2.8.1

[-] Plugins
    - tauri-plugin-log 🦀: 2.6.0
    - @tauri-apps/plugin-log : 2.6.0
    - tauri-plugin-fs 🦀: 2.4.2
    - @tauri-apps/plugin-fs : not installed!
    - tauri-plugin-deep-link 🦀: 2.4.1, (outdated, latest: 2.4.2)
    - @tauri-apps/plugin-deep-link : not installed!
    - tauri-plugin-http 🦀: 2.5.2
    - @tauri-apps/plugin-http : not installed!
    - tauri-plugin-opener 🦀: 2.2.3, (outdated, latest: 2.5.0)
    - @tauri-apps/plugin-opener : not installed!

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
[...]

PR #14122, opened 2025-08-29
feat(cli): set default log level when adding the log plugin

lucasfernog:
needs https://github.com/tauri-apps/plugins-workspace/pull/2965

ref #14075

### DIFF
--- a/crates/tauri-cli/src/add.rs
+++ b/crates/tauri-cli/src/add.rs
@@ -132,6 +132,8 @@ pub fn run(options: Options) -> Result<()> {
     "Builder::new(todo!()).build()"
   } else if plugin == "single-instance" {
     "init(|app, args, cwd| {})"
+  } else if plugin == "log" {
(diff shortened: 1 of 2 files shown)


Legend-Master (dismissed):
I think `Info` is probably a better default?

crates/tauri-cli/src/add.rs
```diff
@@ -132,6 +132,8 @@ pub fn run(options: Options) -> Result<()> {
     "Builder::new(todo!()).build()"
   } else if plugin == "single-instance" {
     "init(|app, args, cwd| {})"
+  } else if plugin == "log" {
+    "Builder::new().level(tauri_plugin_log::log::LevelFilter::Debug).build()"
```
lucasfernog:
```suggestion
    "Builder::new().level(tauri_plugin_log::log::LevelFilter::Info).build()"
```

Legend-Master (approved):

PR #14126, opened 2025-08-31
refactor(cli): improve errors

lucasfernog:
I know this is controversial but I've had a weird "directory not empty" error when building iOS apps that seems to be related to the `gen/apple/build` folder, but I had to guess it and it's impossible to trace it.. so I decided to drop `anyhow` to force us to be more explicit about what happened when something goes wrong.

This PR:

- drops anyhow
- uses thiserror instead, with variants including more context about what actually failed
- keeps the anyhow::Context thingy because that's so cool
- doesn't try to do too much, keeping some tech debt specially in the bundler error type
- it's so damn verbose :( and it took away 4 hours of my life


lucasfernog:
pretty sure CI will fail here, I only did the macOS part so far

lucasfernog:
will be paired with https://github.com/tauri-apps/cargo-mobile2/pull/485 on release

Legend-Master:
Didn't look at the whole thing yet, but I think if we want to actually enforce this, we need a helper function like `context` for the fs errors or it will just be too annoying that nobody actually wants to use it 😂

lucasfernog:
> Didn't look at the whole thing yet, but I think if we want to actually enforce this, we need a helper function like `context` for the fs errors or it will just be too annoying that nobody actually wants to use it 😂

that's a good idea, i wanted to do a second pass on it.. i'm running some prompts now

crates/tauri-bundler/src/error.rs
```diff
@@ -3,19 +3,41 @@
...
+    command: String,
+    /// Error that occurred.
+    error: io::Error,
+  },
   /// Error running tauri_utils API.
   #[error("{0}")]
   Resource(#[from] tauri_utils::Error),
-  /// Bundler error.
-  #[error("{0:#}")]
-  BundlerError(#[from] anyhow::Error),
```
FabianLars, line 18: Can we revert this and add a todo and change the docs saying that this variant is unused now? There are projects using tauri-bundler and this is a breaking api change.

FabianLars: otherwise this looks good i think (at least for a start). couldn't test it _that_ much though cause i can't think of many ways to trigger errors on windows. i guess time will tell how good of a start this is :D

FabianLars (dismissed):

crates/tauri-cli/src/error.rs
```diff
@@ -0,0 +1,257 @@
+// Copyright 2019-2024 Tauri Programme within The Commons Conservancy
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: MIT
+
+use std::{borrow::Cow, fmt::Display, path::PathBuf};
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
```
Legend-Master, line 8: To be honest, I'm a bit concerned about so many errors, this could be quite hard to follow through and find the right error to propagate, and in most cases, CLI errors are just some plain texts explaining what's going on

lucasfernog: most of them are using #[from] except the ones that have context

alternatively we can remove most of the variants here and enforce usage of the context trait (which was the idea here tbh, adding context to every single error out there, but it's impossible to do so using the anyhow crate because it doesn't enforce it, so i had to break compilation for the entire codebase to spot the places that needed it)

Legend-Master: > alternatively we can remove most of the variants here and enforce usage of the context trait (which was the idea here tbh, adding context to every single error out there, but it's impossible to do so using the anyhow crate because it doesn't enforce it, so i had to break compilation for the entire codebase to spot the places that needed it)

That sounds much better to me, enforcing an error context is much more useful than having an error saying it is a `JsonError` (not sure how annoying would that be when writing staffs though 😂, we could try that out anyways)

lucasfernog: pushed some further cleanup; i'd say we should make additional changes in a separate PR :joy:

Legend-Master: Looks much better, we could maybe add a few variant and helps like `fs_context` for some commonly used ones in the future

Legend-Master (dismissed):
Not entirely sure how I would test this (e.g. making something fail), code and design look good though

lucasfernog:
easiest thing you can do is get something to fail :joy: break cargo.toml / tauri.conf.json for instance

Issue #12486, opened 2025-01-23
[bug] use register_listener instead of registerListener.


GlosZhao:
### Describe the bug

When using addPluginListener, I found that registerListener is not allowed. Upon further investigation, I discovered that addPluginListener internally uses registerListener. I believe this does not conform to Rust's naming conventions, and it should be corrected to the previous register_listener.

### Reproduction

https://github.com/tauri-apps/tauri/blob/f5a59b93bfefb43ff131a7870b3c5d5e48c1ca1e/packages/api/src/core.ts#L165

### Expected behavior

use register_listener instead of registerListener.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 15.1.1 arm64 (X64)
    ✔ Xcode Command Line Tools: installed
    ✔ rustc: 1.83.0 (90b35a623 2024-11-26) (Homebrew)
    ✔ cargo: 1.83.0
    ✔ rustup: 1.27.1 (2024-04-24)
    ✔ Rust toolchain: 
    - node: 18.20.2
    - pnpm: 9.10.0
    - npm: 10.5.0

[-] Packages
    - tauri 🦀: 2.1.1
    - tauri-build 🦀: 2.0.3
    - wry 🦀: 0.47.2
    - tao 🦀: 0.30.8
    - @tauri-apps/api : 2.0.0 (outdated, latest: 2.2.0)
    - @tauri-apps/cli : 2.0.3 (outdated, latest: 2.2.5)

[-] Plugins
    - tauri-plugin-shell 🦀: 2.0.2
    - @tauri-apps/plugin-shell : 2.0.0 (outdated, latest: 2.2.0)
    - tauri-plugin-http 🦀: 2.0.3
    - @tauri-apps/plugin-http : 2.0.0 (outdated, latest: 2.2.0)
    - tauri-plugin-os 🦀: 2.0.1
    - @tauri-apps/plugin-os : 2.0.0 (outdated, latest: 2.2.0)
    - tauri-plugin-fs 🦀: 2.0.3
    - @tauri-apps/plugin-fs : not installed!

[-] App
    - build-type: bundle
    - CSP:  xxxx
    - frontendDist: ../dist
    - devUrl: http://localhost:1420/
    - framework: Vue.js
    - bundler: Vite
```

### Stack trace

```text

```

### Additional context

I am not clear about the rationale behind this change. I would appreciate it if I could receive a response explaining the reasoning.

FabianLars:
[...]

PR #14132, opened 2025-08-31
fix(api): "command not found" error when running addPluginListener

lucasfernog:
the backend expects the command name to be in snake case

we've made this change already for check_permissions and request_permissions, but missed register_listener


FabianLars:
No we did not miss it, we reverted it because for some unknown reason snake case didn't work: https://github.com/tauri-apps/tauri/pull/11423

lucasfernog:
oh ok i'll need to dig deeper

FabianLars:
good luck 😅 i hope you can find it. i couldn't make sense of it. especially since it appears to be a bit inconsistent at times.

lucasfernog:
well like someone once said, the whole permission validation is a bit.. messy, crazy flow

crates/tauri/src/webview/mod.rs
```diff
@@ -1715,14 +1715,34 @@ tauri::Builder::default()
...
-      );
+      let acl = runtime_authority
+        .resolve_access(
+          &request.cmd,
+          &window_label,
+          message.webview.label(),
+          &acl_origin,
+        )
+        .or_else(|| {
+          if request.cmd.chars().any(|c| c.is_uppercase()) {
```
Legend-Master: I honestly don't like this, it means we now match the commands in different casing without formal documentation, and that's confusing, I'd rather use a try catch call in the `addPluginListener` function

Legend-Master (commented):
Correct me if I'm wrong

- On desktop, we match the ACL and the command by the exact command names
- On mobile, we still use the exact command names when matching ACLs, but we transform the command calls to camelCase

https://github.com/tauri-apps/tauri/blob/f5851ee00d6d1f4d560a220ca5a728fedd525092/crates/tauri/src/webview/mod.rs#L1870

And from the current casing conventions, `registerListener` seems to be an outlier? I know we can't change that in v2 or we break existing ones that uses the camelCase though

crates/tauri/src/webview/mod.rs
lucasfernog: maybe we just have a particular case for registerListener then? i don't really want to make two calls for a single command :/

lucasfernog: well i'll just go with your idea.. better than bloating the rust side with a lot of temp logic :(

chrox:
We'd better reference [tauri-apps/tauri#12486](https://github.com/tauri-apps/tauri/issues/12486) in this PR to inform developers whose products currently rely on `registerListener`, so they can update their code accordingly and avoid potential breaking changes.

This is the error message in Tauri 2.9.* if we only allow `registerListener` as in Tauri 2.8.* and below:
```
Failed to setup TTS event listener: native-tts.register_listener not allowed. Command not found
setupEventListener | @ | page-ae56bef211a9a7b4.js:1
```

PR #14133, opened 2025-08-31
feat(core): back button event on Android, closes #8142

### LINKED ISSUE #8142: [feat] Custom back button/gesture behavior on Android
### Describe the problem

Currently, on Tauri v2, if the back button or gesture is used on Android, and the app is a single page app (e.g., manages everything on the same page without managing custom history), the app just closes.

I would like to be able to implement custom behavior for the Android back button/gesture. This could possibly be done through a global event that can be listened for, but there should also be a way to disable the default behavior for it. There could also be a way to register a handler while initializing Tauri.

### Alternatives considered

This could also be done by giving Tauri a way to manage the custom browser history, and telling the frontend or backend when the back/forward browser events are triggered. This could also be managed through the developer's frontend framework of choice, but it might require a refactor if the app framework isn't already using custom routing or separate browser pages.

### Additional context

[The Discord support thread that prompted this feature request](https://discord.com/channels/616186924390023171/1167318179043868772)

lucasfernog:
I've used https://github.com/ionic-team/capacitor-plugins/blob/main/app/android/src/main/java/com/capacitorjs/plugins/app/AppPlugin.java as a reference here, checking if there's a back button event handler with a default of webview's goBack implementation


lucasfernog:
the exit impl still needs to be tested

lucasfernog:
I've removed the App::exit impl - after further testing, the existing API is working fine

FabianLars:
If i register a callback it seems to work fine but if i don't register one, i get `[TAURI] Couldn't find callback id ...` errors when i use the back button.
Also it likely was because of my scuffed testing but i couldn't really make use of `canGoBack: true` since it just navigates in that case. Any tips how to test this?

p.s. i had to merge #14132 into my local clone of this branch for the listener to work.

lucasfernog:
> If i register a callback it seems to work fine but if i don't register one, i get `[TAURI] Couldn't find callback id ...` errors when i use the back button. Also it likely was because of my scuffed testing but i couldn't really make use of `canGoBack: true` since it just navigates in that case. Any tips how to test this?
> 
> p.s. i had to merge #14132 into my local clone of this branch for the listener to work.

ohh i didn't realize wry handled this.. i'll push an update

FabianLars:
One really weird thing (that i've seen elsewhere as well, just not this reliably) is that i have to manually shutdown the emulator if i want to re-run `tauri android dev` after registering a listener.

Otherwise this seems to work good now though i wonder if someone will ask for a way to remove the listener to restore native behavior at runtime. I assume that'll be rare so i think we can wait until that's actually needed?

FabianLars (dismissed):

lucasfernog:
> Otherwise this seems to work good now though i wonder if someone will ask for a way to remove the listener to restore native behavior at runtime. I assume that'll be rare so i think we can wait until that's actually needed?

onBackButtonPress returns PluginListener which is a class that has a unregister function (I also forgot about this, had to check lol)
it's similar to how listen() works but uses a class instead of a closure :|

lucasfernog:
> One really weird thing (that i've seen elsewhere as well, just not this reliably) is that i have to manually shutdown the emulator if i want to re-run tauri android dev after registering a listener.

I think it's unrelated. I've seen this when i use Android Studio and i click run or debug way too many times - it even breaks Android Studio itself after a while. I think it's related to the adb port forwarding, it gets stuck eventually.

lucasfernog:
> One really weird thing (that i've seen elsewhere as well, just not this reliably) is that i have to manually shutdown the emulator if i want to re-run tauri android dev after registering a listener.

@FabianLars you're right this indeed makes it a lot easy to reproduce this.. but only if you no-op on the back button handler.. if I do this it doesn't seem like it breaks:
`await __TAURI__.app.onBackButtonPress(() => __TAURI__.process.exit())`
maybe the emulator gets in a weird state? i'm not sure..

I've opened a couple PRs to help: https://github.com/tauri-apps/cargo-mobile2/pull/493 https://github.com/tauri-apps/cargo-mobile2/pull/495

PR #14148, opened 2025-09-04
feat(ios): support async Swift plugin methods (`completionHandler:`) in PluginManager

Choochmeque:
This adds support for invoking plugin methods that Swift exposes as async / async throws.
In the Objective-C runtime, such methods are automatically bridged into selectors with a completionHandler: parameter (e.g. commandName:completionHandler:).
	•	Detects and invokes completionHandler: selectors with a block that rejects if an error is passed.
	•	Falls back to the existing error: selector for throwing methods.
	•	No breaking changes — plugins can now use either throws or async/await Swift APIs.

Code example:

```objc
    @objc public func getProducts(_ invoke: Invoke) async throws {
        let args = try invoke.parseArgs(GetProductsArgs.self)
        
        do {
            let products = try await Product.products(for: args.productIds)
            var productsArray: [[String: Any]] = []
            
            // Other code

            invoke.resolve(["products": productsArray])
        } catch {
            invoke.reject("Failed to fetch products: \(error.localizedDescription)")
        }
    }	
```

Without this change, async Swift plugin methods cannot be called and result is "No command getProducts found for plugin"

### DIFF
--- a/.changes/ios-async-support.md
+++ b/.changes/ios-async-support.md
@@ -1,5 +1,5 @@
 ---
-"tauri-ios-api": minor
+"tauri": minor:feat
 ---
 
 Support async Swift plugin methods (`completionHandler:`) in PluginManager
--- a/.changes/ios-async-support.md
+++ b/.changes/ios-async-support.md
@@ -1,0 +1,5 @@
+---
+"tauri-ios-api": minor
+---
+
+Support async Swift plugin methods (`completionHandler:`) in PluginManager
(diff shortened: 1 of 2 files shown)


Choochmeque:
@FabianLars, could you please review this PR?

FabianLars:
I'm out of office for the next 2-3 so no promises that I can squeeze it in in that timeframe.

Choochmeque:
@FabianLars No worries at all.
I have a workaround in place for now:

```
    @objc public func getProducts(_ invoke: Invoke) throws {
        Task {
            try await self.getProducts(invoke)
        }
    }
```

lucasfernog (dismissed):
ohh nice catch!

PR #14194, opened 2025-09-17
feat(android): add auto_increment_version_code option for Android builds 

### LINKED ISSUE #14135: feat(android): add auto_increment_version_code option for Android builds
### Describe the problem

# 📌 Feature Request: Add `auto_increment_version_code` option for Android builds

## 📝 Problem

Currently, `versionCode` for Android is written into `tauri.properties` during the build process:

- If `bundle.android.version_code` is set in `tauri.conf.json`, it is always used.  
- Otherwise, Tauri derives `versionCode` deterministically from the semver version (`major*1000000 + minor*1000 + patch`).  

This approach has two main issues:

1. **Google Play requires a strictly increasing integer** for `versionCode`. Repeated builds of the same app version produce the same code, which will be rejected.  
2. Developers must manually bump `bundle.android.version_code` for every release, which is inconvenient for some cases.

## 💡 Proposed Solution

Introduce a new config option under `bundle.android`:

```jsonc
"bundle": {
  "android": {
    "versionCode": 2,                 // manual override (existing)
    "autoIncrementVersionCode": false  // NEW option
  }
}
```

- `false` (default): keep current behavior (manual or semver-derived).  
- `true`: automatically increment `versionCode` on each build by:
  - reading the previous value from `tauri.properties`,
  - incrementing it by `+1`,
  - writing it back for the next build.

------

## ✅ Behavior Summary

| Config setting | Result |
|---|---|
| `bundle.android.versionCode = 123` | Always `123` (explicit override wins) |
| `autoIncrementVersionCode = true` | Increments `versionCode` each build (`1 → 2 → 3 …`) |
| Neither set | Fallback: semver-derived (`major*1000000 + minor*1000 + patch`) |

---

## 🔧 Example

**`tauri.conf.json`**
```json
"bundle": {
  "android": {
    "autoIncrementVersionCode": true
  }
}
```

**Build 1 → `tauri.properties`**
```properties
tauri.android.versionName=1.0.0
tauri.android.versionCode=1
```

**Build 2 → `tauri.properties`**
```properties
tauri.android.versionName=1.0.0
tauri.android.versionCode=2
```

### Alternatives considered

_No response_

### Additional context

_No response_

bipsBro:
closes #14135


lucasfernog (approved):
well done! I just had to move the logic to the CLI because the CLI runs `cargo build` once per target, so it would get crazy versionCode bumps :)

PR #14199, opened 2025-09-20
fix: macOS automatic help menu detection

### LINKED ISSUE #13605: [bug] Window and Help menu are not working on MacOS
### Describe the bug

- When setting up the `Window` or `Help` menu on macOS the default menu items are missing -> For the `Help` menu that would be the search bar and for the the `Window` menu that would be the `Move & Resize` and `Full Screen Tile` items in the menu.
- Also, the documentation says that the `Help` menu should be recognized automatically when named `Help`, but it is not.
    - https://v2.tauri.app/reference/javascript/api/namespacemenu/#setashelpmenufornsapp
- Related but from the JS side: https://github.com/tauri-apps/tauri/issues/12652

### Reproduction

This is how I setup the menu:

```rust
    let report_issue = MenuItem::with_id(
        app,
        "report-issue",
        "Report Issue...",
        true,
        Some("Ctrl+F"),
    )?;

    let help_submenu = SubmenuBuilder::new(app, "Help") // also does not work when setting with tauri::menu::HELP_SUBMENU_ID 
        .item(&report_issue)
        .build()?;

    let window = SubmenuBuilder::with_id(app, WINDOW_SUBMENU_ID, "Window")
        .build()?; // No Move & Resize and Full Screen Tile menu items (does not work)

    let menu = MenuBuilder::new(app)
        .items(&[&help_submenu, &window])
        .build()?;

    Ok(menu)

```

### Full `tauri info` output

```text
[✔] Environment
    - node: 22.14.0
    - npm: 11.3.0
    - bun: 1.2.15

[-] Packages
    - tauri 🦀: 2.5.1
    - tauri-build 🦀: 2.2.0
    - wry 🦀: 0.51.2
    - tao 🦀: 0.33.0
    - @tauri-apps/api : 2.5.0
    - @tauri-apps/cli : 2.5.0

[-] Plugins
    - tauri-plugin-opener 🦀: 2.2.7
    - @tauri-apps/plugin-opener : 2.2.7
    - tauri-plugin-dialog 🦀: 2.2.2
    - @tauri-apps/plugin-dialog : 2.2.2
    - tauri-plugin-fs 🦀: 2.3.0

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - framework: Vue.js
    - bundler: Vite
```

probablykasper:
Fixes #13605

By default, macOS auto-detects the `Help` submenu and adds a search bar to it. But since muda's `set_as_help_menu_for_nsapp` is broken (https://github.com/tauri-apps/muda/issues/263), the search bar would never show at for any app ever - So I've disabled that code for the time being.


crates/tauri/src/menu/menu.rs
```diff
@@ -177,10 +177,11 @@ impl<R: Runtime> Menu<R> {
       HELP_SUBMENU_ID,
       "Help",
       true,
-      &[
-        #[cfg(not(target_os = "macos"))]
-        &PredefinedMenuItem::about(app_handle, None, Some(about_metadata))?,
-      ],
+      &[&PredefinedMenuItem::about(
```
FabianLars, line 180: On macOS the About stuff is already in the first menu item (the one with the appname as label)

probablykasper: I didn't change anything there. I assume it's there kinda just to have something in there (looks weird otherwise)

probablykasper: Oh nvm I'm stupid, I did change it. Well yeah, I think it's best to have _something_ there

probablykasper: Electron's default menubar has Electron's own links in there, that's an option

image

FabianLars: i agree that it looks weird without anything (at least as far as i can remember, didn't see check for ages)

FabianLars: It actually doesn't look _that_ weird on macos 26 anymore with all that added padding. Can we perhaps still revert this part? I'm completely on board with the hotfix of the actual issue so i'd like to get it in asap but i think what we put into the help menu needs a bit more thinking since right now it's only added for visuals.

> Electron's default menubar has Electron's own links in there, that's an option

Actually not a fan of that at all since it basically forces everyone to create their own menu even if they want just the default (cause they actually don't want a menu at all in the context of windows/linux)

crates/tauri/src/menu/plugin.rs
```diff
@@ -824,12 +824,13 @@ fn set_as_help_menu_for_nsapp<R: Runtime>(
...
-    let resources_table = webview.resources_table();
-    let submenu = resources_table.get::<Submenu<R>>(rid)?;
-    submenu.set_as_help_menu_for_nsapp()?;
-  }
+  // Broken: https://github.com/tauri-apps/muda/issues/263
+  // #[cfg(target_os = "macos")]
+  // {
+  //   let resources_table = webview.resources_table();
+  //   let submenu = resources_table.get::<Submenu<R>>(rid)?;
+  //   submenu.set_as_help_menu_for_nsapp()?;
```
lucasfernog, line 832: we're also calling it incorrectly here: https://github.com/tauri-apps/tauri/blob/4188ffdafcaa7c3c8b000973146c9384a873e741/crates/tauri/src/menu/plugin.rs#L814

amrbashir:
fixed upstream in https://github.com/tauri-apps/muda/pull/335

PR #14209, opened 2025-09-23
FIX: Fixed GDI object leak when a resizable window is created and then closed on Windows platform

gribunin:
When a window is created in a Tuari application with, for example, call to new WebviewWindow from the Javascript code and later is closed -- exactly 1 GDI object leaks. This happens because of the function set_drag_hwnd_rgn where 2 regions are created, one of them becomes managed by Windows  after call to SetWindowRgn but another one remains dangling. Added correct release of the region handle after it is not used anymore.

### DIFF
--- a/crates/tauri-runtime-wry/src/undecorated_resizing.rs
+++ b/crates/tauri-runtime-wry/src/undecorated_resizing.rs
@@ -430,6 +430,7 @@ mod windows {
 
     CombineRgn(Some(hrgn1), Some(hrgn1), Some(hrgn2), RGN_DIFF);
 
+    DeleteObject(hrgn2.into());
     SetWindowRgn(hwnd, Some(hrgn1), true);
   }
 
--- a/crates/tauri-runtime-wry/src/undecorated_resizing.rs
+++ b/crates/tauri-runtime-wry/src/undecorated_resizing.rs
@@ -420,18 +420,24 @@ mod windows {
     let border_x = util::get_system_metrics_for_dpi(SM_CXFRAME, dpi);
     let border_y = util::get_system_metrics_for_dpi(SM_CYFRAME, dpi);
 
-    let hrgn1 = CreateRectRgn(0, 0, width, height);
+    // hrgn1 must be mutable to call .free() later
+    let mut hrgn1 = CreateRectRgn(0, 0, width, height);
 
     let x1 = if only_top { 0 } else { border_x };
     let y1 = border_y;
     let x2 = if only_top { width } else { width - border_x };
     let y2 = if only_top { height } else { height - border_y };
-    let hrgn2 = CreateRectRgn(x1, y1, x2, y2);
+    
+    // Wrap hrgn2 in Owned so it is automatically freed when going out of scope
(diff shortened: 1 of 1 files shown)


crates/tauri-runtime-wry/src/undecorated_resizing.rs
```diff
@@ -430,6 +430,7 @@ mod windows {
 
     CombineRgn(Some(hrgn1), Some(hrgn1), Some(hrgn2), RGN_DIFF);
 
+    DeleteObject(hrgn2.into());
```
amrbashir: This is not needed as `HRGN` type [implements the `Free` trait](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Graphics/Gdi/struct.HRGN.html#impl-Free-for-HRGN), which means it automatically calls `DeleteObject` on drop

Are you sure this leaked? If so, can you give me instructions that I can use to reproduce and inspect?

gribunin: It implements Free trait but not Drop trait. As far as I understand that means that there are two patterns in windows-rs:

**Raw handle:**
If you’re just holding an HRGN, nothing will free it automatically. You must call DeleteObject yourself.

**Owning handle wrapper:**
If you wrap it in something like Owned<HRGN> or Option<Owned<HRGN>>, then the Drop implementation of Owned<T> calls Free::free automatically, so the handle is released when it goes out of scope.

In our case CreateRectRgn return HRGN and it's not released automatically.

amrbashir: You're right, in that case I think we should probably just wrap it in `Owned` then like this
```rs
Owned::new(CreateRectRgn(...));
```
so that we don't need to worry about which method we should call to free an object since the `Free` trait already does that

gribunin: Or just hrgn2.free(), I guess.

amrbashir: Yeah, but `Owned` is better so it is freed at the end of scope implicitly using the `Drop` trait which is more aligned with Rust

gribunin: And while you are there you can add some other fix of much less importance, actually just in case of pure art. Theoretically if SetWindowRgn fails the hrgn1 GDI object would also will leak in this function as it's not owned by Windows. Better to write it as:
```rust
if !SetWindowRgn(hwnd, Some(hrgn1), true).as_bool() {
        // If it fails, we must free hrgn1 manually
        hrgn1.free();
    }
```

In practice if SetWindowRgn fails something much more worse than GDI leak happens of course.

amrbashir: yeah sounds good to me, feel free to update that as well

gribunin: yup, done

amrbashir (dismissed):
Thank you

amrbashir (approved):

PR #14223, opened 2025-09-27
Add support for adaptive and themed  icons on android 

kandrelczyk:
As discussed here:
https://github.com/tauri-apps/tauri/discussions/13860

It's backwards compatible when passing single file as input.
When passing manifest file:
- when manifest only contains default image result is the same as passing file argument
- when bg_color is passed it will override ios-color param
- android_fg, android_bg and android_monochrome can be passed optionally to be used for android

### DIFF
--- a/crates/tauri-cli/src/icon.rs
+++ b/crates/tauri-cli/src/icon.rs
@@ -59,7 +59,7 @@ struct Manifest {
 #[derive(Debug, Parser)]
 #[clap(about = "Generate various icons for all major platforms")]
 pub struct Options {
-  /// Path to the source icon (squared PNG or SVG file with transparency) a manifest file.
+  /// Path to the source icon (squared PNG or SVG file with transparency) or a manifest file.
   /// 
   /// The manifest file is a JSON file with the following structure:
   /// {
--- a/crates/tauri-cli/src/icon.rs
+++ b/crates/tauri-cli/src/icon.rs
@@ -41,6 +41,7 @@ struct PngEntry {
 }
 
 struct AndroidEntries {
+  icon: Vec<PngEntry>,
   foreground: Vec<PngEntry>,
   background: Vec<PngEntry>,
   monochrome: Vec<PngEntry>,
@@ -379,6 +380,7 @@ fn android(
         foreground_size: 432,
       },
     ];
+    let mut icon_entries = Vec::new();
     let mut fg_entries = Vec::new();
     let mut bg_entries = Vec::new();
     let mut monochrome_entries = Vec::new();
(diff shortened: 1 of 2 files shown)


crates/tauri-cli/src/icon.rs
```diff
@@ -40,10 +40,39 @@ struct PngEntry {
...
+  android_bg: Option<String>,
+  android_fg: Option<String>,
+  android_monochrome: Option<String>,
+}
+
 #[derive(Debug, Parser)]
 #[clap(about = "Generate various icons for all major platforms")]
 pub struct Options {
-  /// Path to the source icon (squared PNG or SVG file with transparency).
+  /// Path to the source icon (squared PNG or SVG file with transparency) a manifest file.
```
FabianLars:
```suggestion
  /// Path to the source icon (squared PNG or SVG file with transparency) or a manifest file.
```

FabianLars:
Any idea how to integrate multiple input icons (multiple sizes for the same icon type) into this manifest - doesn't have to be implemented now but we have to keep that requirement in mind for the config design here.

kandrelczyk:
I would suggest simply allowing each entry to be a directory. So for example
`{ "default": "default_icon", "android_bg": "bg_icon"}`. We can check if the values point to a file (current behavior) or a directory. If it's directory we would expect it to contain files like `32.png`, `128.png` and so on. I think that's the most convenient way and current implementation can be easily expanded this way without changing the manifest schema

lucasfernog:
> I would suggest simply allowing each entry to be a directory. So for example `{ "default": "default_icon", "android_bg": "bg_icon"}`. We can check if the values point to a file (current behavior) or a directory. If it's directory we would expect it to contain files like `32.png`, `128.png` and so on. I think that's the most convenient way and current implementation can be easily expanded this way without changing the manifest schema

yes either that or an object like `{ "default": { "32": "./d/32.png", "128": "128.png" } }`
i like how the object is more explicit, easier to spot typos etc

lucasfernog (approved):
i made some changes to be explicit about the input path being the manifest file, actually overlap the bg and fg icons (like Android Studio does) and round the generated icons (again similar to how Android Studio work), i think it's working well

FabianLars (approved):
(untested)

crates/tauri-cli/src/icon.rs
```diff
@@ -40,10 +40,48 @@ struct PngEntry {
...
+
+struct AndroidEntries {
+  icon: Vec<(PngEntry, AndroidIconKind)>,
+  foreground: Vec<PngEntry>,
+  background: Vec<PngEntry>,
+  monochrome: Vec<PngEntry>,
+}
+
+#[derive(Deserialize)]
+struct Manifest {
```
Legend-Master, line 56: Just a note that we might want to generate a schema file and put it into https://schema.tauri.app/ (also some documentations would be nice)

kandrelczyk:
I checked everything again on Android and Linux and looks like it's working fine. 
Regarding documentation, is updating https://v2.tauri.app/develop/icons/ enough? Is there anything else?

Legend-Master:
> Regarding documentation, is updating https://v2.tauri.app/develop/icons/ enough? Is there anything else?

This could be a good start, if we want an even better experience, we could add a json schema like our config files so you get autocompletes in IDEs

> https://github.com/tauri-apps/tauri/pull/14223#discussion_r2394448665

> https://github.com/tauri-apps/create-tauri-app/blob/8d15234e7b473afe68c8601e88f11d1646157404/templates/_base_/src-tauri/%25(v2)%25tauri.conf.json.lte#L2

kandrelczyk:
Ok, I will try to do this but I'm traveling during the next two weeks so probably will do it when I'm back.

PR #14241, opened 2025-10-02
fix(bundler): set APPIMAGE_EXTRACT_AND_RUN env var as well for linuxdeploy

FabianLars:
ever since we updated the appimage output plugin users had linuxdeploy error out with an error not telling them anything. I heard about fuse issues twice though https://github.com/tauri-apps/tauri/issues/8929#issuecomment-3361352855 
Since we already set that env var's cli arg equivalent, this change should be a nobrainer and if it doesn't help, it at least can't realistically make it worse either.

### DIFF
--- a/crates/tauri-bundler/src/bundle/linux/appimage.rs
+++ b/crates/tauri-bundler/src/bundle/linux/appimage.rs
@@ -190,6 +190,8 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
   let mut cmd = Command::new(linuxdeploy_path);
   cmd.env("OUTPUT", &appimage_path);
   cmd.env("ARCH", tools_arch);
+  // Looks like the cli arg isn't enough for the updated AppImage output-plugin.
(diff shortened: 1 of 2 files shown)


crates/tauri-bundler/src/bundle/linux/appimage.rs
```diff
@@ -190,6 +190,8 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
   let mut cmd = Command::new(linuxdeploy_path);
   cmd.env("OUTPUT", &appimage_path);
   cmd.env("ARCH", tools_arch);
+  // Looks like the cli arg isn't enough for the updated AppImage output-plugin.
+  cmd.env("APPIMAGE_EXTRACT_AND_RUN", "1");
```
velocitysystems, line 194: Possibly this is the issue mentioned in this [comment](https://github.com/linuxdeploy/linuxdeploy/issues/104#issuecomment-554803081).

lucasfernog (approved):

PR #14260, opened 2025-10-07
fix(core): parse Android plugin args starting with `is`, closes #14254

### LINKED ISSUE #14254: [bug] plugin boolean value not parsed
### Describe the bug

I have created a plugin and Rust code pass in the boolean value but it is not parsed properly by Kotlin code

Here is the mobile.rs
` pub fn fullscreen(&self, payload: FullScreenRequest) -> crate::Result<()> {
      println!("called fullscreen {:#?}", payload);
      self.0
          .run_mobile_plugin("setFullScreen", payload)
          .map_err(Into::into)
  }`

confirmed that payload contains the value passed in by JS

10-05 18:35:36.773 26325 26373 I RustStdoutStderr: called fullscreen FullScreenRequest {
10-05 18:35:36.773 26325 26373 I RustStdoutStderr:     isFullScreen: true,
10-05 18:35:36.773 26325 26373 I RustStdoutStderr: }

Kotlin code
@InvokeArg
class FullscreenArgs {
  var isFullScreen: Boolean = false
}

 @Command
    fun setFullScreen(invoke: Invoke) {
        val argsMap: Map<String, Any> = invoke.parseArgs(Map::class.java) as Map<String, Any>
        println("Raw JSON: ${argsMap}")
        val visible = invoke.parseArgs(FullscreenArgs::class.java)
        println("Raw visible: ${visible.toString()}")
        if(visible.isFullScreen)
        {
          println("VISIBLE TRUE")
        }
        else {
          println("VISIBLE FALSE")
        }
        println("Hello, Kotlin!")
}

confirmed this code is called and isFullScreen is always false, even tho rust is passing in as true
Just to confirm i have logged the raw value using Map
Here is the output for the map
10-05 18:35:36.786 26325 26325 I System.out: Raw JSON: {isFullScreen=true}

and after that i prints
10-05 18:40:42.996 26933 26933 I System.out: VISIBLE FALSE

So not sure why it is not properly mapped 

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 15.6.1 arm64 (X64)
    ✔ Xcode: 16.4
    - node: 22.14.0
    - pnpm: 9.11.0
    - npm: 10.9.2

[-] Packages
    - tauri 🦀: 2.8.5
    - tauri-build 🦀: 2.4.1
    - wry 🦀: 0.53.3
    - tao 🦀: 0.34.3
    - @tauri-apps/api : 2.8.0
    - @tauri-apps/cli : 2.8.4

[-] Plugins
    - tauri-plugin-os 🦀: 2.3.1
    - tauri-plugin-window-state 🦀: 2.4.0
    - @tauri-apps/plugin-window-state : 2.4.0
    - tauri-plugin-log 🦀: 2.6.0, (outdated, latest: 2.7.0)
    - tauri-plugin-fs 🦀: 2.4.1, (outdated, latest: 2.4.2)
    - tauri-plugin-http 🦀: 2.5.1, (outdated, latest: 2.5.2)
    - @tauri-apps/plugin-http : 2.5.2

[-] App
    - build-type: bundle
    - CSP: script-src 'self' 'unsafe-inline' 'unsafe-eval' tauri: asset: file:; connect-src 'self' https: http: ws: wss: data: tauri:; style-src 'self' 'unsafe-inline' https: http: https://fonts.googleapis.com file:; frame-src 'self' https: http: blob: data:; font-src 'self' https://fonts.gstatic.com data:; media-src 'self' https: http: blob: data: file:; img-src 'self' data: asset: https: http: blob: file: 'unsafe-eval' 'unsafe-inline'; worker-src 'self' blob: data:; default-src 'self' data: tauri: asset: blob: https: http: 'unsafe-eval' 'unsafe-inline' file:
    - frontendDist: ../dist/browser
    - devUrl: http://127.0.0.1:4200/
    - framework: Angular
    - bundler: Webpack
```

### Stack trace

```text

```

### Additional context

_No response_

lucasfernog:
by default Jackson treats the `isX` as a getter, so it looks for the `x` key in the JSON. To match behavior on other platforms we now configure Jackson to treat it as the field name itself.

### DIFF
--- a/.changes/android-plugin-bool-is-deserialize.md
+++ b/.changes/android-plugin-bool-is-deserialize.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch:bug
+---
+
+Properly deserialize Android plugin args with key starting with `is` (previously treated as a getter instead of a field name).
--- a/crates/tauri/mobile/android/src/main/java/app/tauri/plugin/PluginManager.kt
+++ b/crates/tauri/mobile/android/src/main/java/app/tauri/plugin/PluginManager.kt
@@ -17,6 +17,8 @@ import app.tauri.annotation.InvokeArg
 import app.tauri.FsUtils
 import app.tauri.JniMethod
 import app.tauri.Logger
+import com.fasterxml.jackson.annotation.JsonAutoDetect
+import com.fasterxml.jackson.annotation.PropertyAccessor
 import com.fasterxml.jackson.databind.DeserializationFeature
 import com.fasterxml.jackson.databind.JsonNode
 import com.fasterxml.jackson.databind.ObjectMapper
@@ -69,6 +71,7 @@ class PluginManager(val activity: AppCompatActivity) {
     jsonMapper = ObjectMapper()
       .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
       .enable(DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES)
+      .setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY)
 
     val channelDeserializer = ChannelDeserializer({ channelId, payload ->
       sendChannelData(channelId, payload)


FabianLars:
wtf is going on in java land

FabianLars (approved):

PR #14265, opened 2025-10-08
fix(core): SHA256 hash for JS scripts CSP on Windows

lucasfernog:
we hash JS scripts as SHA256 for the Content-Security-Policy (CSP) header. The isolation pattern is broken on Windows due to the hash including carriage return characters, which are not processed when the webview checks the script hash to see if the CSP allows the script.

### DIFF
--- a/crates/tauri-utils/src/html.rs
+++ b/crates/tauri-utils/src/html.rs
@@ -296,23 +296,23 @@ pub fn normalize_script_for_csp(input: &[u8]) -> Vec<u8> {
 
   let mut i = 0;
   while i < input.len() {
-      match input[i] {
-          b'\r' => {
-              if i + 1 < input.len() && input[i + 1] == b'\n' {
-                  // CRLF → LF
-                  output.push(b'\n');
-                  i += 2;
-              } else {
-                  // Lone CR → LF
-                  output.push(b'\n');
-                  i += 1;
-              }
-          }
-          _ => {
-              output.push(input[i]);
-              i += 1;
-          }
+    match input[i] {
+      b'\r' => {
+        if i + 1 < input.len() && input[i + 1] == b'\n' {
+          // CRLF → LF
+          output.push(b'\n');
+          i += 2;
+        } else {
+          // Lone CR → LF
+          output.push(b'\n');
+          i += 1;
+        }
       }
+      _ => {
+        output.push(input[i]);
+        i += 1;
+      }
+    }
   }
 
   output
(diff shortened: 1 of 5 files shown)


FabianLars (approved):

crates/tauri-utils/src/html.rs
```diff
@@ -286,6 +286,38 @@ pub fn inline_isolation(document: &NodeRef, dir: &Path) {
   }
 }
 
+/// Normalize line endings in script content to match what the browser uses for CSP hashing.
+///
+/// According to the HTML spec, browsers normalize:
```
Legend-Master, line 291: Is there a source that we could put a link here?

lucasfernog: i tried to find one but couldn't 😂 maybe this was AI hallucination

Legend-Master: I think it's better be clear about this, I'm afraid to introduce regressions on other platforms if this is not true for them (although rare, they could also have files ending in CRLF), and if that's a behavior we shouldn't rely on, it could break in the future

Legend-Master: Just some information: I was wondering how this worked for the regular inline script in user provided HTML, and from testing, `kuchiki::parse_html` seems to convert `\r\n` to `\n` which seemed to make it work, but for the isolation case, we inserted the js file in `inline_isolation` directly to the document tree, so it preserved the CRLF

https://github.com/tauri-apps/tauri/blob/c57e098d662bc70a8f3eed3c3aa52f13e54ab1f3/crates/tauri-codegen/src/context.rs#L47-L66

lucasfernog: ohh i was wondering the same - good to know

lucasfernog: oh so maybe we should fix inline_isolation instead? :|

Legend-Master:
I don't know how you even found this out, good job! 😂

crates/tauri-utils/src/html.rs
Legend-Master: No idea, I feel like this is pure magic now, I have no idea in which step, the CRLF became LF (using save as in the webview2 devtools on the `index.html` response gave me LF while we surely returned one ending in CRLF, don't know if it's the devtools or it's something else) (don't know if the CSP is calculated from something like `innerText` or from the HTML file directly either), I would say switching to use `nouce` in `inline_isolation` is a more realistic fix 😂

PR #14269, opened 2025-10-09
feat(core): optimize raw payload in event system, closes #13405

### LINKED ISSUE #13405: [feat] Additionally support pushing array buffers with the event system.
More concise description (the bulk of this conversation was debugging a niche GPU rendering issue)

Tauri currently supports sending binary data from the FE to Rust with [array buffers](https://v2.tauri.app/develop/calling-rust/#returning-array-buffers).

It would be great if the event system (Rust -> FE) could skip this serialization cost as well and send array buffers.

Some potential workarounds for those that need this functionality now:
- Rust pings the FE to make a request for a command which sends an array buffer
- Long polling

---

### Describe the problem

I'm prototyping an app that does heavy image processing. Images can easily be 50 megapixels, and the processing pipeline treats each pixel component as a 32 bit float. (this is a common approach. See Darktable) The issue though, is that the payload for a single image in a raw format can easily be up to  `50_000_000 * 3 * 4` bytes (600 MB) If that sounds scary, it already takes 200ms to send only `3MB` of data over IPC. The image data could be represented in 16 bit color (8 bit is insufficient for many cases) but that doesn't make things much better.

It's well documented that IPC in Tauri is slow for large payloads, but workarounds or knowledge about how this could be improved today or in the future is hard to find. It seems that there are a few approaches:

Potential improvements today:
- Implement streaming. working example: https://github.com/tauri-apps/tauri/tree/dev/examples/streaming
  - This example mentions an already built in `asset` protocol but no mention of how to use it / where to learn more
- Shrink payload (even at 100x smaller this would be too slow)

Potential improvements in the future:
- Expose a more developer friendly API for sending binary data
- Shared memory. It sounds like there's security concerns with implementing this
- Use a memory mapped file (similar security concerns?)

A faster way to share / get large amounts of binary data to the front-end, or if it's possible to draw directly from Rust to a canvas element. This is possible if rust is running in WASM, but is likely much more difficult / impossible when Rust is running natively and a separate process.

### Alternatives considered

With Tauri: 
- Creating a separate window exclusively controlled by Rust and GPU. **Would this work across Windows/Linux/MacOS?**
  - I'd still want to send smaller thumbnails, histograms, etc to the web UI
- Streaming (couldn't get this working. probably still too slow with current architecture)
- Rendering directly to a canvas element from Rust (not possible?)

Without Tauri:
- JS + Rust WASM & WGPU
- JS & WGPU

### Additional context

I'm an experienced FE developer and very new to Rust.

I've scoured over a dozen Tauri threads over this already. It's starting to feel like building an image editing app based on Tauri is not possible with current technology, and the only viable alternative if I want to leverage Web UI is to do everything in the browser (eg: Adobe's browser version of Photoshop) but then I'm hit with RAM limitations.

Similar thread: https://github.com/tauri-apps/tauri/discussions/4177

lucasfernog:
leverages the IPC to send raw data through the event system (similar approach to how channels send raw data through - associate data with an ID, and eval a script that fetches the actual data using the IPC to avoid serialization).

The JS API automatically uses the optimized interface, but Rust emitters should use `Emitter::emit_raw` and `Emitter::emit_raw_to` to emit raw payloads, and read `Event::payload_raw` instead of `Event::payload`.
Ideally we would change types a bit to support both JSON and raw payloads, but we're introducing new methods to avoid breaking changes.


lucasfernog:
we might need to introduce new JS functions too.. this actually breaks the data received on the backend side if you are sending array buffers using the current event system (which serializes it)

packages/api/src/event.ts
```diff
@@ -212,10 +235,38 @@ async function emitTo<T>(
...
+ * await emitTo2('main', 'frontend-loaded', { loggedIn: true, token: 'authToken' });
+ * ```
+ *
+ * @param target Label of the target Window/Webview/WebviewWindow or raw {@link EventTarget} object.
+ * @param event Event name. Must include only alphanumeric characters, `-`, `/`, `:` and `_`.
+ * @param payload Event payload.
+ *
+ * @since 2.10.0
+ */
+async function emitTo2<T>(
```
Legend-Master, line 258: Is this new function necessary? We're sending the same command to our rust side anyways no (as long as we handle both cases in rust side which I assume it's already the case when are calling it like this, it should be backward compatible)?

crates/tauri/src/manager/mod.rs
```diff
@@ -577,18 +616,50 @@ impl<R: Runtime> AppManager<R> {
...
+      .values()
+      .cloned()
+      .collect::<Vec<_>>();
+
     let emit_args = match payload {
       EmitPayload::Serialize(payload) => EmitArgs::new(event, payload)?,
-      EmitPayload::Str(payload) => EmitArgs::new_str(event, payload)?,
-    };
+      EmitPayload::Str(payload) => EmitArgs::new_str(event, payload),
+      EmitPayload::Binary(payload) => {
```
Legend-Master, line 631: This should be done with the same check (size check instead of binary type check) we did for the channel implementation or it will actually slow down smaller payloads

Legend-Master (commented):
I kinda feel like it's a time to unify some channel and event staffs together 😂

packages/api/src/event.ts
lucasfernog: https://github.com/tauri-apps/tauri/pull/14269#issuecomment-3386550130

Legend-Master: I'm not sure I get it, isn't the new `emit2` code the same as `emit` in the previous revisions in the PR?

PR #14289, opened 2025-10-12
feat(core): added WindowConfig::on_new_window, closes #14263

### LINKED ISSUE #14263: [bug] Pop-up window creation is blocked in webviews no matter what permissions are set
### Describe the bug

Since some of earlier updates to Tauri or some plugin, webviews are unable to open detached pop-up windows and it blocks many authentication flows (Google, Apple) and other use-cases. I've tried to set all possible local and remote permissions, but nothing helps.

### Reproduction

1. Create a webview of any website that uses pop-ups, i.e. Anthropic
```js
	const webview = new WebviewWindow("anthropic", {
		title: `Anthropic Console`,
		url: "https://console.anthropic.com/login",
		width: 1280,
		height: 800,
	});
```
2. Click on "Continue with Google"

### Expected behavior

A pop-up with Google auth should open. Instead, there is an error:

Image

Here is what I get in Foundry VTT when trying to create a detached window for character sheet:

Image

It's just a couple of examples but I encountered it on many websites and apps.

### Full `tauri info` output

```text
[✔] Environment
    ✔ WebView2: 141.0.3537.57
    - node: 22.20.0
    - npm: 10.9.3
    - bun: 1.2.23

[-] Packages
    - tauri 🦀: 2.8.5
    - tauri-build 🦀: 2.4.1
    - wry 🦀: 0.53.4
    - tao 🦀: 0.34.3
    - @tauri-apps/api : 2.8.0
    - @tauri-apps/cli : 2.8.4

[-] Plugins
    - tauri-plugin-fs 🦀: 2.4.2
    - tauri-plugin-shell 🦀: 2.3.1
    - tauri-plugin-dialog 🦀: 2.4.0
    - @tauri-apps/plugin-dialog : 2.4.0
    - tauri-plugin-global-shortcut 🦀: 2.3.0
    - @tauri-apps/plugin-global-shortcut : 2.3.0
```

### Additional context

It cause a lot of issues in my Tauri app so currently I have to use very outdated Tauri and plugins versions where this issue does not exist.

https://github.com/phenomen/flc/issues/86
https://github.com/phenomen/flc/issues/87
https://github.com/phenomen/flc/issues/88
https://github.com/phenomen/flc/issues/89

lucasfernog:


Legend-Master:
I would assume that this will force the `urlpattern > regex` dependency (increases the binary size by quite a bit) even for people who's not using this feature? (previously it's only the case if you use a remote URL at least from my own experience, rust's dead code elimination will remove it if you're not using it)

Is it possible that we switch to something like `glob` here, like we did for the `opener` plugin?

lucasfernog:
is urlpattern currently only enabled when you have remote URLs in the capability files? Rust optimizes that away right?

if so then we should optimize it indeed

Legend-Master:
From testing this out, it seems to be the case, it's wild that the rust compiler is this smart to remove it to be honest, and with this PR, the file size remains the same if I use remove unused commands and don't include `"core:webview:allow-create-webview-window"`, then the moment I do, the file size increases by ~1MB

packages/api/src/webview.ts
```diff
@@ -897,8 +901,31 @@ interface WebviewOptions {
...
+   *
+   * ## Platform-specific
+   *
+   * - **Android / iOS**: Not supported.
+   *
+   * */
+  onNewWindow?: onNewWindow
 }
 
+type onNewWindow =
```
Legend-Master:
```suggestion
type OnNewWindow =
```

crates/tauri/src/webview/mod.rs
```diff
@@ -704,30 +712,88 @@ tauri::Builder::default()
...
+                    format!("{label}-created-{number}"),
+                    WebviewUrl::External("about:blank".parse().unwrap()),
+                  )
+                  .window_features(features)
+                  .on_document_title_changed(|window, title| {
+                    window.set_title(&title).unwrap();
+                  })
+                  .title(url.as_str());
+
+                  let window = builder.build().unwrap();
```
Legend-Master: Not sure if we should use `unwrap` here?

crates/tauri/src/webview/mod.rs
```diff
@@ -704,30 +712,93 @@ tauri::Builder::default()
...
+                    created_window_count.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+
+                  let builder = WebviewWindowBuilder::new(
+                    &app_handle,
+                    format!("{label}-created-{number}"),
+                    WebviewUrl::External("about:blank".parse().unwrap()),
+                  )
+                  .window_features(features)
+                  .on_document_title_changed(|window, title| {
+                    window.set_title(&title).unwrap();
```
Legend-Master, line 773: please don't `unwrap` here 🙃

crates/tauri/src/webview/mod.rs
```diff
@@ -704,30 +712,93 @@ tauri::Builder::default()
...
+                  tauri_runtime::webview::NewWindowResponse::Deny
+                }
+              }
+              OnNewWindow::AllowDefault { urls: None } => {
+                tauri_runtime::webview::NewWindowResponse::Allow
+              }
+              #[cfg(desktop)]
+              OnNewWindow::AllowTauriWindow { urls: None } => {
+                tauri_runtime::webview::NewWindowResponse::Allow
+              }
```
Legend-Master, line 796: Shouldn't this mean creating a tauri window like the match above? (`OnNewWindow::AllowTauriWindow { urls: Some(urls) }`)

crates/tauri/src/webview/mod.rs
```diff
@@ -704,30 +712,93 @@ tauri::Builder::default()
...
+            Box::new(
+              move |url, features: NewWindowFeatures| match handler(url, features) {
+                NewWindowResponse::Allow => tauri_runtime::webview::NewWindowResponse::Allow,
+                #[cfg(mobile)]
+                NewWindowResponse::Create { window: _ } => {
+                  tauri_runtime::webview::NewWindowResponse::Allow
+                }
+                #[cfg(desktop)]
+                NewWindowResponse::Create { window } => {
+                  tauri_runtime::webview::NewWindowResponse::Create {
```
Legend-Master, line 732: Maybe it's just me, but I still don't quite understand what does `Create` do in `wry`, I tried to look at the docs but I still don't really get it, and trying to call this in the `simple` example in wry but this seems to just freeze the entire app

```rust
    .with_new_window_req_handler(move |url, features| {
      println!("new window req: {url} {features:?}");
      wry::NewWindowResponse::Create {
        webview: features.opener.webview,
      }
    });
```

crates/tauri/src/lib.rs
```diff
@@ -37,6 +37,7 @@
 //! - **macos-proxy**: Adds support for [`WebviewBuilder::proxy_url`] on macOS. Requires macOS 14+.
 //! - **specta**: Add support for [`specta::specta`](https://docs.rs/specta/%5E2.0.0-rc.9/specta/attr.specta.html) with Tauri arguments such as [`State`](crate::State), [`Window`](crate::Window) and [`AppHandle`](crate::AppHandle)
 //! - **dynamic-acl** *(enabled by default)*: Enables you to add ACLs at runtime, notably it enables the [`Manager::add_capability`] function.
+//! - **url-pattern**: Enables using the [URLPattern] spec for URL scopes on JavaScript APIs.
```
Legend-Master, line 40: We probably need the link for this to work

```suggestion
//! - **url-pattern**: Enables using the [`URLPattern`](https://developer.mozilla.org/en-US/docs/Web/API/URLPattern) spec for URL scopes on JavaScript APIs.
```

PR #14293, opened 2025-10-13
feat(windows): keep dwm shadow on frameless windows

### LINKED ISSUE #13134: [feat] Make borders optional instead of binding shadows
### Describe the problem

In Windows 10/11, I am unable to remove borders while having shadows, This limits the implementation of some UI solutions.

I want a parameter to choose whether to keep the border, independent of the shadow option. 
Because for us developers, customizing a border while preserving shadows is easy (such as defining a border for the body and customizing colors and other effects), and removing the original border is next to impossible (at least I don't seem to see an effective way)

Image

### Alternatives considered

I remember the old version seemed to achieve this effect, but I don't know which specific version it is. 
Perhaps the new version can also be implemented? Just need to put in effort?

### Additional context

In my opinion, those who develop applications using webview generally have a high pursuit of application UI, which may require the technology stack to provide higher degrees of freedom
I like to use Tauri, hoping that at least in terms of UI, what others can do, Tauri can do too, and what others cannot do, Tauri can do even better.

chen-zeong:
Fixes #13134.

## Summary
- keep the native DWM shadow when Windows decorations are disabled so frameless windows stay resizable
- store the shadow preference on WindowWrapper and drive it from SetShadow/SetDecorations handlers
- add a dedicated shadow helper to toggle and reset DWM attributes

## Testing
- set `decorations: false` and `shadow: true` in examples/helloworld/tauri.conf.json
- cargo run --example helloworld (Windows 10) and verify the window stays frameless, keeps its shadow, and remains resizable from every edge

### DIFF
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -141,6 +141,8 @@ type IpcHandler = dyn Fn(Request<String>) + 'static;
 #[cfg(not(debug_assertions))]
 mod dialog;
 mod monitor;
+#[cfg(windows)]
+mod shadow;
 #[cfg(any(
   windows,
   target_os = "linux",
@@ -2443,6 +2445,8 @@ pub struct WindowWrapper {
   #[cfg(windows)]
   is_window_transparent: bool,
   #[cfg(windows)]
+  shadow_requested: AtomicBool,
+  #[cfg(windows)]
   surface: Option<softbuffer::Surface<Arc<Window>, Arc<Window>>>,
   focused_webview: Arc<Mutex<Option<String>>>,
 }
@@ -3309,10 +3313,7 @@ fn handle_user_message<T: UserEvent>(
             if !resizable {
               undecorated_resizing::detach_resize_handler(window.hwnd());
             } else if !window.is_decorated() {
-              undecorated_resizing::attach_resize_handler(
-                window.hwnd(),
-                window.has_undecorated_shadow(),
-              );
+              undecorated_resizing::attach_resize_handler(window.hwnd(), false);
             }
           }
           WindowMessage::SetMaximizable(maximizable) => window.set_maximizable(maximizable),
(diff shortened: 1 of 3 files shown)


Think-M:
@FabianLars Does this provide a good solution? : )

Think-M:
@chen-zeong Brother, can you continue to handle this issue? I urgently need it here. Thank you for your contribution!

chen-zeong:
> @chen-zeong Brother, can you continue to handle this issue? I urgently need it here. Thank you for your contribution!

this solution isn't perfect, but it does work.

When it launches, you'll see a border, but it's gone in like half a second.

I've looked at all the popular Tauri projects, and on Windows 10, they all have those unacceptable triple borders because they're trying to use a custom title bar.

Right now, I'm using a modified Tauri version for both of [my apps](https://github.com/chen-zeong/CompressGo/releases). 

You can just look at [my forked branch](https://github.com/chen-zeong/tauri/tree/frameless-toggle) and use it locally.

Think-M:
> > @chen-zeong Brother, can you continue to handle this issue? I urgently need it here. Thank you for your contribution!
> 
> this solution isn't perfect, but it does work.
> 
> When it launches, you'll see a border, but it's gone in like half a second.
> 
> I've looked at all the popular Tauri projects, and on Windows 10, they all have those unacceptable triple borders because they're trying to use a custom title bar.
> 
> Right now, I'm using a modified Tauri version for both of [my apps](https://github.com/chen-zeong/CompressGo/releases).
> 
> You can just look at [my forked branch](https://github.com/chen-zeong/tauri/tree/frameless-toggle) and use it locally.

It's already quite good to achieve such an effect, at least much better than the border that has always existed before. 

I strongly recommend merging it into the official version, which may require you to push it forward further. 

Half a second is not a big problem at all. I think users can bypass it by delaying the display window by half a second. Anyway, this turns a big problem that cannot be bypassed into a small problem that can be bypassed.

Borber:
I am also looking forward to this feature

PR #14323, opened 2025-10-17
deps: apple-codesign 0.27.0 -> 0.29.0

sftse:
I dont know what the MSRV policy of Tauri is, so feel free to close if since #12110 nothing has changed.

Just noticed this update leads to a massive net -26 dependencies (-28 +2). The current MSRV of `apple-codesign 0.29.0` is v1.81.0

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -240,9 +240,9 @@ dependencies = [
 
 [[package]]
 name = "apple-bundles"
-version = "0.19.0"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


FabianLars:
Yeah, the MSRV is why the automatic PRs were closed. In v2 we're stuck with 1.77 for now :/

FabianLars:
not that it _really_ matters for cli deps but for simplicity i'm trying to keep the same msrv for all packages.

sftse:
Missed the MSRV policy when I originally went looking for it, for reference: #13221

PR #14333, opened 2025-10-20
Improve CLI robustness, safety, and readability

KushalMeghani1644:
This PR makes several critical improvements to the Tauri CLI to enhance stability, performance, and maintainability:

1. **Thread safety and resilience:** The stdout/stderr reader threads now lock the mutex only per iteration, and poisoned locks are handled gracefully. This prevents deadlocks and panics when running commands concurrently, especially on long-running build processes.

2. **Performance and clarity:** Command argument formatting was refactored from fold to join, reducing memory allocations and improving readability.

3. **Cross-platform safety:** File extension parsing now correctly handles `OsStr` conversions, preventing potential panics or incorrect parsing on non-UTF-8 paths.

4. **Verbosity consistency:** Introduced a helper function to centralize verbosity parsing from environment variables and CLI flags, ensuring predictable logging levels across all subcommands.

5. **Default values:** Added `Default` derivations to config-related structs to make deserialization more robust and prevent unnecessary unwraps.

6. **Maintainability:** Standardized error handling and logging formatting for clearer messages and easier future modifications.

7. **Regression safety:** Added a minimal CLI help test to catch accidental changes to subcommand names or descriptions.

These changes are necessary to ensure the CLI behaves reliably under all conditions, handles errors gracefully, and remains maintainable as Tauri continues to grow.


crates/tauri-cli/src/lib.rs
```diff
@@ -115,7 +106,7 @@ impl Display for RunMode {
   }
 }
 
-#[derive(Deserialize)]
+#[derive(Deserialize, Default)]
```
FabianLars: Is that used anywhere? Same for PackageJson

crates/tauri-cli/src/lib.rs
```diff
@@ -178,6 +169,13 @@ fn format_error<I: CommandFactory>(err: clap::Error) -> clap::Error {
   err.format(&mut app)
 }
 
+fn get_verbosity(cli_verbose: u8) -> u8 {
```
FabianLars, line 176: > Verbosity consistency: Introduced a helper function to centralize verbosity parsing from environment variables and CLI flags, ensuring predictable logging levels across all subcommands.

This function is only used once, where the code was inlined before. I don't mind having this in a separate function but wanted to mention this.

crates/tauri-cli/src/lib.rs
```diff
@@ -332,18 +315,25 @@ impl CommandExt for Command {
...
     self.stdout(os_pipe::dup_stdout()?);
     self.stderr(os_pipe::dup_stderr()?);
+
     let program = self.get_program().to_string_lossy().into_owned();
-    log::debug!(action = "Running"; "Command `{} {}`", program, self.get_args().map(|arg| arg.to_string_lossy()).fold(String::new(), |acc, arg| format!("{acc} {arg}")));
+    let args = self
+      .get_args()
+      .map(|a| a.to_string_lossy())
+      .collect::<Vec<_>>()
+      .join(" ");
```
FabianLars, line 332: i wonder if that's slower than fold but i agree that it's easier on the eyes.

crates/tauri-cli/src/lib.rs
```diff
@@ -186,10 +184,6 @@ fn format_error<I: CommandFactory>(err: clap::Error) -> clap::Error {
 /// 1. `tauri-cli 1 2 3` -> `1 2 3`
 /// 2. `cargo tauri 1 2 3` -> `1 2 3`
 /// 3. `node tauri.js 1 2 3` -> `1 2 3`
-///
-/// The passed `bin_name` parameter should be how you want the help messages to display the command.
-/// This defaults to `cargo-tauri`, but should be set to how the program was called, such as
-/// `cargo tauri`.
```
FabianLars, line 192: Were these comments incorrect?

crates/tauri-cli/src/lib.rs
```diff
@@ -359,16 +349,17 @@ impl CommandExt for Command {
...
       loop {
         line.clear();
         match stdout.read_line(&mut line) {
           Ok(0) => break,
           Ok(_) => {
             log::debug!(action = "stdout"; "{}", line.trim_end());
-            lines.extend(line.as_bytes().to_vec());
+            if let Ok(mut lines) = stdout_lines_.lock() {
+              lines.extend(line.as_bytes());
+            }
```
FabianLars: i think we should keep the lock() where it was for performance reasons. the next time we will need a lock is after these threads are stopped.

crates/tauri-cli/src/lib.rs
```diff
@@ -359,16 +349,17 @@ impl CommandExt for Command {
...
           }
-          Err(_) => (),
+          Err(_) => break,
```
FabianLars: probably not a good idea, just because one line may error out doesn't mean we should stop reading completely. Except of course if we've had issues with the old behavior which i'm not aware of.

crates/tauri-cli/src/lib.rs
```diff
@@ -378,16 +369,17 @@ impl CommandExt for Command {
...
           Ok(0) => break,
           Ok(_) => {
             log::debug!(action = "stderr"; "{}", line.trim_end());
-            lines.extend(line.as_bytes().to_vec());
+            if let Ok(mut lines) = stderr_lines_.lock() {
+              lines.extend(line.as_bytes());
+            }
           }
-          Err(_) => (),
+          Err(_) => break,
```
FabianLars: same

crates/tauri-cli/src/lib.rs
```diff
@@ -423,4 +415,10 @@ mod tests {
   fn verify_cli() {
     Cli::command().debug_assert();
   }
+
+  #[test]
+  fn help_output_includes_build() {
+    let help = Cli::command().render_help().to_string();
+    assert!(help.contains("Build"));
+  }
```
FabianLars, line 431: i'm kinda failing to see the value of this test but it's not technically wrong so fair enough i guess

crates/tauri-cli/src/lib.rs
```diff
@@ -65,28 +65,19 @@ impl FromStr for ConfigValue {
...
-        Some(ext) if ext == "json5" => {
-          Ok(Self(::json5::from_str(&raw).with_context(|| {
-            format!("failed to parse config at {} as JSON5", path.display())
-          })?))
-        }
-        // treat all other extensions as json
-        _ => Ok(Self(
-          // from tauri-utils/src/config/parse.rs:
-          // we also want to support **valid** json5 in the .json extension
-          // if the json5 is not valid the serde_json error for regular json will be returned.
```
FabianLars, line 83: since the logic is still the same we should probably keep the comments as well

crates/tauri-cli/src/lib.rs
KushalMeghani1644: I made this function because the logic may be reused (e.g. if we later need verbosity in other commands),
but I’m fine inlining it if preferred.

KushalMeghani1644:
Hey! @FabianLars I have made the changes, you reviewed/requested... I have added back the comments that I had blindly removed... removed the `Default` as-well. I hope the changes made are fine. :)

crates/tauri-cli/src/lib.rs
FabianLars: @KushalMeghani1644 can you copy these comments over as well?

KushalMeghani1644: Oh yes! I missed these! ~~though~~ Now I have added them back now. Sorry for the inconsistency

FabianLars (approved):
thanks :)

crates/tauri-cli/src/lib.rs
```diff
@@ -222,23 +224,17 @@ where
     Err(e) => e.exit(),
   };
 
-  let verbosity_number = std::env::var("TAURI_CLI_VERBOSITY")
-    .ok()
-    .and_then(|v| v.parse().ok())
-    .unwrap_or(cli.verbose);
-  // set the verbosity level so subsequent CLI calls (xcode-script, android-studio-script) refer to it
```
Legend-Master, line 229: Can we add back those comments? Not sure why you removed them

crates/tauri-cli/src/lib.rs
```diff
@@ -222,23 +224,17 @@ where
...
+  let verbosity_number = get_verbosity(cli.verbose);
   std::env::set_var("TAURI_CLI_VERBOSITY", verbosity_number.to_string());
 
   let mut builder = Builder::from_default_env();
-  let init_res = builder
+  if let Err(err) = builder
     .format_indent(Some(12))
     .filter(None, verbosity_level(verbosity_number).to_level_filter())
-    // golbin spams an insane amount of really technical logs on the debug level so we're reducing one level
```
Legend-Master, line 236: same here

crates/tauri-cli/src/lib.rs
```diff
@@ -222,23 +224,17 @@ where
...
     .filter(
       Some("goblin"),
       verbosity_level(verbosity_number.saturating_sub(1)).to_level_filter(),
     )
-    // handlebars is not that spammy but its debug logs are typically far from being helpful
```
Legend-Master, line 241: same here

crates/tauri-cli/src/lib.rs
```diff
@@ -321,8 +314,6 @@ fn prettyprint_level(lvl: Level) -> &'static str {
 }
 
 pub trait CommandExt {
-  // The `pipe` function sets the stdout and stderr to properly
```
Legend-Master, line 324: same here

Legend-Master (changes requested):

crates/tauri-cli/src/lib.rs
FabianLars: I could swear all your marked comments were added back... Maybe it's time for me to retire lol

KushalMeghani1644: Oops! I overlooked these comments, I shall stop using neovim... Jokes aside, I will have them added back through another Pr (since this is merged).

PR #14340, opened 2025-10-21
perf: optimize CSP nonce generation

Tunglies:
Pre-allocate nonces Vec and reuse a single buffer with `write!` instead of `format!`, reducing heap allocations when replacing tokens.

Adjust CSP nonce buffer size based on target pointer width


.changes/change-pr-14340.md
```diff
@@ -0,0 +1,18 @@
...
+"tauri-runtime-wry": patch
+"tauri-codegen": patch
+"tauri-macros": patch
+"tauri-plugin": patch
+"tauri-build": patch
+"tauri": patch
+"@tauri-apps/cli": patch
+"tauri-cli": patch
+"tauri-driver": patch
+---
```
Legend-Master, line 3:
```suggestion
---
"tauri": patch
---
```

This should do, no need to bump all dependencies 🙃

crates/tauri/src/manager/mod.rs
```diff
@@ -127,7 +128,7 @@ fn replace_csp_nonce(
   directive: &str,
   hashes: Vec<String>,
 ) {
-  let mut nonces = Vec::new();
+  let mut nonces = Vec::with_capacity(asset.matches(token).count());
   *asset = replace_with_callback(asset, token, || {
     #[cfg(target_pointer_width = "64")]
     let mut raw = [0u8; 8];
```
Legend-Master, line 134: @lucasfernog do you remember why we used `usize` here at the first place (tracing back to https://github.com/tauri-apps/tauri/commit/cf54dcf9c81730e42c9171daa9c8aa474c95b522)? It doesn't quite make sense to me

.changes/change-pr-14340.md
Tunglies: Sry for that. The bot recommend by default.

crates/tauri/src/manager/mod.rs
vrmiguel: Maybe I don't get the question, but the code _is_ still using a `usize` there, just that it's using its underlying byte representation instead, right?

In any case, seems like this code can be simplified down to:

```rust
let mut raw = [0u8; std::mem::size_of::<usize>()];
getrandom::fill(&mut raw).expect("failed to get random bytes");
let nonce = usize::from_ne_bytes(raw);
```

Or even:

```rust
let mut raw = 0_usize.to_ne_bytes();
getrandom::fill(&mut raw).expect("failed to get random bytes");
let nonce = usize::from_ne_bytes(raw);
```

crates/tauri/src/manager/mod.rs
```diff
@@ -141,17 +142,23 @@ fn replace_csp_nonce(
...
+    if !sources.contains(self_source) {
       sources.push(self_source);
     }
-    sources.extend(nonce_sources);
+    #[cfg(target_pointer_width = "64")]
+    let mut buf = String::with_capacity(28);
+    #[cfg(target_pointer_width = "32")]
+    let mut buf = String::with_capacity(20);
+    #[cfg(target_pointer_width = "16")]
+    let mut buf = String::with_capacity(14);
```
vrmiguel, line 156: Are these capacities meant to be how many bytes to alloc to store `'nonce-{0..usize::MAX}'`? If so, isn't it a bit oversized for 32 and 16 ptr widths? Feels like it should be 18 and 13 respectively, unless I'm missing something 

If we want to be a bit fancy-pants (😛) we could use this helper function:

```rust
const fn nonce_source_capacity() -> usize {
    // 'nonce-' prefix + digits of usize::max + closing quote
    8 + usize::MAX.ilog10() as usize + 1
}
```

(You can test that by replacing usize for u64, u32 and u16 respectively)

crates/tauri/src/manager/mod.rs
```diff
@@ -141,17 +142,23 @@ fn replace_csp_nonce(
...
+    for nonce in nonces {
+      buf.clear();
+      write!(&mut buf, "'nonce-{}'", nonce).unwrap();
+      sources.push(buf.clone());
+    }
```
vrmiguel, line 161: Reusing `buf` doesn't seem helpful in this scenario, if we'll clone it anyway. It means that the code will allocate `nonces.len()+1` times

```rust
for nonce in nonces {
      let mut buf = String::with_capacity(nonce_capacity());
      write!(&mut buf, "'nonce-{}'", nonce).unwrap();
      sources.push(buf);
}
```
Something like that ^ would allocate `n` times rather than `n+1` times

Legend-Master: I honestly doubt these optimizations will actually make any differences, format! Itself is quite optimized to estimate the buffer size already

crates/tauri/src/manager/mod.rs
Legend-Master: I meant it doesn't make sense to make nonce length depends on target pointer size

PR #14347, opened 2025-10-22
chore: remove `--cfg docsrs` to fix docs.rs builds

FabianLars:
this worked in tao but i forgot to include it in tauri for 2.9


FabianLars:
tested in a local docs.rs environment.

PR #14356, opened 2025-10-23
fix: don't try to parse inexistent icons

WhyNotHugo:
When Tauri is explicitly configured to use no icons via:

    "bundle": {
      "icon": []
    }  

It still tries to compile icons in a fallback path of `icons/icon.png`. This blocks compiling any project unless artwork is in place, even though this might be entirely beyond scope for a project.

This issue stems from the fact that `Vec<String>` is used to deserialise icons, so there is no way to distinguish "no explicit value" from "explicitly instructed to use no value".

Use an `Option<Vec<String>>` to deserialise configured icons, where no value (None) means "nothing explicitly specified; use the default), and Some(EmptyVec) means "explicitly requested no icons".

This allows compiling small proof of concept projects without having to worry about artwork.

Fixes: https://github.com/tauri-apps/tauri/discussions/14355

### DIFF
--- a/crates/tauri-build/src/codegen/context.rs
+++ b/crates/tauri-build/src/codegen/context.rs
@@ -101,11 +101,13 @@ impl CodegenContext {
       }
       _ => (),
     }
-    for icon in &config.bundle.icon {
-      println!(
-        "cargo:rerun-if-changed={}",
-        config_parent.join(icon).display()
-      );
+    if let Some(icons) = &config.bundle.icon {
+      for icon in icons {
+        println!(
+          "cargo:rerun-if-changed={}",
+          config_parent.join(icon).display()
+        );
+      }
     }
     if let Some(tray_icon) = config.app.tray_icon.as_ref().map(|t| &t.icon_path) {
       println!(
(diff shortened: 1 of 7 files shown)


WhyNotHugo:
v2: fixed issue in `cargo check`.

Building a downstream project with no icons now works fine :+1:

Legend-Master (commented):
Have not tested the build without icons yet, just some early feedback on the design

I think we don't really need to use `Option<Vec<_>>` here, we can simply skip the icon if the user didn't provide one, either by using an empty list or didn't add that field

matthiasbeyer:
I am interested in this as well. What is necessary to land this patch?

matthiasbeyer:
Coming back again, how about this gets landed?
I still cannot use tauri in my project because of this issue...

matthiasbeyer:
FWIW, I applied [this patch with a small fixup](https://github.com/matthiasbeyer/tauri/tree/tauri-v2.10.3-patched) to 2.10.3 and was able to start my app.

CC @lucasfernog @Legend-Master

Legend-Master:
I don't think this works on Windows yet, an icon is required in `tauri-build` as well

matthiasbeyer:
What exactly is not working on windows?

Legend-Master:
We have an assertion on putting an icon to the resources (.rc file for the linker), not sure if that's needed

matthiasbeyer:
Can you point me to it?

Legend-Master:
https://github.com/tauri-apps/tauri/blob/a30dca4820d0ad681f04737a1819e6a6fab9fe84/crates/tauri-build/src/lib.rs#L668

matthiasbeyer:
Cool, thanks.

@WhyNotHugo do you have a windows machine to test this on?

WhyNotHugo:
I don’t

PR #14357, opened 2025-10-23
feat: Add support for setting icon & template at same time

ykogan-discord:
Calling set_icon and then set_icon_as_template in sequence cause a flicker as they both run on the main thread and update the UI. This exposes a single function to do both at once, preventing the flicker.


FabianLars:
Thanks for the PR! Would you also be interested in adding the corresponding JS api? Would require changes in https://github.com/ykogan-discord/tauri/blob/yonatron/set-icon-and-template-at-same-time/crates/tauri/src/tray/plugin.rs and https://github.com/ykogan-discord/tauri/blob/yonatron/set-icon-and-template-at-same-time/packages/api/src/tray.ts i think

ykogan-discord:
> Thanks for the PR! Would you also be interested in adding the corresponding JS api? Would require changes in https://github.com/ykogan-discord/tauri/blob/yonatron/set-icon-and-template-at-same-time/crates/tauri/src/tray/plugin.rs and https://github.com/ykogan-discord/tauri/blob/yonatron/set-icon-and-template-at-same-time/packages/api/src/tray.ts i think

Heyo! Yes happy to do this. Unsure of my exact timing but I'll try to get it done sooner than later.

ykogan-discord:
> Thanks for the PR! Would you also be interested in adding the corresponding JS api? Would require changes in https://github.com/ykogan-discord/tauri/blob/yonatron/set-icon-and-template-at-same-time/crates/tauri/src/tray/plugin.rs and https://github.com/ykogan-discord/tauri/blob/yonatron/set-icon-and-template-at-same-time/packages/api/src/tray.ts i think

Okay, I made these changes just via pattern matching but I am unsure how to test them. I'm also getting a failure in a snapshot test, which is instructing me to run `cargo insta review` but `cargo insta` doesn't exist. Do I need to install it globally?

Apologies, a bit of a rust noob.

FabianLars:
yeah testing is a bit manual at the moment, at least for the js api, so we typically just use the example app.

i usually just ignore the snapshot stuff tbh x)

crates/tauri/src/tray/mod.rs
```diff
@@ -576,6 +576,38 @@ impl<R: Runtime> TrayIcon<R> {
...
 
+/// Sets the tray icon and template status atomically. **macOS only**.
+///
+/// On macOS, calling `set_icon` followed by `set_icon_as_template` causes a visible
+/// flicker as the icon is rendered twice. This method sets both atomically to prevent that.
+///
+/// ## Platform-specific:
+///
+/// - **Linux / Windows:** Falls back to calling `set_icon`.
+pub fn set_icon_with_template(
```
FabianLars: wonder if we should match the name used in tray-icon `set_icon_as_with_template` (as much as i hate it) 🤔

ykogan-discord: yeah that makes sense to me :)

ykogan-discord:
Okay turns out I needed to change one more file to expose the method in the plugin but I tested this and it's working!

ykogan-discord:
@FabianLars anything else you need from me here or do you do the merge?

FabianLars:
Ah no sorry, you're good! Just waiting for the merge window (which we didn't define yet)

andrewhamon:
> Just waiting for the merge window (which we didn't define yet)

Friendly bump on this. I'm a colleague of Yonatan and would love to stop carrying this as a patch :)

FabianLars:
still waiting for that 2.10 merge window. i highly doubt it'll be before or during the holidays.

andrewhamon:
@FabianLars it seems like 2.10 has come and gone :/

FabianLars (dismissed):

lucasfernog (approved):
well done

Issue #14186, opened 2025-09-14
[bug] Bundler fails to patch __TAURI_BUNDLE_TYPE when `strip` enabled in Cargo.toml


btjawa:
### Describe the bug

When enabling `strip = true` in `[profile.release]` of `Cargo.toml`, the bundler fails to patch the `__TAURI_BUNDLE_TYPE` marker. Without `strip`, everything works fine.

This only occurs on Linux (Windows and macOS works fine with strip enabled).

GitHub Actions (ubuntu-latest runner) reproduces this as well, so i guess this can also be stably reproduced on other machines.

### Reproduction

1. Create a new tauri-app via `sh <(curl https://create.tauri.app/sh)`.
2. Edit `src-tauri/Cargo.toml` and add:
```toml
[profile.release]
strip = true
```
3. Run `pnpm tauri build`.

### Expected behavior

If `strip` is enabled, the bundler should still be able to patch `__TAURI_BUNDLE_TYPE`.

Or at minimum, there should be a clear warning/documentation that `strip` should not be enabled on Linux builds.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Ubuntu 24.4.0 x86_64 (X64) (ubuntu on x11)
    ✔ webkit2gtk-4.1: 2.48.5
    ✔ rsvg2: 2.58.0
    ✔ rustc: 1.89.0 (29483883e 2025-08-04)
    ✔ cargo: 1.89.0 (c24e10642 2025-06-23)
    ✔ rustup: 1.28.2 (e4f3ad6f8 2025-04-28)
    ✔ Rust toolchain: stable-x86_64-unknown-linux-gnu (default)
    - node: 22.19.0
    - pnpm: 10.16.1
    - npm: 11.6.0

[-] Packages
    - tauri 🦀: 2.8.5
    - tauri-build 🦀: 2.4.1
    - wry 🦀: 0.53.3
    - tao 🦀: 0.34.3
    - @tauri-apps/api : 2.8.0
    - @tauri-apps/cli : 2.8.4

[-] Plugins
    - tauri-plugin-opener 🦀: 2.5.0
    - @tauri-apps/plugin-opener : 2.5.0

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:1420/
    - framework: Vue.js
    - bundler: Vite
```

### Stack trace

```text

```

### Additional context

`pnpm tauri build`:

```
[...]

PR #14368, opened 2025-10-26
add info to error message (fix 14186)

kandrelczyk:
Temporary fix to https://github.com/tauri-apps/tauri/issues/14186

Adding info about symbol stripping to the warning message.

### DIFF
--- a/crates/tauri-bundler/src/bundle.rs
+++ b/crates/tauri-bundler/src/bundle.rs
@@ -116,7 +116,7 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
     }
 
     if let Err(e) = patch_binary(&main_binary_path, package_type) {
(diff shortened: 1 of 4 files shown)


crates/tauri-bundler/src/bundle.rs
```diff
@@ -116,7 +116,7 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
     }
 
     if let Err(e) = patch_binary(&main_binary_path, package_type) {
-      log::warn!("Failed to add bundler type to the binary: {e}. Updater plugin may not be able to update this package. This shouldn't normally happen, please report it to https://github.com/tauri-apps/tauri/issues");
+      log::warn!("Failed to add bundler type to the binary: {e}. Updater plugin may not be able to update this package. Make sure symbol stripping is disabled (https://doc.rust-lang.org/cargo/reference/profiles.html#strip). This shouldn't normally happen, please report it to https://github.com/tauri-apps/tauri/issues");
```
Legend-Master: This is only the case for Linux right? Maybe we could only include this part for the linux bundle targets

Legend-Master (dismissed):
Thanks! Also do you mind also adding a change file?

kandrelczyk:
Added change file. It also makes more sense to have this in the error message. The full warning reads better this way.

Legend-Master (approved):
> It also makes more sense to have this in the error message. The full warning reads better this way.

Nice one, thanks!

PR #14379, opened 2025-10-28
Respect required-features field from Cargo.toml

moubctez:
`Cargo.toml` allows binary target definition to contain `required-features` field (see [The Cargo Book](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-required-features-field)). If defined, the binary is built only when a certain feature is activated.

Tauri pulls the list of binaries from `Cargo.toml` regardless of the presence of the `required-features` field and tries to add them to the final package. But the binaries may not be there.

This pull request addresses this problem.

Example excerpt from `Cargo.toml`:

```
[[bin]]
name = "defguard-client"

[[bin]]
name = "defguard-service"
required-features = ["service"]
```

Example build with `cargo-tauri`:

```shell
cargo tauri build
```

The above will exclude `defguard-service` from the final package, while this:

```shell
cargo tauri build --features service
```

will include all binaries.


FabianLars:
thanks for the pr! would you mind adding a small changefile as well please?

moubctez:
> thanks for the pr! would you mind adding a small changefile as well please?

Done

FabianLars (approved):

crates/tauri-cli/src/interface/rust.rs
```diff
@@ -955,6 +957,14 @@ impl AppSettings for RustAppSettings {
...
         .unwrap_or_default();
       for bin in bins {
+        if let (Some(req_features), Some(opt_features)) =
+          (&bin.required_features, &options.features)
+        {
+          // Check if all required features are enabled.
+          if !req_features.iter().all(|feat| opt_features.contains(feat)) {
+            continue;
+          }
+        }
```
sftse, line 967: Shouldn't this have the same code path if `options.features == None` vs `options.features == Some(["a", "b"])` when `req_features == Some(["c"])`? The `continue` will be hit in the latter case, but not in the former.

Looking into this due to #14607

Issue #12172, opened 2025-01-03
[bug] App does not initialize when debugging from Xcode


velocitysystems:
### Describe the bug

When debugging a Tauri app in an iOS simulator or device from Xcode, the app _deploys_ but does not _initialize_. An empty white webview is show.

### Reproduction

Clone the example app in the `plugins-workspace` [repository](https://github.com/tauri-apps/plugins-workspace).
Confirm the app builds/deploys using the CLI to an iOS simulator.

Run the following command:
`pnpm tauri ios dev --open`

![Screenshot 2025-01-01 at 16 21 41](https://github.com/user-attachments/assets/c9851e97-2b3c-4a15-95e0-a26fa317ae7c)

### Expected behavior

App should build, deploy and run on the iOS simulator.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 15.2.0 arm64 (X64)
    ✔ Xcode Command Line Tools: installed
    ✔ rustc: 1.83.0 (90b35a623 2024-11-26)
    ✔ cargo: 1.83.0 (5ffbef321 2024-10-29)
    ✔ rustup: 1.27.1 (54dd3d00f 2024-04-24)
    ✔ Rust toolchain: stable-aarch64-apple-darwin (default)
    - node: 22.12.0
    - pnpm: 9.15.2
    - npm: 10.9.0

[-] Packages
    - tauri 🦀: 2.1.1
    - tauri-build 🦀: 2.0.3
    - wry 🦀: 0.47.2
    - tao 🦀: 0.30.8
    - @tauri-apps/api : 2.1.1 (outdated, latest: 2.2.0)
    - @tauri-apps/cli : 2.1.0 (outdated, latest: 2.2.1)
```

davehorner:
I also also experiencing this on macOS.  I am able to run the app as a macOS application.  I can hit the frontend from the simulator safari.  It shows the title for a quick second and then shows all white aside from the time and battery indicators.

I saw some errors about "View Controller" is unreachable with no entry points.  I tried setting the view controller to "Is Initial View Controller" this stopped the warnings but did not get the application to initialize properly.

I can attach with a debugger but breaking doesn't get me anywhere of interest.  I can say that it doesn't initialize properly when run from the command line without the --open too.

velocitysystems:
[...]

PR #14385, opened 2025-10-28
refactor(core): improve iOS log messages from stdout/stderr

lucasfernog:
move the stdout/stderr forward logic to Swift so it does not consume a Rust thread and never deadlocks on the simulator

I had to work on this because i'm facing #12172 again on latest Xcode (26.1)

### DIFF
--- a/.changes/refactor-ios-stdout-stderr.md
+++ b/.changes/refactor-ios-stdout-stderr.md
@@ -1,0 +1,6 @@
+---
+"tauri": path:bug
+"tauri-macros": patch:bug
+---
+
+Fix iOS deadlock when running on the simulator from Xcode by properly piping stdout/stderr messages through the Xcode console and OSLog.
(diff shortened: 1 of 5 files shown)


lucasfernog:
another benefit is that the Xcode console now properly shows all messages (OSLog events and stdout/stderr)

will3942:
@lucasfernog This is great and affecting us locally - would be great to see this shipped.

FabianLars:
@will3942 did you try it in your project? did it work well?

lucasfernog:
sometimes it logs stdout/stderr messages twice, I didn't have much more time to investigate. But at least it never freezes, which is really annoying.

FabianLars (approved):
(untested)

FabianLars:
> sometimes it logs stdout/stderr messages twice, I didn't have much more time to investigate. 

probably not too much of an issue, so can be looked into later imo

PR #14398, opened 2025-10-31
fix(cli): skip checks ndk when NDK_HOME set

mslxl:
In the original code, the way to ensure NDK installation is to check whether the ndk folder exists under the `$ANDROID_HOME` directory. If it exists, the `NDK_HOME` environment variable is set; otherwise, the NDK will be installed. This approach overrides any user-set `NDK_HOME` variable. However, on some systems (like NixOS), the NDK installed via package managers may not be located in the `$ANDROID_HOME/ndk` directory but elsewhere. Ignoring the `NDK_HOME` variable would render such system-installed NDKs unusable.

This PR bypasses the NDK check when `NDK_HOME` has already been set.


Legend-Master:
Hi, I have merged #15344 since it had a change file and passed the CI tests, thanks anyways!

Issue #3725, opened 2022-03-17
[bug] Loading a video/audio as an asset does not work


firstdorsal:
### Describe the bug

Loading a video as an asset does not work.

I followed the guide here: https://tauri.studio/docs/api/js/modules/tauri

It works fine with image files (png/jpg etc.) but not with videos. 
I tried loading the video https://www.w3schools.com/html/mov_bbb.mp4 directly from the URL, and this worked without problems.
But using the `asset:///home/paul/Downloads/mov_bbb.mp4` URL this does not work
Loading images from the same folder does work.

In the network panel you can see that the request failed but no further reason is specified
![image](https://user-images.githubusercontent.com/10563452/158854074-2a3b165e-8490-4ae2-b418-2ff682bac0a1.png)

![image](https://user-images.githubusercontent.com/10563452/158854189-c1b2de0a-6fd5-4b1d-9bb3-a0a8bc8ccf0b.png)

The loading of assets this way is IMO not very good explained and when it fails, for some reason, it is very difficult to find out why.

Thank you for creating Tauri, it is awesome!

### Platform and versions

```shell
Operating System - Arch Linux, version Rolling Release X64

Node.js environment
  Node.js - 17.7.1
  @tauri-apps/cli - 1.0.0-rc.6
  @tauri-apps/api - 1.0.0-rc.2

Global packages
  npm - 8.5.4
  pnpm - Not installed
  yarn - 1.22.17

Rust environment
  rustup - 1.24.3
  rustc - 1.59.0
  cargo - 1.59.0
  toolchain - stable-x86_64-unknown-linux-gnu 

App directory structure
/dist
/node_modules
/src-tauri
/src
/.git
/public

App
  tauri - 1.0.0-rc.4
  tauri-build - 1.0.0-rc.4
  tao - 0.6.4
  wry - 0.13.3
  build-type - build
  CSP - default-src tauri: asset: https://asset.localhost
  distDir - ../dist
  devPath - http://localhost:3000/
  framework - React
```

red010182:
This config works for me.
```
    "allowlist": {
      ...
      "protocol": {
        "asset": true,
        "assetScope": ["**"]
      }
    },
    "security": {
      "csp": "default-src 'self'; img-src 'self' asset: https://asset.localhost; media-src 'self' asset: https://asset.localhost"
    }

```

---
Edit:

Don't do this. It's dangerous: https://github.com/tauri-apps/tauri/issues/3735#issuecomment-1072978250

[...]

PR #14402, opened 2025-11-01
fix(linux): adding Gstreamer plugin for handling asset:// 

aaalloc:
asset:// is registered as a custom protocol through WebKit, but when an audio or video is got from that protocol, gstreamer is used in the background and can't play the media because there is no handler for asset://
(related to #3725)

This PR creates a Gstreamer Plugin that will handle asset:// url.

If you want to test it out: build the plugin so that you will have a .so, `export GST_PLUGIN_PATH=../../target/debug` (where the build .so lies), and then launch your tauri app or simply try it out with `gst-launch-1.0 uridecodebin uri=asset:///your/path/to/video/or/audio.mp3 ! videoconvert ! autovideosink`

If we want to make things work we will need to place the .so into Gstreamer plugins search path. There's some information about it here : https://gstreamer.freedesktop.org/documentation/gstreamer/gstregistry.html?gi-language=c#gst_registry_add_plugin 
```
On startup, plugins are searched for in the plugin search path. The following locations are checked in this order:

    location from --gst-plugin-path commandline option.
    the GST_PLUGIN_PATH environment variable.
    the GST_PLUGIN_SYSTEM_PATH environment variable.
    default locations (if GST_PLUGIN_SYSTEM_PATH is not set). Those default locations are: $XDG_DATA_HOME/gstreamer-$GST_API_VERSION/plugins/ and $prefix/libs/gstreamer-$GST_API_VERSION/. [$XDG_DATA_HOME](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) defaults to $HOME/.local/share.
```

A solution would be to add the .so where every other gstreamer plugins lies, in my case I have simply put it in `/usr/lib64/gstreamer-1.0` and Gstreamer detected it.


crates/tauri-asset-gst-plugin/rust-toolchain.toml
aaalloc, line 1: This may be unnecessary

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,95 @@
...
+
+impl URIHandlerImpl for TauriAsset {
+  const URI_TYPE: gst::URIType = gst::URIType::Src;
+
+  fn protocols() -> &'static [&'static str] {
+    &[ASSET_URI_SCHEME]
+  }
+
+  fn uri(&self) -> Option<String> {
+    None
```
aaalloc: I don't think it should be like this, but I don't think it is a problem for now

DEVGOD2020:
For anyone with special charecters in their file names...
```rust
use percent_encoding::percent_decode_str;

    // directly having full path after asset:// or having localhost
    let location = location.strip_prefix("localhost").unwrap_or(location).to_string();
    
    //Fix URI UTF8 encoding tauri does, fix file names with special characters
    let location = percent_decode_str(&location)
        .decode_utf8().expect("Valid UTF-8")
        .to_string();
```
In imp.rs you could also use the percent_encoding crate to handle UTF8 encoding file name related issues.
Spaces, dashes, and Chinese characters in audio file names were causing errors for my tauri music player without the UTF8 decoding.

aaalloc:
> For anyone with special charecters in their file names...
> 
> ```rust
> use percent_encoding::percent_decode_str;
> 
>     // directly having full path after asset:// or having localhost
>     let location = location.strip_prefix("localhost").unwrap_or(location).to_string();
>     
>     //Fix URI UTF8 encoding tauri does, fix file names with special characters
>     let location = percent_decode_str(&location)
>         .decode_utf8().expect("Valid UTF-8")
>         .to_string();
> ```
> 
> In imp.rs you could also use the percent_encoding crate to handle UTF8 encoding file name related issues. Spaces, dashes, and Chinese characters in audio file names were causing errors for my tauri music player without the UTF8 decoding.

@DEVGOD2020  I fixed this case, thank you !

aaalloc:
any status about this ?

cjpais:
showing support for this as well from https://github.com/cjpais/Handy/issues/234

crates/tauri-asset-gst-plugin/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+authors = ["Alexander Yanovskyy <contact@yanovskyy.com>"]
+license = "MIT OR Apache-2.0"
+edition = "2018"
+repository = "https://github.com/tauri-apps/tauri"
+description = "Gstreamer plugin for asset protocol"
+
+[dependencies]
+gstreamer = "0.24"
+gstreamer-base = "0.24"
+percent-encoding = "2.3.2"
```
snue, line 12: The `Cargo.lock` is still at `2.3.1` for that, which gives issues for vendored deps packaging.

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,104 @@
+use gst::glib::subclass::types::ObjectSubclassExt;
+use gst::prelude::{ElementExt, GhostPadExt, GstBinExt, PadExt};
+use gst::subclass::prelude::{BinImpl, ElementImpl, ObjectImpl, URIHandlerImpl};
+use gst::subclass::prelude::{GstObjectImpl, ObjectSubclass};
+use gst::{glib, GhostPad, PadDirection};
+use gstreamer_base::gst;
+
+const ASSET_URI_SCHEME: &str = "asset";
```
snue, line 22: Not sure if this is expected, but the application I am trying to get sound of uses a `tauri://` URI scheme.

```
ERROR      webkitmediaplayer MediaPlayerPrivateGStreamer.cpp:2034:handleMessage:<media-player-1> No URI handler implemented for "tauri". (url=tauri://localhost/sound/standard/Move.mp3) (code=12)
```

When I change the scheme, it fails to find the file. It's obviously not using a full filesystem path.

```
0:00:00.076053309 1151971 0x558b05662860 WARN                 filesrc gstfilesrc.c:585:gst_file_src_start:<filesrc> error: No such file "/sound/standard/Move.mp3"
0:00:00.076074098 1151971 0x558b05662860 WARN                 basesrc gstbasesrc.c:3616:gst_base_src_start:<filesrc> error: Failed to start
0:00:00.076247474 1151971 0x558b05662860 WARN                 filesrc gstfilesrc.c:585:gst_file_src_start:<filesrc> error: No such file "/sound/standard/Move.mp3"
0:00:00.076262393 1151971 0x558b05662860 WARN                 basesrc gstbasesrc.c:3616:gst_base_src_start:<filesrc> error: Failed to start
0:00:00.076280507 1151971 0x558b05662860 WARN                 filesrc gstfilesrc.c:585:gst_file_src_start:<filesrc> error: No such file "/sound/standard/Move.mp3"
0:00:00.076288652 1151971 0x558b05662860 WARN                 basesrc gstbasesrc.c:3616:gst_base_src_start:<filesrc> error: Failed to start
0:00:00.076296787 1151971 0x558b05662860 WARN                 basesrc gstbasesrc.c:3984:gst_base_src_activate_push:<filesrc> Failed to start in push mode
0:00:00.076301656 1151971 0x558b05662860 WARN                GST_PADS gstpad.c:1158:gst_pad_set_active:<filesrc:src> Failed to activate pad
0:00:00.076514136 1151971 0x558b05662860 WARN       webkitmediaplayer MediaPlayerPrivateGStreamer.cpp:1083:changePipelineState:<media-player-0> Changing state to PAUSED from NULL with VOID_PENDING pending failed
0:00:00.088278467 1151971 0x558b05662860 WARN               structure gststructure.c:3100:priv_gst_structure_append_to_gstring: No value transform to serialize field 'gerror' of type 'GError'
0:00:00.088310788 1151971 0x558b05662860 ERROR           webkitcommon GStreamerCommon.cpp:927:operator():<media-player-0> Got message: error message: 0x558b060d56a0, time 99:99:99.999999999, seq-num 76, element 'filesrc', GstMessageError, gerror=(GError)NULL, debug=(string)"../plugins/elements/gstfilesrc.c\(585\):\ gst_file_src_start\ \(\):\ /GstPlayBin:media-player-0/GstURIDecodeBin:uridecodebin0/GstTauriAsset:source/GstFileSrc:filesrc:\012No\ such\ file\ \"/sound/standard/Move.mp3\"";
0:00:00.088337317 1151971 0x558b05662860 ERROR      webkitmediaplayer MediaPlayerPrivateGStreamer.cpp:2034:handleMessage:<media-player-0> Resource not found. (url=tauri://localhost/sound/standard/Move.mp3) (code=3)
```

I wonder if this is a separate issue (needs its own gst plugin), or whether I am missing something to enable use of the `asset` protocol for the application.

snue (commented):
With debug output enabled, GStreamer is printing a warning about the missing/unknown license. This should be fixed. 

```
WARN      GST_PLUGIN_LOADING gstplugin.c:520:gst_plugin_register_func: plugin "/nix/store/idnib7y1qg9ngy9hm9n11x1jq19vd3ii-tauri-asset-gst-plugin-0.1.0/lib/libgsttauriasset.so" has unknown license "The Plugin's License"
```

Apart from that the plugin loads and does what it is supposed to do. Unfortunately it didn't solve my case, as the app that gives me issues uses a different URI scheme (see below).

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
aaalloc: Could you explain how did you setup your code for using tauri:// ? And yes this pr is for #3725, for asset://, your supposed to use absolute path

snue: I'm on NixOS, trying to package `pawn-appetit` using existing wrappers for building Rust/Tauri applications. This may be something done by those packaging scripts in the background, or it is from the original code base. I am not very familiar with the internal workings of either these or the application code base yet. I definitely didn't add any custom protocol handling myself.

I am still trying to get it to work and this PR got me quite the way to at least understand what is going on. I'll post an update when I figure out where the `tauri://` protocol usage comes from.

FabianLars: tauri:// is what tauri uses to ship your frontendDist files so i assume your frontend has audio files.
asset:// is to load files from the file system.

snue: Yes, that application bundles the sounds in the frontend files. But the request to play still ends up with GStreamer? No idea how that is supposed to work.

I decided to try bundling the files as resources instead, and get the absolute path with `resolveResource`. When I then adapt this plugin to handle `tauri://`, it actually plays most of the sounds! But I see scary warnings from GStreamer and even experienced some hangs/slowness in the program.

```
1:20:36.591930192 1560551 0x559624405f80 WARN                   pulse pulsesink.c:616:gst_pulseringbuffer_open_device:<autoaudiosink335-actual-sink-pulse> error: Failed to connect: Internal error
1:20:36.599630444 1560551 0x7fa37000e690 WARN                    alsa conf.c:5695:snd_config_expand: alsalib error: Unknown parameters {AES0 0x02 AES1 0x82 AES2 0x00 AES3 0x02}
1:20:36.599652175 1560551 0x7fa37000e690 WARN                    alsa pcm.c:2793:snd_pcm_open_noupdate: alsalib error: Unknown PCM default:{AES0 0x02 AES1 0x82 AES2 0x00 AES3 0x02}
1:20:36.605221065 1560551 0x7fa3700120b0 WARN            webkitcommon GStreamerCommon.cpp:253:getStreamIdFromPad:<mpg123audiodec328:sink> Got invalid stream-id from pad: e409256de6db5a9ce9e0bbb6611d5c75edf69d6e0d4d7bcc0a33bd68457abb3e
1:20:36.605245361 1560551 0x7fa3700120b0 WARN       webkitmediaplayer MediaPlayerPrivateGStreamer.cpp:3363:operator():<media-player-335> Caps event received before stream-start. This shouldn't happen!
```

When I pass the path from `resolveResource` through `convertFileSrc` as well, I get a mostly percent-escaped path using the `asset` protocol (it starts with `asset://localhost/%2F...`). But that doesn't seem to trigger the plugin for me and it doesn't play any sound. I'm still investigating whether the plugin actually gets to see the request in that case.

aaalloc: > But the request to play still ends up with GStreamer? No idea how that is supposed to work.

Yes, 

note that that when you are in dev and doing 
``` 
const audio = new Audio("someaudio.mp3")
``` 
the audio is played through http url (`http://localhost..../someaudio.mp3`)
and not `tauri://`

> I decided to try bundling the files as resources instead, and get the absolute path with resolveResource. When I then adapt this plugin to handle tauri://, it actually plays most of the sounds!

As it should be with `asset://`, but now I guess I should add `tauri://` to handle also but I still not quite understand if `tauri://` might in the scope of this pr or not, what do you think @FabianLars ? 

> When I pass the path from resolveResource through convertFileSrc as well, I get a mostly percent-escaped path using the asset protocol (it starts with asset://localhost/%2F...).

That's ok

> But that doesn't seem to trigger the plugin for me and it doesn't play any sound.

I don't have any info of what you did so hard to help

snue: [...]
> 
> note that that when you are in dev and doing
> 
> ```
> const audio = new Audio("someaudio.mp3")
> ```
> 
> the audio is played through http url (`http://localhost..../someaudio.mp3`) and not `tauri://`

That was probably the intention of the code. I never set up a dev environment for this, just trying the packaged result. Should `Audio()` work with `asset://` URIs though? It does not look like the request ends up with GStreamer if I use it that way. It never attempts to load the plugin and does not complain about the missing `asset://` handler either. That only works for the plain absolute paths that end up with `tauri://` URIs.

These absolute paths would probably break in the dev environment then, trying to access `http://localhost.../absolute/path`. So I guess the plugin shouldn't handle `tauri://` here, as it cannot deal with relative paths easily. And the absolute paths should always use `asset://`. I can always patch it up for NixOS as I need it, but would be helpful to know how to play Audio correctly :wink:

I do not want to derail the discussion here about the code issues of the specific application, though. Not sure if the gstreamer warnings are a concern. I usually don't run apps with `GST_DEBUG=2`. The sound generally works fine now for me in the hacked up version.

aaalloc: > These absolute paths would probably break in the dev environment then, trying to access http://localhost.../absolute/path

Yes that is normal because root would be not root of your fs, but root of the webserver running

> Should Audio() work with asset:// URIs though?

Well it depends on what src you give to `Audio`, 
```js
      const path = "asset:///home/../streaming_example_test_video.mp4";
      const path = convertFileSrc("/home/../streaming_example_test_video.mp4");
      // not a sound but a video, but it still get audio so don't mind about that detail
      const audioTest = new Audio(path);
      audioTest.play();
``` 
this works , but from what we've seen if no protocol is specified, it falls back to `tauri://` so question can be made if this should be handled or not and if the user could or not use audio/video path like that. 

> But I see scary warnings from GStreamer and even experienced some hangs/slowness in the program.

for testing I tried with `cargo run --example streaming` (with some  authorization tweaks in tauri.conf.json for `asset://` that you maybe should also apply for your case) and I don't have the warning that you have

> I can always patch it up for NixOS as I need it, but would be helpful to know how to play Audio correctly

You could mix `resolveResource` and `convertFileSrc` like you did and it should work

snue: Indeed, I was just missing `media-src 'self' asset:;` in the `csp` in `tauri.conf.json`. Now I can use `resolveResource` with `convertFileSrc` and it ends up with the plugin using the `asset` protocol. This plays sound, but I still get the gstreamer warnings about invalid stream IDs and others. If you don't see such warnings, I'm willing to attribute this to just my setup. I'll look into the excessive debug logs a bit more to figure out what the issue is.

crates/tauri-asset-gst-plugin/Cargo.toml
snue: (fixed now, thanks)

crates/tauri-asset-gst-plugin/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+fn plugin_init(plugin: &gst::Plugin) -> Result<(), glib::BoolError> {
+  tauri_asset::register(plugin)
+}
+
+gst::plugin_define!(
+  tauriasset,
+  env!("CARGO_PKG_DESCRIPTION"),
+  plugin_init,
+  concat!(env!("CARGO_PKG_VERSION"), "-", env!("COMMIT_ID")),
+  "MIT",
```
snue: This still causes a warning:

```
 WARN      GST_PLUGIN_LOADING gstplugin.c:520:gst_plugin_register_func: plugin "/nix/store/9bjqvs1s16i3dmmyp1kqk7k17fcbxjqk-tauri-asset-gst-plugin-0.1.0/lib/libgsttauriasset.so" has unknown license "MIT"
```

GStreamer calls it `MIT/X11`. Accepted values: https://github.com/GStreamer/gstreamer/blob/1ee723fa0c6fa9e2b7023d0ab8a544d3ac3f2f7e/subprojects/gstreamer/gst/gstplugin.c#L89-L98

aaalloc: arf, thank you thats fixed !

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,104 @@
...
+      .ok_or_else(|| {
+        glib::Error::new(
+          gst::URIError::BadUri,
+          "Could not get src pad from internal source",
+        )
+      })?;
+
+    let ghostpad = GhostPad::new(PadDirection::Src);
+    ghostpad
+      .set_target(Some(&srcpad))
```
snue: With a known target:
```suggestion
    let ghostpad = GhostPad::with_target(&srcpad)
```

Makes the use of `GhostPadExt` and `PadDirection` obsolete.

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,104 @@
...
+        glib::Error::new(
+          gst::URIError::BadUri,
+          "Could not decode percent-encoded URI",
+        )
+      })?
+      .to_string();
+
+    // now location is like: /path/to/asset
+    let internal_src = gst::ElementFactory::make("filesrc")
+      .name("filesrc")
```
snue: If you don't set a name, it should get a guaranteed unique name based on factory and a number. I don't see an immediate issue with the fixed name, but it also doesn't add anything.

snue (commented):
My GStreamer warnings may be decoder related and vary with the chosen sound effect palette in the application. Not an issue with the URI handling plugin as it seems. But eventually I seem to run into issues when playing repeated small sound effects. It may recover and then throw new warnings. 

At some point I reach a critical "could not open socket" because of too many open file descriptors.

```
0:13:17.522676433 524169 0x565202fa9830 WARN                GST_POLL gstpoll.c:724:gst_poll_new: 0x565203c6ccc0: can't create socket pair !

(WebKitWebProcess:524169): GStreamer-CRITICAL **: 18:59:55.319: gst_poll_get_read_gpollfd: assertion 'set != NULL' failed
shared memfd open() failed: Too many open files
Failed to create secure directory (/run/user/1000/pulse): Too many open files
```

Maybe it is not a good idea to create a new `filesrc` Element with `GhostPad` for the same asset over and over. Or does it leak fds? I have nothing to prove that, though. It could also be fallout of the previous warnings.

Otherwise just minor code suggestions. Really nice to have some sound output finally!

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
aaalloc: Fixed on [a1b2e3e](https://github.com/tauri-apps/tauri/pull/14402/commits/a1b2e3ea6cc85144854f0c00ab2a0f34e0eb0aa0) i justed removed the name you are right it doesn't help that much

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
aaalloc: done in [a1b2e3e](https://github.com/tauri-apps/tauri/pull/14402/commits/a1b2e3ea6cc85144854f0c00ab2a0f34e0eb0aa0) 

thank you for the suggestion thats better !

aaalloc:
@snue 
> My GStreamer warnings may be decoder related and vary with the chosen sound effect palette in the application. Not an issue with the URI handling plugin as it seems. But eventually I seem to run into issues when playing repeated small sound effects. It may recover and then throw new warnings.
> 
> At some point I reach a critical "could not open socket" because of too many open file descriptors.
> 
> ```
> 0:13:17.522676433 524169 0x565202fa9830 WARN                GST_POLL gstpoll.c:724:gst_poll_new: 0x565203c6ccc0: can't create socket pair !
> 
> (WebKitWebProcess:524169): GStreamer-CRITICAL **: 18:59:55.319: gst_poll_get_read_gpollfd: assertion 'set != NULL' failed
> shared memfd open() failed: Too many open files
> Failed to create secure directory (/run/user/1000/pulse): Too many open files
> ```
> 
> Maybe it is not a good idea to create a new `filesrc` Element with `GhostPad` for the same asset over and over. Or does it leak fds? I have nothing to prove that, though. It could also be fallout of the previous warnings.

I've reproduced the same error while creating many HTML Audio, filesrc element should not be recreated if the element has already been created you are right, then I should also implement 
```rust
impl ObjectImpl for TauriAsset {
  fn dispose(&self) {
    // Cleanup filesrc
    ...
  }
}
``` 
after the audio is removed from DOM

If you have any suggestion handling that properly feel free to tell them ! I will try to propose something when I will have something that will satisfy me and handle the problem nicely

snue:
> If you have any suggestion handling that properly feel free to tell them ! I will try to propose something when I will have something that will satisfy me and handle the problem nicely

FWIW, I found the upstream `gst-plugins-rs` package has a [`FileSrc` util](https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/blob/main/generic/file/src/filesrc/imp.rs#L339-368) in the generic `file` plugin. I could patch its `URIHandlerImpl`  to also accept `asset://` and replace it with `file://`. I still see the same gstreamer warnings with that. The application creates throw-away Audio objects, which doesn't play nice in either way. And eventually that patched `FileSrc` runs into the same file descriptor exhaustion.

snue:
> FWIW, I found the upstream `gst-plugins-rs` package has a [`FileSrc` util](https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/blob/main/generic/file/src/filesrc/imp.rs#L339-368) in the generic `file` plugin. I could patch its `URIHandlerImpl` to also accept `asset://` and replace it with `file://`. I still see the same gstreamer warnings with that. The application creates throw-away Audio objects, which doesn't play nice in either way. And eventually that patched `FileSrc` runs into the same file descriptor exhaustion.

Creating new Audio elements caused the `WebKitWebProcess` to increase by about 30 open file descriptors for each new gstreamer pipeline. These seem to be cleaned up correctly already, but not fast enough in all cases. I don't think the resource handling of this plugin is to blame, as the issue is already with the newly created pipelines.

I patched the application to only use a single `Audio` element and update its `.src` instead. This works surprisingly well. No more intermittent application hangs from excessive sound playing and the `WebKitWebProcess` `fd` count stays completely under control. I still see some gstreamer warnings from playing sound. I figured out at least some were caused by having `pavucontrol` watch what is going on. Now I only see occasional underflows on ever increasing `autoaudiosink<nnn>-actual-sink-pulse` that are reliably caused by certain sound files.

My latest tests were with the patched `FileSrc`, but I also re-tested the patched application version with the latest commits on this branch, and it works just as well! But the invalid stream ID warnings are back with that:

```
0:04:17.323207234 1764930 0x7f8a24000e20 WARN            webkitcommon GStreamerCommon.cpp:253:getStreamIdFromPad:<mpg123audiodec198:sink> Got invalid stream-id from pad: f758182a86618c2fdcf30ce6b37d6287834a9706f737cc7d11dfdf3082fc53d6
0:04:17.323230658 1764930 0x7f8a24000e20 WARN       webkitmediaplayer MediaPlayerPrivateGStreamer.cpp:3363:operator():<media-player-198> Caps event received before stream-start. This shouldn't happen!
0:04:17.325467266 1764930 0x555f010b6260 WARN            webkitcommon GStreamerCommon.cpp:253:getStreamIdFromPad:<inputselector198:sink_0> Got invalid stream-id from pad: f758182a86618c2fdcf30ce6b37d6287834a9706f737cc7d11dfdf3082fc53d6
0:04:17.346366999 1764930 0x555f010b6260 WARN            webkitcommon GStreamerCommon.cpp:253:getStreamIdFromPad:<inputselector198:sink_0> Got invalid stream-id from pad: f758182a86618c2fdcf30ce6b37d6287834a9706f737cc7d11dfdf3082fc53d6
```

As the sound still plays, I suspect it's something the specific decoder is picky about. But I don't see these with the patched `FileSrc` (which is a `BaseSrc`, not a `Bin` with `GhostPad`).

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,101 @@
...
+      .ok()
+      .ok_or_else(|| glib::Error::new(gst::URIError::BadUri, "Could not create ghost pad"))?;
+
+    ghostpad
+      .set_active(true)
+      .ok()
+      .ok_or_else(|| glib::Error::new(gst::URIError::BadUri, "Could not activate ghost pad"))?;
+
+    let element = self.obj();
+    element.add_pad(&ghostpad).expect("Failed to add ghost pad");
```
snue: Adding the pad to the element before activating fixes the `invalid stream-id` and `caps before stream start` warnings I have seen. 

```suggestion

    element.add_pad(&ghostpad).expect("Failed to add ghost pad");

    ghostpad
      .set_active(true)
      .ok()
      .ok_or_else(|| glib::Error::new(gst::URIError::BadUri, "Could not activate ghost pad"))?;

```

aaalloc: Oh nice, I will check and update this when I will have time !

aaalloc:
@snue 
> I patched the application to only use a single Audio element and update its .src instead. This works surprisingly well. No more intermittent application hangs from excessive sound playing and the WebKitWebProcess fd count stays completely under control. I still see some gstreamer warnings from playing sound. I figured out at least some were caused by having pavucontrol watch what is going on. Now I only see occasional underflows on ever increasing autoaudiosink<nnn>-actual-sink-pulse that are reliably caused by certain sound files.

My latest tests were with the patched FileSrc, but I also re-tested the patched application version with the latest commits on this branch, and it works just as well! 

If I understand correctly, source code of gstreamer-rs has to be changed ? 

> As the sound still plays, I suspect it's something the specific decoder is picky about. But I don't see these with the patched FileSrc (which is a BaseSrc, not a Bin with GhostPad).

With what you've tested and said before, do you think that theres maybe a better approach for FileSrc+Bin+Ghostpad in our case ?

snue:
> My latest tests were with the patched FileSrc, but I also re-tested the patched application version with the latest commits on this branch, and it works just as well!
> 
> If I understand correctly, source code of gstreamer-rs has to be changed ?

Not `gstreamer` or the `gstreamer-rs` bindings. The `gst-plugins-rs` is just an upstream collection of plugins written in rust. And the contained `file` plugin provides the `GstRsFileSrc`, which can be easily patched to support `asset://` next to `file://`. This is an alternative to the plugin here, but not necessarily better. Not sure if the upstream repository would want to implement it like that. I mainly did it to see how that behaves in comparison regarding file descriptors and other resources.

> 
> > As the sound still plays, I suspect it's something the specific decoder is picky about. But I don't see these with the patched FileSrc (which is a BaseSrc, not a Bin with GhostPad).
> 
> With what you've tested and said before, do you think that theres maybe a better approach for FileSrc+Bin+Ghostpad in our case ?

When you reorder the activation of the ghostpad with adding it to the element (see https://github.com/tauri-apps/tauri/pull/14402#discussion_r2649182527), both plugins do exactly the same thing for me. That is, it works really well as long as I don't spam it with `Audio` elements. Getting this plugin into tauri seems more appropriate to me than hacking the upstream file plugin.

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
snue: You can then also omit the `?;` and use this as function return value directly instead of the `Ok(())` below.

snue:
Heads up, with recent WebKitGTK you additionally need to allow the `asset` protocol through another environment variable for media playback:

`WEBKIT_GST_ALLOWED_URI_PROTOCOLS="asset"`

See https://github.com/WebKit/WebKit/commit/7e871b506eaba52c7d74519aeb0047d702101331

aaalloc:
> Heads up, with recent WebKitGTK you additionally need to allow the `asset` protocol through another environment variable for media playback:
> 
> `WEBKIT_GST_ALLOWED_URI_PROTOCOLS="asset"`
> 
> See [WebKit/WebKit@7e871b5](https://github.com/WebKit/WebKit/commit/7e871b506eaba52c7d74519aeb0047d702101331)

Good to know, that is something to set in packaging phase (in `tauri-bundler` crates). Talking about `tauri-bundler`, this plugin has to be packed like [nsis-tauri-utils](https://github.com/tauri-apps/nsis-tauri-utils), when I created that PR I was not aware of this so I don't know how you @FabianLars or anyone else want to manage that.

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,145 @@
...
+  fn constructed(&self) {
+    self.parent_constructed();
+
+    let element = self.obj();
+    let filesrc = gst::ElementFactory::make("filesrc")
+      .build()
+      .unwrap_or_else(|err| {
+        gst::error!(CAT, imp = self, "Failed to create filesrc element: {err}");
+        panic!("Creating filesrc element failed");
+      });
```
snue: You can avoid the `panic` and return instead. It requires an explicit `match` statement or the [`unwrap_or` crate macros](https://crates.io/crates/unwrap_or), because the `?` shortcut does not work for void functions (that don't return anything), and `return` wouldn't work in the closure.


```suggestion
    let filesrc = match gst::ElementFactory::make("filesrc").build()
      {
        Ok(src) => src,
        Err(err) => {
          gst::error!(CAT, imp = self, "Failed to create filesrc element: {err}");
          return;
        },
      };
```

Same for the other `panic!`s in this function. The `set_uri` already deals nicely with the `filesrc` not being valid, but that will never happen when you already panic here.

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,145 @@
...
+    gst::debug!(CAT, imp = self, "URI from \"{}\" to \"{}\"", uri, &location);
+
+    self
+      .filesrc
+      .get()
+      .ok_or_else(|| {
+        let msg = "filesrc element is not initialized";
+        gst::error!(CAT, imp = self, "{msg}");
+        glib::Error::new(gst::URIError::BadUri, &msg)
+      })
```
snue: I was shortly wondering why this isn't `map_err()` as above. I think you achieve the same with `map_err` and simply not using `?` here. That way you may be able to reuse the actual error message as well.

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,145 @@
...
+      .map(|filesrc| {
+        filesrc.set_property("location", &location);
```
snue: You should return the result of the `set_property`.


```suggestion
        filesrc.set_property("location", &location)
```

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
aaalloc: done in [2706cd3](https://github.com/tauri-apps/tauri/pull/14402/commits/2706cd33c356251d72c82d9f7527e81db367e0be)

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
aaalloc: in last commit I wrapped the result of `set_property("location", &location)` and propagated the error with ?, what do you think about this way ?

snue: Sorry, I was mistaken about the signature of `set_property()`. That simply panics on issues:
https://gtk-rs.org/gtk-rs-core/git/docs/glib/object/trait.ObjectExt.html#tymethod.set_property

So I didn't expect the need to wrap in `Ok`, but yes, that's fine (except you don't need the `return` when you leave out the final semicolon, the last statement is the implicit return value). The code before was maybe even more readable as now you have the `?` nested within the `Ok()`. I'd call that a matter of taste in the end. It looks like it does the right thing, and the compiler probably figures out the rest. If `cargo clippy` doesn't complain, I'm fine with the code.

I didn't try or build/check the latest version locally yet, but I'll do so shortly.

snue: > [!TIP]
> This may be a perfect opportunity for exploring https://rustcombinators.com/ to solve this with an elegant chain from what `filesrc().get()` returns to what `set_uri()` must return, propagating any errors encountered along the way.

For completeness, this goes from `Option<T>` to `Result<T,E>` to `Result<(), E>`, so `.ok_or_else()` is perfect here, but could be combined with `map()` to avoid the extra `Ok()` wrapping. It could alternatively do `Option<T>` to `Option<()>` to `Result<(), E>`. with `.map(...).ok_or_else(...)`.

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    gst::debug!(CAT, imp = self, "URI from \"{}\" to \"{}\"", uri, &location);
+
+    return Ok(self
+      .filesrc
+      .get()
+      .ok_or_else(|| {
+        let msg = "filesrc element is not initialized";
+        gst::error!(CAT, imp = self, "{msg}");
+        glib::Error::new(gst::URIError::BadUri, &msg)
```
snue: clippy found this reference to be unneeded (this isn't `String` as the others)
:
```
warning: this expression creates a reference which is immediately dereferenced by the compiler
   --> crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs:162:49
    |
162 |         glib::Error::new(gst::URIError::BadUri, &msg)
    |                                                 ^^^^ help: change this to: `msg`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow
    = note: `#[warn(clippy::needless_borrow)]` on by default
```

```suggestion
        glib::Error::new(gst::URIError::BadUri, msg)
```

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,166 @@
...
+      })?.set_property("location", &location)
+    );
```
snue: Clippy doesn't like this either, but it's suggestion is pretty horrendous for once.

```
warning: passing a unit value to a function
   --> crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs:156:12
    |
156 |       return Ok(self
    |  ____________^
157 | |       .filesrc
158 | |       .get()
159 | |       .ok_or_else(|| {
...   |
163 | |       })?.set_property("location", &location)
164 | |     );
    | |_____^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unit_arg
    = note: `#[warn(clippy::unit_arg)]` on by default
help: move the expression in front of the call and replace it with the unit literal `()`
    |
156 ~     return {
157 +         let _: () = self
158 +         .filesrc
159 +         .get()
160 +         .ok_or_else(|| {
161 +           let msg = "filesrc element is not initialized";
162 +           gst::error!(CAT, imp = self, "{msg}");
163 +           glib::Error::new(gst::URIError::BadUri, &msg)
164 +         })?.set_property("location", &location);
165 +         Ok(()
166 +         )
167 ~     };
```

This looks nice to me (and satisfies clippy):

```suggestion
   self
      .filesrc
      .get()
      .map(|src| src.set_property("location", &location))
      .ok_or_else(|| {
        let msg = "filesrc element is not initialized";
        gst::error!(CAT, imp = self, "{msg}");
        glib::Error::new(gst::URIError::BadUri, msg)
      })
```

(it's pretty similar to your first attempt)

snue (approved):
I tested the latest version and it works! Produces sound for me reliably. 🎵 

Clippy has some complaints and you could also run `rustfmt` over the code to fix some minor style issues. Otherwise great work! Thanks!

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
aaalloc: done in [bc502b4](https://github.com/tauri-apps/tauri/pull/14402/commits/bc502b45a6995aab5687ce3e50b3d218ca5ee105)

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
aaalloc: > Tip
> 
> This may be a perfect opportunity for exploring https://rustcombinators.com/ to solve this with an elegant chain from what `filesrc().get()` returns to what `set_uri()` must return, propagating any errors encountered along the way.
> 
> For completeness, this goes from `Option<T>` to `Result<T,E>` to `Result<(), E>`, so `.ok_or_else()` is perfect here, but could be combined with `map()` to avoid the extra `Ok()` wrapping. It could alternatively do `Option<T>` to `Option<()>` to `Result<(), E>`. with `.map(...).ok_or_else(...)`.

really really nice site, thank you for the sharing !

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,164 @@
...
+      }
+    };
+
+    match filesrc.static_pad("src") {
+      Some(pad) => pad,
+      None => {
+        gst::error!(CAT, imp = self, "filesrc has no src pad");
+        return;
+      }
+    };
```
rramphal: Why is this block here twice? The second one seems to be dead code since nothing is ever done with the result.

```suggestion
```

crates/tauri-asset-gst-plugin/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "tauri-asset-gst-plugin"
+version = "0.1.0"
+authors = ["Alexander Yanovskyy <contact@yanovskyy.com>"]
+license = "MIT OR Apache-2.0"
+edition = "2018"
```
rramphal: To match [`tauri`'s edition](https://github.com/tauri-apps/tauri/blob/dev/Cargo.toml):
```suggestion
edition = "2021"
```

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,164 @@
...
+impl URIHandlerImpl for TauriAsset {
+  const URI_TYPE: gst::URIType = gst::URIType::Src;
+
+  fn protocols() -> &'static [&'static str] {
+    &[ASSET_URI_SCHEME]
+  }
+
+  fn uri(&self) -> Option<String> {
+    None
+  }
```
rramphal, line 147: The `URIHandler` interface expects `uri()` to return the current URI. Returning `None` prevents downstream elements and the application from successfully performing URI queries to identify the active resource.

```suggestion
fn uri(&self) -> Option<String> {
    self.filesrc
        .get()
        .and_then(|src| src.property::<Option<String>>("location"))
        .filter(|path| !path.is_empty())
        .map(|path| format!("{}://{}", ASSET_URI_SCHEME, path))
}
```

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,164 @@
...
+#[derive(Default)]
+pub struct TauriAsset {
+  pub filesrc: OnceLock<gst::Element>,
+}
+
+impl TauriAsset {
+  pub fn new() -> Self {
+    Self::default()
+  }
+}
```
rramphal: Technically, because the struct derives `Default`, `TauriAsset::default()` works, so the `new()` doesn't add anything. There's no harm in keeping it, but I believe you could just remove it.


```suggestion
```

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,164 @@
...
+use gst::glib::subclass::types::ObjectSubclassExt;
+use gst::prelude::{ElementExt, GstBinExt, PadExt};
+use gst::subclass::prelude::{BinImpl, ElementImpl, ObjectImpl, URIHandlerImpl};
+use gst::subclass::prelude::{GstObjectImpl, ObjectSubclass};
+use gst::{glib, GhostPad};
+use gstreamer::glib::object::ObjectExt;
+use gstreamer::glib::subclass::object::ObjectImplExt;
+use gstreamer_base::gst;
+
+// GST_DEBUG=tauri_asset:5 ...
```
rramphal: Did you mean to leave this in?

```suggestion
```

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,164 @@
...
+impl GstObjectImpl for TauriAsset {}
+impl ObjectImpl for TauriAsset {
+  fn constructed(&self) {
+    self.parent_constructed();
+
+    let element = self.obj();
+    let filesrc = match gst::ElementFactory::make("filesrc").build() {
+      Ok(src) => src,
+      Err(err) => {
+        gst::error!(CAT, imp = self, "Failed to create filesrc element: {err}");
```
rramphal: Log to GStreamer debug logs *and* post an error on the Bus so the application can receive construction failures:

```suggestion
        gst::element_error!(element, gst::LibraryError::Init, ("Failed to create filesrc: {}", err));
```

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,164 @@
...
+      }
+    };
+
+    match element.add(&filesrc) {
+      Ok(_) => (),
+      Err(err) => {
+        gst::error!(CAT, imp = self, "Failed to add filesrc to bin: {err}");
+        return;
+      }
+    };
```
rramphal, line 65: Log to GStreamer debug logs *and* post an error on the Bus (while simplifying the syntax since only error case matters):

```suggestion
    if let Err(err) = element.add(&filesrc) {
        gst::element_error!(element, gst::LibraryError::Init, ("Failed to add filesrc to bin: {}", err));
        return;
    }
```

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    let srcpad = match filesrc.static_pad("src") {
+      Some(pad) => pad,
+      None => {
+        gst::error!(CAT, imp = self, "filesrc has no src pad");
```
rramphal: Log to GStreamer debug logs *and* post an error on the Bus:

```suggestion
        gst::element_error!(element, gst::LibraryError::Init, ("filesrc has no src pad"));
```

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,164 @@
...
+        return;
+      }
+    };
+
+    let ghostpad = match GhostPad::with_target(&srcpad) {
+      Ok(pad) => pad,
+      Err(err) => {
+        gst::error!(CAT, imp = self, "Failed to create ghost pad: {err}");
```
rramphal: Log to GStreamer debug logs *and* post an error on the Bus:

```suggestion
        gst::element_error!(element, gst::LibraryError::Init, ("Failed to create ghost pad: {}", err));
```

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,164 @@
...
+      }
+    };
+
+    match element.add_pad(&ghostpad) {
+      Ok(_) => (),
+      Err(err) => {
+        gst::error!(CAT, imp = self, "Failed to add ghost pad to bin: {err}");
+        return;
+      }
+    };
```
rramphal, line 101: Log to GStreamer debug logs *and* post an error on the Bus (while simplifying the syntax since only error case matters):

```suggestion
    if let Err(err) = element.add_pad(&ghostpad) {
        gst::element_error!(element, gst::LibraryError::Init, ("Failed to add ghost pad to bin: {}", err));
        return;
    }
```

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    match ghostpad.set_active(true) {
+      Ok(_) => (),
+      Err(err) => {
+        gst::error!(CAT, imp = self, "Failed to activate ghost pad: {err}");
+        return;
+      }
+    };
```
rramphal, line 113: Log to GStreamer debug logs and post an error on the Bus (while simplifying the syntax since only error case matters):

```suggestion
    if let Err(err) = ghostpad.set_active(true) {
        gst::element_error!(element, gst::LibraryError::Init, ("Failed to activate ghost pad: {}", err));
        return;
    }
```

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    match self.filesrc.set(filesrc) {
+      Ok(_) => (),
+      Err(_) => {
+        gst::error!(CAT, imp = self, "Failed to set filesrc OnceLock");
+        return;
+      }
+    }
```
rramphal, line 125: Log to GStreamer debug logs *and* post an error on the Bus (while simplifying the syntax since only error case matters and the error is not useful):

```suggestion
    if self.filesrc.set(filesrc).is_err() {
        gst::element_error!(element, gst::LibraryError::Init, ("filesrc OnceLock already set"));         
        return;
    }
```

crates/tauri-asset-gst-plugin/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+path = "src/lib.rs"
+
+[build-dependencies]
+gst-plugin-version-helper = "0.8.3"
+
+[profile.release]
+lto = true
+codegen-units = 1
+opt-level = "z"
+panic = 'abort'
```
rramphal: Silence `No newline at end of file` warning:

```suggestion
panic = 'abort'

```

rramphal (changes requested):
Hey @aaalloc, thank you for your work -- I'm eagerly awaiting for it to be merged in!

I left some comment suggesting changes hoping some would be helpful to you. I'm no expert, so apologies for any errors. Also some of the indentation might be off, but nothing `cargo fmt` can't fix. Also, `cargo clippy` would be good.

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
rramphal: See [comment below](https://github.com/tauri-apps/tauri/pull/14402/changes#r2744321235).

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
aaalloc: Yeah you are right it might a copy past that I forget to remove, thank you !

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
aaalloc: Yes i meaned it, but it isn't as explicit as it would. That was for someone who would wanted to debug this and tell to the reader that he could debug the plugin using that variable as for example `GST_DEBUG=tauri_asset:5 pnpm tauri dev`

aaalloc:
@rramphal thank you for your review !

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,184 @@
...
+use gst::prelude::{ElementExt, GstBinExt, PadExt};
+use gst::subclass::prelude::{BinImpl, ElementImpl, ObjectImpl, URIHandlerImpl};
+use gst::subclass::prelude::{GstObjectImpl, ObjectSubclass};
+use gst::{glib, GhostPad};
+use gstreamer::glib::object::ObjectExt;
+use gstreamer::glib::subclass::object::ObjectImplExt;
+use gstreamer_base::gst;
+
+// for debugging
+// GST_DEBUG=tauri_asset:5 <your-binary-using-the-plugin>
```
rramphal, line 13: @aaalloc I was thinking that if I were working on this plugin, I'd expect this kind of documentation in the README.md. I noticed that you didn't add a README file yet, but I think it might be a good opportunity to explain the purpose of this plugin and provide exactly the type of guidance you're trying to document here.

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -0,0 +1,184 @@
...
+    self.filesrc.get().and_then(|src| {
+      src
+        .property::<Option<String>>("location")
+        .filter(|p| !p.is_empty())
+        .map(|p| format!("{}://{}", ASSET_URI_SCHEME, p))
+    })
+  }
+
+  fn set_uri(&self, uri: &str) -> Result<(), glib::Error> {
+    // uri is like: asset://path/to/asset or asset://localhost/path/to/asset
```
rramphal, line 150: It would be great to add some unit tests here to make sure these assumptions are correct. You can check happy-path valid URIs, paths that are percent-encoded, paths prefixed with `localhost`, malformed URIs, etc.

crates/tauri-asset-gst-plugin/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+version = "0.1.0"
+authors = ["Alexander Yanovskyy <contact@yanovskyy.com>"]
+license = "MIT OR Apache-2.0"
+edition = "2021"
+repository = "https://github.com/tauri-apps/tauri"
+description = "Gstreamer plugin for asset protocol"
+
+[dependencies]
+gstreamer = "0.24"
+gstreamer-base = "0.24"
```
rramphal: You have `use gstreamer_base::gst;`, but `gstreamer-base` just re-exports `gst` from `gstreamer` anyway which you're already importing on line 11. If I'm not mistaken, you could get rid of this dependency and update the code by putting `use gstreamer as gst` at the top instead.

```suggestion
```

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
aaalloc: done in [27ac8ca](https://github.com/tauri-apps/tauri/pull/14402/commits/27ac8ca148cd3ad4f558e8ff0797e7c121386d5f), tell me if i have forgot a test

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
aaalloc: readme added in [27ac8ca](https://github.com/tauri-apps/tauri/pull/14402/commits/27ac8ca148cd3ad4f558e8ff0797e7c121386d5f), hope it is better this way

crates/tauri-asset-gst-plugin/tests/test_mono_s16.flac
aaalloc, line 1: took from https://github.com/GStreamer/gst-plugins-rs/blob/main/audio/claxon/tests/test_mono_s16.flac

crates/tauri-asset-gst-plugin/src/tauri_asset/imp.rs
```diff
@@ -150,7 +150,7 @@ impl URIHandlerImpl for TauriAsset {
     // uri is like: asset://path/to/asset or asset://localhost/path/to/asset
     let sep = format!("{}://", ASSET_URI_SCHEME);
     let mut split = uri.split(sep.as_str());
-    let location = split.nth(1).ok_or_else(|| {
+    let location = split.nth(1).filter(|l| !l.is_empty()).ok_or_else(|| {
```
aaalloc, line 153: didn't handled the case the string could be only asset://, then the location was a empty string before the filter was added

rramphal (approved):

rramphal:
Hey @aaalloc, I know you said on your [blog post](https://yanovskyy.com/blog/en/tauri-webkit) that you're coordinating getting this merged on Discord? Are there any updates on this?

aaalloc:
> Hey @aaalloc, I know you said on your [blog post](https://yanovskyy.com/blog/en/tauri-webkit) that you're coordinating getting this merged on Discord? Are there any updates on this?

Hi ! It will be merged when @FabianLars will have more time to tackle this PR.

FabianLars:
Yep, this delay is totally my fault. Drowning in backlog atm 😂 For now, thanks to the author and the awesome community involvement! I hope i can get to it soon...

PR #14408, opened 2025-11-03
fix(bundler/cli): set user-agent when fetching build tools

FabianLars:


crates/tauri-bundler/src/utils/http_utils.rs
```diff
@@ -47,30 +49,36 @@ fn generate_github_alternative_url(url: &str) -> Option<(ureq::Agent, String)> {
 
   generate_github_mirror_url_from_template(url)
     .or_else(|| generate_github_mirror_url_from_base(url))
-    .map(|alt_url| (ureq::agent(), alt_url))
+    .map(|alt_url| {
+      (
+        ureq::Agent::config_builder()
```
FabianLars, line 54: we may want to add the platform certs here as well but nobody complained so far so idk

Legend-Master: Any reason not to use `base_ureq_agent` here?

crates/tauri-bundler/src/utils/http_utils.rs
```diff
@@ -47,30 +49,36 @@ fn generate_github_alternative_url(url: &str) -> Option<(ureq::Agent, String)> {
...
+      )
+    })
 }
 
 fn create_agent_and_url(url: &str) -> (ureq::Agent, String) {
   generate_github_alternative_url(url).unwrap_or((base_ureq_agent(), url.to_owned()))
 }
 
 pub(crate) fn base_ureq_agent() -> ureq::Agent {
+  let mut config_builder = ureq::Agent::config_builder()
```
Legend-Master, line 69:
```suggestion
  #[allow(unused_mut)]
  let mut config_builder = ureq::Agent::config_builder()
```

Probably doesn't really matter though

crates/tauri-cli/src/helpers/http.rs
```diff
@@ -2,23 +2,24 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-License-Identifier: MIT
 
-use ureq::{http::Response, Body};
+use ureq::{http::Response, Agent, Body};
+
+const CLI_USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"),);
 
 pub fn get(url: &str) -> Result<Response<Body>, ureq::Error> {
+  let mut config_builder = ureq::Agent::config_builder()
```
Legend-Master, line 11: Same as https://github.com/tauri-apps/tauri/pull/14408/files#r2488534622

crates/tauri-bundler/src/utils/http_utils.rs
FabianLars: i assumed that setting the proxy may be a bad idea here. but idk i just kept it as-is cause i wasn't involved in that part at all and idk the requirements or expectations.

crates/tauri-bundler/src/utils/http_utils.rs
FabianLars: hmm right, didn't catch that because the feature is enabled by default

crates/tauri-bundler/src/utils/http_utils.rs
Legend-Master: Alright, let's go with this for now and see if anyone using it actually wants to change this

Legend-Master (approved):

Issue #14420, opened 2025-11-05
[bug] State does not `drop` fields at application exit


raphaelmenges:
### Describe the bug

Hello 👋

I noticed that fields in the [managed state](https://v2.tauri.app/develop/state-management) seem to be not dropped at application exit. I created a minimal example based on the template to reproduce the issue:

```rust
use tauri::{async_runtime::Mutex, Manager};

// Learn more about Tauri commands at https://tauri.app/develop/calling-rust/
#[tauri::command]
fn greet(name: &str) -> String {
    format!("Hello, {}! You've been greeted from Rust!", name)
}

pub struct MyType {}

impl MyType {
    pub fn new() -> MyType {
        println!("Create MyType."); // Appears in the output.
        MyType {}
    }
}

impl Drop for MyType {
    fn drop(&mut self) {
        println!("Drop MyType."); // Does not appear in the output.
    }
}

pub struct State {
    pub my_type: MyType,
}

impl State {
    pub fn new() -> State {
        State {
            my_type: MyType::new(),
        }
    }
}

#[cfg_attr(mobile, tauri::mobile_entry_point)]
pub fn run() {
    tauri::Builder::default()
        .setup(move |app| {
            app.manage(Mutex::new(State::new()));
            Ok(())
        })
        .plugin(tauri_plugin_opener::init())
        .invoke_handler(tauri::generate_handler![greet])
        .run(tauri::generate_context!())
        .expect("error while running tauri application");
}
```

Output when ran and exited properly:

```
raphael@Raphaels-MBP tauri-state-drop-bug % npm run tauri dev

> tauri-state-drop-bug@0.1.0 tauri
> tauri dev

     Running BeforeDevCommand (`npm run dev`)

> tauri-state-drop-bug@0.1.0 dev
> vite


  VITE v6.4.1  ready in 100 ms

  ➜  Local:   http://localhost:1420/
     Running DevCommand (`cargo  run --no-default-features --color always --`)
        Info Watching /Users/raphael/Git/tauri-state-drop-bug/src-tauri for changes...
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.32s
     Running `target/debug/tauri-state-drop-bug`
Create MyType.
raphael@Raphaels-MBP tauri-state-drop-bug %
```

### Reproduction

1. Clone <https://github.com/raphaelmenges/tauri-state-drop-bug>.
2. Run `npm install` in the directory.
3. Run `npm run tauri dev` in the directory.

[...]

PR #14423, opened 2025-11-05
fix(core): clear State map on exit

FabianLars:
related to #14420 

This is very fucked up and shouldn't be done right?

### DIFF
--- a/crates/tauri/src/app.rs
+++ b/crates/tauri/src/app.rs
@@ -1037,6 +1037,7 @@ macro_rules! shared_app_impl {
         for (_, webview) in self.manager.webviews() {
           webview.resources_table().clear();
         }
+        unsafe { self.manager.state.clear() };
       }
 
       /// Gets the invoke key that must be referenced when using [`crate::webview::InvokeRequest`].
--- a/crates/tauri/src/state.rs
+++ b/crates/tauri/src/state.rs
@@ -149,6 +149,13 @@ impl StateManager {
     Some(*value)
   }
 
+  /// SAFETY: This will cause all references obtained through [Self::try_get] to dangle.
+  /// This must only be called on app exit (in cleanup_before_exit).
+  pub(crate) unsafe fn clear(&self) {
+    let mut map = self.map.lock().unwrap();
+    map.clear();
+  }
+
   /// Gets the state associated with the specified type.
   pub fn get<T: Send + Sync + 'static>(&self) -> State<'_, T> {
     self


crates/tauri/src/app.rs
```diff
@@ -1037,6 +1037,7 @@ macro_rules! shared_app_impl {
         for (_, webview) in self.manager.webviews() {
           webview.resources_table().clear();
         }
+        unsafe { self.manager.state.clear() };
```
Legend-Master, line 1040: This is probably not what we want, someone could still be holding onto one of the pointers returned from the state manager while calling this (unless we change the states to `Arc`/`Rc`)

I honestly feel like the entire state system is a bit redundant considering it just stores things globally which you can already do with `static` except for maybe multiple tauri instances

FabianLars: Yeah exactly. I don't know how much that matters on app shutdown though.

But I only opened this PR to learn something, not to actually merge this.

raphaelmenges: > I honestly feel like the entire state system is a bit redundant considering it just stores things globally which you can already do with static except for maybe multiple tauri instances

You mean something like this?

```rust
static STATE: OnceLock<std::sync::Mutex<State>> = OnceLock::new();

pub fn get_state() -> &'static std::sync::Mutex<State> {
    STATE.get_or_init(|| {
        std::sync::Mutex::new(State::new())
    })
}
```

Legend-Master: Yeah something like that

WSH032: > This is probably not what we want, since someone could still be holding onto one of the pointers returned from the state manager while calling this (unless we change the states to Arc/Rc).

The [state](https://docs.rs/tauri/latest/tauri/trait.Manager.html#method.state) has the same lifetime as the manager, which to some extent mitigates this issue, because [you shouldn’t be using tauri’s API after exit](https://docs.rs/tauri/latest/tauri/struct.App.html#method.cleanup_before_exit).  
~~Nonetheless, I think this might be a breaking change, as it alters the previous behavior.~~
(Oh, I see — the cyclic references caused the memory leak. Then I think this is reasonable, as long as we note this in the `Manager::state` documentation.)

raphaelmenges: OnceLock appears to [also not drop its contents at the end](https://doc.rust-lang.org/reference/items/static-items.html#r-items.static.lifetime):

> Static items have the static lifetime, which outlives all other lifetimes in a Rust program. Static items do not call [drop](https://doc.rust-lang.org/reference/destructors.html) at the end of the program.

Legend-Master: > The [state](https://docs.rs/tauri/latest/tauri/trait.Manager.html#method.state) has the same lifetime as the manager, which to some extent mitigates this issue, because [you shouldn’t be using tauri’s API after exit](https://docs.rs/tauri/latest/tauri/struct.App.html#method.cleanup_before_exit).

Yeah.. but from looking at how we use them in `RunEvent::Exit`, we can't change this until v3 at least

https://github.com/tauri-apps/plugins-workspace/blob/3019063ae1a3cda689cff6b8b786c8b770b0496b/plugins/single-instance/src/platform_impl/windows.rs#L111-L129

> Static items have the static lifetime, which outlives all other lifetimes in a Rust program. Static items do not call [drop](https://doc.rust-lang.org/reference/destructors.html) at the end of the program.

Yeah, you'll need to do it manually as well...

PR #14443, opened 2025-11-10
Add cleanup_before_exit plugin hook, CLI kills entire app tree

### LINKED ISSUE #14360: [feat] Add kill tree for sidecar
### Describe the problem

My sidecar is a PyInstaller Python app. Because it's a single file how it works is a PyInstaller bootloader creates a child process for my app, and thus the "app" works. There are two issues right now.

1. When I exit the Tauri app, the sidecar does not get killed.
2. When I manually kill the child, its child process is still alive

Well I'd rather just that the Tauri sidecar cleans up for me. However, if that's not possible, I'd like a kill_all method like there used to be?

### Alternatives considered

```rs
struct MyStruct {
fn kill(&mut self) {
    if let Some(child) = self.child.take() {
           kill_process_tree(child.pid()).unwrap()
    }
  }
}


fn kill_process_tree(pid: u32) -> Result<(), String> {
    let mut system = System::new_all();
    system.refresh_all();

    let root_pid = Pid::from_u32(pid);
    let mut to_kill = HashSet::new();

    // Collect all descendant PIDs
    collect_descendants(&system, root_pid, &mut to_kill);
    to_kill.insert(root_pid);

    // Kill all processes (children first, then parent)
    for &pid in &to_kill {
        if let Some(process) = system.process(pid) {
            process.kill_with(Signal::Kill);
        }
    }

    Ok(())
}

fn collect_descendants(system: &System, pid: Pid, descendants: &mut HashSet<Pid>) {
    for (child_pid, process) in system.processes() {
        if let Some(parent_pid) = process.parent() {
            if parent_pid == pid {
                descendants.insert(*child_pid);
                collect_descendants(system, *child_pid, descendants);
            }
        }
    }
}

```

### Additional context

_No response_

sakina1303:
Adds a best-effort process-tree killer and a lightweight runtime sidecar PID registry, ensuring that descendant processes spawned by sidecars are properly terminated when their parent sidecar is killed.

This resolves the core issue described in #14360 by introducing both API and runtime cleanup mechanisms for sidecar process management.

Details:

- tauri::process::kill_process_tree(pid: u32) — POSIX shell & PowerShell based, best-effort.
- Sidecar PID registry via AppHandle::register_sidecar(pid) / unregister_sidecar(pid) for cleanup on exit.
- App::cleanup_before_exit() drains registry and calls the kill helper.
- CLI dev-run now attempts process-tree cleanup when stopping dev children.

Files Changed:

- process.rs: Added kill_process_tree() (POSIX + PowerShell).
- mod.rs: Added sidecar PID tracking + helper methods.
- app.rs: Added register/unregister APIs; updated cleanup.
- desktop.rs: Added kill-tree logic in dev-run path.
- CHANGELOG.md / PR_DRAFT.md: Updated with notes.

Testing:

- [ ✅ ] cargo test -p tauri, all tests passed.
- [ ✅ ] Manual: verified registry and cleanup invoke kill helper.
- Runtime cleanup is opt-in:
   Call app_handle.register_sidecar(child.id() as u32) after spawning.

Fixes #14360


crates/tauri/src/app.rs
```diff
@@ -1037,6 +1045,12 @@ macro_rules! shared_app_impl {
         for (_, webview) in self.manager.webviews() {
           webview.resources_table().clear();
         }
+  let sidecar_pids = self.manager.drain_sidecar_pids();
```
Legend-Master: I think maybe this is better done as a generic API that allows you to register functions to be run in `cleanup_before_exit`?

And since we already have the side car APIs in the shell plugin, maybe it makes sense to have the kill logic moved there?

cc @lucasfernog @FabianLars any thoughts?

crates/tauri-cli/src/interface/rust/desktop.rs
```diff
@@ -28,6 +28,41 @@ pub struct DevChild {
 
 impl DevProcess for DevChild {
   fn kill(&self) -> std::io::Result<()> {
+    if let Some(pid) = self.dev_child.id() {
+      #[cfg(windows)]
+      {
+        use std::process::Command;
+        let ps = format!(
+          "function Kill-Tree {{ Param([int]$ppid); Get-CimInstance Win32_Process | Where-Object {{ $_.ParentProcessId -eq $ppid }} | ForEach-Object {{ Kill-Tree $_.ProcessId }}; Stop-Process -Id $ppid -ErrorAction SilentlyContinue }}; Kill-Tree {}",
```
elibroftw, line 36: I will say there is a `taskkill /T /PID <PID>` which is built into Windows (I have not test in Tauri code)

crates/tauri/src/app.rs
sakina1303: @Legend-Master Implemented the suggested change, kindly have a look now. The sidecar cleanup logic has been moved to the shell plugin, and a generic cleanup_before_exit hook API is now in place. Let me know if any adjustments are needed.

crates/tauri-cli/src/interface/rust/desktop.rs
sakina1303: @elibroftw I’ve documented kill_process_tree in process.rs, making it prefer taskkill /T /PID /F on Windows with a PowerShell fallback. Also updated PR_DRAFT.md and ran cargo check -p tauri, kindly have a look now.

crates/tauri/CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## \[Unreleased]
+
+### Enhancements
```
Legend-Master: We use change files to manage change logs, instead of change the change log directly, take a look at https://github.com/tauri-apps/tauri/blob/dev/.changes/README.md

crates/tauri/PR_DRAFT.md
```diff
@@ -0,0 +1,54 @@
+Title: Use plugin cleanup hooks for sidecar shutdown; add kill-tree helper (Fixes #14360)
```
Legend-Master: Also this is probably not supposed to be part of the PR 🙃

crates/tauri/src/plugin.rs
```diff
@@ -104,6 +104,15 @@ pub trait Plugin<R: Runtime>: Send {
...
   fn on_event(&mut self, app: &AppHandle<R>, event: &RunEvent) {}
 
+  /// Callback invoked when the application is performing cleanup before exit.
+  ///
+  /// Plugins can use this hook to perform any process shutdown/cleanup they need
+  /// to do before the runtime exits (for example, killing sidecars or stopping
+  /// background tasks). This is guaranteed to run from the thread performing
+  /// the app cleanup/exit sequence.
+  #[allow(unused_variables)]
+  fn cleanup_before_exit(&mut self, app: &AppHandle<R>) {}
```
Legend-Master, line 114: We would also want to implement this for the `TauriPlugin` so that `tauri::plugin::Builder` could have access to this API

crates/tauri/src/app.rs
Legend-Master: This is great 🙃 We have thought about this before in https://github.com/tauri-apps/plugins-workspace/issues/1014#issuecomment-1994857440 but never actually worked on it

Legend-Master (commented):
Thanks for working on this!

crates/tauri/src/plugin.rs
```diff
@@ -104,6 +104,15 @@ pub trait Plugin<R: Runtime>: Send {
   #[allow(unused_variables)]
   fn on_event(&mut self, app: &AppHandle<R>, event: &RunEvent) {}
 
+  /// Callback invoked when the application is performing cleanup before exit.
+  ///
+  /// Plugins can use this hook to perform any process shutdown/cleanup they need
+  /// to do before the runtime exits (for example, killing sidecars or stopping
+  /// background tasks). This is guaranteed to run from the thread performing
```
Legend-Master: > This is guaranteed to run from the thread performing the app cleanup/exit sequence.

This is a bit confusing to me, let's just say it will be run inside `App::cleanup_before_exit`

crates/tauri/src/process.rs
```diff
@@ -128,3 +128,96 @@ fn restart_macos_app(current_binary: &std::path::Path, env: &Env) {
...
+/// the previous implementation.
+///
+/// On Unix-like systems a small shell function using `pgrep -P` is used to
+/// collect child PIDs and send `SIGKILL` to descendants and the root PID.
+///
+/// Note: terminating processes is inherently best-effort and may fail for
+/// protected or system processes, or when the caller lacks sufficient
+/// privileges. Callers should handle and log any errors returned by this
+/// function.
+pub fn kill_process_tree(pid: u32) -> std::io::Result<()> {
```
FabianLars, line 149: since this is a generic function and not bound to this app instance imo this should be moved into https://github.com/tauri-apps/plugins-workspace/tree/v2/plugins/process ~~(and in theory that plugin's exit and restart should probably be part of the `app` core plugin instead but that's a discussion for another day)~~.

Legend-Master:
I think we have now gathered a few different problems around this now:

- Only the side car process is killed on app exit, not the entire process tree
  - Just from testing, it seems like some of the processes are killed while some others are not???
- `tauri dev` doesn't kill side cars (it kills the app forcefully by sending `SIGKILL` on mac and linux, `TerminateProcess` on Windows to `cargo`)
  - It seems to me that it does kill the side car but not always the entire process tree???
- NSIS installers don't kill side cars, they kill the app forcefully through `TerminateProcess` (~~the Wix .msi installers do, they send [`WM_QUERYENDSESSION`](https://github.com/wixtoolset/wix/blob/2ce9b1c79a8fb55e123363e49dd1b3c406536278/src/ext/Util/ca/CloseApps.cpp#L368) to the app which is now handled after https://github.com/tauri-apps/tao/pull/1126~~ no, I thought they do but what actually happened is just it killed the tray icon by sending `WM_CLOSE`)
- Updater plugin exits the app through `std::process::exit` which doesn't trigger the `Exit` event so side cars are not killed

My brain is basically fried right now, no more processes 😭

FabianLars:
1) and perhaps 2) should imo be fixed via https://github.com/tauri-apps/plugins-workspace/issues/1332 and not by trying to kill a process tree. prob doesn't help with the rest though.

crates/tauri/src/plugin.rs
sakina1303: Got it, that makes sense! I’ll update the doc comment to clarify that the hook runs inside App::cleanup_before_exit. Thanks for pointing this out!

crates/tauri/src/plugin.rs
sakina1303: for sure! I’ll add this to the TauriPlugin trait as well so it’s accessible via tauri::plugin::Builder.

crates/tauri/PR_DRAFT.md
sakina1303: Sorry, I’ll remove it from the PR, Thanks for spotting it!

crates/tauri/CHANGELOG.md
sakina1303: Oh I see, got it! I’ll drop the CHANGELOG.md edits and use a .changes file

crates/tauri/src/process.rs
sakina1303: Would it be okay if I leave a TODO for now and move it in a follow-up PR?

FabianLars: No, we at least have to decide how to proceed here because once this is added in tauri we cannot remove it without a major release since removing apis is a breaking change.

sakina1303: Should I move kill_process_tree to the process plugin now as you suggested? I’ll update this PR to avoid a future breaking change.

FabianLars: i think that would make sense, yes. we can always move it back if needed (that way around works)

Tunglies:
> I think we have now gathered a few different problems around this now:
> 
> * Only the side car process is killed on app exit, not the entire process tree
>   
>   * Just from testing, it seems like some of the processes are killed while some others are not???
> * `tauri dev` doesn't kill side cars (it kills the app forcefully by sending `SIGKILL` on mac and linux, `TerminateProcess` on Windows to `cargo`)
>   
>   * It seems to me that it does kill the side car but not always the entire process tree???
> * NSIS installers don't kill side cars, they kill the app forcefully through `TerminateProcess` (~the Wix .msi installers do, they send [`WM_QUERYENDSESSION`](https://github.com/wixtoolset/wix/blob/2ce9b1c79a8fb55e123363e49dd1b3c406536278/src/ext/Util/ca/CloseApps.cpp#L368) to the app which is now handled after [fix(windows): emit LoopDestroyed on WM_ENDSESSION tao#1126](https://github.com/tauri-apps/tao/pull/1126)~ no, I thought they do but what actually happened is just it killed the tray icon by sending `WM_CLOSE`)
> * Updater plugin exits the app through `std::process::exit` which doesn't trigger the `Exit` event so side cars are not killed
> 
> My brain is basically fried right now, no more processes 😭

I fixed point 1, 2 in my(our) project. 3  and 4 not test on my machine but has no community report related. With sidecar, I hold a custom strcture to wrap ProcessChild and give impl Drop for it to handle kill when droping. The entire app(which is main process) listen on system signal and excute tauri exiting and custom behavior, it will kill itself and its sidecar. They works well both in tauri dev and release build.

Tunglies:
> > I think we have now gathered a few different problems around this now:
> > 
> > * Only the side car process is killed on app exit, not the entire process tree
> >   
> >   * Just from testing, it seems like some of the processes are killed while some others are not???
> > * `tauri dev` doesn't kill side cars (it kills the app forcefully by sending `SIGKILL` on mac and linux, `TerminateProcess` on Windows to `cargo`)
> >   
> >   * It seems to me that it does kill the side car but not always the entire process tree???
> > * NSIS installers don't kill side cars, they kill the app forcefully through `TerminateProcess` (~the Wix .msi installers do, they send [`WM_QUERYENDSESSION`](https://github.com/wixtoolset/wix/blob/2ce9b1c79a8fb55e123363e49dd1b3c406536278/src/ext/Util/ca/CloseApps.cpp#L368) to the app which is now handled after [fix(windows): emit LoopDestroyed on WM_ENDSESSION tao#1126](https://github.com/tauri-apps/tao/pull/1126)~ no, I thought they do but what actually happened is just it killed the tray icon by sending `WM_CLOSE`)
> > * Updater plugin exits the app through `std::process::exit` which doesn't trigger the `Exit` event so side cars are not killed
> > 
> > My brain is basically fried right now, no more processes 😭
> 
> I fixed point 1, 2 in my(our) project. 3 and 4 not test on my machine but has no community report related. With sidecar, I hold a custom strcture to wrap ProcessChild and give impl Drop for it to handle kill when droping. The entire app(which is main process) listen on system signal and excute Apphandle::Exit, it will kill itself and its sidecar. They works well.

We found it's hard to fully trust tauri exit behavior, and external listen system signal mannuly  in [clash-verge-rev/crates/signal](https://github.com/clash-verge-rev/clash-verge-rev/tree/dev/crates/clash-verge-signal). But behavior of macOS system shutdown signal seems hooked by tauri, can only processing with `tauri::RunEvent::Exit` otherwise will skip signal hanlde, but ubuntu and other linux works fine.

Tunglies:
This PR using system shell. For Windows example in some case, user can not invoke system powershell due to their machine permission setting or even disable powershell usage. We might not want to handle PID with shell diretcly, and there are potential security problems whether if cross-platform.

Legend-Master:
> With sidecar, I hold a custom strcture to wrap ProcessChild and give impl Drop for it to handle kill when droping.

I think this is more or less something we should provide in shell plugin?

> We found it's hard to fully trust tauri exit behavior

Could you explain about this a bit more?

> This PR using system shell.

Not the biggest fan of this either

Tunglies:
> I think this is more or less something we should provide in shell plugin?

Yes, and the auto-cleanup or likely behavior can be toggled via a field setting whether if spawn or runtime. We did not provide this behavior before, might destroy downstream program behavior if enable by default.

> Could you explain about this a bit more?

Months ago, did not remember too much details. 
- The sidecar plugin's process will not be killed when exiting `cli tauri dev` from terminal. 
- We tried handle with `RunEvent::ExitRequested` and `RunEvent::Exit` to reset system network settings when system shudown. None of linux, macOS and windows tasks works as expected. Might just we did not learned the proper way.

More, hanlding with resue `tauri::asyncruntime` to excute exiting or system shutdown operations in Windows will produce more likehood to failed or skipped. With a new tokio spawn would be totally fixed on that and works fine across Windows, macOS and Linux. https://github.com/clash-verge-rev/clash-verge-rev/pull/5533#issuecomment-3562934643 (Chinese)

Legend-Master:
> We tried handle with RunEvent::ExitRequested and RunEvent::Exit to reset system network settings when system shudown. None of linux, macOS and windows tasks works as expected. Might just we did not learned the proper way.

`RunEvent::Exit` event should now fire on system shutdown on Windows (https://github.com/tauri-apps/tao/pull/1126), not on macOS and Linux yet though

> More, hanlding with resue tauri::asyncruntime to excute exiting or system shutdown operations in Windows will produce more likehood to failed or skipped.

Hmm, that sounds weird, the tauri async runtime by default should be very much the same as using a tokio one directly

Tunglies:
> should be very much the same as using a tokio one directly

Our program continuosly handle differents tauri commands, data-related processing, system setting processing. Might be a little more complex than genral program. The async_runtime::spawn `The singleton async runtime used by Tauri and exposed to users.` that we already used for spawn some inner tasks. A new tokio runtime would not be stuggle with tauri. 

```rust
/// async_runtime::spawn
pub fn spawn<F>(task: F) -> JoinHandle<F::Output>
where
  F: Future + Send + 'static,
  F::Output: Send + 'static,
{
  let runtime = RUNTIME.get_or_init(default_runtime);
  runtime.spawn(task)
}
```

Tunglies:
> not on macOS and Linux yet though

Wierd on my test on macOS, signal_hook can not handle shutdown signal, but SIGTERM, SIGTERM interminal or standalone were fine. Macos shutdown handle only works with 

```rust
#[cfg(target_os = "macos")]
        tauri::RunEvent::Exit => async_runtime::block_on(async {
            some_clean_up()
        }),
```

Linux works with competely works with signal_hook. Seemed tauri hooked macOS shutdown signal and use `tauri::RunEvent::Exit` for case.

FabianLars:
> Macos shutdown handle only works with

Not gonna lie, this sounds like something macOS would. Not sending posix signals in favor of the NSApplication delegate bullshit would be such an Apple move 🙃 
I'm somewhat sure it's not us doing that.

Qodo-Free-For-OSS:
Hi, `App::cleanup_before_exit()` unconditionally unwraps the plugins mutex lock; if plugin code panicked earlier and poisoned the mutex, shutdown will panic instead of performing best-effort cleanup.

**How to fix:** Handle poisoned mutex gracefully

> ### Issue description
> `cleanup_before_exit` should be best-effort. Panicking on a poisoned plugins mutex defeats shutdown cleanup.
> 
> ### Issue Context
> This is invoked during application teardown.
> 
> ### Fix Focus Areas
> - crates/tauri/src/app.rs[1042-1048]
>   - Replace `.lock().unwrap()` with logic that handles poisoning, e.g. `match lock()` then `into_inner()` on `PoisonError`, or skip plugin cleanup with a log.
> - crates/tauri/src/plugin.rs[991-998]
>   - (Optional) consider guarding individual plugin hook panics if the runtime’s policy is to continue best-effort cleanup.

---

aka-danielZhang:
Field data + implementation notes from shipping a very similar system, in case it helps this PR move.

**Field data for the `tauri dev` path specifically.** We maintain a Tauri 2 desktop app (dsh-desktop) whose Rust shell spawns a long-lived Node sidecar (a local web server the webview loads). One icon-retune session left **9 orphaned sidecars** — each holding a random loopback port and a shared state directory, with live sessions inside them. Root cause chain, confirmed against tauri-cli 2.11.4 sources:

- `interface/rust.rs` restarts the app on watch with `child.kill()` — SIGKILL on Unix, so no `RunEvent::Exit`, signal handler, `Drop`, or in-app registry can observe it. In-app cleanup is unreachable *by construction* on this path.
- `kill_before_dev_process`'s tree-kill only ever covers the `beforeDevCommand` child, not the app's own descendants.
- The watcher's built-in ignore list (`node_modules/ target/ gen/ Cargo.lock .DS_Store`) doesn't include `icons/`, so even icon-only changes trigger the kill-restart loop with zero Rust changes.

That's why the CLI-side part of this PR (the `desktop.rs` dev-run cleanup) is the piece that actually fixes our orphan class — the runtime registry alone can't cover it. If splitting the PR into "CLI dev-run cleanup" and "runtime registry" helps either half land, we'd strongly encourage it.

**Registry semantics we learned the hard way** (we ship a `~/.dsh-desktop/sidecars.json` with sweep-on-next-boot):

- **pid recycling**: never reap on pid alone. We record `ps lstart` at spawn time and require equality before killing; a recycled pid simply reads as dead. This matters more than you'd expect on long-running desktop sessions.
- **fail open**: a corrupt or missing registry must read as empty, never brick the boot. (Related to @Qodo-Free-For-OSS's note: best-effort cleanup should also survive a poisoned mutex — `lock().unwrap_or_else(|e| e.into_inner())` is the usual pattern for exactly this.)
- **act only on registered pids**: never scan the process table by name or cmdline — a terminal user running the same server binary must be unreachable by mistake.

**On the primitive itself**: we ended up *not* walking the process tree. On Unix, spawning the sidecar into its own process group (`CommandExt::process_group(0)`, std since 1.64) turns `kill(-pgid, SIGTERM)` into an atomic, TOCTOU-free tree signal with zero shell involvement — which also sidesteps the PowerShell-availability and shell-quoting concerns raised above. We follow up with `SIGKILL` on the same group after a grace period so servers get a chance to flush (SIGTERM-first also matters for sidecars that own state, like ours). The same spawn site can adopt `process-wrap` for Windows Job Objects later, which is the direction of plugins-workspace#1332 — so the runtime half may naturally converge there.

Happy to contribute whichever half the team prefers to move first.

lucasfernog (approved):

crates/tauri-cli/src/interface/rust/desktop.rs
```diff
@@ -28,6 +28,42 @@ pub struct DevChild {
...
+  kill -9 "$p" 2>/dev/null
+done
+true
+"#;
+
+      let pid = pid.to_string();
+      let _ = Command::new("sh")
+        .args(["-c", KILL_TREE, "sh", pid.as_str()])
+        .status();
+    }
```
Legend-Master, line 65: @lucasfernog shouldn't this be done after `self.dev_child.kill()?;`?

FabianLars: I'd assume the parent pid may be gone at that point no?

Edit: as in the os process/I'd not the id() function above. So that we'd call task kill on a pid that the os discarded already

Legend-Master: Hmm I see, does `self.dev_child.kill()` kill the process forcefully like `taskkill /F` does?

FabianLars: Yes, it uses https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-terminateprocess under the hood.

Sorry for being slow to respond, bad connection on the train today

Legend-Master: Got it thanks.

Your slow is my fast 😂

justanotheranonymoususer:
This change introduced a deadlock, please take a look, a fix: https://github.com/tauri-apps/tauri/pull/16090

PR #14454, opened 2025-11-12
feat: add liquid glass window effects

FabianLars:
i intentionally only exposed 2 of all the new effects because those are the only stable ones and i'd like to have the other values investigated first - eg got no source for them

also, we really need to fix our tauri-utils code, we can't keep making breaking changes (this one is even worse with the enum)


FabianLars:
grafik

that forced copilot thing is making me go crazy

WSH032:
> also, we really need to fix our tauri-utils code, we can't keep making breaking changes (this one is even worse with the enum)

[effect](https://docs.rs/tauri/latest/tauri/window/enum.Effect.html) has been re-exported as part of the tauri crate's public API, I think it might be a problem?

FabianLars:
yes exactly! everything sucks

WSH032:
If this PR is going to be merged, I would suggest adding `#[non_exhaustive]`. Since compatibility is already broken, might as well go all the way 🤣.
Another option would be to put them behind an `unstable`/`v3` gated feature?

FabianLars:
yeah no i won't let this merged as is. i'll get back to it later. thanks for the ideas!

lucasfernog:
I also noticed some enum variants like RunEvent::Opened that should have the non_exhaustive attribute but don't :( something for v3

FabianLars:
looking through https://github.com/search?q=utils%253A%253AWindowEffect&type=code i cannot see a single case we'd break by adding new enum variants (and i highly doubt anyone `match`es that enum without a `_` case since it's so comically large) so i'm thinking to just bite the bullet here. Probably makes sense to add non_exhaustive as well like WSH032 but i'm not sure if that can increase the attack surface of the breaking change here.

@Legend-Master wdyt? And yes, i will keep pinging you everywhere, better get used to it 😂 

Edit: open for other ideas, just can't think of anything nice

~~Edit2: Probably better to wait with the non_exhaustive for v3, just to be safe. I assume/hope there won't be new variants till then...~~

Edit3 (lol): I forgot that there already were new variants for the initial liquid glass stuff which i just didn't add yet...

Legend-Master:
Sounds safe enough. And yeah, if we're going to break it, break it with a `non_exhaustive`, a full match code breaks on both anyways.

Just want to ask will this be hard to put in a feature? I know it's a bit bad, but I just don't want breakages anymore 😂

And I don't mind the pings, feel free to abuse them 🙃

FabianLars:
Shouldn't be too hard. I think we can mirror macosPrivateApis here - not sure if a separate tauri.conf.json config like that or if we should derive it from windowEffects (plus a cargo feature for those creating windows in js/rust)

FabianLars:
so that's what the macosprivateapi-like approach would look like. not the biggest fan especially since we cannot have an error here similar to https://github.com/tauri-apps/tauri/blob/feat/macos-liquid-glass/crates/tauri-runtime-wry/src/lib.rs#L889 without hijacking the deserializer (and maybe the JsonSchema generator?)

crates/tauri-utils/src/config.rs
```diff
@@ -3098,6 +3110,9 @@ impl AppConfig {
     if self.macos_private_api {
       features.push("macos-private-api");
     }
+    if self.macos_liquid_glass {
```
Legend-Master: Maybe just go through `self.windows` instead of a new config field?

FabianLars: Yes I wanted to do that at first but this would suck for users that create windows in rust or js

Legend-Master: I feel like enabling a new cargo feature isn't harder then changing a new config field is it? 😂

FabianLars: nope but we usually don't do that with tray-icon being the exception for some reason - and to a degree i can get behind that, the amount of times i had to walk people through how to add the devtools flag to the tauri dep is a bit mind boggling

since this is only temporary i wouldn't mind a feature flag only approach though - not sure if i'd even bother about reading self.windows here 🤔

FabianLars: still thinking about just doing the breaking change here - this feels so stupid 😂

Legend-Master: let's do the break, hope we're not the next time (pun intended)

hongguang121:
**Heads-up before merging liquid glass into packaged macOS 27: two AppKit OBJC aborts with WKWebView + `NSGlassEffectView`**

We reproduced this on a Tauri **2.11.2** packaged app (wry 0.55.1, tao 0.35.2) that inserts a public `NSGlassEffectView` the same way this PR / `window-vibrancy` 0.7–0.8 intend to. `tauri dev` is fine; `tauri build` is not.

Window: `transparent` + `titleBarStyle: Overlay` + `decorations` + `resizable`. Glass is a **theme-frame sibling behind** TaoView/WK (insert once; no live `addSubview` / `setFrame` on `Resized`).

### 1. Packaged native resize — `objc_initWeak`

Incident Identifier: `00238567-94AA-41AB-91CA-689FB9EAF214`

DirectTouch (`Gestures.framework`) → `_resizeChangedForDirectTouch:…` → `_oldPlaceWindow:fromServer:YES` → `_layoutViewTree` → glass Swift layout → `objc_initWeak` → `Namespace OBJC` SIGABRT.

In-webview CSS handles that call `setSize` (`fromServer:NO`) do not crash, and they **cannot** stop DirectTouch: the recognizer runs in `-[NSWindow sendEvent:]` before WK.

### 2. Packaged custom-protocol load — `objc_storeWeak`

Incident Identifier: `8F80BDC0-2F0A-4768-A7EC-CBF7E03BC305`

`WebKit::WebURLSchemeHandlerCocoa::platformStartTask` → `_realMakeFirstResponder:` → `_setFirstResponderIvar:` → `objc_storeWeak` → same `Namespace OBJC` abort.

`tauri dev` uses HTTP `devUrl` without the custom-protocol feature; that path never hits `platformStartTask` on first paint.

### What we are not asking Tauri to do

Please do **not** workaround this with `object_setClass` on `NSWindow`, `class_replaceMethod` on Gesture-Fusion / `NSApplication`, a 2-arg DirectTouch stub, or reparenting WK into `NSGlassEffectView.contentView` (that last one is what [window-vibrancy#199](https://github.com/tauri-apps/window-vibrancy/pull/199) currently documents; it also collides with tao 0.35 `TaoView` weak `ns_window`).

We filed these as **AppKit** crashes with Apple Feedback Assistant (FB numbers in a follow-up). Our packaged 2.11.2 apps currently skip native glass and use CSS only.

If 2.12 exposes `windowEffects` liquid glass, it would help to document that **packaged** macOS 27 + transparent overlay + WK is currently an abort, even though **dev** looks correct.

Minimal repro: empty Tauri 2.11.2 project, overlay transparent window, one sibling `NSGlassEffectView`, `npm run tauri:build`, native-edge drag on macOS 27.0 (26A428), Apple Silicon.

lucasfernog:
I went with the non_exhaustive thingy btw

crates/tauri/src/vibrancy/macos.rs
```diff
@@ -77,3 +115,32 @@ pub fn apply_effects(window: impl HasWindowHandle, effects: WindowEffectsConfig)
...
+
+pub fn clear_effects(window: impl HasWindowHandle) {
+  let _ = window_vibrancy::clear_vibrancy(&window);
+  let _ = window_vibrancy::clear_liquid_glass(&window);
+}
+
+/// Sets `NSGlassEffectView.effectIsInteractive` on the glass view window-vibrancy inserted.
+///
+/// The property is macOS 27.0+, so it is a no-op on older systems.
+fn set_glass_interactive(window: &impl HasWindowHandle, interactive: bool) {
```
FabianLars: looks like something that should be in window-vibrancy

PR #14457, opened 2025-11-13
perf(codegen): wrap generated context in a fn

Legend-Master:
Wrap the generated context code in a function to make rust analyzer faster, this is probably a hack but it works 😂

Before:

https://github.com/user-attachments/assets/be46e01e-d307-4243-b1ac-dea2b1d0220c

After:

https://github.com/user-attachments/assets/9c04dd23-f19d-4c1b-9075-84e03793d28b

### DIFF
--- a/.changes/wrap-generate-context.md
+++ b/.changes/wrap-generate-context.md
@@ -1,0 +1,5 @@
+---
+"tauri-codegen": "patch:perf"
+---
+
+Wrap the generated context code in a function to make rust analyzer faster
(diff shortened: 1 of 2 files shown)


FabianLars:
How did you come up with the idea to wrap it in a function?

FabianLars (approved):

Legend-Master:
> How did you come up with the idea to wrap it in a function?

It was quite a journey 😂 So basically I was wondering why the completion's so slow in that function a long time ago, and I found out if I remove this macro, the problem goes away, then I tried to poke around all sorts of different methods but nothing worked so I left it until now

The only things I noticed that worked before were

1. Remove all the capability staffs
2. Move the entire `generate_context` macro to a function (by the way, a closure doesn't work here)

I was trying to put up a documentation on how to improve this by letting rust analyzer ignore this macro with `rust-analyzer.procMacro.ignored` but thought I'd give it a final shot here we are 🙃

PR #14464, opened 2025-11-14
fix(core): properly handle async errors in addPluginListener

funnydino:
The previous implementation used .then() after invoke() without await, which prevented the catch block from handling rejected promises. Now using await to properly catch errors and allow fallback to camelCase registerListener method.

## What kind of change does this PR introduce?
Bug fix

## What is the current behavior?
The `addPluginListener` function doesn't properly catch async errors because `.then()` is called on the Promise without `await`, preventing the `catch` block from handling rejected promises.

## What is the new behavior?
Now using `await` before `invoke()` calls to properly catch async errors and allow fallback to the camelCase `registerListener` method.

## Does this PR introduce a breaking change?
No


Legend-Master (approved):
Thanks! (force pushed for the code sign requirement we had to merge, hope you don't mind)

Issue #14173, opened 2025-09-10
[feat] Support `alias` for command macro


bowenxuuu:
### Describe the problem

Currently, each command function needs to be defined with a unique name, which is extremely difficult in large-scale projects.

### Describe the solution you'd like

It is hoped that alias support can be added to the Command macro, with an example as follows:

``` Rust
mod user{
    #[command(alias = "/user/list")]
    pub async fn list(limit: i64) -> Result<Vec<User>, Error> {
        // todo
    }
}

mod book{
    #[command(alias = "/book/list")]
    pub async fn list(limit: i64) -> Result<Vec<Book>, Error> {
        // todo
    }
}

mod post{
    #[command]
    pub async fn post_list(limit: i64) -> Result<Vec<Post>, Error> {
        // todo
    }
}
```

``` JavaScript
invoke('/user/list', { limit: 10 })

invoke('/book/list', { limit: 10 })

invoke('post_list', { limit: 10 })
```

I hope aliases are optional: if an alias is specified, the command is invoked via the alias; if no alias is set, the command is invoked via the function name.

Tunglies:
I like this ideal :)

PR #14473, opened 2025-11-16
feat(macros): add support for rename command macro in tauri-macros #14173

Tunglies:
Related issues: https://github.com/tauri-apps/tauri/issues/14173

---

Preview

image

---

IDE & Rust Analyzer

image

image

image

---

Test code:

```rust
mod me {
    #[tauri::command(alias = "greet_from_me")]
    pub fn hello(name: &str) -> String {
        format!("Hello, {}! Me hello from Rust!", name)
    }
}

mod you {
    #[tauri::command(alias = "greet_from_you")]
    pub fn hello(name: &str) -> String {
        format!("Hi, {}! You hello from Rust!", name)
    }
}

#[tauri::command(alias = "salute")]
fn hello(name: &str) -> String {
    format!("Hello, {}! You've been greeted from Rust!", name)
}

pub fn run() {
    tauri::Builder::default()
        .invoke_handler(tauri::generate_handler![hello, me::hello, you::hello])
}
```

```typescript
import { invoke } from "@tauri-apps/api/core";
import { useState } from "react";
import "./App.css";
import reactLogo from "./assets/react.svg";

function App() {
  const [greetMsg, setGreetMsg] = useState("");
  const [name, setName] = useState("");

  async function handleGreetAll() {
    const [salute, fromMe, fromYou] = await Promise.all([
      invoke<string>("salute", { name }),
      invoke<string>("greet_from_me", { name }),
      invoke<string>("greet_from_you", { name }),
    ]);
    setGreetMsg([salute, fromMe, fromYou].join("\n"));
  }

  return (
    <main className="container">
      <h1>Welcome to Tauri + React</h1>

      <div className="row">
        <a href="https://vite.dev" target="_blank">
          <img src="/vite.svg" className="logo vite" alt="Vite logo" />
        </a>
        <a href="https://tauri.app" target="_blank">
          <img src="/tauri.svg" className="logo tauri" alt="Tauri logo" />
        </a>
        <a href="https://react.dev" target="_blank">
          <img src={reactLogo} className="logo react" alt="React logo" />
        </a>
      </div>
      <p>Click on the Tauri, Vite, and React logos to learn more.</p>

      <form
        className="row"
        onSubmit={(e) => {
          e.preventDefault();
          handleGreetAll();
        }}
      >
        <input
          id="greet-input"
          onChange={(e) => setName(e.currentTarget.value)}
          placeholder="Enter a name..."
        />
        <button type="submit">Greet</button>
      </form>
      <pre>{greetMsg}</pre>
    </main>
  );
}

export default App;
```


Legend-Master:
Didn't look much into this yet, but I think if the idea is to rename the command, not having both command names valid, we should call it `rename` similar to `#[serde(rename = "name")]`

Tunglies:
> Didn't look much into this yet, but I think if the idea is to rename the command, not having both command names valid, we should call it `rename` similar to `#[serde(rename = "name")]`

I initially thought of using `rename` with `#[serde(rename = "name")]`, but ended up keeping `alias` for functionality first and reference the issue. See if any review ideals I can go further.

Tunglies:
> Didn't look much into this yet, but I think if the idea is to rename the command, not having both command names valid, we should call it `rename` similar to `#[serde(rename = "name")]`

It should be clearer now, changed to `#[tauri::command(rename = "name")]` . Since the new serde method exposed to user, might conflect with `#serde(rename_all = "snake_case")`, but the priority of `rename_all` is higher. We might need document on this behavior.

lucasfernog:
pushed some fixes for commands defined in a separate module (the commands example is a good way to test this). Unfortunately we need to always export the macro, and preserve the condition of unique command function names :/

examples/commands/main.rs
```diff
@@ -252,6 +252,7 @@ fn main() {
       async_stateful_command,
       command_arguments_wild,
       renamed_command,
+      renamed_command_in_mod,
```
Tunglies, line 255: How abort export same function name from different module?
```rust
A::renamed_command,
B::renamed_command,
```

lucasfernog: see https://github.com/tauri-apps/tauri/pull/14473#issuecomment-3697119435

lucasfernog: there's no way to support all use cases :/ not exporting macros breaks some other scenarios (like importing the function and then using it directly IIRC)

lucasfernog (approved):

PR #14475, opened 2025-11-16
perf: remove needless clones in various files for improved performance

Tunglies:


Tunglies:
Perhaps we could consider adding Clippy rules to the repo?

FabianLars:
> Perhaps we could consider adding Clippy rules to the repo?

What do you have in mind? `pedantic` used to be too much

Tunglies:
> > Perhaps we could consider adding Clippy rules to the repo?
> 
> What do you have in mind? `pedantic` used to be too much

Maybe we can try `nursery` rules first, e.g. `clippy::significant_drop_tightening` and  `clippy::or_fun_call`. `pedantic ` rules producing some noise, but helps in over all so we can try avoid expensive performance operations.

Tunglies:
The issue in https://github.com/tauri-apps/tauri/pull/14476, where an unwrap causes a panic, could be avoided by enabling the Clippy rule clippy::unwrap_used with warn or deny.

FabianLars (approved):

Tunglies:
> > Perhaps we could consider adding Clippy rules to the repo?
> 
> What do you have in mind? `pedantic` used to be too much

I lost FabianLars, too much .expect and .unwarp uses XD

FabianLars:
yeah, the cli (and its subcrates) is quite heavy on that but since it's a cli it's more or less alright. the tauri runtime crates would be more important but even that has many unwraps :/

Tunglies:
> yeah, the cli (and its subcrates) is quite heavy on that but since it's a cli it's more or less alright. the tauri runtime crates would be more important but even that has many unwraps :/

I heared that recently Cloudflare gets 500 Internal server error due to ONE line rust  unwarp operation.  LOL

FabianLars:
luckily we're not cloudflare right? 😂 Though it sounds like that unwrap wasn't really the cause but just the end of a chain of errors though using `expect` should have been the bare minimum if you can't or don't want to propagate the error up

FabianLars:
same for us of course, most unwraps should be replaced with expect even if that results in larger file size

PR #14479, opened 2025-11-16
enhance(nsis): use restart manager to close app

### LINKED ISSUE #39: When using KillProcess or KillProcessCurrentUser to close the tauri application during uninstallation, the tauri application cannot perform pre exit processing

### LINKED ISSUE #9950: [bug] nsis installer won't stop spawned sidecar when upgrading
### Describe the bug

If the tauri app is runing with a spawned background sidecar process, when clicking the nsis bundle to upgrade the tauri app it would pop up a window asking to close the tauri app("xxx is runing. Click OK to close it"). But the spawned sidecar process won't be terminated after the OK button is clicked.

Shall we call tauri cleanup so that "RunEvent::Exit" or "RunEvent::ExitRequest" event handler could be called, or call "taskkill /f /t " to terminate the tauri app process tree so that all the child process would be killed together like described in https://github.com/electron-userland/electron-builder/issues/2516.

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
$ tauri info

[✔] Environment
    - OS: Windows 10.0.22631 X64
    ✔ WebView2: 125.0.2535.79
    - node: 18.18.2
    - yarn: 1.22.21
    - npm: 9.8.1

[-] Packages
    - tauri [RUST]: 2.0.0-beta.22
    - tauri-build [RUST]: 2.0.0-beta.17
    - wry [RUST]: 0.40.0
    - tao [RUST]: 0.28.0
    - @tauri-apps/api [NPM]: 2.0.0-beta.13
    - @tauri-apps/cli [NPM]: 2.0.0-beta.17

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - framework: Vue.js
    - bundler: Vite
Done in 83.82s.
```

### Stack trace

_No response_

### Additional context

_No response_

Legend-Master:
Closes #9950
Closes https://github.com/tauri-apps/nsis-tauri-utils/issues/39
Reference https://github.com/tauri-apps/tao/issues/1115

Use [Restart Manager](https://learn.microsoft.com/en-us/windows/win32/rstmgr/about-restart-manager) to close running app instead of our own `nsis_tauri_utils::KillProcess`, this makes it so that we send a `WM_ENDSESSION` signal to the app for it to gracefully shutdown (paired with https://github.com/tauri-apps/tao/pull/1126)

This also fixed a problem where we would wrongly detect other programs that have the same name as us (e.g. both named `api.exe`) and shut them down

~~One more thing to go though, we should probably also switch `nsis_tauri_utils::FindProcess` to `RmGetList` and use a fully qualified path here, so we fully fix this problem~~ done in 68b9d41


Legend-Master:
I know the code is a bit wacky, but honestly I can't quite think of anything better here

Since there's a cost with `RmRegisterResources`, I'm not sure if it makes sense to just replace the implementation in `FindProcess` and `KillProcess`

> https://learn.microsoft.com/en-us/windows/win32/api/restartmanager/nf-restartmanager-rmregisterresources#:~:text=Each%20call%20to%20the%20RmRegisterResources%20function%20performs%20relatively%20expensive%20write%20operations.%20Do%20not%20call%20this%20function%20once%20per%20file%2C%20instead%20group%20related%20files%20together%20into%20components%20and%20register%20these%20together.

Legend-Master:
Actually, this will cause the app to crash on https://github.com/tauri-apps/tao/blob/3c2b4447aa53151ae96d30a60928d1d71e9bb5fc/src/platform_impl/windows/event_loop/runner.rs#L368 because Restart Manager works a bit differently from system shutdown, it doesn't terminate us right away after returning 0 from `WM_ENDSESSION`

Not sure if we should just call `std::process::exit` when we get `WM_ENDSESSION` to keep the behavior consistent though

amrbashir (approved):

Issue #1633, opened 2021-04-27
AppImage icon not working


lucasfernog:
Some users reported an issue on AppImage icon. We should investigate that.

lucasfernog:
This was a desktop environment (or configuration) issue and not a Tauri issue, as we've seen it working on two different OS, and the issue was only reported on Manjaro with KDE.

PR #14484, opened 2025-11-17
feat(mobile): multi-window support

lucasfernog:
Leverages [scenes](https://developer.apple.com/documentation/uikit/scenes) on iOS and [Activity embedding](https://developer.android.com/develop/ui/views/layout/activity-embedding) on Android.

All window APIs are now exposed on mobile too. Some of them are still ignored though (like setting a title).

# iOS
- Added RunEvent::SceneRequested (on iPad the user can request a new window to be open - e.g. by long pressing the app icon and selecting "New window")

# Android
- Added builder methods to determine the activity to be created. System determines what to do with the activity (new stack, next to another one.. based on the embedding rules)

needs https://github.com/tauri-apps/tao/pull/1154 and https://github.com/tauri-apps/wry/pull/1633


lucasfernog:
here's an example: https://github.com/lucasfernog/tauri-mobile-multi-window-example

crates/tauri/src/webview/plugin.rs
```diff
@@ -163,6 +128,42 @@ mod desktop_commands {
...
 
+  #[cfg(not(feature = "unstable"))]
+  #[command(root = "crate")]
+  pub async fn create_webview() -> crate::Result<()> {
+    Err(crate::Error::UnstableFeatureNotSupported)
+  }
+
+  #[cfg(feature = "unstable")]
+  #[command(root = "crate")]
+  pub async fn create_webview<R: Runtime>(
```
velocitysystems, line 139: While we may not wish to enforce this (since it is device-specific) there is a _theoretical limit_ to the number of web-views which can be created based on the available memory.

If the app begins to use too much memory the OS may mark it for termination particularly if it doesn't respond to memory warnings and attempt to free memory.

velocitysystems, line 139: WebViews are full browser engines that continue consuming significant CPU, GPU, and battery resources (15-25% per hour) even when the app is backgrounded, unless _explicitly paused_ through lifecycle event handlers (`onPause`/`onResume` on Android or scene lifecycle hooks on iOS).

This can reduce background drain to 1-3% per hour—a critical requirement for app store approval and acceptable user experience.

crates/tauri/src/webview/plugin.rs
```diff
@@ -5,86 +5,51 @@
...
+async fn create_webview_window<R: Runtime>(
+  app: AppHandle<R>,
+  options: WindowConfig,
+) -> crate::Result<()> {
+  WebviewWindowBuilder::from_config(&app, &options)?.build()?;
+  Ok(())
+}
 
+#[cfg(desktop)]
+mod desktop_commands {
```
velocitysystems, line 44: Apps can't currently detect if multi-window is available. Should we add an API like this? e.g.

Typescript
```ts
if (await app.supportsMultiWindow()) {
    // Create new window
} else {
    // Ignore
}
```

Rust
```rust
pub fn supports_multi_window() -> bool {
    #[cfg(target_os = "iOS")]
    {
        ios_version() >= 13.0
    }
    #[cfg(target_os = "android")]
    {
        android_api_level() >= 32 // Android 12L
    }
    #[cfg(desktop)]
    {
        true
    }
}
```

velocitysystems (commented):
We're very excited about this feature @lucasfernog. I've left comments on [#1154](https://github.com/tauri-apps/tao/pull/1154) and [#1633](https://github.com/tauri-apps/wry/pull/1633). Here are some additional general questions:

**Platforms & Compatibility**
- How will this affect Tauri's minimum supported versions?(i.e. Requires iOS 13+ and Android 12L)
- What is the fallback behavior for unsupported platforms?
- Which features of the Window API are available vs not available on mobile?

**Performance & Resource Management**
- How do we intend to handle app lifecycle events? e.g. sleep/resume
- How can we optimize multi-window support so as not to affect battery life?

crates/tauri/src/webview/plugin.rs
lucasfernog: right, that's not something we're worried - otherwise we would need to implement such check for all platforms

lucasfernog: for Android the system pauses the activity automatically when it moves to the background.. I would expect iOS to do the same

crates/tauri/src/webview/plugin.rs
lucasfernog: good idea [`7f02cf5`](https://github.com/tauri-apps/tauri/pull/14484/commits/7f02cf5863143e81bb4fec887c78f5cb5030ebed)

lucasfernog:
> We're very excited about this feature @lucasfernog. I've left comments on [#1154](https://github.com/tauri-apps/tao/pull/1154) and [#1633](https://github.com/tauri-apps/wry/pull/1633). Here are some additional general questions:
> 
> **Platforms & Compatibility**
> 
> * How will this affect Tauri's minimum supported versions?(i.e. Requires iOS 13+ and Android 12L)

The default minimum version for iOS is already 14+ but min Android is 7, but both are configurable. The `supports_multiple_windows` will be helpful for this.

> * What is the fallback behavior for unsupported platforms?

The activity gets replaced on Android (new activity in the stack).
The webview gets replaced on iOS.

> * Which features of the Window API are available vs not available on mobile?

That's something I wanted to do a second pass in the future, most of the APIs can be supported.

> 
> **Performance & Resource Management**
> 
> * How do we intend to handle app lifecycle events? e.g. sleep/resume

That's something the system handles. We can add sleep/resume events per window in the future.

> * How can we optimize multi-window support so as not to affect battery life?

As far as I know there's nothing special we can/need to do.

crates/tauri/src/webview/plugin.rs
velocitysystems: @lucasfernog Unlike iOS I don't believe Android _automatically_ pauses the WebView's internal threads or JS timers when the Activity's `onPause(..)` is called. The Android [WebView](https://developer.android.com/reference/android/webkit/WebView.html) has explicit methods `onPause(..)` and `pauseTimer(..)`.

See this SO [discussion](https://stackoverflow.com/questions/19252948/android-webview-not-stopping-after-user-presses-back) for further discussion.

lucasfernog: I've implemented that but it didn't help much [tauri-apps/wry@`0904dc9` (#1633)](https://github.com/tauri-apps/wry/pull/1633/commits/0904dc98d630157d344178dfb7349e8bb34741ab)
i think pauseTimer is something that each user should decide if they want to call or not - they can do so in their TauriActivity subclass.

nicoburns:
Would the approach used in this series of PRs also be suitable for having (potentially mutliple) Tauri view's embedded within a single native view hierarchy? e.g. Using things like iOS/Android native bottom tabs, bottom sheets, or stack navigation views and having Tauri/Wry views within those.

velocitysystems:
> Would the approach used in this series of PRs also be suitable for having (potentially mutliple) Tauri view's embedded within a single native view hierarchy? e.g. Using things like iOS/Android native bottom tabs, bottom sheets, or stack navigation views and having Tauri/Wry views within those.

The current implementation doesn't support embedding Tauri/Wry views within native view hierarchies like tabs, bottom sheets, or navigation stacks. That would require a different architectural approach at the TAO and WRY layers. This PR enables multiple independent full-screen windows, each with their own webview.                                                                                           
                                                                                                         
It's worth noting that Tauri's mobile architecture is _webview-centric_ more broadly — there's currently no supported way to embed native UI components alongside the webview. Plugins can access native platform APIs (Swift/Kotlin), but not modify the view hierarchy. There's some discussion around this in #10993.

FabianLars (approved):

crates/tauri/mobile/android/src/main/java/app/tauri/plugin/PluginManager.kt
```diff
@@ -26,7 +25,7 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.databind.module.SimpleModule
 import java.lang.reflect.InvocationTargetException
 
-class PluginManager(val activity: AppCompatActivity) {
+object PluginManager {
```
Legend-Master, line 28: @lucasfernog this seems to cause #15506, do we have a reason for the change from class to object here?

Just a note that since we have `android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale|smallestScreenSize|screenLayout|uiMode"` in the template manifest, the activity should not normally get recreated, but this is still a problem

Legend-Master: I see the problem here... With multiple activities, `PluginManager` can no longer be bind to one activity.

PR #14486, opened 2025-11-17
feat(mobile): add file association support

lucasfernog:
needs https://github.com/tauri-apps/tao/pull/1155
needs https://github.com/tauri-apps/wry/pull/1634

needs #14484 to be merged first


crates/tauri-build/src/mobile.rs
```diff
@@ -2,11 +2,130 @@
...
+      }
+
+      filters.push_str("</intent-filter>\n");
+    }
+  }
+
+  filters
+}
+
+fn extension_to_mime_type(ext: &str) -> Option<String> {
```
velocitysystems, line 115: While this covers ~25 file types it is important to include support for _custom file extensions/MIME types_.

We need to be able to define these in **tauri.conf.json** for desktop/mobile so they are available to the builder in order to generate correct platform-specific metadata e.g. info.plist, AndroidManifest.xml.

crates/tauri-build/src/mobile.rs
```diff
@@ -2,11 +2,130 @@
...
+        ));
+
+        // Add file scheme and path patterns for extensions
+        if !association.ext.is_empty() {
+          // Create path patterns for each extension
+          // Android's pathPattern needs \\. (double backslash-dot) in XML to match a literal dot
+          let path_patterns: Vec<String> = association
+            .ext
+            .iter()
+            .map(|ext| format!(".*\\\\.{}", ext.0))
```
velocitysystems, line 80: Android's pathPattern has known limitations e.g. Doesn't match paths with multiple dots (e.g., file.backup.png).
- https://stackoverflow.com/questions/3400072
- https://issuetracker.google.com/issues/36983133

Workarounds:
1. Use multiple patterns
```xml
<data android:pathPattern=".*\\.png" />
<data android:pathPattern=".*\\..*\\.png" />
<data android:pathPattern=".*\\..*\\..*\\.png" />
```

2. Use MIME type only
```xml
<intent-filter>
    <action android:name="android.intent.action.SEND" />
    <data android:mimeType="image/png" />
    <!-- No pathPattern - rely on MIME type -->
</intent-filter>
```

3. Scheme + MIME Type
```xml
<intent-filter>
    <action android:name="android.intent.action.VIEW" />
    <data android:scheme="file" />
    <data android:scheme="content" />
    <data android:mimeType="image/png" />
</intent-filter
```

If filters are not generated properly, the only way forward is then to use a **wildcard (*) filter** which associates the app with all file/MIME types - this is not an ideal UX experience.

crates/tauri-build/src/mobile.rs
```diff
@@ -2,11 +2,130 @@
...
+    }
+
+    // If we have mime types, create intent filters
+    if !mime_types.is_empty() {
+      for mime_type in &mime_types {
+        filters.push_str("<intent-filter>\n");
+        filters.push_str("    <action android:name=\"android.intent.action.SEND\" />\n");
+        filters.push_str("    <action android:name=\"android.intent.action.SEND_MULTIPLE\" />\n");
+        filters.push_str("    <category android:name=\"android.intent.category.DEFAULT\" />\n");
+        filters.push_str("    <category android:name=\"android.intent.category.BROWSABLE\" />\n");
```
velocitysystems, line 67: We are missing `ACTION_VIEW` e.g. user taps in  file manager. Ideally, if would be good if we could define which actions we which to listen to.

velocitysystems, line 67: From [Android Docs](https://developer.android.com/guide/components/intents-filters):

> BROWSABLE - The target activity allows itself to be started by a web browser to display data referenced by a link, such as an image or an e-mail message.

This is related to deep-linking not file association. Does this need to be handled differently? How will this interact with the [deep-link](https://v2.tauri.app/plugin/deep-linking/) plugin?

crates/tauri-build/src/mobile.rs
```diff
@@ -2,11 +2,130 @@
...
+        if let Some(mime) = extension_to_mime_type(&ext.0) {
+          mime_types.insert(mime);
+        }
+      }
+    }
+
+    // If we have mime types, create intent filters
+    if !mime_types.is_empty() {
+      for mime_type in &mime_types {
+        filters.push_str("<intent-filter>\n");
```
velocitysystems, line 48: Tauri config has **rank** but this is not being used on Android.
```json
{
  "ext": ["png"],
  "mimeType": "image/png",
  "rank": "Owner"  // ← This exists in config
}
```

Android can specify a _priority_ on intent filters which is a signed integer between -1000 to 1000.

```xml
<intent-filter android:priority="1000">
    <action android:name="android.intent.action.VIEW" />
    <data android:mimeType="image/png" />
</intent-filter>
```

crates/tauri-utils/src/config.rs
```diff
@@ -1201,6 +1201,221 @@ pub struct ExportedFileAssociation {
...
+  }
+  if !document_types.is_empty() {
+    plist.insert("CFBundleDocumentTypes".into(), Value::Array(document_types));
+  }
+
+  Some(Value::Dictionary(plist))
+}
+
+/// Maps file extensions to their standard UTIs for macOS/iOS share sheet support
+fn extension_to_uti(ext: &str) -> Option<&'static str> {
```
velocitysystems, line 1408: Same content RE: file extensions and standard UTIs. While this covers ~30 file types it is important to include support for custom file extensions/MIME types in tauri.conf.json.

crates/tauri-build/src/mobile.rs
lucasfernog: these are just some default values because a mime type is required for the iOS file association to work; you can still define those manually in tauri.conf.json to support other types

crates/tauri-build/src/mobile.rs
lucasfernog: this can be accomplished by just setting a mimeType in the file association config

crates/tauri-build/src/mobile.rs
lucasfernog: [`9d20d69`](https://github.com/tauri-apps/tauri/pull/14486/commits/9d20d6920fe9e6144cafd4fdfb2ddd49ed35d84f)

lucasfernog: the deep-link plugin will need a small change to actually detect that the intent contains a URL that was registered as a deep link - currently it always emits the event

crates/tauri-build/src/mobile.rs
lucasfernog: the priority is only related to other intent filters in the same app
rank is a macOS config that describes whether the app is the owner of a file type or just a secondary viewer.

crates/tauri-utils/src/config.rs
lucasfernog: content types can be defined in tauri.conf.json, this is just a fallback

crates/tauri-build/src/mobile.rs
velocitysystems: Thanks @lucasfernog. Is this task being tracked somewhere? Ensuring that both features continue to work together is an important requirement.

FabianLars (approved):

PR #14493, opened 2025-11-18
Feat/cleanup autogenerated permissions

### LINKED ISSUE #14320: [feat] Remove outdated autogenerated permission files
### Describe the problem

The `tauri_plugin` build script seems to only generate new permission files according to the provided list of commands but never attempts to remove any outdated ones. Not sure if the case is the same for in-app commands and inlined plugins with `tauri_build`.

However, IIUC it seems that the capabilities schema is generated by globbing the permissions directory. This will result in no build/compilation error for the outdated permissions and bugs will be discovered only when actually running the app and seeing the console error. This might be less of a problem for public plugins, but for non-public plugins (especially in my case where the autogenerated directory is gitignored), I think this is worth fixing.

Automatically remove outdated autogenerated permission files.

Potential implementation: Traverse the autogenerated permissions directory and compare against the list of commands passed into the builder. Remove those not in the list. The rest of the logic remains the same.

### Alternatives considered

Alternative implementation: Remove the autogenerated permissions directory and its contents every time (maybe provide an option to turn this behavior on/off). Drawback is the `write_if_changed` benefits are gone.

### Additional context

_No response_

Mishra-coder:
### Description

This PR adds automatic cleanup of outdated autogenerated permission files.  
Previously, old `.toml` files stayed in the `autogenerated` folder even after commands were removed, causing stale permissions to load at runtime.

### What’s changed

- Compare existing autogenerated files with the current command list
- Remove files that no longer match any command

### Why

Prevents stale permission files, reduces confusion, and keeps generated output accurate.

### Related Issue
Fixes  https://github.com/tauri-apps/tauri/issues/14320


Mishra-coder:
@FabianLars,@Legend-Master  PR is ready for mearge please review and suggest changes if required..

crates/tauri-utils/src/acl/build.rs
```diff
@@ -316,6 +316,25 @@ commands.deny = ["{command}"]
       .push(format!("deny-{slugified_command}"));
   }
 
+  use std::collections::HashSet;
+
+  let expected_files: HashSet<String> =
```
Legend-Master: We can push a list of generated files from this time near the `let out_path = path.join(format!("{command}.toml"));` above

crates/tauri-utils/src/acl/build.rs
```diff
@@ -316,6 +316,25 @@ commands.deny = ["{command}"]
...
+    commands.iter().map(|cmd| format!("{}.toml", cmd)).collect();
+
+  if let Ok(entries) = fs::read_dir(path) {
+    for entry in entries.flatten() {
+      if let Ok(file_type) = entry.file_type() {
+        if file_type.is_file() {
+          let file_name = entry.file_name().to_string_lossy().to_string();
+          if !expected_files.contains(&file_name) {
```
Legend-Master: I think we probably should be doing path check here

Legend-Master (commented):
Thanks! Also do you mind adding a change file?

Mishra-coder:
@Legend-Master Please review and let me know if anything else is needed.

package.json
```diff
@@ -25,6 +25,7 @@
     "example:api:dev": "pnpm run --filter \"api\" tauri dev"
   },
   "devDependencies": {
+    "@changesets/cli": "^2.29.7",
```
Legend-Master: We have this setup in the CI (also we use convector not changesets), please remove this

crates/tauri-utils/src/acl/build.rs
```diff
@@ -286,12 +286,15 @@ pub fn autogenerate_command_permissions(
...
 
+  let mut generated_files = Vec::new();
+
   for command in commands {
     let slugified_command = command.replace('_', "-");
 
     let toml = format!(
       r###"{license_header}# Automatically generated - DO NOT EDIT!
 {schema_entry}
+
```
Legend-Master: Please revert this

.changes/cleanup-autogenerated-permissions.md
```diff
@@ -0,0 +1,6 @@
+---
+'tauri-utils': 'patch'
+'tauri-cli': 'patch'
```
Legend-Master:
```suggestion
'tauri-utils': 'patch:enhance'
'tauri': 'patch:enhance'
'tauri-build': 'patch:enhance'
'tauri-plugin': 'patch:enhance'
```

crates/tauri-utils/src/acl/build.rs
```diff
@@ -316,6 +321,24 @@ commands.deny = ["{command}"]
       .push(format!("deny-{slugified_command}"));
   }
 
+  use std::collections::HashSet;
+
+  let expected_files: HashSet<PathBuf> = generated_files.into_iter().collect();
```
Legend-Master: Let's just use the set above, replacing `let mut generated_files = Vec::new();`

Legend-Master (commented):
Please revert unrelated changes and address the review comments above

crates/tauri-utils/src/acl/build.rs
```diff
@@ -316,6 +322,20 @@ commands.deny = ["{command}"]
       .push(format!("deny-{slugified_command}"));
   }
 
+  if let Ok(entries) = fs::read_dir(path) {
```
Legend-Master, line 323:
```suggestion
  // Remove outdated permission files
  if let Ok(entries) = fs::read_dir(path) {
```

Legend-Master (dismissed):
Thanks!

And to merge this, we'll need all your commits to be signed

crates/tauri-utils/src/acl/build.rs
```diff
@@ -281,11 +281,15 @@ pub fn autogenerate_command_permissions(
     "".to_string()
   };
 
+  use std::collections::HashSet;
```
Legend-Master: Let's import this in the top of this file

Legend-Master:
You'll also need to sign all your past commits

> https://stackoverflow.com/questions/41882919/is-there-a-way-to-gpg-sign-all-previous-commits

FabianLars:
Seems like they are signed but you didn't upload the key into github yet https://docs.github.com/en/authentication/managing-commit-signature-verification/adding-a-gpg-key-to-your-github-account

idk what's with the first commit as that one is associated with a different account (but i assume that's still you judging by the similarity to your first name). if you can't fix that i can bypass the merge requirements as an exception if Tony signs off on the changes.

PR #14503, opened 2025-11-19
feat(window): add drag-stopped and resize-stopped events using move/r…

### LINKED ISSUE #14489: [feat] Drag-stopped and resize-stopped event are needed
### Describe the problem

ref: https://github.com/tauri-apps/tauri/discussions/14446

Sometimes we can't use `mouseup` event to detect the end of a drag (or resize) operation.

So, we may need drag-stopped and resize-stopped event.

### Alternatives considered

_No response_

### Additional context

_No response_

Khyati-Kapil:
feat(window-events): add drag-stopped and resize-stopped events

- Added listener for `tauri://move` to detect when window dragging stops
  using a 150ms idle timeout.
- Added listener for `tauri://resize` to detect resize-stop events.
- Ensures consistent behavior across macOS, Windows, and Linux.
This closes #14489


crates/tauri/src/window/scripts/drag.js
```diff
@@ -72,4 +97,11 @@
       }
     })
   }
+
+  window.addEventListener('tauri://resize', () => {
```
Legend-Master, line 101: This is not reliable to be honest

crates/tauri/src/window/scripts/drag.js
```diff
@@ -43,9 +63,14 @@
       // https://github.com/tauri-apps/tauri/issues/2549#issuecomment-1250036908
       e.stopImmediatePropagation()
 
-      // start dragging if the element has a `tauri-drag-region` data attribute and maximize on double-clicking it
-      const cmd = e.detail === 2 ? 'internal_toggle_maximize' : 'start_dragging'
-      window.__TAURI_INTERNALS__.invoke('plugin:window|' + cmd)
+      // maximize on double-clicking it, or start dragging on single click
+      if (e.detail === 2) {
```
Legend-Master, line 67: So I believe this only works for the dragging from our custom draggable region (`data-tauri-drag-region`)?

Legend-Master (commented):
Seeing the limitation on this implementation, I think this should be done on user land or we'll need a more complete solution

crates/tauri/src/window/scripts/drag.js
Khyati-Kapil: Thanks for the feedback, I will try to provide the complete solution using another approach.

Khyati-Kapil:
@Legend-Master  As suggested, I didn’t implement drag-stopped / resize-stopped inside the
runtime because the JS-side detection isn’t reliable enough for all
platforms.

Instead, I added a userland example to the Tauri documentation that shows how
developers can detect when window dragging or resizing has stopped by
listening to `tauri://move` and `tauri://resize` events and debouncing them.

Link to PR in tauri-docs:
https://github.com/tauri-apps/tauri-docs/pull/3631

This keeps the runtime unchanged while still giving developers a clear and
recommended pattern to use until native backend support is added.
Happy to adjust or expand the docs if needed.

Legend-Master:
I'll close this for now, feel free to re-open if you came up with something more complete in the future, thanks anyways!

PR #14521, opened 2025-11-22
fix: issues when patching binary files 

### LINKED ISSUE #14186: [bug] Bundler fails to patch __TAURI_BUNDLE_TYPE when `strip` enabled in Cargo.toml

kandrelczyk:
This fixes https://github.com/tauri-apps/tauri/issues/14186

Alternative patch method based on string value change. Since this is now common for all platforms I removed the no longer necessary code. Also clean up error that is no longer needed. Tested this on Linux with stripped binary and using updater plugin unit tests.


crates/tauri-bundler/src/bundle.rs
```diff
@@ -15,29 +16,33 @@ mod windows;
...
-        binary,
-        package_type.short_name()
-      );
-      windows::patch_binary(binary, package_type)?;
-    }
-    _ => (),
-  }
+  let mut file_data = std::fs::read(binary).expect("Could not read binary file.");
+
+  if let Some(bundle_var_index) = kmp::index_of(BUNDLE_VAR_TOKEN, &file_data) {
```
Legend-Master, line 27: Didn't look at it closely yet, but I think this would fail when patching the binary for the second bundle type right? (e.g `.msi` first and `__TAURI_BUNDLE_TYPE` becomes `__TAURI_BUNDLE_TYPE_VAR_MSI` already that when we do it for the NSIS bundle, we can no longer find `__TAURI_BUNDLE_TYPE_VAR_UNK`)

kandrelczyk: There was an issue with double signing the binaries or patching already signed binary (I don't remember exactly) and the fix was to copy the original binary before each bundle so this should be fine. I tested on linux with standard deb,rpm and appImage and didn't get any errors.

FabianLars: do you remember if after the bundling is all done the binary in target/release/ will be a fresh one or the one from the last patch?

kandrelczyk: I've double checked and something is not right. I will fix it so that we always patch and bundle a copy of the binary and at the end we're left with the original, unpatched one.

Legend-Master: I am still a bit concerned about this approach, like if we stopped at one of these steps (user pressed ctrl-c or we had an error somewhere), there's no way to recover/bundle again other than rebuilding the app's binary through cargo, and since we're matching any string in the binary, although unlikely, we risk replacing the wrong things

In my opinion, this method should not be used for the binary patching/main detection method

FabianLars: > I am still a bit concerned about this approach, like if we stopped at one of these steps (user pressed ctrl-c or we had an error somewhere), there's no way to recover/bundle again other than rebuilding the app's binary through cargo

if we make sure to never touch the binary in target/release/ and always work on the copy somewhere else then you can just re-run tauri bundle but not sure if that's what you mean

> and since we're matching any string in the binary, although unlikely, we risk replacing the wrong things

True but this is so unlikely i wouldn't even count this as an argument

> this method should not be used for the binary patching

well, any ideas left for the patching idea?

> main detection method

we can still offer the registry / file approach as an alternative as well though tbh i kinda see the appeal of an "invisible" approach

Legend-Master: > if we make sure to never touch the binary in target/release/ and always work on the copy somewhere else then you can just re-run tauri bundle but not sure if that's what you mean

Ah, didn't think of that, that should probably fix this problem, it might leave files in the temp folder but I guess if that's somewhere inside `target`, the next `cargo clean` removes it anyways

> True but this is so unlikely i wouldn't even count this as an argument

Fair, I just don't really like that chance to be honest 😂

> well, any ideas left for the patching idea?

I messed up the words when moving them, I was thinking it shouldn't be the 'main binary patching/detection' method (e.g. use the current one and adding this for a backup)

FabianLars: > I messed up the words when moving them, I was thinking it shouldn't be the 'main binary patching/detection' method (e.g. use the current one and adding this for a backup)

ah my bad. I mean honestly fair enough! we can keep both 🤷

kandrelczyk:
```
  // When packaging multiple binary types, we make a copy of the unsigned main_binary so that we can
  // restore it after each package_type step. This avoids two issues:
  //  - modifying a signed binary without updating its PE checksum can break signature verification
  //    - codesigning tools should handle calculating+updating this, we just need to ensure
  //      (re)signing is performed after every `patch_binary()` operation
  //  - signing an already-signed binary can result in multiple signatures, causing verification errors
  let main_binary_reset_required = matches!(target_os, TargetPlatform::Windows)
    && settings.windows().can_sign()
    && package_types.len() > 1;
```

So I remembered correctly we do this but it's just for Windows right now. I'm assuming simply extending this approach to Linux will be enough, right? It's basically boils down to the same issue. I will work on this later today.

FabianLars: in the future we should look into the "work on temp copy only" approach but that's not urgent

crates/tauri-bundler/src/bundle.rs
```diff
@@ -92,22 +97,16 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
     .expect("Main binary missing in settings");
   let main_binary_path = settings.binary_path(main_binary);
 
-  // When packaging multiple binary types, we make a copy of the unsigned main_binary so that we can
-  // restore it after each package_type step. This avoids two issues:
+  // We make a copy of the unsigned main_binary so that we can restore it after each package_type step.
```
Legend-Master, line 114: Let's add a TODO for switching to the "work on temp copy only" approach

crates/tauri-bundler/src/bundle.rs
```diff
@@ -172,6 +162,14 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
       package_type: package_type.to_owned(),
       bundle_paths,
     });
+
+    // Restore unsigne and unpatched binary
```
```suggestion
    // Restore unsigned and unpatched binary
```

crates/tauri-bundler/src/error.rs
```diff
@@ -96,24 +96,14 @@ pub enum Error {
   #[error("Binary parse error: `{0}`")]
   BinaryParseError(#[from] goblin::error::Error),
   /// Package type is not supported by target platform
-  #[error("Wrong package type {0} for platform {1}")]
-  InvalidPackageType(String, String),
+  #[error("Wrong package type {0}")]
```
Legend-Master: Not sure if we could make breaking changes here

@FabianLars any thoughts?

crates/tauri-bundler/src/bundle.rs
Legend-Master: So just to be clear, are we going to use this method for binary patching exclusively or do we want to use this as a fallback? (e.g. the find link section thing -> find the string and replace)

crates/tauri-bundler/src/error.rs
kandrelczyk: This was only used by the binary patching code. I don't think any external code will use this error as it's very specific.

kandrelczyk:
I've change this to be the only method. I don't see any benefits in using the old method over this.

crates/tauri-bundler/src/error.rs
FabianLars: we have downstream bundler users so ideally we don't break here. also not sure if we wanted to keep the old logic around or not @Legend-Master ?

FabianLars:
since the old method didn't work well i agree but i really don't mind either so i'm fine with what tony decides lol - i do like how this here has a seemingly way simpler impl

Legend-Master:
Let's just roll with this one exclusively

crates/tauri-bundler/src/error.rs
Legend-Master: I think we keep `InvalidPackageType` the same and mark the other variants deprecated?

FabianLars: yeah yeah - my ping was a bit offtopic here. Just deprecate it, change the comment to say that it's unused and move on.

.changes/fix-binary-patching.md
```diff
@@ -0,0 +1,8 @@
+---
+"tauri": patch:perf
+"tauri-cli": patch:perf
+"tauri-bundler": patch:perf
+"@tauri-apps/cli": patch:perf
```
Legend-Master:
```suggestion
"tauri": patch:changes
"tauri-cli": patch:changes
"tauri-bundler": patch:changes
"@tauri-apps/cli": patch:changes
```

crates/tauri-bundler/src/error.rs
```diff
@@ -96,24 +96,18 @@ pub enum Error {
...
-    error("__TAURI_BUNDLE_TYPE variable not found in binary. Make sure tauri crate and tauri-cli are up to date")
-  )]
+  #[error("__TAURI_BUNDLE_TYPE variable not found in binary. Make sure tauri crate and tauri-cli are up to date")]
   MissingBundleTypeVar,
   /// Failed to write binary file changed
   #[error("Failed to write binary file changes: `{0}`")]
   BinaryWriteError(String),
-  /// Invalid offset while patching binary file
-  #[error("Invalid offset while patching binary file")]
-  BinaryOffsetOutOfRange,
```
Legend-Master, line 116: We need to keep this (and deprecate it)

crates/tauri-bundler/src/error.rs
```diff
@@ -96,24 +96,18 @@ pub enum Error {
   #[error("Binary parse error: `{0}`")]
   BinaryParseError(#[from] goblin::error::Error),
   /// Package type is not supported by target platform
+  #[deprecated(note = "User WrongPackageType instead")]
   #[error("Wrong package type {0} for platform {1}")]
   InvalidPackageType(String, String),
```
Legend-Master, line 100: Any reason we change this to `WrongPackageType` and removed the platform info?

Legend-Master (commented):
Generally looking good, seems to work on Windows from testing, thanks again for the help on this
Just some nitpicks

crates/tauri-bundler/src/error.rs
kandrelczyk: I merged both Linux and Windows cases in one function so I thought the platform validation doesn't apply anymore but after thinking about it I split it again 

```
   #[cfg(target_os = "linux")]
    let bundle_type = match package_type {
      crate::PackageType::Deb => b"__TAURI_BUNDLE_TYPE_VAR_DEB",
      crate::PackageType::Rpm => b"__TAURI_BUNDLE_TYPE_VAR_RPM",
      crate::PackageType::AppImage => b"__TAURI_BUNDLE_TYPE_VAR_APP",
      _ => {
        return Err(crate::Error::InvalidPackageType(
          package_type.short_name().to_owned(),
          "Linux".to_owned(),
        ))
      }
    };
    #[cfg(target_os = "windows")]
    let bundle_type = match package_type {
      crate::PackageType::Nsis => b"__TAURI_BUNDLE_TYPE_VAR_NSS",
      crate::PackageType::WindowsMsi => b"__TAURI_BUNDLE_TYPE_VAR_MSI",
      _ => {
        return Err(crate::Error::InvalidPackageType(
          package_type.short_name().to_owned(),
          "Windows".to_owned(),
        ))
      }
    };
'''

So I've restored this error.

**Comment by kandrelczyk** [comment]
Totally forgot about Mac. I disabled patching on Mac since we already handle it in `pub fn bundle_type()` in platform.rs. Or do we prefer to patch it and change this method? 

```
match __TAURI_BUNDLE_TYPE {
      "__TAURI_BUNDLE_TYPE_VAR_DEB" => Some(BundleType::Deb),
      "__TAURI_BUNDLE_TYPE_VAR_RPM" => Some(BundleType::Rpm),
      "__TAURI_BUNDLE_TYPE_VAR_APP" => Some(BundleType::AppImage),
      "__TAURI_BUNDLE_TYPE_VAR_MSI" => Some(BundleType::Msi),
      "__TAURI_BUNDLE_TYPE_VAR_NSS" => Some(BundleType::Nsis),
      _ => {
        if cfg!(target_os = "macos") {
          Some(BundleType::App)
        } else {
          None
        }
      }
```

would be

```
match __TAURI_BUNDLE_TYPE {
      "__TAURI_BUNDLE_TYPE_VAR_DEB" => Some(BundleType::Deb),
      "__TAURI_BUNDLE_TYPE_VAR_RPM" => Some(BundleType::Rpm),
      "__TAURI_BUNDLE_TYPE_VAR_APP" => Some(BundleType::AppImage),
      "__TAURI_BUNDLE_TYPE_VAR_MSI" => Some(BundleType::Msi),
      "__TAURI_BUNDLE_TYPE_VAR_NSS" => Some(BundleType::Nsis),
      "__TAURI_BUNDLE_TYPE_VAR_MAC" => Some(BundleType::App),
      _ => None
      }
```

**Comment by Legend-Master** [comment]
I think we could change that later if we actually add another bundler for mac

Also, we probably want to fix those warnings 🙃

> https://github.com/tauri-apps/tauri/actions/runs/19749488088/job/56598652144?pr=14521

**Comment by kandrelczyk** [comment]
Fixed :+1:

File: .changes/fix-binary-patching.md [hunk 8]
```diff
@@ -0,0 +1,8 @@
+---
+"tauri": patch:changes
```
Legend-Master: @FabianLars do we want to delay this to next minor? since this isn't compatible with our current tauri and cli minor versions combo

Legend-Master (dismissed):

.changes/fix-binary-patching.md
FabianLars: sounds reasonable

FabianLars:
tested with strip=true on linux, will do windows later

FabianLars (approved):

PR #14560, opened 2025-11-25
fix(cli): apply CliOptions to the initial target build

lucasfernog:
I noticed this in a workspace where i needed to disable default-features otherwise the crate wouldn't compile for Android, and the build failed with the crate I needed to disable being compiled anyway. This happens because we run an initial build for caching and we were using CliOptions::default() instead of actually creating a proper instance with the provided values.


FabianLars (approved):
untested

lucasfernog:
fixed by #14932

PR #14561, opened 2025-11-25
fix: remove \\r from schema files on windows

FabianLars:
i know this is insane but i've been losing my mind over the last months over this


Legend-Master (dismissed):
Does it make sense to also do it to `generate_capability_schema`?

FabianLars:
i don't think i've seen this in those files yet but it can't hurt

Legend-Master (approved):

PR #14567, opened 2025-11-26
feat: add recent document api support.

Kokoro2336:
add recent documents api support for issue  @open#14460
now only available on macos and windows


FabianLars:
Thanks for the PR! Unfortunately i think this should live in https://github.com/tauri-apps/plugins-workspace/ instead :/

Kokoro2336:
> Thanks for the PR! Unfortunately i think this should live in https://github.com/tauri-apps/plugins-workspace/ instead :/

fine, i'll migrate it later. but one thing i'm confused about is which plugin should these apis lie in. i'm new to this proj, could you give me some hints please?

FabianLars:
it should be its own plugin - you can copy the template from https://github.com/tauri-apps/plugins-workspace/tree/v2/shared/template into the plugins dir. It doesn't have to be perfect, I can help with if needed since we didn't exactly streamlined this yet.

Kokoro2336:
> it should be its own plugin - you can copy the template from https://github.com/tauri-apps/plugins-workspace/tree/v2/shared/template into the plugins dir. It doesn't have to be perfect, I can help with if needed since we didn't exactly streamlined this yet.

thank you so much!

PR #14578, opened 2025-11-29
docs(android): add dynamic library name guide for plugins (fix #14569)

### LINKED ISSUE #14569: [docs] [android] retrieve the app lib name dynamically in Kotlin
Hi ! 

First, thank you for this whole project.

I'm building a plugin that calls some Rust methods as explained in [this section](https://tauri.app/develop/plugins/develop-mobile/#calling-rust-from-mobile-plugins) of the docs.
It works as intended, except that the lib name to load is not dynamic at all. 

We should provide a way for plugin writers to retrieve the name of the generated lib by rust (I didn't find myself how to do this). Otherwise plugins that rely on this feature are not really plugin, since they require changing manually the name of the lib to link. 

If someone knows a workaround, I would be happy to open a PR to document this case.

Subham-KRLX:
Closes #14569

Documents how to properly load native libraries in Android plugins with dynamic naming instead of hardcoding `System.loadLibrary("app_lib")`.

Adds guide explaining:
- Library naming convention from Cargo.toml
- Portable code pattern using constants
- Error handling and quick reference

Reference for updating tauri-apps/tauri-docs.


Subham-KRLX:
@IT-ess ,@FabianLars can you please review this.

cntrvsy:
> @IT-ess ,@cntrvsy can you please review this.
 Wish I could, try reaching out on the discord 
https://discord.gg/tauri

IT-ess:
Thanks for the PR ! 
This is a first step, but this doc will mainly be useful for the users of plugin that do need Rust lib linking. The actual issue is for plugin developers to make the lib linking dynamic, and that part should be documented instead. This kind of knowledge would allow plugin devs to make more flexible plugins while keeping them "plug and play". 
I don't know if this requires changes in Tauri's code, or if the solution is already out there but I just don't know how to do it.

Subham-KRLX:
Thanks for the feedback @IT-ess!
To clarify the scope: should this PR document how plugin **developers** can implement automatic library loading (so users don't need to configure anything), or should it stay focused on helping plugin **users** adopt better practices?
If it's the former, does Tauri already support automatic library name detection, or would that require code changes? Happy to research and expand the PR accordingly.

IT-ess:
It is the former, and I don't know if that requires modifications in the code. I guess it does need some, otherwise it would probably have been documented either

Subham-KRLX:
@IT-ess Found the solution! Plugin devs can auto-generate the library name using Gradle BuildConfig:

**In `android/build.gradle.kts`:**
```kotlin
android {
    defaultConfig {
        // Read from Cargo.toml at build time
        val cargoToml = file("../../Cargo.toml")
        val libraryName = cargoToml.readText()
            .lines()
            .find { it.trim().startsWith("name") }
            ?.substringAfter("\"")?.substringBefore("\"")
            ?.replace("-", "_") ?: "unknown"
        
        buildConfigField("String", "LIBRARY_NAME", "\"$libraryName\"")
    }
    buildFeatures { buildConfig = true }
}
In Example.kt:

kotlin
companion object {
    init {
        System.loadLibrary(BuildConfig.LIBRARY_NAME) // Auto-generated!
    }
}

**Comment by IT-ess** [comment]
I'm not sure but the config should be added to the build.gradle.kts of the plugin user, not the plugin itself ? :thinking: 
I'll give it a try later, I don't have my computer with me right now

**Comment by Subham-KRLX** [comment]
@IT-ess It goes in the **plugin's** [build.gradle.kts](cci:7://file:///Users/subhamsangwan/tauri/crates/tauri-cli/templates/plugin/android/build.gradle.kts:0:0-0:0) not the user's app.

Flow:
1. Plugin developer (you) adds BuildConfig to your plugin's [android/build.gradle.kts](cci:7://file:///Users/subhamsangwan/tauri/crates/tauri-cli/templates/plugin/android/build.gradle.kts:0:0-0:0) once
2. Plugin users just install your plugin - no config needed ✅

**Structure:**
tauri-plugin-yourplugin/ └── android/ ├── build.gradle.kts ← BuildConfig HERE (plugin dev sets up once) └── src/.../Example.kt ← Uses BuildConfig.LIBRARY_NAME

user-app/ └── (just imports plugin, no changes needed)


This makes **your plugin** automatically load the correct library so users don't have to configure anything. Try it out when you get a chance!

**Comment by IT-ess** [comment]
I just tested that, and it doesn't work at all :face_with_diagonal_mouth:

**Comment by Subham-KRLX** [comment]
@IT-ess Could you share what error you got when testing the BuildConfig approach? That will help me find the right solution.

**Comment by IT-ess** [comment]
Your method doesn't work at all. 
I found out an easy way to inject the app lib through gradle config: 
1. When loading the native lib, use : `System.loadLibrary(BuildConfig.TAURI_LIBRARY_NAME)`
2. In the `build.gradle.kts` file of the plugin, add those two lines here : 
```kts
android {
 ...

    defaultConfig {
        ...
        val libName = project.findProperty("tauri_app_lib_name") as? String ?: "tauri_app_lib"
        buildConfigField("String", "TAURI_LIBRARY_NAME", "\"$libName\"")
    }
    buildFeatures {
        buildConfig = true
    }
...
}
```
3. In the `gradle.properties` file of the plugin user, add a line and paste the app lib name in this variable `tauri_app_lib_name=tauri_app_lib`

I used this approach in my [share intent plugin](https://github.com/IT-ess/tauri-plugin-mobile-sharetarget). The best would be to have a standard variable that all plugin devs would use, I used `tauri_app_lib_name`, it seems quite straightforward to me.

PR #14586, opened 2025-11-30
feat: change Webview::navigate to accept WebviewUrl instead of Url

### LINKED ISSUE #14583: [feat] `Webview::navigate` should take `WebviewUrl` as param instead of raw `Url`
### Describe the problem

I'm trying to perform a navigation to an external Url and then back to the app after the external Url invokes some command. I'm facing an issue with the navigation back to the app.

I used the following to go back to the app (since I'm on mobile platform):

```
webview_window.navigate("http://tauri.localhost/".parse()?);
```

The problem I am facing is that the proxy to the development viteserver is not working?? (not sure):

Image

Probably because there is a lot of logic for handling the App url when setting up a webview:
https://github.com/tauri-apps/tauri/blob/dev/crates/tauri/src/manager/webview.rs#L414-L432

I would like to know  how to perform the navigation but also I think `webview.navigate` should actually take a `WebviewUrl` as input since that is also what we use when setting up the webview programmatically!

### Alternatives considered

_No response_

### Additional context

_No response_

surajkewat72:
This commit addresses issue #14583 by changing the  and
 methods to accept  as a parameter
instead of the raw  type.

Changes:
- Updated  to accept  parameter
- Updated  to accept  parameter
- Added  helper method that converts  to using the same logic as webview initialization (handles dev server proxying, app URLs, external URLs, and custom protocols)
- Added comprehensive documentation examples for both methods

Benefits:
- Type-safety: Ensures URLs passed to navigate use the same type as webview creation
- Consistency: Uses the same URL handling logic as webview initialization
- Fixes dev server proxy issues: Properly handles localhost/dev server URLs
- Better API: More intuitive for developers using WebviewUrl elsewhere

Breaking Change:
This is a breaking change as the method signature has changed from:
  pub fn navigate(&self, url: Url) -> crate::Result<()>
to:
  pub fn navigate(&self, webview_url: WebviewUrl) -> crate::Result<()>

Users will need to update their code to use WebviewUrl instead of raw Url.

Migration example:
Before: webview.navigate("http://tauri.localhost/".parse()?) After:  webview.navigate(WebviewUrl::App("index.html".into()))

Closes #14583


crates/tauri/src/webview/mod.rs
```diff
@@ -1667,8 +1667,77 @@ tauri::Builder::default()
       .map(|url| url.parse().map_err(crate::Error::InvalidUrl))?
   }
 
+  /// Converts a [`WebviewUrl`] to a [`Url`] using the same logic as webview initialization.
+  fn webview_url_to_url(&self, webview_url: WebviewUrl) -> crate::Result<Url> {
```
yashsriv, line 1671: Can we share this logic between the two callsites instead of duplicating it?

Issue #12276, opened 2025-01-06
[bug] App filesystem paths do not resolve correctly on iOS and Android


velocitysystems:
### Describe the bug

[appDataDir](https://v2.tauri.app/reference/javascript/api/namespacepath/#appdatadir) does not resolve filesystem paths correctly (or consistently) on iOS and Android.

| OS | Expected | Actual | Comments |
| --- | --- | --- | --- |
| macOS | `.../Library/Application Support/com.company.appname` | `.../Library/Application Support/com.company.appname` | ✅ |
| Android | `/data/user/0/com.company.appname` | `/data/user/0/com.company.appname` |  ✅  |
| iOS | `.../Data/Application/5537853B-26D5-4B30-B99E-373146D19848` | `.../Data/Application/5537853B-26D5-4B30-B99E-373146D19848/Library/Application Support/com.company.appname` |  ❌  |

### Reproduction

Use the [appDataDir](https://v2.tauri.app/reference/javascript/api/namespacepath/#appdatadir) API from @tauri-apps/api.

### Expected behavior

**macOS**
- Unsigned applications typically store application data under: `~/Library/Application Support/[App Name]`
- Signed (sandboxed) macOS applications have directories located under the ~/Library/Containers folder. E.g. `~/Library/Containers/[Bundle Identifier]/Data/[Subfolder]`

**iOS**
- Sandbox paths are managed by the system and vary for each app. Each app gets its unique directory under /var/mobile/Containers E.g. `/var/mobile/Containers/Data/Application/[UUID]/[Subfolder]`

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 15.2.0 arm64 (X64)
    ✔ Xcode Command Line Tools: installed
    ✔ rustc: 1.83.0 (90b35a623 2024-11-26)
    ✔ cargo: 1.83.0 (5ffbef321 2024-10-29)
    ✔ rustup: 1.27.1 (54dd3d00f 2024-04-24)
    ✔ Rust toolchain: stable-aarch64-apple-darwin (default)
    - node: 22.12.0
    - pnpm: 9.15.2
    - npm: 10.9.0

[-] Packages
    - tauri 🦀: 2.1.1
    - tauri-build 🦀: 2.0.3
    - wry 🦀: 0.47.2
    - tao 🦀: 0.30.8
    - @tauri-apps/api : 2.1.1 (outdated, latest: 2.2.0)
[...]

PR #14598, opened 2025-12-02
feat: Add AppLibrary path (#12276)

onehumandev:
feat: Add support for AppLibrary (#12276)

Previously, there was no access to this specific path used by iOS.

We have added support for this directory.

There was a thought that we would be able to add access to this directory
by setting a feature flag in the plugin and having conditional retrieval of
the needed directories.

However, this proved to be much more complex than needed, since in reality we are able to 
make this a purely additive change.

Comparisons of the returned values before and after these changes for both Android and iOS are
attached here.
Android comparison:
[PostChangeResults-Android.csv](https://github.com/user-attachments/files/23892108/PostChangeResults-Android.csv)

iOS comparison:
[PostChangeResults-iOS.csv](https://github.com/user-attachments/files/23892112/PostChangeResults-iOS.csv)

In this process, we have added a page to the sample views to immediately
see the output of the different path methods.

This required some visual fixes of the examples/api app for the menu button
to be usable on modern Android devices to work with insets.)


.changes/add-ios-paths.md
```diff
@@ -0,0 +1,7 @@
+---
+"@tauri-apps/api": patch:enhance
+---
+
+Add support for AppDocuments, AppLibrary, and AppTemp folders.
```
velocitysystems: The change notes mention "AppDocuments, AppLibrary and AppTemp" but the PR title mentions "AppLibrary, AppTemp and AppCache". Does one of these need to be updated?

crates/tauri/src/path/desktop.rs
```diff
@@ -268,6 +268,29 @@ impl<R: Runtime> PathResolver<R> {
       .map(|dir| dir.join(&self.0.config().identifier))
   }
 
+  /// Returns the path to the app's documents directory.
+  ///
+  /// Not supported on desktop, returns error.
+  pub fn app_documents_dir(&self) -> Result<PathBuf> {
+    Err(Error::UnknownPath)
```
velocitysystems: Wouldn't we want these methods to resolve (or fall back) on desktop? Otherwise we'd need to handle desktop/mobile paths separately in the application.

onehumandev: I'm not sure what a sensible fallback would be for these on Desktop implementations would be, since these may not have direct translations to MacOS and Windows apps.
Do you have any thoughts on what a reasonable fallback would be for these methods on desktop?

crates/tauri/src/path/android.rs
```diff
@@ -152,6 +152,21 @@ impl<R: Runtime> PathResolver<R> {
     self.call_resolve("getCacheDir")
   }
 
+  /// Returns the path to the app's documents directory.
+  pub fn app_documents_dir(&self) -> Result<PathBuf> {
+    self.call_resolve("getFilesDir")
```
velocitysystems: Both of these methods call `getFilesDir` and return the same path. What about external files directory on Android? Would this be a better fit for `app_documents_dir`? How is this pattern handled in other platforms like Expo, MAUI etc?

crates/tauri/src/path/ios.rs
```diff
@@ -0,0 +1,257 @@
...
+  /// Returns the path to the suggested directory for your app's cache files.
+  ///
+  /// Resolves to `FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first!`
+  pub fn app_cache_dir(&self) -> Result<PathBuf> {
+    self.call_resolve("getCacheDir")
+  }
+
+  /// Returns the path to the app's documents directory.
+  ///
+  /// Resolves to `FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first!.
```
velocitysystems: Nit: Missing closing backtick (`) at the end.

examples/api/src-tauri/tauri.conf.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "../../../crates/tauri-schema-generator/schemas/config.schema.json",
   "productName": "Tauri API",
   "version": "1.0.0",
-  "identifier": "com.tauri.api",
+  "identifier": "com.onehumandev.tauri.api",
```
velocitysystems: Nit: Roll back this change.

velocitysystems (changes requested):

crates/tauri/src/path/desktop.rs
velocitysystems: I guess the key question is: **What is the intent of these APIs?**

| Directory | iOS Purpose | Potential Desktop Equivalent |
|-----------|-------------|------------------------------|
| `appDocumentsDir` | User-visible files, backed up to iCloud | `documentDir()` or `appDataDir()` |
| `appLibraryDir` | App support files, hidden from user | `appDataDir()` or `appLocalDataDir()` |
| `appTempDir` | Temporary files, can be purged | `tempDir()` |

1. **If we want "a place to store user documents":** Fallback to `document_dir()` makes sense
2. **If we want "iOS-specific iCloud-synced location":** Fallback would be misleading; error is better
3. **If we want "temporary scratch space":** `app_temp_dir()` → `temp_dir()` is a safe fallback

@onehumandev Using Claude to do some research on the matter....
### How Cross-Platform Frameworks Handle Path APIs
| Framework | Strategy | Unavailable Path Behavior |
|-----------|----------|---------------------------|
| **Flutter** | Unified methods, platform-specific implementations | Throws exception |
| **Electron** | Unified methods, always finds equivalent | Throws error (rare) |
| **React Native** | Platform-specific constants | Returns `undefined` |
| **Expo** | Unified `Paths` object, mostly cross-platform | Property may be empty |
| **Tauri (PR)** | Mobile-specific methods | Returns `Err(UnknownPath)` |

---

### Flutter (`path_provider`)

**One method per concept**, resolves to closest platform equivalent:

```dart
getApplicationDocumentsDirectory()  // Works everywhere
getTemporaryDirectory()             // Works everywhere
getApplicationSupportDirectory()    // Works everywhere
```

### Electron

**Unified `app.getPath(name)` API** - always returns a path:

```javascript
app.getPath('documents')  // All platforms
app.getPath('userData')   // All platforms  
app.getPath('temp')       // All platforms
```

### React Native (`react-native-fs`)

**Platform-specific constants**, some only available on certain platforms:

```javascript
RNFS.DocumentDirectoryPath    // iOS + Android
RNFS.LibraryDirectoryPath     // iOS only
RNFS.DownloadDirectoryPath    // Android + Windows only
```

### Expo (`expo-file-system`)

**Unified `Paths` object** with cross-platform directories:

```javascript
import { Paths } from 'expo-file-system';

Paths.document  // iOS + Android (app's document directory)
Paths.cache     // iOS + Android (can be purged by system)
Paths.bundle    // iOS + Android (bundled assets, read-only)
```

crates/tauri/src/path/android.rs
onehumandev: I have mixed thoughts on where the entry point for `getExternalFilesDir` should exist. 
I lean strongly toward it being a completely separate method because it simply doesn't exist on the other platforms, and I don't know how  much we are expecting devs to fully read the docs to understand the mapping from API method to the actual implementation.  
However, if this all should be handled in Tauri V3, perhaps this new method can point to `getExternalFilesDir`, and we can add comments at the Rust and TS layers describing the behavior on Android.
Does that make sense @velocitysystems ?

## Cross-platform comparison (brought to you by Claude 😄 ) 

### Expo (React / Expo FileSystem)
Exposes only app-scoped directories (`documentDirectory`, `cacheDirectory`).
On Android these map to internal app storage (effectively `Context.getFilesDir()` / `getCacheDir()`).
There is no API surface for `getExternalFilesDir()`; user-visible files are handled via sharing or pickers.
**Rationale:** maintain strict iOS parity and hide Android-specific storage topology.

### .NET MAUI
Exposes `AppDataDirectory` and `CacheDirectory`, which on Android map to internal app storage (`getFilesDir()`, `getCacheDir()`).
MAUI does not expose external app storage in the cross-platform API.
Access to `getExternalFilesDir()` requires platform-specific Android code, intentionally outside the portable abstraction.

### Capacitor (used with web stacks, often compared to Electron on mobile)
Exposes a `Directory` enum (`Data`, `Cache`, `Documents`, `External`, `ExternalStorage`, etc.).
On Android, the native plugin explicitly maps:

- `Data` → `getFilesDir()` (internal storage)
- `Cache` → `getCacheDir()` (internal storage)
- `Documents` → `getExternalFilesDir(Environment.DIRECTORY_DOCUMENTS)` (app-specific external storage)
- `External` → `getExternalFilesDir(null)` (app-specific external storage root)
- `ExternalStorage` → primary shared/external storage (separate from `getExternalFilesDir()`; requires special permissions on Android 10+)

This is the only examined cross-platform stack that intentionally exposes Android external app storage as a first-class concept.

### Electron (desktop)
Not applicable to Android; uses OS-level concepts (`userData`, `documents`, `downloads`).
There is no internal/external distinction analogous to Android; mobile behavior depends on wrappers (e.g., Capacitor).

## Summary
Most cross-platform frameworks (Expo, MAUI) collapse Android internal vs external app storage into a single "app data" abstraction to preserve iOS parity. Capacitor is the notable exception, explicitly modeling `getExternalFilesDir()` in its cross-platform API, with the caveat that `ExternalStorage` access requires additional permissions on Android 10+ due to scoped storage restrictions.

crates/tauri/mobile/ios-api/Sources/Tauri/PathPlugin.swift
```diff
@@ -0,0 +1,54 @@
+// Copyright 2019-2024 Tauri Programme within The Commons Conservancy
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: MIT
+
+import Foundation
+
+class PathPlugin: Plugin {
+  private let fileManager = FileManager.default
```
FabianLars: Is this not available in objective-c? I would prefer a rust solution if possible.
i do see eg https://developer.apple.com/documentation/foundation/filemanager/urls(for:in:)?language=objc which is here https://docs.rs/objc2-foundation/0.3.2/objc2_foundation/struct.NSFileManager.html

the only relevant comment on this topic i see is
> This also required the creation of a Swift plugin to allow us to access
these iOS APIs.

so i'm not sure about the exact reasons here.

crates/tauri/src/path/ios.rs
```diff
@@ -0,0 +1,257 @@
+// Copyright 2019-2024 Tauri Programme within The Commons Conservancy
```
FabianLars: if my other comment about using objc is possible to do i think i'd prefer to handle this in desktop.rs via a couple cfg flags than all of this duplication

packages/api/src/path.ts
```diff
@@ -179,6 +225,75 @@ async function appCacheDir(): Promise<string> {
...
+ * const appTempDirPath = await appTempDir();
+ * ```
+ *
+ * @since 2.0.0
+ */
+async function appTempDir(): Promise<string> {
+  return invoke('plugin:path|resolve_directory', {
+    directory: BaseDirectory.AppTemp
+  })
+}
```
FabianLars: i still believe appTempDir should not be a thing, if needed we can "break" tempDir but looking at your csv files there's no difference between appTempDir and tempDir?

i admit that i forgot most of the conversations so i may be missing something

FabianLars (commented):
not a full review yet, just getting back into it 🙃

packages/api/src/path.ts
onehumandev: My question is how this would translate to desktop.
Specifically, for desktop platforms, is `tempDir` expected to go to a directory owned by the application itself (within the app sandbox), or to an OS-defined temp directory?
If the former, then I concur that `appTempDir` is superfluous and can be removed.
If the latter, then I am thinking that desktop implementations could benefit from `appTempDir` as well .

Apologies if this was clarified earlier, as I too have forgotten many of the details when this was initially discussed :)

FabianLars: hmm, that's actually a good point. Looking at the thousands of files in the root of my temp dir i'm not sure how valuable an app-specific dir really is, especially since you probably still need some kind of randomizer/versioning even in an app-specific dir (or for that dir itself). 🤔 

i'm kinda leaning towards it now but i'll ask the others for opinions later.

Legend-Master: Is `AppTemp` a different path from `Temp` on iOS?  From the docs, it seems to be the same to me

> https://doc.rust-lang.org/nightly/std/env/fn.temp_dir.html

FabianLars: no, on mobile both are the same, see csv in pr description. the question is really just about desktop now. but we could also remove this part from the pr and revisit later.

Legend-Master: My brain can't handle those paths right now, I think we should remove for now if that's not implemented anywhere

onehumandev: Updated the PR to remove `appTempDir` from the code, and from the PR title/description.
We can use `tempDir` for now on iOS and Android.
c. @FabianLars @velocitysystems

packages/api/src/path.ts
```diff
@@ -180,6 +222,64 @@ async function appCacheDir(): Promise<string> {
...
+ * #### Platform-specific
+ *
+ * - **iOS:** Resolves to `NSTemporaryDirectory()` (native Foundation APIs)
+ * - **Android:** Resolves to `context.cacheDir` (via `getCacheDir()` in PathPlugin.kt)
+ * - **Desktop:** Not supported, returns null
+ *
+ * @example
+ * ```typescript
+ * ```
+ *
```
FabianLars: leftover?

FabianLars:
I am kinda failing to see the point of appDocumentsDir. On iOS it's the same as documentsDir (and user-facing) and on android it's the same as appLibraryDir (non user facing) which a) sounds confusing and b) doesn't solve anything(?)
Edit: Maybe it makes sense to further split up the PR either way to finally get the least controversial one in 😂 

I feel like with all the discussions and the long breaks (which is totally our fault) we kinda lost the plot here. Like, what are we actually trying to solve? => **Which paths are we currently missing in Tauri?**

Bonus questions:
Which paths in the current API are incorrect (pending fix in v3?)
Which path functions would be nice to have? For example to solve a case where you use different method per platform to store the same kind of data.

.changes/add-ios-paths.md
```diff
@@ -0,0 +1,9 @@
+---
+"@tauri-apps/api": patch:enhance
```
FabianLars, line 3:
```suggestion
tauri: patch:enhance
"@tauri-apps/api": patch:enhance
```

crates/tauri/src/path/android.rs
```diff
@@ -152,6 +152,21 @@ impl<R: Runtime> PathResolver<R> {
...
+
+  /// Returns the path to the app's library directory.
+  pub fn app_library_dir(&self) -> Result<PathBuf> {
+    self.call_resolve("getFilesDir")
+  }
+
+  /// Returns the path to the app's temporary directory.
+  pub fn app_temp_dir(&self) -> Result<PathBuf> {
+    self.call_resolve("getCacheDir")
+  }
```
FabianLars: i don't have my glasses on but i think this is a leftover as it's part of what was removed from this PR per the discussion with Tony?

crates/tauri/src/path/desktop.rs
```diff
@@ -262,12 +262,54 @@ impl<R: Runtime> PathResolver<R> {
...
+  /// This method is meant for sandboxed apps, specifically on mobile platforms.
+  /// On iOS, resolves to `NSTemporaryDirectory()`.
+  /// On desktop, resolves to [`temp_dir`](Self::temp_dir).
+  pub fn app_temp_dir(&self) -> Result<PathBuf> {
+    #[cfg(target_os = "ios")]
+    return super::ios_native::app_temp_dir().ok_or(Error::UnknownPath);
+    #[cfg(not(target_os = "ios"))]
+    self.temp_dir()
+  }
+}
```
FabianLars: same

crates/tauri/src/path/ios_native.rs
```diff
@@ -0,0 +1,56 @@
+// Copyright 2019-2024 Tauri Programme within The Commons Conservancy
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: MIT
+
+//! Native iOS path resolution using Foundation APIs via objc2-foundation.
+//! Replaces the Swift PathPlugin for app_cache_dir, app_documents_dir,
+//! app_library_dir, and app_temp_dir.
```
FabianLars:
```suggestion
//! app_library_dir
```

same

crates/tauri/src/path/ios_native.rs
```diff
@@ -0,0 +1,56 @@
...
+
+/// Returns the app's temporary directory (equivalent to NSTemporaryDirectory()).
+pub fn app_temp_dir() -> Option<PathBuf> {
+  let s = unsafe { NSTemporaryDirectory() };
+  let path = s.to_string();
+  if path.is_empty() {
+    return None;
+  }
+  Some(PathBuf::from(path))
+}
```
FabianLars: same

velocitysystems:
Thanks for the review @FabianLars. Just weighing in on the above comments. @onehumandev I'd suggest we narrow the scope of this PR to just `appLibraryDir` — access to the iOS Library folder. That's the one path Tauri genuinely doesn't expose right now.

If we can slim the PR down to just that (plus cleanup), we should have a clear path to merge. Then everything else — `appDocumentsDir`, the broader platform-specific API surface — belongs back in the [issue](https://github.com/tauri-apps/tauri/issues/12276).

onehumandev:
@FabianLars @velocitysystems  Updated per recommendations, so we should be good to go. Thanks!

FabianLars:
ngl now that we talked about the new platform specific path apis i kinda want to expose appLibrary as an iOS specific one as well instead 😅 

i'll leave this pr open just a few more days to see if that discussion is going anywhere. sorry for the multiple delays :/

velocitysystems:
Thanks @FabianLars. After discussing with @onehumandev, we’ve decided to abandon this PR and instead move forward with a custom plugin that implements the proposed platform-specific API [approach](https://github.com/tauri-apps/tauri/issues/12276#issuecomment-4112014834). The intention is to mature this externally first, with the proposal of integrating it into Tauri’s FS plugin or core at a later stage.

FabianLars:
Sounds good! We often push that plugin-first approach around here so that makes sense to me. I'll close this PR in the meantime then.
Apologies for essentially wasting your time here as well :/

PR #14604, opened 2025-12-03
perf(cli): optimize npm package version fetching

Magnus1X:
## Description
This PR optimizes the `tauri info` command by batching npm package version checks. Previously, the CLI spawned a new process for each package to check its version, which could be slow.
## What’s changed
- Modified `crates/tauri-cli/src/info/packages_nodejs.rs` and `crates/tauri-cli/src/info/plugins.rs` to use `current_package_versions`.
- The CLI now fetches all required package versions in a single command execution instead of sequentially.
- Added a release log file for the performance improvement.
## Why
This change significantly reduces the overhead of running `tauri info`, especially in projects with many plugins, by minimizing the number of shell commands executed. This results in a faster and more responsive CLI experience.

### DIFF
--- a/crates/tauri-cli/src/info/packages_nodejs.rs
+++ b/crates/tauri-cli/src/info/packages_nodejs.rs
@@ -137,12 +137,27 @@ pub fn items(
 ) -> Vec<SectionItem> {
   let mut items = Vec::new();
   if let Some(frontend_dir) = frontend_dir {
-    for (package, version) in [
-      ("@tauri-apps/api", None),
-      ("@tauri-apps/cli", Some(metadata.js_cli.version.clone())),
-    ] {
+    let mut packages = vec![("@tauri-apps/api", None)];
+    if let Some(cli) = &metadata.js_cli.version {
+      packages.push(("@tauri-apps/cli", Some(cli.clone())));
+    }
+
+    let versions = package_manager
+      .current_package_versions(
+        &packages.iter().map(|(p, _)| p.to_string()).collect::<Vec<_>>(),
+        frontend_dir,
+      )
+      .unwrap_or_default();
+
+    for (package, version) in packages {
       let frontend_dir = frontend_dir.clone();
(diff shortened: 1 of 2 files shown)


Tunglies:
I am dobtful with this really improved performance, and memory side
- Additional vec! allocated and iter it again
- Additional collect::<Vec<_>> inplace
- More clone
- More String allocate
- More collect behavior
- HashMap?

crates/tauri-cli/src/info/packages_nodejs.rs
```diff
@@ -155,12 +170,17 @@ pub fn nodejs_section_item(
   version: Option<String>,
   frontend_dir: PathBuf,
   package_manager: PackageManager,
+  current_version: Option<semver::Version>,
```
Legend-Master, line 173: Is there any differences between this and the `version` parameter above?

crates/tauri-cli/src/info/packages_nodejs.rs
```diff
@@ -137,12 +137,27 @@ pub fn items(
...
   let mut items = Vec::new();
   if let Some(frontend_dir) = frontend_dir {
-    for (package, version) in [
-      ("@tauri-apps/api", None),
-      ("@tauri-apps/cli", Some(metadata.js_cli.version.clone())),
-    ] {
+    let mut packages = vec![("@tauri-apps/api", None)];
+    if let Some(cli) = &metadata.js_cli.version {
+      packages.push(("@tauri-apps/cli", Some(cli.clone())));
+    }
```
Legend-Master, line 143: This change doesn't really do anything, please revert

Legend-Master (changes requested):
Thanks for contributing! Do you mind also adding a change file?

> I am dobtful with this really improved performance, and memory side

This is mainly for batching the package manager calls through migrating to `current_package_versions`, that plays a much bigger role in terms of performance so I think this is worth while the trade offs

PR #14607, opened 2025-12-04
refactor(cli): simplify `features: Option<Vec<String>>` to `Vec<String>`

sftse:
Noticed a code pattern where `None` is used in place of an empty vector. The resulting code has to transform between different representations and liberally uses `get_or_insert` on the `None`. This is a bit roundabout, so use empty `Vec` as a replacement for `None`.

Empty `Vec`s don't allocate, so this changes nothing except make the code shorter.

Not entirely sure, but also think this fixes a bug where the code paths of `None` and `Some(Vec::new())` would have diverged.

### DIFF
--- a/.changes/empty-vec-instead-of-none.md
+++ b/.changes/empty-vec-instead-of-none.md
@@ -3,4 +3,4 @@
 "tauri-cli": patch:enhance
 ---
 
-Replace Option<Vec> with simple Vec
+Simplified internal representation of `features: Option<Vec<String>>` with `Vec<String>`, no user facing changes
--- a/.changes/empty-vec-instead-of-none.md
+++ b/.changes/empty-vec-instead-of-none.md
@@ -1,0 +1,6 @@
+---
+"@tauri-apps/cli": patch:enhance
+"tauri-cli": patch:enhance
+---
+
+Replace Option<Vec> with simple Vec
--- a/crates/tauri-cli/src/interface/rust.rs
+++ b/crates/tauri-cli/src/interface/rust.rs
@@ -51,7 +51,7 @@ pub struct Options {
   pub runner: Option<RunnerConfig>,
   pub debug: bool,
   pub target: Option<String>,
-  pub features: Option<Vec<String>>,
+  pub features: Vec<String>,
   pub args: Vec<String>,
   pub config: Vec<ConfigValue>,
   pub no_watch: bool,
@@ -108,7 +108,7 @@ impl From<crate::dev::Options> for Options {
 #[derive(Debug, Clone)]
 pub struct MobileOptions {
   pub debug: bool,
-  pub features: Option<Vec<String>>,
+  pub features: Vec<String>,
   pub args: Vec<String>,
   pub config: Vec<ConfigValue>,
   pub no_watch: bool,
@@ -393,7 +393,7 @@ fn dev_options(
   mobile: bool,
   args: &mut Vec<String>,
   run_args: &mut Vec<String>,
-  features: &mut Option<Vec<String>>,
+  features: &mut Vec<String>,
   app_settings: &RustAppSettings,
 ) {
   let mut dev_args = Vec::new();
@@ -429,9 +429,7 @@ fn dev_options(
       })
       .collect();
     args.push("--no-default-features".into());
-    if !enable_features.is_empty() {
-      features.get_or_insert(Vec::new()).extend(enable_features);
-    }
+    features.extend(enable_features);
   }
 }
 
@@ -498,15 +496,8 @@ fn get_watch_folders(additional_watch_folders: &[PathBuf]) -> crate::Result<Vec<
 }
 
 impl Rust {
-  pub fn build_options(
-    &self,
-    args: &mut Vec<String>,
-    features: &mut Option<Vec<String>>,
-    mobile: bool,
-  ) {
-    features
-      .get_or_insert(Vec::new())
-      .push("tauri/custom-protocol".into());
+  pub fn build_options(&self, args: &mut Vec<String>, features: &mut Vec<String>, mobile: bool) {
+    features.push("tauri/custom-protocol".into());
     if mobile {
       args.push("--lib".into());
     } else {
@@ -957,11 +948,12 @@ impl AppSettings for RustAppSettings {
         .clone()
         .unwrap_or_default();
       for bin in bins {
-        if let (Some(req_features), Some(opt_features)) =
-          (&bin.required_features, &options.features)
-        {
+        if let Some(req_features) = &bin.required_features {
           // Check if all required features are enabled.
-          if !req_features.iter().all(|feat| opt_features.contains(feat)) {
+          if !req_features
+            .iter()
+            .all(|feat| options.features.contains(feat))
+          {
             continue;
           }
         }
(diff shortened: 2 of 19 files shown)


crates/tauri-cli/src/interface/rust.rs
```diff
@@ -957,11 +948,12 @@ impl AppSettings for RustAppSettings {
         .clone()
         .unwrap_or_default();
       for bin in bins {
-        if let (Some(req_features), Some(opt_features)) =
```
sftse, line 960: This code previously had different behavior if options.features was `Some(Vec::new())` (`continue` hit) vs `None` (`continue` not hit). This is odd given the variable names which imply the required features should be present.

crates/tauri-cli/src/mobile/android/mod.rs
```diff
@@ -128,19 +128,18 @@ pub fn command(cli: Cli, verbosity: u8) -> Result<()> {
...
-  }
+  android_options.features.extend_from_slice(features);
 
   let raw = RawAndroidConfig {
-    features: android_options.features.clone(),
+    features: if android_options.features.is_empty() {
+      None
+    } else {
+      Some(android_options.features.clone())
+    },
```
sftse: Did not check how this external type behaves, so this if branch might be overly conservative and replaceable with `Some(android_options.features.clone())`

crates/tauri-cli/src/mobile/android/mod.rs
```diff
@@ -161,7 +160,11 @@ pub fn get_config(
   let metadata = AndroidMetadata {
     supported: true,
     cargo_args: Some(android_options.args),
-    features: android_options.features,
+    features: if android_options.features.is_empty() {
+      None
+    } else {
+      Some(android_options.features)
+    },
```
sftse: Same here, if branch maybe removable.

crates/tauri-cli/src/mobile/mod.rs
```diff
@@ -178,10 +178,16 @@ impl Default for DevHost {
...
+  v.is_empty()
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct CliOptions {
   pub dev: bool,
-  pub features: Option<Vec<String>>,
+  #[serde(default)]
+  #[serde(skip_serializing_if = "is_empty")]
+  pub features: Vec<String>,
```
sftse: This `Serialize` and `Deserialize` impl is the only publically visible way the change of internal fields might be observable. This uses `serde` configuration so serialization strings remain identical.

sftse:
@FabianLars 
Fixed CI

sftse:
I'm a bit confused by how the change sets work, the bot suggested
```text
---
"@tauri-apps/api": patch
"tauri-utils": patch
"tauri-macos-sign": patch
"tauri-bundler": patch
"tauri-runtime": patch
"tauri-runtime-wry": patch
"tauri-codegen": patch
"tauri-macros": patch
"tauri-plugin": patch
"tauri-build": patch
"tauri": patch
"@tauri-apps/cli": patch
"tauri-cli": patch
"tauri-driver": patch
---
```
even though only `tauri-cli` has any changes. Is this correct?

FabianLars:
no the bot just lists all existing packages. just remove all that you didn't touch (eg in your case only tauri-cli needs to be there)

crates/tauri-cli/src/mobile/android/mod.rs
Legend-Master: I don't think it's being used at all currently, so wouldn't matter either ways

crates/tauri-cli/src/mobile/ios/mod.rs
```diff
@@ -232,7 +227,11 @@ pub fn get_config(
             }
           }
         }),
-    ios_features: ios_options.features.clone(),
+    ios_features: if ios_options.features.is_empty() {
```
Legend-Master: Same here

crates/tauri-cli/src/mobile/mod.rs
Legend-Master: I think this is also different from before, it used to be something like `"features": null` in JSON

Since I'm not the most familiar with where is this used, I'll leave it to @FabianLars @lucasfernog

Legend-Master (dismissed):
Thanks

crates/tauri-cli/src/mobile/android/mod.rs
sftse: I'll check, just to make sure this is future-proof, even if it's not used.

Alternatively, if things are not used, should we remove them?

crates/tauri-cli/src/mobile/mod.rs
sftse: Think I checked this (not at workstation rn) and only used internally, will get back to you.

crates/tauri-cli/src/mobile/android/mod.rs
Legend-Master: > Alternatively, if things are not used, should we remove them?

I'm not the most familiar with it, maybe @lucasfernog knows about them a bit more

crates/tauri-cli/src/mobile/mod.rs
sftse: Made a commit highlighting where serialization is used. This seems purely internal, so going with the simpler serialization.

crates/tauri-cli/src/mobile/android/mod.rs
sftse: `cargo-mobile2` does not have control flow that depends on these fields, going with the simpler code.

crates/tauri-cli/src/mobile/ios/mod.rs
sftse: `cargo-mobile2` does not have control flow that depends on these fields, going with the simpler code.

sftse:
RFR

.changes/empty-vec-instead-of-none.md
```diff
@@ -0,0 +1,6 @@
+---
+"@tauri-apps/cli": patch:enhance
+"tauri-cli": patch:enhance
+---
+
+Replace Option<Vec> with simple Vec
```
Legend-Master:
```suggestion
Simplified internal representation of `features: Option<Vec<String>>` with `Vec<String>`, no user facing changes
```

Legend-Master (dismissed):

Legend-Master (dismissed):
Thanks

crates/tauri-cli/src/interface/rust.rs
Legend-Master: Should be a miss in the previous PR

crates/tauri-cli/src/mobile/ios/mod.rs
```diff
@@ -232,7 +227,11 @@ pub fn get_config(
...
+        None
+    } else {
+        Some(ios_options.features.clone())
+    },
```
Legend-Master:
```suggestion
    ios_features: Some(ios_options.features.clone()),
```

leftover, https://github.com/tauri-apps/tauri/pull/14607/files#r2649702462

crates/tauri-cli/src/mobile/mod.rs
Legend-Master: Unresolved this so we get to see this more obviously if anyone ever get back to this PR

Legend-Master (approved):

PR #14611, opened 2025-12-05
feat(bundler): add macos pkg installer support with custom signing

pierrekin:
## Summary

  Adds support for creating macOS PKG installers with both native and custom signing capabilities. This enables distribution
   of Tauri applications via PKG installers and supports HSM-based signing workflows.

  ## Features

  ### PKG Installer Support
  - Creates macOS PKG installers using `pkgbuild` and `productbuild`
  - Requires `distribution.xml` in project root for PKG customization
  - Native signing support via `productsign` using `signingIdentity` or `APPLE_CERTIFICATE` environment variable
  - Respects `--no-sign` flag

  ### Custom Signing Commands
  Adds support for custom signing commands to integrate external signing tools (e.g., HSM-based solutions):
  - **`appSignCommand`**: Custom command for signing .app bundles
  - **`pkgSignCommand`**: Custom command for signing .pkg installers
  - **`dmgSignCommand`**: Custom command for signing .dmg disk images

  Custom commands are checked before native signing and use `%1` placeholder for the artifact path. Commands run in the
  directory where `tauri build` was executed, allowing relative paths.

  ### Configuration Example
  ```json
  {
    "bundle": {
      "macOS": {
        "appSignCommand": {
          "cmd": "./shims/sign_app.sh",
          "args": ["%1"]
        },
        "pkgSignCommand": {
          "cmd": "./shims/sign_pkg.sh",
          "args": ["%1"]
        },
        "dmgSignCommand": {
          "cmd": "./shims/sign_dmg.sh",
          "args": ["%1"]
        }
      }
    }
  }
  ```

  Test Plan

  - Create PKG installer without signing
  - Create PKG installer with native signing (productsign)
  - Test custom signing commands for .app, .pkg, and .dmg
  - Verify relative paths work in custom signing commands
  - Verify --no-sign flag is respected
  - Test with distribution.xml customization

  Modified Crates

  - tauri-bundler: PKG bundler implementation and custom signing logic
  - tauri-utils: Configuration schema for custom signing commands
  - tauri-cli: Schema updates for new configuration fields

### DIFF
--- a/crates/tauri-bundler/src/bundle.rs
+++ b/crates/tauri-bundler/src/bundle.rs
@@ -70,11 +70,15 @@ pub struct Bundle {
 /// Returns the list of paths where the bundles can be found.
 pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
   let mut package_types = settings.package_types()?;
+  log::info!("Initial package types: {:?}", package_types);
+
   if package_types.is_empty() {
+    log::warn!("Package types is empty, returning early");
     return Ok(Vec::new());
   }
 
   package_types.sort_by_key(|a| a.priority());
+  log::info!("Sorted package types: {:?}", package_types);
 
   let target_os = settings.target_platform();
 
@@ -110,8 +114,11 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
   let mut main_binary_signed = false;
   let mut bundles = Vec::<Bundle>::new();
   for package_type in &package_types {
+    log::info!("Processing package type: {:?}", package_type);
+
     // bundle was already built! e.g. DMG already built .app
     if bundles.iter().any(|b| b.package_type == *package_type) {
+      log::info!("Skipping {:?}, already built", package_type);
       continue;
     }
 
@@ -149,6 +156,18 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
           });
         }
         bundled.dmg
+      }
+      // pkg is dependent of MacOsBundle, we send our bundles to prevent rebuilding
+      #[cfg(target_os = "macos")]
+      PackageType::Pkg => {
+        let bundled = macos::pkg::bundle_project(settings, &bundles)?;
+        if !bundled.app.is_empty() {
+          bundles.push(Bundle {
+            package_type: PackageType::MacOsBundle,
+            bundle_paths: bundled.app,
+          });
+        }
+        bundled.pkg
       }
 
       #[cfg(target_os = "windows")]
--- a/.changes/macos-pkg-installer.md
+++ b/.changes/macos-pkg-installer.md
@@ -1,0 +1,20 @@
+---
+tauri-bundler: minor:feat
+tauri-utils: minor:feat
+tauri-cli: minor:feat
+---
+
+Add macOS PKG installer support with custom signing commands.
+
+Implements support for creating macOS PKG installers using pkgbuild and productbuild, with native signing via productsign and support for custom signing commands (useful for HSM-based signing solutions).
+
+Features:
+- Create PKG installers from .app bundles using distribution.xml from project root
+- Native PKG signing with productsign using signingIdentity or APPLE_CERTIFICATE
+- Custom signing command support for .app bundles, .pkg installers, and .dmg disk images
+- Custom commands use %1 placeholder for artifact path and run in build directory for relative path support
+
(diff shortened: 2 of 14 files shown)


crates/tauri-bundler/src/bundle.rs
```diff
@@ -70,11 +70,15 @@ pub struct Bundle {
...
   let mut package_types = settings.package_types()?;
+  log::info!("Initial package types: {:?}", package_types);
+
   if package_types.is_empty() {
+    log::warn!("Package types is empty, returning early");
     return Ok(Vec::new());
   }
 
   package_types.sort_by_key(|a| a.priority());
+  log::info!("Sorted package types: {:?}", package_types);
```
FabianLars, line 81: i won't annotate every line but logs like this one are honestly a bit too verbose, at least for the `info` level, perhaps tracing or debug.

FabianLars (commented):
> Adds support for custom signing commands to integrate external signing tools (e.g., HSM-based solutions):

Are there HSM based solutions for macOS? i know of tools like https://github.com/indygreg/apple-platform-rs (which i'd love to integrate into tauri at some point) but didn't hear about hsm ones yet (or many local solutions either).

For the signCommand config, if we keep it, i think it would make more sense to match the Windows behavior and have a single config and leave it to the scripts to handle the file type check if needed.
Also, for the future it would make reviews much easier to have the pkg impl and the custom signing feature in 2 separate PRs :)

Lastly, we also need to check how to make the updater work with this, eg do we still use the .app bundle to update?

pierrekin:
Thanks so much for this feedback!

> Are there HSM based solutions for macOS? i know of tools like https://github.com/indygreg/apple-platform-rs (which i'd love to integrate into tauri at some point) but didn't hear about hsm ones yet (or many local solutions either).

I can't share too many details, but at our large enterprise we have a team that maintains signing infrastructure that is able to sign iOS and macOS entities in compliance with all of apple's rules and protocols, but (it's my understanding) using HSM to store the actual keys. I am aware of "userspace" rewrites of Apple's macOS tooling, and it would be my guess that they are using one such rewrite ;)

When we use the 'signCommand' we don't directly use the HSM, we use a client binary which speaks to this internal signing infrastructure.

> For the signCommand config, if we keep it, i think it would make more sense to match the Windows behavior and have a single config and leave it to the scripts to handle the file type check if needed.

Fair point, however in the case of signing a package that contains an application bundle, tauri would need to call the custom signing command twice. I guess that is no problem, as long as the signing command understands what to do in each case.

> Also, for the future it would make reviews much easier to have the pkg impl and the custom signing feature in 2 separate PRs :)

Yes, my apologies, I was not fast enough to mark this as draft and keep things self contained. I propose to break this into three manageable PRs for the Tauri team. **Does the following sound reasonable?**

**PR 1: Custom signing for the macOS app bundle types**
  - Custom signing command support for existing macOS app bundle type
  - This is a complete, testable feature that works end-to-end
  
**PR 2: Support for PKG bundles**
  - Support for building PKG bundles
  - With a basic templating system similar to Wix so that users can provide their own distribution.xml and tauri can populate variables like the application name etc
  - With the ability to add extra "products" to the "productbuild" call
  - Builds on top of PR 1 in that it also supports custom signing
  - This is a complete, testable feature that works end-to-end
  
**PR 3: Custom notarisation support for existing bundle types and PKG bundle type**
  - Similar to PR 1 in that it adds custom support, this time for notarisation
  - Builds on top of PR2 in that it can notarize packages
  - This is a complete, testable feature that works end-to-end

> Lastly, we also need to check how to make the updater work with this, eg do we still use the .app bundle to update?
This is an interesting point. I think that still using the `.app` bundle to update is the most "obvious" way, but it is pretty limiting.

I think many more advanced projects package into a `.pkg` because they would like to use the scripting and other features of `.pkg` files to do maintenance or migrations as part of the install.

Another factor is that there is sort of an "outer" .pkg package, and then an inner ".pkg" component within that package. Perhaps the component is what should be downloaded and installed with the automatic updater.

I think philosophically it may come down to, should the updater update the "installation" or the "application"? 🤔 

> i won't annotate every line but logs like this one are honestly a bit too verbose, at least for the info level, perhaps tracing or debug.

My apologies, these were left over from development when I pushed to the wrong branch. 🙈

pierrekin:
Closing this PR, I am going to make new PRs according to the plan documented [in this comment](https://github.com/tauri-apps/tauri/pull/14611#issuecomment-3632757368). I plan on starting work on this in the first few weeks of Jan 2026.

Issue #3109, opened 2021-12-24
[feat] Portable app release?


vanowm:
### Describe the problem

Run application without installing

### Describe the solution you'd like

Is there a way create an app that doesn't require installation and can be run from a removable storage on any computer (assuming the same OS). On top of that any settings stored in the same directory as the executable?

For now I'm only interested in Windows OS.

Archieeeeee:
Same question here, is it possible to use cargo build to release and it will be great if theres a guide to do so

lucasfernog:
You could just ship the `.exe` file if you don't use resources or sidecars. But the user will need to have the Webview runtime installed, since our installer is responsible for managing it.

vanowm:
And the settings?

lucasfernog:
The only settings stored are the Webview ones, which can be customized with the [data directory](https://docs.rs/tauri/1.0.0-beta.8/tauri/struct.WebviewAttributes.html#method.data_directory).

PR #14620, opened 2025-12-07
feat(core): allow config app dir override

### LINKED ISSUE #3020: Specify where the Window State plugin's JSON file is stored
### Problem
Currently, the Window State plugin saves the `.window-state.json` file in a fixed location: `${dataDir}/${bundleIdentifier}/`. This creates issues for applications that need to maintain full portable installations where all data stays in the app/binary directory.

### Proposed Solution
Add a `with_dir()` method to the `Builder` that allows specifying a custom directory for the window state file, similar to how `with_filename()` allows custom filenames.

**Suggested API:**
```rust
pub fn with_dir<P: AsRef<std::path::Path>>(self, dir: P) -> Self
```

**Example Usage:**
```rust
// Relative path
tauri_plugin_window_state::Builder::new()
    .with_state_flags(StateFlags::SIZE | StateFlags::POSITION)
    .with_dir("./config")
    .with_filename("my-window-state.json")
    .build()

// Absolute path
tauri_plugin_window_state::Builder::new()
    .with_state_flags(StateFlags::SIZE | StateFlags::POSITION)
    .with_dir("C:\\Users\\Username\\AppData\\Local\\MyApp")
    .with_filename("my-window-state.json")
    .build()
```

Legend-Master:
Reference: https://github.com/tauri-apps/tauri/issues/3109

Closes https://github.com/tauri-apps/plugins-workspace/issues/3020
We could still do https://github.com/tauri-apps/plugins-workspace/pull/3022 by changing `filename` to take a path (which can be relative)


crates/tauri/src/path/desktop.rs
```diff
@@ -276,6 +292,10 @@ impl<R: Runtime> PathResolver<R> {
   /// - **macOS:** Resolves to [`home_dir`](Self::home_dir)`/Library/Logs/${bundle_identifier}`
   /// - **Windows:** Resolves to [`local_data_dir`](Self::local_data_dir)`/${bundle_identifier}/logs`.
   pub fn app_log_dir(&self) -> Result<PathBuf> {
+    if let Some(app_directories_override) = self.app_directories_override() {
+      return Ok(app_directories_override?.join("logs"));
```
Legend-Master, line 296: Very much open to suggestion on those paths, and we could document them once settled

FabianLars:
This either needs to be windows only or have more warnings. Once you package the app on Linux and macOS most  paths won't work anymore while on Windows they'd still work (on perUser installs).

Speaking about paths, would it make sense to still resolve the `$DESKTOP` etc variables (except the APP ones of course) ?

For Linux another one like `$APPIMAGE` or simply resolving the path relative to the APPIMAGE env var instead of current_exe would help as well since appimages are in lines with the portable app idea here, the current_exe is just a temp read-only path

Legend-Master:
> Once you package the app on Linux and macOS most paths won't work anymore while on Windows they'd still work (on perUser installs).

So my original idea was to only use this in non-packaged environment like portable apps and during debugging (in `target` directory), but I wasn't sure what name to use here and just went with this, we could expand the documentations of course

> Speaking about paths, would it make sense to still resolve the $DESKTOP etc variables (except the APP ones of course) ?

Most of the paths still need to work I think, people might want to access those directories not just for storing their own data?

> For Linux another one like $APPIMAGE or simply resolving the path relative to the APPIMAGE env var instead of current_exe would help as well since appimages are in lines with the portable app idea here, the current_exe is just a temp read-only path

Nice call, I'll change it to use `current_binary` here instead

FabianLars:
> Most of the paths still need to work I think, people might want to access those directories not just for storing their own data?

I meant in the new path config, so you could say your override should be somewhere on the Desktop or in the Documents dir.

> Nice call, I'll change it to use current_binary here instead

totally forgot that's a thing x)

Legend-Master:
> I meant in the new path config, so you could say your override should be somewhere on the Desktop or in the Documents dir.

Ah I see what you mean now, good idea I guess

amrbashir:
I don't see much value in this tbh, and I see it as feature creep. the linked issue/PR should be handled there instead.

IMO this can bring more headache than it solves:
- if people want to use this for testing, I'd argue its better to use a dev-only identifier aka `com.example.app-dev` that way you are actually testing your app against the same prod directory structure
- if people want to place files next to their executable aka portable apps, users should just use the `current_binray() + filename` instead

Legend-Master:
> * if people want to use this for testing, I'd argue its better to use a dev-only identifier aka `com.example.app-dev` that way you are actually testing your app against the same prod directory structure

Yeah I get this, I mainly just want this so each time I download a new bug repro example, I could use this so I don't have to clean up a whole bunch of directories

> * if people want to place files next to their executable aka portable apps, users should just use the `current_binray() + filename` instead

This is a bit hard since you need to know and change every file storage settings to point to a custom location. If you also publish both a portable version and a non portable version, it adds in even more complexity

That being said, I do see your points. Maybe keep the PR around without merging see if people find it useful?

amrbashir:
> > * if people want to use this for testing, I'd argue its better to use a dev-only identifier aka `com.example.app-dev` that way you are actually testing your app against the same prod directory structure
> 
> Yeah I get this, I mainly just want this so each time I download a new bug repro example, I could use this so I don't have to clean up a whole bunch of directories

I feel you but I wouldn't consider this use case as a reason to introduce such API.

> > * if people want to place files next to their executable aka portable apps, users should just use the `current_binray() + filename` instead
> 
> This is a bit hard since you need to know and change every file storage settings to point to a custom location. If you also publish both a portable version and a non portable version, it adds in even more complexity

I don't understand where the complexity comes from, we have 4 or 5 `app_*_dir` APIs, users will just replace any usage of these APIs with their own functions that does:
1. check if portable and based on that
2. delegate to Tauri APIs + filename
3. or use `current_binary` + filename


> That being said, I do see your points. Maybe keep the PR around without merging see if people find it useful?

yeah totally

Legend-Master:
> users will just replace any usage of these APIs with their own functions that does:

The thing is, it's not just the app's code but also tauri and plugins. Which API needs to be replaced is not obvious sometimes and you need to catch all of them. Also monitor the changelogs in case there're new files being saved. And if a plugin doesn't provide an option like this, you'll either need to patch it or let it go

amrbashir:
Any plugin that uses a file for storage should expose an API to change it, because plugins can still choose not to use tauri path APIs and go with their own implementation or use `dirs` crate directly. 

I recently had this exact situation where I was authoring a plugin for a client that needed to download and save files before even running the tauri application and so I was not able to use tauri path APIs at all.

lucasfernog:
also useful if you're migrating from another framework to tauri - so you can reuse app paths (we've seen this problem before with some folks migrating their mobile apps to

lucasfernog (approved):

FabianLars:
2 votes against and still merged without any further discussion?

Issue #14626, opened 2025-12-08
[bug] tauri/wry leaking WebKitNetworkProcess processes


qu1ck:
### Describe the bug

When closing all windows of a tauri app WebKitNetworkProcess is left running. 
This is after closing and opening main window of my app 4 times:

<img width="350" height="220" alt="Image" src="https://github.com/user-attachments/assets/8431d5f5-605e-46ba-b553-955671cbe17e" />

### Reproduction

Have the app destroy() and recreate the window a few times.

### Expected behavior

Processes are cleaned up or at least reused when windows are closed/recreated.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Debian 13.0.0 x86_64 (X64) (KDE on wayland)
    ✔ webkit2gtk-4.1: 2.50.1
    ✔ rsvg2: 2.60.0
    ✔ rustc: 1.91.1 (ed61e7d7e 2025-11-07)
    ✔ cargo: 1.91.1 (ea2d97820 2025-10-10)
    ✔ rustup: 1.28.2 (e4f3ad6f8 2025-04-28)
    ✔ Rust toolchain: stable-x86_64-unknown-linux-gnu (default)
    - node: 24.11.1
    - npm: 11.6.2

[-] Packages
    - tauri 🦀: 2.9.1, (outdated, latest: 2.9.4)
    - tauri-build 🦀: 2.5.1, (outdated, latest: 2.5.3)
    - wry 🦀: 0.53.5
    - tao 🦀: 0.34.5
    - @tauri-apps/api  ⱼₛ: 2.9.0 (outdated, latest: 2.9.1)
    - @tauri-apps/cli  ⱼₛ: 2.9.1 (outdated, latest: 2.9.5)

[-] Plugins
    - tauri-plugin-notification 🦀: 2.3.1, (outdated, latest: 2.3.3)
    - @tauri-apps/plugin-notification  ⱼₛ: 2.3.1 (outdated, latest: 2.3.3)
    - tauri-plugin-shell 🦀: 2.3.1, (outdated, latest: 2.3.3)
    - @tauri-apps/plugin-shell  ⱼₛ: 2.3.1 (outdated, latest: 2.3.3)
    - tauri-plugin-clipboard-manager 🦀: 2.3.0, (outdated, latest: 2.3.2)
    - @tauri-apps/plugin-clipboard-manager  ⱼₛ: 2.3.0 (outdated, latest: 2.3.2)
[...]

PR #14628, opened 2025-12-08
fix(linux): reuse WebContext to prevent WebKitNetworkProcess leak

KushalMeghani1644:
This PR fixes the issue where `WebKitNetworkProcess` instances are left running after closing Tauri windows on Linux.

When a webview is destroyed on Linux, the `WebContext` is removed from the context store if no other webviews reference it. When a new window is created later, a new `WebContext` is allocated, which spawns a new `WebKitNetworkProcess`. WebKit does not terminate these network processes when the context is dropped, leading to process accumulation.

### Solution applied in this PR: 

Keep the `WebContext` alive on Linux/BSD platforms even when all referencing webviews are closed. This allows the context to be reused by future webviews, ensuring only a single `WebKitNetworkProcess` is spawned during the application's lifetime.

### Resolves

Issue: #14626


FabianLars:
> WebKit does not terminate these network processes when the context is dropped, leading to process accumulation.

I feel like fixing this would be the more appropriate but i wouldn't mind your approach that much if that's not possible, assuming that it's properly killed once the app actually exits.
I'm thinking about long-running tray apps here that only rarely open a window.

FabianLars:
And perhaps it should also be investigated a bit upstream: https://github.com/tauri-apps/wry/issues/590 though it may very well be possible that we just have to document that downstream (in this case tauri) has to handle this.

KushalMeghani1644:
Hey @FabianLars 
>I feel like fixing this would be the more appropriate but i wouldn't mind your approach that much if that's not possible, assuming that it's properly killed once the app actually exits.
I'm thinking about long-running tray apps here that only rarely open a window.

for your concern of the tray applications I have added docs in the code that makes the behavior clear that the process will still terminate when the app exits since on Linux that's the expected behavior whenever an app closes!

>And perhaps it should also be investigated a bit upstream: https://github.com/tauri-apps/wry/issues/590 though it may very well be possible that we just have to document that downstream (in this case tauri) has to handle this.

Upon investigating the issue you mentioned I feel like the patch belongs to this repo only!

FabianLars:
> for your concern of the tray applications I have added docs in the code that makes the behavior clear that the process will still terminate when the app exits since on Linux that's the expected behavior whenever an app closes!

i was thinking that this would be kinda wasteful to keep processes open that aren't used. Like the rust process is like 5mb and we keep a 25mb process around for perhaps no reason? doesn't sound in line with tauri's philosophy to me.

KushalMeghani1644:
>i was thinking that this would be kinda wasteful to keep processes open that aren't used. Like the rust process is like 5mb and we keep a 25mb process around for perhaps no reason? doesn't sound in line with tauri's philosophy to me.

I understand your concerns @FabianLars but my idea or approach is that if we keep the process alive and the window closes there would be no further needs to restart that process again! which could speed up things! and as far as I know not many people would care about extra ~25Mb being used in their RAM as long as it's useful + gives a advantage! And if that doesn't sound good we can add a configurable option so the users can go to the tauri-conf.json and set true if they want the process alive or killed after the window is closed. What do you think?

FabianLars:
> and as far as I know not many people would care about extra ~25Mb being used in their RAM as long as it's useful + gives a advantage!

More efficient ram usage is one of the major reasons people switch to tauri and there's really not much of an advantage here as re-starting the network process is faster than re-starting the other processes so this should be done before you even notice it.

> And if that doesn't sound good we can add a configurable option so the users can go to the tauri-conf.json and set true if they want the process alive or killed after the window is closed. What do you think?

For that we first have to fix the issue that prevents it from being killed :D I'd be okay with merging your change if it means we will investigate the actual issue and switch to the fix then. == We'd take your PR as a temporary workaround to get rid of the tens of zombie processes for now.

KushalMeghani1644:
Thanks for the feedback @FabianLars I get it that keeping WebKitNetworkProcess alive is a minor memory overhead and that restarting it would be fast enough in most cases. So if you are fine with it... you can happily merge my PR as a temp work around and I will work on fixing this bug and adding the configurable option if you are fine with it :)

FabianLars:
i'd very much appreciate it if you could look into the root issue here and fix it! Don't bother with a config though, that really isn't worth it, we don't keep any webview processes running on any platform so there's no reason to allow devs to do that for this single process :)

KushalMeghani1644:
Thanks! @FabianLars I will investigate the root issue here and work on fixing this ASAP :D

KushalMeghani1644:
Umm... hey! @FabianLars I did a bit of thorough investigation, and the results are quite interesting!
What I found?

WebKit itself is responsible, not Tauri :O

The root cause is in WebKit's architecture:

- WebKit spawns one WebKitNetworkProcess per WebsiteDataManager (which is wrapped by WebContext in webkit2gtk). This is by design, the NetworkProcess handles cookies, caching, and networking for all WebViews sharing that context.

- WebKit does NOT terminate the NetworkProcess when the WebContext is dropped this is intentional behavior in WebKit's multi-process model. The NetworkProcess is designed to persist because:
       - It holds persistent data (cookies, cache)
       - Restarting it is expensive
       - It's supposed to be shared across all WebViews
       - The bug is that each new WebContext spawns its own NetworkProcess, and when that WebContext is destroyed, the old NetworkProcess becomes orphaned but WebKit doesn't clean it up.

What to do?

The proper fix would be in WebKit upstream to either:

- Terminate NetworkProcess when all associated WebContexts are destroyed
- Share a single NetworkProcess across all WebContexts

So seems like my work-around works xD now my confusion is that if this is a WebKit issue on linux itself... I don't think there's much we can do, Maybe go and fix WebKit? xD IDK if you already knew this or not but yeah this is all I was able to find really.

FabianLars:
This is so stupid but it's also such a webkit thing to do. Really makes no sense to me that it behaves that way _with no way to opt-out_

Thanks for the investigation nonetheless!

KushalMeghani1644:
Glad it helped! @FabianLars, anyways... curious about what to do next for this issue?

FabianLars:
hmm i guess nothing to do for you anymore. We need to test it ourselves now but i know that we're all insanely busy at work right now. Since i'm in the car with a hotspot for now maybe i can do it quickly now, can't actually work here anyway.

FabianLars:
ah and we need a changefile similar to https://github.com/tauri-apps/tauri/pull/14620/changes#diff-4d219b4237dee6e7ef1d053a6c0ea396331048f5fcc8eb105d4b9813ee9e5e6b (but `tauri-runtime-wry` and `patch`)

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -2421,7 +2421,16 @@ impl Drop for WebviewWrapper {
 
       if let Some(web_context) = context_store.get_mut(&self.context_key) {
         web_context.referenced_by_webviews.remove(&self.label);
-
+        // On Linux/BSD, we keep the WebContext alive even when all windows are closed.
+        // This ensures only one WebKitNetworkProcess is spawned for the lifetime of the app.
+        // For tray-only or long-running apps, the process will still terminate when the application exits.
```
FabianLars:
```suggestion
		// https://github.com/tauri-apps/tauri/issues/14626
        // Because WebKit does not close its network process even when no webviews are running, we need to ensure to re-use the existing process on Linux by keeping the WebContext alive for the lifetime of the app.
        // WebKit on macOS handles this itself.
```
formatting will be off but whatever

FabianLars:
seems to work well in my quick test, thanks

KushalMeghani1644:
> seems to work well in my quick test, thanks

Thanks @FabianLars I have applied your suggestion with a clear formatting... if there are other changes please let-me know I will apply them ASAP :)

FabianLars:
Thanks! I don't expect any changes but I'm away over the weekend (already am) so won't be able to merge until Monday earliest :)

KushalMeghani1644:
>Thanks! I don't expect any changes but I'm away over the weekend (already am) so won't be able to merge until Monday earliest :)

No problems! @FabianLars enjoy, take your time, and have fun!

KushalMeghani1644:
Sorry to ping @FabianLars if you were busy... are we ready to merge these changes? or are there more changes to make? happy to make them if so!

FabianLars:
nono you're good. the last work week was just busy and now it's christmas. i have a free second rn though

FabianLars (approved):

KushalMeghani1644:
Thanks for putting your time! And sorry to disturb you! Merry christmas! :D

PR #14632, opened 2025-12-08
Cleanups

sftse:
Got sidetracked, best read commit by commit.

### DIFF
--- a/.changes/change-pr-14632.md
+++ b/.changes/change-pr-14632.md
@@ -1,0 +1,7 @@
+---
+"tauri-utils": patch:enhance
+"tauri-build": patch:enhance
+"tauri-cli": patch:enhance
+---
+
+Small code refactors for improved code readability
--- a/.changes/change-pr-14632.md
+++ b/.changes/change-pr-14632.md
@@ -4,4 +4,4 @@
 "tauri-cli": patch:enhance
 ---
 
-Small code refactors for improved code readability
+Small code refactors for improved code readability. No user facing changes.
--- a/crates/tauri-cli/src/build.rs
+++ b/crates/tauri-cli/src/build.rs
@@ -144,7 +144,7 @@ pub fn command(mut options: Options, verbosity: u8) -> Result<()> {
 pub fn setup(
   interface: &AppInterface,
   options: &mut Options,
-  config_: &ConfigMetadata,
+  config: &ConfigMetadata,
   mobile: bool,
 ) -> Result<()> {
   let tauri_path = tauri_dir();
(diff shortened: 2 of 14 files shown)


crates/tauri-cli/src/dev.rs
```diff
@@ -216,14 +216,13 @@ pub fn setup(interface: &AppInterface, options: &mut Options, config: ConfigHand
           let status = child_
             .wait()
             .expect("failed to wait on \"beforeDevCommand\"");
-          if !(status.success() || KILL_BEFORE_DEV_FLAG.get().unwrap().load(Ordering::Relaxed)) {
+          if !(status.success() || KILL_BEFORE_DEV_FLAG.load(Ordering::Relaxed)) {
```
sftse, line 221: Although implausible, there's nothing stopping the child to exit before the flag gets initialized. Because the thread is spawned, this panic might be hard to detect.

Legend-Master:
I am a little concerned about https://github.com/tauri-apps/tauri/pull/14632/commits/d97b190268699e7945c1d9263ba855a63dc17fe4 that it changed the timing on when we resolve the tauri directories which might get affected by the working directory (we do change them later on in some places)

sftse:
Removed the commit until can split it into changes more obviously correct.
If the globals can be removed think that would be quite an improvement.

RFR

sftse:
Fixed CI

crates/tauri-cli/src/build.rs
```diff
@@ -144,7 +144,7 @@ pub fn command(mut options: Options, verbosity: u8) -> Result<()> {
 pub fn setup(
   interface: &AppInterface,
   options: &mut Options,
-  config: ConfigHandle,
+  config_: &ConfigMetadata,
```
Legend-Master:
```suggestion
  config: &ConfigMetadata,
```

Maybe?

Legend-Master (dismissed):
Looking good, thanks!

And you need to sign your commits 🙃

sftse:
GTM

.changes/change-pr-14632.md
```diff
@@ -0,0 +1,7 @@
+---
+"tauri-utils": patch:enhance
+"tauri-build": patch:enhance
+"tauri-cli": patch:enhance
+---
+
+Small code refactors for improved code readability
```
FabianLars:
```suggestion
Small code refactors for improved code readability. No user facing changes.
```
maybe just me but i like to mention this

sftse: Isn't this implied by "patch"?

FabianLars: hmm, i guess so. i often just add it for refactors that for example don't fix bugs some users had either. so no user facing changes in API and behavior.

edit: i guess that comes from the experience that you can't rely on something to be implied, i assume that's the heavy language barriers we're dealing with here. so this way i'm just trying to make sure the "Small code refators for improved code readability." text cannot be misinterpreted at all.

but like i said, i think that's just me here 😂

FabianLars (approved):

Legend-Master (approved):

PR #14633, opened 2025-12-08
fix(tauri-cli, tauri-build): copy external binaries only during bundling

### LINKED ISSUE #14602: [bug] Resource paths available only after full builds are checked in non-builds
### Describe the bug

Some resource paths in `tauri.conf.json` seem to be checked in every build, even the resource paths like additional binaries that are only known after a command like `cargo build --release --bins`.

This `tauri.conf.json` currently does not permit running `cargo clippy` or `cargo check`. 

```text
{
  "$schema": "https://schema.tauri.app/config/2.0.0-rc",
  "productName": "project",
  "version": "0.1.0",
  "identifier": "project.app",
  "build": {
    "beforeDevCommand": "npm run dev",
    "devUrl": "http://localhost:1420",
    "beforeBuildCommand": "npm run build",
    "frontendDist": "../build"
  },
  "app": {
    "windows": [
      {
        "title": "project",
        "width": 800,
        "height": 600
      }
    ],
    "security": {
      "csp": null,
      "pattern": {
        "use": "brownfield"
      }
    }
  },
  "bundle": {
    "active": true,
    "targets": "all",
    "icon": [
      "icons/32x32.png",
      "icons/128x128.png",
      "icons/128x128@2x.png",
      "icons/icon.icns",
      "icons/icon.ico"
    ],
    "externalBin": ["../../target/release/process"]
  }
}
```

```text
error: failed to run custom build command for `project v0.1.0 (project/project/src-tauri)`

Caused by:
  process didn't exit successfully: `project/target/debug/build/project-e52c3da1635962f4/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-changed=src/tauri.conf.json
  cargo:rerun-if-env-changed=TAURI_CONFIG
  cargo:rustc-check-cfg=cfg(desktop)
  cargo:rustc-cfg=desktop
  cargo:rustc-check-cfg=cfg(mobile)
  cargo:rerun-if-changed=project/project/src-tauri/tauri.conf.json
  cargo:rustc-env=TAURI_ANDROID_PACKAGE_NAME_APP_NAME=app
  cargo:rustc-env=TAURI_ANDROID_PACKAGE_NAME_PREFIX=<project>
  cargo:rustc-check-cfg=cfg(dev)
  cargo:rustc-cfg=dev
  cargo:PERMISSION_FILES_PATH=project/target/debug/build/project-8d94d68162fcccc9/out/app-manifest/__app__-permission-files
  cargo:rerun-if-changed=capabilities
  cargo:rerun-if-env-changed=REMOVE_UNUSED_COMMANDS
  cargo:rustc-env=TAURI_ENV_TARGET_TRIPLE=x86_64-unknown-linux-gnu
  resource path `../../target/release/process-x86_64-unknown-linux-gnu` doesn't exist

```

tauri-build is at 2.5.2

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
[✔] Environment
    - OS: Artix Linux Rolling Release x86_64 (X64) (Unknown DE on tty)
    ✔ webkit2gtk-4.1: 2.50.1
    ✔ rsvg2: 2.61.2
    ✔ rustc: 1.89.0-nightly (255aa2208 2025-06-19)
    ✔ cargo: 1.89.0-nightly (2251525ae 2025-06-16)
    ✔ Rust toolchain: nightly-x86_64-unknown-linux-gnu (default)
    - node: 24.9.0
    - npm: 11.6.2

[-] Packages
    - tauri 🦀: 2.9.2, (outdated, latest: 2.9.4)
    - tauri-build 🦀: 2.5.2, (outdated, latest: 2.5.3)
    - wry 🦀: 0.53.5
    - tao 🦀: 0.34.5
    - @tauri-apps/api  ⱼₛ: 2.9.0 (outdated, latest: 2.9.1)
    - @tauri-apps/cli  ⱼₛ: 2.9.4 (outdated, latest: 2.9.5)

[-] Plugins
    - tauri-plugin-shell 🦀: 2.3.3
    - @tauri-apps/plugin-shell  ⱼₛ: 2.3.3
    - tauri-plugin-dialog 🦀: 2.4.2
    - @tauri-apps/plugin-dialog  ⱼₛ: 2.4.2
    - tauri-plugin-fs 🦀: 2.4.4
    - @tauri-apps/plugin-fs  ⱼₛ: 2.4.4

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../build
    - framework: Svelte
    - bundler: Vite
```

### Stack trace

```text

```

### Additional context

_No response_

sftse:
Should fix #14602 

I'm a bit unsure about moving things from the build script to the `tauri-cli`, since parts of the bundler were placed deliberately in the build script.

If possible, moving as much as possible from the build script into the `tauri-cli` could improve code comprehension.

### DIFF
--- a/crates/tauri-build/src/lib.rs
+++ b/crates/tauri-build/src/lib.rs
@@ -16,7 +16,7 @@ use cargo_toml::Manifest;
 
 use tauri_utils::{
   config::{BundleResources, Config, WebviewInstallMode},
-  resources::{external_binaries, ResourcePaths},
+  resources::ResourcePaths,
 };
 
 use std::{
@@ -50,37 +50,6 @@ fn copy_file(from: impl AsRef<Path>, to: impl AsRef<Path>) -> Result<()> {
   let dest_dir = to.parent().expect("No data in parent");
   fs::create_dir_all(dest_dir)?;
   fs::copy(from, to)?;
-  Ok(())
-}
-
-fn copy_binaries(
-  binaries: ResourcePaths,
-  target_triple: &str,
-  path: &Path,
-  package_name: Option<&String>,
-) -> Result<()> {
-  for src in binaries {
-    let src = src?;
-    println!("cargo:rerun-if-changed={}", src.display());
-    let file_name = src
-      .file_name()
-      .expect("failed to extract external binary filename")
-      .to_string_lossy()
-      .replace(&format!("-{target_triple}"), "");
-
-    if package_name == Some(&file_name) {
-      return Err(anyhow::anyhow!(
-        "Cannot define a sidecar with the same name as the Cargo package name `{}`. Please change the sidecar name in the filesystem and the Tauri configuration.",
-        file_name
-      ));
-    }
-
-    let dest = path.join(file_name);
-    if dest.exists() {
-      fs::remove_file(&dest).unwrap();
-    }
-    copy_file(&src, &dest)?;
-  }
   Ok(())
 }
 
@@ -527,15 +496,6 @@ pub fn try_build(attributes: Attributes) -> Result<()> {
     .unwrap()
     .parent()
     .unwrap();
-
-  if let Some(paths) = &config.bundle.external_bin {
-    copy_binaries(
-      ResourcePaths::new(&external_binaries(paths, &target_triple, &target), true),
-      &target_triple,
-      target_dir,
-      manifest.package.as_ref().map(|p| &p.name),
-    )?;
-  }
 
   #[allow(unused_mut, clippy::redundant_clone)]
   let mut resources = config
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8627,6 +8627,7 @@ dependencies = [
  "axum",
  "base64 0.22.1",
  "cargo-mobile2",
+ "cargo_toml",
  "clap",
  "clap_complete",
  "colored",
--- a/crates/tauri-cli/Cargo.toml
+++ b/crates/tauri-cli/Cargo.toml
@@ -114,6 +114,7 @@ rand = "0.9"
 zip = { version = "4", default-features = false, features = ["deflate"] }
 which = "8"
 rayon = "1.10"
+cargo_toml = "0.22.3"
 
 [dev-dependencies]
 insta = "1"
--- a/crates/tauri-cli/src/bundle.rs
+++ b/crates/tauri-cli/src/bundle.rs
@@ -10,10 +10,12 @@ use std::{
 
 use clap::{builder::PossibleValue, ArgAction, Parser, ValueEnum};
 use tauri_bundler::PackageType;
+use tauri_utils::platform::target_triple;
 use tauri_utils::platform::Target;
+use tauri_utils::resources::{external_binaries, ResourcePaths};
 
 use crate::{
-  error::{Context, ErrorExt},
+  error::{bail, Context, ErrorExt},
   helpers::{
     self,
     app_paths::tauri_dir,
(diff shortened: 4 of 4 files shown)


sftse:
It's not a particularly clean fix, just moving enough out of `tauri-build` and duplicating some code into `tauri-cli`.

crates/tauri-build/src/lib.rs
```diff
@@ -53,37 +53,6 @@ fn copy_file(from: impl AsRef<Path>, to: impl AsRef<Path>) -> Result<()> {
...
 
-fn copy_binaries(
-  binaries: ResourcePaths,
-  target_triple: &str,
-  path: &Path,
-  package_name: Option<&String>,
-) -> Result<()> {
-  for src in binaries {
-    let src = src?;
-    println!("cargo:rerun-if-changed={}", src.display());
```
sftse, line 64: This line was necessary because `copy_binaries` was in the build script. Moving it to the bundler will always refresh the binaries before bundling, which is fine unless they take a long time to copy.

FabianLars:
> If possible, moving as much as possible from the build script into the tauri-cli could improve code comprehension.

The build script does so much shit so that users can use eg `cargo run` directly so there's only so much we can remove without breaking that.

sftse:
Is it currently possible to build a fully functional Tauri app without calling `tauri build` or `tauri bundle`?

sftse:
I've tried before and simply running `cargo run --bin <app>` failed

sftse:
Ok, so the helloworld example does run with just cargo, never mind then.. Must be something else preventing my other app from running.

sftse:
This fix has been working for me so far, I guess this would hinge on whether it is expected that bundling only works with tauri-cli.

The only way to break even less workflows would be to copy the binaries as best possible, ignoring errors in tauri-build and then copy again with strict error checking in tauri-cli, but this seems even less pretty.

Thoughts?

Legend-Master:
Does this change break `tauri dev`/`cargo run` for cases where you do have the `externalBin` available (it seems like it)? I think the build script used to provide you the `externalBin` files during development without running `tauri bundle` right?

sftse:
That is a valid point and a use case which would be broken through this change. We could as said opt to be more conservative and ignore some types of errors (NotFound) when called from the build script.

If somebody is relying on wrong paths in externalBins being found at build time, he would then receive a more obscure error when running `tauri dev`.

Legend-Master:
This is quite tricky, I'm not sure if ignoring error here is a good choice here either since you wouldn't notice the error until much later when your app or other things call it

sftse:
I understand this is technically breaking, but to play devil's advocate, the externalBin configuration is under `bundle`, as if it was never intended to be checked at build time. The directory we're copying into is the cargo target directory, which would be unusual to try call binaries from.

Somebody doing this would need to distinguish whether their app is built in debug or release mode to reliably call the right binaries, is that possible?

FabianLars:
i don't care about `cargo build` but `cargo run` must keep working. still, tauri is in a position where even these kind of breaking changes should be made carefully (including to consider waiting for v3). we're not some small hobby project and we're officially following semver.

Legend-Master:
I am not sure if this is a good idea even without the breaking change, to me, I would expect the whole thing to just work when I run `tauri dev` or `cargo run` (`cargo build` and then running the binary manually in extend), even though we did a lot of things that were probably not intended to be done in build scripts, they indeed reduced friction (I really hated the experience having to use `mach` in servo, they basically semi-followed the build script guide lines and put those things in their CLI but I honestly hate it more than what we did)

About the fact that those settings are under `bundle`, it sure is kind of confusing but I don't know if there's a better name / put it in another field

PR #14648, opened 2025-12-12
fix(cef/macos): fix TitleBarStyle::Visible have full size content view

amrbashir:


lucasfernog:
i think i just copied it from tao :D

PR #14653, opened 2025-12-14
fix: default `WindowConfig::focus` to false in `Default::default`

Legend-Master:
Not exactly sure if this change will be breaking... We're technically fixing it but maybe someone's relying on it


FabianLars:
I have a hard time believing anyone relying on that 🤔

FabianLars (approved):

PR #14664, opened 2025-12-17
perf: optimize memory allocation for TaoIcon handling

Tunglies:


crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -486,7 +487,12 @@ pub struct TaoIcon(pub TaoWindowIcon);
 impl TryFrom<Icon<'_>> for TaoIcon {
   type Error = Error;
   fn try_from(icon: Icon<'_>) -> std::result::Result<Self, Self::Error> {
-    TaoWindowIcon::from_rgba(icon.rgba.to_vec(), icon.width, icon.height)
+    let rgba = match icon.rgba {
+      Cow::Borrowed(bytes) => bytes.to_vec(),
+      Cow::Owned(bytes) => bytes,
+    };
```
FabianLars, line 493: is Rust not smart enough to optimize this itself? :(

Tunglies:
```rust
pub fn to_vec(&self) -> Vec<T>
    where
        T: Clone,
    {
        self.to_vec_in(Global)
    }

pub fn to_vec_in<A: Allocator>(&self, alloc: A) -> Vec<T, A>
    where
        T: Clone,
    {
        hack::to_vec(self, alloc)
    }
```

Seemed call `.to_vec`, it will allocate everytime.

Legend-Master: Does something like to_owned work here?

Tunglies: Nope. I missunderstanding at previous, I find it `Icon::from_rgba` needed Vec<u8> now. Sry for borthering guys.

Tunglies: I will close this PR later,  incase if there is more sharing from you :)

Legend-Master: I think `into_owned` should work here right? (I was out 4 hours ago and I meant `into_owned` but can't remember the name)

PR #14666, opened 2025-12-17
perf: change EmitArgs to use RawValue for zero-copy payload

Tunglies:


Legend-Master (commented):
Hmm, do you mind explaining a bit about what zero copy this change made? `to_raw_value` uses `serde_json::to_string` internally as well, no?

Tunglies:
I see the source code now. This is invalid PR. My hands writes while brain not XD

PR #14667, opened 2025-12-17
perf: replace Arc<Mutex> with DashMap in ChannelDataIpcQueue for improved performance

Tunglies:
Introducing Dashmap dependency. Test with example, trigge the spawn button on UI.

Some testing usage:

```rust
let start_insert_time = std::time::Instant::now();
            webview
              .state::<ChannelDataIpcQueue>()
              .0
              .lock()
              .unwrap()
              .insert(data_id, body);
            let elapsed_insert = start_insert_time.elapsed();
              println!(
                "[tauri::ipc::channel] inserted channel data id {data_id} in {:?}",
                elapsed_insert
              );
```

```rust
// The repo's example cmd.rs
#[command]
pub fn spam(channel: Channel<Vec<u8>>) -> tauri::Result<()> {
  // make greater than MAX_RAW_DIRECT_EXECUTE_THRESHOLD(1024)
  let size = 50_000usize;
  let data = vec![0u8; size];
  channel.send(data)?;
  Ok(())
}
```

Previous Arc<Mutex<HashMap>>
```
[tauri::ipc::channel] inserted channel data id 0 in 34.334µs
```

Current Dashmap
```
[tauri::ipc::channel] inserted channel data id 0 in 15.958µs
```

Roughly improved 53% cost time.

### DIFF
--- a/.changes/improve-performance-when-payload-oversize-MAX_RAW_DIRECT_EXECUTE_THRESHOLD.md
+++ b/.changes/improve-performance-when-payload-oversize-MAX_RAW_DIRECT_EXECUTE_THRESHOLD.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch:minor
+---
+
+Introducing Dashmap instead of Arc<Mutex> in ChannelDataIpcQueue, improve performance when payload larger than MAX_RAW_DIRECT_EXECUTE_THRESHOLD. No user facing changes.
--- a/.changes/improve-performance-when-payload-oversize-MAX_RAW_DIRECT_EXECUTE_THRESHOLD.md
+++ b/.changes/improve-performance-when-payload-oversize-MAX_RAW_DIRECT_EXECUTE_THRESHOLD.md
@@ -1,5 +1,5 @@
 ---
-"tauri": patch:minor
+"tauri": patch:perf
 ---
 
 Introducing Dashmap instead of Arc<Mutex> in ChannelDataIpcQueue, improve performance when payload larger than MAX_RAW_DIRECT_EXECUTE_THRESHOLD. No user facing changes.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8499,6 +8499,7 @@ dependencies = [
  "bytes",
  "cargo_toml",
  "cookie",
+ "dashmap",
  "data-url",
  "dirs 6.0.0",
  "dunce",
--- a/crates/tauri/Cargo.toml
+++ b/crates/tauri/Cargo.toml
@@ -83,6 +83,7 @@ specta = { version = "^2.0.0-rc.16", optional = true, default-features = false, 
 ] }
 # WARNING: cookie::Cookie is re-exported so bumping this is a breaking change, documented to be done as a minor bump
 cookie = "0.18"
+dashmap = "6.1.0"
 
 # desktop
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd", target_os = "netbsd", target_os = "windows", target_os = "macos"))'.dependencies]
--- a/crates/tauri/src/ipc/channel.rs
+++ b/crates/tauri/src/ipc/channel.rs
@@ -3,14 +3,14 @@
 // SPDX-License-Identifier: MIT
 
 use std::{
-  collections::HashMap,
   str::FromStr,
   sync::{
     atomic::{AtomicU32, AtomicUsize, Ordering},
-    Arc, Mutex,
+    Arc,
   },
 };
 
+use dashmap::DashMap;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 use crate::{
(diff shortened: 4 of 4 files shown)


.changes/improve-performance-when-payload-oversize-MAX_RAW_DIRECT_EXECUTE_THRESHOLD.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri": patch:minor
```
Tunglies: Since introducing a new dependency, should go for a patch or a minor?

FabianLars: patch is fine

sftse:
Is there a more macroscopic example how this benefits performance? I'm in favor but for other reasons, already a transitive dependency and gets rid of unwraps for data structures that don't benefit from poisoning.

Legend-Master:
I am not sure how much this would benefit performance to be honest, with the test code you provided, my results are like this in release build (size optimized not speed)

> [tauri::ipc::channel] inserted channel data id 0 in 1.1µs
> [tauri::ipc::channel] inserted channel data id 1 in 700ns

And the whole fetch process takes milliseconds, I'm not sure if it makes sense to bring in a new dependency just for this

> already a transitive dependency

This is in the `tauri-cli`'s dependency tree, but in not `tauri`'s

sftse:
In that case no longer in favor.

On another note, removing the Arc here is almost certainly going to change behavior unless the derived `Clone` is never called.

Tunglies:
> Clone is never called.

Yes, there is no any clone with ChannelDataIpcQueue.

> macroscopic example & how much this would benefit performance

In the [clash-verge-rev](https://github.com/clash-verge-rev/clash-verge-rev) project, the optimization reduced the average IPC channel insertion time from approximately 2.36µs to 1.15µs, achieving a significant performance gain of over 50%.

> bring in a new dependency

Since Dashmap already in the lock file, bringing it as tauri's dependency did not need compile new units.

Legend-Master:
> In the [clash-verge-rev](https://github.com/clash-verge-rev/clash-verge-rev) project, the optimization reduced the average IPC channel insertion time from approximately 2.36µs to 1.15µs, achieving a significant performance gain of over 50%.

Could you share some benchmark results on the actual time used for a whole send operation for the channel? I think this insert call is most definitely not a bottleneck

> Since Dashmap already in the lock file, bringing it as tauri's dependency did not need compile new units.

As I said, it is only in a dependency of `tauri-cli`, if we add it to `tauri`, we need to consider the trade offs around the extra compile time and final binary size

Tunglies:
Compile Config:
```
[profile.release]
panic = "abort"
codegen-units = 1
lto = "thin"
opt-level = 3
debug = false
strip = true
overflow-checks = false
rpath = false
```

Compile Command:
```
// Equal to cargo build --release --timings
pnpm build -- --timings
```

Baseline:
```
Compile Units: 1008
Compile Time: 6m14s
Binary Size: 27272512
```

With Dashmap as dependency:
```
Compile Units: 1011
Compile Time: 6m33s
Binary Size: 27272528
```
---

> some benchmark results on the actual time used for a whole send operation for the channel

This PR forcusing on reduce IPC communication lantancy when inserting data when IPC payload over MAX_RAW_DIRECT_EXECUTE_THRESHOLD. 

I would like to share some of actually runtime measurement with same compile config. I measure with launch to main dashboard, and do no extra user-operations, but cancle run mannualy. The order may not exactually same since async.
[insert_lantancy_baseline.txt](https://github.com/Tunglies/sharning/blob/main/insert_lantancy_baseline.txt)
[insert_lantancy_with_dashmap.txt](https://github.com/Tunglies/sharning/blob/main/insert_lantancy_with_dashmap.txt)

> need to consider the trade offs

Sorry to bother you, I'm used to developing user applications and may lack experience in considering as upstream aspects.

Legend-Master:
> I would like to share some of actually runtime measurement with same compile config. I measure with launch to main dashboard, and do no extra user-operations, but cancle run mannualy. The order may not exactually same since async.
[insert_lantancy_baseline.txt](https://github.com/Tunglies/sharning/blob/main/insert_lantancy_baseline.txt)
[insert_lantancy_with_dashmap.txt](https://github.com/Tunglies/sharning/blob/main/insert_lantancy_with_dashmap.txt)

The data seems to suggest the 2 took very similar times, no? With `insert_lantancy_with_dashmap.txt` having some more outliers that took milliseconds

> Sorry to bother you, I'm used to developing user applications and may lack experience in considering as upstream aspects.

No worries, we all learned our way through 🙃

Tunglies:
Tauri already has MAX_RAW_DIRECT_EXECUTE_THRESHOLD(1024) to graceful data offloading which rarely triggered in normal use. The Dashmap makes inner split locks and cause memory footprint heavy and do not helps in general usecase. Could be better if lockless optimize.

PR #14671, opened 2025-12-18
feat(bundler): support Liquid Glass icons, closes #14207

### LINKED ISSUE #14207: [feat] Support Liquid Glass Icons for macOS 26
### Describe the problem

Apple introduced a new way to create icons using [Icon Composer](https://developer.apple.com/icon-composer/). This is not a standard image format and has assets & layers bundled inside it.

However, this app creates .icon files which does not currently work with tauri icon.

You can use .icon files as parameters for the tauri icon command, or set .icon files as icons for iOS and macOS.

### Alternatives considered

_No response_

### Additional context

_No response_

### LINKED ISSUE #14979: [feat] Support MacOS 26 Icons (Dark Mode & Clear)
### Describe the problem

It shows the normal colored icon and it makes it stand out a lot more when the other icons are all tinted or clear and just the Tauri app isn't?

I want it to show the icon as a clear or tinted one and not the normal colored icon. I don't know if this is an issue that also occurs on the latest version of iOS? I haven't tested

### Alternatives considered

_No response_

### Additional context

_No response_

lucasfernog:
the `icon` config now supports loading an Assets.car directly or a `.icon` (Icon Composer asset) that gets compiled into an Assets.car file

here's a dramatic demo:

https://github.com/user-attachments/assets/ff57a1e0-feaf-4354-9990-661296f9d0a6


doroved:
Will this appear in 2.9.6? :)

crates/tauri-bundler/src/bundle/macos/app.rs
```diff
@@ -362,6 +360,27 @@ fn create_info_plist(
...
+        .to_string_lossy()
+        .into_owned()
+        .into(),
+    );
+  }
+
+  if let Some(assets_car_file) = assets_car_file {
+    if let Some(icon_name) = app_icon_name_from_assets_car(&assets_car_file) {
+      plist.insert("CFBundleIconName".into(), icon_name.clone().into());
+      plist.insert("CFBundleIconFile".into(), icon_name.into());
```
FabianLars: Do we set the name for the .Icon / Assets.car thingy? Just concerned about breaking the .icns fallback here

UsamaKhan:
Is this still planned/being worked on? @lucasfernog

FabianLars:
planned for 2.11

crates/tauri-bundler/src/bundle/macos/app.rs
lucasfernog-crabnebula: I've pushed a fix for it, looks like that's the appropriate way to manage this

CFBundleIconName for the CAR, CFBundleIconFile for the ICNS
i tested this by deleting the Assets.car file, running `touch my.app` and rerunning it, it picks up the icns then

alternatively we could make sure the file names always match (which means renaming the icns file to AppIcon.icns), that's what the Icon Composer app does anyway

FabianLars (approved):

PR #14700, opened 2025-12-26
fix: BSD support in tauri-runtime

mewi99:
Hello. The most important change in this PR is the addition of the *bsd configuration options to the cfg attribute on top of Runtime's new_any_thread() method signature. 
This ensures that implementing Runtime for Wry in tauri-runtime-wry doesn't cause a `function not a member of trait Runtime` error.

I also added the same options to places where, due to the practical similarities with Linux, it makes sense to have them and not treat the BSDs like they are entirely different platforms.

This does not mean full Tauri support on the BSDs, but after these changes and a few patches to a couple of dependecies, I was able to run the helloworld example successfully on FreeBSD, and I was able to compile another project that uses Tauri.

### DIFF
--- a/.changes/runtime-bsd.md
+++ b/.changes/runtime-bsd.md
@@ -1,0 +1,6 @@
+---
+tauri-runtime: patch:bug
+tauri-runtime-wry: patch:bug
+---
+
+Fix compilation errors when targeting BSD.
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -37,7 +37,13 @@ use tauri_runtime::{
 use objc2::rc::Retained;
 #[cfg(target_os = "macos")]
 use tao::platform::macos::{EventLoopWindowTargetExtMacOS, WindowBuilderExtMacOS};
-#[cfg(target_os = "linux")]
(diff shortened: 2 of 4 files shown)


meow2149:
> Hello. The most important change in this PR is the addition of the *bsd configuration options to the cfg attribute on top of Runtime's new_any_thread() method signature. This ensures that implementing Runtime for Wry in tauri-runtime-wry doesn't cause a `function not a member of trait Runtime` error.
> 
> I also added the same options to places where, due to the practical similarities with Linux, it makes sense to have them and not treat the BSDs like they are entirely different platforms.
> 
> This does not mean full Tauri support on the BSDs, but after these changes and a few patches to a couple of dependecies, I was able to run the helloworld example successfully on FreeBSD, and I was able to compile another project that uses Tauri.

I really enjoyed reading this PR — the BSD changes and your explanations were thoughtful, clear, and honestly a pleasure to go through.

It made me curious about the person behind the code, so I wanted to say hi 🙂  
If you’re open to it, I’d like to follow each other on GitHub and stay in touch. I have a feeling I’d enjoy seeing more of what you work on.

FabianLars (approved):
thanks for the pr :)

PR #14702, opened 2025-12-27
docs: pixel units

Legend-Master:


FabianLars (approved):
(didn't really double check tbh)

PR #14708, opened 2025-12-29
refactor: use u64 instead of usize for nonce gen

Legend-Master:
`nonce` string length shouldn't be related to platform pointer size really

### DIFF
--- a/crates/tauri/src/manager/mod.rs
+++ b/crates/tauri/src/manager/mod.rs
@@ -129,14 +129,7 @@ fn replace_csp_nonce(
 ) {
   let mut nonces = Vec::new();
   *asset = replace_with_callback(asset, token, || {
-    #[cfg(target_pointer_width = "64")]
-    let mut raw = [0u8; 8];
-    #[cfg(target_pointer_width = "32")]
-    let mut raw = [0u8; 4];
-    #[cfg(target_pointer_width = "16")]
-    let mut raw = [0u8; 2];
-    getrandom::fill(&mut raw).expect("failed to get random bytes");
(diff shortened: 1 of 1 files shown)


meow2149:
> `nonce` string length shouldn't be related to platform pointer size really

cool

lucasfernog (approved):

sftse:
Isn't 64 bits a bit on the short side anyway? Collisions every 2^32 on average is usually not considered acceptable.

Legend-Master:
64 bit would be colliding every 2^64 (18_446_744_073_709_551_616) nonce on average? Unless the page's filled with attacking script tags (I doubt the page would even load if you put in so many script tags 😂), the chance for this collision should be really low

sftse:
The probability not to generate a collision within N nonce generations is reduced by the amount of nonces generated, if you successfully generated a non-colliding nonce then that gets added to the pile of nonces you must never regenerate, this is the [birthday bound](https://en.wikipedia.org/wiki/Birthday_attack). This probability grows quadratically and is the reason AEAD encryption algorithms take much longer nonces than simply 64 bits.
Chacha20Poly1305 takes a 96 bit nonce because of this.

According to the [CSP Spec](https://w3c.github.io/webappsec-csp/#security-nonces) this nonce SHOULD be at least 128 bits.

Legend-Master:
It still feels like the chance's very low to me, in that table, even when we bump the amount to 190,000 (which is already highly unlikely), it still only have a 10^9 chance of collision

Anyways, we could use 128 bits if the spec suggests so

PR #14717, opened 2025-12-30
refactor: use 128 bits for nonce generation

Legend-Master:
Also, re-use the nonce through out a single HTML file to save compute and limit collisions

Reference: https://github.com/tauri-apps/tauri/pull/14708#issuecomment-3697174838

### DIFF
--- a/crates/tauri/src/manager/mod.rs
+++ b/crates/tauri/src/manager/mod.rs
@@ -135,7 +135,7 @@ fn replace_csp_nonce(
       // > https://w3c.github.io/webappsec-csp/#security-nonces
       let mut buffer: [u8; 16] = [0; 16];
       getrandom::fill(&mut buffer).expect("failed to get random bytes");
-      format!("{buffer:x?}")
+      u128::from_ne_bytes(buffer).to_string()
     })
   });
 
--- a/crates/tauri/src/manager/mod.rs
+++ b/crates/tauri/src/manager/mod.rs
@@ -4,6 +4,7 @@
 
 use std::{
   borrow::Cow,
+  cell::OnceCell,
   collections::HashMap,
   fmt,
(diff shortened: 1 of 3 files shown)


crates/tauri/src/manager/mod.rs
```diff
@@ -127,24 +128,27 @@ fn replace_csp_nonce(
   directive: &str,
   hashes: Vec<String>,
 ) {
-  let mut nonces = Vec::new();
+  let nonce = OnceCell::new();
```
sftse, line 131: If we generate the nonce outside the callback we can replace `OnceCell` with a regular local, remove `fn replace_with_callback` entirely and use `str::replace`.

sftse (commented):
Is it fine to reuse the nonce here? I'm not familiar with what this does, but "reusing nonce" usually indicates something is seriously wrong.

crates/tauri/src/manager/mod.rs
Legend-Master: I'm not entirely sure about this but random number generation is usually slow, so in cases where we don't have inline scripts/styles, we don't pay the price for that

Legend-Master:
> Is it fine to reuse the nonce here? I'm not familiar with what this does, but "reusing nonce" usually indicates something is seriously wrong.

I believe this is the desired way to do this, the re-use only happens in a single request. We insert nonce to inline script and style tags in the HTML and add them to the CSP header. If we don't reuse a single nonce here, we will end up with more nonce that could be attacked and waste time generating them

For example, YouTube does this

crates/tauri/src/manager/mod.rs
sftse: Right, the semantics need to reflect it is only initialized if used, so `OnceCell` is a good fit. We could simplify the generic closure though and `replace_with_cache(&OnceCell<String>)`

PR #14725, opened 2026-01-03
refactor(cli): disable jsonschema resolving external resources

Legend-Master:
We only ever use it on the bundled schema for `tauri.conf.json` and it doesn't have any external resources that needs to be resolved, this removed the dependency on reqwest

Also moved this to be a `OnceCell` as it should never change


FabianLars (approved):
untested

PR #14726, opened 2026-01-03
fix: map `rustls-tls` to `reqwest/rustls-no-provider`

Legend-Master:
Reference: https://github.com/tauri-apps/tauri/pull/14724#issuecomment-3706984601

Correct me if I'm wrong, we only use `reqwest` in `tauri` to proxy requests to localhost so it's ok to use platform verifier here without `web-pki-roots` from `reqwest@0.12`'s `rustls-tls` feature

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8538,6 +8538,7 @@ dependencies = [
  "quickcheck_macros",
  "raw-window-handle",
  "reqwest 0.13.1",
+ "rustls 0.23.35",
  "serde",
  "serde_json",
  "serde_repr",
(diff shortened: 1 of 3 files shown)


FabianLars:
do we need a provider for the certs to work?

for reference, in the http plugin i wanted to do this https://github.com/tauri-apps/plugins-workspace/pull/3124/changes#diff-dee408ed956e8dd38398059404d9d9ea771834fb2792fc93ac816043cef088a1R293-R300

Legend-Master:
> do we need a provider for the certs to work?

Seems like it... Good catch, the whole TLS selection thing is so confusing to be honest 😫

> https://github.com/seanmonstar/reqwest/issues/2924

I'll get back to this tomorrow or you can take over if you want

FabianLars (approved):

Issue #4617, opened 2022-07-07
[feat] Add .taurignore and turn off the default behavior of watching all workspace members


simonhyll:
### Describe the problem

I use a monolithic style repository. Crates in my repository can be completely unrelated to the Tauri app, such as the REST API. But even if it's related to the app, for example the frontend resides in a separate workspace member than the Tauri source code. The result of watching all workspace members then is that rather than having hot reloading for my frontend, any change to the frontend results in the Tauri app being rebuilt as well. If I update the API, Tauri rebuilds, for absolutely no good reason. Even if I just change the README in another project, it triggers a rebuild of the Tauri app.

The only solution right now for me that sort of works is by passing an ignore file as an environment variable in the format of `TAURI_DEV_WATCHER_IGNORE_FILE=.taurignore`. It works, but passing that environment variable is incredibly annoying.

### Describe the solution you'd like

- Disable watching all workspace members by default. Add a CLI flag such as `--watch-members` to enable it. It's a good feature but I highly doubt most people using workspaces are using every single crate in their workspace for the Tauri project. If you disagree, at least add the reverse option, e.g. `--no-watch-members`
- Support reading files in the workspace called `.taurignore` that disable watchers for ignored files

### Alternatives considered

None really. I'll use my workaround until either of the above solutions are implemented, or until someone else suggests something better.

One potential improvement to the watched files could be only watching for changes in `*.rs` files in other workspace members, since it's pretty much guaranteed that the only reason you want to rebuild your Tauri app based on a workspace member change is when you've updated the Rust code. Icons and such most likely reside in the Tauri project itself so watching the entire Tauri project makes sense, but watching all files in all workspace members is almost guaranteed to be unnecessary and unwanted behavior, watching all Rust files should be the only thing people really need. Ignore files would still be required because again, if I have a REST API, I don't want those Rust source files to rebuild my Tauri app as well, but it'd be a start.

mokeyish:
This feature is very nice. i need it

lucasfernog:
What if we add a [tauri] section to the workspace Cargo.toml file where you can configure which packages to include or which packages to ignore?

simonhyll:
> What if we add a [tauri] section to the workspace Cargo.toml file where you can configure which packages to include or which packages to ignore?

[...]

PR #14747, opened 2026-01-08
fix(cli): only watch dependent workspace members

Legend-Master:
Reference: https://github.com/tauri-apps/tauri/issues/4617#issuecomment-3722331235

We can also add in support for watching all path dependencies in the future if we want


crates/tauri-cli/src/interface/rust.rs
```diff
@@ -1227,7 +1192,85 @@ pub(crate) fn get_cargo_metadata(tauri_dir: &Path) -> crate::Result<CargoMetadat
...
+  Ok(output.stdout)
+}
+
+pub(crate) fn get_cargo_metadata(tauri_dir: &Path) -> crate::Result<CargoMetadata> {
+  let output = run_cargo_metadata(tauri_dir)?;
+  serde_json::from_slice(&output).context("failed to parse cargo metadata")
+}
+
+#[derive(Deserialize)]
+pub struct CargoMetadataExpended {
```
FabianLars: Does it really need a second struct for this?

Legend-Master: Hmm, I'm not sure, did this because I thought it makes the ones not using it faster to parse? Never benchmarked this so I don't know, fine with merging/keeping both

FabianLars: i don't think this will make a real world difference. I think a better optimization would be to not call it as often as we do now (or cache it) but even that doesn't seem that bad atm.

i'd prefer the cleaner code over a nanosecond faster cli

Legend-Master: Merged the two

FabianLars (dismissed):

Legend-Master:
I wonder do we need to still check for workspace `.gitignore` and `.taurignore` files?

Legend-Master:
let's still load them to be safe

Edit: We never watched workspace root explicitly before... I though we did but it's really just the common ancestor of all the member crates. Doesn't hurt to add though

FabianLars (approved):

PR #14759, opened 2026-01-11
fix: fix leftover inconsistent env var in `tauri signer sign` command

### LINKED ISSUE #14758: enh: Inconsistent env vars for signing keys in tauri-cli
### Inconsistent env vars for signing keys in tauri-cli

There is an inconsistency in the environment variables used for code signing between the **Signer** and the **Bundler**.

- **Signer** (`crates/tauri-cli/src/signer/sign.rs`) uses:
  - `TAURI_PRIVATE_KEY`
  - `TAURI_PRIVATE_KEY_PASSWORD`

- **Bundler** (`crates/tauri-cli/src/bundle.rs`) uses:
  - `TAURI_SIGNING_PRIVATE_KEY`
  - `TAURI_SIGNING_PRIVATE_KEY_PASSWORD`

This appears to be an oversight from PR https://github.com/tauri-apps/tauri/pull/7949, where the env naming was updated.

#### Expected behavior
Both components should use the same environment variable names for signing keys to avoid confusion and misconfiguration.

#### Impact
Users configuring signing via environment variables may experience unexpected failures depending on which command is used.

amrbashir:
closes #14758

### DIFF
--- a/.changes/signing-env-vars.md
+++ b/.changes/signing-env-vars.md
@@ -1,0 +1,14 @@
+---
+"tauri-cli": patch:enhance
+"@tauri-apps/cli": patch:enhance
+---
+
+Added new environment variables for `tauri signer sign` command, to align with existing environment variables used in `tauri build`, `tauri bundle` and `tauri signer generate`
+- `TAURI_SIGNING_PRIVATE_KEY`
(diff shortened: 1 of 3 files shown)


Legend-Master (approved):
(untested)

PR #14766, opened 2026-01-13
build(tauri-macos-sign): remove once-cell-regex

sftse:
Tauri has a lot of dependencies, as fruitless as the quest appears to be to trim them, I did find one that is simple enough.

Matching [PR](https://github.com/tauri-apps/cargo-mobile2/pull/500) to `cargo-mobile2`.


crates/tauri-macos-sign/Cargo.toml
```diff
@@ -15,7 +15,8 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 tempfile = "3"
 x509-certificate = "0.23"
-once-cell-regex = "0.2"
+once_cell = "1.20"
+regex = "1.11"
```
FabianLars:
```suggestion
once_cell = "1"
regex = "1"
```

i try not to dictate a minimum minor version if possible.

sftse: Just for completeness, are you aware that in general just fixing the major version is a mistake?

I looked it up in the old APIs for the crates just to be sure, the API we're using was there.

FabianLars: > Just for completeness, are you aware that in general just fixing the major version is a mistake?

Not sure what you mean. If you mean that we should find out the minimum supported minor version for all crates then yes, i'm aware, it's just too much work atm because like you said, tauri has _a lot_ of dependencies.

FabianLars (approved):

Issue #14714, opened 2025-12-29
[feat] Export the exit command on AppPlugin


NyaomiDEV:
### Describe the problem

Despite exit() being a command in AppPlugin.kt (as of Tauri 2.9), I see no immediate counterpart inside the documentation, nor it being exposed as a callable function via the JS API.

### Describe the solution you'd like

Please make it available and documented.

### Additional context

https://github.com/tauri-apps/tauri/commit/3397fd9bfe5f6b1337110149f6c34731b8a44bb3#diff-d2def9c3a9932c15bc46247935a1e6edf7c7aa5064c99972b4ee750cca46c084R50

FabianLars:
Seems like it was a left-over from https://github.com/tauri-apps/tauri/pull/14133/changes/29a7dd84947415f31156a5a4f751228c8616a7aa 

According to Lucas, the normal exit methods of tauri work fine on Android so as i understand it, the kotlin code should be removed, right @lucasfernog

FabianLars:
Oh and if you want to exit the app, you're looking for https://docs.rs/tauri/latest/tauri/struct.AppHandle.html#method.exit or https://v2.tauri.app/reference/javascript/process/#exit

NyaomiDEV:
Not calling `activity.finish()` on Android causes the application to close unexpectedly (ie without a close animation.)

While this works for the intended purpose, not having an animation, or worse, making the screen blink to white and then blink back to the user homescreen, can be an issue if the user is photosensitive.

While I do agree to use `tauri-plugin-process` now that I know that there are already methods built for this, I do think this blinking bug should be handled transparently in some way.

PR #14767, opened 2026-01-13
feat: expose app exit via JS app API

seekskyworld:
Issue: https://github.com/tauri-apps/tauri/issues/14714

Summary:
- add core app exit command and expose it as JS process.exit
- remove the leftover Android AppPlugin exit command
- document app exit permissions and add a change note

Motivation:
- make the existing exit capability available through the supported API surface

Validation:
- cargo check -p tauri
- pnpm -C packages/api ts:check


FabianLars:
~~i feel like you kinda ignored my comments in the linked issue?~~ edit: or maybe not, and perhaps you just missed the last link.

anyway, i'm not opposed to an exit js command but it should be part of the `app` plugin/module or really anything other than "process" because we already have a process _plugin_.

FabianLars:
cc @whoever-in-the-working-group-cares anyone against having app.exit in js? Then the process plugin can be just for "other" processes than the current one (or both i guess)

packages/api/src/process.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @example
+ * ```typescript
+ * import { exit } from '@tauri-apps/api/process';
+ * await exit();
+ * ```
+ *
+ * @since 2.10.0
+ */
+async function exit(exitCode?: number): Promise<void> {
+  const payload = exitCode === undefined ? {} : { exitCode }
```
Legend-Master, line 27: I think we could put the default value in the js side? (e.g. `{ exitCode: exitCode ?? 0 }`)

packages/api/src/process.ts
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @example
+ * ```typescript
+ * import { exit } from '@tauri-apps/api/process';
+ * await exit();
+ * ```
+ *
+ * @since 2.10.0
+ */
+async function exit(exitCode?: number): Promise<void> {
```
Legend-Master, line 26: Let's move this to `app` (`packages/api/src/app.ts`)

Legend-Master (commented):
> cc @whoever-in-the-working-group-cares anyone against having app.exit in js? Then the process plugin can be just for "other" processes than the current one (or both i guess)

Moving `exit` to core (`app`) makes a lot of sense to me at least

NyaomiDEV:
> Moving exit to core (app) makes a lot of sense to me at least

It makes total sense, and actually it's unclear why the `process` plugin exists at all, as it only exposes exiting and restarting, something that the app should be able to do by itself.

FabianLars:
the process plugin was supposed to become more than that, more similar to nodejs' process module

crates/tauri/src/app/plugin.rs
```diff
@@ -46,6 +46,11 @@ pub fn app_hide<R: Runtime>(app: AppHandle<R>) -> crate::Result<()> {
   Ok(())
 }
 
+#[command(root = "crate")]
+pub fn exit<R: Runtime>(app: AppHandle<R>, exit_code: Option<i32>) {
+  app.exit(exit_code.unwrap_or(0));
```
lucasfernog, line 51: on Android this should route to the AppPlugin for graceful shutdown

PR #14786, opened 2026-01-17
feat(path): add WebviewWindow::convert_file_src

### LINKED ISSUE #12022: [feat] Rust equivalent convertFileSrc()
### Describe the problem

I'm currently trying to write a markdown editor in Tauri. For this I need to convert the file paths the user enters to tauris file paths.
Since I  use a Rust library to parse the markdown to HTML, I need to do the file path conversion in Rust.

For this reason I'd like to have the _convertFileSrc()_ function in Rust.

### Alternatives considered

Currently I'm trying to either call the Javascript version of the function from within Rust but this is hacky and I'd like to have a native way.

I'm also considering just passing a template to rust at the start of the program, that describes the transformation, but this feels also somewhat prone to breaking.

### Additional context

_No response_

AndyBodnar:
## Summary

Adds a Rust equivalent of the JavaScript `convertFileSrc` function to the `WebviewWindow` API. This enables file-to-URL conversion from Rust code, which is useful when processing file paths in Rust (e.g., parsing markdown with image references) and needing to convert them to URLs that the webview can load.

The implementation:
- Uses `percent-encoding` for URL-safe path encoding (already a dependency)
- Uses `dunce::simplified` to normalize Windows paths (already a dependency)
- Returns HTTP URLs on Windows/Android and custom protocol URLs elsewhere, matching the JS implementation

## Usage

```rust
use tauri::Manager;

tauri::Builder::default()
    .setup(|app| {
        let video_path = app.path().app_data_dir()?.join("video.mp4");
        let window = app.get_window("main").unwrap();
        let url = window.convert_file_src(&video_path, None);
        // On Windows: http://asset.localhost/C%3A%5CUsers%5C...
        // On macOS/Linux: asset://localhost/%2FUsers%2F...
        println!("URL: {}", url);
        Ok(())
    });
```

## Test plan

- [ ] Verify build passes in CI

Closes #12022

### DIFF
--- a/crates/tauri/src/path/android.rs
+++ b/crates/tauri/src/path/android.rs
@@ -4,6 +4,7 @@
 
 use super::Result;
 use crate::{plugin::PluginHandle, Runtime};
+use percent_encoding::{utf8_percent_encode, NON_ALPHANUMERIC};
 use std::path::{Path, PathBuf};
 
 /// A helper class to access the mobile path APIs.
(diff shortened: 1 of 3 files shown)


AndyBodnar:
This adds the Rust equivalent of the JavaScript `convertFileSrc` function. Followed the existing patterns in the codebase:

- Desktop version uses `dunce::simplified` for path canonicalization and percent-encodes the path
- Android version uses the same approach (HTTP protocol like Windows)
- iOS version also added with the same pattern

Platform-specific URL schemes:
- Windows/Android: `http://{protocol}.localhost/{encoded_path}`
- macOS/Linux/iOS: `{protocol}://localhost/{encoded_path}`

Socket Security checks passed.

AndyBodnar:
All checks passed.

**What this adds**

The Rust equivalent of the JavaScript `convertFileSrc()` function. This lets you convert a local file path into a URL that can be loaded by the webview, which is useful for displaying local images, videos, or other assets.

**Implementation details**

- Added `convert_file_src()` to `PathResolver` for desktop, Android, and iOS
- Uses `dunce::simplified()` on desktop to clean up Windows UNC paths
- Percent-encodes the path to handle special characters and spaces
- Returns platform-appropriate URLs:
  - Windows/Android: `http://{protocol}.localhost/{encoded_path}`
  - macOS/Linux/iOS: `{protocol}://localhost/{encoded_path}`
- Default protocol is "asset" but can be customized

**Why this matters**

Previously you had to use the JS bridge to call `convertFileSrc()` from Rust code, or manually construct the URL. This gives Rust-side code a clean way to generate asset URLs without going through JS.

**Testing**

- Socket Security checks passed
- Followed the existing patterns in the codebase for path handling and URL construction

AndyBodnar:
@zeenix

crates/tauri/src/path/desktop.rs
```diff
@@ -293,4 +294,43 @@ impl<R: Runtime> PathResolver<R> {
...
+  ///   });
+  /// ```
+  pub fn convert_file_src<P: AsRef<Path>>(&self, path: P, protocol: Option<&str>) -> String {
+    let protocol = protocol.unwrap_or("asset");
+    let path = dunce::simplified(path.as_ref());
+    let encoded = utf8_percent_encode(&path.to_string_lossy(), NON_ALPHANUMERIC).to_string();
+
+    #[cfg(windows)]
+    {
+      format!("http://{protocol}.localhost/{encoded}")
```
FabianLars: this can also be `https` https://docs.rs/tauri/latest/tauri/webview/struct.WebviewWindowBuilder.html#method.use_https_scheme - since this is a per-webview setting i guess the function needs another argument.

crates/tauri/src/path/android.rs
```diff
@@ -175,4 +176,36 @@ impl<R: Runtime> PathResolver<R> {
...
+  ///     Ok(())
+  ///   });
+  /// ```
+  pub fn convert_file_src<P: AsRef<Path>>(&self, path: P, protocol: Option<&str>) -> String {
+    let protocol = protocol.unwrap_or("asset");
+    let path = path.as_ref();
+    let encoded = utf8_percent_encode(&path.to_string_lossy(), NON_ALPHANUMERIC).to_string();
+
+    // Android uses HTTP protocol like Windows
+    format!("http://{protocol}.localhost/{encoded}")
```
FabianLars: same

FabianLars (changes requested):
Thanks for the PR! On top of the https comments, i need you to [sign your commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) to be able to merge the PR. Lastly, a small changefile will be needed (check .changes dir).

Thanks again :)

AndyBodnar:
Added the `use_https` parameter to `convert_file_src` so it works with webviews that have `use_https_scheme` enabled. Passing `Some(true)` will generate `https://` URLs instead of `http://` on Windows and Android. Also added the changefile and signed the commit.

PR #14791, opened 2026-01-18
refactor(cli): remove mutex on config

Legend-Master:
Follow up of #14668

### DIFF
--- a/crates/tauri-cli/src/interface/rust.rs
+++ b/crates/tauri-cli/src/interface/rust.rs
@@ -597,7 +597,7 @@ impl Rust {
               if is_configuration_file(self.app_settings.target_platform, event_path)
                 && reload_config(config, merge_configs, dirs.tauri).is_ok()
               {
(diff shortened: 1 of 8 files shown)


crates/tauri-cli/src/interface/rust.rs
```diff
@@ -539,15 +539,17 @@ impl Rust {
     .map(|c| Box::new(c) as Box<dyn DevProcess + Send>)
   }
 
-  fn run_dev_watcher<F: Fn(&mut Rust) -> crate::Result<Box<dyn DevProcess + Send>>>(
+  fn run_dev_watcher<
```
sftse, line 542: Shouldn't need higher-rank lifetimes here?

Legend-Master: Nice catch, was a left over back when I was tinkering with the function signatures

FabianLars (approved):
i wonder if this will bite us back someday

PR #14793, opened 2026-01-18
fea(wix): add minimum webview2 version support

Legend-Master:
Port #10339 to MSI (Wix) installer

The old `bundle > windows > nsis > minimumWebview2Version` is now deprecated in favor of `bundle > windows > minimumWebview2Version`

---

A note, in case we ever want to support this for MSIX, it seems like it's possible through https://learn.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-win32dependencies-externaldependency


crates/tauri-bundler/src/bundle/windows/msi/main.wxs
```diff
@@ -279,16 +279,17 @@
 
         {{#if install_webview}}
         <!-- WebView2 -->
-        <Property Id="WVRTINSTALLED">
-            <RegistrySearch Id="WVRTInstalledSystem" Root="HKLM" Key="SOFTWARE\Microsoft\EdgeUpdate\Clients\{F3017226-FE2A-4295-8BDF-00C3A9A7E4C5}" Name="pv" Type="raw" Win64="no" />
-            <RegistrySearch Id="WVRTInstalledUser" Root="HKCU" Key="SOFTWARE\Microsoft\EdgeUpdate\Clients\{F3017226-FE2A-4295-8BDF-00C3A9A7E4C5}" Name="pv" Type="raw"/>
+        <Property Id="INSTALLED_WEBVIEW2_VERSION">
```
Legend-Master, line 286: Can we can change these IDs freely or is this considered a breaking change?

crates/tauri-bundler/src/bundle/settings.rs
```diff
@@ -587,6 +591,10 @@ pub struct WindowsSettings {
   /// if you are on another platform and want to cross-compile and sign you will
   /// need to use another tool like `osslsigncode`.
   pub sign_command: Option<CustomSignCommandSettings>,
+  /// Try to ensure that the WebView2 version is equal to or newer than this version,
+  /// if the user's WebView2 is older than this version,
+  /// the installer will try to trigger a WebView2 update.
+  pub minimum_webview2_version: Option<String>,
```
Legend-Master, line 597: Adding a field here is technically breaking as well...

FabianLars: any better idea than making it all Builder patterns in v3?

crates/tauri-bundler/src/bundle/windows/msi/main.wxs
FabianLars: Should be fine but may be an annoyance for someone providing a custom template/fragment. Highly doubt they care about these ids though.

crates/tauri-bundler/src/bundle/settings.rs
Legend-Master: marking minors as breaking? 😂

crates/tauri-bundler/src/bundle/windows/msi/main.wxs
Legend-Master: I'll add some notes in the change file I guess

crates/tauri-bundler/src/bundle/settings.rs
FabianLars: time to stop using semver

crates/tauri-bundler/src/bundle/windows/msi/main.wxs
```diff
@@ -283,38 +283,62 @@
...
+            <RegistrySearch Id="Webview2VersionSystemx64" Root="HKLM" Key="SOFTWARE\WOW6432Node\Microsoft\EdgeUpdate\Clients\{F3017226-FE2A-4295-8BDF-00C3A9A7E4C5}" Name="pv" Type="raw" />
+            <RegistrySearch Id="Webview2VersionSystemx86" Root="HKLM" Key="SOFTWARE\Microsoft\EdgeUpdate\Clients\{F3017226-FE2A-4295-8BDF-00C3A9A7E4C5}" Name="pv" Type="raw" />
+            <RegistrySearch Id="Webview2VersionUser" Root="HKCU" Key="SOFTWARE\Microsoft\EdgeUpdate\Clients\{F3017226-FE2A-4295-8BDF-00C3A9A7E4C5}" Name="pv" Type="raw"/>
```
FabianLars, line 289: This is the same as the old version that used Win64="no" but i guess being more explicit is good.

FabianLars (approved):

FabianLars:
p.s. couldn't test it but i trust you :P

crates/tauri-bundler/src/bundle/windows/msi/main.wxs
Legend-Master: The description of the `Win64` attribute is kind of confusing that I never really understood it, so I just mirrored the checks from the NSIS script

> https://docs.firegiant.com/wix3/xsd/wix/registrysearch/

Legend-Master (commented):
> p.s. couldn't test it but i trust you :P

I have tested locally by setting the min version to a super high value and it worked, that's all I have tested though 🙃

PR #14797, opened 2026-01-19
Fix unsound Send+Sync impl

sftse:
This is a tiny breaking change to fix an unsoundness in the public API of tauri.
There is no non-breaking change that could fix this, and anybody without a Sync+Send error type is doing something shady anyway.

The issue is we can have an error type with an `Rc` that is then laundered by the Tauri `unsafe impl` into a `SetupError` which can be freely sent to a different thread. See this modification of helloworld.rs

```rust
fn main() {
  use std::cell::RefCell;
  use std::rc::Rc;

  thread_local! {
      static LOCAL: RefCell<Option<Rc<()>>> = RefCell::new(None);
  };
  LOCAL.with(|m| *m.borrow_mut() = Some(Rc::new(())));
  // Rc -> SetupError(Box<dyn std::error::Error + Send + Sync>) Laundromat
  // free to send one end to another thread
  if let tauri::Error::Setup(e) = tauri::Builder::default()
    .setup(move |_| Err(Error(LOCAL.with(|m| m.borrow_mut().take().unwrap())).into()))
    .invoke_handler(tauri::generate_handler![])
    .run(tauri::generate_context!(
      "../../examples/helloworld/tauri.conf.json"
    ))
    .unwrap_err()
  {
    std::thread::spawn(move || e);
  };
}

#[derive(Clone)]
struct Error(std::rc::Rc<()>);

impl std::fmt::Debug for Error {
  fn fmt(&self, _: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
    todo!()
  }
}

impl std::fmt::Display for Error {
  fn fmt(&self, _: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
    todo!()
  }
}

impl std::error::Error for Error {}
```

---------------------------------------------------------------------

There is another soundness issue in `tauri-runtime-wry`'s [`WindowsStore`](https://github.com/tauri-apps/tauri/blob/e919a760edfc115f9e4b5d841e29cc38d5535ed1/crates/tauri-runtime-wry/src/lib.rs#L428-L437) where there's an unsafe `Sync` impl even though the inner type is public. I haven't been able to verify whether the code was ever sound, but since the inner field was made public it definitely became unsound, `RefCell` is not `Sync`, but the wrapper type is, so we can trivially send references to `WindowsStore` to multiple threads and then open them up to references to `RefCell`.
The safety comments "// SAFETY: we ensure this type is only used on the main thread." are incorrect.

### DIFF
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -804,10 +804,6 @@ impl std::fmt::Debug for WindowBuilderWrapper {
   }
 }
 
-// SAFETY: this type is `Send` since `menu_items` are read only here
-#[allow(clippy::non_send_fields_in_send_ty)]
-unsafe impl Send for WindowBuilderWrapper {}
-
 impl WindowBuilderBase for WindowBuilderWrapper {}
 impl WindowBuilder for WindowBuilderWrapper {
   fn new() -> Self {
@@ -5224,5 +5220,14 @@ fn to_tao_theme(theme: Option<Theme>) -> Option<TaoTheme> {
     Some(Theme::Light) => Some(TaoTheme::Light),
     Some(Theme::Dark) => Some(TaoTheme::Dark),
     _ => None,
+  }
+}
+
+#[cfg(test)]
+mod tests {
+  fn is_send<T: Send>() {}
+  #[test]
+  fn are_send() {
+    is_send::<super::WindowBuilderWrapper>();
   }
 }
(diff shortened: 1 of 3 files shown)


FabianLars:
@WSH032 if you're still around, will this cause issues for you?

sftse:
While on the topic of unsoundness, I previously filed #13257 and [wry PR1571](https://github.com/tauri-apps/wry/pull/1571) to tighten up the code and be a bit more defensive about unsafe in general, but it seems they never got the necessary traction. Would appreciate some feedback regarding how to proceed with them.

sftse:
Woops, I should have run the code instead of just compiling it, my bad. The setup error is caught [here](https://github.com/tauri-apps/tauri/blob/53611c4d7bdaf89b9a5d7c46a9c4bf4e34216148/crates/tauri/src/app.rs#L1298-L1300) and stringified. So luckily there isn't a way to get an unsound crash from it. I'll open a separate issue for the remaining issue.

```rust
        if let Err(e) = setup(&mut self) {
          panic!("Failed to setup app: {e}");
        }
```

PR #14804, opened 2026-01-21
fix(cli): xcode-script tauri path resolution

lucasfernog:
regression from #14668
when running the tauri CLI via cargo, the CWD is changed by the xcode-script command - moves from src-tauri/gen/apple to src-tauri so the CLI can resolve the app path properly resolving early breaks this, so we must be careful with the #14668 change

### DIFF
--- a/crates/tauri-cli/src/mobile/ios/mod.rs
+++ b/crates/tauri-cli/src/mobile/ios/mod.rs
@@ -102,18 +102,23 @@ enum Commands {
 
 pub fn command(cli: Cli, verbosity: u8) -> Result<()> {
   let noise_level = NoiseLevel::from_occurrences(verbosity as u64);
-  let dirs = crate::helpers::app_paths::resolve_dirs();
   match cli.command {
-    Commands::Init(options) => init_command(
-      MobileTarget::Ios,
-      options.ci,
-      options.reinstall_deps,
-      options.skip_targets_install,
-      options.config,
-      &dirs,
-    )?,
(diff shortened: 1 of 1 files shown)


FabianLars (approved):

sftse:
Thanks, I should have caught that in review.

crates/tauri-cli/src/mobile/ios/mod.rs
```diff
@@ -102,18 +102,23 @@ enum Commands {
...
-    Commands::Init(options) => init_command(
-      MobileTarget::Ios,
-      options.ci,
-      options.reinstall_deps,
-      options.skip_targets_install,
-      options.config,
-      &dirs,
-    )?,
+    Commands::Init(options) => {
+      let dirs = crate::helpers::app_paths::resolve_dirs();
```
Legend-Master, line 107: Let's move these `let dirs = crate::helpers::app_paths::resolve_dirs();` to each sub commands?

Legend-Master (commented):
> I am a little concerned about https://github.com/tauri-apps/tauri/pull/14632/commits/d97b190268699e7945c1d9263ba855a63dc17fe4 that it changed the timing on when we resolve the tauri directories which might get affected by the working directory (we do change them later on in some places)
> 
> _Originally posted by @Legend-Master in https://github.com/tauri-apps/tauri/issues/14632#issuecomment-3627361907_

Guess I wasn't careful about this after all 😂

FabianLars:
#14833

PR #14805, opened 2026-01-21
refactor: move unsafe send sync impl up to `Context`

### LINKED ISSUE #14801: [bug] Unsound Sync implementation
### Describe the bug

[WindowsStore](https://github.com/tauri-apps/tauri/blob/e919a760edfc115f9e4b5d841e29cc38d5535ed1/crates/tauri-runtime-wry/src/lib.rs#L428-L437) is a public type with an `unsafe impl Sync` override, even though the inner field is a public `RefCell`. Rough sketch how to cause UB:

```rust
use std::cell::RefCell;

unsafe impl Sync for BadSync {}

struct BadSync(RefCell<()>);

fn main() {
    let badsync = BadSync(RefCell::new(()));
    let x = &badsync;
    std::thread::scope(|s| {
        s.spawn(|| loop {
            *x.0.borrow_mut() = ();
        });
        s.spawn(|| loop {
            *x.0.borrow_mut() = ();
        });
    });
}
```

This doesn't crash on my machine, but we can run it as `cargo miri run` to confirm it is unsound.
This inner `RefCell` is accessible to any implementer of `Plugin`

```rust
use tao::event::Event;
use tao::event_loop::{ControlFlow, EventLoopProxy, EventLoopWindowTarget};
use tauri_runtime::UserEvent;
use tauri_runtime_wry::EventLoopIterationContext;
use tauri_runtime_wry::Message;
use tauri_runtime_wry::Plugin;
use tauri_runtime_wry::WebContextStore;
use tauri_runtime_wry::WindowsStore;

impl<T: UserEvent> Plugin<T> {
    fn on_event(
        &mut self,
        event: &Event<'_, Message<T>>,
        event_loop: &EventLoopWindowTarget<Message<T>>,
        proxy: &EventLoopProxy<Message<T>>,
        control_flow: &mut ControlFlow,
        context: EventLoopIterationContext<'_, T>,
        web_context: &WebContextStore,
    ) -> bool {
        let windows: &WindowsStore = &*context.windows;
        std::thread::scope(|s| {
            s.spawn(|| loop {
                *windows.0.borrow_mut() = todo!();
            });
            s.spawn(|| loop {
                *windows.0.borrow_mut() = todo!();
            });
        });
        true
    }
}
```

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
master
```

### Stack trace

```text

```

### Additional context

_No response_

sftse:
Fixes #14801

Depends on [wry PR 1657](https://github.com/tauri-apps/wry/pull/1657).

I'm not intimately familiar with what is going on in this code, so I'm using heuristics to guide my fix:

- most of the arguments passed to `fn on_event` from the `Plugin` trait are not Sync
- `WindowsStore` is sandwiched between two types that are both not Sync, it contains types that are not Sync and it itself is contained in
```rust
unsafe impl<T: UserEvent> Sync for DispatcherMainThreadContext<T> {}

pub struct DispatcherMainThreadContext<T: UserEvent> {
  pub window_target: EventLoopWindowTarget<Message<T>>,
  pub web_context: WebContextStore,
  // changing this to an Rc will cause frequent app crashes.
  pub windows: Arc<WindowsStore>,
  #[cfg(feature = "tracing")]
  pub active_tracing_spans: ActiveTraceSpanStore,
}
```
and
```rust
pub struct EventLoopIterationContext<'a, T: UserEvent> {
  pub callback: &'a mut (dyn FnMut(RunEvent<T>) + 'static),
  pub window_id_map: WindowIdStore,
  pub windows: Arc<WindowsStore>,
  #[cfg(feature = "tracing")]
  pub active_tracing_spans: ActiveTraceSpanStore,
}
```
of which the last is not Sync.

For basic soundness based on the examples in #14801 if a non-Sync field is public in a public struct that struct *must* be not Sync as well.
We have two choices, make both these structs that contain public `WindowsStore`s `Sync` or make them both not Sync. Attempting to make both of these structs `Sync` is a rat's nest, so let's take the alternative, let's assume they must be both not Sync. This implies the Sync impl of both `WindowsStore` and `DispatcherMainThreadContext` are mistaken.

Removing the Sync impl of `DispatcherMainThreadContext` leads to another rat's nest of errors. `Context` used to inherit Sync based on its fields, so removing Sync from `DispatcherMainThreadContext` makes it not Sync as well.

I can't reason whether `Context` should or should not be Sync, but in any case we are overriding Sync "at the edges" of the code, so there is a larger space in which the absence of Sync will be caught by the compiler.

In summary, these changes strictly reduce the ways to be unsound. More cases are caught that were previously not sound.

### DIFF
--- a/.changes/fix-send-sync.md
+++ b/.changes/fix-send-sync.md
@@ -1,0 +1,5 @@
+---
+"tauri-runtime-wry": minor:breaking
+---
+
+`WindowsStore` and `DispatcherMainThreadContext` are no longer `Send` and `Sync`, and unsafe impl has been moved to `Context` directly.
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -249,6 +249,7 @@ pub(crate) fn send_user_message<T: UserEvent>(
   }
 }
 
+unsafe impl<T: UserEvent> Send for Context<T> {}
 unsafe impl<T: UserEvent> Sync for Context<T> {}
 
 #[derive(Clone)]
@@ -434,10 +435,6 @@ pub enum ActiveTracingSpan {
 #[derive(Debug)]
 pub struct WindowsStore(pub RefCell<BTreeMap<WindowId, WindowWrapper>>);
 
-// SAFETY: we ensure this type is only used on the main thread.
-#[allow(clippy::non_send_fields_in_send_ty)]
-unsafe impl Send for WindowsStore {}
-
 #[derive(Debug, Clone)]
 pub struct DispatcherMainThreadContext<T: UserEvent> {
   pub window_target: EventLoopWindowTarget<Message<T>>,
@@ -447,10 +444,6 @@ pub struct DispatcherMainThreadContext<T: UserEvent> {
   #[cfg(feature = "tracing")]
   pub active_tracing_spans: ActiveTraceSpanStore,
 }
-
-// SAFETY: we ensure this type is only used on the main thread.
-#[allow(clippy::non_send_fields_in_send_ty)]
-unsafe impl<T: UserEvent> Send for DispatcherMainThreadContext<T> {}
 
 impl<T: UserEvent> fmt::Debug for Context<T> {
   fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -249,9 +249,6 @@ pub(crate) fn send_user_message<T: UserEvent>(
   }
 }
 
-unsafe impl<T: UserEvent> Send for Context<T> {}
-unsafe impl<T: UserEvent> Sync for Context<T> {}
-
 #[derive(Clone)]
 pub struct Context<T: UserEvent> {
   pub window_id_map: WindowIdStore,
@@ -265,6 +262,9 @@ pub struct Context<T: UserEvent> {
   next_webview_event_id: Arc<AtomicU32>,
   webview_runtime_installed: bool,
 }
+
+unsafe impl<T: UserEvent> Send for Context<T> {}
+unsafe impl<T: UserEvent> Sync for Context<T> {}
 
 impl<T: UserEvent> Context<T> {
   pub fn run_threaded<R, F>(&self, f: F) -> R
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -249,6 +249,8 @@ pub(crate) fn send_user_message<T: UserEvent>(
   }
 }
 
+unsafe impl<T: UserEvent> Sync for Context<T> {}
+
 #[derive(Clone)]
 pub struct Context<T: UserEvent> {
   pub window_id_map: WindowIdStore,
@@ -436,10 +438,6 @@ pub struct WindowsStore(pub RefCell<BTreeMap<WindowId, WindowWrapper>>);
 #[allow(clippy::non_send_fields_in_send_ty)]
 unsafe impl Send for WindowsStore {}
 
-// SAFETY: we ensure this type is only used on the main thread.
-#[allow(clippy::non_send_fields_in_send_ty)]
-unsafe impl Sync for WindowsStore {}
-
 #[derive(Debug, Clone)]
 pub struct DispatcherMainThreadContext<T: UserEvent> {
   pub window_target: EventLoopWindowTarget<Message<T>>,
@@ -453,10 +451,6 @@ pub struct DispatcherMainThreadContext<T: UserEvent> {
 // SAFETY: we ensure this type is only used on the main thread.
 #[allow(clippy::non_send_fields_in_send_ty)]
 unsafe impl<T: UserEvent> Send for DispatcherMainThreadContext<T> {}
-
-// SAFETY: we ensure this type is only used on the main thread.
-#[allow(clippy::non_send_fields_in_send_ty)]
-unsafe impl<T: UserEvent> Sync for DispatcherMainThreadContext<T> {}
 
 impl<T: UserEvent> fmt::Debug for Context<T> {
   fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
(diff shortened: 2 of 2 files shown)


sftse:
Because of
```rust
unsafe impl<T: UserEvent> Send for DispatcherMainThreadContext<T> {}

#[derive(Clone)]
pub struct DispatcherMainThreadContext<T: UserEvent> {
  pub window_target: EventLoopWindowTarget<Message<T>>,
  pub web_context: WebContextStore,
  // changing this to an Rc will cause frequent app crashes.
  pub windows: Arc<WindowsStore>,
  #[cfg(feature = "tracing")]
  pub active_tracing_spans: ActiveTraceSpanStore,
}
```
there's another change necessary, originally missed.
`Arc<T>` is only Send if `T: Send`, but the Send on `DispatcherMainThreadContext` allows an unsound bypass:
Clone `DispatcherMainThreadContext` -> send clone to another thread -> take inner `Arc<WindowsStore>` -> access `WindowsStore` unsynchronized from both threads. So we have to do the same fix as with Sync and remove the Send impl, delay it until `Context`.

Maybe it is still unsound that `Context` is Send and Sync, but the scope for unsound mistakes is reduced.

sftse:
As the broken CI on [wry PR 1657](https://github.com/tauri-apps/wry/pull/1657) shows, that fix was a bit optimistic.

I don't know when I can further look into this, but I've pushed a fix that pulls the unsafety at least to the internals of Tauri instead of exposing it publically. Maybe this is enough of an improvement to be merged?

Summarizing, tauri currently exposes unsoundness in its public api, we might be interested in merging something that breaks as little as possible to fix this, this is such a minimal suggestion.

The fact that `wry` does end up needing Send [here](https://github.com/tauri-apps/wry/blob/9fa99b14fd8fb8aa739a17c6bb6e074c858da0c8/src/webview2/mod.rs#L759) for the request handler is a bit rough though. Somebody with a better grasp of the code should probably be looking into this.

```rust
          std::thread::spawn(move || match new_window_req_handler(uri, features) { ... });
```

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -2795,6 +2782,7 @@ impl<T: UserEvent> Wry<T> {
     let main_thread_id = current_thread().id();
     let web_context = WebContextStore::default();
 
+    #[allow(clippy::arc_with_non_send_sync)]
```
Legend-Master, line 2833: The code below should not be needed now with https://github.com/tauri-apps/wry/pull/1660 in

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -253,6 +253,9 @@ pub(crate) fn send_user_message<T: UserEvent>(
   }
 }
 
+unsafe impl<T: UserEvent> Send for Context<T> {}
```
Legend-Master: The way this moved the unsafe `Send` and `Sync` impl to `Context` makes sense to me

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -428,14 +431,6 @@ pub enum ActiveTracingSpan {
 #[derive(Debug)]
 pub struct WindowsStore(pub RefCell<BTreeMap<WindowId, WindowWrapper>>);
```
Legend-Master, line 436: I think maybe we can do this the thing I did in https://github.com/tauri-apps/tauri/pull/15411/changes/021476b4db623051c1494e19a1cb85289d2ff45d

sftse:
This would probably fix #10001

Legend-Master:
https://github.com/tauri-apps/tauri/issues/10001 is different that it's a panic on `global_shortcut_manager` that no longer exists in v2

sftse:
Might make sense to close it?

crates/tauri-runtime-wry/src/lib.rs
sftse: Can we elaborate on this in another PR? I don't have the bandwidth except to see this merged.

Legend-Master (dismissed):
Since this is breaking change in the `tauri-runtime-wry` crate, I will wait a bit to merge this in tauri 2.12 window to publish it as a minor bump

.changes/fix-send-sync.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri": patch:sec
```
Legend-Master:
```suggestion
"tauri-runtime-wry": minor:breaking
```

.changes/fix-send-sync.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Redraw Send/Sync boundaries to fix unsoundness
```
Legend-Master:
```suggestion
`WindowsStore` and `DispatcherMainThreadContext` are no longer `Send` and `Sync`, and unsafe impl has been moved to `Context` directly.
```

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -249,6 +249,9 @@ pub(crate) fn send_user_message<T: UserEvent>(
   }
 }
 
+unsafe impl<T: UserEvent> Send for Context<T> {}
+unsafe impl<T: UserEvent> Sync for Context<T> {}
```
Legend-Master: A small nitpick, maybe moving this to below `pub struct Context`

Legend-Master (changes requested):

crates/tauri-runtime-wry/src/lib.rs
Legend-Master: Basically put the entire `DispatcherMainThreadContext` behind an `Arc` and store it in `Wry`, then only store a `Weak<DispatcherMainThreadContext>` in the `Context`.

Anyways, that should be explored in #15411 instead.

sftse:
@Legend-Master 
Is it fine with you to ping me to rebase my PRs? I understand this is a bit a different workflow than what is common with `tauri` where branches are updated by merging but it really helps me keep the history readable, I'll try to keep a short turnaround time.

Legend-Master:
If it works better for you, sure. I personally don't like to do that because it removes the history in the PR that when you go back some time after, you kinda forget what each review comments were about

sftse:
I fully agree, GH forces a tradeoff here that is not great. Stacked diffs are the future..

Another papercut is that by squashing onto master `git pull --rebase origin dev` simply doesn't work reliably without phantom merge conflicts.

Legend-Master (approved):
let's roll

PR #14812, opened 2026-01-23
fix(specta): don't use `#[specta(rename = ...)]` with `tauri::ipc::Channel`

oscartbeaumont:
With some futures changes coming in the next Specta release this will cause problems.

This change of code is backwards compatible because Specta does something like `attributes.rename.unwrap_or_else(|| ident.to_string())` so removing the `attributes.rename` is fine if we change `ident` to match it's value.

### DIFF
--- a/.changes/change-pr-14812.md
+++ b/.changes/change-pr-14812.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch
+---
+
+fix(specta): don't use `#[specta(rename = ...)]` with `tauri::ipc::Channel`
--- a/.changes/change-pr-14812.md
+++ b/.changes/change-pr-14812.md
@@ -1,5 +1,5 @@
 ---
-"tauri": patch
+"tauri": patch:bug
 ---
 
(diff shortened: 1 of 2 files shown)


oscartbeaumont:
I think the change tag CI workflow is not working? I have got a changeset file which marks tauri as needing a patch but it's saying it doesn't.

I don't think the Clippy failure is the fault of this PR.

FabianLars:
>  I have got a changeset file which marks tauri as needing a patch but it's saying it doesn't.

For some reason we enforce patch types in this repo, so for example instead of `tauri: patch` you need `tauri: patch:bug` - list of tags https://github.com/tauri-apps/tauri/blob/dev/.changes/config.json#L3-L12

> I don't think the Clippy failure is the fault of this PR.

yeah, we didn't adapt to the 1.93 release from yesterday yet

oscartbeaumont:
Ohhh, that makes sense! Updated the changeset.

FabianLars (approved):

PR #14833, opened 2026-01-26
Less statics fixup

sftse:
You break, you fix.

Supersedes #14804 

I rechecked #14668 for unintended changes.

If we cherry-pick f7d6378e952b678c8464c7d976c6473445d47352 onto the changes #14668 and run `git checkout pr-14668-plus-cherry && git diff HEAD~2` we can manually grep for `resolve\(` and check that all the original calls that were removed are replaced with an identical call to the new `fn resolve_dirs`. The one or two changes that are not of the form
```text
- crate::helpers::app_paths::resolve();
+ let dirs = crate::helpers::app_paths::resolve_dirs();
```
are within the same hunk and there are no function calls between the call sites that could affect the resolution.

### DIFF
--- a/crates/tauri-cli/src/mobile/android/mod.rs
+++ b/crates/tauri-cli/src/mobile/android/mod.rs
@@ -104,17 +104,19 @@ enum Commands {
 }
 
 pub fn command(cli: Cli, verbosity: u8) -> Result<()> {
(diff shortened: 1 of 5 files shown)


sftse:
Fixed CI

sftse:
Windows runner died at random?

Legend-Master (approved):

PR #11694, opened 2024-11-14
refactor(cli): use `cargo run` for `tauri dev`

FabianLars:
closes #4724
edit: closes #8169

Soooo i have small problem here. I debugged this for so long (hours) that i don't even remember why i was debugging it. The only difference between the current behavior and this PR is a new error message on ctrlc but i think this `cargo run` is printing that.

i would appreciate some testing and telling me what i missed.

p.s. the cargo check workflow will fail. didn't want to remove DevCleanup and manually_killed until someone tells me what i missed.

p.p.s. i'll try to do some refactoring after this gets merged. the whole cli dev/build impl is a bit convoluted 😅 

p.p.p.s idk why github messes up the diff this much, it's better in vscode for me

PR #14836, opened 2026-01-27
refactor tauri-cli

sftse:
Make more things concrete rather than abstract.

### DIFF
--- a/.changes/change-pr-14836.md
+++ b/.changes/change-pr-14836.md
@@ -1,5 +1,6 @@
 ---
-"tauri-cli": patch
+"@tauri-apps/cli": patch:changes
+"tauri-cli": patch:changes
 ---
 
 Continued refactors of tauri-cli, fix too weak atomics.
--- a/.changes/change-pr-14836.md
+++ b/.changes/change-pr-14836.md
@@ -1,0 +1,5 @@
+---
+"tauri-cli": patch
+---
+
+Continued refactors of tauri-cli, fix too weak atomics.
--- a/crates/tauri-cli/src/dev.rs
+++ b/crates/tauri-cli/src/dev.rs
@@ -25,13 +25,13 @@ use std::{
   process::{exit, Command, Stdio},
   sync::{
     atomic::{AtomicBool, Ordering},
-    Arc, Mutex, OnceLock,
+    OnceLock,
   },
 };
 
 mod builtin_dev_server;
 
-static BEFORE_DEV: OnceLock<Mutex<Arc<SharedChild>>> = OnceLock::new();
+static BEFORE_DEV: OnceLock<SharedChild> = OnceLock::new();
 static KILL_BEFORE_DEV_FLAG: AtomicBool = AtomicBool::new(false);
 
 #[cfg(unix)]
@@ -205,11 +205,10 @@ pub fn setup(
 
         let child = SharedChild::spawn(&mut command)
           .unwrap_or_else(|_| panic!("failed to run `{before_dev}`"));
-        let child = Arc::new(child);
-        let child_ = child.clone();
 
+        let child = BEFORE_DEV.get_or_init(move || child);
         std::thread::spawn(move || {
-          let status = child_
+          let status = child
             .wait()
             .expect("failed to wait on \"beforeDevCommand\"");
           if !(status.success() || KILL_BEFORE_DEV_FLAG.load(Ordering::Relaxed)) {
@@ -217,8 +216,6 @@ pub fn setup(
             exit(status.code().unwrap_or(1));
           }
         });
-
-        BEFORE_DEV.set(Mutex::new(child)).unwrap();
 
         let _ = ctrlc::set_handler(move || {
           kill_before_dev_process();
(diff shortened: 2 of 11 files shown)


crates/tauri-cli/src/interface/rust.rs
```diff
@@ -582,17 +579,12 @@ impl Rust {
                 display_path(event_path.strip_prefix(dirs.frontend).unwrap_or(event_path))
               );
 
-              let mut p = process.lock().unwrap();
-              p.kill().context("failed to kill app process")?;
+              child.kill().context("failed to kill app process")?;
 
               // wait for the process to exit
               // note that on mobile, kill() already waits for the process to exit (duct implementation)
-              loop {
```
sftse, line 590: This code is busy looping waiting for the child to exit, better block and save those CPU cycles.

crates/tauri-cli/src/dev.rs
```diff
@@ -205,21 +205,18 @@ pub fn setup(
...
-        let child = Arc::new(child);
-        let child_ = child.clone();
 
+        let child = BEFORE_DEV.get_or_init(move || child);
         std::thread::spawn(move || {
-          let status = child_
+          let status = child
             .wait()
             .expect("failed to wait on \"beforeDevCommand\"");
-          if !(status.success() || KILL_BEFORE_DEV_FLAG.load(Ordering::Relaxed)) {
```
sftse, line 215: The use of `Ordering::Relaxed` is almost always incorrect without use of memory barriers, but most CPUs don't even have real relaxed orderings, so the bug will hardly surface in practice.

crates/tauri-cli/src/interface/mod.rs
```diff
@@ -104,35 +98,3 @@ pub enum ExitReason {
   /// Regular exit.
   NormalExit,
 }
-
-pub trait Interface: Sized {
```
Legend-Master, line 108: I don't mind removing this, but I think we should do it in a separate PR

Legend-Master (commented):
To make this easier for us to look back in history when we squash merge them, I think we should split the PR into 3 different ones:

1. busy loop fix
2. get config and os string refactors
3. remove `Interface`

sftse:
I'm not opposed to splitting this PR, but it seems a bit synthetic if it's just done to work well with a squash.
While in the middle of a refactor I occasionally find things to fix and splitting this into different PRs can mean a lot of rebasing ceremony to satisfy these constraints.

That said, I'm not familiar with the benefits of squashing, so interested to hear a different take.

Legend-Master:
First off, we have set to only allow squash merge (could of course make exceptions)

To me with squash merge, we maintain a linear history and can create as many small commits in a PR without having to worry about polluting the main history
Also I like to treat each PR as a separate topic, which is pretty inline with what squash merge does

> While in the middle of a refactor I occasionally find things to fix and splitting this into different PRs can mean a lot of rebasing ceremony to satisfy these constraints.

I fully understand this though, I just think there're too much unrelated things to the pull request topic, like the entire busy loop fix is ~10 lines in 8634d96b82383a4dd4a85e1f30d20518363baa5e but the refactor of other things took ~200 lines (very much appreciate the clear commit messages though)

And also in this specific case, the removal of the `Interface` related change might also need some extra look from other team members since I don't really know the reason why they existed at the first place

crates/tauri-cli/src/interface/rust.rs
sftse: Split out into #14839

sftse:
Managed to mess up the change file, rerun.

> I fully understand this though, I just think there're too much unrelated things to the pull request topic

Yeah, I agree in this specific case, splitting is fairly easy. If I'm already 20 commits in and splitting would be a gnarly rebase, I hope we can be flexible if the need arises :)

crates/tauri-cli/src/mobile/mod.rs
```diff
@@ -86,7 +81,7 @@ impl DevProcess for DevChild {
   }
 
   fn manually_killed_process(&self) -> bool {
```
Legend-Master, line 83: Looking at the code, it seems like the use for this is removed in #11694 and we should be able to remove this entirely right? cc @FabianLars

.changes/change-pr-14836.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri-cli": patch
```
Legend-Master:
```suggestion
"@tauri-apps/cli": patch:changes
"tauri-cli": patch:changes
```

Legend-Master (dismissed):
Would like to wait a bit for the response and see if we can remove `manually_killed_process` entirely

crates/tauri-cli/src/mobile/mod.rs
sftse: I haven't removed it because I may still need it for an upcoming PR that hasn't been started yet.

FabianLars: > Looking at the code, it seems like the use for this is removed in https://github.com/tauri-apps/tauri/pull/11694

Hmm yeah, now that you mention it it's not actually used on mobile, i didn't check that in the linked PR and just left mobile untouched.

Legend-Master: > I haven't removed it because I may still need it for an upcoming PR that hasn't been started yet.

We can leave it for now then

Legend-Master (approved):

PR #14840, opened 2026-01-28
refactor(tauri-cli): remove trait implemented only once

sftse:
@Legend-Master

### DIFF
--- a/crates/tauri-cli/src/build.rs
+++ b/crates/tauri-cli/src/build.rs
@@ -11,7 +11,7 @@ use crate::{
     config::{get_config, ConfigMetadata, FrontendDist},
   },
   info::plugins::check_mismatched_packages,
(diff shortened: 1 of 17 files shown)


Legend-Master:
cc @lucasfernog since this predates back to #4447, I'm not extract sure if it's fine to remove

PS: I'm good with the change personally

sftse:
When I was doing the previous refactors, the trait kept getting in the way, since the changes in the impl always had to be mirrored onto the trait itself. I think if there's a single impl the costs are hardly ever worth it.

FabianLars:
Not Lucas here, let's just do it. I don't see a reason for this either.

FabianLars (approved):

Legend-Master (approved):

PR #12164, opened 2025-01-03
fix(cli): Ignore file access events

FabianLars:
fixes #12141

The new notify / debouncer-mini versions now also report Access events on Linux which breaks tauri dev

PR #14844, opened 2026-01-29
refactor(cli): reduce some nesting code

Legend-Master:


crates/tauri-cli/src/interface/rust.rs
```diff
@@ -553,43 +554,46 @@ impl Rust {
...
-          }
+    while let Ok(events) = rx.recv() {
+      for event in events {
+        if event.kind.is_access() {
+          continue;
+        }
+
+        let Some(event_path) = event.paths.first() else {
+          continue;
+        };
```
sftse: Looks a bit odd only using the first path? Could maybe use a comment.

Legend-Master: Not sure why, I simply did a `Convert to guarded return` refactor using rust analyzer here

sftse: Clearly a behavior-preserving refactor, just a drive-by remark.

Legend-Master: Seems like it's from #12164 when we switched to debouncer full which uses it for rename events, I'll see if we want to handle it here

FabianLars: > I'll see if we want to handle it here

i don't see why we would want to

Legend-Master: For also accounting the renamed to file?

FabianLars: haha yes obviously. I can't remember what I was getting at anymore, just that I think there's little benefit in behavior compared to the lines of code added - since it's just 5 or so you can ignore me and go on with your day x)

FabianLars (approved):

PR #14816, opened 2026-01-23
feat: add convert_file_src function to path module

LouisLau-art:
## Summary
Adds a new public method `PathResolver::convert_file_src` that converts a file path to a Tauri asset URL. This function is useful when you need to reference local files in your frontend (e.g., in markdown editors where you're parsing markdown to HTML in Rust).

## Implementation
- Uses `dunce` to canonicalize paths (normalizes paths like `C:\` on Windows)
- Uses `percent-encoding` to URL-encode the path
- All dependencies (`dunce`, `percent-encoding`) are already in Cargo.toml

## Platform-specific behavior
- **Windows and Android**: Returns URLs in format `http://asset.localhost/<path>`
- **Other platforms**: Returns URLs in format `asset://localhost/<path>`

## Example usage
\`\`\`rust
use tauri::Manager;
use std::path::Path;

tauri::Builder::default()
  .setup(|app| {
    let path = Path::new("/path/to/image.png");
    let asset_url = app.path().convert_file_src(path)?;
    println!("Asset URL: {}", asset_url);
    Ok(())
  });
\`\`\`

## Motivation
Previously, users had to manually call the JavaScript version of `convertFileSrc()` from Rust, which was hacky. This provides a native Rust way to convert file paths to Tauri asset URLs.

Closes #12022

PR #14854, opened 2026-01-29
feat(path): add Rust convert_file_src function

### LINKED ISSUE #12022: [feat] Rust equivalent convertFileSrc()

ddmoney420:
## Summary
Adds a Rust equivalent of the JavaScript `convertFileSrc()` function, allowing backend code to convert device file paths to URLs that can be loaded by the webview.

## Motivation
This is useful when processing file paths in Rust (e.g., markdown parsers, image processors) that need to generate URLs for the webview to load assets. Currently this requires hacky workarounds like calling JavaScript from Rust or manually implementing the URL transformation.

## Implementation
The function in `tauri::path::convert_file_src`:
- URL encodes the file path using `percent_encoding`
- Uses platform-appropriate scheme (`http://` on Windows/Android, custom protocol on other platforms)
- Supports custom protocols via optional parameter
- Matches the behavior of the JavaScript `convertFileSrc()` exactly

## Usage
```rust
use tauri::path::convert_file_src;

// Default "asset" protocol
let url = convert_file_src("/path/to/video.mp4", None);
// Returns: "asset://localhost/%2Fpath%2Fto%2Fvideo.mp4" (on macOS/Linux)
// Returns: "http://asset.localhost/%2Fpath%2Fto%2Fvideo.mp4" (on Windows/Android)

// Custom protocol
let url = convert_file_src("/path/to/file", Some("custom"));
```

## Test plan
- Added unit tests for default protocol, custom protocol, and special character handling
- Tests verify platform-specific URL schemes

Fixes #12022

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/crates/tauri/src/path/mod.rs
+++ b/crates/tauri/src/path/mod.rs
@@ -7,6 +7,56 @@ use std::{
   str::FromStr,
 };
 
+/// Convert a device file path to a URL that can be loaded by the webview.
+///
+/// Note that `asset:` and `http://asset.localhost` must be added to
+/// [`app.security.csp`](https://v2.tauri.app/reference/config/#csp-1) in `tauri.conf.json`.
+///
+/// Additionally, `"enable": true` must be added to
+/// [`app.security.assetProtocol`](https://v2.tauri.app/reference/config/#assetprotocolconfig)
+/// in `tauri.conf.json` and its access scope must be defined on the `scope` array.
+///
+/// # Arguments
+///
+/// * `file_path` - The file path to convert.
+/// * `protocol` - The protocol to use. Defaults to `"asset"`. You only need to set this
+///   when using a custom protocol.
+///
+/// # Example
+///
+/// ```rust,no_run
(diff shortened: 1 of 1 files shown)


FabianLars:
Thanks for the PR but just like #14816 i'll close it in favor of #14786.
Funny that all 3 PRs made the same mistake of missing the `https` option though 😅

PR #14856, opened 2026-01-30
perf: optimize event emission by passing filters by reference

Tunglies:


Legend-Master (changes requested):
I don't think it really makes any differences passing a reference here, I believe rust compiler is smart enough that they would produce the exact same final code

PR #14857, opened 2026-01-30
refactor: rewrite some `&String` to  `&str`

Tunglies:


Tunglies:
Wonder if prefer separate PR or combine them in one. Looking for the more performance improvements.

crates/tauri-build/src/lib.rs
```diff
@@ -593,8 +593,8 @@ pub fn try_build(attributes: Attributes) -> Result<()> {
         .icon
         .iter()
         .find(|i| predicate(i))
-        .cloned()
-        .unwrap_or_else(|| default.to_string());
+        .map(AsRef::as_ref)
+        .unwrap_or_else(|| default);
       icon_path.into()
     }
```
sftse: The `Fn(&&String) -> bool` would also be more natural as `Fn(&str) -> bool` and we don't need to clone into a `PathBuf` to return it. But then we need to both annotate lifetimes and the `find_icon` name is a bit unidiomatic, un-std-like in that it returns something else than an `Option<&T>`.

Suggestion, we make the function mirror the stdlib function, and move the `unwrap_or` to the callsite.

```rust
    fn find_icon<F: Fn(&str) -> bool>(config: &Config, predicate: F) -> Option<&str> {
      config
        .bundle
        .icon
        .iter()
        .find(|i| predicate(i))
        .map(|s| s.as_str())
    }

    let window_icon_path = attributes
      .windows_attributes
      .window_icon_path
      .unwrap_or_else(|| {
        find_icon(&config, |i| i.ends_with(".ico"))
          .unwrap_or("icons/icon.ico")
          .into()
      });
```

This is just marginally longer, but bubbles clones up the call chain. The higher up, the better visibility we have about whether they are really necessary or just an artifact of our own making.

sftse (commented):
Personal opinion, I'm generally in favor of removing clones because they pollute my reading cache, but rarely declare them as `perf`. If I do declare them as `perf`, it should be justified by benchmarks, bonus points for a macroscopic example that had real user impact because of the changes.

Single clones, not called in a loop, are not perf-work to me.

Leaning into borrowing more strongly also helps to catch bugs from refactors, as the code is very "tight" in a sense, has high visibility to the borrow checker and we cannot mutate one code path without forgetting to update the other.

```text
A -> B -> \  // two code paths that borrow from A, encodes tight dependency on the data
 \ -> C -> D
```
If the code paths `A -> B` and `A -> C` clone to ownership, there is a chance to introduce changes just to one codepath but not the other, and they mess up whatever we were doing in `D`.

.changes/perf-tauri-build-string.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri-build": perf
+---
+
+Refactor find_icon to use &str to remove unnecessary clones. No user facing changes.
```
Legend-Master: Those refactors are small enough that I think we could merge in without putting them in the change log 🙃

crates/tauri-build/src/lib.rs
```diff
@@ -593,8 +593,8 @@ pub fn try_build(attributes: Attributes) -> Result<()> {
         .icon
         .iter()
         .find(|i| predicate(i))
-        .cloned()
-        .unwrap_or_else(|| default.to_string());
+        .map(AsRef::as_ref)
```
Legend-Master: To be honest, I think we should just remove this function and inline to caller

Legend-Master (changes requested):
> Wonder if prefer separate PR or combine them in one. Looking for the more performance improvements.

If they do something on the same topic, it's ok to combine, and separate PRs are fine here too

And I agree with @sftse that refactor is a better title here

Tunglies:
Thanks for reviewing and giving advices.

- Removed the Changelog
- Inline the find_iconn
- Modified  another `&&String` to `&str`

Tunglies:
```bash
   Compiling serde v1.0.228
error: failed to run custom build command for `libc v0.2.178`

Caused by:
  process didn't exit successfully: `/target/debug/build/libc-922dd68bc8706b43/build-script-build` (exit status: 1)
  --- stderr
  /target/debug/build/libc-922dd68bc8706b43/build-script-build: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /target/debug/build/libc-922dd68bc8706b43/build-script-build)
  /target/debug/build/libc-922dd68bc8706b43/build-script-build: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /target/debug/build/libc-922dd68bc8706b43/build-script-build)
  /target/debug/build/libc-922dd68bc8706b43/build-script-build: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /target/debug/build/libc-922dd68bc8706b43/build-script-build)
warning: build failed, waiting for other jobs to finish...
error: failed to run custom build command for `serde_core v1.0.228`
```

Is it due to my RR? XD

crates/tauri-codegen/src/context.rs
```diff
@@ -493,14 +493,14 @@ pub fn context_codegen(data: ContextData) -> EmbeddedAssetsResult<TokenStream> {
 fn find_icon(
   config: &Config,
   config_parent: &Path,
-  predicate: impl Fn(&&String) -> bool,
+  predicate: impl Fn(&str) -> bool,
```
Legend-Master: I think keeping the old one here is fine, it's a bit tighter

crates/tauri-build/src/lib.rs
```diff
@@ -587,21 +587,19 @@ pub fn try_build(attributes: Attributes) -> Result<()> {
...
       .window_icon_path
-      .unwrap_or_else(|| find_icon(&config, |i| i.ends_with(".ico"), "icons/icon.ico"));
+      .unwrap_or_else(|| {
+        config
+          .bundle
+          .icon
+          .iter()
+          .find(|i| i.ends_with(".ico"))
+          .map(AsRef::as_ref)
+          .unwrap_or_else(|| "icons/icon.ico")
```
Legend-Master:
```suggestion
          .unwrap_or("icons/icon.ico")
```

Legend-Master:
> ```shell
>    Compiling serde v1.0.228
> error: failed to run custom build command for `libc v0.2.178`
> 
> Caused by:
>   process didn't exit successfully: `/target/debug/build/libc-922dd68bc8706b43/build-script-build` (exit status: 1)
>   --- stderr
>   /target/debug/build/libc-922dd68bc8706b43/build-script-build: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /target/debug/build/libc-922dd68bc8706b43/build-script-build)
>   /target/debug/build/libc-922dd68bc8706b43/build-script-build: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /target/debug/build/libc-922dd68bc8706b43/build-script-build)
>   /target/debug/build/libc-922dd68bc8706b43/build-script-build: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /target/debug/build/libc-922dd68bc8706b43/build-script-build)
> warning: build failed, waiting for other jobs to finish...
> error: failed to run custom build command for `serde_core v1.0.228`
> ```
> 
> Is it due to my RR? XD

Not sure what's happening here, definitely not related to this PR though

Tunglies:
There you go

Legend-Master (approved):

FabianLars:
> Not sure what's happening here, definitely not related to this PR though

Yeah no the cache for the android ci just breaks from time to time, iirc some cross issue, can't remember. i'll delete the cache.

PR #14865, opened 2026-01-31
feat: expose wry permission handler API in tauri

F0RLE:
## Description

This PR exposes the expanded `wry` **Permission Handler API** through Tauri, allowing applications to intercept and respond to WebView permission requests across supported backends.

It wires the permission handler through `tauri-runtime`, `tauri-runtime-wry`, the global Tauri builder, and individual webview builders. Until the matching `wry` changes from tauri-apps/wry#1654 are released, `tauri-runtime-wry` points at the merged Wry revision.

### Key Changes
- **Permission types:** Added the expanded `PermissionKind` and `PermissionResponse` types to `tauri-runtime`, mirroring Wry's permission API and platform-specific behavior.
- **Global app hook:** Added `on_permission_request` to `tauri::Builder`, allowing applications to handle permission requests globally for all webviews.
- **Per-webview hook:** Added `on_permission_request` to `tauri::webview::WebviewBuilder` and `tauri::window::WebviewWindowBuilder` for individual webviews and windows.
- **Preserved native defaults:** When neither a per-webview nor an app handler is configured, Tauri does not install a Wry permission handler, preserving the existing platform behavior.

### Permission Resolution Order
1. **Webview-level handler:** Checked first when configured for the individual webview.
2. **App-level handler:** Checked when the webview handler returns `Default` or is not configured.
3. **Wry / OS behavior:** A configured handler returning `Default` delegates to Wry's platform behavior. If no handler is configured, the existing native path is left unchanged.


F0RLE:
Updated this PR after tauri-apps/wry#1654 was merged.

Changes in the latest push:
- merged the current `dev` branch to resolve the dirty state;
- switched the temporary Wry dependency from my fork to `tauri-apps/wry` at the merged Wry commit `e7c84dc2ee460933a4f61bab3c9f1b2d3d33e337`;
- resolved the `Cargo.lock` conflict.

I also ran `cargo check -p tauri-runtime-wry -p tauri-runtime -p tauri --locked` locally and it passes.

Legend-Master:
Let's sign the past commits in this PR first, so we can focus on the actual changes

y4my4my4m:
These changes would significantly help my project. Thank you for working on it. Hopefully it can be merged soon. Cheers.

F0RLE:
Done. I rebuilt the PR on top of the latest `dev` with a signed commit and kept the Wry dependency pointed at the merged `tauri-apps/wry` commit.

Legend-Master:
I think you have quite a few merge conflicts that bought back the old code

F0RLE:
Fixed. The previous rebuild accidentally brought back unrelated `tauri-runtime-wry` event-loop changes from the old conflict resolution.

Legend-Master:
We can ignore the apple clippy for now and fix it in another PR with a `objc2-*` bump maybe paired with a wry bump.

F0RLE:
One thing I noticed while checking this again: `prepare_webview` always installs a permission handler wrapper, even when no per-webview or app handler is configured and plugins do not handle the request.

On WKWebView that changes the existing no-handler behavior: no handler uses `WKPermissionDecision::Grant`, while the wrapper returns `PermissionResponse::Default`, which maps to `WKPermissionDecision::Prompt`.

Is that change intended, or should I preserve the old no-handler behavior before this is merged? I can fix it, but wanted to confirm the expected behavior first.

crates/tauri/src/plugin.rs
```diff
@@ -487,6 +501,33 @@ impl<R: Runtime, C: DeserializeOwned> Builder<R, C> {
     self
   }
 
+  /// Callback invoked when the webview requests a permission.
```
Legend-Master: Let's document that the first non-default return will be used.

crates/tauri/src/webview/mod.rs
```diff
@@ -693,6 +700,52 @@ tauri::Builder::default()
...
+  /// > This handler only triggers for new permission requests. If the user has already
+  /// > allowed or denied a permission persistently within the webview, the browser
+  /// > will use the saved preference instead of calling this handler.
+  ///
+  /// ## Platform-specific:
+  ///
+  /// - **Windows**: Fully supported via WebView2's PermissionRequested event.
+  /// - **macOS / iOS**: Fully supported via WKUIDelegate's requestMediaCapturePermission.
+  /// - **Linux**: Fully supported via WebKitGTK's permission-request signal.
+  /// - **Android**: Experimental support implemented via JNI.
```
Legend-Master: We need to mirror the updated docs from wry

crates/tauri-runtime/src/webview.rs
```diff
@@ -72,6 +74,97 @@ pub enum DownloadEvent<'a> {
   },
 }
 
+/// Permission request response.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Default)]
```
Legend-Master: Also need an update. And we should make a new file for this, let's call it `webview_permissions.rs` for example.

crates/tauri/src/manager/webview.rs
```diff
@@ -310,6 +312,37 @@ impl<R: Runtime> WebviewManager<R> {
         }
       }));
 
+    let label_ = pending.label.clone();
+    let app_manager_ = manager.manager_owned();
+    let permission_request_handler = pending.permission_request_handler.take();
+    pending
+      .permission_request_handler
```
Legend-Master: This is indeed a bit problematic... Let's preserve the old behaviors for now and remove the plugin support, maybe with a TODO comment.

F0RLE: Could we keep plugin support by carrying `Option<PermissionResponse>` internally? `None` would mean that no webview, app, or plugin handler actually handled the request, while `Some(Default)` would mean an explicit default response. The plugin hook would return `None` by default, and `PluginStore` would still select the first non-default response. At the Wry boundary, a final `None` can preserve the legacy behavior (`Allow` on Apple, `Default` elsewhere). That avoids changing existing apps without dropping plugin support. Would this approach work for you?

Legend-Master: Hmm, that can be a bit confusing, but probably fine as long as the documentation mentions that clearly in all places. Also cc @FabianLars for some opinions.

FabianLars: not sure if this is relaly something we should expose to plugins tbh. i think this would add more confusion than actual value.

F0RLE:
Addressed the review feedback in `2bad4b91`:

- removed the plugin permission API;
- only install the Wry handler wrapper when a per-webview or app handler is configured, preserving the existing native behavior otherwise;
- moved the runtime permission types to `webview_permissions.rs` and mirrored the current Wry platform docs;
- updated the WebviewBuilder Android docs;
- added a regression test for the no-handler path.

Local checks:
- `cargo test -p tauri --lib --locked` (59 passed)
- `cargo check -p tauri-runtime-wry -p tauri-runtime -p tauri --locked`
- `cargo clippy -p tauri-runtime-wry -p tauri-runtime -p tauri --locked -- -D warnings`

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -4870,6 +4870,48 @@ You may have it installed on another user account, but it is not available for t
...
+        wry::PermissionKind::AutomaticDownloads => {
+          tauri_runtime::webview::PermissionKind::AutomaticDownloads
+        }
+        wry::PermissionKind::FileSystemAccess => {
+          tauri_runtime::webview::PermissionKind::FileSystemAccess
+        }
+        wry::PermissionKind::Autoplay => tauri_runtime::webview::PermissionKind::Autoplay,
+        wry::PermissionKind::Other => tauri_runtime::webview::PermissionKind::Other,
+        _ => tauri_runtime::webview::PermissionKind::Other,
+      };
```
Legend-Master: let's move this to `webview_permissions.rs`

crates/tauri/src/webview/webview_window.rs
```diff
@@ -434,6 +434,20 @@ tauri::Builder::default()
     self
   }
 
+  /// Defines a closure to be executed when a permission is requested.
```
Legend-Master, line 437: We should mirror the docs from the `webview.rs`

crates/tauri/src/app.rs
```diff
@@ -1788,6 +1792,24 @@ tauri::Builder::default()
     self
   }
 
+  /// Defines a closure to be executed when a permission is requested.
```
Legend-Master, line 1789: Also mirror the docs from `webview.rs`

crates/tauri-runtime-wry/src/lib.rs
F0RLE: Moved both Wry/Tauri permission mappings to crates/tauri-runtime-wry/src/webview_permissions.rs in d7f2c093.

crates/tauri/src/webview/webview_window.rs
F0RLE: Mirrored the permission handler note, platform-specific behavior, and an API-specific example here in d7f2c093.

crates/tauri/src/app.rs
F0RLE: Mirrored the permission handler note, platform-specific behavior, and an app-wide example here in d7f2c093.

crates/tauri-runtime-wry/src/webview_permissions.rs
```diff
@@ -0,0 +1,35 @@
+// Copyright 2019-2024 Tauri Programme within The Commons Conservancy
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: MIT
+
+use tauri_runtime::webview::{PermissionKind, PermissionResponse};
+
+pub(super) fn map_permission_kind(kind: wry::PermissionKind) -> PermissionKind {
```
Legend-Master:
```suggestion
pub(super) fn to_wry_permission_kind(kind: wry::PermissionKind) -> PermissionKind {
```

crates/tauri-runtime-wry/src/webview_permissions.rs
```diff
@@ -0,0 +1,35 @@
...
+    wry::PermissionKind::PointerLock => PermissionKind::PointerLock,
+    wry::PermissionKind::AutomaticDownloads => PermissionKind::AutomaticDownloads,
+    wry::PermissionKind::FileSystemAccess => PermissionKind::FileSystemAccess,
+    wry::PermissionKind::Autoplay => PermissionKind::Autoplay,
+    wry::PermissionKind::Other => PermissionKind::Other,
+    _ => PermissionKind::Other,
+  }
+}
+
+pub(super) fn map_permission_response(response: PermissionResponse) -> wry::PermissionResponse {
```
Legend-Master:
```suggestion
pub(super) fn to_wry_permission_response(response: PermissionResponse) -> wry::PermissionResponse {
```

Legend-Master (commented):
Thanks! Just a bit of nitpicks and we can get this merged after the wry release

crates/tauri-runtime-wry/src/webview_permissions.rs
F0RLE: Renamed the helper and its call site to to_wry_permission_kind in d173e8cd.

crates/tauri-runtime-wry/src/webview_permissions.rs
F0RLE: Renamed the helper and its call site to to_wry_permission_response in d173e8cd.

Legend-Master (approved):
Thanks for the long standing effort!

PR #14871, opened 2026-02-02
fix(tauri-driver): don't inherit stdout from parent

goosewobbler:
**Context:**
I'm working on the official WDIO integration for Tauri and found that `tauri-driver` was corrupting the wdio.log file.

**Issue:**
When tauri-driver spawns the native WebDriver binary (WebKitWebDriver/msedgedriver), the child process inherits the parent's stdout by default. The native WebDriver outputs HTTP protocol data and diagnostic information to stdout, which gets interleaved with tauri-driver's own stdout output.
This causes issues because test frameworks (such as WebDriverIO, Selenium, etc.) capture tauri-driver's stdout to parse responses and coordinate test execution. The extraneous WebDriver output corrupts this communication channel, leading to:
- Parse errors in test framework protocol handling
- Unexpected data in test logs
- Potential test failures due to stdout pollution

**Solution:**
Redirect the native WebDriver binary's stdout to Stdio::null() while preserving stderr inheritance. This ensures WebDriver errors remain visible for debugging, while preventing stdout pollution of the parent process.

Original issue: https://github.com/webdriverio/desktop-mobile/issues/118


FabianLars:
thanks for the PR! can you add a small changefile like this as well please? https://github.com/tauri-apps/tauri/pull/14873/changes (using tauri-driver instead of course).

ah and we also require all commits to be [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) if you can make that happen.

lucasfernog (approved):

PR #14878, opened 2026-02-03
chore(deps): remove unused goblin

Legend-Master:
Remove a left over dependency from #14521

### DIFF
--- a/crates/tauri-bundler/src/error.rs
+++ b/crates/tauri-bundler/src/error.rs
@@ -92,9 +92,6 @@ pub enum Error {
   /// Failed to validate downloaded file hash.
   #[error("hash mismatch of downloaded file")]
   HashError,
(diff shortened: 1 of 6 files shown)


crates/tauri-bundler/src/error.rs
```diff
@@ -92,9 +92,6 @@ pub enum Error {
   /// Failed to validate downloaded file hash.
   #[error("hash mismatch of downloaded file")]
   HashError,
-  /// Failed to parse binary
-  #[error("Binary parse error: `{0}`")]
-  BinaryParseError(#[from] goblin::error::Error),
```
FabianLars, line 97: that's a breaking change

Legend-Master: Oh well, guess we'll see it later then

PR #14880, opened 2026-02-03
feat: re-export `url::form_urlencoded`

Tunglies:
Useful when something likes

```rust
let mut url = match Url::parse(input)?;
url.query_pairs_mut()
                .extend_pairs(form_urlencoded::parse(dirty_params.as_bytes()));
```


FabianLars:
i don't think we should add any further re-exports. We plan to re-iterate all the imports in v3 as it's quite a mess atm.

Issue #14777, opened 2026-01-15
[feat] Android support productFlavors


kupi-kupi:
### Describe the problem

I want to build an application using "productFlavors", to split into DEV and PROD, as was done in Flutter via the --flavor flag.

Are there any plans to add support?

### Describe the solution you'd like

-

FabianLars:
> Are there any plans to add support?

Nothing planned yet. This is the first time this is requested. Can you perhaps expand a bit further on how that would look like / what you need?

kupi-kupi:
@FabianLars 
I want to be able to use "applicationIdSuffix" for an application in development mode, but so that it is created as a "release" and not as a "debug"
Also be able to change the name, add the prefix DEV and change the icon.

More details can be found here: https://developer.android.com/build/build-variants#product-flavors

FabianLars:
cool, thanks. Agreed that this could be a cool feature. I just can't give you an ETA _but_ you can modify the android project so in theory you could add flavors yourself in the meantime.

aiHiNiHao:
I'm also really looking forward to this feature. If we could pass variant parameters through Tauri's CLI tool, it would eliminate the need to open Android Studio.

onehumandev:
I'm currently working on a possible fix for this, as it is an active blocker for further development for us. I'm hoping to have something (at least for a starting point) by the end of the week.

PR #14886, opened 2026-02-03
feat: Add support for Android build variants (feat #14777)

onehumandev:
# Problem

When using `--config` to override the `identifier` field for Android dev/build commands ([per Tauri documentation](https://v2.tauri.app/develop/configuration-files/#extending-the-configuration)), Tauri fails with a "Project directory does not exist" error. The same approach works correctly for iOS, where the bundle identifier is successfully overridden at runtime.

This prevents a common development workflow: having both debug and release versions of an app installed side-by-side on the same device
For example, a developer may need to test a debug build while also having the App Store release version installed.
This is especially important when testing device-specific issues, or crashes that occur only in either debug or release, but not the other, such as bugs related to code obfuscation for release builds.

# Resolution
We have added the `applicationIdSuffix` to the `build.gradle.kts` file when we init the Android project, and default the value to `.debug`.
This matches to what happens when a new Android project is created in Android Studio, and the approach described in the Android docs for [configuring build variants](https://developer.android.com/build/build-variants#build-types).

With these changes, when the user runs `tauri android dev`, the debug build variant will be run.
In practice, this creates an app with the identifier `[base identifier][applicationIdSuffix]`, such as `com.owner.exampleapp.debug`.

When the user runs `tauri android dev --release`, the release build variant is used, which will have only the base identifier, such as `com.owner.exampleapp`.

# Possible further changes
If needed in the future, this approach could be changed to allow more build variants (besides debug and release), and for release variants to have a suffix as well.

However, that is a much larger change that would require reading from the gradle files themselves when trying to start the Android app through tauri-cli.

The changes in the MR account for the vast majority case, which is needing simply one debug and one release variant.

# Dependencies

This is dependent on the changes in this [cargo-mobile2 MR](https://github.com/tauri-apps/cargo-mobile2/pull/505).


crates/tauri-utils/src/config.rs
```diff
@@ -3205,6 +3205,12 @@ pub struct AndroidConfig {
   /// Note that to use this feature, you should remove `/tauri.properties` from `src-tauri/gen/android/app/.gitignore` so the current versionCode is committed to the repository.
   #[serde(alias = "auto-increment-version-code", default)]
   pub auto_increment_version_code: bool,
+
+  /// Application ID suffix to append for debug builds.
+  /// This allows installing debug and release versions side-by-side on the same device.
+  /// Example: ".debug" will make debug builds use "com.example.app.debug" as the application ID.
+  #[serde(alias = "debugApplicationIdSuffix", default = "default_debug_application_id_suffix")]
```
FabianLars, line 3212: 1) alias should be `debug-application-id-suffix` - iirc that's for TOML
2) defaulting to .debug will require devs to rerun android init after updating. The easiest solution would be to just default to None, and if we wanted to add the default to the init templates. Alternatively we could probably inject it into existing gradle configs but i couldn't tell you how from the top of my head.

onehumandev: @FabianLars 
1. My bad; fixed.
2. Apologies if I misunderstand; I was under the assumption that a re-init would be required to get the application suffix regardless of what option we take.
I attempted the solution of changing the init templates, and when re-initting the android project, the application suffix was not correctly populated. Can you confirm the change would look like such in [tauri.config.schema.json](https://github.com/tauri-apps/tauri/blob/b3f2d12b89daefe528e562b93871db62f77973b9/crates/tauri-cli/tauri.config.schema.json#L3083-L3106):
```ts
"debugApplicationIdSuffix": {
  "description": "Application ID suffix to append for debug builds.\n This allows installing debug and release versions side-by-side on the same device.\n Example: \".debug\" will make debug builds use \"com.example.app.debug\" as the application ID.",
  "type": [
    "string",
    "null"
  ],
  "default": ".debug"
}
```
If yes, then when would be required from a dev to get the `.debug` suffix for their Android app after pulling in the latest Tauri version? Wouldn't they still need to re-init the Android app?

lucasfernog:
also needed https://github.com/tauri-apps/cargo-mobile2/pull/509

crates/tauri-cli/src/mobile/android/mod.rs
```diff
@@ -188,6 +190,335 @@ pub fn get_config(
   (config, metadata)
 }
 
+pub(super) fn sync_application_id_suffix(
```
onehumandev: @lucasfernog To confirm: is the purpose of this method to change the build.gradle.kts if the user makes any changes to the applicationIdSuffix in the Cargo.toml, so that running `tauri android init` is not required if any change is needed to the suffix?

crates/tauri-utils/src/config.rs
```diff
@@ -3232,6 +3232,15 @@ pub struct AndroidConfig {
...
+  /// This allows installing build variants side-by-side on the same device.
+  /// Example: `{ "debug": ".debug" }` will make debug builds use "com.example.app.debug" as the application ID.
+  #[serde(
+    alias = "application-id-suffix",
+    default = "default_application_id_suffix"
+  )]
+  pub application_id_suffix: BTreeMap<String, String>,
 }
 
 impl Default for AndroidConfig {
```
velocitysystems, line 3243: Will this change silently mutate existing projects on upgrade? E.g. After upgrading the CLI, it changes the `applicationId` from `com.example.app` to `com.example.app.debug`. This could break project-specific configuration and any external integrations which rely on a stable package name e.g. Firebase, deep-linking.

If this change _does_ mutate existing projects, should we consider applying this to new projects only and leave existing projects untouched?

crates/tauri-cli/src/mobile/android/mod.rs
lucasfernog: yes

crates/tauri-cli/src/mobile/android/dev.rs
```diff
@@ -370,8 +381,17 @@ fn run(
 
   let build_app_bundle = metadata.asset_packs().is_some();
 
+  let application_id_suffix = super::application_id_suffix_for_build_type(
+    tauri_config,
+    if profile == Profile::Debug {
+      "debug"
+    } else {
```
onehumandev: If I'm reading this right, this means that we always set an application id suffix by default either `debug` or `release`; is that accurate?

If so, this is not aligned with standard Android practice. Typically, the release build configuration does not have a suffix by default. 
Out of the box, when creating an Android project with the identifier `com.app.myapp`, the identifiers would be:
* Release: `com.app.myapp`
* Debug: `com.app.myapp.debug`

Is there any particular reason we wouldn't want to follow the same pattern here?

[Android docs](https://developer.android.com/build/build-variants#change-app-id) for more context and usage of suffixes:
```
When you build an APK or AAB for your app, 
the build tools tag the app with the application 
ID defined in the defaultConfig block from the 
build.gradle.kts file, as shown in the following 
example. 

However, if you want to create different versions 
of your app to appear as separate listings on 
Google Play Store, such as a "free" and "pro" 
version, you need to create separate build variants 
that each have a different application ID.
```

lucasfernog: I realized the map for applicationIdSuffix would be more useful for a productFlavor thingy instead of a buildVariant (these concepts seem pretty similar i confused them), I'll revert to debug_application_id_suffix

onehumandev:
@lucasfernog  Tested everything out with both a a fresh and already existent Tauri Android app, and changing the applicationIdSuffix between debug and release configs in the Cargo.toml, and all works great. Thanks very much!

c. @velocitysystems

crates/tauri-cli/src/mobile/android/mod.rs
```diff
@@ -188,6 +190,335 @@ pub fn get_config(
...
+struct BuildTypeBlock {
+  name: String,
+  opening_brace: usize,
+  closing_brace: usize,
+}
+
+fn build_type_blocks(build_gradle: &str, start: usize, end: usize) -> Vec<BuildTypeBlock> {
+  static BUILD_TYPE_RE: OnceLock<regex::Regex> = OnceLock::new();
+
+  let build_type_re = BUILD_TYPE_RE.get_or_init(|| {
```
velocitysystems: Nit: The build-type regex only recognizes two declaration forms:
- `getByName("...")`
- `debug { }`, `release { }`

It misses `create("...")`, `register("...")`, etc — all standard ways to declare a build type in Kotlin DSL. With `create("staging") { ... }`, the regex doesn't match the block at all. Should this be resolved?

lucasfernog:
> @lucasfernog Tested everything out with both a a fresh and already existent Tauri Android app, and changing the applicationIdSuffix between debug and release configs in the Cargo.toml, and all works great. Thanks very much!
> 
> c. @velocitysystems

thanks. I just pushed a change to revert the config back to the debug_application_id_suffix format for consistency - we can add a config that maps product flavor names to a config object in the future to further enhance this

crates/tauri-cli/src/mobile/android/mod.rs
```diff
@@ -188,6 +190,335 @@ pub fn get_config(
...
+      continue;
+    }
+
+    if character == '/' && chars.peek().is_some_and(|(_, next)| *next == '*') {
+      in_block_comment = true;
+      previous = character;
+      continue;
+    }
+
+    if character == '"' || character == '\'' {
```
velocitysystems, line 393: Nit: Does not handle Kotlin raw strings ("""...""") which are common for Proguard rules etc.

velocitysystems (changes requested):

crates/tauri-cli/src/mobile/android/mod.rs
lucasfernog: well since we're only touching debug, the other variants won't be used (i've reverted the change to support arbitrary build variants too - since we'd need to add CLI flags to select a build variant aswell)
future work involves extending this to support product flavors too

crates/tauri-utils/src/config.rs
lucasfernog: yeah, let's apply that change too

crates/tauri-cli/src/mobile/android/mod.rs
lucasfernog: [`a063a37` (this PR)](https://github.com/tauri-apps/tauri/pull/14886/commits/a063a3776379c92228e43d7a9f27c403cce9fdb9)

crates/tauri-utils/src/config.rs
lucasfernog: https://github.com/tauri-apps/tauri/pull/14886/changes/52d908ce2b3090be112acc5ee472690eae3d9f45

onehumandev: Testing this presently; will post findings within a couple minutes

onehumandev: Confirmed, all changes look good.
I created a new Tauri app with current Tauri prod, then ran `tauri android dev` while pointing to these changes, and verified that the `applicationIdSuffix` is not added to the `build.gradle.kts`.
I then deleted src-tauri/gen/android, and reran `tauri android init`, and verified the  `applicationIdSuffix` is defined in `build.gradle.kts`.

One question: with these latest changes, it seems that users will have to add the `"debugApplicationIdSuffix": ".debug"` to the Cargo.toml before re-initting so it would look as such:
```
"android": {
  "versionCode": 4,
  "minSdkVersion": 24,
  "debugApplicationIdSuffix": ".debug"
}
``` 
Is this intended? As in, do we expect the applicationIdSuffix to be populated only if the Cargo.toml describes, and there is no default value provided when initting the Android project?

lucasfernog: > One question: with these latest changes, it seems that users will have to add the "debugApplicationIdSuffix": ".debug" to the TAURI.CONF.JSON before re-initting

that is true unfortunately, having a default value would be a breaking change as mentioned by @velocitysystems :
> This could break project-specific configuration and any external integrations which rely on a stable package name e.g. Firebase, deep-linking.

onehumandev: Sounds good; I definitely agree with not having the suffix changed on `tauri android dev`, just wanted to make sure that we didn't want a default value with `tauri android init` as well.
Thanks!

lucasfernog: some people do not commit their Android projects so having a default on `android init` would still be a breaking change

lucasfernog (approved):

PR #14899, opened 2026-02-05
fix(runtime-wry): fix cross-thread window creation race condition

npiesco:
Fix race condition where `create_window()` called from background threads returns before window is inserted into map, causing `window_handle()` to hang forever.

Add completion channel `(Sender<Result<()>>)` to `CreateWindow` message. Handler signals after `windows.insert()`, caller waits on `rx.recv()` before returning.

Tested on Windows ARM64 and Linux (WSL)


npiesco:
Issue extends to CreateWebView, going to handle this too.

PR #14905, opened 2026-02-06
feat(async_runtime): enable track_caller attribute for async_runtime

### LINKED ISSUE #14904: [feat] async_runtime for better async-task tracing
### Describe the problem

In tokio-console, tracing warnings tasks becomes hard when they warp as location `tauri::async_runtime::spawn`. Have no clue which exactly task lead to.

Some additional track_caller attribute will handle property
```diff
+ #[track_caller]
pub fn block_on<F: Future>(task: F) -> F::Output {
  let runtime = RUNTIME.get_or_init(default_runtime);
  runtime.block_on(task)
}
```

Add on

Image

No add
Image

### Alternatives considered

It could be add on when enable the tracing feature.

### Additional context

_No response_

Tunglies:
Related issues: #14904


Legend-Master:
I have read about this on https://doc.rust-lang.org/reference/attributes/codegen.html#the-track_caller-attribute but I'm still not sure the best practices of this attribute, do we have some pointers for me to look at to understand this better? Or maybe someone else on the team understands it better?

Tunglies:
Hey @Legend-Master, I thinks these is another small reference. In https://github.com/tokio-rs/console/discussions/120. His usage is exactly as same as tauri's.

Tunglies:
Also in the source code of tokio

https://github.com/tokio-rs/tokio/blob/6c03e03898d71eca976ee1ad8481cf112ae722ba/tokio/src/task/builder.rs#L86-L98

crates/tauri/src/async_runtime.rs
```diff
@@ -42,6 +42,7 @@ impl GlobalRuntime {
     }
   }
 
+  #[cfg_attr(feature = "tracing", track_caller)]
```
Tunglies: Better to enable `track_caller` by default to give more flexiability?

Legend-Master: Yeah, doesn't feel like it belongs to the tracing feature

lucasfernog (approved):
nice catch, thanks!

PR #14908, opened 2026-02-07
feat: Allow getting inner PathBuf from SafePathBuf

joshuamegnauth54:
`SafePathBuf` implements `AsRef<Path>` which is ergonomic and useful. However, some APIs take owned `PathBuf`s. This leads to clunky code where the caller has to get a `&Path` from the `SafePathBuf` then take ownership of that path. Ideally, if a user has a `SafePathBuf` and needs a `PathBuf`, they won't need to allocate again just to get the inner `PathBuf` back.


sftse:
Is `SafePathBuf` being used standalone by a consumer? I'm not sure why the type was exposed as `pub`, I can't find public functions that return or take this type except the constructors.

Legend-Master:
> Is `SafePathBuf` being used standalone by a consumer? I'm not sure why the type was exposed as `pub`, I can't find public functions that return or take this type except the constructors.

`tauri-plugin-fs` for an instance uses it, not familiar with the history about this struct though

> https://docs.rs/tauri-plugin-fs/latest/tauri_plugin_fs/enum.SafeFilePath.html

Legend-Master (dismissed):

sftse:
@Legend-Master 
CI is upset about something, do we need to rerun?

joshuamegnauth54:
Hi! Sorry for the late response. I force pushed to add a changes file. Sorry, I didn't know I needed one since the change was trivial. :sweat_smile:

.changes/safepathbuf_into_pathbuf.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri-runtime": patch:enhance
```
Legend-Master:
```suggestion
"tauri": patch:enhance
```

Legend-Master (approved):
Thanks, I will merge this after we got all the patch version bump PRs released and starting to merge for the next minor (2.11)

The change file is needed since this is a user facing change despite being trivial

PR #14911, opened 2026-02-08
fix(cef): adjust inner size on Windows to account for borders

amrbashir:


FabianLars (approved):
tested and seems to work. i need this together with another commit so i'll merge it. if you need to fix something please do so in a follow up commit, sorry :D

PR #14917, opened 2026-02-09
fix: call on_window_destroyed for all window kinds

FabianLars:
This is one possible fix for the "double X required to close app" issue but i feel like it's not the correct approach? @lucasfernog 

also seems to fix the macos exit crash


FabianLars:
still not that familiar with this part of the runtime/cef

lucasfernog (approved):

PR #14925, opened 2026-02-10
feat: add `eval_with_callback` to Webview and WebviewWindow

### LINKED ISSUE #5441: [feat] Call javascript function from Rust (expose wry's eval_with_callback in tauri)
### Describe the problem

I recently have a project that needs to call a certain calculation function of a javascript library and get its result, which needs to be called in the rust code.

I have found the discussion https://github.com/tauri-apps/tauri/discussions/3018.
But the `window.eval("window['function_name']()")` solution can not be called with arguments and can not return data.

I would like to call a javascript function with arguments in rust code and can get its return data.

### Alternatives considered

_No response_

### Additional context

_No response_

lanyeeee:
close #5441

This PR adds a new method `eval_with_callback` to **tauri::webview::Webview** and **tauri::webview::WebviewWindow**, providing a way to evaluate JavaScript in the webview and receive the evaluation result (serialized as a JSON string) via a callback function. 

The feature addresses the long-standing request in #5441

The change is non-breaking, fully backward-compatible, and ready for integration.

#### Implementation Details

- Added `eval_script_with_callback` to the **tauri_runtime::WebviewDispatch** trait.
- Implemented the dispatcher method in **tauri-runtime-wry** by:
  - Introducing a new **WebviewMessage::EvaluateScriptWithCallback** variant (with conditional fields for tracing support, mirroring the pattern used for **WebviewMessage::EvaluateScript**).
  - Sending the message via the existing user message channel.
  - Handling the message in `handle_user_message` by forwarding to `wry::WebView::evaluate_script_with_callback`.
- Updated the mock runtime to record the evaluated script (callback is not invoked, as it's a mock).
- Exposed a convenient public API `eval_with_callback` on both **tauri::webview::Webview** and **tauri::webview::WebviewWindow**.
- Added documentation comments explaining the behavior, including JSON serialization and platform-specific notes.

The implementation closely follows the existing pattern for `eval_script` to ensure consistency, minimal code duplication, and proper tracing span propagation when the `tracing` feature is enabled.

#### Platform Support (Android)

The underlying **wry::WebView::evaluate_script_with_callback** method is documented as **"Android: Not implemented yet."**

In my own real-world testing on Android devices, the method actually works correctly — the callback is invoked with the expected JSON-serialized result, and no errors occur.

However, to respect the wry documentation and avoid promising unsupported behavior,  `eval_with_callback` documentation includes the same warning: **"Android: Not implemented yet."** Users should be aware that while it may work in practice (as verified in testing), it is not officially guaranteed by wry.

#### Tested Example

The following command was tested successfully on both **Windows** and **Android**, correctly returning the evaluated result on both platforms:

```rust
use std::sync::mpsc::channel;

use tauri::{AppHandle, Manager};

#[tauri::command(async)]
fn get_document_html(app: AppHandle) -> Result<String, String> {
    let window = app.get_webview_window("main").unwrap();

    let (tx, rx) = channel::<String>();

    let _ = window.eval_with_callback("document.documentElement.innerHTML", move |res| {
        tx.send(res).unwrap();
    });

    let result = rx.recv().unwrap();

    Ok(result)
}
```


lanyeeee:
There are two test failures in this run, but they appear to be flaky. I pushed the [exact same commit](https://github.com/lanyeeee/tauri/pull/1) to my fork and all tests passed there without issues.

Could you please re-run the GitHub Actions? @FabianLars

FabianLars:
don't worry about it. the android builds regularly fail cause the cache breaks for some reason. meant to delete the caches for another PR but the mobile gh app is horrible x) will do now

crates/tauri/src/webview/mod.rs
```diff
@@ -1901,6 +1901,24 @@ tauri::Builder::default()
       .map_err(Into::into)
   }
 
+  /// Evaluate JavaScript with callback function on this window.
+  /// The evaluation result will be serialized into a JSON string and passed to the callback function.
+  ///
+  /// Exception is ignored because of the limitation on windows. You can catch it yourself and return as string as a workaround.
+  ///
+  /// - **Android:** Not implemented yet.
```
Legend-Master:
```suggestion
  /// Evaluate JavaScript with callback function on this webview.
  /// The evaluation result will be serialized into a JSON string and passed to the callback function.
  ///
  /// Exception is ignored because of the limitation on Windows. You can catch it yourself and return as string as a workaround.
  ///
  /// ## Platform-specific
  ///
  /// - **Android:** Not implemented yet.
```

crates/tauri-runtime/src/lib.rs
```diff
@@ -576,6 +576,18 @@ pub trait WebviewDispatch<T: UserEvent>: Debug + Clone + Send + Sync + Sized + '
   /// Executes javascript on the window this [`WindowDispatch`] represents.
   fn eval_script<S: Into<String>>(&self, script: S) -> Result<()>;
 
+  /// Evaluate JavaScript with callback function on the window this [`WindowDispatch`] represents.
+  /// The evaluation result will be serialized into a JSON string and passed to the callback function.
+  ///
+  /// Exception is ignored because of the limitation on windows. You can catch it yourself and return as string as a workaround.
+  ///
+  /// - **Android:** Not implemented yet.
```
Legend-Master:
```suggestion
  /// Evaluate JavaScript with callback function on the webview this [`WebviewDispatch`] represents.
  /// The evaluation result will be serialized into a JSON string and passed to the callback function.
  ///
  /// Exception is ignored because of the limitation on Windows. You can catch it yourself and return as string as a workaround.
  ///
  /// ## Platform-specific
  ///
  /// - **Android:** Not implemented yet.
```

crates/tauri/src/webview/webview_window.rs
```diff
@@ -2324,6 +2324,20 @@ impl<R: Runtime> WebviewWindow<R> {
     self.webview.eval(js)
   }
 
+  /// Evaluate JavaScript with callback function on this window.
+  /// The evaluation result will be serialized into a JSON string and passed to the callback function.
+  ///
+  /// Exception is ignored because of the limitation on windows. You can catch it yourself and return as string as a workaround.
+  ///
+  /// - **Android:** Not implemented yet.
```
Legend-Master:
```suggestion
  /// Evaluate JavaScript with callback function on this webview.
  /// The evaluation result will be serialized into a JSON string and passed to the callback function.
  ///
  /// Exception is ignored because of the limitation on Windows. You can catch it yourself and return as string as a workaround.
  ///  
  /// ## Platform-specific
  ///
  /// - **Android:** Not implemented yet.
```

Legend-Master (commented):
Looking good from a glance, just some nitpicks on the docs, thanks!

Do you mind also adding a change file?

crates/tauri/src/webview/mod.rs
Legend-Master: Actually, this is implemented for Android in https://github.com/tauri-apps/wry/pull/1133 just like your test results shown, let's remove this part

And you can open a PR to correct the docs on wry if you will, nice catch! 🙃

lanyeeee: Thanks for the review!

I've pushed a follow-up commit that updates the docs as suggested, removes the outdated Android note, and adds a change file

I also opened https://github.com/tauri-apps/wry/pull/1708 for the corresponding wry docs fix

Feel free to take another look whenever you have time

Legend-Master (approved):
Thanks

Issue #14019, opened 2025-08-17
[bug] Tauri can't open multiple windows in multi monitor screen


arifur-rahman-arif:
### Describe the bug

# Tauri Multi-Monitor Window Placement Issue

## Description

In our application, we are attempting to create overlay windows on each available monitor in a multi-monitor setup using Tauri. Despite multiple approaches to position the windows on different monitors, all overlay windows continue to open on the same monitor (typically the primary monitor).

This issue aligns with the existing feature request [#6394](https://github.com/tauri-apps/tauri/issues/6394) for the ability to programmatically assign windows to specific monitors/displays. However, we would like to provide additional context and request assistance or updates on this functionality.

## Environment

- **Operating System**: Linux (Fedora 41, Kernel 6.15.6)
- **Tauri Version**: 2 (as per Cargo.toml)
- **Application**: Algenius (a screenshot capture tool)

## Problem Details

We are developing a screenshot capture tool that requires overlay windows to be displayed on each monitor for user interaction. The goal is to have one overlay window per monitor, positioned and sized according to each monitor's dimensions and location.

### Steps Taken

We've tried several approaches to achieve multi-monitor window placement:
1. **Explicit Positioning**: Setting the window position using monitor coordinates (`window.set_position(PhysicalPosition { x: monitor_pos.x, y: monitor_pos.y })`).
2. **Offset Positioning**: Applying a small offset for each window to encourage placement on different monitors.
3. **Delayed Showing**: Spawning windows hidden, positioning them, and only showing them after a delay to ensure proper placement.
4. **Maximized Property**: Using `.maximized(true)` to attempt full-screen overlays on each monitor.
5. **Size Matching**: Setting window size to match monitor dimensions (`.inner_size(monitor_size.width, monitor_size.height)`).

Despite these efforts, all overlay windows open on the same monitor. We've adjusted delays between window creation and showing (from 50ms to 500ms) to account for initialization timing, but this did not resolve the issue.

### Current Code Snippet

```rust
#[tauri::command]
pub async fn take_screenshot(app: AppHandle, _screenshot_type: String) -> Result<String, String> {
    // ... (monitor detection code) ...
    
    // Create overlay on each monitor
    for (i, monitor) in monitors_vec.iter().enumerate() {
        let overlay_label = if i == 0 { base_overlay_label.to_string() } else { format!("{}_{}", base_overlay_label, i) };
        let m_size = monitor.size();
        let m_pos = monitor.position();
        
[...]

Issue #7376, opened 2023-07-07
[bug] Unable to set x or y window position on Ubuntu production builds / bundle


dreilacadin:
### Describe the bug

We have a Retail POS application which has two windows. 1 main `POS` window and a customer facing `Cart` window. On MacOS, dev and prod builds, setting position for the second window is easy and is working by doing something like this onMount (Were using SvelteKit as our frontend):

```ts
const monitors = await availableMonitors() // get all available monitors

// The cart monitor is the first monitor that has a non-zero x position (monitor facing the customer)
// As this indicates that the monitor is not the primary monitor

let cartMonitor: Monitor | undefined = monitors.find((monitor) => monitor.position.x > 0)

if (!cartMonitor) {
  console.warn("No monitor with x pos greater than 0 detected")
  cartMonitor = monitors[1] // Fallback to selecting the second monitor
}

console.log("Cart monitor detected: ", cartMonitor)

new WebviewWindow("cart", {
  url: "/cart",
  title: "Cart",
  height: cartMonitor.size.height,
  width: cartMonitor.size.width,
  x: cartMonitor.position.x,
  y: cartMonitor.position.y,
})
```

With this, the second window correctly positions to the second monitor. 

The problem is however for Ubuntu, it only works for `dev` mode but not for `production builds`. The 2nd window just doesn't want to move. Even if you do something like

`await appWindow.setPosition(new LogicalPosition(1920, 0));`

It only works on `dev` but again not on `prod builds`. The windows just stays on the first monitor. Has anyone encountered this issue as well on ubuntu and how were you able to address this? (Using Ubuntu 22.04.2 LTS)

### Reproduction

1. Clone the reproduction repo: `git clone https://github.com/dreilacadin/ubuntu-tauri.git`
2. Run `yarn install` or however you'd like to install the dependencies.
3. Make sure to connect a secondary monitor or display.
4. Run `yarn tauri dev`, hopefully you would see that in dev mode, the cart window spawns to the correct secondary monitor.
5. Run `yarn tauri build` to build the production bundle.
6. Install the bundle in your Ubuntu machine.
7. Run the `ubuntu-tauri` app, check if the `Cart` window does indeed stay in your primary monitor.

### Expected behavior

The main window should, as expected, spawn in the primary monitor, then another window (cart) window should spawn to the secondary monitor.

Expected - (Works in Dev mode):

[...]

PR #14926, opened 2026-02-11
feat(window): add command to set fullscreen on specific monitor

Oscar3170:
Add a new command to the window core plugin `set_fullscreen_on_monitor` which receives a position and sets the window as fullscreen on the monitor that contains the position.

Since tao already exposes a way to specify the monitor for the fullscreen function, it was only necessary to expose this parameter in the tauri-runtime-wry and window plugin.

Since monitor names are not unique the new function needs the monitor's position to know which monitor we actually want.

This is the most useful on wayland systems, since the window can't move itself any other way, and this at least gives an option for applications that need to be fullscreen. 

Workaround for issues: #14019 and #7376.

### DIFF
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -1381,6 +1381,7 @@ pub enum WindowMessage {
   SetSizeConstraints(WindowSizeConstraints),
   SetPosition(Position),
   SetFullscreen(bool),
+  SetFullscreenOnMonitor(PhysicalPosition<f64>),
   #[cfg(target_os = "macos")]
   SetSimpleFullscreen(bool),
   SetFocus,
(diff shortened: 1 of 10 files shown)


Oscar3170:
The naming and function parameters could cause confusion, since the other set_fullscreen functions can set the fullscreen state on or off, while this new function only sets fullscreen on.

I figured it would be best to leave the job of turning off the fullscreen state to the regular set_fullscreen function, since the monitor is irrelevant in this case.

Be free to give feedback on if I should change the name or add a bool parameter to the function.

Qodo-Free-For-OSS:
Hi, `SetFullscreenOnMonitor` does nothing when `monitor_from_point` returns `None`, but the command still resolves successfully, so callers cannot detect failure or provide feedback.

**Severity:** action required | **Category:** correctness

**How to fix:** Return error when monitor missing

> ### Issue description
> `set_fullscreen_on_monitor` currently succeeds even when no monitor is found for the provided point, causing a silent no-op.
> 
> ### Issue Context
> The handler branches on `if let Some(monitor)` and otherwise does nothing. Since the dispatch path returns `Ok(())` when the message is delivered, callers can't detect failure.
> 
> ### Fix Focus Areas
> - crates/tauri-runtime-wry/src/lib.rs[3433-3437]
> 
> ### Proposed fix
> Change the handler to emit a warning and/or surface a failure back to the caller. Options:
> - Convert this message to a request/response pattern (like other getters) returning `Result<()>` through a channel, and return `Err(Error::FailedToGetMonitor)` when `monitor_from_point` is `None`.
> - If you want to keep fire-and-forget semantics, at minimum log a warning when `monitor_from_point` is `None` so the failure is observable.

---
*[Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review - free for open-source.*

PR #14938, opened 2026-02-13
fix(tauri): (end, start) should be (start, end)

cuiweixie:

### DIFF
--- a/crates/tauri/src/protocol/asset.rs
+++ b/crates/tauri/src/protocol/asset.rs
@@ -190,7 +190,7 @@ fn get_response(
         // multi-part range header
         let mut buf = Vec::new();
 
-        for (end, start) in ranges {
+        for (start, end) in ranges {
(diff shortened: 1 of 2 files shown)


FabianLars:
did this cause issues to you or how did you find that? i don't think i've seen bug reports that sound related but i can't imagine that 🤔

cuiweixie:
> did this cause issues to you or how did you find that? i don't think i've seen bug reports that sound related but i can't imagine that 🤔 

I am learning rust recently. tauri is the most famous cross platform framework as far as i know.So i am just reading the code. and find the bug by accident.

FabianLars:
ah cool! would you mind also updating examples/streaming.rs ? it's the same code with the same issue.

cuiweixie:
> ah cool! would you mind also updating examples/streaming.rs ? it's the same code with the same issue.

Done.

FabianLars (approved):

PR #14945, opened 2026-02-15
fix(bundler): swap WIX registry search order to prioritize InstallDir

### LINKED ISSUE #14213: [bug] tauri wix template issue with registry search for install location
### Describe the bug

First of all, this is my first issue post and I am still a bit unsure whether this is even considered a bug or not or just a weird edge case.

We recently migrated from tauri v1 to v2 and noticed that even though we might install the app in `C:\Program Files` (using the msi install option), after we update with the tauri updater plugin it automatically gets moved the the corresponding local appdata folder instead and then Program Files app folder gets wiped. This never happened when we used tauri v1. The installed location never changed after an update. We noticed by comparing the wix templates tauri provides for the v1 and v2 versions that the v2 version has the following property

```
        <Property Id="INSTALLDIR">
          <!-- First attempt: Search for "InstallDir" -->
          <RegistrySearch Id="PrevInstallDirWithName" Root="HKCU" Key="Software\\{{manufacturer}}\\{{product_name}}" Name="InstallDir" Type="raw" />

          <!-- Second attempt: If the first fails, search for the default key value (this is how the nsis installer currently stores the path) -->
          <RegistrySearch Id="PrevInstallDirNoName" Root="HKCU" Key="Software\\{{manufacturer}}\\{{product_name}}" Type="raw" />
        </Property>
```

while the v1 has 

```
        <Property Id="INSTALLDIR">
            <RegistrySearch Id="PrevInstallDirReg" Root="HKCU" Key="Software\\{{manufacturer}}\\{{product_name}}" Name="InstallDir" Type="raw"/>
        </Property>
```

I also noticed that the registry key in HKCU still had the (default) key set even when we uninstall the app (perhaps some legacy stuff from v1 or when we temporarily tried out nsis that never got cleaned up?). The problem is that in the v2 version our registry have both the (default) key and the (InstallDir) key in the registry and when our tauri updater plugin updates with the .msi updater it automatically chooses the default key instead. We tried swapping the order of the two registry searches in our own app and it now works as expected. Just thought I would open an issue here but not sure if the current template implementation (https://github.com/tauri-apps/tauri/blob/dev/crates/tauri-bundler/src/bundle/windows/msi/main.wxs) is just expected behaviour or not? If so I can close this issue and just cotinue with our "registry swapped template". I would just assume it would always go with the InstallDir registry key over the default one if both exists?

### Reproduction

_No response_

### Expected behavior

Install location of the tauri app shouldn't change after using the tauri updater.

### Full `tauri info` output

```text
[✔] Environment
    ✔ WebView2: 140.0.3485.81
    ✔ MSVC:
        - Visual Studio Professional 2019
    - node: 22.14.0
    - pnpm: 8.10.5
    - npm: 11.2.0

[-] Packages
    - tauri 🦀: 2.8.5
    - tauri-build 🦀: 2.4.1
    - wry 🦀: 0.53.3
    - tao 🦀: 0.34.3
    - @tauri-apps/api : 2.8.0
    - @tauri-apps/cli : 2.8.4

[-] Plugins
    - tauri-plugin-http 🦀: 2.5.2
    - @tauri-apps/plugin-http : 2.5.2
    - tauri-plugin-dialog 🦀: 2.4.0
    - @tauri-apps/plugin-dialog : 2.4.0
    - tauri-plugin-os 🦀: 2.3.1
    - tauri-plugin-log 🦀: 2.7.0
    - tauri-plugin-deep-link 🦀: 2.4.3
    - @tauri-apps/plugin-deep-link : 2.4.3
    - tauri-plugin-fs 🦀: 2.4.2
    - @tauri-apps/plugin-fs : 2.4.2
    - tauri-plugin-single-instance 🦀: 2.3.4
    - tauri-plugin-shell 🦀: 2.3.1
    - tauri-plugin-notification 🦀: 2.3.1
    - @tauri-apps/plugin-notification : 2.3.1
```

### Stack trace

```text

```

### Additional context

_No response_

veeceey:
Fixes #14213

When both the NSIS default registry key and the named `InstallDir` key exist (e.g. after migrating from NSIS to WIX), the WIX installer was using the wrong install path on update. This happened because WIX evaluates multiple `RegistrySearch` entries within a `Property` and uses the last value found.

The previous order searched for `InstallDir` first and the NSIS default key second, so the default key would win when both existed. This caused apps installed to `C:\Program Files` to silently move to the AppData folder on update via the updater plugin.

This swaps the order so the NSIS default key is searched first (as a fallback), and the named `InstallDir` key is searched second, giving it priority when present.

As @FabianLars noted in the issue thread, swapping the two searches is the correct approach.

### DIFF
--- a/.changes/fix-wix-registry-search-order.md
+++ b/.changes/fix-wix-registry-search-order.md
@@ -1,0 +1,5 @@
+---
+"tauri-bundler": patch:bug
+---
+
+Fix WIX installer registry search order so that the named `InstallDir` key takes priority over the NSIS default key, preventing install location from changing on updates.
--- a/crates/tauri-bundler/src/bundle/windows/msi/main.wxs
+++ b/crates/tauri-bundler/src/bundle/windows/msi/main.wxs
@@ -70,6 +70,10 @@
         <Property Id="ARPURLUPDATEINFO" Value="{{homepage}}"/>
         {{/if}}
 
+        <!-- NOTE: The order of RegistrySearch elements below matters. In WIX, when multiple
+             RegistrySearch elements are listed under a single Property, the LAST successful
+             match wins. We list the NSIS default-key search first and the MSI InstallDir
+             search second so that the MSI-specific path takes priority when both keys exist. -->
         <Property Id="INSTALLDIR">
           <!-- First attempt: Search for the default key value (this is how the nsis installer stores the path) -->
           <RegistrySearch Id="PrevInstallDirNoName" Root="HKCU" Key="Software\\{{manufacturer}}\\{{product_name}}" Type="raw" />
--- a/crates/tauri-bundler/src/bundle/windows/msi/main.wxs
+++ b/crates/tauri-bundler/src/bundle/windows/msi/main.wxs
@@ -71,11 +71,11 @@
         {{/if}}
 
         <Property Id="INSTALLDIR">
-          <!-- First attempt: Search for "InstallDir" -->
-          <RegistrySearch Id="PrevInstallDirWithName" Root="HKCU" Key="Software\\{{manufacturer}}\\{{product_name}}" Name="InstallDir" Type="raw" />
-
-          <!-- Second attempt: If the first fails, search for the default key value (this is how the nsis installer currently stores the path) -->
+          <!-- First attempt: Search for the default key value (this is how the nsis installer stores the path) -->
           <RegistrySearch Id="PrevInstallDirNoName" Root="HKCU" Key="Software\\{{manufacturer}}\\{{product_name}}" Type="raw" />
+
+          <!-- Second attempt: Search for "InstallDir" which takes priority if found (this is how the msi installer stores the path) -->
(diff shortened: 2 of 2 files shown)


FabianLars:
> As @FabianLars noted in the issue thread, swapping the two searches is the correct approach.

I couldn't remember my comment over there at all anymore 😂

FabianLars:
Thanks for the PR! Would you mind [signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) (and force-pushing) your commit? Our branch protection rules require that :)

FabianLars:
Perhaps a small comment in the wix file that explains why the order is like it is would be nice too i think so we don't end up swapping it again 😅

veeceey:
Added a NOTE comment above the RegistrySearch block explaining that in WIX the last successful match wins, so the ordering is intentional. Should help prevent future accidental swaps.

Regarding commit signing — I'll get that sorted and force-push the signed commits shortly.

veeceey:
Commits are now signed and force-pushed. Also added the NOTE comment you suggested about the registry search ordering.

FabianLars:
Awesome, thanks!

FabianLars (approved):

PR #14662, opened 2025-12-16
fix: resource path handles `./` path differently

Legend-Master:
Fix #14659

Also added CI for `tauri-utils`

PR #14959, opened 2026-02-17
refactor: replace `kuchikiki` with `dom_query`

### LINKED ISSUE #14958: `kuchikiki` dependency is dragging in very old crates
I wanted to report / start a discussion about the `kuchikiki` dependency. It is being brought in by `tauri-build` and it brings with it a lot of very old dependencies. For example, `convert_case` version 0.4.0 is over 5 years old: https://crates.io/crates/convert_case/0.4.0

```
convert_case v0.4.0
└── derive_more v0.99.20 (proc-macro)
    └── selectors v0.24.0
        └── kuchikiki v0.8.8-speedreader
            └── tauri-utils v2.8.1
                ├── tauri-build v2.5.3
                │   [build-dependencies]
                │   ├── firezone-gui-client v1.5.11 (/home/thomas/src/github.com/firezone/firezone/rust/gui-client/src-tauri)
                │   └── tauri v2.9.5
                │       ├── firezone-gui-client v1.5.11 (/home/thomas/src/github.com/firezone/firezone/rust/gui-client/src-tauri)
                │       ├── tauri-plugin-dialog v2.6.0
                │       │   └── firezone-gui-client v1.5.11 (/home/thomas/src/github.com/firezone/firezone/rust/gui-client/src-tauri)
                │       ├── tauri-plugin-fs v2.4.5
                │       │   └── tauri-plugin-dialog v2.6.0 (*)
                │       ├── tauri-plugin-opener v2.5.3
                │       │   └── firezone-gui-client v1.5.11 (/home/thomas/src/github.com/firezone/firezone/rust/gui-client/src-tauri)
                │       ├── tauri-plugin-shell v2.3.4
                │       │   └── firezone-gui-client v1.5.11 (/home/thomas/src/github.com/firezone/firezone/rust/gui-client/src-tauri)
                │       └── tauri-specta v2.0.0-rc.21
                │           └── firezone-gui-client v1.5.11 (/home/thomas/src/github.com/firezone/firezone/rust/gui-client/src-tauri)
                ├── tauri-codegen v2.5.2
                │   └── tauri-macros v2.5.2 (proc-macro)
                │       └── tauri v2.9.5 (*)
                ├── tauri-macros v2.5.2 (proc-macro) (*)
                └── tauri-plugin v2.4.0
                    [build-dependencies]
                    ├── tauri-plugin-dialog v2.6.0 (*)
                    ├── tauri-plugin-fs v2.4.5 (*)
                    ├── tauri-plugin-opener v2.5.3 (*)
                    └── tauri-plugin-shell v2.3.4 (*)
                [build-dependencies]
                ├── tauri v2.9.5 (*)
                └── tauri-plugin-fs v2.4.5 (*)
```

This is coming up a lot in our dependency management because we by-default ban duplicate dependencies.

I am not quite sure what the maintenance story around https://github.com/brave/kuchikiki is but it hasn't had a release in almost 3 years and the release tag that Tauri is depending on is this rather odd-looking `v0.8.8-speedreader` version.

How critical is this dependency for Tauri? From my own research, it appears that it is only used for this functionality here: https://github.com/tauri-apps/tauri/blob/6252432f0757d896d7a61819bbff127efac5a156/crates/tauri-cli/src/dev/builtin_dev_server.rs#L130-L155

It appears to me that this should be possible to re-implement quite easily without this dependency?

thomaseizinger:
This PR removes the dependency on `kuchikiki` from the `tauri-cli` and `tauri-utils` crates. It is still being pulled in by `wry` but that is not part of this repository and needs to be dealt with separately.

Resolves: #14958


crates/tauri-utils/src/html.rs
```diff
@@ -158,32 +70,27 @@ pub fn inject_nonce_token(
 }
 
 /// Injects a content security policy to the HTML.
-pub fn inject_csp(document: &NodeRef, csp: &str) {
-  with_head(document, |head| {
-    head.append(create_csp_meta_tag(csp));
+pub fn inject_csp(document: &mut NodeRef, csp: &str) {
+  with_head(document, |doc| {
```
Legend-Master: why is `with_head` taking in a `doc` here? (same for the other places)

crates/tauri-utils/src/html.rs
```diff
@@ -26,127 +21,44 @@ use crate::{
...
-    f(node.as_node())
-  } else {
-    let node = NodeRef::new_element(
-      QualName::new(None, ns!(html), LocalName::from("head")),
-      None,
-    );
-    f(&node);
-    document.prepend(node)
+fn with_head<F: FnOnce(&mut NodeRef)>(document: &mut NodeRef, f: F) {
+  if document.select("head").is_empty() {
```
Legend-Master: We should keep the old behavior

Legend-Master (commented):
Thanks for working on this

Could we preserve the non mut version of those functions? Asking because it is technically going to break people updating `tauri-utils` without bumping `tauri` (we don't fully follow semver on tauri-utils but unfortunately we didn't mark tauri and other dependencies to only take a smaller version range yet)

Also we'll need a change file here and the wry one, and this one will need to wait for the wry one to go through first

thomaseizinger:
> Could we preserve the non mut version of those functions?

That is a bit tricky to do. We'd have to make a wrapper for the document that uses interior mutability. Is that something you want?

crates/tauri-utils/src/html.rs
thomaseizinger: Good question! Sorry for being sloppy here, I should have double-checked this.

Legend-Master:
> We'd have to make a wrapper for the document that uses interior mutability.

Doesn't `NodeRef` from `dom_query` already do that with its nodes?

thomaseizinger:
> > We'd have to make a wrapper for the document that uses interior mutability.
> 
> Doesn't `NodeRef` from `dom_query` already do that with its nodes?

Yep, I just discovered that as well! Sorry for the bad first iteration.

I've revised this to not be mutable but it is still a breaking change because we used to have `kuchikiki`'s types in the API.

thomaseizinger:
Not sure what to do about the CI failures regarding `serde`? Seems unrelated to this PR.

Legend-Master:
> I've revised this to not be mutable but it is still a breaking change because we used to have kuchikiki's types in the API.

Yes, I'm happy as long as old tauri still work with the new tauri-utils though 😂 cc @FabianLars any thoughts on this?

To me I think we should probably mark the entire tauri-utils crate as if it's pre-1.0, meaning 2.5 -> 2.6 will contain breaking changes and adjust our crates to rely on minor versions. Not sure if we should do this before 3.0 though

> Not sure what to do about the CI failures regarding serde? Seems unrelated to this PR.

No worries, that's a cache problem for quite some time now

FabianLars:
> To me I think we should probably mark the entire tauri-utils crate as if it's pre-1.0, meaning 2.5 -> 2.6 will contain breaking changes and adjust our crates to rely on minor versions. Not sure if we should do this before 3.0 though

problem is that we re-export tauri-utils in tauri as tauri::utils

> No worries, that's a cache problem for quite some time now

i removed the cache, next run should work.

Legend-Master:
> problem is that we re-export tauri-utils in tauri as tauri::utils

If we decide to do that, we would also need something like we did for `cookie` and platform webview crates

Or we should re-think about if we should have an internal util crate or something in v3

FabianLars:
yeah, something like this i guess. or we make a different design choice for the utils crate. eg no breaking changes (builder pattern or non_exhaustive?) and no third party types in the api and whatever else we can think of.

but yes, we talked about splitting the utils crate up a few times as well

thomaseizinger:
> > To me I think we should probably mark the entire tauri-utils crate as if it's pre-1.0, meaning 2.5 -> 2.6 will contain breaking changes and adjust our crates to rely on minor versions. Not sure if we should do this before 3.0 though
> 
> problem is that we re-export tauri-utils in tauri as tauri::utils

Yeah that means any breaking change to `tauri-utils` is a breaking change to `tauri`.

If you want to be strict about semver, my suggestion would be:

- Re-export as little as possible from `tauri`, also not `tauri-utils`
- Do not re-export types from other crates
- Version crates like `tauri-utils` independently so you can make breaking changes to it without having to break `tauri` itself. Yes this means that users have to depend on them separately.
- Use a tool like `cargo-semver-checks` to automate this process. Not breaking semver is a very hard thing to do without any tool support.

What is the current policy here? Will this have to wait until Tauri v3?

FabianLars:
> What is the current policy here? Will this have to wait until Tauri v3?

breaking changes in struct fields are fairly common but this here is a bit much for 2.x i think

thomaseizinger:
> > What is the current policy here? Will this have to wait until Tauri v3?
> 
> breaking changes in struct fields are fairly common but this here is a bit much for 2.x i think

Because we are exposing a type with a different API? What if we create a wrapper around `dom_query::Document` that mimics the previous API?

FabianLars:
if we could mimic the api in a way that would not break "old" tauri-codegen and tauri versions that somehow use newer tauri-utils versions that would be ideal. I assume virtually nobody uses the html-manipulation feature directly so i don't really care about anything else (including the use in tauri-cli)

not sure how reasonable/realistic that wish is

thomaseizinger:
We can mimic it such that zero code changes are required when you upgrade both. But you still need to upgrade both because they are just not the same type. If you only upgrade `tauri-utils`, it is not going to work.

I was under the impression that all tauri crates need to be bumped together anyway always.

There is one approach that we can take if you want to be fully backwards-compatible:

- Leave all the previous code in place
- Add an entirely new module that uses `dom_query`
- Add a new feature-flag
- Require a more recent version of `tauri-utils` in all other crates and stop using the old API. That way, the public API here only receives additions and it is thus backwards-compatible.

Let me know what you prefer.

Legend-Master:
Looking at the usage, mimicking the methods can be quite hard to maintain since we'll need to wrap them with our own structs

Feature flags can work, but it's also quite hard, we'll need a new default feature set and enable the old feature for `tauri-utils`, then use `default-features = false` + enable the new feature in dependent crates

If the dependency issue of kuchikiki isn't critical, I am leaning towards delaying this to v3

And in v3, I would really like to bump all tauri crate minor versions together, and make the dependency rely on minors (e.g. `~3.3`), also marking internal dependencies non-stable between minor bumps

(or alternatively, change up the way we release things by bumping major versions and use a 'forth' digit to indicate major feature/change releases (like 301.5.2 to indicate major version 3, breaking change 1, minor 5, patch 2) 😂 https://antfu.me/posts/epoch-semver)

thomaseizinger:
> Feature flags can work, but it's also quite hard, we'll need a new default feature set and enable the old feature for `tauri-utils`, then use `default-features = false` + enable the new feature in dependent crates

I don't think we need to? `tauri-utils` currently doesn't have any default features and the `build` feature can stay as is. We just need to introduce a new `html-manipulation-2` feature that uses `dom_query` and internally, between Tauri crates, we just stop using the `build` feature and list the other ones out explicitly, replacing `html-manipulation` with `html-manipulation-2`.

That is entirely backwards-compatible and allows anyone to remove `kuchikiki` from their dependency tree.

thomaseizinger:
Let me know what you think of the latest version. If I am not mistaken, this is fully backwards-compatible. Feature-naming subject to bike-shedding of course, I just wanted to demonstrate the concept.

FabianLars:
i for one could totally live with this approach :)

thomaseizinger:
> i for one could totally live with this approach :)

Cool! I'll fix the clippy warnings and then test in our repo to ensure the dependencies do go away.

thomaseizinger:
Okay, CI should be passing now. I've patched in this branch into my repo and it seems to work. The thing though is, all plugins will need to be updated to no longer use the `build` feature.

For now, I've unrolled the `build` feature in this PR. Should we introduce a `build-2` (name subject to bike-shedding) that can be used in place of the old one? With Tauri v3, that can then be removed again in favor of `build`.

thomaseizinger:
> The thing though is, all plugins will need to be updated to no longer use the `build` feature.

Actually, it is only the `fs` plugin so maybe we don't need to introduce `build-2`.

thomaseizinger:
> > The thing though is, all plugins will need to be updated to no longer use the `build` feature.
> 
> Actually, it is only the `fs` plugin so maybe we don't need to introduce `build-2`.

I ended up introducing it because there is also a dependency on the `build` feature to enable the `build` module which I've now enabled in both cases (`build` and `build-2`).

Legend-Master:
Juggling feature flags is not exactly easy, it feels to me that this is too much effort for upgrading a non-critical dependency here. And because of the way features work, it is possible for projects to end up with both branches pulled in and compiled.

That being said, I'm not opposed to this if that's not me working on it 😂 Thanks for the effort!

thomaseizinger:
I am a bit confused, do you see this approach merging or not? 

I am happy to figure out the details. As far as I understand, we are only doing this to keep things backwards-compatible and most users are unlikely to use the re-export `html` module anyway.

Legend-Master:
> I am a bit confused, do you see this approach merging or not?

I am not the biggest fan of this approach since it adds in lots of complexity, but I'm happy to take it as long as I am not spending all that time to figure out how to do it (I still kind of have to since I'll need to review this), anyways, if you make this approach work, I'm happy to merge it

> we are only doing this to keep things backwards-compatible and most users are unlikely to use the re-export html module anyway.

Yes, mainly because we didn't lock the `tauri-utils` version before, so if the user updated only `tauri-utils`, they would end up with a broken build

crates/tauri/Cargo.toml
```diff
@@ -167,7 +166,7 @@ glob = "0.3"
 heck = "0.5"
 tauri-build = { path = "../tauri-build/", default-features = false, version = "2.5.5" }
 tauri-utils = { path = "../tauri-utils/", version = "2.8.2", features = [
-  "build",
+  "build-2",
```
thomaseizinger, line 172: So this is a very subtle breaking change for all users that use the `html` module from `tauri-utils` via the `tauri` crate.

Basically, if someone uses `tauri::utils::html::*` today, their code will break until they add a dependency to `tauri-utils` and activate the `build` or `html-manipulation` feature.

Is that acceptable?

thomaseizinger:
All done for a full review @Legend-Master !

crates/tauri-utils/src/html.rs
```diff
@@ -164,6 +164,15 @@ pub fn inject_csp(document: &NodeRef, csp: &str) {
   });
 }
 
+/// Injects a content security policy to the HTML.
+pub fn append_script_to_head(document: &NodeRef, script: &str) {
```
Legend-Master: Where is this used?

crates/tauri-utils/Cargo.toml
```diff
@@ -68,6 +70,14 @@ build = [
   "swift-rs",
   "html-manipulation",
 ]
+build-2 = [
```
Legend-Master, line 75: Let's document this

crates/tauri-codegen/src/context.rs
```diff
@@ -44,13 +46,14 @@ pub struct ContextData {
   pub test: bool,
 }
 
-fn inject_script_hashes(document: &NodeRef, key: &AssetKey, csp_hashes: &mut CspHashes) {
-  if let Ok(inline_script_elements) = document.select("script:not(:empty)") {
+fn inject_script_hashes(document: &Document, key: &AssetKey, csp_hashes: &mut CspHashes) {
+  let inline_script_elements = document.select("script:not(:empty)");
+  if !inline_script_elements.is_empty() {
```
Legend-Master: The for loop already checks for `is_empty`

crates/tauri-utils/src/html2.rs
```diff
@@ -0,0 +1,338 @@
...
+pub fn serialize_node(document: &Document) -> Vec<u8> {
+  document.html().as_bytes().to_vec()
+}
+
+/// Parses the given HTML string.
+pub fn parse(html: String) -> Document {
+  Document::from(html.as_str())
+}
+
+fn with_head<F: FnOnce(NodeRef<'_>)>(document: &Document, f: F) {
```
Legend-Master: We can change this to something like I did in https://github.com/tauri-apps/wry/pull/1671

```rust
  let head = document.head().unwrap_or_else(|| {
    let html = document.html_root();
    let head = document.tree.new_element("head");
    html.prepend_child(&head);
    head
  });
```

crates/tauri-utils/src/html2.rs
```diff
@@ -0,0 +1,338 @@
...
+use crate::pattern::isolation::IsolationJavascriptCodegen;
+use crate::{
+  assets::{SCRIPT_NONCE_TOKEN, STYLE_NONCE_TOKEN},
+  config::{DisabledCspModificationKind, PatternKind},
+};
+
+pub use dom_query::Document;
+
+/// Serializes the document to HTML.
+pub fn serialize_node(document: &Document) -> Vec<u8> {
```
Legend-Master: We should change it to take `NodeRef` or rename it to `serialize_document`

Legend-Master:
Actually, there's one more downside here, the `build2` feature is something people need to migrate to and then migrate back in v3...

thomaseizinger:
> Actually, there's one more downside here, the `build2` feature is something people need to migrate to and then migrate back in v3...

If people migrate at all? I don't see this as something that people should migrate to. It is more of a thing that people can use if they want to clean up their dependency tree. Not everyone cares that much but the people that do care will appreciate the possibility and gladly change up a feature-flag I'd assume.

Legend-Master:
This is more of a burden on the community plugin developers I think, not many people use tauri-utils with build feature in an app presumably. Not a deal breaker, just someone I would like to mention

crates/tauri-utils/src/html.rs
thomaseizinger: Here: https://github.com/tauri-apps/tauri/pull/14959/changes#diff-ab633d2a6a078076838a6bdcb1dc5493ee362a6014bc77951b924b8f9c83f140R131

crates/tauri-codegen/src/context.rs
thomaseizinger: Refactored to a more functional style in https://github.com/tauri-apps/tauri/pull/14959/commits/21db23e0c315848ca53611f6bcb0bd429ce2397e.

crates/tauri-utils/src/html.rs
Legend-Master: That one directly points to `html2` not `html`, no?

crates/tauri-utils/src/html2.rs
thomaseizinger: Renamed functions to `serialize_doc` and `parse_doc` in https://github.com/tauri-apps/tauri/pull/14959/commits/626893aee6c7805f2dc72fb86bef83a2af622ed2.

crates/tauri-utils/Cargo.toml
thomaseizinger: https://github.com/tauri-apps/tauri/pull/14959/commits/5f790d1cef2dae5c6fb1a72a084b548d9225263e

crates/tauri-utils/src/html2.rs
thomaseizinger: https://github.com/tauri-apps/tauri/pull/14959/commits/5197aa37983759e6b8482f3824e350015a540a26

Legend-Master (commented):
Looking good, could you also add a change file?

Also: https://github.com/tauri-apps/tauri/pull/14959/changes#r2868485152

Just a note so we don't forget: we'll need to do a change for `plugin-fs` as well once the new version gets released

thomaseizinger:
> Looking good, could you also add a change file?

Done! Let me know if `minor:enhance` is a good tag.

Legend-Master:
> > Looking good, could you also add a change file?
> 
> Done! Let me know if `minor:enhance` is a good tag.

Probably should be `minor:deps`? Fine for me though

I think should also add a notice about the stability that we might update `dom_query` in minor version bumps

With the new `wry` release, I think we can pull that in now and see the lock file change more clearly

And also https://github.com/tauri-apps/tauri/pull/14959#discussion_r2868485152

crates/tauri-utils/src/html.rs
thomaseizinger: Ah yeah! It is unused on the old one, you are right!

thomaseizinger:
> I think should also add a notice about the stability that we might update `dom_query` in minor version bumps

Where would you like to have that?

Legend-Master:
Maybe at module level? I think it probably makes sense to mark this new module (html2) unstable now with the chance we got

@FabianLars any thoughts?

thomaseizinger:
> I think should also add a notice about the stability that we might update `dom_query` in minor version bumps

The alternative would be to straight away introduce it with a wrapper type to hide `dom_query` from the public API.

Legend-Master:
Just a note an the CI errors, the tauri-utils tests were not run in the CI before #14662

crates/tauri-utils/src/html2.rs
```diff
@@ -0,0 +1,330 @@
...
+pub fn append_script_to_head(document: &Document, script: &str) {
+  let head = ensure_head(document);
+  let script_tag = format!(r#"<script>{script}</script>"#);
+  head.prepend_html(script_tag);
+}
+
+/// The shape of the JavaScript Pattern config
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "lowercase", tag = "pattern")]
+pub enum PatternObject {
```
Legend-Master: @chippers I wonder where do we use it, it seems like we are having it in both here and `crates/tauri/src/pattern.rs` while this one was never used?

Since we're creating a new file with many old things for the compatibility, I want to make sure we don't leave too much unresolved things here that we won't be able to trace the origin using git blame easily

Legend-Master:
Will wait for new wry and dom query versions and then we can merge this (done as of 
6f1f9ca)

And with the chance we got, we might want to also clean up as many `pub` in the new module as possible, for example https://github.com/tauri-apps/tauri/pull/14959#discussion_r2935374537 and mark the module unstable, and lock tauri-utils versions in tauri to minors if possible

crates/tauri-utils/src/html2.rs
thomaseizinger: Looks like it can be removed yeah!

FabianLars:
> And with the chance we got, we might want to also clean up as many pub in the new module as possible, for example https://github.com/tauri-apps/tauri/pull/14959#discussion_r2935374537 and mark the module unstable

Splitting it up into tauri-utils and tauri-internals or something is stupid right?

thomaseizinger:
> > And with the chance we got, we might want to also clean up as many pub in the new module as possible, for example [#14959 (comment)](https://github.com/tauri-apps/tauri/pull/14959#discussion_r2935374537) and mark the module unstable
> 
> Splitting it up into tauri-utils and tauri-internals or something is stupid right?

In my humble opinion, anything "utils" is always a code-smell because it always end up being a catch-all bucket for all kinds of functionality and causes an entangled mess. Of course, it comes with its own bit of complexity to create a `tauri-html` or other things but I do think that esp. when it comes to semver, it is worth to split code into cohesive buckets. That way, version bumps are a lot less common and have less of a ripple effect through the ecosystem. With more cohesive crates, the reason why one package depends on another is solely for the functionality it wants from it and not thus is not impacted if something else in its API changes that it is not even using.

So no, I don't think it is a stupid idea. I think it is actually a great idea IF the boundaries of the crates are designed around their functionality, i.e. `tauri-html` etc.

FabianLars:
yeah i think we can all agree that tauri-utils contains way too much 😅 but perhaps it's not the time to split it up and simply wait for v3 (iirc getting the config part out of utils was also very tricky?)

Legend-Master:
An internal crate sounds good to me, but in the other hand, I never even realized that the tauri-utils crate is actually a public crate (and tauri even re-exports it) until recently say half a year ago 😂

I don't know how common is a non-semver following internal crate practice in the rust world, but it seems to be something quite common for the npm packages, like `docusaurus` having [@docusaurus/utils](https://npmx.dev/package/@docusaurus/utils), parcel having [@parcel/utils](https://npmx.dev/package/@parcel/utils)

In the other hand, some other packages/crates choose to 0.x for internal packages/crates that does follow semver

Both approaches have their own trade offs though

> iirc getting the config part out of utils was also very tricky?

Can't remember the details anymore, I think it was mainly about the ACL and other things being shared by both the config and tauri APIs like `TitleBarStyle`

FabianLars:
i'd still follow semver for external crates just to not make cargo's life harder than necessary but imo we don't need to follow tauri's version e.g. we can keep it in 0.x or like chrome on v145+ 🤷 

That would also make _some_ de-duplication possible assuming that not every single change we have there is breaking

thomaseizinger:
> An internal crate sounds good to me, but in the other hand, I never even realized that the tauri-utils crate is actually a public crate (and tauri even re-exports it) until recently say half a year ago 😂

I think that is the root of the issue here. utils crates are fine as long as you are not making them part of your public API. They are "public" in the sense of everyone can depend on it but the versioning doesn't matter then, right? It wouldn't matter if `tauri-utils` is version 37 as long as it is only used internally. That is what `tokio` and `tokio-util` are doing for example and I think that is a very sensible approach.

lucasfernog (dismissed):
nice one!

FabianLars (dismissed):

FabianLars (approved):

crates/tauri-utils/src/html2.rs
Legend-Master: Since chippers didn't respond in time, I have gone ahead and removed these in 91888fb, we can always add them if needed

Legend-Master (approved):
Thanks again for the patience @thomaseizinger !

PR #14967, opened 2026-02-18
fix(deps): disable default-features in objc2 crates

montyc1999:
We are mistakenly using default features in the crates: `objc2-foundation`, `objc2-app-kit`, and `objc2-web-kit`.

Also, `tauri-runtime-wry` doesn't need to depend on `objc2-foundation` at all: `objc2_foundation::MainThreadMarker` is deprecated and was moved to `objc2::MainThreadMarker`.

Here's the before/after result of `CARGO_PROFILE_RELEASE_LTO=fat cargo llvm-lines -p tauri --example helloworld --release`:

Before:
```
  Lines                  Copies                Function name
  -----                  ------                -------------
  2427243                107287                (TOTAL)
```

After:

```
  Lines                  Copies               Function name
  -----                  ------               -------------
  2234525                93674                (TOTAL)
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5072,36 +5072,8 @@ checksum = "5906f93257178e2f7ae069efb89fbd6ee94f0592740b5f8a1512ca498814d0fb"
 dependencies = [
  "bitflags 2.7.0",
  "block2 0.6.0",
- "libc",
  "objc2 0.6.0",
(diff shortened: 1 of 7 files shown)


FabianLars:
good catch, thanks!

FabianLars (approved):

PR #14977, opened 2026-02-20
feat(cli): optional banner generation for Android TV target

setoelkahfi:
I tried to initiate [the discussion here](https://github.com/orgs/tauri-apps/discussions/14889), so I tried this way.

This addition is not intrusive, as in it's an optional parameter. Background is that my users are mostly on `Android TV` and Tauri has been really great platform to share codebase among my apps.

### DIFF
--- a/crates/tauri-cli/src/icon.rs
+++ b/crates/tauri-cli/src/icon.rs
@@ -48,11 +48,19 @@ enum AndroidIconKind {
   Rounded,
 }
 
+struct BannerEntry {
+  name: String,
+  width: u32,
+  height: u32,
+  out_path: PathBuf,
+}
+
 struct AndroidEntries {
   icon: Vec<(PngEntry, AndroidIconKind)>,
   foreground: Vec<PngEntry>,
   background: Vec<PngEntry>,
   monochrome: Vec<PngEntry>,
+  banner: Vec<BannerEntry>,
 }
 
 #[derive(Deserialize)]
@@ -63,6 +71,7 @@ struct Manifest {
   android_fg: Option<String>,
   android_monochrome: Option<String>,
   android_fg_scale: Option<f32>,
+  android_tv_banner: Option<String>,
 }
 
 #[derive(Debug, Parser)]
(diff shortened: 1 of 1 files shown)


FabianLars:
looking at my tv i'm not sure the text input parameter is really the right call. barely any (serious looking apps) use it. Maybe we can accept a banner-ish icon input instead? So for example Netflix, one icon for the symbols with just the `N` and one for banners with `Netflix` spaced out? 🤔

setoelkahfi:
> looking at my tv i'm not sure the text input parameter is really the right call. barely any (serious looking apps) use it. Maybe we can accept a banner-ish icon input instead? So for example Netflix, one icon for the symbols with just the `N` and one for banners with `Netflix` spaced out? 🤔

That's a nice addition.

The issue is with the standard app review on PlayStore for Android TV. The script provides a predictable result based on [the guideline](https://developer.android.com/design/ui/tv/guides/system/tv-app-icon-guidelines#banner) from Google and I tested it on a production app. I tried submitting my app with only an icon as the banner, which is similar to Netflix spaced out icon option, and the app didn't pass the review.

FabianLars:
Perhaps not the best example, they're probably not as strict for netflix and similar large apps x)

i have to admit though that i didn't check the code before my earlier comment so i thought we'd write the text ourselves which sounds horrible (cause how to define font size, weight, color, etc). Still though, if it's an image already, might as well take the "full" banner image as an input and not just the text so that devs/designers can configure the size and spacing and whatever themselves. There's imho not much the code here does to "help" with the guidelines that they couldn't do themselves just as easy.

setoelkahfi:
I see what you're coming from. I don't have a dedicated designer, that's important context. So, the script is my way to automate the creation from the existing app-icon that the `cargo tauri icon` command already use.

if we go with your suggestion, doesn't mean that it should be a dedicated command? or what do you think the command would be?

FabianLars:
> if we go with your suggestion, doesn't mean that it should be a dedicated command? or what do you think the command would be?

essentially same as it is now but instead of a path to the text you provide a path to the full banner icon. Another PR will also add json file inputs.

setoelkahfi (commented):
@FabianLars been long time since I synced with the dev branch, could you take a look again?

PR #14986, opened 2026-02-22
fix(tauri-utils): sort csp/plugin/header config maps during codegen so `generate_context!` is deterministic

### LINKED ISSUE #14978: [bug] `generate_context!` macro doesn't sort CSP object keys, breaks incremental cache
### Describe the bug

When we use a CSP configuration like this:

```json
"csp": {
  "img-src": "'self'",
  "style-src": "'self' 'unsafe-inline'",
  "default-src": "'self'",
  "connect-src": "ipc: http://ipc.localhost"
}
```

The `generate_context!` macro produces something like this, but the lines are randomly ordered:

```rust
Csp::DirectiveMap({
    let mut map = HashMap::new();
    map.insert("connect-src".into(), CspDirectiveSources::Inline("ipc: http://ipc.localhost".into()));
    map.insert("style-src".into(), CspDirectiveSources::Inline("'self' 'unsafe-inline'".into()));
    map.insert("img-src".into(), CspDirectiveSources::Inline("'self'".into()));
    map.insert("default-src".into(), CspDirectiveSources::Inline("'self'".into()));
    map
})
```

This breaks the incremental compilation cache.

### Reproduction

Minimal reproduction here:
https://github.com/montyc1999/tauri-nondeterministic-csp-order-demo

Note that I set `codegen-units = 1` to better demonstrate the issue.
But the issue is much greater in my actual project with the default `codegen-units = 256`, and  adds several seconds to compilation time.

### Expected behavior

The `generate_context!` macro should be deterministic.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 14.8.1 arm64 (X64)
    ✔ rustc: 1.93.1 (01f6ddf75 2026-02-11)
    ✔ cargo: 1.93.1 (083ac5135 2025-12-15)
```

### Stack trace

```text

```

### Additional context

This issue might also affect other configuration objects that pass a HashMap to `map_lit`, like `PluginConfig` and `HeaderSource`.

montyc1999:
Fixes https://github.com/tauri-apps/tauri/issues/14978

It would be nicer to convert these HashMaps to BTreeMaps, but that would be a breaking change.

### DIFF
--- a/.changes/fix-nondeterministic-context-codegen.md
+++ b/.changes/fix-nondeterministic-context-codegen.md
@@ -1,0 +1,5 @@
+---
+'tauri-utils': 'patch:bug'
+---
+
+Sort csp/plugin/header configs when generating HashMap constructors so that `generate_context!` is deterministic.
--- a/.changes/fix-nondeterministic-context-codegen.md
+++ b/.changes/fix-nondeterministic-context-codegen.md
@@ -3,3 +3,5 @@
 ---
 
 Sort csp/plugin/header configs when generating HashMap constructors so that `generate_context!` is deterministic.
+
+See: https://github.com/tauri-apps/tauri/issues/14978 for more information
--- a/crates/tauri-utils/src/config.rs
+++ b/crates/tauri-utils/src/config.rs
@@ -3842,6 +3842,9 @@ mod build {
           quote!(#prefix::Policy(#policy.into()))
         }
         Self::DirectiveMap(list) => {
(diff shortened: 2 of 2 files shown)


Legend-Master:
Nice catch, could you also add some comments about the reason for this and a todo about changing it them to `BTreeMap`s?

montyc1999:
@Legend-Master sure, added. How's that?

Legend-Master (approved):
Thanks!

PR #14993, opened 2026-02-24
refactor: use `html` module from `tauri-utils`

thomaseizinger:
As a first step towards removing the outdated `kuchikiki` dependency, we remove them from the `tauri-cli` crate and instead utilize the existing `html` module from `tauri-utils`.

Related: #14959


thomaseizinger:
@Legend-Master Could we merge this so CI doesn't have to constantly get approved in #14959 ?

Legend-Master:
Merging this means we will need to bump a minor version of tauri-utils which I don't really want right now, it is easier for us to release patch versions between minor versions without some of the minor bump changes in the tree

Legend-Master:
Closing in favor of #14959

PR #14997, opened 2026-02-25
fix(core): replace unsafe unwraps with proper error handling

RainyNight9:

### DIFF
--- a/crates/tauri/src/ipc/protocol.rs
+++ b/crates/tauri/src/ipc/protocol.rs
@@ -65,9 +65,14 @@ pub fn get<R: Runtime>(manager: Arc<AppManager<R>>) -> UriSchemeProtocolHandler 
               span.record(
                 "request",
                 match &request.body {
-                  super::InvokeBody::Json(j) => serde_json::to_string(j).unwrap_or_else(|e| e.to_string()),
-                  super::InvokeBody::Raw(b) => serde_json::to_string(b).unwrap_or_else(|e| e.to_string()),
-                },
+                  super::InvokeBody::Json(j) => serde_json::to_string(j),
+                  super::InvokeBody::Raw(b) => serde_json::to_string(b),
+                }
+                .unwrap_or_else(|e| {
+                  let msg = format!("failed to serialize request body: {e}");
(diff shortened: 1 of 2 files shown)


sftse:
I dont think silencing potential errors by returning invalid json is proper error handling.

RainyNight9:
> I dont think silencing potential errors by returning invalid json is proper error handling.

Thanks for the feedback! You are right that returning a raw error string creates invalid JSON and might silence the error.

I've updated the code to:

1. Log the error using tracing::debug! so it's visible in the logs.
2. Return a valid JSON string (using serde_json::Value::String ) containing the error message, ensuring the response remains valid JSON.
Please let me know if this looks good to you.

sftse:
Is there conscious thought involved with this PR?

crates/tauri/src/window/mod.rs
```diff
@@ -333,7 +333,11 @@ tauri::Builder::default()
     let pending_webview = webview.into_pending_webview(self.manager, &self.label)?;
     let window = self.build_internal(Some(pending_webview))?;
 
-    let webview = window.webviews().first().unwrap().clone();
+    let webview = window
+      .webviews()
+      .first()
+      .cloned()
+      .ok_or(crate::Error::WebviewNotFound)?;
```
FabianLars, line 340: the webview was created in 333/334 so there should be no way for the unwrap to panic.

crates/tauri/src/ipc/protocol.rs
```diff
@@ -65,9 +65,14 @@ pub fn get<R: Runtime>(manager: Arc<AppManager<R>>) -> UriSchemeProtocolHandler
               span.record(
                 "request",
                 match &request.body {
-                  super::InvokeBody::Json(j) => serde_json::to_string(j).unwrap(),
-                  super::InvokeBody::Raw(b) => serde_json::to_string(b).unwrap(),
-                },
+                  super::InvokeBody::Json(j) => serde_json::to_string(j),
+                  super::InvokeBody::Raw(b) => serde_json::to_string(b),
```
FabianLars, line 69: this was already validated in parse_invoke_request

PR #14999, opened 2026-02-25
chore: fix clippy warnings

RainyNight9:

### DIFF
--- a/crates/tauri-bundler/src/bundle/macos/app.rs
+++ b/crates/tauri-bundler/src/bundle/macos/app.rs
@@ -136,7 +136,7 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
         if matches!(e, NotarizeAuthError::MissingTeamId) {
           return Err(e.into());
         } else {
-          log::warn!("skipping app notarization, {}", e);
+          log::warn!("skipping app notarization, {e}");
         }
       }
     }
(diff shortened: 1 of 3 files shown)


crates/tauri-bundler/src/bundle/macos/app.rs
```diff
@@ -136,7 +136,7 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
         if matches!(e, NotarizeAuthError::MissingTeamId) {
           return Err(e.into());
         } else {
-          log::warn!("skipping app notarization, {}", e.to_string());
+          log::warn!("skipping app notarization, {}", e);
```
FabianLars: if to_string can be removed, does this work as well?

```suggestion
          log::warn!("skipping app notarization, {e}");
```

RainyNight9: > if to_string can be removed, does this work as well?

Yes, it works! Since Rust 1.58, we can capture arguments from the surrounding scope. I've updated the code.

FabianLars (approved):

FabianLars:
Ah, one more thing. We have branch protection rules in place that requrie _all_ commits of a PR to be signed, can you perhaps set that up and force-push the signed commits? https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

RainyNight9:
> Ah, one more thing. We have branch protection rules in place that requrie _all_ commits of a PR to be signed, can you perhaps set that up and force-push the signed commits? https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

done

FabianLars (approved):

PR #15010, opened 2026-02-27
feat(cef): add WebviewBuilder::on_address_change

david-crabnebula:
The current `on_navigation` callback only fires before full page loads. CEF `DisplayHandler::on_address_change` fires for ALL URL changes including `pushState`


FabianLars:
can you remember if tauri+wry behaves like this?

david-crabnebula:
> can you remember if tauri+wry behaves like this?

good question, probably, because navigation to SPA like nextjs that use push state will not work (or at least the `on_navigation` hook would never fire, only on the index)

lucasfernog:
> can you remember if tauri+wry behaves like this?

wry doesn't treat SPA route changes as navigation

Issue #15003, opened 2026-02-25
[bug] `WindowsStore` can be called outside of main thread, causing crashes due to previous assumptions


chippers:
### Describe the bug

The `handle_event_loop` and `handle_user_message` functions use `borrow()` and `borrow_mut()` due to previous assumptions that the `WindowsStore` is only callable from the main thread. Not only is that not true anymore, but that same assumption causes an unsound implementation for `Sync` as seen in #14801. After the unsoundness is fixed, the assumption that `WindowsStore` can only be called from the main thread and therefore can do a panicking `.borrow()`/`.borrow_mut()` may need to be changed.

Originally it was a Mutex, and moved to an `Arc<RefCell` to a `Rc<RefCell` to a `Arc<RefCell` with an unsound `Sync` impl. Moving it back to a mutex or a RwLock would be the most straighforward move.

### Reproduction

https://github.com/chippers/tauri-runtime-wry-naughty-refcell-plugin

### Full `tauri info` output

```text
[✘] Environment
    - OS: Mac OS 26.3.0 arm64 (X64)
    ✔ Xcode Command Line Tools: installed
    ✘ Xcode: not installed!
    ✔ rustc: 1.93.1 (01f6ddf75 2026-02-11)
    ✔ cargo: 1.93.1 (083ac5135 2025-12-15)
    ✔ rustup: 1.28.2 (e4f3ad6f8 2025-04-28)
    ✔ Rust toolchain: stable-aarch64-apple-darwin (default)
    - node: 25.6.1
    - pnpm: 10.29.3
    - yarn: 1.22.22
    - npm: 11.9.0
    - deno: deno 2.6.10

[-] Packages
    - tauri 🦀: 2.10.2
    - tauri-build 🦀: 2.5.5
    - wry 🦀: 0.54.2
    - tao 🦀: 0.34.5
    - @tauri-apps/api  ⱼₛ: not installed!
    - @tauri-apps/cli  ⱼₛ: 2.10.0

[-] Plugins

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../src
```

### Stack trace

```text
thread 'main' (1726941) panicked at /Users/chip/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tauri-runtime-wry-2.10.0/src/lib.rs:4268:31:
RefCell already mutably borrowed
stack backtrace:
[...]

PR #15020, opened 2026-03-03
`tauri-runtime-wry` windows store `RefCell` refactor

chippers:
Draft as this is experimental, I don't know if this is the proper solution.

Related to #15003, this attempts to fix the underying panic by moving all calls to `try_` variants and adding logs where it cannot fail. The newtype for `WindowsStore` is written using callbacks to help reduce the changes of a borrow guard lasting longer than it should by accident.

I was originally going to write this with a Mutex/RwLock, but since I was triggering it with only a main thread (based on my debugging) the same faulty logic should be able to create a deadlock.  At least this gives insight when borrowed and not


sftse:
The underlying issue was fixed with merging of #14805

PR #15029, opened 2026-03-04
Revert "Apply Version Updates From Current Changes"

Legend-Master:
Reverts tauri-apps/tauri#14897

It can't seem to go through `cargo audit` because of old `time` crate (CI didn't catch it because it re-generated the lock file, see #15030), but since we can't pull in the new version in the repo due to a high MSRV


FabianLars:
if only #15030 is needed we don't need to revert. it will automatically retry to publish next commit

Legend-Master:
Ah, right, we'll also need to add the time error to ignore list

FabianLars:
might as well do that in 15030 too imo

PR #15030, opened 2026-03-04
fix(ci): bump rustsec/audit-check to v2 and ignore time audit

Legend-Master:
Use the change made in https://github.com/rustsec/audit-check/pull/20 to prevent things like
https://github.com/tauri-apps/tauri/actions/runs/22661517484/job/65685733208

Also adding RUSTSEC-2026-0009 to the ignore list

> time crate can't be updated in the repo because of MSRV, users are unaffected

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -6,4 +6,6 @@ ignore = [
   "RUSTSEC-2020-0095",
   # proc-macro-error is unmaintained
   "RUSTSEC-2024-0370",
+  # time crate can't be updated in the repo because of MSRV, users are unaffected
+  "RUSTSEC-2026-0009",
 ]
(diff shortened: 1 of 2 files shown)


Legend-Master:
Looking at where this was ran

> https://github.com/jbolda/covector/blob/eff3ad5a4e0aebc701ca8cefa8afd9906b21eb43/packages/covector/src/init.ts#L470

The audit run was not known to me, do we also have this check for the js? (doesn't seem like it) Asking since our js one is also failing because of eslint and rollup plugin terser

FabianLars:
puhh, don't think so? pretty sure i published a few times with failing js audits

FabianLars (approved):

PR #15031, opened 2026-03-04
chore: update js dependencies to fix audit

Legend-Master:
Don't understand why `pnpm update typescript-eslint` didn't work, went ahead did a `pnpm update --recursive` 🤷‍♂️

Also removed outdated overrides, moved `onlyBuiltDependencies` to `pnpm-workspace.yaml`

### DIFF
--- a/crates/tauri-schema-worker/package.json
+++ b/crates/tauri-schema-worker/package.json
@@ -8,6 +8,6 @@
     "dev": "wrangler dev"
   },
   "devDependencies": {
-    "wrangler": "^4.61.1"
+    "wrangler": "^4.70.0"
   }
 }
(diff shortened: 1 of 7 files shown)


FabianLars:
> Don't understand why pnpm update typescript-eslint didn't work, 

the pnpm update command genuinely sucks and is my biggest (and only?) issue with pnpm. That you cannot simply update nested dependencies drives me crazy.

FabianLars:
for `serialize-javascript` i added an overwrite in tauri-action as well but we may wanna check if we can bully @jbolda into updating it in covector

Legend-Master:
Just looked at the `serialize-javascript` v6 to v7 breaking change, and it seems to be just a node version requirement bump, likely safe to override

FabianLars (approved):

PR #15038, opened 2026-03-05
refactor(macos-sign): use base64 crate instead of Command

amrbashir:
Avoids the mentioned issue by stripping whitespace before decoding. 

This fixes a niche case for me, where I have base64 installed on macOS from GNU coreutils, which doesn't use the same flags and builds were failing for me.

### DIFF
--- a/crates/tauri-macos-sign/src/lib.rs
+++ b/crates/tauri-macos-sign/src/lib.rs
@@ -64,6 +64,8 @@ pub enum Error {
   },
   #[error("failed to encode DER: {error}")]
   FailedToEncodeDER { error: std::io::Error },
+  #[error("failed to decode base64 certificate: {0}")]
+  Base64Decode(base64::DecodeError),
   #[error("certificate missing common name")]
   CertificateMissingCommonName,
   #[error("certificate missing organization unit for common name {common_name}")]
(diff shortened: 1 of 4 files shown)


zubinqayam (approved):
K01

zubinqayam (approved):
Done33

zubinqayam (approved):
Done55

Legend-Master (dismissed):
Not sure if we need a change file?

crates/tauri-macos-sign/src/lib.rs
```diff
@@ -329,36 +331,23 @@ impl NotarytoolCmdExt for Command {
...
-      .arg(&src_path)
-      .arg("-o")
-      .arg(out_path)
-      .piped(),
-    "failed to decode certificate",
-  )
-  .map_err(|error| Error::CommandFailed {
-    command: "base64 --decode".to_string(),
+  // strip whitespace before decoding
+  let cleaned: String = input.chars().filter(|c| !c.is_ascii_whitespace()).collect();
```
Legend-Master, line 342: Not exactly sure the behavior with the OS base64, but I think we should add a bit more context here in the comments

amrbashir: I don't have enough information about what was the previous problem exactly, so maybe @lucasfernog can shed some light on it

lucasfernog: this is just tech debt from the days David wrote this - like 5 years ago?? omg i'm old

lucasfernog (dismissed):

PR #15061, opened 2026-03-09
feat(ios): add --no-sign and --archive-only flags to ios build

### LINKED ISSUE #14940: [feat] Add a way to build iOS archives without signing
### Describe the problem

It would be good practice to have a way for CI/CD to build iOS `.xcarchive`s without signing them, so that we can convert them to `.ipa` at a later stage and sign them offline (ourselves as developers or single users via AltStore et cetera). Currently, one can build via a combination of `tauri ios build --open` and launching `xcodebuild` from another terminal:

```
xcodebuild archive \
          -project gen/apple/${NAME}.xcodeproj \
          -scheme "${NAME}"_iOS \
          -archivePath gen/apple/build/App.xcarchive \
          -configuration Release
          CODE_SIGNING_REQUIRED=NO 
          CODE_SIGNING_ALLOWED=NO
```

Something like `tauri ios build --archive --no-codesign`?

### Alternatives considered

Making two concurrent jobs? Seems hard to do? I don't know

### Additional context

Intended for use in GitHub Actions etc.

PixPMusic:
closes https://github.com/tauri-apps/tauri/issues/14940

Of note, `cargo test` is throwing a failed test, but it seems to be due to a snapshot change outside of my work? clippy is passing fine.

This adds two flags:

* `--no-sign`: uses xcodebuild archive to output a .xarchive, then pulls the .app out of that and produces an IPA for use in third-party distribution (AltStore/SideStore)
* `--archive-only`: uses xcodebuild archive to output a .xarchive, following your signing configuration. Can be combined with `--no-sign` as well

Places I'm unable to check for regression: App Store Connect use should be unaffected, but I'm unable to verify as I don't have an Apple Developer subscription.


PixPMusic:
I am going to take note of the following in SideStore:

https://github.com/LiveContainer/SideStore/blob/bff192be4e2b99e0e7cd717ba3d1871d3a7b2136/Makefile#L252-L268

I may want to consider the impact of entitlements in tauri projects, both for things like storage, but also for custom native extensions requiring advanced platform-specific configuration.

lucasfernog (dismissed):

PR #15064, opened 2026-03-10
fix: ohos build error

richerfu:
When i try to run tauri in ohos, i'll get the crash. I think this change is important. cc @frankplus @FabianLars @lucasfernog


FabianLars:
if we keep using git deps for longer we should look into pinning them to a commit hash but let's go with this for now.

FabianLars (approved):

PR #15101, opened 2026-03-12
fix(ios): fix deadlock during Swift plugin command handling

PrinceTumby:
Hi!

I was developing a Tauri plugin in Swift for iOS, and after changing a number of commands to be `async`, I noticed my entire app started freezing.
After some debugging, it seemed like there was a deadlock happening between two threads running `tauri::plugin::run_command` and its nested `plugin_command_response_handler` function.

I haven't had time to do a full investigation (it's a little difficult to reproduce with a simple example), but this fixes the issue by just making sure the `PENDING_PLUGIN_CALLS` mutex lock is held for just long enough to remove the handler, whereas previously Rust was implicitly extending the lifetime of the lock to cover the entire `if let` block.

I didn't immediately see any issues talking about this, but happy to tag them if there are any.

### DIFF
--- a/crates/tauri/src/plugin/mobile.rs
+++ b/crates/tauri/src/plugin/mobile.rs
@@ -373,12 +373,12 @@ pub(crate) fn run_command<R: Runtime, C: AsRef<str>, F: FnOnce(PluginResponse) +
         CStr::from_ptr(payload)
       };
 
-      if let Some(handler) = PENDING_PLUGIN_CALLS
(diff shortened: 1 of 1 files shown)


FabianLars (approved):

FabianLars:
my brain is foggy lately but i'm having a hard time seeing how this would behave differently - i mean, i see what you're thinking, i just can't really believe rust does this - anyway, doesn't hurt so i merged it anyway. Then we'll see over time if it really helped :)

PR #15105, opened 2026-03-13
fix: proper transparency on macOS and Windows

amrbashir:


crates/tauri-runtime-cef/src/cef_impl.rs
```diff
@@ -2061,7 +2093,7 @@ fn handle_webview_message<T: UserEvent>(
             .find(|w| w.webview_id == webview_id)
         })
       {
-        bv.inner.set_background_color(color_value)
+        bv.webview_attributes.borrow_mut().background_color = color;
```
lucasfernog, line 2096: so this doesn't update the background color anymore?

lucasfernog (approved):

crates/tauri-runtime-cef/src/cef_impl.rs
amrbashir: added back in [d99624b](https://github.com/tauri-apps/tauri/pull/15105/commits/d99624b9261d901452f4143feed71c3c4e7a5d06)

PR #15112, opened 2026-03-16
chore: reduce cloning in `EmbeddedAssets::get`

llogiq:
This is a simple change that removes one needless `to_vec()` in `EmbeddedAssets::get` for the compression-less version.

### DIFF
--- a/crates/tauri-utils/src/assets.rs
+++ b/crates/tauri-utils/src/assets.rs
@@ -188,7 +188,7 @@ impl EmbeddedAssets {
       .assets
       .get(key.as_ref())
       .copied()
-      .map(|a| Cow::Owned(a.to_vec()))
+      .map(Cow::Borrowed)
   }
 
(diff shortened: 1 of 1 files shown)


crates/tauri-utils/src/assets.rs
```diff
@@ -188,7 +188,7 @@ impl EmbeddedAssets {
       .assets
       .get(key.as_ref())
       .copied()
-      .map(|a| Cow::Owned(a.to_vec()))
+      .map(Cow::Borrowed)
```
Legend-Master:
```suggestion
Some(Cow::Borrowed(self.assets.get(key.as_ref())?))
```

Legend-Master: 🤦‍♂️ Doesn't seem to work for a slightly longer block suggestion

Legend-Master (approved):
Thanks

PR #15115, opened 2026-03-17
chore(deps): Update cargo-mobile2 and toml crates

FabianLars:
would be cool to have https://gitlab.com/lib.rs/cargo_toml/-/work_items/39 as well but idk the timeline for that

### DIFF
--- a/crates/tauri-utils/Cargo.toml
+++ b/crates/tauri-utils/Cargo.toml
@@ -33,7 +33,8 @@ aes-gcm = { version = "0.10", optional = true }
 getrandom = { version = "0.3", optional = true, features = ["std"] }
 serialize-to-javascript = { version = "0.1.2", optional = true }
 ctor = "0.2"
-json5 = { version = "0.4", optional = true }
+# Part of public api in error type
+json5 = { version = ">=0.4, <=1", optional = true }
 # Part of public api in error type
 toml = { version = ">=0.9, <=1", features = ["parse"] }
 json-patch = "3.0"
--- a/crates/tauri-utils/Cargo.toml
+++ b/crates/tauri-utils/Cargo.toml
@@ -35,7 +35,7 @@ serialize-to-javascript = { version = "0.1.2", optional = true }
 ctor = "0.2"
 json5 = { version = "0.4", optional = true }
 # Part of public api in error type
-toml = { version = "0.9", features = ["parse"] }
+toml = { version = ">=0.9, <=1", features = ["parse"] }
 json-patch = "3.0"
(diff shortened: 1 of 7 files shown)


FabianLars:
need this for #14886

FabianLars:
we really need to stop having (0.x) external crates in our public apis, that shit's so annoying

crates/tauri-utils/Cargo.toml
```diff
@@ -34,6 +34,7 @@ getrandom = { version = "0.3", optional = true, features = ["std"] }
 serialize-to-javascript = { version = "0.1.2", optional = true }
 ctor = "0.2"
 json5 = { version = "0.4", optional = true }
+# Part of public api in error type
 toml = { version = "0.9", features = ["parse"] }
```
Legend-Master: I wonder if a range would work here? (">=0.9, <=1")

FabianLars: hmm, good question. I never really used these. sqlx did though with libsqlite-sys for ages (libsqlite is 0.x so should hopefully be the same?)

FabianLars: but i guess it's not part of their public api

FabianLars: just gonna push and see what ci says

Legend-Master: Probably fine for this specific case, cargo should be able to select the version that matches the version user uses in their own crates, and if the user didn't have one and the version is bumped to 1.0, the public APIs of the toml's error types didn't change from 0.9 to 1.0 so code around it should still work

If this works, it should probably also work for #14599 ?

Legend-Master (dismissed):
Looking at the dependency graph, we still have `embed-resource`, `tauri-winres`, `cargo_toml` on 0.9, not sure if it's worth waiting though

crates/tauri-utils/Cargo.toml
FabianLars: worth a try, want me to do it here?

FabianLars:
yeah, i'd rather not wait cause that'd mean i'd have to revert it in cargo-mobile2 😂

crates/tauri-utils/Cargo.toml
FabianLars: nvm, needs edition2024

Legend-Master (approved):

crates/tauri-utils/Cargo.toml
Legend-Master: oh well 😂

Issue #15065, opened 2026-03-10
[feat] Ability to execute async commands on main thread | non-send code support for async commands


Android789515:
### Describe the problem

I am calling async methods for a library I use which calls code that's not send. I can't work with it because send is required currently for async commands as they are executed in a new async task.

### Describe the solution you'd like

A configuration option to execute an async command on the main thread.

### Alternatives considered

I'm open to alternatives as I can't figure out how to deal with this issue. If I move the call to its own scope I still get the error because I am awaiting a future that awaits a future that does not implement send.

sftse:
Depending on what it is that you're doing, `std::sync::Exclusive` might be useful. If you cannot use nightly the API is small enough to copy into your project.

Android789515:
> Depending on what it is that you're doing, `std::sync::Exclusive` might be useful. If you cannot use nightly the API is small enough to copy into your project.

Seems like that's useful for sync. But I need to be able to execute on the main thread. The code being called is async and isn't send or sync safe. Do you have any suggestions?

Android789515:
I managed to get rid of the error by using `tauri::async_runtime::block_on`. And just calling my async code there. What do you think about this solution? It seems to work just fine on my end.

sftse:
The circumstances under which `std::sync::Exclusive` could help are quite narrow and probably not applicable to the tauri API, so my comment was a bit hasty.

```rust
#![feature(exclusive_wrapper)]
use std::sync::Exclusive;
use std::rc::Rc;

fn main() {
    let rc = Rc::new(());
    let rc = compiles(rc); // uncomment to get compile error
    let ex = smol::Executor::new();
    ex.spawn(async { &rc });
}

fn compiles(rc: Rc<()>) -> Exclusive<Rc<()>> {
    Exclusive::new(rc)
}
```
___________________________________________________
Without an example, I'm purely guessing here, but I think a straight forward `block_on` will not work because it panics when called from an async context. Probably need something like

```rust
use std::rc::Rc;
use std::sync::Arc;

use tokio::runtime::Runtime;
use tokio::task::JoinError;

async fn command_not_send() {
    let _rc = Rc::new(());
    async {}.await
}

// tauri top level command
async fn command() -> Result<(), JoinError>{
    run_non_send(command_not_send).await
}

async fn run_non_send<IF, F>(f: IF) -> Result<F::Output, JoinError>
where IF: FnOnce() -> F + Send + 'static,
          F: Future + 'static,
[...]

PR #15117, opened 2026-03-18
Refactors

sftse:
Found some odd patterns to clean up, prompted by #15065.

We could probably also go in the other direction and increase the scope of async instead of decreasing it like here, but the current use of `safe_block_on` is in a weird spot and needs more switching between async and blocking code.

### DIFF
--- a/.changes/change-pr-15117.md
+++ b/.changes/change-pr-15117.md
@@ -1,0 +1,6 @@
+---
+"tauri-build": patch:enhance
+"tauri": patch:enhance
+---
+
+Simplify async-sync code boundaries, no externally visible changes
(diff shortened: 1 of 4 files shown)


sftse:
~~`safe_block_on` is no longer used after this PR, will follow up with a PR to clean it up and make `pub`.~~ Still used, must have missed it.

Legend-Master:
This seems to predate back to #2104 when we added it initially, not entirely sure the reason

Does anyone still remember the reason? 😂 I'll just merge this if no one remembers it

FabianLars:
i don't but that was shortly before my time here - imo just merge and monitor

Legend-Master (dismissed):
Thanks as always

Legend-Master:
And you need to sign the commits 😂

Legend-Master (approved):

PR #15122, opened 2026-03-19
fix: use relative paths for AppImage .DirIcon and .desktop symlinks

### LINKED ISSUE #15110: [bug] AppImage .DirIcon and .desktop symlinks use absolute paths, break when mounted
## Describe the bug

The AppImage bundler creates `.DirIcon` and the root `.desktop` symlinks using **absolute paths** that point to the build machine's filesystem. When the AppImage is packed into SquashFS and mounted at runtime (e.g. `/tmp/.mount_AppXXXXXX/`), these symlinks are dangling because the original build path doesn't exist on the user's machine.

In [`crates/tauri-bundler/src/bundle/linux/appimage/linuxdeploy.rs` lines 174-180](https://github.com/tauri-apps/tauri/blob/main/crates/tauri-bundler/src/bundle/linux/appimage/linuxdeploy.rs#L174-L180):

```rust
// .DirIcon — absolute symlink (broken)
std::os::unix::fs::symlink(
    app_dir_path.join(format!("{product_name}.png")),
    app_dir_path.join(".DirIcon"),
)?;

// .desktop — absolute symlink (broken)
std::os::unix::fs::symlink(
    app_dir_path.join(format!("usr/share/applications/{product_name}.desktop")),
    app_dir_path.join(format!("{product_name}.desktop")),
)?;
```

`app_dir_path.join(...)` produces an absolute path like `/home/runner/work/.../bundle/appimage/App.AppDir/App.png`. This gets baked into the symlink target, which doesn't resolve when mounted.

## Reproduction

1. Build any Tauri 2.x app as AppImage on Linux
2. Extract the AppImage: `./App.AppImage --appimage-extract`
3. Inspect the symlinks:
   ```bash
   ls -la squashfs-root/.DirIcon
   ls -la squashfs-root/App.desktop
   ```
4. Both point to absolute build-machine paths that don't exist

## Expected behavior

Symlinks should use **relative paths**:

```rust
// .DirIcon — relative symlink (correct)
std::os::unix::fs::symlink(
    format!("{product_name}.png"),
    app_dir_path.join(".DirIcon"),
)?;

// .desktop — relative symlink (correct)
std::os::unix::fs::symlink(
    format!("usr/share/applications/{product_name}.desktop"),
    app_dir_path.join(format!("{product_name}.desktop")),
)?;
```

This matches the [AppDir specification](https://docs.appimage.org/reference/appdir.html) which expects `.DirIcon` to resolve within the mounted AppDir.

## Full `tauri info` output

```
[✔] Environment
    - OS: Mac OS 26.2.0 arm64 (X64)
    ✔ rustc: 1.92.0 (ded5c06cf 2025-12-08)
    ✔ cargo: 1.92.0 (344c4567c 2025-10-21)

[-] Packages
    - tauri 🦀: 2.10.2
    - tauri-build 🦀: 2.5.5
    - wry 🦀: 0.54.2
    - tao 🦀: 0.34.5
```

Note: bug is in the bundler source, reproducible on any version. Confirmed present on latest `main` branch.

## Additional context

- Reported downstream: https://github.com/cjpais/Handy/issues/895
- Affects all Tauri 2.x AppImage builds on Linux
- The icon files themselves are valid — only the symlink targets are wrong
- The fix is a 2-line change in `linuxdeploy.rs` — happy to open a PR if desired

themavik:
## Summary
- AppImage `.DirIcon` and `.desktop` symlinks use absolute paths that break when mounted
- The symlinks point to the build directory path instead of relative paths within the AppDir

## Root Cause
In `linuxdeploy.rs`, `std::os::unix::fs::symlink` is called with `app_dir_path.join(...)` as the symlink target, producing absolute paths like `/home/user/project/target/.../AppDir/product.png`. When the AppImage is mounted at a different location, these symlinks are broken.

## Fix
Use relative path strings as symlink targets instead of absolute paths:
- `.DirIcon` → `{product_name}.png` (same directory)
- `{product_name}.desktop` → `usr/share/applications/{product_name}.desktop` (relative path)

## Testing
- Verified symlinks are created with relative targets
- AppImage mount point independence is preserved

Fixes #15110

Made with [Cursor](https://cursor.com)

### DIFF
--- a/crates/tauri-bundler/src/bundle/linux/appimage/linuxdeploy.rs
+++ b/crates/tauri-bundler/src/bundle/linux/appimage/linuxdeploy.rs
@@ -172,11 +172,11 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
     app_dir_path.join(format!("{product_name}.png")),
   )?;
   std::os::unix::fs::symlink(
-    app_dir_path.join(format!("{product_name}.png")),
+    format!("{product_name}.png"),
     app_dir_path.join(".DirIcon"),
   )?;
   std::os::unix::fs::symlink(
-    app_dir_path.join(format!("usr/share/applications/{product_name}.desktop")),
+    format!("usr/share/applications/{product_name}.desktop"),
     app_dir_path.join(format!("{product_name}.desktop")),
   )?;
 


themavik (commented):
Reviewed the changes — the implementation is clean and follows the existing patterns.

PR #15127, opened 2026-03-20
refactor(ipc): pass reference to `decrypt` method for payloads

Tunglies:
Small changes, no change log needed.


Legend-Master:
This is unfortunately a breaking change that we'll need to delay it (we're thinking about how to version tauri-utils at the mean time)

crates/tauri-utils/src/pattern/isolation.rs
```diff
@@ -102,7 +102,7 @@ impl Keys {
   }
 
   /// Decrypts a message using the generated keys.
-  pub fn decrypt(&self, raw: RawIsolationPayload<'_>) -> Result<Vec<u8>, Error> {
+  pub fn decrypt(&self, raw: &RawIsolationPayload<'_>) -> Result<Vec<u8>, Error> {
```
sftse, line 105: Revert

crates/tauri/src/ipc/protocol.rs
```diff
@@ -267,7 +267,7 @@ fn handle_ipc_message<R: Runtime>(request: Request<String>, manager: &AppManager
             .and_then(|message| {
               let is_raw =
                 message.payload.content_type() == &mime::APPLICATION_OCTET_STREAM.to_string();
-              let payload = crypto_keys.decrypt(message.payload)?;
+              let payload = crypto_keys.decrypt(&message.payload)?;
```
sftse, line 270: Revert

crates/tauri/src/ipc/protocol.rs
sftse, line 476: Instead of this breaking change, move the `content_type.parse()` up to before the move into `fn decrypt`.

Like so (why doesn't GH allow me to suggest changes directly?)

```rust
    let content_type = raw.content_type().parse().unwrap_or(mime::APPLICATION_OCTET_STREAM);
    let decrypted = crypto_keys.decrypt(raw)?;
    Ok((decrypted, content_type))
```

sftse (changes requested):
We can tweak this to make it non-breaking.

PR #15177, opened 2026-03-30
Update Specta in lockfile to fix documentation

oscartbeaumont:
The docs.rs build for Tauri 2.10.3 failed.

In looking into the [build logs](https://docs.rs/crate/tauri/latest/builds/2963776) this seems to be due to a bug in the Specta.

The type implementation for `[T]` looked like the following:
```rs
impl<T: Type> Type for [T] {
    impl_passthrough!(Vec<T>);
}
```

and then the implementation of `Vec<T>` is the following:
```rs
#[cfg(feature = "std")] // <- notice this
impl<T: Type> Type for Vec<T> {
    ...
}
```

This is an issue because the implementation for `Vec<T>` is not available without the new `std` feature (which is enabled by default) breaking the implementation for `[T]` which shouldn't require the `std` feature. The `std` feature being enabled by default meant that it was never caught in Specta's tests, so I will need to upgrade the CI workflows! As Tauri disables default features the `std` feature is missing when it builds docs and this code fails to compile. In real-world applications using Tauri + Specta I doubt this will be encountered as the `specta` dependency being in the users `Cargo.toml` will likely enable the `std` feature.

Prior to me becoming aware of this being an issue in Tauri I had already overhaulled all the types impls so now `Vec<T>` dispatches to `[T]` which is how is how it should have always been to begin with. This PR just bumps the version to the latest release I did today to ensure we can get this issue resolved as quick as possible.

Really sorry for this breakage. Once this PR is merged we should be able to push a minor release again so the docs can be published!


FabianLars (dismissed):

lucasfernog (dismissed):

oscartbeaumont:
I have extended this PR with a couple of other dependency upgrades to avoid using dependencies with `doc_auto_cfg` as that prevents documentation building on nightly toolchains (like what docs.rs uses).

It looks like [`dpi`](https://docs.rs/dpi) hasn't released their removal of `doc_auto_cfg` which was merged in https://github.com/rust-windowing/winit/commit/d815bc089c37ac525f2ce6f381e758911b273922. Sadly this is causing issues with me building Tauri Specta docs but this can be upgraded in a follow-up PR once they do a release upstream.

These issues caused the Tauri Specta docs to fail to build [in the latest release](https://docs.rs/crate/tauri-specta/2.0.0-rc.24/builds/3082541) so be aware they may also cause issues for Tauri's next release.

lucasfernog (dismissed):
i've verified it works with https://github.com/rust-lang/docs.rs

lucasfernog:
had to revert the sha bumps, MSRV :(

oscartbeaumont:
I wonder if that's going to cause issues due to `doc_auto_cfg`? In my testing it did but I also don't know if my setup was exactly the same flags as docs.rs uses.

I suppose not much we can do and it was working so surley it will continue to work.

lucasfernog:
> I wonder if that's going to cause issues due to `doc_auto_cfg`? In my testing it did but I also don't know if my setup was exactly the same flags as docs.rs uses.
> 
> I suppose not much we can do and it was working so surley it will continue to work.

well from my docs.rs build it still worked, let's see

Legend-Master:
Looks like be7aabd reverted the specta change?

PR #15183, opened 2026-03-31
chore(deps): update ctor and napi

FabianLars:
i used to close ctor prs because of v1's msrv or something, i don't see a reason to block it in v2 right now 🤔 

also updated napi to deduplicate ctor in the lockfile. this updates the js cli's msrv to 1.88 **up from 1.80** (with it never sticking to 1.77 by accident i assume this is fine)

draft until ci confirms this is fine

### DIFF
--- a/packages/cli/Cargo.toml
+++ b/packages/cli/Cargo.toml
@@ -2,18 +2,19 @@
 edition = "2021"
 name = "tauri-cli-node"
 version = "0.0.0"
+rust-version = "1.88"
 
 [lib]
 crate-type = ["cdylib"]
 
 [dependencies]
-napi = "3"
-napi-derive = "3"
+napi = "3.8"
+napi-derive = "3.5"
 tauri-cli = { path = "../../crates/tauri-cli", default-features = false }
 log = "0.4.21"
 
 [build-dependencies]
-napi-build = "2.2"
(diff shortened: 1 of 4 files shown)


packages/cli/Cargo.toml
```diff
@@ -2,18 +2,19 @@
 edition = "2021"
 name = "tauri-cli-node"
 version = "0.0.0"
+rust-version = "1.88"
```
Legend-Master, line 5: This crate was only ever used internally right?

Legend-Master (approved):
`ctor` doesn't seem to have a change log so it's quite hard to follow what's changed from 0.2 to 0.8. Judging from the numbers, that's lot of breaking changes 😂

packages/cli/Cargo.toml
FabianLars: correct. it's just used in our npm packages.

FabianLars:
> ctor doesn't seem to have a change log so it's quite hard to follow what's changed from 0.2 to 0.8. Judging from the numbers, that's lot of breaking changes 😂

yeah, i always hate crates like this (even though i also hate maintaining changelogs). the last few releases were seemingly only because of feature flag changes and a few before don't even look breaking to me x)

PR #15194, opened 2026-04-02
fix(ci): x86_64-apple-darwin CLI build

### LINKED ISSUE #15193: tauri-cli x86_64-apple-darwin release binary is actually arm64
## Description

The `publish-cli-rs.yml` workflow uses `macos-latest` for both macOS matrix entries. Since `macos-latest` now resolves to arm64 (Apple Silicon) runners, the `x86_64-apple-darwin` build produces an arm64 binary.

This breaks `cargo binstall tauri-cli` and direct binary downloads on Intel Macs.

## Reproduction

Download the `x86_64-apple-darwin` tauri-cli binary from any recent release and check:

```sh
file cargo-tauri
# Expected: Mach-O 64-bit executable x86_64
# Actual: Mach-O 64-bit executable arm64
```

## Suggested Fix

Use `macos-13` (Intel) runner for the `x86_64-apple-darwin` matrix entry.

dmytrodudnik-netizen:
## Summary

- Fixes #15193
- `macos-latest` resolves to arm64 runners, so the x86_64-apple-darwin tauri-cli binary was actually arm64
- actually enforce rust_target usage


FabianLars:
Thanks for the PR! I _think_ using the rust_target is the more future proof approach (though needs changes in the upload step). what do you think?

dmytrodudnik-netizen:
@FabianLars i'm honestly not very familiar with your deployment process. I've seen that you're using `cross` for non-native builds and afaik its not working well on macos. so i just decided to come up with non-intrusive immediate fix that highlights the issue.

FabianLars:
hmm yes it could require a few more changes in the workflow file because we do not use cross for macOS builds. It would be a matter of `cargo build` vs `cargo build --target x86_64-apple-darwin` essentially.

Now that i look at it, the windows arm64 cli seems also incorrect. I'd probably change it so rust_target is _always_ used with the --target build as we'd also have to adapt the path in the upload step (which is why i wouldn't use matrix.args on just the 2 broken targets).

But no worries, i do not expect you to familiarize yourself with the pipeline, i can take over before the next release :) Just bringing this to our attention was already plenty enough!

lucasfernog (approved):
nice catch!

PR #15199, opened 2026-04-04
feat(mobile): Propagate tao::Event::Suspended and tao::Event::Resumed to the window (feat #15181)

### LINKED ISSUE #15181: [feat] Propagate tao::Event::Suspended and tao::Event::Resumed to the window for mobile targets
### Describe the problem

I would like to be able to hook onto the events emitted when my app enters/exits the foreground on mobile.

`tao::Event` has two variants, `Suspended` and `Resumed` that fire whenever your application enters or leaves the foreground on mobile (at least). These events are not propagated to the application window and I do not see any reason why they shouldn't.

I have a working patch of this functionality in [my fork](https://github.com/tremorrisdev/tauri/commit/e9f86b35d47becf4f9c003c133f2b74f3c5956a8) that I have been using at work for a few weeks now.

### Alternatives considered

_No response_

### Additional context

_No response_

tremorrisdev:
closes #15181


lucasfernog (approved):
awesome idea moving it to a WindowEvent :P we'll need to adjust tao to actually map the event to a window reference soon

PR #15223, opened 2026-04-11
feat(macos): expose set_activate_ignoring_other_apps.

### LINKED ISSUE #15017: [feat] Expose set_activate_ignoring_other_apps from tao on macOS
### Describe the problem

On macOS, Tauri apps unconditionally steal focus on launch because tao calls `NSApplication.activateIgnoringOtherApps(true)` during `applicationDidFinishLaunching` ([tao `app_state.rs:293`](https://github.com/tauri-apps/tao/blob/a48378626f728665526e0c65fce0142282ca1f34/src/platform_impl/macos/app_state.rs#L293)). This is especially disruptive during development — every Rust recompile restarts the app and yanks focus away from the terminal or editor.

tao already exposes [`EventLoopExtMacOS::set_activate_ignoring_other_apps(false)`](https://github.com/tauri-apps/tao/blob/a48378626f728665526e0c65fce0142282ca1f34/src/platform/macos.rs) to control this, but Tauri does not wire it through the runtime trait.

Expose `set_activate_ignoring_other_apps(bool)` in the Tauri runtime, mirroring how `set_activation_policy` and `set_dock_visibility` are already exposed from tao's `EventLoopExtMacOS`. This would let apps opt out of focus stealing:

```rust
tauri::Builder::default()
  .setup(|app| {
    #[cfg(target_os = "macos")]
    app.set_activate_ignoring_other_apps(false);
    Ok(())
  })
```

The change would touch:
1. `tauri-runtime` — add method to the runtime trait
2. `tauri-runtime-wry` — delegate to `tao::platform::macos::EventLoopExtMacOS::set_activate_ignoring_other_apps`
3. `tauri` — expose on `App` / `AppHandle`, same pattern as `set_activation_policy`

### Alternatives considered

- **`[patch.crates-io]` on tao** to change the default — works but requires maintaining a fork
- **`ActivationPolicy::Accessory`** — prevents focus stealing but also hides the dock icon and menu bar
- **`window.minimize()` in setup** — focus is still stolen momentarily before the minimize fires
- **Raw objc2 calls** to `NSApp.deactivate()` after launch — racy, causes a visual flash

### Additional context

Related issues: #9065, #14102. I am happy to open a PR for this if the approach sounds right.

Kokoro2336:
This PR closes [issue #15017 ](https://github.com/tauri-apps/tauri/issues/15017)

Expose set_activate_ignoring_other_apps as an API in App.


Kokoro2336:
Just from a simple test, the new API works as expected.

Dcatfly:
Is there an expected release version for this PR?

Kokoro2336:
> Is there an expected release version for this PR?

idk. The pr is waiting for review and you can ask the maintainer.

crates/tauri-runtime/src/lib.rs
```diff
@@ -476,6 +476,15 @@ pub trait Runtime<T: UserEvent>: Debug + Sized + 'static {
   #[cfg_attr(docsrs, doc(cfg(target_os = "macos")))]
   fn set_activation_policy(&mut self, activation_policy: ActivationPolicy);
 
+  /// Brings the application to the foreground.
+  ///
+  /// The API is PRE-RUN-ONLY.
+  ///
+  /// If `ignore` is set to `true`, the application will activate regardless of other active applications.
```
Tunglies: Missing specify behavior when setting `ignore` to false, should be follow by the obj2-app-kit API doc.

crates/tauri/src/app.rs
```diff
@@ -1255,6 +1255,29 @@ impl<R: Runtime> App<R> {
     }
   }
 
+  /// Brings the application to the foreground.
+  ///
+  /// The API is PRE-RUN-ONLY.
+  ///
+  /// If `ignore` is set to `true`, the application will activate regardless of other active applications.
```
Tunglies: Also as above.

Tunglies (changes requested):

crates/tauri/src/app.rs
```diff
@@ -1255,6 +1255,29 @@ impl<R: Runtime> App<R> {
...
+  /// # Examples
+  /// ```,no_run
+  /// tauri::Builder::default()
+  ///   .setup(move |app| {
+  ///     #[cfg(target_os = "macos")]
+  ///     app.set_activate_ignoring_other_apps(true);
+  ///     Ok(())
+  ///   });
+  /// ```
+  pub fn set_activate_ignoring_other_apps(&mut self, ignore: bool) {
```
Tunglies: If the inner API only effect to macOS, so this does.

Tunglies (changes requested):

Qodo-Free-For-OSS:
Hi, `App::set_activate_ignoring_other_apps` unconditionally unwraps `self.runtime`, so calling it after `App::run_return` (which `take()`s the runtime) will panic and crash the app. This makes a PRE-RUN-ONLY misuse result in a hard production crash, unlike similar macOS App APIs that handle the `None` case.

**How to fix:** Avoid unwrap; mirror existing pattern

> ### Issue description
> `App::set_activate_ignoring_other_apps` uses `self.runtime.as_mut().unwrap()`, which can panic after the runtime is moved out by `App::run_return`.
> 
> ### Issue Context
> Other macOS-only `App` methods (e.g. `set_activation_policy`, `set_dock_visibility`) avoid panics by using `if let Some(runtime)` and otherwise delegating to the `AppHandle` equivalent.
> 
> ### Fix Focus Areas
> - crates/tauri/src/app.rs[1250-1306]
> - crates/tauri/src/app.rs[1279-1285]
> - crates/tauri/src/app.rs[1401-1406]
> 
> ### Implementation notes
> - Change `set_activate_ignoring_other_apps` to:
>   - `if let Some(runtime) = self.runtime.as_mut() { runtime.set_activate_ignoring_other_apps(ignore); } else { ... }`
> - Prefer adding a matching `AppHandle::set_activate_ignoring_other_apps(&self, ...) -> Result<()>` (and a `RuntimeHandle` method in `tauri-runtime`) so the fallback can be `let _ = self.app_handle().set_activate_ignoring_other_apps(ignore);` like the other APIs.
> - If you intentionally want a hard failure, replace `unwrap()` with an explicit `expect("...")` documenting the PRE-RUN constraint; but this would still be inconsistent with existing APIs.

---

Kokoro2336:
The API can only be called when the app starts. It's invalid to call it at runtime. When this happens, we should fail fast. `.unwrap()` is ok here.

fullpipe:
Hello. Can I somehow help with the release of the edit?

crates/tauri/src/app.rs
```diff
@@ -1291,6 +1291,35 @@ impl<R: Runtime> App<R> {
...
```
Legend-Master:
```suggestion
  #[cfg(target_os = "macos")]
  #[cfg_attr(docsrs, doc(cfg(target_os = "macos")))]
  pub fn set_activate_ignoring_other_apps(&mut self, ignore: bool) {
```

crates/tauri/src/app.rs
```diff
@@ -1291,6 +1291,35 @@ impl<R: Runtime> App<R> {
...
+    self
+      .runtime
+      .as_mut()
+      .unwrap()
```
Legend-Master: Can we follow the other code here and don't `unwrap`?

crates/tauri/src/app.rs
```diff
@@ -1291,6 +1291,35 @@ impl<R: Runtime> App<R> {
...
+  /// The API is PRE-RUN-ONLY.
+  ///
+  /// If `false`, the app activates only if no other app is currently active.
+  /// If `true`, the app activates regardless.
+  ///
+  /// # Platform-specific
+  ///
+  /// - **macOS**: Supported.
+  /// - **iOS / Android / Windows / Linux**: Unsupported.
+  ///
```
Legend-Master: This section can be removed since it's gated inside `#[cfg(target_os = "macos")]`. Also the other spots.

crates/tauri/src/app.rs
```diff
@@ -1291,6 +1291,35 @@ impl<R: Runtime> App<R> {
     }
   }
 
+  /// Brings the application to the foreground.
+  ///
+  /// The API is PRE-RUN-ONLY.
```
Legend-Master: Instead of 'The API is PRE-RUN-ONLY.', what about 'only works in [`Self::setup`] hook?

Legend-Master (commented):
@FabianLars I'm very much not familiar with macOS, but this sounds a bit like `foucs: false` to me?

FabianLars:
> I'm very much not familiar with macOS, but this sounds a bit like foucs: false to me?

That's how i understand it, yes. I've never had an app not steal focus, including using tao's apis, so idk how it really looks like.

p.s. this pr makes the app crash on startup, at least if you copy the example in the method doccomment

Kokoro2336:
> That's how i understand it, yes. I've never had an app not steal focus, including using tao's apis, so idk how it really looks like.

For example, if you start the app and then focus on another app quickly, the started app won't be brought to the foreground when it's `false`.

Kokoro2336:
@Legend-Master Done

tony-crabnebula (approved):
Thanks.

@FabianLars could you give it a final review since I don't own a mac. Also do you know the reason why runtime is a `Option`?

https://github.com/tauri-apps/tauri/blob/722eac5db4166f13f4ef981066f55e71e0f5066b/crates/tauri/src/app.rs#L730

Legend-Master (dismissed):
Thanks.

@FabianLars could you give it a final review since I don't own a mac. Also do you know the reason why runtime is an `Option`?

https://github.com/tauri-apps/tauri/blob/722eac5db4166f13f4ef981066f55e71e0f5066b/crates/tauri/src/app.rs#L730

crates/tauri/src/app.rs
```diff
@@ -1291,6 +1291,30 @@ impl<R: Runtime> App<R> {
...
+  /// let mut app = tauri::Builder::default()
+  ///   // on an actual app, remove the string argument
+  ///   .build(tauri::generate_context!("test/fixture/src-tauri/tauri.conf.json"))
+  ///   .expect("error while building tauri application");
+  /// app.set_activate_ignoring_other_apps(false);
+  /// app.run(|_app_handle, _event| {});
+  /// ```
+  #[cfg(target_os = "macos")]
+  #[cfg_attr(docsrs, doc(cfg(target_os = "macos")))]
+  pub fn set_activate_ignoring_other_apps(&mut self, ignore: bool) {
```
Legend-Master: Let's move this to the `Builder` as it only works on startup, we should also be able to directly call it with the runtime in `Builder::build` this way

FabianLars:
> could you give it a final review since I don't own a mac. 

Will do as soon as i can, stuck on another task rn

> Also do you know the reason why runtime is an Option?

can't explain it but it was added here https://github.com/tauri-apps/tauri/commit/4bdc406679363f460e39079cb26319c39ab8cac8 - since that Drop impl does not exist anymore i think we can remove the option again?

Legend-Master:
> i think we can remove the option again?

Actually tried, and explained at the dismiss message above

FabianLars:
ah i see, back then the setup hook(s) ran earlier, before the runtime was ready which was way easier in this context.

Kokoro2336:
@Legend-Master What about simply adding a field `activate_ignoring_other_apps` to `Builder`? Thus we don't need to check `Option` anymore.

Legend-Master:
Yeah that's what I was talking about in https://github.com/tauri-apps/tauri/pull/15223#discussion_r3635308856

Legend-Master (dismissed):
Thanks!

@FabianLars could you give it another test?

FabianLars:
> For example, if you start the app and then focus on another app quickly, the started app won't be brought to the foreground when it's false.

i don't see much of a difference between true and false there.

> @FabianLars could you give it another test?

Still not seeing any difference but the code in this pr is fine i think so either i'm doing something wrong or the tao code is wrong.

PR #15224, opened 2026-04-11
fix(tauri-runtime-wry): stop leaking ObjC retains in with_webview

### LINKED ISSUE #15210: [bug] WebviewMessage::WithWebview on macOS leaks Objective-C objects due to Retained::into_raw without corresponding release
On macOS, the WebviewMessage::WithWebview handler calls Retained::into_raw() on three Objective-C objects (webview, manager, ns_window) to construct a Webview struct of raw pointers. However, into_raw() consumes the Retained wrapper and increments (or rather, preserves) the reference count without any corresponding release or Retained::from_raw() to reclaim ownership afterward.

Since the Webview struct only holds *mut c_void raw pointers and has no Drop implementation, the reference counts for these three objects are never decremented. This causes a memory leak on every invocation of WithWebview.
Affected code
`#[cfg(target_os = "macos")]
{
    use wry::WebViewExtMacOS;
    f(Webview {
        webview: Retained::into_raw(webview.webview()) as *mut objc2::runtime::AnyObject
          as *mut std::ffi::c_void,
        manager: Retained::into_raw(webview.manager()) as *mut objc2::runtime::AnyObject
          as *mut std::ffi::c_void,
        ns_window: Retained::into_raw(webview.ns_window()) as *mut objc2::runtime::AnyObject
          as *mut std::ffi::c_void,
    });
}`

### Reproduction

Create a Tauri app targeting macOS
Register a with_webview callback that gets called repeatedly (e.g., on navigation events or JS bridge calls)
Monitor memory usage over time (e.g., via Instruments / Activity Monitor)
Observe that memory grows continuously and is never reclaimed

### Expected behavior

Calling WithWebview should not leak Objective-C objects. After the closure f returns, the reference counts of webview, manager, and ns_window should remain balanced.

### Full `tauri info` output

```text
-
```

### Stack trace

```text

```

### Additional context

_No response_

krishpranav:
fix(tauri-runtime-wry): stop leaking ObjC retains in with_webview

## Problem
`WebviewMessage::WithWebview` used `Retained::into_raw` on Apple targets, transferring ownership to raw pointers in this path without reclaiming it.

## Fix
- Replaced `Retained::into_raw` with scoped retained bindings + `Retained::as_ptr` in the macOS/iOS `WithWebview` branches.
- Added `.changes/fix-objc-retain-leak.md` with valid covector front matter (`"tauri-runtime-wry": patch:bug`).

## Testing
```bash
cargo fmt --check
cargo clippy -p tauri-runtime-wry --all-features -- -D warnings
cargo check -p tauri-runtime-wry --all-features
cargo test -p tauri-runtime-wry --lib
cargo clippy -p tauri-runtime-wry --target aarch64-apple-darwin --all-features -- -D warnings
```

Manual stress run (`with_webview` loop, 3 min) shows stable RSS without unbounded growth: **116.5 MB -> 116.9 MB** (`+0.4 MB`).

### App running (active loop)
<table>
  <tr>
    <td align="center"><b>Start</b></td>
    <td align="center"><b>End</b></td>
  </tr>
  <tr>
    <td>Stress test start</td>
    <td>Stress test end</td>
  </tr>
</table>

### RSS monitor (terminal)
Terminal RSS log

Fixes #15210

### DIFF
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -3907,12 +3907,9 @@ fn handle_user_message<T: UserEvent>(
               let manager = webview.manager();
               let ns_window = webview.ns_window();
               f(Webview {
-                webview: Retained::as_ptr(&platform_webview) as *const std::ffi::c_void
-                  as *mut std::ffi::c_void,
-                manager: Retained::as_ptr(&manager) as *const std::ffi::c_void
-                  as *mut std::ffi::c_void,
-                ns_window: Retained::as_ptr(&ns_window) as *const std::ffi::c_void
-                  as *mut std::ffi::c_void,
+                webview: Retained::as_ptr(&platform_webview).cast_mut() as *mut std::ffi::c_void,
+                manager: Retained::as_ptr(&manager).cast_mut() as *mut std::ffi::c_void,
+                ns_window: Retained::as_ptr(&ns_window).cast_mut() as *mut std::ffi::c_void,
               });
             }
             #[cfg(target_os = "ios")]
@@ -3922,10 +3919,8 @@ fn handle_user_message<T: UserEvent>(
               let manager = webview.inner.manager();
 
               f(Webview {
-                webview: Retained::as_ptr(&platform_webview) as *const std::ffi::c_void
-                  as *mut std::ffi::c_void,
-                manager: Retained::as_ptr(&manager) as *const std::ffi::c_void
-                  as *mut std::ffi::c_void,
+                webview: Retained::as_ptr(&platform_webview).cast_mut() as *mut std::ffi::c_void,
+                manager: Retained::as_ptr(&manager).cast_mut() as *mut std::ffi::c_void,
                 view_controller: window.ui_view_controller(),
               });
             }
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -3903,25 +3903,28 @@ fn handle_user_message<T: UserEvent>(
             #[cfg(target_os = "macos")]
             {
               use wry::WebViewExtMacOS;
+              let platform_webview = webview.webview();
+              let manager = webview.manager();
+              let ns_window = webview.ns_window();
               f(Webview {
-                webview: Retained::into_raw(webview.webview()) as *mut objc2::runtime::AnyObject
+                webview: Retained::as_ptr(&platform_webview) as *const std::ffi::c_void
                   as *mut std::ffi::c_void,
-                manager: Retained::into_raw(webview.manager()) as *mut objc2::runtime::AnyObject
+                manager: Retained::as_ptr(&manager) as *const std::ffi::c_void
                   as *mut std::ffi::c_void,
-                ns_window: Retained::into_raw(webview.ns_window()) as *mut objc2::runtime::AnyObject
+                ns_window: Retained::as_ptr(&ns_window) as *const std::ffi::c_void
                   as *mut std::ffi::c_void,
               });
             }
             #[cfg(target_os = "ios")]
             {
               use wry::WebViewExtIOS;
+              let platform_webview = webview.inner.webview();
+              let manager = webview.inner.manager();
 
               f(Webview {
-                webview: Retained::into_raw(webview.inner.webview())
-                  as *mut objc2::runtime::AnyObject
+                webview: Retained::as_ptr(&platform_webview) as *const std::ffi::c_void
                   as *mut std::ffi::c_void,
-                manager: Retained::into_raw(webview.inner.manager())
-                  as *mut objc2::runtime::AnyObject
+                manager: Retained::as_ptr(&manager) as *const std::ffi::c_void
                   as *mut std::ffi::c_void,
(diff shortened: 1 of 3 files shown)


krishpranav:
Hey @Legend-Master

Just a small fix for the Apple WithWebview pointer handoff. Swapped Retained::into_raw for a scoped Retained + as_ptr to keep ownership balanced. Also fixed the .changes metadata so the checks pass cleanly. No API changes or anything big. Attached a stress run with stable RSS.

Lemme know!

sftse:
Transforming a `&T` into `*mut T` through `*const T` is not a good pattern and can easily lead to UB, although it's probably safe here as they point to an ObjC `Arc` which doesn't have sensible mutation. Could fix this by changing the pointers in the `WebView` to `*const`.

krishpranav:
Just a follow-up, any updates?

min-median-max:
Hi @krishpranav — are you still working on this?

Asking because we're seeing the same leak in a shipping app. On macOS each `with_webview` call moves the WKWebView / manager / NSWindow out via `Retained::into_raw` and never recovers them, leaking one retain each — for us that keeps the WebContent process alive for the whole lifetime of the app; closing a view never releases it, so they accumulate. We confirmed it's the runtime and not our own misuse (no leak with our native hooks off; per-view PID deltas show zero survivors only after the `into_raw` → `as_ptr` change). So this fix matters to us too.

On @sftse's cast note — we borrowed with `Retained::as_ptr` and made the `Webview` pointer fields `*const`, which avoids the `*mut` cast without touching the public accessors.

Would you be able to update the PR along those lines?

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -3903,25 +3903,28 @@ fn handle_user_message<T: UserEvent>(
             #[cfg(target_os = "macos")]
             {
               use wry::WebViewExtMacOS;
+              let platform_webview = webview.webview();
+              let manager = webview.manager();
+              let ns_window = webview.ns_window();
               f(Webview {
-                webview: Retained::into_raw(webview.webview()) as *mut objc2::runtime::AnyObject
+                webview: Retained::as_ptr(&platform_webview) as *const std::ffi::c_void
```
Legend-Master:
```suggestion
                webview: Retained::as_ptr(&platform_webview).cast_mut()
```

Also the other ones

Legend-Master (dismissed):
Thanks, just a bit of nitpicks and you'll need to sign your commits for me to merge this

Also, let's add a todo comment on those fields to change the types to `*const std::ffi::c_void` in v3, as well as documenting you should not try to mutate those pointer (there shouldn't be a case in normal use cases for you to mutate them anyways)

krishpranav:
Hi, sure will do those changes.

krishpranav:
@Legend-Master rebased and signed all commits, added that TODO/doc comment you asked for should be good for a re-review whenever you get a chance!

Legend-Master:
In case you didn't see

> https://github.com/tauri-apps/tauri/pull/15224#discussion_r3548796668

krishpranav:
@Legend-Master it's done.

Legend-Master (approved):
Thanks

sftse:
This stops the leak for sure because it doesn't take "ownership" of the object anymore, but it's not super clear to me whether this occurs at the cost of a crash. The `Webview` struct has no lifetime constraint and is passed to some opaque closure which could be sending it anywhere. Even if this is currently sound, this looks to be separated from disaster by merely one small, remote refactor.

Legend-Master:
The lifetime is a problem here for sure... Although you should not do, you can send the `PlatformWebview` to outside of the closure which will result in dangling pointers.

Since `tauri-runtime-wry` is not a stable crate, we can change the `Webview` fields, what about we change them to `Retain<...>`  and only do the `as_ptr` in tauri's `PlatformWebview` methods (`PlatformWebview` will be safe while the raw pointers are always unsafe with no lifetime guarantees)? @sftse 

We can also deprecate the old methods and adding in new safe `objc2-*` equivalents like what we did for other platforms.

sftse:
> Since `tauri-runtime-wry` is not a stable crate, we can change the `Webview` fields, what about we change them to `Retain<...>` and only do the `as_ptr` in tauri's `PlatformWebview` methods (`PlatformWebview` will be safe while the raw pointers are always unsafe with no lifetime guarantees)?

Yeah, was thinking along those lines. There's a lot of channels involved with Tauri, and I honestly don't have the full picture of what the lifecycle of the objects are. As a rule of thumb, borrowing and channels interact poorly, it's either borrowing with apis like in `rayon` that are able to close over the lifetime (block until borrow is no longer needed) or channels + refcounting. I'd say this fix went in the wrong direction.

Legend-Master:
Actually, `PlatformWebview` is not `Send` while `with_webview` requires the closure to be `Send` meaning you can't use a channel to move out the `PlatformWebview`. That being said, storing the `objc-*` `Retain<...>`s still have other benefits.

weareoutman:
Hi team, I'm wondering would this fix be backported to tauri v2?

FabianLars:
the PR was merged into the v2 branch. There just haven't been a release on that branch for a while for various reasons. Soon ™️ though.

PR #15241, opened 2026-04-15
fix(codegen): select largest ICO entry instead of first

### LINKED ISSUE #14596: [bug] [Windows] Window icon appears blurry - question about ICO multi-size handling
### Describe the bug

On Windows, I noticed that the runtime window icon (displayed in taskbar and titlebar) appears blurry/pixelated, even when using a properly formatted ICO file containing multiple size entries (16x16, 24x24, 32x32, 48x48, 64x64, 256x256).

**What I found in the code:**

Looking at `crates/tauri-codegen/src/image.rs`, the `new_ico()` function (lines 49-72) seems to only read the first entry from the ICO file:

```rust
pub fn new_ico(root: &TokenStream, icon: &Path) -> EmbeddedAssetsResult<Self> {
    let buf = Self::open(icon);

    let icon_dir = ico::IconDir::read(Cursor::new(&buf))
        .unwrap_or_else(|e| panic!("failed to parse icon {}: {}", icon.display(), e));

    let entry = &icon_dir.entries()[0];  // Only takes the first entry
    let rgba = entry
        .decode()
        .unwrap_or_else(|e| panic!("failed to decode icon {}: {}", icon.display(), e))
        .rgba_data()
        .to_vec();

    Cached::try_from(rgba).map(|cache| Self {
        cache,
        root: root.clone(),
        format: IconFormat::Image {
            width: entry.width(),
            height: entry.height(),
        },
    })
}
```

It appears that only `entries()[0]` is read from the ICO file, and the other size entries are not used. This single RGBA image seems to be used for all icon contexts at runtime.

**My understanding of Windows icon behavior:**

As far as I know, Windows uses different icon sizes for different contexts (e.g., small icons for titlebar, larger icons for taskbar). ICO files support multiple embedded sizes for this reason. I'm wondering if this might be related to the blurry icon issue I'm seeing.

**Is this intentional?** I'd appreciate any clarification on whether this is by design (perhaps for memory/performance reasons) or if it's something that could be improved.

### Reproduction

1. Create an ICO file with multiple sizes (16, 24, 32, 48, 64, 256) using a tool like ImageMagick:
   ```bash
   magick icon.png -define icon:auto-resize=256,64,48,32,24,16 icon.ico
   ```

2. Use this ICO as window icon in a Tauri app

3. Build and run on Windows

4. Observe: The titlebar icon appears blurry/pixelated

5. Compare: Right-click the EXE in Explorer - the file icon is sharp (Windows reads the ICO resource correctly)

**Workaround discovered:**

Creating a Windows shortcut (.lnk) pointing to the EXE makes the **taskbar** icon clear, because Windows shortcuts read the icon from the EXE's embedded resource (which preserves all ICO sizes). However, the **titlebar** icon remains blurry as it uses the runtime RGBA data.

### Expected behavior

I expected the window icon to appear sharp/crisp, similar to how the file icon appears in Windows Explorer (which reads the ICO resource directly and displays the appropriate size).

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.26200 x86_64 (X64)
    ✔ WebView2: 142.0.3595.94
    ✔ rustc: 1.91.1 (ed61e7d7e 2025-11-07)
    ✔ cargo: 1.91.1 (ea2d97820 2025-10-10)
    ✔ Rust toolchain: 1.91.1-x86_64-pc-windows-msvc
    - node: 24.11.1
    - pnpm: 10.20.0
    - npm: 11.6.2

[-] Packages
    - @tauri-apps/cli: ^2.9.0
    - @tauri-apps/api: ^2.9.0
    (via Pake CLI v3.5.3)

[-] Plugins

[-] App
```

### Stack trace

```text

```

### Additional context

_No response_

### LINKED ISSUE #15269: [bug] [Windows] Task Manager child window icons are severely corrupted (Win11)
### Describe the bug

Win11:
reinamanager:
Image

clash verge rev:
Image

yaak:
Image

I don't seem to have encountered this problem on Windows 10.

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
n/a
```

### Stack trace

```text

```

### Additional context

_No response_

Razshy:
Closes #14596

`new_ico()` in `tauri-codegen` uses `entries()[0]` to select which image from an ICO file to embed as the window icon. Since ICO files conventionally store entries smallest-first, this picks the 16x16 image. When the OS needs a larger size (taskbar, alt-tab), it scales up from 16x16, resulting in a blurry icon.

This selects the largest entry instead, giving the OS a high-resolution source to downscale from.

- One line changed in `crates/tauri-codegen/src/image.rs`
- No new dependencies
- Passes `rustfmt --check` and `cargo check -p tauri-codegen`

### DIFF
--- a/crates/tauri-codegen/src/image.rs
+++ b/crates/tauri-codegen/src/image.rs
@@ -54,7 +54,11 @@ impl CachedIcon {
     let icon_dir = ico::IconDir::read(Cursor::new(&buf))
       .unwrap_or_else(|e| panic!("failed to parse icon {}: {}", icon.display(), e));
 
-    let entry = icon_dir.entries().iter().max_by_key(|e| e.width()).unwrap();
+    let entry = icon_dir
+      .entries()
+      .iter()
+      .min_by_key(|e| e.width().abs_diff(32))
+      .unwrap();
     let rgba = entry
       .decode()
       .unwrap_or_else(|e| panic!("failed to decode icon {}: {}", icon.display(), e))
--- a/crates/tauri-codegen/src/image.rs
+++ b/crates/tauri-codegen/src/image.rs
@@ -54,11 +54,7 @@ impl CachedIcon {
     let icon_dir = ico::IconDir::read(Cursor::new(&buf))
       .unwrap_or_else(|e| panic!("failed to parse icon {}: {}", icon.display(), e));
 
-    let entry = icon_dir
-      .entries()
-      .iter()
-      .min_by_key(|e| e.width().abs_diff(32))
-      .unwrap();
+    let entry = icon_dir.entries().iter().max_by_key(|e| e.width()).unwrap();
     let rgba = entry
       .decode()
       .unwrap_or_else(|e| panic!("failed to decode icon {}: {}", icon.display(), e))
--- a/crates/tauri-codegen/src/image.rs
+++ b/crates/tauri-codegen/src/image.rs
@@ -54,7 +54,7 @@ impl CachedIcon {
     let icon_dir = ico::IconDir::read(Cursor::new(&buf))
       .unwrap_or_else(|e| panic!("failed to parse icon {}: {}", icon.display(), e));
 
-    let entry = &icon_dir.entries()[0];
+    let entry = icon_dir.entries().iter().max_by_key(|e| e.width()).unwrap();
     let rgba = entry
       .decode()
       .unwrap_or_else(|e| panic!("failed to decode icon {}: {}", icon.display(), e))
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -48,7 +48,7 @@ use tao::platform::macos::{EventLoopWindowTargetExtMacOS, WindowBuilderExtMacOS}
 ))]
 use tao::platform::unix::{WindowBuilderExtUnix, WindowExtUnix};
 #[cfg(windows)]
-use tao::platform::windows::{IconExtWindows, WindowBuilderExtWindows, WindowExtWindows};
+use tao::platform::windows::{WindowBuilderExtWindows, WindowExtWindows};
 #[cfg(windows)]
 use webview2_com::{ContainsFullScreenElementChangedEventHandler, FocusChangedEventHandler};
 #[cfg(windows)]
@@ -1241,14 +1241,6 @@ impl WindowBuilder for WindowBuilderWrapper {
     self.inner = self
       .inner
       .with_window_icon(Some(TaoIcon::try_from(icon)?.0));
-
-    #[cfg(windows)]
-    {
-      if let Ok(resource_icon) = TaoWindowIcon::from_resource(32512, None) {
-        self.inner = self.inner.with_taskbar_icon(Some(resource_icon));
-      }
-    }
-
     Ok(self)
   }
 
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -48,7 +48,7 @@ use tao::platform::macos::{EventLoopWindowTargetExtMacOS, WindowBuilderExtMacOS}
 ))]
 use tao::platform::unix::{WindowBuilderExtUnix, WindowExtUnix};
 #[cfg(windows)]
-use tao::platform::windows::{WindowBuilderExtWindows, WindowExtWindows};
+use tao::platform::windows::{IconExtWindows, WindowBuilderExtWindows, WindowExtWindows};
 #[cfg(windows)]
 use webview2_com::{ContainsFullScreenElementChangedEventHandler, FocusChangedEventHandler};
 #[cfg(windows)]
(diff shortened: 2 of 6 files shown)


Enigmatikk (approved):

Legend-Master:
Thanks for contributing! Using the biggest image in the `.ico` is probably the correct thing to do, but I am a bit worried about people will now ship raw bytes of a large image by default (256 × 256 × 4 = 262,144 bytes for example)

It seems like the way this was somewhat tackled way back in https://github.com/tauri-apps/tauri/pull/4992/changes/a80d5b7138f993f86d891c8e96cd3c0907d569d1 by putting the 32x32 one to be the first entry 😂

Perhaps we could make it select the biggest entry under say 32x32 or 64x64 (64 × 64 × 4 = 16,384 bytes) for now, this should be a sufficient sized icon for the taskbar icons on most devices, and ideally we could add support of using the embedded resource icons as mentioned in https://github.com/tauri-apps/tauri/issues/14596#issuecomment-3606515455

FabianLars:
we should also look into using tao's https://docs.rs/tao/latest/x86_64-pc-windows-msvc/tao/platform/windows/trait.WindowExtWindows.html#tymethod.set_taskbar_icon for large icons since you mentioned alt-tab which will use this one.
We currently only set the small icon, which as it name suggests should be somewhat small (as it's used in the window titlebar for example)

Razshy:
Ok so Instead of using a decoded RGBA buffer, `tauri-runtime-wry` loads both `ICON_SMALL` and `ICON_BIG` directly from the exe embedded Ico resource `IconExtWindows::from_resource`  so Windows picks the right size for (titlebar, taskbar, alt-tab). It falls back to the existing RGBA path if resource loading fails. @Legend-Master @FabianLars

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -1219,6 +1219,19 @@ impl WindowBuilder for WindowBuilderWrapper {
   }
 
   fn icon(mut self, icon: Icon) -> Result<Self> {
+    #[cfg(windows)]
+    {
+      let small = TaoWindowIcon::from_resource(32512, Some(TaoPhysicalSize::new(16, 16)));
```
Legend-Master: This would prevent the user from setting an icon themselves

Razshy: Good point

I changed ICON_SMALL to use the users customization and runtime `set_icon`
 
And the resource is only used for ICON_BIG which wasn't set at all before form what I know. Also capped the codegen RGBA icon at 32x32 to keep the binary small. 

Also i think exposing a public `set_taskbar_icon` API could be a follow up PR.

crates/tauri-codegen/src/image.rs
```diff
@@ -54,7 +54,11 @@ impl CachedIcon {
     let icon_dir = ico::IconDir::read(Cursor::new(&buf))
       .unwrap_or_else(|e| panic!("failed to parse icon {}: {}", icon.display(), e));
 
-    let entry = &icon_dir.entries()[0];
+    let entry = icon_dir
+      .entries()
+      .iter()
+      .min_by_key(|e| e.width().abs_diff(32))
```
Legend-Master: I just opened #15274, with that, I think we might be able to remove this limit here

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -1222,6 +1222,14 @@ impl WindowBuilder for WindowBuilderWrapper {
     self.inner = self
       .inner
       .with_window_icon(Some(TaoIcon::try_from(icon)?.0));
+
+    #[cfg(windows)]
+    {
+      if let Ok(resource_icon) = TaoWindowIcon::from_resource(32512, None) {
```
Legend-Master: Not sure if this is a good default, the task bar icon seems to fallback from `ICON_BIG > ICON_SMALL > exe icon` 🤔

And I don't think this belongs to here, it should be in the build method

Legend-Master (dismissed):
Thanks! I will merge this one after #15274 so we don't embed huge images

> https://github.com/tauri-apps/tauri/pull/15241#issuecomment-4279915579

crates/tauri/src/image/mod.rs
```diff
@@ -121,6 +126,62 @@ unsafe fn read_bgra(hbm: HBITMAP, width: i32, height: i32) -> crate::Result<Vec<
...
 }
 
+/// Reads the `RT_GROUP_ICON` directory of an icon resource and returns the dimensions of
+/// its largest image.
+///
+/// # Safety
+///
+/// `resource_id` must be a valid resource name or `MAKEINTRESOURCE` id.
+#[cfg(windows)]
+unsafe fn largest_icon_size(module: HMODULE, resource_id: PCWSTR) -> crate::Result<(u32, u32)> {
```
Legend-Master: @lucasfernog I don't think we want to always use the largest entry as default here, by default we only use it for things like task bar icon and some other even smaller things. The default ICO can and should contain a large entry to be displayed in file explorer.

> https://github.com/tauri-apps/tauri/pull/15274#discussion_r3117111753

PR #15242, opened 2026-04-15
Improve NSIS installer script

bovirus:
- Add for each installer windows on top left after program name the info about program version (currently show only program name)
- Add in exe file properties for File description "Program name" + "installer" (currently is only program name and is a duplicated info because "Program Name" is also in other place)

### DIFF
--- a/crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
+++ b/crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
@@ -70,7 +70,7 @@ Var NoShortcutMode
 Var WixMode
 Var OldMainBinaryName
 
-Name "${PRODUCTNAME}"
+Name "${PRODUCTNAME} ${VERSION}"
 BrandingText "${COPYRIGHT}"
 OutFile "${OUTFILE}"
 
@@ -82,7 +82,7 @@ InstallDir "${PLACEHOLDER_INSTALL_DIR}"
 
 VIProductVersion "${VERSIONWITHBUILD}"
 VIAddVersionKey "ProductName" "${PRODUCTNAME}"
-VIAddVersionKey "FileDescription" "${PRODUCTNAME}"
+VIAddVersionKey "FileDescription" "${PRODUCTNAME} installer"
 VIAddVersionKey "LegalCopyright" "${COPYRIGHT}"
 VIAddVersionKey "FileVersion" "${VERSION}"
 VIAddVersionKey "ProductVersion" "${VERSION}"


bovirus:
**Exe file properties - File description**

Currently with tauri and NSIS Product Name nd Product description are the same

Example by Open Video Downloader (compiled with tauri and NSIS)

image

With my proposal is like Mp3tag installer (file description is "Program name" + "installer")

image

bovirus:
**Caption - Program version showed after program name**

Example: Open VideoDownloader installer use tauri and NSIS script and don't show the version info

image

With my prosal changes is like Mp3tag installer - caption showed after program name the info about program version

image

Legend-Master:
So just a heads up, I am quite busy at the moment to review this. I will take a look when I get some time off

bovirus:
@Legend-Master 

My proposal are very simple and based on official NSIS documentaion.
My suggestion are based on experience about installer created with NSIS.

Currently the project based on tauri (like Open Video Downlaoder) don't show (as explained with screenshots) on top left program version and don't show in exe file properties the right info for file description (mssing "installer" string).

To enable all projects based on tauri to get the benefits of these cimple changes (that I hope are quite clear) I believe that you can simply add it.
I didn't see negative points about these changes.

crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
```diff
@@ -70,7 +70,7 @@ Var NoShortcutMode
 Var WixMode
 Var OldMainBinaryName
 
-Name "${PRODUCTNAME}"
+Name "${PRODUCTNAME} ${VERSION}"
```
Legend-Master, line 73: Looking at what this will affect, I don't think we should change the `Name` to something other than a plain `${PRODUCTNAME}`

> https://github.com/NSIS-Dev/nsis/blob/5b9af08eae993ba6f12a27a95f0bfc533279e2c8/Source/lang.cpp#L50

crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
```diff
@@ -82,7 +82,7 @@ InstallDir "${PLACEHOLDER_INSTALL_DIR}"
 
 VIProductVersion "${VERSIONWITHBUILD}"
 VIAddVersionKey "ProductName" "${PRODUCTNAME}"
-VIAddVersionKey "FileDescription" "${PRODUCTNAME}"
+VIAddVersionKey "FileDescription" "${PRODUCTNAME} installer"
```
Legend-Master, line 85: this change looks good to me, do you know if we can translate this?

crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
bovirus: @Legend-Master 

I believe that you have a misunderstanding about ${PRODUCTNAME}, ${VERSION} and Name.
Please check NSIS documentation.
Name (gerenic - don't mean productname) is a variable that will be replaced with the relative value only on top left of the installer (and where you use ^Name in the strings).
Are two different value for variable.
${PRODUCTNAME} is only ProductName
${VERSION} is only ProductVersion
Name is typically ProductName+Version and should replace in the context that I explained.

Then no negative points to use it.
Most of the installer based on NSIS use 
Name "${PRODUCTNAME} ${VERSION}"

crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
bovirus: I believe that could be translated.
But typically is leaved in English.
Becasue the installer select the right term in local language when you select language at first step of the installer.
But didn't work wiuth exe -> right click -> File properties (you don't select a language).

crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
Legend-Master: No, you can this from the screenshot in your comments, many other 'Mp3tag' got replaced by 'Mp3tag v3.34'

> https://github.com/tauri-apps/tauri/pull/15242#issuecomment-4250049901

crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
Legend-Master: I believe this can be done through adding something like

`VIAddVersionKey /LANG=${LANG_SPANISH} "FileDescription" "${PRODUCTNAME} instalar"` to `Spanish.nsh`
`VIAddVersionKey /LANG=${LANG_SIMPCHINESE} "FileDescription" "${PRODUCTNAME} 安装程序"` to `SimpChinese.nsh`

bovirus: @Legend-Master 

Yoiu can setup any function that you like but I beleive (you can test it), that the strings should be localzied when you run the instalelr and select the language.
Without this the strings remain in English.
Then when you make right click -> Properties trhe instalelr is not run and teh strings is in english.
All program that I know based on SNSI have the string in fiel properties as "instalelr" in English.

crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
bovirus: @Legend-Master 

Example by official doc provided by NSIS (Chapter2)

https://nsis.sourceforge.io/Docs/Chapter2.html

!define VERSION "1.0.3"
Name "My Program ${VERSION}"
OutFile "My Program Installer - ${VERSION}.exe"
 
that could be setiing up program name with a variable

!define VERSION "1.0.3"
!define PROGRAM_NAME "tauri"
Name "${PROGRAM_NAME} ${VERSION}"
OutFile "${PROGRAM_NAME} ${VERSION} installer.exe"

crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
Legend-Master: With the code above, I was able to make it show the version based on the system locale tested on my machine, so this is definitely doable

crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
Legend-Master: Not sure if I get what you meant, I am saying this name should be the product name because it is used everywhere NSIS uses the app name, you won't want that to always contain the versions

crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
bovirus: @Legend-Master 

In NSIS There are many way to setup the function (I posted below the offcciail doc by NSIS).
Please use what is for you the ebst.

My scope is to add the fuctions that I asked

- Show program version of program name on top left of each installer window
- Change the file description ad"ProgramName Installer".

Thanks.

Legend-Master: I would like a have a more complete way of doing this before we ship it to all tauri apps, that's why I don't want to merge straight away.
And I don't really have enough time to make the whole change and do all the testing to make sure they work as expected (especially NSIS staff) right now

crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
bovirus: @Legend-Master 

As you can see in the example that I provided that is coming from offciial NSIS documentation is the same NSIS developer that suggest to use in this way. It's a logical use of this.

It's a trick to use a string that is combinationation of ProductName + ProductVersion (not exist as direct variable that is ProductName + ProductVersion).

If the official developer suggest to use in this way why didn't do this?

Legend-Master: As I said, we should not replace all app names with app names + versions. And the example you're pointing to is a show case of the conditional compilation feature what just happens to use

> ```
> !define VERSION "1.0.3"
> Name "My Program ${VERSION}"
> ```

If you look at the link to the actual docs about the `Name`, it says

> Sets the name of the installer. The name is usually simply the product name such as 'MyApp' or 'CrapSoft MyApp'.
> https://nsis.sourceforge.io/Docs/Chapter4.html#aname

bovirus: @Legend-Master 

The description in NSIS doc not consider that doing that you will not have version on top left for each instalelr windows.
To change "Name" as productName+ProduyctVersion (as expolained in the example) is the best and easy solution that didn't required addition strings adding (for each language).

Could be for you a possible solution without touch "Name"?

LangString CaptionText ${LANG_ENGLISH} "${PRODUCT_NAME} ${VERSION} Installer"
LangString CaptionText ${LANG_ITALIAN} "Installazione di ${PRODUCT_NAME} ${VERSION}"
etc etc for other language
Caption "$(CaptionText)"

Just setup caption title.

crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
bovirus: @Legend-Master 

I beleive that you can start just make for the english and use correct file description (now is only prioduct name and is not right).

I posted a possible solution to make without toch Name variable.

bovirus: @Legend-Master 

example of use of NSI - "BUtil" program

https://github.com/drweb86/butil/blob/master/sources/setup.nsi

; Installer attributes
Name "${PRODUCT_NAME} ${PRODUCT_VERSION}"

bovirus: @Legend-Master 

As suggest by Anders (NSIS developer) you can setup in this way

```
LangString ^SetupCaption ${LANG_ENGLISH} "Setup ${PRODUCTNAME} ${VERSION}"
LangString ^SetupCaption ${LANG_ITALIAN} "Installazione di ${PRODUCTNAME} ${VERSION}"
LangString ^UninstallCaption ${LANG_ENGLISH} "Uninstall ${PRODUCTNAME} ${VERSION}"
LangString ^UninstallCaption ${LANG_ITALIAN} "Disinstallazione di ${PRODUCTNAME} ${VERSION}"
```

PR #15252, opened 2026-04-16
fix(cef): close all browsers before shutdown to fix null pointer dereference on macOS

amrbashir:

### DIFF
--- a/crates/tauri-runtime-cef/src/lib.rs
+++ b/crates/tauri-runtime-cef/src/lib.rs
@@ -1988,7 +1988,6 @@ impl<T: UserEvent> CefRuntime<T> {
 
             tx.send(objc2_app_kit::NSApplicationTerminateReply::TerminateCancel)
               .unwrap();
-            cef_impl::close_all_windows(&windows_);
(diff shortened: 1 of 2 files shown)


crates/tauri-runtime-cef/src/lib.rs
```diff
@@ -1973,10 +1974,21 @@ impl<T: UserEvent> CefRuntime<T> {
...
+            // Start closing all browsers (including devtools). The actual
+            // destruction is async and completes via the CEF message loop.
+            //
+            // Signal the main loop to exit. The post-loop safety net will
+            // pump the message loop until all browsers are fully destroyed
+            // before calling cef::shutdown().
+
             tx.send(objc2_app_kit::NSApplicationTerminateReply::TerminateCancel)
               .unwrap();
+            cef_impl::close_all_windows(&windows_);
```
lucasfernog: I think this is redundant, the exit below will break the event loop and run the rest of the cleanup, tho I didn't check

lucasfernog (approved):

PR #15255, opened 2026-04-16
feat(core): add android activityEmbedding config

lucasfernog:
automatically configures the project to support Activity embedding and generate split rules from config


lucasfernog:
follow-up for #14484

.changes/android-embedded-activity-config.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri-utils": minor:feat
+---
+
+Added `tauri.conf.json > bundle > android > activityEmbedding` config to enable Activity embedding and define split rules for Android multi-window support.
```
velocitysystems, line 5: The new splitRules entry declares `"secondary": "DetailActivity"`, but no `DetailActivity` Kotlin class is generated by tauri-build or declared in the example. Is the expectation this too is generated or is manually supplied by the user?

crates/tauri-build/src/mobile.rs
```diff
@@ -199,3 +215,256 @@ dependencies {
...
+     }\n",
+  );
+
+  let initializer_path = source_dir.join("TauriSplitInitializer.kt");
+  write_if_changed(&initializer_path, kt).context("failed to write TauriSplitInitializer.kt")?;
+
+  Ok(())
+}
+
+fn append_split_rule(kt: &mut String, i: usize, rule: &SplitPairRule, package_name: &str) {
```
velocitysystems, line 341: Both `SplitType::Ratio(f64)` and `EmbeddingAspectRatio::Ratio(f64)` variants accept a bare `f64`. The comments document the SDK's contract (0.0 < r < 1.0 for SplitType::Ratio, > 1.0 for EmbeddingAspectRatio::Ratio), but nothing enforces it. Should this be enforced?

Nit: Kotlin treats values like `$identifier` and `${expr}` as string interpolation. Should we guard against this? (Even if these are invalid for this use-case).

PR #15262, opened 2026-04-19
Non-breaking version of PR 15127

sftse:
Supersedes #15127


Legend-Master (approved):
I don't mind this change either ways to be honest, but I think in v3, we should still make `Keys::decrypt` take a reference here and go with #15127

PR #15263, opened 2026-04-19
feat(windows): append .rc content support.

### LINKED ISSUE #14784: [feat] Allow extending Windows resources.rc file
### Describe the problem

The current state of Tauri does not easily allow users to integrate additional resource files to link against the executable.

The current solution to embed additional resources is to use the [`embed-resource`](https://crates.io/crates/embed-resource) crate during the build phase to compile and link additional resources. This solution works for most of the cases.

However, the default resource.rc file produced by Tauri includes an ICON definition (which is mandatory for having an app's icon), which prevents including other resources that also contain an ICON declaration due to a linker fatal error about duplicated resources related to the ICON declarations. The current solution is to manually append the desired ICON entry to the `resource.rc` file generated by Tauri and rebuild it during the build phase.
Another solution may be to add the `/TLBID` flag to the linker to specify a correct resource ID for the corresponding resource file, but this solution is suboptimal and was not tested. More, the documentation does not mention the content of the resource file generated by Tauri, which may introduce undefined behavior or errors during the build with other resource files.

`tauri-winres` exposes the [`append_rc_content`](https://docs.rs/tauri-winres/0.3.5/tauri_winres/struct.WindowsResource.html#method.append_rc_content) method, which allows merging multiple resource files. However, this isn't available via `tauri-build` nor `tauri.conf.json` file. One solution would be to allow providing a list of additional resources in the `tauri.conf.json` file.

### Alternatives considered

Alternatively, allowing multiple icons to be linked against the executable on Windows should also resolve the problem. In this case, adding a note on the documentation on the recommended way of adding additional resources seems a good idea! 

### Additional context

Having multiple icons inside the executable is a standard practice on Windows, notably when adding an entry to the jumplist menu with an alternative icon than the one used by the app executable.

Kokoro2336:
This pr closes issue #14784 .

Allow users to append extra .rc content by a new api in `WindowsAttributes`.

### DIFF
--- a/crates/tauri-build/src/lib.rs
+++ b/crates/tauri-build/src/lib.rs
@@ -245,7 +245,7 @@ pub struct WindowsAttributes {
   /// A string containing additional .rc content to be appended to the generated resource file on Windows.
   ///
   /// Defaults to `None`.
-  append_rc_content: Option<String>,
+  append_rc_content: Vec<String>,
 }
 
 impl Default for WindowsAttributes {
@@ -260,7 +260,7 @@ impl WindowsAttributes {
     Self {
       window_icon_path: Default::default(),
       app_manifest: Some(include_str!("windows-app-manifest.xml").into()),
-      append_rc_content: None,
+      append_rc_content: Vec::new(),
     }
   }
 
@@ -270,7 +270,7 @@ impl WindowsAttributes {
     Self {
       app_manifest: None,
       window_icon_path: Default::default(),
-      append_rc_content: None,
+      append_rc_content: Vec::new(),
     }
   }
 
@@ -345,7 +345,7 @@ impl WindowsAttributes {
   /// This can be called multiple times to append multiple contents.
   #[must_use]
   pub fn append_rc_content<S: Into<String>>(mut self, content: S) -> Self {
-    self.append_rc_content = Some(content.into());
+    self.append_rc_content.push(content.into());
     self
   }
 }
@@ -627,8 +627,8 @@ pub fn try_build(attributes: Attributes) -> Result<()> {
       res.set_manifest(&manifest);
     }
 
-    if let Some(append_rc_content) = attributes.windows_attributes.append_rc_content {
-      res.append_rc_content(&append_rc_content);
+    for content in attributes.windows_attributes.append_rc_content {
+      res.append_rc_content(&content);
     }
 
     if let Some(version_str) = &config.version {
--- a/crates/tauri-build/src/lib.rs
+++ b/crates/tauri-build/src/lib.rs
@@ -341,6 +341,8 @@ impl WindowsAttributes {
     self
   }
 
+  /// Append additional .rc content to the generated resource file on Windows.
+  /// This can be called multiple times to append multiple contents.
   #[must_use]
   pub fn append_rc_content<S: AsRef<str>>(mut self, content: S) -> Self {
     self.append_rc_content = Some(content.as_ref().to_string());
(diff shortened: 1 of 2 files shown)


Kokoro2336:
Weixin Image_20260419224635_28_89
Weixin Image_20260419225036_29_89
Weixin Image_20260419225050_30_89

I've made a simple test on the api. For simplicity I used `Cargo.toml` under `src-tauri` directly. I tried to use `strings` to read the content linked into the `.exe`, and it indicated that the content in `Cargo.toml` had been successfully written.

crates/tauri-build/src/lib.rs
```diff
@@ -334,6 +340,14 @@ impl WindowsAttributes {
     self.app_manifest = Some(manifest.as_ref().to_string());
     self
   }
+
+  /// Append additional .rc content to the generated resource file on Windows.
+  /// This can be called multiple times to append multiple contents.
+  #[must_use]
+  pub fn append_rc_content<S: AsRef<str>>(mut self, content: S) -> Self {
+    self.append_rc_content = Some(content.as_ref().to_string());
```
Legend-Master:
```suggestion
  pub fn append_rc_content<S: Into<String>>(mut self, content: S) -> Self {
    self.append_rc_content = Some(content.into());
```

crates/tauri-build/src/lib.rs
```diff
@@ -334,6 +340,14 @@ impl WindowsAttributes {
     self.app_manifest = Some(manifest.as_ref().to_string());
     self
   }
+
+  /// Append additional .rc content to the generated resource file on Windows.
+  /// This can be called multiple times to append multiple contents.
```
Legend-Master, line 343: This does not seem to be the case, change our internal `append_rc_content` to be a `Vec` and actually implement this or remove this doc

Legend-Master (changes requested):

crates/tauri-build/src/lib.rs
Kokoro2336: you mean a `Vec<String>` rather than a single `Option<String>`?

Legend-Master: Yeah

Kokoro2336: But if you take a look at the inner impl of `append_rc_content` in [winres](https://github.com/tauri-apps/winres/blob/f348f6b94eee3d1db5e0627f8de802165f2cac6d/src/lib.rs#L506) and you will find that the field continuously receives trailing `&str` by calling the api multiple times. So I think a single `Option<String>` it's acceptable?

Legend-Master: Yeah, but we only ever called that function once, no?

Kokoro2336: Ok. I've now updated the logic.

crates/tauri-build/src/lib.rs
```diff
@@ -242,6 +242,10 @@ pub struct WindowsAttributes {
   ///
   /// [application manifest]: https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests
   app_manifest: Option<String>,
+  /// A string containing additional .rc content to be appended to the generated resource file on Windows.
```
Legend-Master: Let's also update the description here

Kokoro2336: updated

Legend-Master (approved):
Thanks

Kokoro2336:
Good！

PR #15266, opened 2026-04-20
fix(tauri): enforce ACL for remote origins even without AppManifest

mtsgrd:
## Summary

Custom (non-plugin) commands registered via `invoke_handler` bypass ACL checks entirely when no `AppManifest` is configured in `build.rs`. This means if a webview navigates away from the bundled frontend to a remote URL, that remote page can invoke any custom IPC command without restriction.

This PR adds `|| !is_local` to the existing ACL enforcement condition in `Webview::on_message()`, so that requests from non-local origins always go through `resolve_access()`:

```rust
// Before:
if (plugin_command.is_some() || has_app_acl_manifest)

// After:
if (plugin_command.is_some() || has_app_acl_manifest || !is_local)
```

### Behavior after this change

| Origin | AppManifest configured? | Result |
|--------|------------------------|--------|
| Local | No | **Unchanged** - ACL skipped, commands work (permissive default) |
| Local | Yes | **Unchanged** - ACL enforced as before |
| Remote | No | **Changed** - ACL enforced, no entries found, rejected |
| Remote | Yes, with explicit `remote` capability | **Unchanged** - `resolve_access` matches, allowed |

### Breaking change

Apps that intentionally allow remote origins to call custom commands **without** configuring an `AppManifest` will need to add an `AppManifest` with explicit `remote` capabilities to restore the previous behavior. This is expected to be uncommon, since most apps either use local-only webviews or already configure an `AppManifest` when working with remote content.

## Test plan

- Added `remote_origin_blocked_for_custom_commands_without_app_manifest` unit test that verifies:
  - A custom command from `https://some-hostile-domain.com` is rejected
  - The same command from `tauri://localhost` is not blocked by the remote-origin guard
- All 53 existing `tauri` lib tests continue to pass

### DIFF
--- a/.changes/enforce-acl-remote-origins.md
+++ b/.changes/enforce-acl-remote-origins.md
@@ -1,0 +1,5 @@
+---
+'tauri': 'patch:sec'
+---
+
+Enforce ACL checks for IPC requests from remote origins even when no `AppManifest` is configured. Previously, custom (non-plugin) commands bypassed ACL entirely without an `AppManifest`, allowing any origin to invoke them. Now, remote origins are always subject to ACL resolution, and can only reach custom commands if an explicit `remote` capability has been granted.
--- a/crates/tauri/src/test/mod.rs
+++ b/crates/tauri/src/test/mod.rs
@@ -216,7 +216,11 @@ pub fn mock_app() -> App<MockRuntime> {
 ///             cmd: "ping".into(),
 ///             callback: tauri::ipc::CallbackFn(0),
 ///             error: tauri::ipc::CallbackFn(1),
-///             url: "http://tauri.localhost".parse().unwrap(),
+///             url: if cfg!(any(windows, target_os = "android")) {
+///                 "http://tauri.localhost"
+///             } else {
+///                 "tauri://localhost"
+///             }.parse().unwrap(),
 ///             body: tauri::ipc::InvokeBody::default(),
 ///             headers: Default::default(),
 ///             invoke_key: tauri::test::INVOKE_KEY.to_string(),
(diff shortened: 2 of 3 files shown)


tweidinger (dismissed):
Just reviewed the code without testing. Breaking should be fine since unintentional and security related. LGTM and thanks!
Maybe next time via the disclosure feature though 😅

tweidinger:
Please fix tests/fmt/clippy before merge @mtsgrd or @FabianLars . I'm on mobile atm.

mtsgrd:
@tweidinger thanks for reviewing! Let me know if there's anything else I can do?

> Maybe next time via the disclosure feature though 😅

Indeed, that would have been better.

mtsgrd:
@tweidinger how can we move this pr forward?

Qodo-Free-For-OSS:
Hi, The new unit test hard-codes `tauri://localhost` as a local origin, but on Windows/Android local origins are represented as `http(s)://tauri.localhost`, so the test’s “local” request can be classified as remote and incorrectly blocked.

**Severity:** action required | **Category:** correctness

**How to fix:** Use platform-specific local URL

> ### Issue description
> `remote_origin_blocked_for_custom_commands_without_app_manifest()` uses `tauri://localhost` as the local origin. On Windows/Android, Tauri local origins are represented as `http(s)://tauri.localhost`, so the test can misclassify the request as remote.
> 
> ### Issue Context
> `Webview::is_local_url()` has platform-specific logic for custom protocols on Windows/Android.
> 
> ### Fix Focus Areas
> - crates/tauri/src/webview/mod.rs[2344-2397]
> 
> ### Suggested change
> In the test, build the local URL similarly to the doctest helpers:
> ```rust
> let local_url: url::Url = if cfg!(any(windows, target_os = "android")) {
>   "http://tauri.localhost"
> } else {
>   "tauri://localhost"
> }.parse().unwrap();
> ```
> Use `local_url` for the local request.

---

tweidinger:
Hey, I am traveling atm. Poking @FabianLars for getting this merged in dev.

FabianLars (approved):

PR #15274, opened 2026-04-21
refactor(windows): read `default_window_icon` from app resources

Legend-Master:
Reference #14596, #15241

Read `default_window_icon` `Image` from the app's resources so we don't need to embed the raw pixels

We should look into keeping the resource `HICON` handle instead of reading out the RGBA values and then creating new icons in tao/muda/tray-icon, it will require some API changes around `default_window_icon` and or `Image` though


crates/tauri-codegen/src/context.rs
```diff
@@ -211,25 +211,7 @@ pub fn context_codegen(data: ContextData) -> EmbeddedAssetsResult<TokenStream> {
...
-        let icon_path = find_icon(
-          &config,
-          &config_parent,
-          |i| i.ends_with(".png"),
-          "icons/icon.png",
-        );
-        let icon = CachedIcon::new(&root, &icon_path)?;
-        quote!(::std::option::Option::Some(#icon))
-      }
+      quote!(#root::image::Image::from_app_icon(64).ok())
```
Legend-Master: Looking at

> https://learn.microsoft.com/en-us/windows/apps/design/iconography/app-icon-construction#icon-scaling

64 (~16KB of memory) should be a ceiling to use, we could go for a bit lower if we want

Legend-Master: Also, because this is most likely to be loaded from the very beginning, `LoadImageW` for the first time will actually take a bit of time (~1.5ms on my machine), not sure if this is worth the trade offs

FabianLars: that sounds pretty _meh_ to be honest

lucasfernog: yikes

Qodo-Free-For-OSS:
Hi, `Image::from_icon_resource` calls `GetIconInfo` but never frees `ICONINFO.hbmColor`/`hbmMask`, leaking GDI objects on every call (startup path on Windows via default window icon). This can exhaust the per-process GDI handle limit and cause UI/resource creation failures.

**How to fix:** Free ICONINFO bitmaps

> ### Issue description
> On Windows, `GetIconInfo` returns `ICONINFO` containing `hbmColor` and `hbmMask` bitmap handles that are owned by the caller. `Image::from_icon_resource` currently never releases these handles, leaking GDI objects per invocation.
> 
> ### Issue Context
> This path is now used for the Windows default `default_window_icon` (via `Image::from_app_icon_resource(...).ok()`), so the leak can occur during normal app startup.
> 
> ### Fix Focus Areas
> - crates/tauri/src/image/mod.rs[141-207]
> 
> ### What to change
> - Ensure `icon_info.hbmColor` and `icon_info.hbmMask` are released with `DeleteObject` on all exit paths (success and error).
> - Prefer a small RAII guard (custom struct with `Drop`) to avoid missing cleanup on early returns.
> - Add the necessary Windows import for `DeleteObject` (and any handle types needed).

---
*Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review. FYI, Qodo is free for open-source.*

lucasfernog (approved):

FabianLars:
Was this solved? https://github.com/tauri-apps/tauri/pull/15274#discussion_r3118779655

Also wonder how much the fairly large amount of added code negates the icon space savings.

Legend-Master:
> Also wonder how much the fairly large amount of added code negates the icon space savings.

image

Seems like it's actually a problem...

Edit: or maybe not, seeing the same jump on Linux as well while all the code here are gated on Windows

image

PR #15276, opened 2026-04-21
Follow up to PR15262

sftse:
Follow up #15262. Most in the structs in `tauri-utils/isolation.rs` are only used in `tauri`, could be a candidate to split and move for v3.


sftse:
Does not work as expected, fails at runtime.

Issue #15277, opened 2026-04-21
Linux/WebKitGTK: getUserMedia(audio) fails with NotAllowedError even with permission-request allow hook


AmeerJ97:
## Summary
On Linux (Ubuntu Neon / KDE Plasma / Wayland), a minimal Tauri app using WebKitGTK fails to access microphone input via `navigator.mediaDevices.getUserMedia({ audio: true })` with:

`NotAllowedError: The request is not allowed by the user agent or the platform in the current context.`

This still happens even when a native WebKitGTK `permission-request` handler is installed and explicitly calls `request.allow()`.

## Reproducer
Minimal public repro repo:
https://github.com/AmeerJ97/tauri-linux-mic-permission-repro

## Environment
- OS: Ubuntu Neon / KDE Plasma / Wayland 6.6
- Tauri JS: 2.5.0
- Tauri Rust: 2.5.6
- WebKitGTK: 2.50.4
- Node: 22.22.2
- Rust: 1.95.0

## Repro steps
1. Clone the repro repo
2. `npm install`
3. `npm run tauri -- dev` or build and run the app
4. Click `Request Microphone`

## Expected behavior
Either:
- WebKitGTK permission prompt appears and microphone access succeeds, or
- the native `permission-request` allow hook is sufficient for microphone access

## Actual behavior
The app returns:
- `NotAllowedError: The request is not allowed by the user agent or the platform in the current context.`

## Notes
The repro installs a Linux-only native permission hook using Tauri's `with_webview()` and WebKitGTK's `connect_permission_request`, then calls `request.allow()`.

That suggests either:
- WebKitGTK user-media permission is not actually being satisfied by this hook in the Tauri runtime path, or
- additional Linux/WebKitGTK setup is required that is not documented/exposed clearly in Tauri.

If this should work differently or belongs in `wry`, I’m happy to move the report, but this is the smallest Tauri-facing reproducer I could produce.

AmeerJ97:
Closing this for now. The main blocker turned out to be our local integration path rather than a confirmed upstream Tauri bug, and we need to validate against the working Handy implementation before reopening or filing a narrower upstream report.

PR #15278, opened 2026-04-21
Allow Linux WebKit user-media permission requests

AmeerJ97:
## Summary
Allow Linux WebKitGTK user-media permission requests in the Tauri runtime by handling WebKit `permission-request` events for `UserMediaPermissionRequest` and calling `request.allow()`.

## Why
This is a candidate fix for:
- https://github.com/tauri-apps/tauri/issues/15277

Minimal reproducer:
- https://github.com/AmeerJ97/tauri-linux-mic-permission-repro

## What changed
In `tauri-runtime-wry` on Linux, after the webview is created:
- connect to WebKitGTK `permission-request`
- if the request is `UserMediaPermissionRequest`, call `allow()` and return `true`
- otherwise return `false`

## Status
- `cargo build -p tauri-runtime-wry` passes locally
- This PR is draft because it still needs runtime verification against the standalone reproducer

### DIFF
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -114,6 +114,13 @@ use wry::{
   target_os = "android"
 )))]
 use wry::{WebViewBuilderExtUnix, WebViewExtUnix};
+#[cfg(not(any(
+  target_os = "windows",
+  target_os = "macos",
+  target_os = "ios",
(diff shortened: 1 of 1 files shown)


AmeerJ97:
Closing this draft PR for now. We need to validate the actual local integration against the existing working Handy implementation before proposing an upstream runtime change.

PR #15288, opened 2026-04-23
fix(tauri-build): set Windows FileVersion/ProductVersion strings from tauri config

pascalandr:
## Summary

This PR sets the Windows `FileVersion` and `ProductVersion` string values from `config.version` in `tauri-build`.

The current `dev` branch sets the numeric `FILEVERSION` / `PRODUCTVERSION` fixed fields from the Tauri config, but does not set the corresponding `FileVersion` / `ProductVersion` string values. In that case those strings fall back to whatever `tauri-winres` inferred from `Cargo.toml`.

That can leave the built `.exe` with mixed version sources:

- `FILEVERSION` / `PRODUCTVERSION` fixed fields come from `config.version`
- `FileVersion` / `ProductVersion` strings come from `Cargo.toml`

This PR restores the missing string setters so that Windows VERSIONINFO uses the version from the Tauri config consistently.

## Why this is safe

- matches the existing `tauri-build` changelog entry that says these fields come from `tauri.conf.json`
- changes only a single existing block in `crates/tauri-build/src/lib.rs`
- adds no new API surface
- changes behavior only for undocumented transitive `tauri-winres` Cargo metadata overrides

## Verification

- `cargo build -p tauri-build`
- manually verified the diff is limited to restoring the two missing string setters plus the `.changes` entry

### DIFF
--- a/.changes/windows-versioninfo-strings.md
+++ b/.changes/windows-versioninfo-strings.md
@@ -1,0 +1,5 @@
+---
+"tauri-build": "patch:bug"
+---
+
+Set the correct Windows `FileVersion` and `ProductVersion` string values using the version from the Tauri config.
--- a/crates/tauri-build/src/lib.rs
+++ b/crates/tauri-build/src/lib.rs
@@ -634,6 +634,8 @@ pub fn try_build(attributes: Attributes) -> Result<()> {
         let version = (v.major << 48) | (v.minor << 32) | (v.patch << 16);
         res.set_version_info(VersionInfo::FILEVERSION, version);
         res.set_version_info(VersionInfo::PRODUCTVERSION, version);
+        res.set("FileVersion", version_str);
+        res.set("ProductVersion", version_str);
       }
     }
 


.changes/windows-versioninfo-strings.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri-build": "patch:bug"
+---
+
+Restore the Windows `FileVersion` and `ProductVersion` string values from `tauri.conf.json` / `config.version`, which were unintentionally dropped in #8168.
```
Legend-Master: We should phrase it a little better, something like setting the correct `FileVersion` and `ProductVersion` string using the version in tauri config first

Legend-Master (dismissed):
Nice catch, thanks!

And just a bit of nitpicks. Also, I don't see anywhere we set these fields before https://github.com/tauri-apps/tauri/pull/8168, could you verify this?

.changes/windows-versioninfo-strings.md
pascalandr: Updated in 655d831, thanks. I rephrased the change entry to focus on setting the correct string values from the Tauri config first.

pascalandr:
Thanks, updated the `.changes` wording in `655d831`.

I also re-verified the history:

- `dc6b0d852` originally had `res.set("FileVersion", version)` and `res.set("ProductVersion", version)`
- `1ca2dd677` kept those setters when `set_version_info(...)` was introduced
- `a54797120` / #8168 only keeps the `set_version_info(...)` calls

So the string setters were present before #8168 and were dropped there during the refactor.

Legend-Master:
Not sure what's wrong with `git blame`, but from some searches, it came from c6c59cf2373258b626b00a26f4de4331765dd487 and probably a result of a merge conflict

Let's update the change file and PR descriptions

Also, you'll need to sign your commits for me to merge this

pascalandr:
Updated:

- PR description corrected 
- `.changes` entry reworded
- branch rewritten as a single signed commit

Legend-Master (dismissed):
Thanks!

pascalandr:
Updated the branch again after the branch-protection experiment note.

The no-op `Cargo.toml` tweak has been reverted, so the PR is back to the intended minimal diff only:

- `crates/tauri-build/src/lib.rs`
- `.changes/windows-versioninfo-strings.md`

The current PR head is a single signed commit on top of the latest `dev`.

Legend-Master:
We could revert, we were just experimenting with the branch protection rules

Legend-Master (approved):

PR #15289, opened 2026-04-23
fix(tauri-build): preserve numeric semver build metadata in Windows FILEVERSION

pascalandr:
## Summary

This PR preserves a numeric semver build identifier in the 4th segment of the Windows `FILEVERSION` fixed field.

Today, `tauri-build` projects `config.version` into the Windows fixed version fields as `major.minor.patch.0`, which means versions such as `1.2.3+42` always become `1.2.3.0` even though the numeric build identifier is representable in the Windows 4th segment.

This PR keeps the current `major.minor.patch` projection and preserves the build metadata only when it is:

- numeric
- representable as `u16`

Otherwise, behavior stays unchanged and the 4th segment remains `0`.

`PRODUCTVERSION` is unchanged.

## Why this is safe

- single-file change in the existing Windows version block
- no new API surface
- no behavior change when build metadata is absent, non-numeric, composite, or out of range
- parsing now matches the stricter numeric handling already used in the Windows NSIS code path
- includes unit tests for numeric, composite, non-numeric, and out-of-range build metadata

## Verification

- `cargo test -p tauri-build`

### DIFF
--- a/crates/tauri-build/src/lib.rs
+++ b/crates/tauri-build/src/lib.rs
@@ -464,7 +464,7 @@ pub fn build() {
   }
 }
 
-fn version_u64(v: &semver::Version) -> u64 {
+fn product_version_u64(v: &semver::Version) -> u64 {
   (v.major << 48) | (v.minor << 32) | (v.patch << 16)
 }
 
@@ -475,7 +475,7 @@ fn file_version_u64(v: &semver::Version) -> u64 {
     .map(u64::from)
     .unwrap_or(0);
 
-  version_u64(v) | build
+  product_version_u64(v) | build
 }
 
 /// Same as [`build()`], but takes an extra configuration argument, and does not panic.
@@ -645,9 +645,9 @@ pub fn try_build(attributes: Attributes) -> Result<()> {
 
     if let Some(version_str) = &config.version {
       if let Ok(v) = Version::parse(version_str) {
-        let version = version_u64(&v);
+        let product_version = product_version_u64(&v);
         res.set_version_info(VersionInfo::FILEVERSION, file_version_u64(&v));
-        res.set_version_info(VersionInfo::PRODUCTVERSION, version);
+        res.set_version_info(VersionInfo::PRODUCTVERSION, product_version);
         res.set("FileVersion", version_str);
         res.set("ProductVersion", version_str);
       }
@@ -754,7 +754,7 @@ mod tests {
 
     assert_eq!(
       crate::file_version_u64(&version),
-      crate::version_u64(&version)
+      crate::product_version_u64(&version)
     );
   }
 
@@ -764,7 +764,7 @@ mod tests {
 
     assert_eq!(
       crate::file_version_u64(&version),
-      crate::version_u64(&version)
+      crate::product_version_u64(&version)
     );
   }
 
@@ -774,7 +774,7 @@ mod tests {
 
     assert_eq!(
       crate::file_version_u64(&version),
-      crate::version_u64(&version)
+      crate::product_version_u64(&version)
     );
   }
 }
--- a/crates/tauri-build/src/lib.rs
+++ b/crates/tauri-build/src/lib.rs
@@ -464,18 +464,14 @@ pub fn build() {
   }
 }
 
-fn product_version_u64(v: &semver::Version) -> u64 {
-  (v.major << 48) | (v.minor << 32) | (v.patch << 16)
-}
-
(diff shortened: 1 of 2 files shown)


Legend-Master:
The semver metadata doesn't directly translate to a revision number in `FILEVERSION`, I don't think we should do this to be honest

pascalandr:
I agree it's not a strict semantic mapping.
My point is just that `FILEVERSION` and `PRODUCTVERSION` currently both collapse to `major.minor.patch.0`, so `1.2.3+41` and `1.2.3+42` become indistinguishable in fixed Windows version info.
This PR only preserves an already-numeric build identifier in `FILEVERSION`; otherwise behavior stays unchanged.

Legend-Master:
I feel like this behavior could be quite unexpected for many people, although it won't really do any harm. Especially since we don't handle other types of metadata/pre-release data

@FabianLars any opinions on this?

FabianLars:
hmm i think that's fine? we do a similar conversion in the .msi already

crates/tauri-build/src/lib.rs
```diff
@@ -464,6 +464,23 @@ pub fn build() {
   }
 }
 
+fn version_u64(v: &semver::Version) -> u64 {
+  (v.major << 48) | (v.minor << 32) | (v.patch << 16)
+}
+
+fn file_version_u64(v: &semver::Version) -> u64 {
```
Legend-Master: We should use the parsed `semver::Version`, see

https://github.com/tauri-apps/tauri/blob/a219ede0003bb8073d8002be42bcf343538c42f8/crates/tauri-bundler/src/bundle/windows/nsis/mod.rs#L150-L172

Legend-Master (changes requested):
Ok, let's do this then

pascalandr:
Addressed in `a1d29fd`.

The build metadata handling now matches the stricter NSIS approach: it must parse as a whole `u16`, and composite metadata like `+42.sha` now falls back to `0` instead of using the first segment.

crates/tauri-build/src/lib.rs
```diff
@@ -631,8 +645,8 @@ pub fn try_build(attributes: Attributes) -> Result<()> {
 
     if let Some(version_str) = &config.version {
       if let Ok(v) = Version::parse(version_str) {
-        let version = (v.major << 48) | (v.minor << 32) | (v.patch << 16);
-        res.set_version_info(VersionInfo::FILEVERSION, version);
+        let version = version_u64(&v);
+        res.set_version_info(VersionInfo::FILEVERSION, file_version_u64(&v));
         res.set_version_info(VersionInfo::PRODUCTVERSION, version);
```
Legend-Master: Any reasons for 2 different versions here?

pascalandr: Yes — the only intentional difference is that `FILEVERSION` may preserve a numeric build identifier, while `PRODUCTVERSION` keeps the existing `major.minor.patch.0` projection.

I renamed the base helper to `product_version_u64` in `9bb1e11` to make that distinction explicit in the code.

Legend-Master: Sorry I meant why we want them to be different

pascalandr: You are right, from the Windows executables I inspected, the product version shown in Explorer comes from the string value, while the file version is driven by the fixed numeric version info !
There is no point to distinguish FILEVERSION and PRODUCTVERSION then ...

I stripped this down so both fixed fields now use the same 4-part numeric value.

crates/tauri-build/src/lib.rs
```diff
@@ -464,6 +464,16 @@ pub fn build() {
   }
 }
 
+fn version_u64(v: &semver::Version) -> u64 {
```
Legend-Master: Maybe something like

```suggestion
fn to_winres_version(v: &semver::Version) -> u64 {
```

pascalandr: Renamed in `3c46a62`.

crates/tauri-build/src/lib.rs
```diff
@@ -464,6 +464,16 @@ pub fn build() {
   }
 }
 
+fn to_winres_version(v: &semver::Version) -> u64 {
```
Legend-Master: Let's move this down to just before the tests

Legend-Master (changes requested):
You'll also need to run `cargo fmt`

Legend-Master (approved):
Thanks!

PR #15290, opened 2026-04-24
emit resumed on resumed

seanaye:
Currently there is not an event emitted out of the tauri wry runtime for app resumed events on ios. This PR forwards the resumed event outwards which allows ios apps to listen to the resume event

### DIFF
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -4166,6 +4166,10 @@ fn handle_event_loop<T: UserEvent>(
       callback(RunEvent::Resumed);
     }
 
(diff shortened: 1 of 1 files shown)


Legend-Master (commented):
This was mapped to `WindowEvent::*` in #15199, @lucasfernog I think it should not map to the `RunEvent::Resumed` now?

FabianLars:
I don't think we should expose the _current_ Resume mobile events on RunEvent as at least on Android they are activity-bound meaning you potentially get multiple RunEvent::Resumed events without the App ever being Suspended.

I'd look into a per-app Resume event on the tao side for mobile first before going this here.

PR #15294, opened 2026-04-25
fix(bundler): add missing MULTIUSER LangString translations for Korean NSIS

Debaq:
## Summary

Adds the 5 missing `MULTIUSER_*` LangString translations to `Korean.nsh` so Korean installers built with \`installMode: \"both\"\` no longer fall back to English for the install-mode dialog.

Currently, building a Tauri app with NSIS bundle that lists \`Korean\` in \`bundle.windows.nsis.languages\` and uses \`installMode: \"both\"\` produces these warnings:

\`\`\`
warning: !warning: LangString \"MULTIUSER_TEXT_INSTALLMODE_TITLE\" for language Korean is missing, using fallback from \"...\\English.nsh\"
warning: !warning: LangString \"MULTIUSER_TEXT_INSTALLMODE_SUBTITLE\" for language Korean is missing, ...
warning: !warning: LangString \"MULTIUSER_INNERTEXT_INSTALLMODE_TOP\" for language Korean is missing, ...
warning: !warning: LangString \"MULTIUSER_INNERTEXT_INSTALLMODE_ALLUSERS\" for language Korean is missing, ...
warning: !warning: LangString \"MULTIUSER_INNERTEXT_INSTALLMODE_CURRENTUSER\" for language Korean is missing, ...
\`\`\`

This PR adds Korean translations for those 5 strings.

## Notes for reviewers

The translations were derived from the English equivalents. Native Korean speaker review would be appreciated to confirm naturalness — happy to amend.

## Test plan

- [ ] Build a sample Tauri app with NSIS bundle, \`installMode: \"both\"\`, and \`Korean\` in \`languages\` — verify no \`MULTIUSER_*\` LangString warnings.
- [ ] Run installer with Korean selected — verify the install-mode picker dialog shows Korean text.

### DIFF
--- a/.changes/nsis-korean-multiuser-strings.md
+++ b/.changes/nsis-korean-multiuser-strings.md
@@ -1,0 +1,5 @@
+---
+"tauri-bundler": patch
+---
+
+Add missing `MULTIUSER_*` LangString translations to Korean NSIS language file (`MULTIUSER_TEXT_INSTALLMODE_TITLE`, `MULTIUSER_TEXT_INSTALLMODE_SUBTITLE`, `MULTIUSER_INNERTEXT_INSTALLMODE_TOP`, `MULTIUSER_INNERTEXT_INSTALLMODE_ALLUSERS`, `MULTIUSER_INNERTEXT_INSTALLMODE_CURRENTUSER`). Previously these fell back to English when a Korean installer used `installMode: both`.
--- a/.changes/nsis-korean-multiuser-strings.md
+++ b/.changes/nsis-korean-multiuser-strings.md
@@ -1,5 +1,5 @@
 ---
-"tauri-bundler": patch
+"tauri-bundler": patch:bug
 ---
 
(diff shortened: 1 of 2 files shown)


Legend-Master:
Hi, thanks for contributing

I don't mind adding this just to tauri, but does it make sense to upstream to https://github.com/NSIS-Dev/nsis/blob/affd49994d092bc0a7ba505d87b41de7aecfb738/Contrib/Language%20files/Korean.nsh instead so this benefits everyone using NSIS?

Debaq:
Good point, hadn't thought of that — you're right that upstreaming benefits the whole NSIS ecosystem.

Happy to open a PR against NSIS-Dev/nsis with the same 5 strings. My only concern is that it could take a while to land and ship in a tagged NSIS release, plus Tauri would then need to bump its embedded NSIS version. So I'd suggest keeping this PR open as a short-term fix in the meantime, and once the upstream change is released and pulled in here, the override can just be removed.

Let me know if that works for you and I'll open the upstream PR today.

Nico.-

Debaq:
Quick update: opened the upstream PR — NSIS-Dev/nsis#32.

Same five `MULTIUSER_INSTALLMODEPAGE` strings, wrapped in the standard `!ifdef MULTIUSER_INSTALLMODEPAGE` block to match how `TradChinese.nsh`, `Vietnamese.nsh`, `Ukrainian.nsh` etc. structure it upstream.

Once that lands and gets pulled into a tauri-bundler NSIS sync, the override here can simply be deleted.

Off-topic, but: thank you to the whole Tauri team for the project. Loving working with it.

Nico!

sredna:
You could perhaps put these new strings inside `!if "${NSIS_PACKEDVERSION}" <= 0x03012000` ... `!endif` to avoid problems when the next NSIS version is released.

PR #15296, opened 2026-04-26
refactor: use more early returns and cleanups

Legend-Master:
(use vscode or something other than github to see the diffs better)


sftse:
What sometimes helps with these ugly diffs is to split it into two commits: One changes just `if let Some(...) = ` into `let .. else { ..early return };` then do the indentation of the rest of the lines. Not a guarantee, but sometimes much easier to read if changing indentation does not also change number of lines.

`rustfmt` does reflow lines when width changes, so foils the plan on occasion.

crates/tauri-runtime-wry/src/lib.rs
sftse, line 524: For consistency, this is the last branch where there is a mix of early-return and fallthrough.

Could hoist it into an own match arm
```suggestion
#[cfg(not(windows))]
TaoWindowEvent::Focused(focused) => WindowEvent::Focused(*focused),
#[cfg(windows)]
TaoWindowEvent::Focused(focused) => {
```

sftse (approved):

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -541,9 +542,13 @@ impl WindowEventWrapper {
...
           // (to prevent double focus event when we click a webview after focusing a window)
-          *focused_webview = (*focused).then(|| FOCUSED_WEBVIEW_MARKER.to_string());
+          *focused_webview = if *focused {
+            Some(FOCUSED_WEBVIEW_MARKER.to_owned())
+          } else {
+            None
+          };
 
-          return Self(Some(WindowEvent::Focused(*focused)));
+          WindowEvent::Focused(*focused)
```
sftse, line 551: This change adds more consistency, since the ones that early-return are all of the form `return Self(None)`

(note to self)

crates/tauri/src/manager/webview.rs
```diff
@@ -268,7 +268,7 @@ impl<R: Runtime> WebviewManager<R> {
       let web_resource_request_handler = pending.web_resource_request_handler.take();
       let protocol = crate::protocol::tauri::get(
         manager.manager_owned(),
-        &window_origin,
+        window_origin.clone(),
```
sftse, line 271: This lifts the clone up the callchain, always cloned internally.

crates/tauri/src/protocol/tauri.rs
```diff
@@ -41,8 +41,6 @@ pub fn get<R: Runtime>(
     url
   };
 
-  let window_origin = window_origin.to_string();
-
```
sftse, line 49: Internal clone occurred here.

sftse (dismissed):
Audited the ugly diff with difftastic, LGTM.

sftse (approved):

PR #15299, opened 2026-04-29
enhance(bundler): some error messages

Legend-Master:


crates/tauri-cli/src/error.rs
```diff
@@ -10,8 +10,8 @@ pub enum Error {
   Context(String, Box<dyn std::error::Error + Send + Sync + 'static>),
   #[error("{0}")]
   GenericError(String),
-  #[error("failed to bundle project {0}")]
-  Bundler(#[from] Box<tauri_bundler::Error>),
+  #[error("failed to bundle project: {0}")]
+  Bundler(#[from] tauri_bundler::Error),
```
sftse, line 14: This is a (minor) breaking change, is this intended?

FabianLars: so far we didn't really care about that in the cli crate. though technically it _is_ a library even if that's just for internal napi-rs consumption.

Legend-Master: Yeah, wouldn't consider it breaking or we basically won't be able to change anything

sftse: Don't mind me then :+1:

Qodo-Free-For-OSS:
Hi, `crates/tauri-bundler/src/bundle.rs` uses `anyhow::Context` for an I/O error, which converts the error into `tauri_bundler::Error::BundlerError(anyhow::Error)` instead of the crate’s structured error variants. This contradicts the bundler’s own comment that the anyhow-based variant is “no longer used”, and leads to inconsistent error formatting and matching.

**Severity:** remediation recommended | **Category:** maintainability

**How to fix:** Use bundler Context/ErrorExt

> ### Issue description
> `crates/tauri-bundler/src/bundle.rs` currently uses `anyhow::Context` to add context to an `std::io::Result` (`tempfile::tempfile()`), which pushes failures into `tauri_bundler::Error::BundlerError(anyhow::Error)`.
> 
> This is inconsistent with the bundler’s structured error types (`Error::Fs`, `Error::IoError`, `Error::Context`) and the existing `ErrorExt::fs_context` helper.
> 
> ### Issue Context
> The bundler error enum explicitly notes the `BundlerError(anyhow::Error)` variant is “no longer used”, so this PR unintentionally reintroduces it on a hot path.
> 
> ### Fix Focus Areas
> - crates/tauri-bundler/src/bundle.rs[17-23]
> - crates/tauri-bundler/src/bundle.rs[135-139]
> 
> ### Suggested change
> Remove `use anyhow::Context;` and avoid `anyhow::Context` here.
> 
> One option:
> - Import the bundler’s `Context` trait (from `crate::error`) and convert the `std::io::Error` first:
>   - `tempfile::tempfile().map_err(crate::Error::IoError).context("failed to create temp file for main binary copy")?;`
> 
> This keeps the error inside `tauri_bundler::Error` variants and still adds human-friendly context.

---
*[Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review - free for open-source.*

Legend-Master:
@lucasfernog just realized that the bundler `Error::Context` variant in `tauri-bundler` wraps it self which doesn't seem right, should we change that or should we make a new variant to prevent breaking here? (the one in the `tauri-cli` should be the correct one)

This bot is surprisingly right twice now by the way, pretty cool

lucasfernog:
> @lucasfernog just realized that the bundler `Error::Context` variant in `tauri-bundler` wraps it self which doesn't seem right, should we change that or should we make a new variant to prevent breaking here? (the one in the `tauri-cli` should be the correct one)
> 
> This bot is surprisingly right twice now by the way, pretty cool

use IoError for now and leave it for v3? omg i hate this, i wish bundler wasn't a lib :joy:

Legend-Master:
> use IoError for now and leave it for v3?

I guess if we don't change it, anyhow should work the same for now (the explicit error types were just added to force us to write contexts)

FabianLars (approved):

PR #15303, opened 2026-04-30
fix(deps): update specta in lockfile

Legend-Master:
cherry-pick of b8bebc8, reference: https://github.com/tauri-apps/tauri/pull/15177#issuecomment-4344816208

PR #15305, opened 2026-04-30
Revert "fix(deps): update specta in lockfile"

Legend-Master:
Reverts tauri-apps/tauri#15303

Looking at https://docs.rs/crate/tauri/latest/builds/2963776, it seems to be pulling in `specta-2.0.0-rc.23` which is not in the lockfile 🤷‍♂️

### DIFF
--- a/.github/workflows/lint-rust.yml
+++ b/.github/workflows/lint-rust.yml
@@ -12,6 +12,8 @@ on:
     paths:
       - '.github/workflows/lint-rust.yml'
       - 'crates/**'
+      - 'Cargo.toml'
+      - 'Cargo.lock'
 
 env:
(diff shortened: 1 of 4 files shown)


FabianLars (approved):
i don't really get why we're reverting, so we can test the minimum version?

Legend-Master:
Basically just for the tests to run, the new specta requires some updates on the tests for clippy and requires edition 2024 which we can't use because of MSRV. I don't exactly know why #15177 tried to bumped it at the first place if the docs.rs build failure was not caused by the locked version, I just assumed it was a mistake the change was reverted in be7aabd. (also didn't notice the CI didn't run)

PR #15306, opened 2026-04-30
fix(tauri-build): add `Attributes::config_path`

### LINKED ISSUE #14229: [feat] specify custom `tauri.conf.json` path when using `tauri_build::try_build`
### Describe the problem

I want to be able to specify `tauri.conf.json` in the root of my workspace. Despite being able to specify `config_path` in `CodegenContext`, the following seems to always read the config from current directory.

https://github.com/tauri-apps/tauri/blob/4188ffdafcaa7c3c8b000973146c9384a873e741/crates/tauri-build/src/lib.rs#L473-L474

`config_path` should be respected

### Alternatives considered

_No response_

### Additional context

Note that using `generate_context!` without the manual `build.rs` does not have the problem. However, `generate_context!` only takes plain literals during macro expansion which is inflexible, and I have other things to configure during the build process.

Kokoro2336:
Fixes #14229 

Fixing config path customization in `try_build()`


Kokoro2336:
This PR introduces a breaking change in `Attributes` and `CodegenContext`, which adds extra field `config_path` in `Attributes` for the convenience of customizing `config_path` under non-`codegen`-feature mode. To protect a single source of truth, I suggest ***deprecating*** `CodegenContext::config_path()` and hoist the API to `Attributes`.

Legend-Master (approved):
Thanks

PR #15310, opened 2026-04-30
fix: pin napi for msrv and Node.js on CI

lucasfernog:
regression from #15183


lucasfernog:
CI doesn't work without this one, docker image runs Rust 1.82

ref https://github.com/napi-rs/napi-rs/issues/2491

PR #14984, opened 2026-02-22
feat: bump Android Gradle Plugin to 9.0.1 and dependencies 

kmiit:
This PR updates the Android build environment and dependencies to leverage latest tooling and frameworks for improved stability and compatibility.

### Dependencies
- **Gradle**: 8.14.3 → 9.3.1
- **Android Gradle Plugin**: 8.11.0 → 9.0.1  
- **Kotlin**: 1.9.25 → 2.3.10
- **Kotlin Gradle Plugin**: 1.8.20 → 9.0.1
- `androidx.core:core-ktx`: 1.7.0 → 1.17.0
- `androidx.appcompat:appcompat`: 1.6.0 → 1.7.1
- `androidx.activity:activity-ktx`: 1.10.1 → 1.12.4
- `androidx.webkit:webkit`: 1.14.0 → 1.15.0
- `com.google.android.material:material`: 1.12.0 → 1.13.0
- `androidx.test.ext:junit`: 1.1.4 → 1.3.0
- `androidx.test.espresso:espresso-core`: 3.5.0 → 3.7.0

### Kotlin
- Update JVM target to Java 24

### Gradle
- Add `-Dlint.use.fir.uast=false` flag for FIR compatibility

PR #15335, opened 2026-05-04
fix(android): Gradle upgrades

meghprkh:
- Android Studio kept giving me these suggestions in the notifications
  - Upgrades Gradle, now no need for kotlinOptions, the global JVM version is used automatically
  - Makes the BuildTask not depend on project so configuration cache can be enabled
  - Uses the gradle daemon and java resolution convention

I only half know what I am doing, please review in depth. I will add some github comments on what was suggested by Android Studio


crates/tauri-cli/templates/mobile/android/app/build.gradle.kts
```diff
@@ -2,7 +2,6 @@ import java.util.Properties
 
 plugins {
     id("com.android.application")
-    id("org.jetbrains.kotlin.android")
```
meghprkh, line 6: I assume this is inbuilt now and had to be removed in newer gradle as it kept complaining about two kotlin plugins

crates/tauri-cli/templates/mobile/android/app/build.gradle.kts
```diff
@@ -50,9 +49,6 @@ android {
             )
         }
     }
-    kotlinOptions {
```
meghprkh: With the gradle upgrade / the new kotlin plugin this option is moved to a kotlin section. But the JVM version should work acc https://issuetracker.google.com/issues/408242956

crates/tauri-cli/templates/mobile/android/buildSrc/src/main/kotlin/BuildTask.kt
```diff
@@ -45,18 +54,18 @@ open class BuildTask : DefaultTask() {
...
-        val target = target ?: throw GradleException("target cannot be null")
-        val release = release ?: throw GradleException("release cannot be null")
+        val rootDirRel = rootDirRel ?: throw GradleException("rootDirRel cannot be null") as Throwable
+        val target = target ?: throw GradleException("target cannot be null") as Throwable
+        val release = release ?: throw GradleException("release cannot be null") as Throwable
         val args = listOf({{quote-and-join tauri-binary-args}});
 
-        project.exec {
-            workingDir(File(project.projectDir, rootDirRel))
+        execOperations.exec {
```
meghprkh, line 60: project.exec no longer works

crates/tauri-cli/templates/mobile/android/buildSrc/src/main/kotlin/BuildTask.kt
```diff
@@ -45,18 +54,18 @@ open class BuildTask : DefaultTask() {
     }
 
     fun runTauriCli(executable: String) {
-        val rootDirRel = rootDirRel ?: throw GradleException("rootDirRel cannot be null")
-        val target = target ?: throw GradleException("target cannot be null")
-        val release = release ?: throw GradleException("release cannot be null")
+        val rootDirRel = rootDirRel ?: throw GradleException("rootDirRel cannot be null") as Throwable
```
meghprkh: This is likely unnecessary but was suggested by the LSP

crates/tauri-cli/templates/mobile/android/buildSrc/src/main/kotlin/BuildTask.kt
```diff
@@ -5,11 +5,20 @@ import org.gradle.api.GradleException
...
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.TaskAction
+import javax.inject.Inject
+import org.gradle.process.ExecOperations
+import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.tasks.Internal
+
+abstract class BuildTask : DefaultTask() {
+    @get:Inject
+    abstract val execOperations: ExecOperations
```
meghprkh, line 13: project.exec replacement. class becomes abstract https://stackoverflow.com/questions/79640295/how-to-replace-project-exec

crates/tauri-cli/templates/mobile/android/buildSrc/src/main/kotlin/BuildTask.kt
```diff
@@ -45,18 +54,18 @@ open class BuildTask : DefaultTask() {
...
         val args = listOf({{quote-and-join tauri-binary-args}});
 
-        project.exec {
-            workingDir(File(project.projectDir, rootDirRel))
+        execOperations.exec {
+            workingDir(File(projectDir, rootDirRel))
             executable(executable)
             args(args)
-            if (project.logger.isEnabled(LogLevel.DEBUG)) {
+            if (logger.isEnabled(LogLevel.DEBUG)) {
```
meghprkh, line 64: If we use `Task.project` within task, build can not be confguration-cached

crates/tauri-cli/templates/mobile/android/gradle.properties
```diff
@@ -7,6 +7,7 @@
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
 org.gradle.jvmargs=-Xmx2048m -Dfile.encoding=UTF-8
+org.gradle.configuration-cache=true
```
meghprkh, line 10: enable configuration cache https://docs.gradle.org/9.4.1/userguide/configuration_cache_enabling.html

crates/tauri-cli/templates/mobile/android/gradle.properties
```diff
@@ -21,4 +22,3 @@ kotlin.code.style=official
 # resources declared in the library itself and none from the library's dependencies,
 # thereby reducing the size of the R class for that library
 android.nonTransitiveRClass=true
-android.nonFinalResIds=false
```
meghprkh: deprecation warning on this

crates/tauri-cli/templates/mobile/android/settings.gradle
```diff
@@ -1,3 +1,7 @@
+plugins {
```
meghprkh, line 1: IDE suggested for automatic resolution of JVM

crates/tauri-cli/templates/plugin/android/settings.gradle
```diff
@@ -10,9 +10,6 @@ pluginManagement {
                 case "com.android.library":
                     useVersion("8.0.2")
                     break
-                case "org.jetbrains.kotlin.android":
```
meghprkh, line 13: I assume this is no longer needed

meghprkh:
Ah I also see https://github.com/tauri-apps/tauri/pull/14984/ . Sorry didnt check draft PRs before making this one. Please close if unneeded

crates/tauri-cli/templates/mobile/android/buildSrc/src/main/kotlin/BuildTask.kt
meghprkh: * https://docs.gradle.org/current/userguide/service_injection.html#execoperations

Legend-Master:
@meghprkh thanks for comments, awesome work!

Have tried this with the dialog plugin and it complained about the duplicated `id("org.jetbrains.kotlin.android")` and deprecated `kotlinOptions` which means we probably can't do this in v2...

@FabianLars since this is kinda important that our current Gradle/AGP versions don't work with the default Java bundled in Android Studio (JBR 25) anymore, do you think it's acceptable to force everyone to do the migration? 😩 (Maybe not, it breaks too much)

Maybe we can try disable the built-in kotlin for now with `android.builtInKotlin=false`?

FabianLars:
breaking the templates should be totally fine, no? 
how does the compatibility of the `crates/tauri-build/src/mobile.rs` and `crates/tauri/mobile/android/build.gradle.kts` look like? does that break projects until they run `android init` again?

missing the whole picture here rn tbh, in theory 2.12 is already kinda breaking but idk how breaking this change really is and whether plugins need to adapt in a breaking way

---

In theory android studio will prompt to download java 21 when you open it but that won't update the gradle config automatically :/

Legend-Master:
> how does the compatibility of the crates/tauri-build/src/mobile.rs and crates/tauri/mobile/android/build.gradle.kts look like? does that break projects until they run android init again?

Didn't try it yet but probably also problematic

> missing the whole picture here rn tbh, in theory 2.12 is already kinda breaking but idk how breaking this change really is and whether plugins need to adapt in a breaking way

Yeah... That's what I'm talking about, the plugins will all need to be updated to compile if we do this...

> In theory android studio will prompt to download java 21 when you open it but that won't update the gradle config automatically :/

Yeah it did prompt but didn't work at all, not sure the reason though. That's not too reliable at least.

FabianLars:
> Yeah... That's what I'm talking about, the plugins will all need to be updated to compile if we do this...

Okay that sucks then. I mean the plan is still to somewhat rush v3 no? Maybe we can delay this PR a bit then. Can we do the below screenshot with the CLI perhaps?

> Yeah it did prompt but didn't work at all, not sure the reason though. That's not too reliable at least.

it does download it fine at least but then on the gradle error you still have to click on `Change Gradle JDK configuration`
grafik

so yeah, not what i'd expect when i click on a "Use Java 21" button 😂

Legend-Master:
I tried this thing as well and it didn't do anything for me either so...

Legend-Master:
Actually 🤦‍♂️

image

FabianLars:
solid product i'd say 😂

Legend-Master:
@meghprkh we have just merged a non-breaking version of this in #15828 that is based on this PR, we will consider merging this one in v3. Thanks!

meghprkh:
Thanks @Legend-Master , no worries!

PR #15344, opened 2026-05-05
fix(cli): Honor a specified NDK version by setting NDK_HOME, if set (#14860)

### LINKED ISSUE #14860: [feat] Honor a specified NDK version, if set
### Describe the problem

Latest Tauri versions try to build using the latest NDK which is installed in the system - not the specified NDK through the NDK_HOME variable. This means it's impossible to reproduce a build within build systems, and this adds unnecessary friction should developers want their apps to be reproducible by others (such as the guys over at F-Droid / IzzyOnDroid.)

If NDK_HOME is set, Tauri should honor that instead of trying to get the latest NDK from `$ANDROID_HOME/ndk`.

Ref: https://github.com/tauri-apps/tauri/blob/32576120fd4c7cb164a546c813cbb014c0b8da54/crates/tauri-cli/src/mobile/android/mod.rs#L347-L365

### Alternatives considered

_No response_

### Additional context

_No response_

raglady:
Previously, if the NDK_HOME environment variable is set or not, the NDK used by the cli is  the lastest version installed at ANDROID_HOME/ndk directory. And if the cli cannot find any ndk inside the ANDROID_HOME/ndk, it try to install the version 29.0.13846066.
We got always this type of output even we set NDK_HOME or not, where the version is the latest or 29.0.13846066:
```
Info Using installed NDK: ~/Android/Sdk/ndk/29.0.13846066
```

Now, after this update, when we launch for example:
```
$  NDK_HOME=/custom/Android/Sdk/ndk/xx.xxxxxx cargo tauri android dev
```
we get the output:
```
Info Using installed NDK: /custom/Android/Sdk/ndk/xx.xxxxxxx
```

Fixes #14860 
Closes #14398

### DIFF
--- a/.changes/fix-ndk-home-not-honor.md
+++ b/.changes/fix-ndk-home-not-honor.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch:bug
+---
+
+Fix NDK_HOME environment variable not honored when set
--- a/crates/tauri-cli/src/mobile/android/mod.rs
+++ b/crates/tauri-cli/src/mobile/android/mod.rs
@@ -591,14 +591,27 @@ fn ensure_ndk(non_interactive: bool) -> Result<()> {
     .or_else(|| std::env::var_os("ANDROID_SDK_ROOT"))
     .map(PathBuf::from)
     .context("Failed to locate Android SDK")?;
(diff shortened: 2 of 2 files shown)


Legend-Master (approved):
Thanks!

PR #15351, opened 2026-05-07
Respect required-features during src/bin binary discovery

mike-lmctl:
Fixes the src/bin fallback in the Tauri CLI bundler so it does not re-add a
binary that Cargo skipped because its required features were not enabled.

I added a regression test with a feature-gated [[bin]] entry and a matching
src/bin file. The focused tauri-cli test passes locally and was also confirmed
in a clean Docker container.

### DIFF
--- a/crates/tauri-cli/src/interface/rust.rs
+++ b/crates/tauri-cli/src/interface/rust.rs
@@ -740,6 +740,23 @@ impl CargoSettings {
   }
 }
 
+fn manifest_bin_required_features_match(bin: &BinarySettings, options: &Options) -> bool {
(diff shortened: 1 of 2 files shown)


mike-lmctl:
All 15 checks are passing now. When someone from the Tauri CLI side has time, I’d appreciate a review.

FabianLars:
Hey, sorry i missed your PR and just merged another one :/ Thanks for the contribution nonetheless.

For the future, we need a changefile (check .changes dir) and more importantly, we need all commits to be [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) to be able to merge PRs :)

Issue #8248, opened 2023-11-18
[feat] Build for the web


Jamzy01:
### Describe the problem

I know that Yew already exists, but implementing a way to build for the web in tauri would be a really useful feature

### Describe the solution you'd like

I don't need this feature urgently, but it would be nice if it could at least be added to the Tauri roadmap by an admin to be implemented at some point. I just want a simple way to build a single page app (or even multi page at some point). I think it'd be a great idea because then the web app can leverage the power of rust whilst still being easy to write and having nice gui.

### Alternatives considered

Of course, there's yew, but that'd mean you'd have to create your app again from the ground up.

amrbashir:
I am not sure what you mean tbh so would need more clarification. Tauri uses web technology to present the UI inside a webview on Desktop, you can still reuse your HTML/CSS/JS project and run it inside the browser (you still need to guard your usage of tauri APIs though so it doesn't error in the browser)

Jamzy01:
> I am not sure what you mean tbh so would need more clarification. Tauri uses web technology to present the UI inside a webview on Desktop, you can still reuse your HTML/CSS/JS project and run it inside the browser (you still need to guard your usage of tauri APIs though so it doesn't error in the browser) 

Yeah I know I wasn't particularly clear, and the reason for that is that I'm not 100% sure of how I'd like it to be implemented. Now yes, I know I can just run it inside a web browser, but as you said, then you can't call tauri commands, and you just overall don't have the rust backend. And I know that for my app, my frontend utilizes the backend for it's speed to perform heavy computations even when native apis aren't even being used (in other words, I'm writing code in rust that could just as well be written in js, but I'm doing it to increase performance). And allowing normal websites to integrate with a rust backend (a backend that I assume would be ran though wasm) would be great because then you can make faster websites still use the same web technologies you're used to (unlike yew where you have to mess around learning their own web framework, which makes it difficult to port from an existing web project).

Also something that would be nice would be some inbuilt apis to rust for the web, like for example still allowing the backend to make http requests, still giving dialog support etc).

And of course I'd want a way to conditionally compile code only for a web build, something like this:

`#[cfg(web)]`

The current app I'm working on should work great as a browser build, because with a few small tweaks it'd work just fine without native apis. And my app would definitely be much faster than just a regular react app levering the power of rust. And I'm also sure that if apps like discord and teams were made using tauri, then this'd make it easy for them to have one codebase for all platforms.

[...]

PR #15354, opened 2026-05-07
Add WebAssembly build support for the JS CLI

birdup000:
# Web Build Support Changes

This note summarizes the `add web build` commit that adds WebAssembly/WASI build
support for the `@tauri-apps/cli` package.

## Summary

The change adds a web build path for the JavaScript CLI package so it can produce
NAPI-RS WebAssembly artifacts in addition to the existing native Node.js
bindings. The new target is `wasm32-wasip1-threads`, and the package exposes a
browser entry that can load the generated WASI binding in web-compatible
environments.

## Main Changes

- Added `wasm32-wasip1-threads` to the CLI package's NAPI targets.
- Added `build:web` and `build:web:debug` scripts for producing the WASI
  artifacts.
- Added a small `build-web.js` helper so the WASI reactor object is linked
  correctly and the generated module does not hang during load.
- Added `collect-artifacts.js` so release artifacts are copied into both the
  root CLI package and the `@tauri-apps/cli-wasm32-wasi` optional package.
- Added `main.browser.js` as the browser-facing wrapper so the public API keeps
  the same Promise-based shape as `main.js`.
- Added package metadata for `@tauri-apps/cli-wasm32-wasi`.
- Updated the publish workflow to build and upload the web artifacts before the
  publish job runs.
- Kept native CLI behavior unchanged by gating the native `tauri-cli`
  dependency and native `run` implementation away from wasm targets.

## Runtime Behavior

The WebAssembly binding can be imported and loaded in web/WASI environments. The
native Tauri CLI commands still require native toolchains and process spawning,
so the wasm `run` export returns a clear error instead of trying to execute
native-only behavior:

> The Tauri CLI WebAssembly binding can be loaded in web/WASI environments, but
> running Tauri commands requires native toolchains and process spawning.

## Validation Performed

- Ran the new web build path with `pnpm --dir packages/cli build:web`.
- Confirmed the generated WASI binding can be loaded from Node.js without
  hanging.
- Confirmed the wasm `run` export returns the expected native-toolchain guard
  error.
- Ran the native CLI package check with `cargo check -p tauri-cli-node`.
- Ran `cargo fmt -p tauri-cli-node --check`.
- Temporarily created and ran a small Tauri/Vite showcase app to verify the
  browser entry loads the WASI binding with the required cross-origin isolation
  headers.

## Review Notes

The new JavaScript and Rust source files include the standard Tauri header:

```text
Copyright 2019-2024 Tauri Programme within The Commons Conservancy
SPDX-License-Identifier: Apache-2.0
SPDX-License-Identifier: MIT
```

I found that the repository has a `check license headers` workflow that requires
this header on new `.rs`, `.js`, `.ts`, `.yml`, `.swift`, and `.kt` files. If
the wording, year range, or placement needs to be changed for this contribution,
I can update the new files to match the maintainers' preferred licensing policy.

Please also let me know if any of the code needs to be adjusted to better match
project conventions, naming, review expectations, or contributor conduct
guidelines. I am able to change the helper script names, package metadata,
workflow shape, feature naming, error message wording, or browser wrapper
structure if a different style would fit the project better.

### DIFF
--- a/.github/workflows/publish-cli-js.yml
+++ b/.github/workflows/publish-cli-js.yml
@@ -150,6 +150,38 @@ jobs:
           name: bindings-${{ matrix.settings.target }}
           path: packages/cli/${{ env.APP_NAME }}.*.node
           if-no-files-found: error
+  build-web:
+    name: stable - wasm32-wasip1-threads - web
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - run: npm i -g --force corepack
+      - name: Setup node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: 'pnpm'
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: wasm32-wasip1-threads
+      - name: Install dependencies
+        run: pnpm i --frozen-lockfile --ignore-scripts
+      - name: Build WebAssembly binding
+        run: pnpm build:web
+      - name: Upload artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: bindings-wasm32-wasip1-threads
+          path: |
+            packages/cli/cli.wasi-browser.js
+            packages/cli/cli.wasi.cjs
+            packages/cli/cli.wasi.d.ts
+            packages/cli/cli.wasm32-wasi.debug.wasm
+            packages/cli/cli.wasm32-wasi.wasm
+            packages/cli/wasi-worker-browser.mjs
+            packages/cli/wasi-worker.mjs
+          if-no-files-found: error
   #  build-freebsd:
   #    runs-on: macos-10.15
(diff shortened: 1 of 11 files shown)


birdup000:
This PR should close this issue #8248

FabianLars:
Thanks for the PR but i honestly don't understand what this would be good for 😅 I can't think of a single reason to use the cli in a wasm context. It's also not related to #8248 - that wants the tauri _app_ in web, not the cli.

Will close this for now but can re-open if there's a good reason for a wasm cli :)

birdup000:
> Thanks for the PR but i honestly don't understand what this would be good for 😅 I can't think of a single reason to use the cli in a wasm context. It's also not related to #8248 - that wants the tauri _app_ in web, not the cli.
> 
> Will close this for now but can re-open if there's a good reason for a wasm cli :)

Thank you for the feedback, will continue working on the web support.

PR #15372, opened 2026-05-14
feat: add Windows VC runtime linking and bundling options

amrbashir:
## Summary

Adds explicit configuration for controlling the Windows Visual C++ runtime behavior.

- Adds `build.windows.staticVCRuntime` to control MSVC static runtime linking.
- Adds `tauri_build::WindowsAttributes::static_vc_runtime` as a build-script override.
- Deprecates the `STATIC_VCRUNTIME` env var and emits a warning when it is used.
- Adds `bundle.windows.bundleVCRuntime` to copy Visual C++ runtime DLLs into Windows MSI/NSIS bundles.
- Moves the embedded `vswhere.exe` into `tauri-bundler` so both the bundler and CLI can reuse it.
- Uses bundled `vswhere.exe` to locate Visual Studio redist files, with `VCTOOLS_REDIST_DIR` available as an override.
- Adds a changeset for the affected crates.

## Details

When `bundle.windows.bundleVCRuntime` is enabled, the Windows bundlers locate the target architecture runtime directory from:

```text
<VCTOOLS_REDIST_DIR>/<latest MSVC version>/<arch>/Microsoft.VC*.CRT
```

or, if `VCTOOLS_REDIST_DIR` is unset:

```text
<Visual Studio install>/VC/Redist/MSVC/<latest MSVC version>/<arch>/Microsoft.VC*.CRT
```

The Visual Studio install is discovered using the embedded `vswhere.exe` and the `Microsoft.VisualStudio.Component.VC.Redist.14.Latest` component. The latest MSVC redist version is selected by parsing version directory names with `semver`.

## Validation

- `cargo fmt --all --check`
- `cargo check -p tauri-utils -p tauri-build -p tauri-bundler -p tauri-cli`
- `cargo run -p tauri-schema-generator`


crates/tauri-bundler/src/bundle/windows/nsis/mod.rs
```diff
@@ -774,6 +774,25 @@ fn generate_resource_data(settings: &Settings) -> crate::Result<ResourcesMap> {
...
 
+  if settings.windows().bundle_vc_runtime {
+    for dll in vc_runtime_dlls(settings.binary_arch())? {
+      let dll = dunce::simplified(&dll).to_path_buf();
+      if added_resources.contains(&dll) {
+        continue;
+      }
+      if settings.windows().can_sign() {
+        try_sign(&dll, settings)?;
+      }
```
Legend-Master: The VC runtime DLLs should be signed by Microsoft already?

amrbashir: correct, removed that bit

crates/tauri-cli/config.schema.json
```diff
@@ -2749,6 +2777,11 @@
               "type": "null"
             }
           ]
+        },
+        "bundleVCRuntime": {
+          "description": "Whether to bundle the Visual C++ runtime DLLs alongside the application.\n\n This is only useful when `build > windows > staticVCRuntime` is set to `false`.",
```
FabianLars: How about a note like "or you're bundling exes/dlls that require it" - you know, like the reason you're opening this for :D

crates/tauri-cli/src/interface/rust/desktop.rs
```diff
@@ -159,10 +159,6 @@ pub fn build(
   let out_dir = app_settings.out_dir(&options, tauri_dir)?;
   let bin_path = app_settings.app_binary_path(&options, tauri_dir)?;
 
-  if !std::env::var_os("STATIC_VCRUNTIME").is_some_and(|v| v == "false") {
```
lucasfernog, line 162: isn't this a breaking change?

amrbashir: not really just moved it to tauri-build

crates/tauri-cli/config.schema.json
amrbashir: added

crates/tauri-cli/src/interface/rust/desktop.rs
lucasfernog: previously even when STATIC_VCRUNTIME was unset this would return true and force it to true

so we need to preserve this - maybe check first the config, then default to reading the env (and using true if None)?

amrbashir: yes that is preserved, the lookup chain is:
STATIC_VCRUNTIME != false,  -> statick linking
STATIC_VCRUNTIME not set -> check tauri-build API and use that
STATIC_VCRUNTIME not set and tauri-build API not used -> check staticVCRuntime config option (and if not set, it defaults to true)

lucasfernog (dismissed):

lucasfernog (approved):

PR #15373, opened 2026-05-14
enhance: include permission identifier in ACL error messages

### LINKED ISSUE #11262: [feat] Plugin names should be able to include underscore '_'
### Describe the problem

When creating my first (internal) plugin, I called it `sqlite_proxy`. Then I spent literally hours chasing down the build issue because the message is not very clear:
```
  thread 'main' panicked at build.rs:11:6:
  failed to run tauri-build: failed to parse JSON: identifiers can only include lowercase ASCII, hyphens which are not leading or trailing, and a single colon if using a prefix at line 16 column 23
```

It does not identify the `.json` file that needs fixing; nor specify that `identifiers` here refers to plugin names. Being new to the plugin ecosystem, I assumed that I simply failed to create some sort of capability file / permission setup. This was not the case. The above error message is a result of https://github.com/tauri-apps/tauri/pull/9952, which improved the state of the world, but punted on documentation changes and the underlying issue.

I believe there are 3 different levels of fixing this, listed in order of difficulty:
- [ ] Update the documentation around plugins to be very explicit about allowed names.
- [ ] Update the error message to:
  1. mention the file it's failing on
  2. mention that the problem is the plugin name.
  3. Optionally add validation _elsewhere_ that fails quickly on disallowed plugin names (e.g. plugin init)
- [ ] The ultimate fix would be to simply remove this constraint.
Note: this was already considered in https://github.com/tauri-apps/tauri/issues/9951: `writing our own deserializer without serde_untagged`. I do not know the specific limitations of serde/serde_untagged regarding this. Sharing context on the reasons it's so painful to change might be useful.

### Alternatives considered

_No response_

### Additional context

I personally wouldn't consider the issue fully solved until the restriction is removed; however the interim steps would help mitigate some of the harm.

MavenRain:
#### What kind of change does this PR introduce?

  Enhancement to ACL error diagnostics.

  #### Description

  Closes #11262 (level 2 of the three-tier ask: identify the file and clarify
  which identifier is invalid).

  - Adds three path-attached variants to the ACL `Error` enum (`TomlFile`,
  `JsonFile`, `Json5File`) carrying the parse error and the file path. The
  legacy `Toml`, `Json`, `Json5` variants stay for non-file parse contexts.
  - Updates the parse call sites in `acl/build.rs` (3) and `acl/capability.rs`
  (3) to use the new variants via `map_err`, so the file path is reported in the
   error.
  - Changes `ParseIdentifierError::InvalidFormat` from a unit variant to
  `InvalidFormat(String)` carrying the offending identifier. Updates the message
   to say "invalid plugin or permission identifier '<value>': ..." rather than
  the previous generic "identifiers can only include ...".

  The composite effect on the user-facing error (the exact scenario from
  #11262):

  Before: `failed to parse JSON: identifiers can only include lowercase ASCII,
  hyphens which are not leading or trailing, and a single colon if using a
  prefix at line 16 column 23`

  After: `failed to parse JSON file '/path/to/permissions/default.toml': invalid
   plugin or permission identifier 'sqlite_proxy:allow-foo': identifiers can
  only include lowercase ASCII letters, digits, hyphens (not leading or
  trailing), and a single colon when using a prefix at line 16 column 23`

  #### Notes for review

  - The `InvalidFormat` variant signature change is technically a breaking API
  change. I verified there are zero in-tree pattern matches on
  `ParseIdentifierError` variants (`grep -rn 'ParseIdentifierError::' crates/`),
   so no internal callers break. External pattern matchers (if any) would need
  to update from `InvalidFormat` to `InvalidFormat(_)`. The covector changeset
  is `minor:enhance`; happy to bump to `major:breaking` if you prefer.
  - Level 1 of the issue (docs covering naming rules) was addressed separately
  at tauri-apps/tauri-docs#3826.
  - Level 3 (relaxing the underscore restriction) is a deeper change tied to
  `serde_untagged` (referenced in #9951) and is intentionally out of scope here.

  Build, test, fmt, clippy run locally on `tauri-utils --features
  build,config-json5`; all clean.


crates/tauri-utils/src/acl/identifier.rs
```diff
@@ -127,8 +127,8 @@ pub enum ParseIdentifierError {
   Humongous(usize),
 
   /// Identifier is not in a valid format.
-  #[error("identifiers can only include lowercase ASCII, hyphens which are not leading or trailing, and a single colon if using a prefix")]
-  InvalidFormat,
+  #[error("invalid plugin or permission identifier '{0}': identifiers can only include lowercase ASCII letters, digits, hyphens (not leading or trailing), and a single colon when using a prefix")]
+  InvalidFormat(String),
```
FabianLars: This unfortunately is a breaking change and has to wait for v3

MavenRain:
Thanks @FabianLars, I agreed.  I just pushed a revision that keeps `ParseIdentifierError::InvalidFormat` as a unit variant (no API change) and instead wraps the offending value into the error at the `Identifier::Deserialize` boundary.  The user-visible serde error message is unchanged in shape (invalid plugin or permission identifier '<value>': <reason>), so the diagnostic improvement carries through every real parse path (TOML/JSON/JSON5 capability and permission files) without breaking anyone matching on the enum.  Direct callers of `Identifier::try_from` see the unit-variant `Display` as previously seen.

The three new path-attached parse variants (`Error::TomlFile`, `Error::JsonFile`, and `Error::Json5File`) are additive and remain in this PR; let me know if you want any of those split-out as well.

crates/tauri-utils/src/acl/mod.rs
```diff
@@ -116,15 +116,28 @@ pub enum Error {
   #[error("failed to parse TOML: {0}")]
   Toml(#[from] toml::de::Error),
```
Legend-Master, line 117: Let's deprecate the old ones and add a todo for the removal in v3

crates/tauri-utils/src/acl/mod.rs
```diff
@@ -116,15 +116,28 @@ pub enum Error {
...
 
+  /// Invalid TOML encountered while parsing a known file.
+  #[error("failed to parse TOML file '{}': {}", _1.display(), _0)]
+  TomlFile(toml::de::Error, PathBuf),
```
Legend-Master: Could we align this with 

https://github.com/tauri-apps/tauri/blob/79e7ab8ad2a4dad9384c03bb50e9dc6917089057/crates/tauri-utils/src/config/parse.rs#L84

e.g.

```rust
    /// The path that failed to parse into JSON.
    path: PathBuf,

    /// The parsing [`serde_json::Error`].
    error: serde_json::Error,
```

(p.s. not the name, just tuple to struct)

crates/tauri-utils/src/acl/mod.rs
Legend-Master: Actually.. We didn't mark this error type `#[non_exhaustive]` so adding a new type is also breaking...

Do you mind splitting the `Identifier` deserialization code to a new PR so we can merge it first? The error message change will probably need to wait for v3

Legend-Master (approved):
Thanks, awesome improvement!

PR #15375, opened 2026-05-14
feat(bundler): support Debian pre-release versioning

### LINKED ISSUE #14869: [feat] bundler(deb): Allow debian pre-release versioning
### Describe the problem

I've stumbled upon this small debian "anomaly" where their versioning is not 100% compatible with semver:
debian uses [the tilde character](https://wiki.debian.org/Versioning) for pre-releases and SemVer uses [a colon](https://semver.org/#spec-item-9)

so if we take the versions in the SemVer example: `1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0`
for the same versions, debian ordering would be `1.0.0 < 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1`

Replacing the first colon (since SemVer allows multiple to appear) with a tilde (most likely in the deb bundling process since it doesn't seem to affect other parts).

The result for the above example is: `1.0.0~alpha < 1.0.0~alpha.1 < 1.0.0~alpha.beta < 1.0.0~beta < 1.0.0~beta.2 < 1.0.0~beta.11 < 1.0.0~rc.1 < 1.0.0`, the desired order.

### Alternatives considered

Leave it as is - after all it's a somewhat niche problem to have (usage of pre-versioning may not be as widespread).

### Additional context

_No response_

timkicker:
Converts the `-` SemVer pre-release separator to `~` when writing the deb filename and control file `Version:` field, so e.g. `1.0.0-alpha` becomes `1.0.0~alpha` and sorts before `1.0.0`. Any `-` inside the pre-release is rewritten to `.` so dpkg doesn't read it as the `upstream_version`/`debian_revision` separator (`1.0.0-rc-1` becomes `1.0.0~rc.1`).

Versions that don't parse as SemVer or have no pre-release suffix are passed through unchanged.

Closes #14869.

### DIFF
--- a/crates/tauri-bundler/src/bundle/linux/debian.rs
+++ b/crates/tauri-bundler/src/bundle/linux/debian.rs
@@ -41,6 +41,31 @@ use std::{
   path::{Path, PathBuf},
 };
 
+/// Converts a SemVer version string to the Debian upstream version format.
+///
+/// SemVer separates the pre-release suffix with `-`, while Debian uses `~`
+/// so that `1.0.0~alpha` sorts before `1.0.0`. Any `-` inside the SemVer
+/// pre-release is rewritten to `.` so the whole suffix stays on the upstream
+/// side of the Debian grammar, which uses `-` to separate `upstream_version`
+/// from `debian_revision`. Two pre-releases that differ only by `-` vs `.`
+/// position (e.g. `rc-1` and `rc.1`) therefore produce the same output.
+/// Inputs that don't parse as SemVer, or have no pre-release, are returned
+/// unchanged. See <https://wiki.debian.org/Versioning>.
+fn to_debian_version(version: &str) -> String {
+  match semver::Version::parse(version) {
+    Ok(v) if !v.pre.is_empty() => {
+      let pre = v.pre.as_str().replace('-', ".");
+      let mut deb = format!("{}.{}.{}~{}", v.major, v.minor, v.patch, pre);
+      if !v.build.is_empty() {
+        deb.push('+');
+        deb.push_str(v.build.as_str());
+      }
+      deb
+    }
+    _ => version.to_string(),
+  }
+}
+
 /// Bundles the project.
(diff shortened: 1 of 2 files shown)


crates/tauri-bundler/src/bundle/linux/debian.rs
```diff
@@ -60,7 +85,7 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
   let package_base_name = format!(
     "{}_{}_{}",
     settings.product_name(),
-    settings.version_string(),
+    to_debian_version(settings.version_string()),
```
FabianLars, line 88: not a review yet, more like a reminder for when we got to this pr: so far we handled filename changes as breaking changes. for example it would break tauri-action's asset detection

WolleTD:
I just stumbled upon this restriction when trying to version packages for different debian releases.

Rather than modifying the given version (and thus introducing a breaking change as @FabianLars stated), I think it would be preferrable to add an additional `revision` setting that isn't constrained to semver (I think there is a spec in the Debian Manual somewhere). If provided, it is appended to the version like `{version}-{revision}`. This would allow me to do stuff like `3.0.0-1~deb12` and `3.0.0-1~deb13`.

It could be unset in a minor update and optionally defaulted to `1` in a major update. While not required by the Debian package version spec, almost all debian packages have it and almost all packagers set it to `1` by default.

FabianLars:
we're also trying to get v3 sooner rather than later so we could do it in breaking ways as well. what you described would kinda match what we do in rpm though so that sounds reasonable to me rn.

so, how about we think about what the ideal result would be and then we go from there and see what's possible in tauri v2 or v3?

PR #15383, opened 2026-05-15
fix(tauri-utils): preserve resource source file name when dest is empty

### LINKED ISSUE #15380: [bug] Regression in tauri-utils 2.8.3: can no longer use "." as a destination path for a file resource
### Describe the bug

Before this change: https://github.com/tauri-apps/tauri/pull/14662

This resource config used to work:

```json
{
  "bundle": {
    "resources": {
      "my-file.txt": "."
    }
  }
}
```

But now `tauri_build::build()` fails with `Is a directory (os error 21)`

Observed failing on both mac and windows.

### Reproduction

Minimal reproduction: https://github.com/montyc1999/tauri-repro-resource-dest-dot-path
1. Observe that it builds.
2. In `Cargo.toml`, change: `tauri-utils = { version = "=2.8.2" }`
    to `tauri-utils = { version = "=2.8.3" }`
    And observe that it fails.

### Additional context

We can still map a folder to ".":

```json
{ "bundle": {
    "resources": {
      "my-folder": "."
    }
} }
```

It's just the single file case that no longer works.

Legend-Master:
Fix #15380, regression from #14662

Apparently https://github.com/tauri-apps/tauri/pull/14662#discussion_r2621860377 is possible

This behavior is confusing, but we should still keep it in v2 for the compatibility

### DIFF
--- a/crates/tauri-utils/src/resources.rs
+++ b/crates/tauri-utils/src/resources.rs
@@ -209,7 +209,21 @@ impl ResourcePathsIter<'_> {
             // preserving the file name as it is
             ResourcePathsInnerIter::Glob { .. } => dest.join(path.file_name().unwrap()),
           },
-          None => dest.clone(),
+          None => {
+            if dest.components().count() == 0 {
+              // if current_dest is empty while processing a file pattern
+              // we preserve the file name as it is
+              //
+              // e.g. `{ "README.md": "" }` is `README.md` -> `$RESOURCE/README.md`
+              //
+              // TODO: This behavior is a confusing special case,
+              // remove this in v3 or make other cases like this work
+              // > `{ "README.md": "./folder/" }` is `README.md` -> `$RESOURCE/folder/README.md` (this gives `$RESOURCE/folder` today)
+              PathBuf::from(path.file_name().unwrap())
+            } else {
+              dest.clone()
+            }
+          }
         }
       } else {
         // If [`ResourcePathsIter::pattern_iter`] is a [`PatternIter::Slice`]
--- a/.changes/no-dest-resource-target.md
+++ b/.changes/no-dest-resource-target.md
@@ -1,0 +1,17 @@
+---
+"tauri-utils": "patch:bug"
+---
+
+Fix a regression in tauri-utils 2.8.3 that made empty path an invalid resource target, e.g.
+
+```json
+{
+  "bundle": {
+    "resources": {
+      "README.md": "",
+    }
+  }
+}
+```
+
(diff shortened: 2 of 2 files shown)


crates/tauri-utils/src/resources.rs
```diff
@@ -209,7 +209,21 @@ impl ResourcePathsIter<'_> {
...
+          None => {
+            if dest.components().count() == 0 {
+              // if current_dest is empty while processing a file pattern
+              // we preserve the file name as it is
+              //
+              // e.g. `{ "README.md": "" }` is `README.md` -> `$RESOURCE/README.md`
+              //
+              // TODO: This behavior is a confusing special case,
+              // remove this in v3 or make other cases like this work
+              // > `{ "README.md": "./folder/" }` is `README.md` -> `$RESOURCE/folder/README.md` (this gives `$RESOURCE/folder` today)
```
amrbashir, line 221: Can you explain how is this confusing? for me at least I think of:
1. `README.md` -> `./` or just `.` as `$Resource/README.md`
2. `README.md` -> `./folder` as `$Resource/folder/README.md` since it is a file
3. `directory` -> `./` or just `.` as `$Resource/directory`

The one I actually think of as confusing is this:
4. `directory` -> `./folder` as `$Resource/folder`, it should be `$Resource/folder/directory`

Overall I think we should be following the behavior of `cp` command on unix

amrbashir (approved):

crates/tauri-utils/src/resources.rs
Legend-Master: > 2. `README.md` -> `./folder` as `$Resource/folder/README.md` since it is a file

We can not tell if the target `./folder` is a folder (we can if it's written as `./folder/` though), so it gives `$RESOURCE/folder` today, and that's the confusing part

Legend-Master: > 2. `README.md` -> `./folder` as `$Resource/folder/README.md` since it is a file
> 3. `directory` -> `./` or just `.` as `$Resource/directory`

I don't think these were the case? At least not when I wrote https://tauri.app/develop/resources/#tab-panel-715 🤔 (I wrote it based on the actual behavior we had I believe)

amrbashir: I really don't know how it behaves atm, but if we plan to break this in a new major, I would just suggest being close as possible to `cp` command, it makes it more obvious and familiar and less confusing

Legend-Master: That makes sense, `{ "README.md": "./folder/" }` must be written as `./folder/` not `./folder` to put it at `$RESOURCE/folder/README.md` though

PR #15395, opened 2026-05-16
feat: add configuration `bundle.unwatchedResources`

### LINKED ISSUE #15390: [feat] Support skip resource watching
### Describe the problem

My app needs bundling a Python embed runtime with tons of site-packages (~4000 dirs). After adding the dir to `tauri.conf.json > bundle > resources`, commands `cargo:rerun-if-changed` messed up the terminal.

https://github.com/tauri-apps/tauri/blob/3b5b2cc125c19cefb23f618d26fd09bfce5db538/crates/tauri-build/src/lib.rs#L63-L64

```diff
 {
   "bundle": {
+    "unwatchedResources": [/* glob */],
   },
 }
```

### Alternatives considered

_No response_

### Additional context

_No response_

typed-sigterm:
closes #15390

### DIFF
--- a/.changes/unwatched-resources.md
+++ b/.changes/unwatched-resources.md
@@ -1,0 +1,9 @@
+---
+'tauri': minor
+'tauri-build': minor
+'tauri-cli': minor
+'tauri-schema-generator': minor
+'tauri-utils': minor
+---
+
+Add configuration `bundle.unwatchedResources`.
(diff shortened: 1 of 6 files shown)


Legend-Master (commented):
I do agree that this is something we want to address, but I'm not if this is the best way to do it...

Legend-Master:
I wonder if #15629 is a better fix than this, instead of emitting `rerun-if-changed` on every file, we do it on the directory level

PR #15411, opened 2026-05-18
fix: keep tao window_target in main thread

### LINKED ISSUE #15408: [bug] WebView2 custom protocol can clone DispatcherMainThreadContext off the main thread
### Describe the bug

A Windows WebView2 custom protocol request can call Tauri's custom protocol handler from a non-main thread. If the handler path looks up and clones a `Webview`, this can clone `tauri_runtime_wry::DispatcherMainThreadContext` off the main thread and trip the unsafe `Rc` precondition inside `tao::EventLoopWindowTarget`.

In our app this was observed as an unsafe precondition violation around `Rc::inc_strong` / `strong_count == 0` while cloning a Tauri webview during custom protocol handling.

The suspected chain is:

``text
Tauri::Webview::clone
  -> DetachedWebview::clone
    -> WryWebviewDispatcher::clone
      -> Context::clone
        -> DispatcherMainThreadContext::clone
          -> EventLoopWindowTarget::clone
            -> Rc<EventLoopRunner>::clone
``

`DispatcherMainThreadContext` has unsafe `Send` / `Sync` impls with the safety comment saying the type is only used on the main thread, but the derived `Clone` can still be reached from non-main-thread custom protocol callbacks.

### Source locations checked

This is still present in the latest published crates at the time of filing:

- `tauri-runtime-wry 2.11.2`, `src/lib.rs`:
  - `Context<T>` is `#[derive(Clone)]` and contains `main_thread: DispatcherMainThreadContext<T>`.
  - `DispatcherMainThreadContext<T>` is `#[derive(Debug, Clone)]` and contains `window_target: EventLoopWindowTarget<Message<T>>`.
  - `DispatcherMainThreadContext<T>` has unsafe `Send` and `Sync` impls with the comment `we ensure this type is only used on the main thread`.
  - The adjacent field `windows` is already an `Arc<WindowsStore>` with the comment `changing this to an Rc will cause frequent app crashes`, but `window_target` still contains tao's internal `Rc` state.
- `tauri-runtime 2.11.2`, `src/webview.rs`:
  - `DetachedWebview::clone` clones `dispatcher`.
- `tauri 2.11.2`, `src/webview/mod.rs`:
  - `Webview::clone` clones the `DetachedWebview`.
- `wry 0.55.1`, `src/webview2/mod.rs`, custom protocol handling:
  - The async response setter dispatches back to the main thread when needed.
  - But `custom_protocol_handler(&webview_id, request, RequestAsyncResponder { ... })` itself is invoked directly on the current `WebResourceRequested` callback thread.

### Why this seems unsafe

The Tauri runtime safety invariant appears to be that `DispatcherMainThreadContext` is only used on the main thread. However, on Windows/WebView2, `WebResourceRequested` can be observed during shutdown/restart or internal WebView2 work from a non-main thread. In that case, the wry custom protocol handler is called without being dispatched to the main thread first.

If app/framework code inside that handler resolves a webview and clones it, `DispatcherMainThreadContext::clone` clones `EventLoopWindowTarget`, which ultimately clones non-atomic `Rc` state from the wrong thread. This violates the safety invariant.

There is an older related issue, #10001, also titled around `DispatcherMainThreadContext` unsafe precondition violations, but the path described here is different: WebView2 custom protocol handling -> webview clone -> `window_target` clone.

### Expected behavior

Custom protocol handling should not be able to clone or otherwise touch `DispatcherMainThreadContext` / `EventLoopWindowTarget` off the main thread.

### Possible fixes

A few possible directions, depending on intended ownership between Tauri/wry/tao:

1. Dispatch the wry `custom_protocol_handler` invocation itself onto the WebView/main thread on Windows, not only the async response setter.
2. Avoid deriving `Clone` for `DispatcherMainThreadContext` / `Context` in a way that clones `EventLoopWindowTarget` from arbitrary threads.
3. Split main-thread-only state out of dispatcher handles so cloning a `Webview`/dispatcher from protocol callbacks does not clone tao `Rc` state.

### Versions

Our affected lockfile currently has:

``text
tauri = 2.10.3
tauri-runtime-wry = 2.10.1
wry = 0.54.4
tao = 0.34.8
``

I also checked the latest published crates and the relevant code shape still appears unchanged:

``text
tauri = 2.11.2
tauri-runtime-wry = 2.11.2
wry = 0.55.1
tao = 0.35.x
``

### Platform

Windows, WebView2 backend.

### Additional context

I do not yet have a minimal public reproduction project. This issue is filed from a crash investigation plus source-level tracing. The important part is that `custom_protocol_handler` can run outside the main thread while Tauri's `Webview::clone` path can clone main-thread-only tao state.

Legend-Master:
An attempt to fix #15408

By keeping an `Arc` in `Wry` and only keep an `Weak` in `Context`, we make sure we only ever upgrade/use/drop the `window_target` on main thread. (previously, the clone/drop can be made in another thread from cloning and sending to another thread)


crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -2784,14 +2790,16 @@ impl<T: UserEvent> RuntimeHandle<T> for WryHandle<T> {
   fn display_handle(
     &self,
   ) -> std::result::Result<DisplayHandle<'_>, raw_window_handle::HandleError> {
-    self.context.main_thread.window_target.display_handle()
+    todo!();
```
Legend-Master: Don't know how can we deal with this life time issue though...

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -2926,12 +2936,14 @@ impl<T: UserEvent> Wry<T> {
...
 
+    let window_target = Arc::new(event_loop.deref().clone());
+
     let context = Context {
       window_id_map,
       main_thread_id,
       proxy: event_loop.create_proxy(),
       main_thread: DispatcherMainThreadContext {
-        window_target: event_loop.deref().clone(),
+        window_target: Arc::downgrade(&window_target),
```
Tunglies, line 2914: Do we actually need Arc? Since keep in main thread isn't cross thread. Or something like `run_on_main` ?

https://github.com/tauri-apps/tao/blob/1080241d20d6bd0044d1cdb5633a25a3560f31cb/src/platform_impl/macos/util/async.rs#L43-L49

Legend-Master: The thing here is this context can be cloned and moved to other threads, and the inner rc from tao can't be cloned/dropped from multiple threads

RyanJamesStewart:
Took the branch for a spin on Windows (`x86_64-pc-windows-msvc`, WebView2 148). It builds clean. For the `display_handle` lifetime question you flagged, the piece that compiles is filling the `todo!()` by upgrading the weak `window_target` and re-borrowing the raw handle:

```rust
fn display_handle(
  &self,
) -> std::result::Result<DisplayHandle<'_>, raw_window_handle::HandleError> {
  let window_target = self
    .context
    .main_thread
    .window_target
    .upgrade()
    .ok_or(raw_window_handle::HandleError::Unavailable)?;
  let raw = window_target.display_handle()?.as_raw();
  // SAFETY: `raw` is a plain `Copy` `RawDisplayHandle` pointing at the event
  // loop's display connection, which stays valid for as long as the runtime
  // (and therefore `&self`) is alive.
  Ok(unsafe { DisplayHandle::borrow_raw(raw) })
}
```

The `RawDisplayHandle` is `Copy`, so this sidesteps tying the returned `DisplayHandle<'_>` to the temporary `Arc` from `upgrade()`; the borrow is anchored to `&self` instead. (Restores the `HasDisplayHandle` import the branch had dropped.) Happy to push it as a commit if useful.

Legend-Master:
Well, with my change, the lifetime of the `EventLoopWindowTarget` is no longer related to the `WryHandle` which means this can cause dangling handles

That being said, even though the old lifetime was correct, it carried the `EventLoopWindowTarget` to the wrong thread to clone and drop

RyanJamesStewart:
You're right, and thanks for the catch. The `upgrade()`-then-`borrow_raw` was unsound: `&self` on a `Weak`-only `WryHandle` doesn't keep the `EventLoopWindowTarget` alive, so on X11/Wayland the borrowed connection pointer can dangle if the main thread drops the loop. The Windows-empty handle hid it in my test.

Digging into what the `todo!()` replaced: before this PR it was `window_target.display_handle()` on the owned target, which returned a working handle on every platform, and `AppHandle`'s public `HasDisplayHandle` impl routes here (it holds a `RuntimeHandle`, not the `Runtime`). So a blanket `Err(Unavailable)` would regress `AppHandle::display_handle()` on Windows/macOS, where the empty handle is the correct value and the HWND/NSView comes off the window handle.

So I think the real choice is binary:

1. **Empty `WindowsDisplayHandle`/`AppKitDisplayHandle` on Windows/macOS, `Err(HandleError::Unavailable)` on X11/Wayland.** Those variants carry no pointer, so borrowing them for any lifetime is sound; this restores prior behavior where it can and degrades only the off-thread Linux case.
2. **Split `HasDisplayHandle` off `RuntimeHandle`** so only main-thread holders (which keep the strong `Arc`) implement it, and serve X11/Wayland soundly there. That's a separate, larger refactor touching the public trait surface.

I'd lean toward (1), scoped to this PR: it's the smallest sound change that keeps the public API working where it can, and the X11/Wayland off-thread path is the one this PR is already making safe, so degrading it to a clean `Err` is honest rather than a real capability loss. (2) feels like its own PR if you want the Linux off-thread handle back.

Either way, scratch my earlier offer to push the `borrow_raw` version as a commit. How would you prefer to shape it?

Legend-Master:
[`HasWindowHandle`](https://docs.rs/raw-window-handle/latest/raw_window_handle/trait.HasWindowHandle.html) requires the handle to be valid as long as the object implementing it which can't be achieved with this weak pointer approach.

I think the direction would be either try another solution for the implicit `window_target` access, or just break the promise (go with the one you proposed) if the trade off is worth it

RyanJamesStewart:
Thanks, agreed the weak pointer can't satisfy the rwh lifetime contract in general.

The split I'd suggest: Windows and macOS return the empty `WindowsDisplayHandle`/`AppKitDisplayHandle`, which carry no pointer, so they're sound for any lifetime and keep the prior behavior with no promise broken. X11/Wayland are the only arm with a real connection pointer.

So the actual decision is just that one arm:

1. `Err(HandleError::Unavailable)` on X11/Wayland: fully sound, off-thread display_handle stops working there (it was already unsound pre-PR via the cross-thread `Rc` path this fixes).
2. `upgrade()` + `borrow_raw` on X11/Wayland with a SAFETY note that the handle must not outlive the runtime: restores prior behavior, breaks the lifetime promise only on Linux off-thread, which is the case you noted the old code was already relying on.

I lean toward (2): it keeps Windows/macOS sound and serves Linux as before, rather than dropping the Linux off-thread path to `Unavailable` when Windows/macOS can stay working. But that's a documented-unsoundness call on a thin slice (off-thread X11/Wayland surface init is rare), so if you'd rather stay strictly sound, (1) is reasonable too. Which would you prefer?

Legend-Master:
Yeah, a small unsoundness would out weight breaking everyone using that API (your `2`), if we decide to still use the weak pointer approach

RyanJamesStewart:
Here's what I'd push. Flagging the SAFETY note for your review since it's the load-bearing part. Good to commit?

```rust
fn display_handle(&self) -> Result<DisplayHandle<'_>, HandleError> {
  // Windows/macOS: the display handle carries no pointer, so borrowing
  // one is sound for any lifetime regardless of the Weak event loop.
  #[cfg(windows)]
  return Ok(unsafe { DisplayHandle::borrow_raw(
    RawDisplayHandle::Windows(WindowsDisplayHandle::new())) });
  #[cfg(target_os = "macos")]
  return Ok(unsafe { DisplayHandle::borrow_raw(
    RawDisplayHandle::AppKit(AppKitDisplayHandle::new())) });

  #[cfg(not(any(windows, target_os = "macos")))]
  {
    let window_target = self
      .context
      .main_thread
      .window_target
      .upgrade()
      .ok_or(HandleError::Unavailable)?;
    let raw = window_target.display_handle()?.as_raw();
    // SAFETY: `raw` borrows the X11/Wayland connection pointer owned by
    // the main-thread EventLoopWindowTarget. The Weak above can't keep
    // that target alive, so this technically violates the rwh lifetime
    // contract: callers must not retain or use this handle past a point
    // where the main thread may have dropped the runtime. In practice
    // this holds for typical surface-init consumers, which use the
    // handle synchronously and don't outlive the runtime. Going via the
    // Weak (rather than dropping to Err) preserves prior behavior on
    // Linux, which was already de-facto relied on.
    Ok(unsafe { DisplayHandle::borrow_raw(raw) })
  }
}
```

Plus a one-line restoration of the `HasDisplayHandle` import the branch dropped, and `WindowsDisplayHandle` / `AppKitDisplayHandle` imports for the Windows/macOS arms.

One thing I want to be straight about on the SAFETY note: "callers must not retain or use this handle past a point where the main thread may have dropped the runtime" is a precondition a wgpu/softbuffer/glutin caller can't actually verify on its own. That's why this arm is documented-unsound rather than sound; there's no formulation that closes the gap without restructuring the trait. The note acknowledges the gap rather than pretending to close it, which is the trade we settled on.

Legend-Master:
Just so we're clear, I would like to hold onto this one see if we can come up with another method (not the weak pointer approach used here) before preceding with this unsound implemention 
And the original one posted in https://github.com/tauri-apps/tauri/pull/15411#issuecomment-4581898174 would be good enough just with some more documentations

RyanJamesStewart:
I went looking for a non-weak method before adding docs, and the useful finding is that two axes are getting conflated: the drop fix and `display_handle` soundness are orthogonal. A non-weak redesign of `window_target` ownership is a legitimate way to fix #15408 and is your call, but it won't make `display_handle` sound on X11/Wayland, because the wall there is independent of weak-vs-strong: the method returns a `Send`-anchored borrow of a display connection owned by the `!Send` main-thread `EventLoopWindowTarget`. Anything reachable from the `Send + Sync` off-thread handle inherits that lifetime gap.

For the off-thread X11/Wayland arm specifically, there are two in-repo endings:

1. `upgrade()` + `borrow_raw` the live connection: documented-unsound, restores prior behavior, confined to that path.
2. Return `Err(HandleError::Unavailable)`: sound, no compile break, keeps the trait impl. This isn't a regression on that arm. The pre-PR behavior there was the cross-thread `Rc` clone/drop that #15408 exists to fix, i.e. live UB, so there were no sound off-thread X11/Wayland callers to lose. `Err` strictly removes unsoundness from a path that had it, with zero behavior loss for any sound caller.

And two that are out of scope for this PR: a tao/winit refactor that splits the connection into a `Send + Sync` object the off-thread handle can hold strongly (sound, non-breaking, but upstream), or dropping the trait impl entirely (sound but a compile break).

My recommendation is `Err` now, with the sound capability tracked as a tao follow-up. The "don't break callers" concern from earlier was about not regressing sound Windows/macOS surface-init, and neither arm touches that, since both keep the empty handle while the runtime is alive. It was never about an off-thread X11/Wayland consumer, because the UB means there isn't a sound one to protect. So `borrow_raw` is only worth it if you can name a workflow that needs the handle off-thread on X11/Wayland today, and for what it's worth I can't from my side: I'm on Windows/WebView2 with no off-thread X11/Wayland consumer.

On the Win/Mac arm, either ending returns the empty handle while the runtime is alive. One behavior delta worth making intentional rather than incidental: routing Win/Mac through `upgrade()` means a dead runtime returns `Err` where the owned target previously returned a valid-but-empty handle. Arguably more correct, but it's a change outside the X11/Wayland scope, so worth stating on purpose.

If you take `Err`, the PR text just needs a capability note: off-thread X11/Wayland `display_handle` declines pending a tao follow-up. If you'd rather keep `borrow_raw`, I'd instead state plainly that it reintroduces a known, pre-existing unsoundness confined to that path, linked to the follow-up, so the SAFETY note isn't later read as a resolution. Glad to draft whichever wording your call selects.

sftse:
Signaling my intent to get up to speed on this PR.

crates/tauri-runtime-wry/src/lib.rs
Legend-Master: @FabianLars ok so the wrong lifetime was already the case for the `WryWindowDispatcher::window_handle` that it doesn't own the tao `Window`. What do you think about breaking this as well 😂?

FabianLars: don't care 😂

Legend-Master: Oh well, one more thing, should we keep the old behavior that returns a `DisplayHandle` even if called off main thread or should we return `Unavailable` here?

PR #15412, opened 2026-05-18
Use workspace dependency management

sftse:
We can simplify dependency management a bit by centralizing the workspace-internal dependencies.


FabianLars:
I was just about to ping @jbolda to ask whether covector supports this but CI already answered my question 😂

FabianLars:
I do like this approach though so maybe we can fix covector

Tunglies:
Any plan for workspace lint rules?

Legend-Master:
> Any plan for workspace lint rules?

@Tunglies Not yet, you can open an issue on which lint rules do you want to enable and we can discuss there

> I was just about to ping @jbolda to ask whether covector supports this but CI already answered my question 😂

@FabianLars I couldn't remember what was the problem back then, it seems to work at least in this run https://github.com/tauri-apps/tauri/actions/runs/28021124759/job/82937404412?pr=15412

Legend-Master (commented):
I like this a lot, but we probably need to fix the feature flags first since some of them actually depends on the default features

FabianLars:
This was the error https://github.com/tauri-apps/tauri/actions/runs/26026811997/job/76506308250 - i think it still does not work but there are no subcrate (eg tauri-utils instead of tauri) changefiles anymore so it doesn't check.

sftse:
> I like this a lot, but we probably need to fix the feature flags first since some of them actually depends on the default features

Can you explain?

Just seen I missed a spot or two that are using their own paths instead of the workspace deps.

bench/tests/cpu_intensive/src-tauri/Cargo.toml
```diff
@@ -6,11 +6,9 @@ edition = "2021"
...
-tauri-build = { path = "../../../../crates/tauri-build", features = [
-  "codegen",
-] }
+tauri-build = { workspace = true, features = ["codegen"] }
 
 [dependencies]
 serde_json = "1"
 serde = { version = "1", features = ["derive"] }
-tauri = { path = "../../../../crates/tauri", features = [] }
+tauri = { workspace = true }
```
Legend-Master: This needs wry, as well as the other bench crates

examples/api/src-tauri/tauri-plugin-sample/Cargo.toml
```diff
@@ -5,12 +5,10 @@ edition = "2021"
 links = "tauri-plugin-sample"
 
 [dependencies]
-tauri = { path = "../../../../crates/tauri" }
+tauri = { workspace = true }
```
Legend-Master, line 8: This should also require wry

examples/api/src-tauri/Cargo.toml
```diff
@@ -11,10 +11,7 @@ name = "api_lib"
 crate-type = ["staticlib", "cdylib", "rlib"]
 
 [build-dependencies]
-tauri-build = { path = "../../../crates/tauri-build", features = [
-  "codegen",
-  "isolation",
-] }
+tauri-build = { workspace = true, features = ["codegen", "isolation"] }
```
Legend-Master: This as well

Legend-Master (commented):
pretty much every example basically

jbolda:
Following up here as we had conversations on other mediums. Working through this support in `covector` which should land soon. Had a lot of blocking upgrades that I needed to land first. They did though and @johncarmack1984 already had [the PR rolling](https://github.com/jbolda/covector/pull/397) for adding this support 🙌 .

FabianLars:
covector 0.13 was released so we can try this again

Legend-Master:
I seem to get some errors running `cargo publish --dry-run` in `./crates/tauri/` 🤔

> ```
> error: failed to prepare local package for uploading                                                                                                      
> 
> Caused by:
>   failed to select a version for `tauri`.
>       ... required by package `tauri v2.11.5 (tauri/crates/tauri)`
>   versions that meet the requirements `^2.11` are: 2.11.5, 2.11.4, 2.11.3, 2.11.2, 2.11.1, 2.11.0
> 
>   package `tauri` links to the native library `Tauri`, but it conflicts with a previous package which links to `Tauri` as well:
>   package `tauri v2.11.5 (C:\tt\Temp Fix\tauri\crates\tauri)`
>   note: only one package in the dependency graph may specify the same links value to ensure that only one copy of a native library is linked in the final binary
>   for more information, see https://doc.rust-lang.org/cargo/reference/resolver.html#links
>   help: try to adjust your dependencies so that only one package uses the `links = "Tauri"` value
> 
>   failed to select a version for `tauri` which could resolve this conflict
> ```

crates/tauri-macros/Cargo.toml
```diff
@@ -20,8 +20,8 @@ proc-macro2 = { version = "1", features = ["span-locations"] }
 quote = "1"
 syn = { version = "2", features = ["full"] }
 heck = "0.5"
-tauri-codegen = { version = "2.6.3", default-features = false, path = "../tauri-codegen" }
```
sftse, line 23: `tauri-macros` used to not depend on default features of `tauri-codegen`, now with workspace dependency all these features are turned on.

crates/tauri-cli/Cargo.toml
```diff
@@ -47,7 +47,7 @@ sublime_fuzzy = "0.7"
 clap_complete = "4"
 clap = { version = "4", features = ["derive", "env"] }
 thiserror = "2"
-tauri-bundler = { version = "2.9.4", default-features = false, path = "../tauri-bundler" }
+tauri-bundler = { workspace = true }
```
sftse, line 50: ~~`tauri-cli` used to not depend on default features of `tauri-bundler`, now with workspace dependency all these features are turned on.~~ misread dependencies, this is correct as-is.

sftse (commented):
What was the issue with `cargo publish` ?

crates/tauri-macros/Cargo.toml
Legend-Master: `tauri-codegen` has no default features so

Legend-Master:
> What was the issue with `cargo publish` ?

I don't quite understand it yet, see the error in my comment above

Legend-Master:
> I seem to get some errors running `cargo publish --dry-run` in `./crates/tauri/` 🤔
> 
> > ```
> > error: failed to prepare local package for uploading                                                                                                      
> > 
> > Caused by:
> >   failed to select a version for `tauri`.
> >       ... required by package `tauri v2.11.5 (tauri/crates/tauri)`
> >   versions that meet the requirements `^2.11` are: 2.11.5, 2.11.4, 2.11.3, 2.11.2, 2.11.1, 2.11.0
> > 
> >   package `tauri` links to the native library `Tauri`, but it conflicts with a previous package which links to `Tauri` as well:
> >   package `tauri v2.11.5 (C:\tt\Temp Fix\tauri\crates\tauri)`
> >   note: only one package in the dependency graph may specify the same links value to ensure that only one copy of a native library is linked in the final binary
> >   for more information, see https://doc.rust-lang.org/cargo/reference/resolver.html#links
> >   help: try to adjust your dependencies so that only one package uses the `links = "Tauri"` value
> > 
> >   failed to select a version for `tauri` which could resolve this conflict
> > ```

Resolved in b692576, pretty weird not gonna lie

Legend-Master:
@jbolda do you know if we can dry run this to test how it does with covector?

Cargo.toml
```diff
@@ -45,6 +45,18 @@ license = "Apache-2.0 OR MIT"
...
+tauri-runtime = { version = "2.11.3", path = "./crates/tauri-runtime" }
+tauri-runtime-wry = { version = "2.11.4", path = "./crates/tauri-runtime-wry", default-features = false }
+tauri-build = { version = "2.6.3", path = "./crates/tauri-build", default-features = false }
+tauri-macros = { version = "2.6.3", path = "./crates/tauri-macros" }
+tauri-codegen = { version = "2.6.3", path = "./crates/tauri-codegen" }
+tauri-utils = { version = "2.9.3", path = "./crates/tauri-utils" }
+tauri-plugin = { version = "2.6.3", path = "./crates/tauri-plugin", default-features = false }
+tauri-bundler = { version = "2.9.4", path = "./crates/tauri-bundler", default-features = false }
+tauri-macos-sign = { version = "2.3.4", path = "./crates/tauri-macos-sign" }
+
```
sftse, line 59: The patch versions aren't enforced by `cargo`. Either we need to keep them in sync by updating them when the patch version in the `Cargo.toml` of the dependency changes or they become silently misleading.

Legend-Master: They should be bumped by covector, I would like to probably pin them to minor versions as well. For example https://github.com/tauri-apps/tauri/commit/6f6ab1207bb3923c2721fbc67d2fdb1c8deb0c7a

jbolda:
@Legend-Master Yea, run it locally with `npx` or the bin directly with `covector version --dry-run` or `covector publish --dry-run`.

Legend-Master:
They seem to work for me locally, thanks! @jbolda

Legend-Master (dismissed):

Legend-Master (approved):

sftse:
@Legend-Master 
Thanks for the fixups!

PR #15413, opened 2026-05-18
refactor: reduced some clones

Legend-Master:

### DIFF
--- a/.changes/runtime-wry-avoid-clones.md
+++ b/.changes/runtime-wry-avoid-clones.md
@@ -1,0 +1,5 @@
+---
+"tauri-runtime-wry": "minor:changes"
+---
+
+`EventLoopIterationContext` and `UserMessageContext` now takes references to avoid clones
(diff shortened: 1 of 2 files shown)


sftse:
I think this is the right direction to go in more in general. Lifetimes seem like a nuisance but help discover hierarchically well structured designs that are much less likely to lead to deadlocks and memory leaks.

I have similar changes I think we should make heading towards v3.

Legend-Master:
Yeah, that's what I thought as well, these changes were made so I could see where are these things used and the flow more clearly

Tunglies:
From `Arc<WindowsStore>` to `&WindowsStore` might changed behavior. 
```rust 
pub struct WindowsStore(pub RefCell<BTreeMap<WindowId, WindowWrapper>>)
```
Now we cannot move ownership cross-thread, forcing processing in single thread stack.

Legend-Master:
> Now we cannot move ownership cross-thread, forcing processing in single thread stack.

Yeah, that's the whole point, they should not be crossing to another thread

sftse:
> Now we cannot move ownership cross-thread, forcing processing in single thread stack.

To be precise, this PR still allows multithreaded access. #14805 will remove it, so `Arc` becomes dead weight.

Tunglies:
Thanks for brings more context, now I see it.

sftse (approved):

PR #11676, opened 2024-11-13
fix(bundler): sign DLLs

thewh1teagle:
Fix https://github.com/tauri-apps/tauri/issues/11673

Now it's signed after bundling:

Log:

<details>

```console
..\..\tauri\target\debug\cargo-tauri.exe bundle
    Signing D:\vibe\target\release\vibe.exe
    Signing D:\vibe\target\release\vibe.exe with identity "45DCFF335BB538C58489BD03BF167D29DDA53110"
    Info "Done Adding Additional Store\r\nSuccessfully signed: D:\\vibe\\target\\release\\vibe.exe\r\r\n"
    Warn NSIS directory contains mis-hashed files. Redownloading them.
    Downloading https://github.com/tauri-apps/nsis-tauri-utils/releases/download/nsis_tauri_utils-v0.4.1/nsis_tauri_utils.dll
    Info validating hash
    Info Target: x64
    Info Signing NSIS plugins
    Signing C:\Users\User\AppData\Local\tauri\NSIS\Plugins\x86-ansi\NSISdl.dll
    Signing C:\Users\User\AppData\Local\tauri\NSIS\Plugins\x86-ansi\NSISdl.dll with identity "45DCFF335BB538C58489BD03BF167D29DDA53110"
    Info "Done Adding Additional Store\r\nSuccessfully signed: C:\\Users\\User\\AppData\\Local\\tauri\\NSIS\\Plugins\\x86-ansi\\NSISdl.dll\r\r\n"
    Signing C:\Users\User\AppData\Local\tauri\NSIS\Plugins\x86-ansi\StartMenu.dll
    Signing C:\Users\User\AppData\Local\tauri\NSIS\Plugins\x86-ansi\StartMenu.dll with identity "45DCFF335BB538C58489BD03BF167D29DDA53110"
    Info "Done Adding Additional Store\r\nSuccessfully signed: C:\\Users\\User\\AppData\\Local\\tauri\\NSIS\\Plugins\\x86-ansi\\StartMenu.dll\r\r\n"
    Signing C:\Users\User\AppData\Local\tauri\NSIS\Plugins\x86-ansi\System.dll
    Signing C:\Users\User\AppData\Local\tauri\NSIS\Plugins\x86-ansi\System.dll with identity "45DCFF335BB538C58489BD03BF167D29DDA53110"
[...]

PR #15422, opened 2026-05-20
fix(nsis): embed signed copies of stock plugins, not unsigned system DLLs

### LINKED ISSUE #14147: [bug] nsis plugins not signed
### Describe the problem

Some Antivirus programs, 360 etc, would block the loading of the plugin from nsis installer silently. From my perspective, adds a digital cert may solve this issue.

Image

* plugin was not loaded
* so the killProcess and strReplacer is not work

Image

Add a option in bundler config, allowing codesign tool to sign nsis plugins.

I can draft a pr to support this case if needed.

### Alternatives considered

_No response_

### Additional context

_No response_

koki-develop:
## Summary

closes #14147

When an NSIS bundle is produced with `signCommand` configured, the stock plugin DLLs (`NSISdl.dll`, `StartMenu.dll`, `System.dll`, `nsDialogs.dll`) unpacked into `$PLUGINSDIR` at install time end up **unsigned**. PR #14627 made the sign step itself run, but the signed local copies were never embedded into the resulting setup.exe.

## Root cause

When `signCommand` is enabled, the bundler copies the system NSIS plugins to `<output>/Plugins/x86-unicode/` and signs them in place. However, `installer.nsi` only has an `!addplugindir` for the `additional/` subdirectory — nothing points makensis at the directory holding the signed stock plugins. makensis therefore falls back to `${NSISDIR}/Plugins/<arch>/` and embeds the unsigned originals.

The signed local copies are never referenced by makensis, so the unsigned versions are what end up inside setup.exe.

## Fix

- `installer.nsi`: between `!include "StrFunc.nsh"` and `${StrCase}`, add an `!addplugindir` for the signed plugin directory (gated on signing being enabled)
- `mod.rs`: expose the matching `signed_plugins_path` template variable only when signing is enabled

## Also: drop the `NSISPLUGINS` env line

Before launching makensis, `mod.rs` set `nsis_cmd.env("NSISPLUGINS", plugins_path)`, but `NSISPLUGINS` is not an environment variable that NSIS recognizes. Grepping the entire `NSIS-Dev/nsis` source tree and running `gh search code --owner NSIS-Dev 'NSISPLUGINS'` both return zero hits — the line was a complete no-op.

The intent was presumably to communicate the local-copy plugin directory to makensis, but the proper route is now established by the `!addplugindir` change above, so this dead line is removed as well.

### DIFF
--- a/crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
+++ b/crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
@@ -13,6 +13,12 @@ ManifestDPIAwareness PerMonitorV2
   SetCompressor /SOLID "{{compression}}"
 !endif
 
+; Keep above !include to stay ahead of any plugin command
+; see https://github.com/tauri-apps/tauri/pull/15422#discussion_r3289239624
+{{#if signed_plugins_path}}
+!addplugindir "{{signed_plugins_path}}"
+{{/if}}
+
 !include MUI2.nsh
 !include FileFunc.nsh
 !include x64.nsh
@@ -22,9 +28,6 @@ ManifestDPIAwareness PerMonitorV2
 !include "Win\COM.nsh"
 !include "Win\Propkey.nsh"
 !include "StrFunc.nsh"
-{{#if signed_plugins_path}}
-!addplugindir "{{signed_plugins_path}}"
-{{/if}}
 ${StrCase}
 ${StrLoc}
 
--- a/crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
+++ b/crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
@@ -22,6 +22,9 @@ ManifestDPIAwareness PerMonitorV2
 !include "Win\COM.nsh"
 !include "Win\Propkey.nsh"
 !include "StrFunc.nsh"
+{{#if signed_plugins_path}}
+!addplugindir "{{signed_plugins_path}}"
+{{/if}}
 ${StrCase}
 ${StrLoc}
 
--- a/crates/tauri-bundler/src/bundle/windows/nsis/mod.rs
+++ b/crates/tauri-bundler/src/bundle/windows/nsis/mod.rs
@@ -282,6 +282,13 @@ fn build_nsis_app_installer(
     to_json(&additional_plugins_path),
   );
 
+  if let Some(plugin_copy_path) = &maybe_plugin_copy_path {
+    data.insert(
+      "signed_plugins_path",
+      to_json(plugin_copy_path.join("x86-unicode")),
+    );
+  }
+
   data.insert("arch", to_json(arch));
   data.insert("bundle_id", to_json(bundle_id));
   data.insert("manufacturer", to_json(manufacturer));
@@ -682,10 +689,6 @@ fn build_nsis_app_installer(
   let mut nsis_cmd = Command::new(nsis_toolset_path.join("makensis.exe"));
   #[cfg(not(target_os = "windows"))]
   let mut nsis_cmd = Command::new("makensis");
-
-  if let Some(plugins_path) = &maybe_plugin_copy_path {
-    nsis_cmd.env("NSISPLUGINS", plugins_path);
-  }
 
   nsis_cmd
     .args(["-INPUTCHARSET", "UTF8", "-OUTPUTCHARSET", "UTF8"])
--- a/.changes/nsis-stock-plugins-embed-signed.md
+++ b/.changes/nsis-stock-plugins-embed-signed.md
@@ -1,0 +1,5 @@
+---
+"tauri-bundler": "patch:bug"
+---
+
+Fix NSIS stock plugins (`NSISdl.dll`, `StartMenu.dll`, `System.dll`, `nsDialogs.dll`) being embedded in the final installer as unsigned despite the signing step succeeding. The signed local copies under `<output>/Plugins/x86-unicode/` were not on makensis' plugin search path, so makensis fell back to the unsigned DLLs from the NSIS toolset directory. The fix adds `!addplugindir` for the signed plugin directory before any plugin command is parsed in the script.


Legend-Master:
https://github.com/tauri-apps/tauri/issues/14147#issuecomment-3259721138 seems like the signing failed not that they were not included?

koki-develop-lot:
@Legend-Master Thanks for taking a look.

You're right that #14147 itself is an issue about signing failing, and that part was resolved by #14627.

What this PR is targeting is a separate problem that's left over: #14627 made the signing step itself succeed, but **those signed copies aren't on makensis' plugin search path, so as a result makensis falls back to the unsigned originals and embeds those instead**.

> PR https://github.com/tauri-apps/tauri/pull/14627 made the sign step itself run, but the signed local copies were never embedded into the resulting setup.exe.
> > https://github.com/tauri-apps/tauri/pull/15422#issue-4486263047

Legend-Master:
cc @lucasfernog since you wrote the change to use this environment variable in https://github.com/tauri-apps/tauri/pull/11676/changes/37b4ca1b58d62fa201e666c62df34f83d3ad8eab, I can't seem to find any references of `NSISPLUGINS` online 🤔

koki-develop-lot:
I can't find any either - https://github.com/search?q=org%3ANSIS-Dev+%22NSISPLUGINS%22&type=code

crates/tauri-bundler/src/bundle/windows/nsis/mod.rs
Legend-Master, line 280: Actually, didn't we already set this path for `!addplugindir`?

koki-develop: @Legend-Master The existing `additional_plugins_path` and the new `signed_plugins_path` resolve to **different directories**.

When signing is enabled, the bundler copies plugins into two separate locations:

| Variable | Resolved path | Contents |
|---|---|---|
| `additional_plugins_path` (existing) | `<output>/Plugins/x86-unicode/additional/` | `nsis_tauri_utils.dll` only |
| `signed_plugins_path` (new) | `<output>/Plugins/x86-unicode/` | stock plugins (`NSISdl` / `StartMenu` / `System` / `nsDialogs`) |

https://github.com/tauri-apps/tauri/blob/41fdca7d1991cf4b4b69a5a758a5371333218487/crates/tauri-bundler/src/bundle/windows/nsis/mod.rs#L201
https://github.com/tauri-apps/tauri/blob/41fdca7d1991cf4b4b69a5a758a5371333218487/crates/tauri-bundler/src/bundle/windows/nsis/mod.rs#L244-L260

The stock plugins live **one level above** where `additional_plugins_path` points, so the existing `!addplugindir` doesn't cover them.

koki-develop:
Just leaving the verification steps here for reference (on macOS).

Extract the plugin DLLs from a `setup.exe` built with signing enabled, and check the signature status of each DLL.

```sh
# extract plugin DLLs from setup.exe
7z x setup.exe -oextracted

# verify each DLL
for dll in extracted/\$PLUGINSDIR/*.dll; do
  echo "=== $(basename "$dll") ==="
  osslsigncode verify "$dll" 2>&1 | head -2
  echo # blank line between entries
done
```

Output:

```
=== nsis_tauri_utils.dll ===
****************:error:********:PKCS7 routines:PKCS7_verify:certificate verify error:crypto/pkcs7/pk7_smime.c:297:Verify error: self-signed certificate in certificate chain
PE checksum   : 0000C602

=== nsDialogs.dll ===
No signature found
Unable to extract existing signature

=== NSISdl.dll ===
No signature found
Unable to extract existing signature

=== StartMenu.dll ===
No signature found
Unable to extract existing signature

=== System.dll ===
No signature found
Unable to extract existing signature
```

Only `nsis_tauri_utils.dll`, copied from `additional/`, has the signed copy embedded — the stock plugins are still the unsigned originals from the system NSIS toolset.

(NOTE: the `verify error: self-signed certificate in certificate chain` on `nsis_tauri_utils.dll` just means macOS's local CA store doesn't include the root CA — the signature itself is present.)

crates/tauri-bundler/src/bundle/windows/nsis/mod.rs
```diff
@@ -282,6 +282,13 @@ fn build_nsis_app_installer(
     to_json(&additional_plugins_path),
   );
 
+  if let Some(plugin_copy_path) = &maybe_plugin_copy_path {
+    data.insert(
+      "signed_plugins_path",
+      to_json(plugin_copy_path.join("x86-unicode")),
```
Legend-Master, line 288: This seems to give me an error here 🤔

> Plugin command System::Call conflicts with a plugin in another directory!

koki-develop: @Legend-Master Hmm - for what it's worth, in my own project the signing flow does work end-to-end with this fork (on `windows-latest` via GitHub Actions).

I also dug into the NSIS source, and as far as I can tell, the current PR shouldn't actually be able to hit the conflict path. Sharing my trace below.

---

The `Plugin command ... conflicts with a plugin in another directory!` error only fires when the DLL name is in makensis' internal `m_dllname_conflicts` set.

https://github.com/NSIS-Dev/nsis/blob/d20198c805a603b2e6359458152987f49f3637cd/Source/script.cpp#L5079-L5083
https://github.com/NSIS-Dev/nsis/blob/d20198c805a603b2e6359458152987f49f3637cd/Source/Plugins.cpp#L246

The only place that inserts into that set is `GetExports` in `Plugins.cpp`,

https://github.com/NSIS-Dev/nsis/blob/d20198c805a603b2e6359458152987f49f3637cd/Source/Plugins.cpp#L151

and its condition is `DllHasDataHandle(dllName) == true` - i.e. that DLL has already been assigned a data handle from a previous plugin command. The handle itself only gets set by `SetDllDataHandle`, which is only called from `TOK__PLUGINCOMMAND` processing:

https://github.com/NSIS-Dev/nsis/blob/d20198c805a603b2e6359458152987f49f3637cd/Source/script.cpp#L5117

So the necessary order for the conflict error to fire is **"a plugin call (e.g. `System::Call`) is processed first → then a `!addplugindir` for a directory containing the same DLL is parsed"**.

---

If I walk through the `installer.nsi` this PR generates and watch the order in which `!addplugindir` and `System::Call` actually appear:

https://github.com/koki-develop/tauri/blob/bea3d5f9aa08ab6887c6e9fbea5d57636757db1a/crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi#L16-L29

1. None of the `!include`s above the new `!addplugindir` emit anything on their own - the `System::Call`s inside them all live within `!macro` definitions, which are just stored as templates until they get expanded by `${MacroName}` or `!insertmacro`. The same applies to tauri's own `utils.nsh` / `FileAssociation.nsh`:

https://github.com/koki-develop/tauri/blob/bea3d5f9aa08ab6887c6e9fbea5d57636757db1a/crates/tauri-bundler/src/bundle/windows/nsis/utils.nsh#L22-L71

https://github.com/koki-develop/tauri/blob/bea3d5f9aa08ab6887c6e9fbea5d57636757db1a/crates/tauri-bundler/src/bundle/windows/nsis/FileAssociation.nsh#L131-L135

2. The new `!addplugindir` runs - at this point no DLL has a data handle yet, so nothing gets added to `m_dllname_conflicts`.

3. `${StrCase}` is expanded, and that's where the very first `System::Call` actually appears in the script (this is also when `System`'s data handle gets set):

https://github.com/NSIS-Dev/nsis/blob/d20198c805a603b2e6359458152987f49f3637cd/Include/StrFunc.nsh#L158-L201

4. After that, the only other `!addplugindir` in the script is the existing one at L94 (`${ADDITIONALPLUGINSPATH}`). That directory only contains `nsis_tauri_utils.dll`, and the only DLL with a data handle so far is `System` (not `nsis_tauri_utils`), so this one doesn't insert into `m_dllname_conflicts` either.

https://github.com/koki-develop/tauri/blob/bea3d5f9aa08ab6887c6e9fbea5d57636757db1a/crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi#L93-L94

https://github.com/koki-develop/tauri/blob/bea3d5f9aa08ab6887c6e9fbea5d57636757db1a/crates/tauri-bundler/src/bundle/windows/nsis/mod.rs#L46-L61

---

So as far as I can see, there's no point in the generated `installer.nsi` where the conflict condition would actually be satisfied.

If possible, could you share the few log lines right before `Plugin command System::Call conflicts ...` fires (ideally with the `.nsi:NNN` line reference)? That'd help pinpoint which `System::Call` is the one triggering it. Also - are you using a custom template by any chance?

Legend-Master: I think I had a git stash conflict that I didn't notice that pushed the `!addplugindir` in this PR a bit lower which gave me

```
Plugin command System::Call conflicts with a plugin in another directory!
Error in macro __NSD_FreeImage on macroline 2
Error in macro MUI_FUNCTION_WELCOMEPAGE on macroline 50
Error in macro MUI_PAGEDECLARATION_WELCOME on macroline 18
Error in macro MUI_PAGE_WELCOME on macroline 6
```

It's working now after resetting the workspace, sorry about that! Appreciate the effort digging this!

crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
```diff
@@ -22,6 +22,9 @@ ManifestDPIAwareness PerMonitorV2
 !include "Win\COM.nsh"
 !include "Win\Propkey.nsh"
 !include "StrFunc.nsh"
+{{#if signed_plugins_path}}
+!addplugindir "{{signed_plugins_path}}"
+{{/if}}
```
Legend-Master: Since I just hit https://github.com/tauri-apps/tauri/pull/15422#discussion_r3289231459, let's move this block to above all the `!include` calls, and add a little comment on this must be before any actual code

Legend-Master (dismissed):
Awesome work, thanks! Really appreciate the detailed answers!

Just a bit of nitpicks and we can merge this one

crates/tauri-bundler/src/bundle/windows/nsis/installer.nsi
koki-develop: @Legend-Master 
Done in 8cf1e53. 
Please feel free to adjust the wording if anything feels off!

Legend-Master (approved):
Thanks!

PR #15429, opened 2026-05-22
fix: check makensis exit code

Legend-Master:

### DIFF
--- a/crates/tauri-bundler/src/bundle/windows/nsis/mod.rs
+++ b/crates/tauri-bundler/src/bundle/windows/nsis/mod.rs
@@ -13,7 +13,7 @@ use crate::{
       },
     },
   },
-  error::ErrorExt,
+  error::{bail, ErrorExt},
   utils::{
     http_utils::{download_and_verify, verify_file_hash, HashAlgorithm},
     CommandExt,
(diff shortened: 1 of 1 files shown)


crates/tauri-bundler/src/bundle/windows/nsis/mod.rs
```diff
@@ -704,6 +704,9 @@ fn build_nsis_app_installer(
       command: "makensis.exe".to_string(),
       error,
     })?;
+  if !status.success() {
+    bail!("Failed to bundle app with makensis");
```
amrbashir, line 711: nit: use `anyhow::bail!`

amrbashir (dismissed):

crates/tauri-bundler/src/bundle/windows/nsis/mod.rs
Legend-Master: Isn't the plan to get rid of anyhow so it forces us to add contexts to errors?

amrbashir: nvm, my bad I thought the import was from anyhow

amrbashir (approved):

PR #15450, opened 2026-05-30
fix: mock runtime run on main thread dead locks

Legend-Master:
Fix `run_main_thread!` macro used by things like `MenuItem` and some other APIs deadlock when called on main thread


FabianLars (dismissed):
untested

FabianLars (dismissed):

FabianLars (approved):
further approvals cost extra

PR #15454, opened 2026-05-30
Fix aliased plugin imports in v1 migration

fallintoplace:
## Summary
- preserve the local alias when migrating pluginified imports from `@tauri-apps/api`
- update the v1 migration fixtures to expect valid namespace imports
- reparse migrated JS, Svelte, and Vue output in tests so invalid syntax is caught

## Root cause
The migrator generated aliased namespace imports as `import * as cli as superCli from "..."`, which is not valid ESM and breaks migrated code for inputs like `import { cli as superCli } from "@tauri-apps/api"`.

## Testing
- `cargo test -p tauri-cli 'migrate::migrations::v1::frontend::tests::migrates'`

### DIFF
--- a/.changes/v1-migrate-aliased-plugin-imports.md
+++ b/.changes/v1-migrate-aliased-plugin-imports.md
@@ -1,0 +1,7 @@
+---
+"tauri-cli": "patch:bug"
+---
+
+Fix `tauri migrate` generating invalid namespace imports for aliased pluginified imports from `@tauri-apps/api`.
+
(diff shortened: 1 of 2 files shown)


Legend-Master (dismissed):

.changes/v1-migrate-aliased-plugin-imports.md
```diff
@@ -0,0 +1,7 @@
+---
+"tauri-cli": "patch:bug"
```
Legend-Master, line 2:
```suggestion
"tauri-cli": "patch:bug"
"@tauri-apps/cli": "patch:bug"
```

Legend-Master (approved):
Thanks!

PR #15464, opened 2026-06-03
feat(cef): add with_webview support

amrbashir:


crates/tauri/src/webview/mod.rs
```diff
@@ -1718,10 +1778,15 @@ impl<R: Runtime> Webview<R> {
   /// Note that `webview2-com`, `webkit2gtk`, `objc2_web_kit` and similar crates may be updated in minor releases of Tauri.
   /// Therefore it's recommended to pin Tauri to at least a minor version when you're using `with_webview`.
   ///
+  #[cfg_attr(
```
FabianLars, line 1667: The comment above should probably mention the cef crate like you did in webview_window.rs

crates/tauri-runtime-cef/src/lib.rs
```diff
@@ -52,6 +52,9 @@ mod cef_impl;
 mod cef_webview;
 mod utils;
 
+/// The `cef` crate used by this runtime, re-exported for convenience.
+pub use cef;
```
FabianLars, line 61: Probably want to mention that this will be updated across breaking/major boundaries all the freaking time

lucasfernog (approved):

PR #15465, opened 2026-06-03
feat(cef): add WebviewEvent::Focus

amrbashir:


lucasfernog (approved):
could we emit Focused(false) to the *other* browsers when on_got_focused is called? (and maybe also emit it when the parent window loses focus)
or is that too much? :D

lucasfernog (approved):

Issue #8920, opened 2024-02-20
`tauri dev` is incredibly slow


seaglade:
### Describe the bug

When running `tauri dev` the app takes *minutes* to launch, and a single change to a file forces it to start over, seemingly from scratch. This happens when running `yarn tauri dev` in the unchanged result of `yarn create tauri-app`. Around 300 rust libraries seem to be getting compiled every time a change is made.

### Reproduction

```sh
yarn create tauri-app # Fill out the form and let it create the project
cd <app-directory>
yarn
yarn tauri dev
```

### Expected behavior

Interface loads quickly and refreshes quickly when files are changed.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.22631 X64
    ✔ WebView2: 121.0.2277.128
    ✔ MSVC: Visual Studio Community 2022
    ✔ rustc: 1.74.1 (a28077b28 2023-12-04)
    ✔ cargo: 1.74.1 (ecb9851af 2023-10-18)
    ✔ rustup: 1.26.0 (5af9b9484 2023-04-05)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (default)
    - node: 21.6.1
    - pnpm: 8.15.3
    - yarn: 1.22.19
    - npm: 10.2.4

[-] Packages
    - tauri [RUST]: 1.6.0
    - tauri-build [RUST]: 1.5.1
    - wry [RUST]: 0.24.7
    - tao [RUST]: 0.16.7
    - @tauri-apps/api [NPM]: 1.5.3
    - @tauri-apps/cli [NPM]: 1.5.10

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../dist
    - devPath: http://localhost:1420/
    - framework: React
    - bundler: Vite
```

seaglade:
Huh, if the change is made to the UI the interface reloads instantly. It's only backend changes that take so long. That makes a bit more sense but is still problematic.

seaglade:
Running `tauri build` and then `tauri dev` seems to have fixed the issue. I'm guessing the `dev` command does something differently that causes it to rebuild everything every single time. Regardless, I hope that this can be documented somewhere and/or fixed.

amrbashir:
This is something related to Rust and its incremental compilation, it shouldn't recompile your dependencies unless necessary. Anyways, there is nothing we can do since this seems like a Rust issue.

kon72:
[...]

PR #15467, opened 2026-06-03
Fixed vite crawler overreaching into cache, and hanging tauri dev

yackermann:
Tauri was generating embedded asset cache files inside `src-tauri/target/**/tauri-codegen-assets/` while preserving original file extension.

This issue affecting frontend serving using `npm run dev` and `npm run tauri dev`

Example: NOT A REAL JS/css/html file, but a binary blob

```text
project/src-tauri/target/release/build/project-91fc920d1289165b/out/tauri-codegen-assets/2de3b7aa36eeec0abc6d73f5fdeeab51e2f99865bbadeab43d14e3be0c1c55e3.js/css/html
```

Vite crawl ended up crawiling them because it though they were real web files, and ended up handing over invalid input

I propose removing extension preservation as it confusing static crawlers, and because those compiled files are not in fact contain any of the extension defined data, but are data blobs.

P.S. The actual asset key is still the same, so Tauri still serves `index.html` normally. Only the internal cache filename changed.

## Before

Cold request timing:

```sh
✗ curl -o /dev/null -s -w '/src/main.tsx -> %{time_total}s\n' \
  http://127.0.0.1:8089/src/main.tsx
/src/main.tsx -> 51.285s
```

## After

```
✗ curl -o /dev/null -s -w '/src/main.tsx -> %{time_total}s\n' \
  http://127.0.0.1:8089/src/main.tsx
/src/main.tsx -> 0.00442s
```

### DIFF
--- a/crates/tauri-codegen/src/embedded_assets.rs
+++ b/crates/tauri-codegen/src/embedded_assets.rs
@@ -344,12 +344,9 @@ impl EmbeddedAssets {
     // get a hash of the input - allows for caching existing files
     let hash = crate::checksum(&input).map_err(EmbeddedAssetsError::Hex)?;
 
-    // use the content hash to determine filename, keep extensions that exist
-    let out_path = if let Some(ext) = path.extension().and_then(|e| e.to_str()) {
-      out_dir.join(format!("{hash}.{ext}"))
-    } else {
-      out_dir.join(hash)
-    };
+    // use only the content hash to determine filename; this prevents frontend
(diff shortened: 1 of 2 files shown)


yackermann:
Ref: #8920

Legend-Master:
I have never had vite taking so long to load a file, do you have a reproducible example for this?

yackermann:
@Legend-Master yes sir. Let me upload it on github.

PR #14629, opened 2025-12-08
fix(cli): Android build --apk and --aab flags requiring a value

lucasfernog:
regression from #14120

PR #14120, opened 2025-08-29
feat(cli): add mobile run commands, closes #13196

lucasfernog:
similar to the dev command, but runs using the production bundle of the app

it is essentially a wrapper to the build command, but now it will

currently can be achieved with `[android|ios] build --open` and running from the IDE.

Edit by @FabianLars
fixes #13196

PR #15470, opened 2026-06-03
fix(cli): skip AAB build during `tauri android run`

### LINKED ISSUE #15419: bug: `tauri android run` builds both APK and AAB, causing ~50% slower deploy
## Describe the bug

`tauri android run` builds **both APK and AAB** even though it only needs the APK for installation. This causes 3 separate Gradle invocations instead of the necessary 1-2, resulting in ~50% slower deploy times.

## Steps to reproduce

1. Run `tauri android run` with a connected device/emulator
2. Observe the build output

## Expected behavior

`android run` should only build the APK needed for installation, skipping the AAB bundle entirely.

## Actual behavior

The command triggers 3 Gradle invocations:

1. **APK build** (`assembleUniversalDebug`) → Rust compiles (~16s)
2. **AAB build** (`bundleUniversalDebug`) → Rust compiles **again** (~15s) ← unnecessary
3. **Install** (`installUniversalDebug`) → Rust cached (0.5s)

## Root cause

In [`crates/tauri-cli/src/mobile/android/run.rs`](https://github.com/tauri-apps/tauri/blob/dev/crates/tauri-cli/src/mobile/android/run.rs), the `run` function calls `super::build::run()` with:

```rust
super::build::Options {
    // ...
    split_per_abi: true,
    apk: false,
    aab: false,
    skip_bundle: false,  // ← this causes both APK + AAB to be built
    // ...
}
```

Then in [`build.rs` lines 252-256](https://github.com/tauri-apps/tauri/blob/dev/crates/tauri-cli/src/mobile/android/build.rs#L252-L256):

```rust
if !(options.skip_bundle || options.apk || options.aab) {
    // if the user didn't specify the format to build, we'll do both
    options.apk = true;
    options.aab = true;
}
```

Since `skip_bundle` is `false` and both `apk`/`aab` are `false`, the condition triggers and **both** formats are built.

## Proposed fix

Set `skip_bundle: true` in `run.rs` to skip AAB generation during `android run`:

```rust
super::build::Options {
    // ...
    skip_bundle: true,  // ← skip AAB, we only need APK for install
    // ...
}
```

Or alternatively, set `apk: true` explicitly to avoid the "build both" fallback.

## Impact

- **Time savings**: ~50% reduction in `android run` time (eliminates one full Rust compilation + Gradle invocation)
- **Disk I/O**: No unnecessary AAB file generation
- **Developer experience**: Faster iteration cycle for mobile development

## Environment

- Tauri CLI: 2.11.2
- Platform: Windows 11
- Target: Android (x86_64 emulator)
- NDK: 27.2.12479018

## Build log excerpt

```
> tauri android run
   Compiling suit v0.1.0 (16.38s)  ← APK build
   Compiling suit v0.1.0 (15.38s)  ← AAB build (unnecessary)
   Compiling suit v0.1.0 (0.47s)   ← Install (cached)
```

Adit-Jain-srm:
## Summary

Fixes #15419 — \	auri android run\ builds both APK and AAB even though only the APK is needed for device installation, causing ~50% slower deploys.

## Problem

In \
un.rs\, the build options are passed with \pk: false, aab: false, skip_bundle: false\. This triggers the fallback in \uild.rs:253\:

\\\
ust
if !(options.skip_bundle || options.apk || options.aab) {
    options.apk = true;
    options.aab = true;  // <- unnecessary for \
un\
}
\\\

Result: 3 Gradle invocations (APK build + AAB build + install) instead of 2 (APK build + install).

## Fix

Set \pk: true\ explicitly in \
un.rs\ so the fallback doesn't trigger. Only the APK is built, and \ab\ remains \alse\.

## Impact

- Eliminates one full Rust compilation + Gradle invocation during \ndroid run\
- ~50% reduction in deploy time (as measured in the issue)
- No behavior change for \	auri android build\ (which still builds both by default)

### DIFF
--- a/crates/tauri-cli/src/mobile/android/run.rs
+++ b/crates/tauri-cli/src/mobile/android/run.rs
@@ -101,7 +101,7 @@ pub fn command(options: Options, noise_level: NoiseLevel) -> Result<()> {
       features: options.features,
       config: options.config.clone(),
       split_per_abi: true,
-      apk: false,
+      apk: true,
       aab: false,
       skip_bundle: false,
       open: options.open,


Adit-Jain-srm:
You're right to close this — my fix was shallow. The correct approach is \skip_bundle: true\ (restoring #14120 behavior that #14629 regressed), not \pk: true\. Apologies for the noise.

PR #15481, opened 2026-06-05
fix(acl): preserve Number::Int on serde_json round-trip

### LINKED ISSUE #15480: [bug] `Number::Int` is silently coerced to `Number::Float` on serde_json round-trip
### Describe the bug

**File:** `crates/tauri-utils/src/acl/value.rs`  
**Severity:** High — data corruption, silent loss of type fidelity  
**Found by:** Property-based testing (`value_to_serdejson_and_back`, `serdejson_to_value_and_back`, `list_structure_preserved`)

---

## Summary

Any `acl::Value::Number(Number::Int(i))` value is silently converted to
`Number::Float(i as f64)` after a round-trip through `serde_json::Value`.
The integer/float distinction is permanently lost. This affects serialization,
deserialization, scope evaluation, and any code that pattern-matches on the
`Number` variant.

---

## Minimal Reproduction

```rust
use tauri_utils::acl::{Number, Value};

let original = Value::Number(Number::Int(0));

let json: serde_json::Value = original.clone().into();
let back: Value = json.into();

// FAILS: left=Number(Int(0)), right=Number(Float(0.0))
assert_eq!(original, back);
```

Proptest shrunk counterexample:
```
input:    Map({"": Number(Int(0))})
expected: Map({"": Number(Int(0))})
actual:   Map({"": Number(Float(0.0))})
```

---

## Root Cause

`From<serde_json::Value> for Value` (line 88–96 of `value.rs`) checks `as_f64()`
before `as_i64()`:

```rust
// Current — broken order
serde_json::Value::Number(n) => Value::Number(if let Some(f) = n.as_f64() {
    Number::Float(f)          // ← always wins for any integer that fits in f64
} else if let Some(n) = n.as_u64() {
    Number::Int(n as i64)
} else if let Some(n) = n.as_i64() {
    Number::Int(n)
} else {
    Number::Int(0)
}),
```

`serde_json::Number::as_f64()` succeeds for every integer in the range
`[-(2^53), 2^53]` (and many outside it), so the `as_i64()` branch is
never reached for typical integer values. Every integer that arrives from
JSON as `Number(0)`, `Number(42)`, etc. is converted to `Float`.

---

## Impact

- **ACL scope evaluation**: Scopes that compare integer values (e.g.
  `{ "allow": [{ "port": 8080 }] }`) may fail to match because the parsed
  value is `Float(8080.0)` instead of `Int(8080)`.
- **TOML → JSON round-trip**: Values written as TOML integers (`port = 8080`)
  are correctly parsed as `Number::Int(8080)` via `From<toml::Value>`, but
  lose that distinction the moment they pass through serde_json.
- **Debugging**: `Debug` output prints `Float(0.0)` where a developer expects
  `Int(0)`, making logs and error messages misleading.

---

## Proposed Fix

Reverse the check order — prefer integer interpretation first:

```rust
// Fixed order
serde_json::Value::Number(n) => Value::Number(if let Some(i) = n.as_i64() {
    Number::Int(i)
} else if let Some(u) = n.as_u64() {
    // WARNING: `u as i64` wraps for values > i64::MAX, producing large negative
    // numbers (e.g. i64::MAX + 1 → i64::MIN). This is a behavioural change from
    // the current code, which converts those values to Float instead. Consider
    // saturating to i64::MAX, returning an error, or keeping them as
    // Number::Float(u as f64) to at least preserve approximate magnitude.
    Number::Int(u as i64)
} else if let Some(f) = n.as_f64() {
    Number::Float(f)
} else {
    Number::Int(0)
}),
```

This matches serde_json's own convention: its `Visitor` implementations
prefer `visit_i64`/`visit_u64` over `visit_f64` when the number has no
decimal component.

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
[✘] Environment
       ✘ Xcode: not installed!
       - node: 25.8.2
       - pnpm: 10.30.3
       - npm: 11.11.1
       - bun: 1.3.13

   [-] Packages
       - tauri 🦀: 2.11.2
       - tauri-build 🦀: 2.6.2
       - wry 🦀: 0.55.0, (outdated, latest: 0.55.1)
       - tao 🦀: 0.35.0, (outdated, latest: 0.35.3)
       - @tauri-apps/cli  ⱼₛ: 2.11.2

   [-] Plugins
       - tauri-plugin-log 🦀: 2.6.0, (outdated, latest: 2.8.0)
       - @tauri-apps/plugin-log  ⱼₛ: not installed!

   [-] App
```

### Stack trace

```text

```

### Additional context

_No response_

thanhtoantnt:
closes #15480

 What changed

 From<serde_json::Value> for Value in crates/tauri-utils/src/acl/value.rs was checking as_f64() before as_i64(). Because serde_json::Number::as_f64() succeeds for any integer that
 fits in an f64, every integer JSON number was silently deserialized as Number::Float — the as_i64() branch was never reached for typical values.

 The check order is now:
 1. as_i64() → Number::Int(i) — handles all signed integers
 2. as_u64() → Number::Float(u as f64) — u64 values above i64::MAX can't be represented without wrapping, so Float preserves approximate magnitude
 3. as_f64() → Number::Float(f) — genuine floating-point numbers
 4. fallback → Number::Int(0)

 This matches serde_json's own visitor convention of preferring integer representations when the number has no decimal component.

 Impact

 - ACL scopes that compare integer values (e.g. { "port": 8080 }) now correctly match Number::Int instead of silently failing against Number::Float(8080.0).
 - TOML integers parsed as Number::Int via From<toml::Value> no longer lose that distinction when passed through serde_json.

### DIFF
--- a/crates/tauri-utils/src/acl/value.rs
+++ b/crates/tauri-utils/src/acl/value.rs
@@ -86,12 +86,14 @@ impl From<serde_json::Value> for Value {
     match value {
       serde_json::Value::Null => Value::Null,
       serde_json::Value::Bool(b) => Value::Bool(b),
-      serde_json::Value::Number(n) => Value::Number(if let Some(f) = n.as_f64() {
+      serde_json::Value::Number(n) => Value::Number(if let Some(i) = n.as_i64() {
+        Number::Int(i)
+      } else if let Some(u) = n.as_u64() {
+        // u64 values that exceed i64::MAX cannot be represented as Int without
+        // wrapping; fall back to Float to preserve approximate magnitude.
+        Number::Float(u as f64)
+      } else if let Some(f) = n.as_f64() {
         Number::Float(f)
-      } else if let Some(n) = n.as_u64() {
-        Number::Int(n as i64)
-      } else if let Some(n) = n.as_i64() {
-        Number::Int(n)
       } else {
         Number::Int(0)
       }),
--- a/crates/tauri-utils/src/acl/value.rs
+++ b/crates/tauri-utils/src/acl/value.rs
@@ -89,9 +89,7 @@ impl From<serde_json::Value> for Value {
       serde_json::Value::Number(n) => Value::Number(if let Some(i) = n.as_i64() {
         Number::Int(i)
       } else if let Some(u) = n.as_u64() {
-        // u64 values that exceed i64::MAX cannot be represented as Int without
-        // wrapping; fall back to Float to preserve approximate magnitude.
-        Number::Float(u as f64)
+        Number::Int(u as i64)
       } else if let Some(f) = n.as_f64() {
         Number::Float(f)
       } else {
--- a/.changes/acl-number-int-preserved-on-serdejson-roundtrip.md
+++ b/.changes/acl-number-int-preserved-on-serdejson-roundtrip.md
@@ -4,4 +4,4 @@
 
 Fix `Number::Int` being silently coerced to `Number::Float` on `serde_json` round-trip.
 
-`From<serde_json::Value> for Value` was checking `as_f64()` first, which succeeds for every integer that fits in an f64, so integer JSON numbers were always deserialized as `Number::Float`. The check order is now `as_i64()` → `as_u64()` (falling back to `Float` for values above `i64::MAX` to avoid silent wrapping) → `as_f64()`, matching serde_json's own visitor convention.
+`From<serde_json::Value> for Value` was checking `as_f64()` first, which succeeds for every integer that fits in an f64, so integer JSON numbers were always deserialized as `Number::Float`. The check order is now `as_i64()` → `as_u64()` (cast to `i64`, wrapping for values above `i64::MAX`) → `as_f64()`, matching serde_json's own visitor convention.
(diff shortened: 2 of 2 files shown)


crates/tauri-utils/src/acl/value.rs
```diff
@@ -86,12 +86,14 @@ impl From<serde_json::Value> for Value {
     match value {
       serde_json::Value::Null => Value::Null,
       serde_json::Value::Bool(b) => Value::Bool(b),
-      serde_json::Value::Number(n) => Value::Number(if let Some(f) = n.as_f64() {
+      serde_json::Value::Number(n) => Value::Number(if let Some(i) = n.as_i64() {
```
Legend-Master, line 89: @lucasfernog ~~do you still remember where is this `From<serde_json:Value> for Value` impl is used? I can't seem to find any and this conversion implementation depends heavily on how it was used~~ never mind, it's used in migration and capability builder

I think should look into just use `serde_json:Value` here in v3

crates/tauri-utils/src/acl/value.rs
```diff
@@ -86,12 +86,14 @@ impl From<serde_json::Value> for Value {
...
+        Number::Int(i)
+      } else if let Some(u) = n.as_u64() {
+        // u64 values that exceed i64::MAX cannot be represented as Int without
```
Legend-Master: Let's keep the i64 cast for now, this is capped vs precision lost + change of type

Legend-Master (approved):
thanks

PR #15482, opened 2026-06-05
fix(acl): map NaN/\u00b1Infinity to JSON null instead of panicking

### LINKED ISSUE #15477: [bug] Panic on `inf`/`-inf`/`NaN` floats when converting `acl::Value` to `serde_json::Value`
## Summary

`From<Value> for serde_json::Value` in `crates/tauri-utils/src/acl/value.rs` calls `serde_json::Number::from_f64(f).unwrap()` without guarding against non-finite float values. `from_f64` returns `None` for `inf`, `-inf`, and `NaN` (which JSON cannot represent), so `.unwrap()` panics. The panic is unrecoverable.

This is reachable through normal TOML config parsing — TOML 1.0 supports `inf`, `-inf`, and `nan` as valid float literals, and the `From<toml::Value> for Value` impl passes them through without validation.

## Environment

- Crate: `tauri-utils`
- File: `crates/tauri-utils/src/acl/value.rs`
- Affects: any app using a `Tauri.toml` capability file with a float-typed scope value

## Steps to Reproduce

**Via TOML config (normal usage):**

```toml
# Tauri.toml or capability file
[scope]
some_value = inf
```

This walks the path `Tauri.toml → toml::Value → acl::Value → serde_json::Value → panic` through ordinary config parsing.

**Programmatically:**

```rust
use tauri_utils::acl::value::{Value, Number};

let v = Value::Number(Number::Float(f64::INFINITY));
let _: serde_json::Value = v.into();
// thread panicked: called `Option::unwrap()` on a `None` value
// at crates/tauri-utils/src/acl/value.rs:69
```

## Root Cause

```rust
// crates/tauri-utils/src/acl/value.rs
impl From<Value> for serde_json::Value {
  fn from(value: Value) -> Self {
    match value {
      Value::Number(Number::Float(f)) => {
        serde_json::Value::Number(serde_json::Number::from_f64(f).unwrap())
        //                                                         ^^^^^^^^
        //                                   panics when f is inf, -inf, or NaN
      }
      // ...
    }
  }
}
```

The TOML entry point has no guard either:

```rust
// ~line 139
Toml::Float(f) => f.into(),  // passes inf/nan directly into Number::Float
```

## Expected Behavior

Non-finite float values should be handled gracefully — either mapped to a fallback, rejected early with a clear error, or excluded at the point of TOML parsing.

## Actual Behavior

The process panics with `called 'Option::unwrap()' on a 'None' value` at `value.rs:69`. The panic is unrecoverable from a `From` impl.

## Proposed Fix

Three options in order of increasing scope:

**Option A — map to null at serialization time (lowest friction):**
```rust
Value::Number(Number::Float(f)) => {
    serde_json::Number::from_f64(f)
        .map(serde_json::Value::Number)
        .unwrap_or(serde_json::Value::Null)
}
```

**Option B — reject at the TOML entry point:**
```rust
Toml::Float(f) if f.is_finite() => f.into(),
Toml::Float(_) => Value::Null,  // or return an error
```

**Option C — reject at construction time** in `Number::Float`, so non-finite values can never be stored. This is the most defensive approach but is a broader API change.

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
[✘] Environment
       ✘ Xcode: not installed!
       - node: 25.8.2
       - pnpm: 10.30.3
       - npm: 11.11.1
       - bun: 1.3.13

   [-] Packages
       - tauri 🦀: 2.11.2
       - tauri-build 🦀: 2.6.2
       - wry 🦀: 0.55.0, (outdated, latest: 0.55.1)
       - tao 🦀: 0.35.0, (outdated, latest: 0.35.3)
       - @tauri-apps/cli  ⱼₛ: 2.11.2

   [-] Plugins
       - tauri-plugin-log 🦀: 2.6.0, (outdated, latest: 2.8.0)
       - @tauri-apps/plugin-log  ⱼₛ: not installed!

   [-] App
```

### Stack trace

```text

```

### Additional context

_No response_

thanhtoantnt:
closes #15477

 What changed

 From<Value> for serde_json::Value in crates/tauri-utils/src/acl/value.rs was calling serde_json::Number::from_f64(f).unwrap(). The JSON spec (RFC 8259 §6) does not permit non-finite
 floats, so from_f64 returns None for NaN, +Infinity, and -Infinity — and the .unwrap() turns that into an unconditional panic.

 ```rust
   // before — panics for non-finite f64
   Value::Number(Number::Float(f)) => {
       serde_json::Value::Number(serde_json::Number::from_f64(f).unwrap())
   }

   // after — non-finite maps to null
   Value::Number(Number::Float(f)) => {
       match serde_json::Number::from_f64(f) {
           Some(n) => serde_json::Value::Number(n),
           None => serde_json::Value::Null, // NaN / ±Infinity are not valid JSON
       }
   }
 ```

 Number::Float accepts any f64 via From<f64>, so non-finite values are constructible without error and the panic was the first signal that anything was wrong — potentially surfacing
 at an unrelated call site well after the bad value was created.

 Non-finite floats now produce serde_json::Value::Null, consistent with the IEEE 754 null-substitution convention used by many JSON serializers.

### DIFF
--- a/crates/tauri-utils/src/acl/value.rs
+++ b/crates/tauri-utils/src/acl/value.rs
@@ -66,7 +66,10 @@ impl From<Value> for serde_json::Value {
       Value::Null => serde_json::Value::Null,
       Value::Bool(b) => serde_json::Value::Bool(b),
       Value::Number(Number::Float(f)) => {
-        serde_json::Value::Number(serde_json::Number::from_f64(f).unwrap())
+        match serde_json::Number::from_f64(f) {
+          Some(n) => serde_json::Value::Number(n),
+          None => serde_json::Value::Null, // NaN / ±Infinity are not valid JSON
+        }
       }
       Value::Number(Number::Int(i)) => serde_json::Value::Number(i.into()),
       Value::String(s) => serde_json::Value::String(s),
--- a/.changes/acl-nonfinite-float-null-instead-of-panic.md
+++ b/.changes/acl-nonfinite-float-null-instead-of-panic.md
@@ -1,0 +1,7 @@
+---
+"tauri-utils": "patch:bug"
+---
+
+Fix panic when converting `Number::Float(NaN)` or `Number::Float(±Infinity)` to `serde_json::Value`.
+
+`From<Value> for serde_json::Value` was calling `serde_json::Number::from_f64(f).unwrap()`, which panics for non-finite floats because the JSON specification (RFC 8259) does not permit them. Non-finite values now map to `serde_json::Value::Null` instead of panicking.


crates/tauri-utils/src/acl/value.rs
```diff
@@ -66,7 +66,10 @@ impl From<Value> for serde_json::Value {
       Value::Null => serde_json::Value::Null,
       Value::Bool(b) => serde_json::Value::Bool(b),
       Value::Number(Number::Float(f)) => {
-        serde_json::Value::Number(serde_json::Number::from_f64(f).unwrap())
+        match serde_json::Number::from_f64(f) {
```
Legend-Master, line 69: As much as this is a problem, from how we use this today, a crash might be better than silently making it `null` here.

PR #15491, opened 2026-06-06
fix(mobile): avoid mutex deadlocks

bclarke123:
This PR removes potential deadlocks when multiple threads call IPC commands concurrently on iOS / Android - in Tauri 2.9 there are four surfaces where this issue can manifest, and in 2.11 one of them has been fixed (ios `run_command`). This PR fixes the other three to acquire locks before entering `if let Some(handler) = ...` to avoid deadlocks on device, which are hard to trace and give the impression of instability.

One note, these changes introduce two `.cloned()` calls on `Channel` structs - these are Arc-backed, and so these clones should be cheap, and are required to release the locks on CHANNELS / PENDING_PLUGIN_CALLS before sending to the WebView.


Legend-Master (dismissed):

bclarke123:
Will do!

Legend-Master (approved):
Thanks!

PR #15495, opened 2026-06-07
 fix(bundler): use relative symlink for .DirIcon in AppImage

### LINKED ISSUE #15488: [bug] AppImage .DirIcon is generated as an absolute symlink to the build machine path
### Describe the bug

Tauri-generated AppImages can contain a broken `.DirIcon` symlink after being moved away from the build machine.

When building an AppImage, the generated AppDir contains the application PNG next to `.DirIcon`, but `.DirIcon` is created as an absolute symlink pointing to the build machine's AppDir path. For example, after extracting the AppImage with `--appimage-extract`, `.DirIcon` points to a path under the build machine workspace instead of to the icon file inside the extracted AppDir.

This makes the final AppImage non-relocatable from an icon metadata perspective and can cause Linux desktops/file managers to show a generic icon.

### Expected behavior

`.DirIcon` should not point to an absolute build-machine path.

It should either be a PNG file copy, or a relative symlink to the root app icon, for example:
```text
.DirIcon -> Jumanji.png
```

Or to an AppDir-local hicolor path. The AppImage/AppDir spec says the root app icon should live in the AppDir and .DirIcon is commonly a symlink to that file. The icon filename should match the desktop Icon=
value.

### Actual behavior

The generated AppImage contains something like:
```text
.DirIcon -> /builds/<group>/<project>/src-tauri/target/release/bundle/appimage/Jumanji.AppDir/Jumanji.png
```

After moving the AppImage to another machine and extracting it, that path does not exist, so .DirIcon is broken. The PNG icon file itself is present in the extracted AppDir root, next to .DirIcon.

### Relevant source

The published tauri-bundler source appears to create the symlink using an absolute AppDir path:

```rs
std::os::unix::fs::symlink(
  app_dir_path.join(format!("{product_name}.png")),
  app_dir_path.join(".DirIcon"),
)?;
```

That produces an absolute symlink target because app_dir_path is absolute. A safer implementation would copy the PNG to .DirIcon, or create a relative symlink target:

```rs
std::os::unix::fs::symlink(
  format!("{product_name}.png"),
  app_dir_path.join(".DirIcon"),
)?;
```

### Why this matters
AppImages are expected to be relocatable. Embedding a build-local absolute path in .DirIcon breaks that expectation and causes icon metadata to fail once the AppImage is distributed.

### Environment
- Tauri CLI: 2.11.1
- tauri-bundler: 2.9.1 / 2.9.2 appears to contain the same code
- Target: Linux x86_64 AppImage
- Build environment: GitLab CI Docker runner
- AppImage extracted with: ./App.AppImage --appimage-extract

### References
- AppImage AppDir spec: https://docs.appimage.org/reference/appdir.html

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
[✘] Environment
    - OS: Mac OS 26.4.0 arm64 (X64)
    ✘ Xcode: not installed!
    ✔ rustup: 1.29.0 (2026-03-05)
    - node: 24.14.1
    - pnpm: 11.1.1
    - yarn: 1.22.22
    - npm: 11.11.0

[-] Packages
    - tauri 🦀: 2.11.1, (outdated, latest: 2.11.2)
    - tauri-build 🦀: 2.6.1, (outdated, latest: 2.6.2)
    - wry 🦀: 0.55.1
    - tao 🦀: 0.35.2, (outdated, latest: 0.35.3)
    - @tauri-apps/api  ⱼₛ: 2.11.0
    - @tauri-apps/cli  ⱼₛ: 2.11.1 (outdated, latest: 2.11.2)

[-] Plugins

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist-ui
    - devUrl: http://127.0.0.1:1420/
    - framework: React
    - bundler: Vite
```

### Stack trace

```text

```

### Additional context

_No response_

aitryhard:
The .DirIcon symlink was pointing to an absolute path on the build
machine, which breaks the AppImage when moved to another system.

Changed to a relative symlink so the icon resolves correctly
regardless of where the AppImage is extracted.

Fixes #15488

### DIFF
--- a/crates/tauri-bundler/src/bundle/linux/appimage/linuxdeploy.rs
+++ b/crates/tauri-bundler/src/bundle/linux/appimage/linuxdeploy.rs
@@ -176,7 +176,7 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
     app_dir_path.join(".DirIcon"),
   )?;
   std::os::unix::fs::symlink(
-    app_dir_path.join(format!("usr/share/applications/{product_name}.desktop")),
+    format!("usr/share/applications/{product_name}.desktop"),
     app_dir_path.join(format!("{product_name}.desktop")),
   )?;
 
--- a/crates/tauri-bundler/src/bundle/linux/appimage/linuxdeploy.rs
+++ b/crates/tauri-bundler/src/bundle/linux/appimage/linuxdeploy.rs
@@ -172,7 +172,7 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<PathBuf>> {
     app_dir_path.join(format!("{product_name}.png")),
   )?;
   std::os::unix::fs::symlink(
-    app_dir_path.join(format!("{product_name}.png")),
+    format!("{product_name}.png"),
     app_dir_path.join(".DirIcon"),
   )?;
   std::os::unix::fs::symlink(
--- a/crates/tauri/Cargo.toml
+++ b/crates/tauri/Cargo.toml
@@ -129,12 +129,12 @@ objc2-web-kit = { version = "0.3", default-features = false, features = [
   "WKWebViewConfiguration",
   "WKUserContentController",
 ] }
-window-vibrancy = "0.6"
+window-vibrancy = "0.7"
 
 # windows
 [target."cfg(windows)".dependencies]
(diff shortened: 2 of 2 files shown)


aitryhard:
Closing this PR. I need to understand the issue better before submitting a fix

PR #15501, opened 2026-06-07
Fix: avoid lock contention in scope/fs.rs

bclarke123:
### Summary

In `tauri::fs::Scope` it's currently possible to deadlock during event dispatch, when a listener re-enters the scope. This PR implements the existing pattern from `tauri::event::listener` to add a Vec of Pending enums when the lock can't be immediately acquired, which is flushed on `emit`.

### Repro

```rust
let scope = /* fs::Scope */;
scope.once(|_| {});
scope.allow_file("/some/path").unwrap(); // -> Scope::emit -> hangs forever
```

allow_file / allow_directory / forbid_file / forbid_directory all call emit, so any of them reproduces it once a once listener (or a re-entrant listener) is registered.

### Tests

This PR includes 4 new unit tests, to ensure the bug is fixed, and existing behaviour is preserved.

`once_listener_does_not_deadlock` - red / green, this test should fail on current `dev` branch and succeed on this branch (based on repro above)
`once_fires_exactly_once` - ensures deferred self-unlisten actually does
`handler_can_listen_without_deadlock` - ensure a handler registering a listener mid-dispatch doesn't deadlock
`handler_can_emit_without_deadlock` - ensure a handler can re-emit mid-dispatch without deadlock

### DIFF
--- a/.changes/scope-fs-event-deadlock.md
+++ b/.changes/scope-fs-event-deadlock.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch:bug
+---
+
+Implement pending queue pattern from `event/listener.rs` in `scope/fs.rs` to avoid lock contention
--- a/crates/tauri/src/scope/fs.rs
+++ b/crates/tauri/src/scope/fs.rs
@@ -29,12 +29,24 @@ pub enum Event {
 
 type EventListener = Box<dyn Fn(&Event) + Send>;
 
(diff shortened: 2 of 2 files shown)


Legend-Master:
This already implemented in https://github.com/tauri-apps/tauri/pull/15474
Thanks anyways!

PR #15508, opened 2026-06-09
fix: wait for window creation no matter if on main thread

### LINKED ISSUE #10157: [feat] Catch tauri errors
### Describe the problem

Some users have an old version of webview2. if it's too old, the internal function create_webview will fail to run, tauri will log the error, but the app will still run in the background without show any error or crash the thread.

Let me catch such errors. currently I don't get anything from it. I even have a panic hook, but it doesn't kick in since seems like the event loop just keep runing.

### Alternatives considered

-

### Additional context

https://github.com/thewh1teagle/vibe/issues/153#issuecomment-2198800744

### LINKED ISSUE #13046: [bug] encountering an error at the line:  let hwnd = window.hwnd().unwrap();
### Describe the bug

When attempting to obtain the hwnd of a child window created via new WebviewWindow(), an error occurs. According to the information obtained from println!, the webview_window object is indeed the created child window object, but an error happens when trying to get its hwnd. Previously, I directly passed window: Window as a method parameter and was able to obtain the window hwnd this way. However, doing so always retrieves the main window's window object, even if a child window object is explicitly passed before the call—it doesn't work as intended. I need the hwnd of the child window created in JavaScript code. Currently, in the Rust part, I can find the correct child window object by its label, but crucially, getting its hwnd results in an error. Does anyone have any solutions for this? Thanks.

Image

Image

### Reproduction

_No response_

### Expected behavior

_No response_

### Full `tauri info` output

```text
[✔] Environment
    ✔ WebView2: 134.0.3124.72
    - node: 20.18.3
    - pnpm: 10.6.2
    - yarn: 1.22.22
    - npm: 10.8.2

[-] Packages
    - tauri 🦀: 2.3.1
    - tauri-build 🦀: 2.0.6
    - wry 🦀: 0.50.5
    - tao 🦀: 0.32.8
    - @tauri-apps/api : 2.3.0 (outdated, latest: 2.4.0)
    - @tauri-apps/cli : 2.3.1 (outdated, latest: 2.4.0)

[-] Plugins
    - @tauri-apps/plugin-opener : 2.2.6

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - framework: Vue.js
    - bundler: Vite
```

### Stack trace

```text

```

### Additional context

_No response_

### LINKED ISSUE #14914: [bug] create_window race condition: HWND not registered before function returns
### Describe bug

`Tauri::WebviewWindowBuilder::build()` can return before HWND is registered in window map on event loop thread. Any immediate call to `.hwnd()` (or any method that reads window map) races against event loop's `Message::CreateWindow` handler and may panic/return stale value.

Discovered while debugging ARM64 WebView2 deadlock (tauri-apps/wry#1665): even after fixing wry-level deadlock, Tauri runtime would intermittently fail because `create_window` returned `WebviewWindow` whose HWND wasn't yet available.

### Root Cause

`tauri-runtime-wry`, `create_window` sends `Message::CreateWindow` to event loop via proxy and returns immediately. Event loop thread creates window, inserts HWND into `windows`, and installs WebView2 but caller is already past `build()` call and may attempt to use window handle before registered.

### Fix

Add `oneshot::Sender<Result<()>>` to `Message::CreateWindow`. Event loop handler sends `Ok(())` after window is fully created + registered. Caller blocks on receiver, ensuring HWND is available before `build()` returns.

### Minimal Reproduction

https://github.com/npiesco/wry-arm64-deadlock

- `main` branch: exhibits the race (and the upstream wry deadlock)
- `fix/vendored-wry-tauri` branch: includes both fixes

Race is most easily observed on Windows ARM64 (Snapdragon X Elite) where timing is different, but race condition exists on all platforms.

### Expected behavior

`WebviewWindowBuilder::build()` should not return until window's HWND is registered + available for use.

### Platform and Versions

- OS: Windows 11 ARM64 (Snapdragon X Elite)
- Rustc: 1.87.0 (aarch64-pc-windows-msvc)
- Tauri: 2.x (tauri-runtime-wry from dev branch)

### Related

- tauri-apps/wry#1665  WebView2 deadlock on ARM64 (root cause, fixed separately)
- tauri-apps/wry#583  Original deadlock report from 2022

Legend-Master:
Closes #10157
Closes #13046
Closes #14914
Closes #14915

Reference: https://github.com/tauri-apps/tauri/issues/10157#issuecomment-2241087241

#14915 did this with the same approach, but the use of LLM made tons of noise, that it covered up what the change was really about...

For reviewer: I don't know if this was a mistake not to do this or there's a legitimate reason for that

### DIFF
--- a/.changes/wait-async-window-creation.md
+++ b/.changes/wait-async-window-creation.md
@@ -1,0 +1,5 @@
+---
+"tauri-runtime-wry": patch:bug
+---
+
+`create_window` should wait for the window creation to complete before returning even if it's off main thread, it should also return the error if it failed
--- a/.changes/wait-async-window-creation.md
+++ b/.changes/wait-async-window-creation.md
@@ -2,4 +2,4 @@
 "tauri-runtime-wry": patch:bug
 ---
 
-`create_window` should wait for the window creation to complete before returning even if it's off main thread, it should also return the error if it failed
+`create_window` and `create_webview` should wait for the window creation to complete before returning even if it's off main thread, it should also return the error if it failed
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -379,6 +379,7 @@ impl<T: UserEvent> Context<T> {
     let window_id_wrapper = Arc::new(Mutex::new(window_id));
     let window_id_wrapper_ = window_id_wrapper.clone();
 
+    let (tx, rx) = channel();
     send_user_message(
       self,
       Message::CreateWebview(
@@ -394,8 +395,11 @@ impl<T: UserEvent> Context<T> {
             options.focused_webview,
           )
         }),
+        tx,
       ),
     )?;
+    rx.recv()
+      .map_err(|_| crate::Error::FailedToReceiveMessage)??;
 
     let dispatcher = WryWebviewDispatcher {
       window_id: window_id_wrapper,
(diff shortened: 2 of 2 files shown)


sftse (dismissed):
LGTM

Legend-Master:
@lucasfernog I'll wait a bit for your review before merging since you might know the history better

amrbashir (approved):

PR #15520, opened 2026-06-11
feat(api): add logical monitor bounds to the Monitor object

### LINKED ISSUE #15519: [feat] Add logical monitor bounds to the JS Monitor API
### Describe the problem

Tauri currently exposes monitor bounds from `availableMonitors()` as physical values:

```ts
const monitors = await availableMonitors()

monitor.position // PhysicalPosition
monitor.size     // PhysicalSize
monitor.scaleFactor
```

However, `WebviewWindow` creation options use logical pixels:

```ts
new WebviewWindow('second-window', {
  x,      // logical pixels
  y,      // logical pixels
  width,  // logical pixels
  height, // logical pixels
})
```

This means that application code cannot safely pass `monitor.position.x` / `monitor.position.y` into `WebviewWindow` options without knowing that it may need to divide by `monitor.scaleFactor`.

This is especially visible on macOS with Retina / mixed-DPI monitor setups. For example:

```text
Built-in display:
  logical position: 0,0
  logical size:     1440x900
  physical size:    2880x1800
  scaleFactor:      2

External display:
  logical position: 1440,0
  logical size:     1920x1080
  scaleFactor:      1
```

On macOS, using the physical monitor position as the initial `WebviewWindow` position can place the window incorrectly because the window API expects logical coordinates.

This is related to prior multi-monitor / macOS positioning reports, including:

- https://github.com/tauri-apps/tauri/issues/12167
- https://github.com/tauri-apps/tauri/issues/7139
- https://github.com/tauri-apps/tao/issues/816
- https://github.com/tauri-apps/tauri/pull/15250

Add logical monitor bounds to the JavaScript `Monitor` object returned by:

```ts
availableMonitors()
currentMonitor()
primaryMonitor()
monitorFromPoint()
```

Suggested API:

```ts
interface Monitor {
  name: string | null

  // Existing fields, unchanged
  position: PhysicalPosition
  size: PhysicalSize
  workArea: {
    position: PhysicalPosition
    size: PhysicalSize
  }
  scaleFactor: number

  // New fields
  logicalPosition: LogicalPosition
  logicalSize: LogicalSize
  logicalWorkArea: {
    position: LogicalPosition
    size: LogicalSize
  }
}
```

The new values can be derived in the JS API mapping layer:

```ts
const position = new PhysicalPosition(m.position)
const size = new PhysicalSize(m.size)
const workAreaPosition = new PhysicalPosition(m.workArea.position)
const workAreaSize = new PhysicalSize(m.workArea.size)

return {
  name: m.name,
  scaleFactor: m.scaleFactor,

  position,
  size,
  workArea: {
    position: workAreaPosition,
    size: workAreaSize,
  },

  logicalPosition: position.toLogical(m.scaleFactor),
  logicalSize: size.toLogical(m.scaleFactor),
  logicalWorkArea: {
    position: workAreaPosition.toLogical(m.scaleFactor),
    size: workAreaSize.toLogical(m.scaleFactor),
  },
}
```

Application code could then use monitor coordinates that match `WebviewWindow` options directly:

```ts
const monitor = (await availableMonitors())[1]

new WebviewWindow('second-window', {
  x: monitor.logicalPosition.x,
  y: monitor.logicalPosition.y,
  fullscreen: true,
})
```

### Alternatives considered

The monitor API returns physical positions/sizes, while `WebviewWindow` options expect logical positions/sizes. This requires userland conversion:

```ts
x: monitor.position.x / monitor.scaleFactor,
y: monitor.position.y / monitor.scaleFactor,
```

This workaround is easy to miss and is not obvious from the API shape.

### Additional context

_No response_

polw1:
Adds `logicalPosition`, `logicalSize` and `logicalWorkArea` to the `Monitor` object returned by `currentMonitor()`, `primaryMonitor()`, `monitorFromPoint()` and `availableMonitors()`.

The values are derived in `mapMonitor()` from the existing physical bounds via `toLogical(scaleFactor)`, so they match the logical units expected by `WebviewWindow` creation options. This removes the need for manual physical-to-logical conversion in userland, which is easy to miss on macOS mixed-DPI setups.

Existing `position`, `size`, `workArea` and `scaleFactor` fields are unchanged, so the change is non-breaking.

closes #15519

### DIFF
--- a/packages/api/src/window.ts
+++ b/packages/api/src/window.ts
@@ -47,26 +47,45 @@ import { Image, transformImage } from './image'
 export interface Monitor {
   /** Human-readable name of the monitor */
   name: string | null
-  /** The monitor's resolution. */
+  /**
+   * The monitor's resolution in physical pixels.
+   *
+   * Use {@linkcode Monitor.scaleFactor} to convert to logical pixels:
+   * ```typescript
+   * const logicalSize = monitor.size.toLogical(monitor.scaleFactor);
+   * ```
+   */
   size: PhysicalSize
-  /** the Top-left corner position of the monitor relative to the larger full screen area. */
+  /**
+   * the Top-left corner position of the monitor relative to the larger full screen area, in physical pixels.
+   *
+   * Note that window creation options such as `x`, `y`, `width` and `height` expect
+   * logical pixels, so convert with {@linkcode Monitor.scaleFactor} first:
+   * ```typescript
+   * import { currentMonitor } from '@tauri-apps/api/window';
+   * import { WebviewWindow } from '@tauri-apps/api/webviewWindow';
+   *
+   * const monitor = await currentMonitor();
+   * const position = monitor.position.toLogical(monitor.scaleFactor);
+   * const webview = new WebviewWindow('my-label', { x: position.x, y: position.y });
+   * ```
+   */
   position: PhysicalPosition
-  /** The monitor's work area. */
+  /**
+   * The monitor's work area (the monitor area excluding taskbars and docks) in physical pixels.
+   *
+   * Use {@linkcode Monitor.scaleFactor} to convert to logical pixels as shown in
+   * {@linkcode Monitor.position}.
+   */
   workArea: {
     position: PhysicalPosition
     size: PhysicalSize
   }
-  /** The scale factor that can be used to map physical pixels to logical pixels. */
+  /**
+   * The scale factor that can be used to map physical pixels to logical pixels,
+   * e.g. `monitor.position.toLogical(monitor.scaleFactor)`.
+   */
   scaleFactor: number
-  /** The monitor's resolution in logical pixels, matching the units used by window creation options. */
-  logicalSize: LogicalSize
-  /** The Top-left corner position of the monitor in logical pixels, matching the units used by window creation options. */
-  logicalPosition: LogicalPosition
-  /** The monitor's work area in logical pixels. */
-  logicalWorkArea: {
-    position: LogicalPosition
-    size: LogicalSize
-  }
 }
 
 type Theme = 'light' | 'dark'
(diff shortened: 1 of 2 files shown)


packages/api/src/window.ts
```diff
@@ -2543,18 +2552,31 @@ interface WindowOptions {
...
+  return {
+    name: m.name,
+    scaleFactor: m.scaleFactor,
+    position,
+    size,
+    workArea: {
+      position: workAreaPosition,
+      size: workAreaSize
+    },
+    logicalPosition: position.toLogical(m.scaleFactor),
```
Legend-Master: Not sure I get this change, couldn't the user do this themselves? Like it's

```ts
const monitor = await currentMonitor()
monitor.logicalPosition
// vs
monitor.position.toLogical(monitor.scaleFactor)
```

polw1: Thanks for pointing that out.

Today, both sides expose plain numbers: `monitor.position.x` / `monitor.position.y` on the monitor side, and `x` / `y` on the window creation side. Because of that, it is easy for someone developing mainly on Windows, or on a scaleFactor=1 setup, to pass the physical monitor coordinates directly into window creation and not notice the mismatch.

That code may appear to work locally, but then behave differently on macOS or on high-DPI / mixed-DPI setups.

So the idea here is not to introduce something completely new, but to make the common path easier and clearer, especially for new users or for users coming from frameworks where monitor bounds can usually be passed directly into window creation without doing a manual scale factor conversion.

Legend-Master: I see, I still feel like this should be done at the user side though... Maybe as an alternative, we could document how to convert them as examples in `Monitor.position` (and a few other fields)?

polw1: No doubt, clearer documentation around the physical/logical distinction would help a lot

Please let me know if this new direction looks good to you.

Legend-Master (approved):
The documentations should help a lot on this, thanks!

Issue #15541, opened 2026-06-14
[bug] Cannot compile brotli-decompressor on Windows — E0277 alloc_no_stdlib duplicate versions


Jerezincoc:
### Describe the bug

Fresh `create-tauri-app` project fails to compile on Windows with Rust 1.88.0. The error is `E0277: the trait bound StandardAlloc: alloc::Allocator<u8> is not satisfied` caused by duplicate versions of `alloc_no_stdlib` (2.0.4 and 3.0.0) in the dependency graph.

### Steps to reproduce

1. `npm create tauri-app@latest my-app` (React + TypeScript + pnpm)
2. `cd my-app && pnpm install`
3. `pnpm tauri dev`

### Expected behavior

Project compiles successfully.

### Actual behavior

```
error[E0277]: the trait bound `StandardAlloc: alloc::Allocator<u8>` is not satisfied
 --> brotli-8.x/src/enc/combined_alloc.rs
note: there are multiple different versions of crate `alloc_no_stdlib` in the dependency graph
  alloc-no-stdlib v2.0.4 (expected)
  alloc-no-stdlib v3.0.0 (found)
error: could not compile `brotli-decompressor` due to 36+ previous errors
```

### Environment

- OS: Windows 11 x64
- rustc: 1.88.0
- cargo: 1.88.0
- node: 20.20.0
- pnpm: 10.x
- create-tauri-app: latest
- tauri: 2.x (latest from create-tauri-app)

### Additional context

Tried:
- `cargo clean` + rebuild → same error
- `cargo update brotli --precise 3.5.0` → incompatible, Tauri requires `^7`
- Downgrading Rust to 1.85 → other dependencies require 1.86+

The root cause appears to be conflicting versions of `alloc_no_stdlib` pulled in by `brotli` and `brotli-decompressor`.

FabianLars:
upstream issues:
https://github.com/dropbox/rust-brotli/issues/256
https://github.com/dropbox/rust-alloc-no-stdlib/issues/22

workaround: `cargo update -p alloc-stdlib --precise 0.2.2` or `cargo update -p alloc-no-stdlib@3.0.0 --precise 2.0.4`

FabianLars:
this one really sucks cause we can't really lock this. cargo could decide not to de-duplicate the versions and pull in the broken stuff regardless. I'll try anyway.

FabianLars:
[...]

PR #15542, opened 2026-06-14
chore(deps): try to lock alloc-(no)-stdlib

FabianLars:
ref #15541


FabianLars:
this does not help at all, nice

FabianLars:
"fixed" upstream

PR #15544, opened 2026-06-15
chore(runtime-wry)!: remove unused dpi wrapper types

Legend-Master:
`tauri-runtime` and `tao` both uses `dpi` crate now that there's no need for the wrappers anymore

### DIFF
--- a/.changes/dpi-type-wrappers.md
+++ b/.changes/dpi-type-wrappers.md
@@ -1,0 +1,5 @@
+---
+"tauri-runtime-wry": "minor:breaking"
+---
+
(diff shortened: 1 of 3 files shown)


FabianLars (approved):

FabianLars:
i'm always anxious about non-major breaking changes x)

Legend-Master:
I'm trying to make those 'breaking changes' in unstable crate to only happen in minor versions as much as possible, so if someone actually depends on them (which they probably shouldn't or should pin the versions anyways), it helps a little

PR #15558, opened 2026-06-18
fix: drop tray icon on main thread

Legend-Master:
Fix #12534

PR #15559, opened 2026-06-18
refactor: simplify menu code

Legend-Master:
Separated from #15558

### DIFF
--- a/crates/tauri/src/menu/mod.rs
+++ b/crates/tauri/src/menu/mod.rs
@@ -76,6 +76,16 @@ macro_rules! gen_wrappers {
         app_handle: $crate::AppHandle<R>,
       }
 
+      impl<R: $crate::Runtime> $inner<R> {
+        fn new(app_handle: $crate::AppHandle<R>, menu: ::muda::$type) -> Self {
+          Self {
+            id: menu.id().clone(),
+            inner: ManuallyDrop::new(menu),
+            app_handle,
+          }
+        }
+      }
+
       /// # Safety
       ///
       /// We make sure it always runs on the main thread.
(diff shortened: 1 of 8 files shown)


crates/tauri/src/menu/mod.rs
```diff
@@ -70,11 +70,20 @@ macro_rules! gen_wrappers {
     $(
       #[tauri_macros::default_runtime(crate::Wry, wry)]
       pub(crate) struct $inner<R: $crate::Runtime> {
-        id: $crate::menu::MenuId,
-        inner: ::std::option::Option<::muda::$type>,
+        // SAFETY: we only call `ManuallyDrop::take` in [`Self::drop`] to drop it on main thread
+        inner: ManuallyDrop<::muda::$type>,
         app_handle: $crate::AppHandle<R>,
```
sftse, line 75: This change replaces a potential code bug that would result in a panic with one that on mistake results in UB. I think this is more complex, not less. We still have to uphold the same invariants, to only take the item from this location a single time, but the stakes are now much higher if it is wrong.

It's a bit unusual to mark the field with a `SAFETY` comment, having it on the unsafe block tends to be a better place.

Legend-Master: Moved this to below in `drop`, see if that makes more sense to you.

I personally don't mind either `Option` or `ManuallyDrop`, I just think `ManuallyDrop` has a better semantic here as well as a tiny bit lesser runtime cost.

(`ManuallyDrop` is safe as long as you use the unsafe `ManuallyDrop` functions like `ManuallyDrop::drop`/`ManuallyDrop::take` correctly)

sftse: I can see how the semantics are a clearer.

Imho the use of unsafety to argue performance has to be accompanied by performance measurements.

sftse (approved):

PR #15560, opened 2026-06-18
docs(cli): document signing private key path env var

nandanpugalia:
## Summary
- document `TAURI_SIGNING_PRIVATE_KEY_PATH` in the CLI environment variables reference
- clarify that it is used by the standalone `tauri signer sign` command

Related to tauri-apps/tauri-docs#3890.

## Verification
- `git diff --check`

Note: `pnpm exec prettier --check crates/tauri-cli/ENVIRONMENT_VARIABLES.md` could not run in this checkout because `prettier` is not installed locally.


Legend-Master:
Please clean up your LLM output. We don't accept fully automated agent PRs.

PR #15573, opened 2026-06-21
chore(deps): bump cloudflare worker versions

Legend-Master:
Hopefully fixes https://github.com/tauri-apps/tauri/actions/runs/27907781850/job/82579404185

### DIFF
--- a/.github/workflows/deploy-schema-worker.yml
+++ b/.github/workflows/deploy-schema-worker.yml
@@ -16,8 +16,8 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: cloudflare/wrangler-action@v3
+      - uses: actions/checkout@v6
+      - uses: cloudflare/wrangler-action@v4
         with:
           command: deploy
           workingDirectory: 'crates/tauri-schema-worker'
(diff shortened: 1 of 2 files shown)


FabianLars:
-17 on the supply chain security - ouch

FabianLars (approved):

Legend-Master:
Of course it doesn't work 😩

> https://github.com/tauri-apps/tauri/actions/runs/27924717300/job/82624833803

Seems like we're hitting

> https://github.com/cloudflare/workers-rs/issues/1014

Legend-Master:
The android tests are failing after this got merged, but I can't really tell why or if it's related 🤔

FabianLars:
Looks like the usual fluke we regularly have. Try deleting the cache (can't find it on my phone)

Edit: because I saw this in another pr yesterday I think

FabianLars:
Ah or maybe not quite, that usually complained about glibc right? Can't remember

Legend-Master:
I tried to clear the cache already, didn't work either so

> https://github.com/tauri-apps/tauri/actions/runs/27942931554/job/82686423245

Legend-Master:
Found the reason, `worker@0.8.5` (which marked its MSRV to be 1.75 which is not true) pulled in `wasm-streams@0.6.0` which is on edition 2024. And because other crates don't use this dependency, the test went fine, but cross calls `cargo metadata` which failed...

FabianLars:
Pain :'))

Legend-Master:
Since we're going to raise MSRV soon, I'm not gonna spend time fixing this then 🙃

PR #15576, opened 2026-06-23
test: regenerate stale macOS acl snapshot for has_app_acl

chippers:
PR #13418 added the has_app_acl field to the Resolved struct and regenerated the linux and windows snapshots but missed macOS. CI does not run this insta test on macOS, so cargo test fails on a fresh clone of dev. This adds the missing field to match the other platforms.


Legend-Master (approved):
Nice catch

PR #15578, opened 2026-06-23
enhance: remove `T: Send + Sync` requirement from `State<'r, T>`

sftse:
#13257 isn't getting any majorities.

This includes just the removal of some unsafe blocks and a bit of cleanup.

### DIFF
--- a/.changes/change-pr-15578.md
+++ b/.changes/change-pr-15578.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch:enhance
+---
+
+`State` had `Send` and `Sync` trait bounds that were already implied, remove them from the struct definition.
--- a/crates/tauri/src/state.rs
+++ b/crates/tauri/src/state.rs
@@ -6,7 +6,6 @@ use std::{
   any::{Any, TypeId},
   collections::HashMap,
   hash::BuildHasherDefault,
-  pin::Pin,
   sync::Mutex,
 };
 
@@ -100,7 +99,7 @@ impl std::hash::Hasher for IdentHash {
 /// Safety:
 /// - The `key` must equal to `(*value).type_id()`, see the safety doc in methods of [StateManager] for details.
 /// - Once you insert a value, you can't remove/mutated/move it anymore, see [StateManager::try_get] for details.
-type TypeIdMap = HashMap<TypeId, Pin<Box<dyn Any + Sync + Send>>, BuildHasherDefault<IdentHash>>;
+type TypeIdMap = HashMap<TypeId, Box<dyn Any + Sync + Send>, BuildHasherDefault<IdentHash>>;
 
 /// The Tauri state manager.
 #[derive(Debug)]
@@ -120,13 +119,12 @@ impl StateManager {
     let type_id = TypeId::of::<T>();
     let already_set = map.contains_key(&type_id);
     if !already_set {
-      let ptr = Box::new(state) as Box<dyn Any + Sync + Send>;
-      let pinned_ptr = Box::into_pin(ptr);
+      let state = Box::new(state) as Box<dyn Any + Sync + Send>;
       map.insert(
         type_id,
         // SAFETY: keep the type of the key is the same as the type of the value，
         // see [try_get] methods for details.
-        pinned_ptr,
+        state,
       );
     }
     !already_set
@@ -137,11 +135,9 @@ impl StateManager {
   pub(crate) unsafe fn unmanage<T: Send + Sync + 'static>(&self) -> Option<T> {
     let mut map = self.map.lock().unwrap();
     let type_id = TypeId::of::<T>();
-    let pinned_ptr = map.remove(&type_id)?;
-    // SAFETY: The caller decides to break the immovability/safety here, then OK, just let it go.
-    let ptr = unsafe { Pin::into_inner_unchecked(pinned_ptr) };
+    let state = map.remove(&type_id)?;
     let value = unsafe {
-      ptr
+      state
         .downcast::<T>()
         // SAFETY: the type of the key is the same as the type of the value
         .unwrap_unchecked()
@@ -160,9 +156,9 @@ impl StateManager {
   pub fn try_get<T: Send + Sync + 'static>(&self) -> Option<State<'_, T>> {
     let map = self.map.lock().unwrap();
     let type_id = TypeId::of::<T>();
-    let ptr = map.get(&type_id)?;
+    let state = map.get(&type_id)?;
     let value = unsafe {
-      ptr
+      state
         .downcast_ref::<T>()
         // SAFETY: the type of the key is the same as the type of the value
         .unwrap_unchecked()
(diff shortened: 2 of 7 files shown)


sftse:
Something is broken about cross compilation.

Legend-Master:
> Something is broken about cross compilation.

Yeah, let's ignore that for now, see https://github.com/tauri-apps/tauri/pull/15573#issuecomment-4770180783

crates/tauri/src/state.rs
```diff
@@ -18,19 +17,21 @@ use crate::{
 /// A guard for a state value.
 ///
 /// See [`Manager::manage`](`crate::Manager::manage`) for usage examples.
-pub struct State<'r, T: Send + Sync + 'static>(&'r T);
+#[derive(Clone)]
+pub struct State<'r, T>(&'r T);
```
Legend-Master, line 20: I feel like dropping the `Send + Sync + 'static` bound here is just making things more confusing since you should only use this struct through the state APIs which all require the `T` to be `Send + Sync + 'static`

sftse: There is a partial consensus in Rust that doing so is not very [helpful](https://stackoverflow.com/questions/49229332/should-trait-bounds-be-duplicated-in-struct-and-impl), and the standard library doesn't do this either, things like `HashMap` do not put their bounds on the struct even though there is no way to insert an element with a key unless it impls `Hash`.

In practice I've found bounds on the struct to be quite annoying if I need to quickly whip up a method on the struct for debugging or exploring something about the code, it forces me to haul the type bounds around even if the thing I'm implementing doesn't need them.
~~This hasn't been true for a while due to [RFC2089](https://rust-lang.github.io/rfcs/2089-implied-bounds.html), not sure why I still believed that.~~
Corrected the correction.

As a secondary reason it makes it harder to see what impact an attempt to change the bounds would have on the code, and if every struct needs the union of the bounds of the structs it contains it can become cumbersome quickly.

Legend-Master: [RFC2089](https://rust-lang.github.io/rfcs/2089-implied-bounds.html) has not been implemented right?

Anyways, I will take a look at https://stackoverflow.com/questions/49229332/should-trait-bounds-be-duplicated-in-struct-and-impl tomorrow and see if I could understand the trade offs better. Thanks for explaining

sftse: > [RFC2089](https://rust-lang.github.io/rfcs/2089-implied-bounds.html) has not been implemented right?

Thanks, right, I got confused how I could've missed that.

crates/tauri/src/state.rs
```diff
@@ -18,19 +17,21 @@ use crate::{
 /// A guard for a state value.
 ///
 /// See [`Manager::manage`](`crate::Manager::manage`) for usage examples.
-pub struct State<'r, T: Send + Sync + 'static>(&'r T);
+#[derive(Clone)]
```
Legend-Master: The derive will require `T` to be clone which is not what we want here

crates/tauri/src/state.rs
```diff
@@ -18,19 +17,21 @@ use crate::{
...
+pub struct State<'r, T>(&'r T);
 
-impl<'r, T: Send + Sync + 'static> State<'r, T> {
+impl<'r, T> State<'r, T> {
   /// Retrieve a borrow to the underlying value with a lifetime of `'r`.
   /// Using this method is typically unnecessary as `State` implements
   /// [`std::ops::Deref`] with a [`std::ops::Deref::Target`] of `T`.
+  #[doc(hidden)]
```
Legend-Master: Looking at how we use this everywhere in https://github.com/tauri-apps/plugins-workspace, I don't think we should hide this

crates/tauri/src/state.rs
Legend-Master: Sound reasonable and since you can't construct a `State` directly, this change works for me

Legend-Master (changes requested):

crates/tauri/src/state.rs
sftse: I see, that is a problem. Idea was to nudge users towards using the `Deref` impl which is a bit easier for us to reason about, it just kind of seems unnecessary to have both `Deref` and `fn inner()`.

Going to revert.

Legend-Master: `inner` has a different lifetime that's quite a bit more ergonomic to use so this is not exactly redundant.

sftse: `State` holds a `&'r T` which `fn inner` opens access to. Ignoring the discussion in #13257 where the privacy of the field is used to delay the destructor of `T`, this is the same as `State(pub &'r T)`. Code that is safe with `fn inner()` can be equivalently expressed with `Deref`.

sftse: The reference is more ergonomic than `State` I agree, but I think this is more because `State` lacks `Copy`, which we could derive if we want to.

Legend-Master: I think even with `Copy`, we will still have `value referencing temporary value` errors sometimes though? A not so good example would be

```rust
impl<R: Runtime, T: Manager<R>> crate::SampleExt<R> for T {
  fn sample(&self) -> &Sample<R> {
    &self.state::<Sample<R>>().deref()
  }
}
```

but you get the idea

sftse: Yes, so my comment may be entirely in the realm of the theoretical because of the breakage it would cause, but simply stating that every user of Tauri admits a refactor that still compiles if we remove `fn inner()`. The lifetime is not critical to callers compiling is what I wanted to say.

Here `fn sample` would have to return `State<'_, Sample<R>>` instead of `&Sample<R>` and every user of that method would not notice a thing because of `Deref` that makes `State<'_, Sample<R>>` ~ `&Sample<R>`.

But yes, this is too insignificant to further pursue.

crates/tauri/src/state.rs
```diff
@@ -98,9 +97,8 @@ impl std::hash::Hasher for IdentHash {
 }
 
 /// Safety:
-/// - The `key` must equal to `(*value).type_id()`, see the safety doc in methods of [StateManager] for details.
```
Legend-Master, line 101: Let's keep this line. We still have an `expect` even without any unsafe code

Legend-Master (commented):
~~Let's also add a change file~~ never mind, stale review comment

.changes/change-pr-15578.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri": patch:enhance
+---
+
+Cleanup state implementation. No user facing changes.
```
Legend-Master: Since `State` is a public type, let's document what actually happened, e.g. the removal of `Send + Sync`

.changes/change-pr-15578.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri": patch:enhance
```
Legend-Master, line 2:
```suggestion
"tauri": minor:enhance
```

sftse: Doesn't make a difference to me, but minor usually implies adding api surface, patch is for non-breaking changes to already existing API, sure?

Legend-Master: Yeah since we're already on the 2.12 train now, no differences.

I was thinking it should be a minor bump because have `impl` more things to `State` and it no longer requires `Send + Sync`

Legend-Master (approved):

PR #15600, opened 2026-06-29
chore(renovate): group napi-rs packages

Legend-Master:
This should probably be merged after the MSRV bump so we can re-evaluate which crates we can get in from the spamming PRs

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -34,6 +34,11 @@
       "description": "Group worker-rs crates",
       "groupName": "worker-rs crates",
       "matchSourceUrls": ["https://github.com/cloudflare/workers-rs"]
+    },
+    {
+      "description": "Group napi-rs packages",
+      "groupName": "napi-rs packages",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -34,6 +34,11 @@
       "description": "Group worker-rs crates",
       "groupName": "worker-rs crates",
       "matchSourceUrls": ["https://github.com/cloudflare/workers-rs"]
+    },
+    {
+      "description": "Group napi-rs packages",
+      "groupName": "napi-rs packages",
+      "matchSourceUrls": ["https://github.com/napi-rs/napi-rs"]
```
Legend-Master, line 41: Honestly I don't know if https://github.com/napi-rs/napi-rs is correct since they don't show the `(source)` link for those crates/packages

> https://github.com/tauri-apps/tauri/pull/15590

FabianLars: only one way to find out - if it doesn't work i think matching on the package names ala `napi*` should be fairly save

FabianLars (approved):

PR #15602, opened 2026-06-29
chore(ci): migrate to use `--cross-compile`

Legend-Master:
Mirror https://github.com/tauri-apps/create-tauri-app/pull/983

Migrate to use `--cross-compile` instead of docker to build cli-js for musl targets, this should unblock us from updating NAPI packages/crates

Reference: https://github.com/tauri-apps/tauri/pull/15310#issuecomment-4355101884

---

Good run: https://github.com/tauri-apps/tauri/actions/runs/28387392871?pr=15602


Legend-Master:
Seems like the `aarch64-unknown-linux-gnu` target just hit https://github.com/briansmith/ring/issues/1728, because NAPI pulls in `@napi-rs/cross-toolchain-x64-target-aarch64@1.0.3` when running through the cross build chain and it has gcc 4.8.5 which seems to be too old for ring... (similar to https://github.com/napi-rs/cross-toolchain/issues/4)

Edit: fixed by moving to `--cross-compile` which uses `cargo-zigbuild`

Legend-Master:
Good run: https://github.com/tauri-apps/tauri/actions/runs/28387392871?pr=15602

FabianLars (approved):

PR #15603, opened 2026-06-29
fix(acl): make ResolvedCommand referenced_by unconditional (fix #15406)

### LINKED ISSUE #13865: [bug] rust-analyzer throws errors about missing field referenced_by when tauri-utils is built without debug_assertions
### Describe the bug

Disabling `debug_assertions` when building tauri-utils causes IDE errors from rust-analyzer when calling `tauri::generate_context!()` claiming that 'referenced_by' field is missing several times:

Image

I discovered this because I with the following in my `Cargo.toml` for improved performance of dev builds:
```toml

# Enable high optimizations for dependencies (incl. Bevy), but not for our code:
[profile.dev]
incremental = true            # Improve consecutive recompile performance
opt-level = 1                 # Enable a small amount of optimization in debug mode

[profile.dev.package."*"]
incremental = true
debug = false
opt-level = 3
codegen-units = 16
overflow-checks = false
debug-assertions = false
```

Using cargo-expand before and after enabling `debug_assertions` I can see that the `ResolvedCommand` struct is generated in both cases, but `referenced_by` field is omitted when it is disabled:

<details>
<summary>cargo-expand output</summary>

`debug_assertions = false`:

```rust
                                    map.insert(
                                        "plugin:app|identifier".into(),
                                        <[_]>::into_vec(
                                            ::alloc::boxed::box_new([
                                                ::tauri::utils::acl::resolved::ResolvedCommand {
                                                    context: ::tauri::utils::acl::ExecutionContext::Local,
                                                    windows: <[_]>::into_vec(
                                                        ::alloc::boxed::box_new(["main".parse().unwrap()]),
                                                    ),
                                                    webviews: ::alloc::vec::Vec::new(),
                                                    scope_id: ::core::option::Option::None,
                                                },
                                            ]),
                                        ),
                                    );
```

`debug_assertions = true`:

```rust
                                    map.insert(
                                        "plugin:app|identifier".into(),
                                        <[_]>::into_vec(
                                            ::alloc::boxed::box_new([
                                                ::tauri::utils::acl::resolved::ResolvedCommand {
                                                    context: ::tauri::utils::acl::ExecutionContext::Local,
                                                    referenced_by: ::tauri::utils::acl::resolved::ResolvedCommandReference {
                                                        capability: "default".into(),
                                                        permission: "allow-identifier".into(),
                                                    },
                                                    windows: <[_]>::into_vec(
                                                        ::alloc::boxed::box_new(["main".parse().unwrap()]),
                                                    ),
                                                    webviews: ::alloc::vec::Vec::new(),
                                                    scope_id: ::core::option::Option::None,
                                                },
                                            ]),
                                        ),
                                    );
```

</details>

Since the app actually launches fine without referenced_by, my guess is the parameter should be made optional to avoid IDE errors.

### Reproduction

1. `cargo tauri init`
2. Edit `cargo.toml` and set `debug_assertions = false`
3. Open lib.rs in IDE

### Expected behavior

No errors

### Full `tauri info` output

```text
[✔] Environment
    ✔ rustc: 1.90.0-nightly (e43d139a8 2025-07-09)
    ✔ cargo: 1.90.0-nightly (930b4f62c 2025-06-28)
    ✔ rustup: 1.28.2 (2025-04-28)
    ✔ Rust toolchain: nightly-aarch64-apple-darwin (overridden by environment variable RUSTUP_TOOLCHAIN)
    - node: 24.2.0
    - npm: 11.3.0

[-] Packages
    - tauri 🦀: 2.6.2
    - tauri-build 🦀: 2.3.0
    - wry 🦀: 0.52.1
    - tao 🦀: 0.34.0
    - tauri-cli 🦀: 2.5.0 (outdated, latest: 2.6.2)

[-] Plugins

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: webui
    - devUrl: localhost:5174
```

### Stack trace

```text

```

### Additional context

_No response_

### LINKED ISSUE #15406: [bug] tauri-utils: cfg(debug_assertions) gates on ResolvedCommand break proc-macro / consumer interop across mixed profiles
### Crates
`tauri-utils` 2.9.1, `tauri-macros` 2.5.5
### Problem
`ResolvedCommand` has a field gated by `#[cfg(debug_assertions)]`:
```rust
pub struct ResolvedCommand {
    pub context: ExecutionContext,
    #[cfg(debug_assertions)]
    pub referenced_by: ResolvedCommandReference,
    pub windows: Vec<glob::Pattern>,
    pub webviews: Vec<glob::Pattern>,
    pub scope_id: Option<ScopeKey>,
}
```
The `ToTokens` impl emits a struct literal whose shape **also** varies on `cfg(debug_assertions)`. Same pattern for `ResolvedCommandReference` and the private `referenced_by_permission_identifier`.
For proc-macros, `cfg(debug_assertions)` reflects the proc-macro's own host build, not the consumer's target. When they disagree, `generate_context!()` emits a struct literal referencing a field the consumer can't see:
```
error[E0560]: struct `tauri::utils::acl::resolved::ResolvedCommand`
              has no field named `referenced_by`
```

### Reproduction

```toml
[dependencies]
tauri = { version = "2.10", features = ["custom-protocol"] }
[build-dependencies]
tauri-build = "2.5"
# Force proc-macros to debug while consumer goes release.
[profile.release.build-override]
debug-assertions = true
[profile.release.package."*"]
debug-assertions = false
```
```rust
// build.rs
fn main() { tauri_build::build(); }
// src/main.rs
fn main() { let _ = tauri::generate_context!(); }
```
`cargo build --release` → field-mismatch error.
Default cargo builds don't hit this, but profile overrides and any build system that compiles proc-macros and consumers under independent profiles do.

### Expected behavior

Make `referenced_by` (and the private field) unconditional. Two `String`s per command — negligible cost, diagnostic-only, callers can ignore in release.

### Full `tauri info` output

```text
- OS: Windows 11 (x86_64)
- rustc: 1.95.0
- tauri: 2.10.3
- tauri-build: 2.5.6
```

### Stack trace

```text

```

### Additional context

_No response_

tenderdeve:
Closes #15406.
Closes #13865

`ResolvedCommand::referenced_by` and `ResolvedCommandReference` were gated behind `#[cfg(debug_assertions)]`. Because a proc-macro's `cfg(debug_assertions)` reflects the macro's own host build (not the consumer's target), `generate_context!`'s `ToTokens` could emit a struct literal whose shape disagreed with the consumer's view of `ResolvedCommand`.

Under mismatched profiles — e.g.:

```toml
[profile.release.build-override]
debug-assertions = true
[profile.release.package."*"]
debug-assertions = false
```

`cargo build --release` fails with `error[E0560]: struct ... has no field named referenced_by`.

Fix: drop the `cfg(debug_assertions)` gating on the field, on `ResolvedCommandReference`, on the `referenced_by_permission_identifier` plumbing, and on the `ToTokens` emission, so the resolved struct shape is identical across build profiles. Cost is two `String`s per command, diagnostic-only.

Tested: `cargo test -p tauri-utils --features build` (41 passing), plus `cargo check -p tauri` in both debug and release.


Legend-Master (dismissed):
See https://github.com/tauri-apps/tauri/issues/15406#issuecomment-4862009787

tenderdeve:
@Legend-Master right that the `cfg` lives entirely in `tauri-utils` — the catch is `tauri-utils` gets compiled twice with two different `debug_assertions` values, and the `cfg` is evaluated independently in each:

- **Host copy** — `generate_context!` (tauri-macros → tauri-codegen → `tauri_utils::acl::resolved::build::ToTokens`) runs in the proc-macro, so it's compiled under the *build-override* profile. Its `debug_assertions` decides whether the emitted struct literal includes `referenced_by`.
- **Target copy** — the `ResolvedCommand` struct the emitted tokens must match is compiled under the app's normal profile. Its `debug_assertions` decides whether the field *exists*.

In a plain `cargo build` / `cargo build --release` both copies share the same `debug_assertions`, so they agree and nothing breaks — that's why it doesn't reproduce normally. It only breaks when the two profiles disagree, which `build-override` lets you do.

Minimal repro (any tauri v2 app), add to `src-tauri/Cargo.toml`:

```toml
[profile.release.build-override]
debug-assertions = true
```

then `cargo tauri build` (or `cargo build --release` in `src-tauri`). Host proc-macro now has `debug_assertions = true` → emits `referenced_by`; runtime `tauri-utils` is release (`debug_assertions = false`) → struct has no such field:

```
error[E0560]: struct `tauri::utils::acl::resolved::ResolvedCommand` has no field named `referenced_by`
```

Flip it (`profile.dev.build-override.debug-assertions = false` + `cargo build`) and you get the mirror error where the token stream omits a field the struct requires. Making the field unconditional removes the profile dependence entirely; cost is one extra `ResolvedCommandReference` (two `String`s) per resolved command in release. Happy to instead gate the emission on the *target* profile if you'd rather keep it debug-only, but there's no clean way for a proc-macro to read the consumer's `debug_assertions`, so unconditional looked like the robust fix.

Legend-Master:
That sounds right in theory, but I can't seem to reproduce it...

tenderdeve:
@Legend-Master got a self-contained repro. Two gotchas that make it not reproduce:

1. **`[profile.*.build-override]` only takes effect from the workspace root** (or a standalone `src-tauri`). If `src-tauri` is a workspace *member*, Cargo silently ignores the profile override — no warning in some setups. That's the most likely reason it didn't repro for you.
2. Needs a clean rebuild so `generate_context!` actually re-expands under the new host profile (`cargo clean` or touch the macro input).

You don't even need tauri to see the class of bug — here's a 3-crate workspace that reproduces the exact `E0560`. The `shared` crate mirrors `ResolvedCommand` (cfg-gated field), `pmac` mirrors the `ToTokens` codegen:

`Cargo.toml` (workspace root):
```toml
[workspace]
members = ["shared", "pmac", "app"]
resolver = "2"
[profile.release.build-override]
debug-assertions = true
```

`shared/src/lib.rs`:
```rust
#[derive(Default)]
pub struct Cmd {
    pub context: u32,
    #[cfg(debug_assertions)]
    pub referenced_by: String,
}
```

`pmac/src/lib.rs` (proc-macro, deps: `shared`):
```rust
use proc_macro::TokenStream;
#[proc_macro]
pub fn make_cmd(_: TokenStream) -> TokenStream {
    let mut fields = String::from("context: 1u32,");
    if cfg!(debug_assertions) {              // HOST copy of `shared`, under build-override
        fields.push_str("referenced_by: String::new(),");
    }
    format!("shared::Cmd {{ {fields} }}").parse().unwrap()
}
```

`app/src/main.rs` (deps: `shared`, `pmac`):
```rust
fn main() { let _c: shared::Cmd = pmac::make_cmd!(); }
```

`cargo build --release`:
```
error[E0560]: struct `Cmd` has no field named `referenced_by`
 --> app/src/main.rs:2:27
```

Host proc-macro copy of `shared` compiles with `debug_assertions = true` (build-override) → emits `referenced_by`; target copy is release `= false` → struct lacks the field. That's precisely what `generate_context!` → `tauri-codegen` → `tauri_utils::acl::resolved::build::ToTokens` does vs the `ResolvedCommand` linked into the app. Making the field unconditional removes the profile dependence for one `ResolvedCommandReference` (two `String`s) per resolved command in release — happy to gate on the target profile instead if you'd prefer debug-only, but a proc-macro can't read the consumer's `debug_assertions`, so unconditional was the robust option.

Legend-Master:
Could you record a short video on the repro steps? Like from `create-tauri-app` to adding what to the `Cargo.toml` and using what to build and the final error.

tenderdeve:
@Legend-Master can't easily attach a video, but here's the full copy-pasteable transcript from `create-tauri-app` to the error — real tauri, no synthetic crates. macOS, `cargo 1.96.0`, `create-tauri-app@4.6.2`, `tauri-utils 2.9.3`.

```
$ npm create tauri-app@latest repro -- --template vanilla --manager pnpm --yes
Template created!

$ cd repro/src-tauri

# append the build-override that lets host & target profiles disagree
$ printf '\n[profile.release.build-override]\ndebug-assertions = true\n' >> Cargo.toml

$ cargo build --release
   Compiling repro v0.1.0 (.../repro/src-tauri)
error[E0422]: cannot find struct, variant or union type `ResolvedCommandReference` in module `::tauri::utils::acl::resolved`
  --> src/lib.rs:12:14
   |
12 |         .run(tauri::generate_context!())
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in `::tauri::utils::acl::resolved`
   |
note: found an item that was configured out
  --> .../tauri-utils-2.9.3/src/acl/resolved.rs:25:12
   |
23 | #[cfg(debug_assertions)]
   |       ---------------- the item is gated here
25 | pub struct ResolvedCommandReference {
   |            ^^^^^^^^^^^^^^^^^^^^^^^^

error[E0560]: struct `ResolvedCommand` has no field named `referenced_by`
  --> src/lib.rs:12:14
   |
12 |         .run(tauri::generate_context!())
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^ `ResolvedCommand` does not have this field
   |
   = note: this error originates in the macro `tauri::generate_context`

error: could not compile `repro` (lib) due to 2 previous errors
```

Three things that make it *not* reproduce (likely why it didn't for you):

1. **`src-tauri` must be the manifest root** where the `[profile.*.build-override]` lives (default `create-tauri-app` layout — good). If `src-tauri` is a workspace *member*, Cargo ignores profile overrides in the member manifest, so the host/target profiles never diverge and it builds fine.
2. **`--release`** — the override flips the *host* build-script/proc-macro profile to `debug_assertions = true` while the target crate stays release (`= false`). Both halves of `tauri-utils` must disagree; a plain `cargo build` keeps them equal.
3. Clean expansion — first `--release` build already re-expands `generate_context!` under the new host profile, so no `cargo clean` needed here, but if you'd already built release before adding the override, touch `src/lib.rs` or `cargo clean`.

The mirror case (`profile.dev.build-override.debug-assertions = false` + `cargo build`) gives the inverse error where the token stream omits a field the struct requires. Making the field unconditional removes the profile dependence entirely.

Legend-Master:
I see why 🤦‍♂️, the project I used to test things didn't have any allowed permissions, thanks for the repro

tenderdeve:
@Legend-Master thanks for confirming. green now except the two android jobs, and those are a cross pre-compile infra thing, nothing to do with this.

review's still on changes-requested from before the repro - could you take another look when you have a sec?

and if you'd rather keep referenced_by debug-only i can gate it on the target profile instead, just went unconditional since the proc-macro can't read the consumer's debug_assertions. either way works for me.

PR #15604, opened 2026-06-29
fix(core): purge JS event listeners when source webview is destroyed (fix #15583)

### LINKED ISSUE #15583: [bug] JS event listeners are not removed when the source Webview is destroyed
### Describe the bug

In Tauri v2, JavaScript event listeners registered from a Webview are stored in the backend `Listeners::js_event_listeners` map, keyed by the source webview label.

When the source Webview/Window is destroyed, Tauri removes the Webview/Window handle from the manager, but it does not remove the JS event listener entries registered by that source Webview.

After the source Webview is destroyed, its JavaScript runtime no longer exists, so those callbacks can never be delivered. The listener metadata is therefore stale unless the frontend successfully calls `unlisten()` before teardown.

This is especially awkward for close handling: an app may need to call `unlisten()` from `onCloseRequested` before allowing/destroying the window, just to avoid stale backend listener entries. That makes the window close path depend on async IPC cleanup.

### Reproduction

I do not have a minimal reproduction repo yet, but the behavior can be seen from the current implementation:

1. From a Webview, register a JS event listener:

   ```ts
   import { listen } from '@tauri-apps/api/event'

   const unlisten = await listen('some-event', () => {
     console.log('event received')
   })
   ```

2. Destroy or close the source Webview/Window without calling `unlisten()`.

3. The Webview/Window handle is removed from the manager, but the entry in `Listeners::js_event_listeners` remains.

Relevant current code:

- `tauri/crates/tauri/src/event/listener.rs`
  - `js_event_listeners: Mutex<HashMap<WebviewLabel, HashMap<crate::EventName, HashSet<JsHandler>>>>,`

### Expected behavior

When a Webview is destroyed, Tauri should remove JS event listener entries that were registered from that source Webview.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 26.5.1 arm64 (X64)
    ✔ Xcode: 26.5
    - node: 26.3.0
    - pnpm: 11.3.0

[-] Packages
    - tauri 🦀: 2.10.3, (outdated, latest: 2.11.3)
    - tauri-build 🦀: 2.5.6, (outdated, latest: 2.6.3)
    - wry 🦀: 0.54.2, (outdated, latest: 0.55.1)
    - tao 🦀: 0.34.8, (outdated, latest: 0.35.3)
    - @tauri-apps/api  ⱼₛ: 2.10.1 (outdated, latest: 2.11.1)
    - @tauri-apps/cli  ⱼₛ: 2.10.1 (outdated, latest: 2.11.3)

[-] Plugins
    - ...

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ./dist/
    - framework: React
    - bundler: Vite
```

### Stack trace

```text
No
```

### Additional context

From the application side, once the source Webview is destroyed, the JS callback cannot be invoked anymore because the JS runtime is gone. Keeping the backend listener entry does not appear useful.

Related issue:
- #8916

*Disclaimer: This issue was submitted by AI and I have reviewed it.*

tenderdeve:
Closes #15583.

JS event listeners registered from a webview are stored in `Listeners::js_event_listeners`, keyed by the source webview label. When the webview is destroyed the manager removes its handle from the store but leaves the listener entries behind. Since the webview's JavaScript runtime is gone, those callbacks can never be delivered — the entries just leak until the app exits, and apps are forced to call `unlisten()` from `onCloseRequested` before closing a window just to keep the backend map clean.

Add `Listeners::remove_webview_events(label)` and call it from `on_webview_close` and `on_window_close` (for each webview of the closing window), so a destroyed webview's JS listeners are dropped alongside its handle.

Added a unit test (`js_listeners_removed_on_webview_close`) covering register → close → gone.

Note: @Legend-Master pointed out the Rust-side `handlers` map may leak similarly. That one is keyed by `EventTarget` rather than the owning webview, so purging it correctly has to account for `Window` vs `Webview` targets and multi-webview windows. Left out of this PR to keep it focused on the reported JS-listener leak — happy to follow up if you'd like it handled here.

Tested: `cargo test -p tauri --lib event::listener` (5 passing), fmt + clippy clean.

### DIFF
--- a/crates/tauri/src/event/listener.rs
+++ b/crates/tauri/src/event/listener.rs
@@ -256,7 +256,7 @@ impl Listeners {
   /// Called when a webview is destroyed: its JavaScript runtime no longer
   /// exists, so the listeners keyed by its label can never be delivered and
   /// would otherwise leak in the map until the app exits.
-  pub(crate) fn remove_webview_events(&self, webview_label: &str) {
+  pub(crate) fn remove_webview_js_listeners(&self, webview_label: &str) {
     self
       .inner
       .js_event_listeners
@@ -431,7 +431,7 @@ mod test {
     assert!(listeners.has_js_listener(event.as_str_event(), |_| true));
 
     // dropping the source webview must drop its JS listeners.
-    listeners.remove_webview_events(webview_label);
+    listeners.remove_webview_js_listeners(webview_label);
     assert!(!listeners.has_js_listener(event.as_str_event(), |_| true));
   }
 }
--- a/crates/tauri/src/event/listener.rs
+++ b/crates/tauri/src/event/listener.rs
@@ -251,6 +251,20 @@ impl Listeners {
     }
   }
 
+  /// Removes all JS event listeners registered from the given webview.
+  ///
+  /// Called when a webview is destroyed: its JavaScript runtime no longer
+  /// exists, so the listeners keyed by its label can never be delivered and
+  /// would otherwise leak in the map until the app exits.
+  pub(crate) fn remove_webview_events(&self, webview_label: &str) {
+    self
+      .inner
+      .js_event_listeners
+      .lock()
+      .unwrap()
+      .remove(webview_label);
+  }
+
   pub(crate) fn has_js_listener<F: Fn(&EventTarget) -> bool>(
     &self,
     event: crate::EventName<&str>,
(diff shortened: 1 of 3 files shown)


crates/tauri/src/event/listener.rs
```diff
@@ -251,6 +251,20 @@ impl Listeners {
     }
   }
 
+  /// Removes all JS event listeners registered from the given webview.
+  ///
+  /// Called when a webview is destroyed: its JavaScript runtime no longer
+  /// exists, so the listeners keyed by its label can never be delivered and
+  /// would otherwise leak in the map until the app exits.
+  pub(crate) fn remove_webview_events(&self, webview_label: &str) {
```
Legend-Master:
```suggestion
  pub(crate) fn remove_webview_js_listeners(&self, webview_label: &str) {
```

Legend-Master (commented):
Looks good otherwise

tenderdeve:
@Legend-Master updated in fdf05c08c — renamed `remove_webview_events` → `remove_webview_js_listeners` across all call sites.

wyhaya:
Can this RP be merged?

Legend-Master (dismissed):
Thanks.

You'll need to run `cargo fmt` and sign your commits for me to merge this

tenderdeve:
@Legend-Master the approval got dismissed when i pushed a small rustfmt fixup (the rename pushed one call site over the line limit, fmt just split it - 0947c2f1b). ci's green again, mind re-approving when you get a sec?

@wyhaya should be good to merge once it's re-approved 👍

Legend-Master:
> and sign your commits for me to merge this

Just so you know there's 2 requests here from the last review

tenderdeve:
@Legend-Master both done — `cargo fmt` applied and all commits are now signed (showing Verified). Ready for re-review 🙏

Legend-Master (approved):
Thanks!

PR #15617, opened 2026-06-30
fix(core): purge Rust event listeners when window/webview is destroyed

### LINKED ISSUE #15613: [bug] event listeners are not cleaned up when its `EventTarget` is destroyed
> Seems like it's also the case for the rust side listeners 

 _Originally posted by @Legend-Master in [#15583](https://github.com/tauri-apps/tauri/issues/15583#issuecomment-4798640712)_

Not exactly easy to fix with how the filters work today though

tenderdeve:
Closes #15613 (Rust-side follow-up to the listener-leak discussion in #15583).

Rust-side listeners registered via `listen`/`once` on a `Window`, `Webview` or `WebviewWindow` were never removed when that target was destroyed, so the handlers leaked for the lifetime of the app.

### Change

- Add `Listeners::remove_listeners_for_label`, which retains only handlers whose `EventTarget` is not bound to the destroyed label. It follows the existing `try_lock` + `Pending` queue convention so it stays safe if called mid-emit.
- Wire it into `AppManager::on_window_close` (window label + each child webview label) and `on_webview_close`.

`App`, `Any` and `AnyLabel` listeners are kept — only `Window`/`Webview`/`WebviewWindow` targets matching the closed label are dropped.

### Open question

I left `AnyLabel { label }` listeners intact, treating them as explicit cross-target subscriptions rather than something owned by a single window/webview. Happy to also purge those on matching label if you think that fits the intended semantics better.

This is scoped to the native (`handlers`) map only; the JS-side `js_event_listeners` cleanup is a separate concern.

### Test

Added `remove_listeners_for_label_drops_only_owned_targets` verifying that `Window`/`Webview`/`WebviewWindow` listeners for the closed label are removed while unrelated-label, `App`, `Any` and `AnyLabel` listeners survive.

### DIFF
--- a/crates/tauri/src/event/listener.rs
+++ b/crates/tauri/src/event/listener.rs
@@ -190,11 +190,8 @@ impl Listeners {
 
   /// Removes every Rust-side event listener registered for the exact given target.
   ///
-  /// Called when a window or webview is destroyed so its listeners don't leak. Matching
-  /// is done on the full [`EventTarget`] (kind + label) rather than the label alone,
-  /// since in multi-webview mode a window and its webviews can carry the same label while
-  /// referring to different targets. Global targets such as [`EventTarget::App`],
-  /// [`EventTarget::Any`] and [`EventTarget::AnyLabel`] are never matched here.
+  /// Called when a window or webview is destroyed so its listeners don't leak.
+  /// Global targets such as [`EventTarget::App`], [`EventTarget::Any`] and [`EventTarget::AnyLabel`] are never matched here.
   pub(crate) fn remove_listeners_for_target(&self, target: EventTarget) {
     match self.inner.handlers.try_lock() {
       Err(_) => self.insert_pending(Pending::RemoveForTarget(target)),
--- a/crates/tauri/src/event/listener.rs
+++ b/crates/tauri/src/event/listener.rs
@@ -25,6 +25,7 @@ enum Pending {
     handler: Handler,
   },
   Emit(EmitArgs),
+  RemoveForTarget(EventTarget),
 }
 
 /// Stored in [`Listeners`] to be called upon, when the event that stored it, is triggered.
@@ -128,6 +129,7 @@ impl Listeners {
         Pending::Unlisten(id) => self.unlisten(id),
         Pending::Listen { id, event, handler } => self.listen_with_id(id, event, handler),
         Pending::Emit(args) => self.emit(args)?,
+        Pending::RemoveForTarget(target) => self.remove_listeners_for_target(target),
       }
     }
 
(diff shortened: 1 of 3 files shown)


Legend-Master:
Please sign your commits and rebase .

tenderdeve:
@Legend-Master rebased on `dev` and signed — commit `5baef0420` is now Verified and the PR is mergeable.

The rebase conflicted with the JS-listener cleanup from #15604 (now on `dev`). Resolved it by keeping **both**: `remove_webview_js_listeners` (JS map) *and* `remove_listeners_for_label` (Rust handlers) run on window/webview destroy. They clean different maps — `remove_listeners_for_label` only touches the Rust `handlers` map, so #15604’s JS cleanup is preserved, not replaced. `cargo check`/`fmt` clean, listener tests pass.

crates/tauri/src/event/listener.rs
```diff
@@ -25,6 +25,7 @@ enum Pending {
     handler: Handler,
   },
   Emit(EmitArgs),
+  RemoveForLabel(String),
```
Legend-Master: This needs to store the event target instead

crates/tauri/src/event/listener.rs
```diff
@@ -316,6 +335,17 @@ impl Listeners {
   }
 }
 
+/// Whether the listener target is owned by the given window/webview label, meaning it
+/// should be dropped when that label is destroyed.
+fn target_bound_to_label(target: &EventTarget, label: &str) -> bool {
+  match target {
+    EventTarget::Window { label: l }
```
Legend-Master: `Window`'s labels are not always the same as `Webview`'s in multi-webview mode

Legend-Master (changes requested):

tenderdeve:
@Legend-Master addressed both, pushed in `724965882` (rebased on `dev`, signed/Verified):

1. **Store the event target, not the label** — `Pending::RemoveForLabel(String)` is now `Pending::RemoveForTarget(EventTarget)`, and the removal matches on the full `EventTarget` (kind + label) via equality instead of a bare label string.

2. **Multi-webview labels** — because matching is now exact-target, a `Window` and a `Webview` that happen to share a label are treated as distinct. `on_window_close` removes the `Window` + `WebviewWindow` targets and, per child webview, the exact `Webview` target; `on_webview_close` removes only that `Webview` target. So closing one webview no longer touches the window's (or a sibling webview's) listeners.

Dropped the old `target_bound_to_label` helper and updated the unit test to assert same-label-different-kind targets survive. `cargo test -p tauri event::listener` green (6/6).

Legend-Master (approved):

PR #15619, opened 2026-06-30
feat(cli): add `--no-binary-patching` flag to build and bundle

### LINKED ISSUE #15591: [feat] add the option to disable binary patching on bundling
### Describe the problem

Right now it is not possible to disable the binary patching that happens when bundling the app:

`Info Patching app\src-tauri\target\release\app.exe with bundle type information: nsis`

This patching makes code-signing the executable (or just bundling the already signed binary) inside the NSIS bundler impossible as this injection invalidates the certificate.

A CLI flag or config value to disable this patching.

### Alternatives considered

_No response_

### Additional context

Is this patching required for the updater to work?

tenderdeve:
Closes #15591.

Tauri patches the main executable with bundle type information (`Patching app.exe with bundle type information: nsis`) so the updater plugin can pick the matching installer format when several bundle types target the same platform. This rewrites the binary after it is built, invalidating an existing code signature on it — which makes bundling an already-signed binary inside e.g. the NSIS bundler impossible.

### Change

Add a `bundle > disableBinaryPatching` config option (default `false`, so patching stays on). When set to `true`, `bundle_project` skips the `patch_binary` call and the post-patch re-sign, leaving the main binary untouched.

I gated the re-sign together with the patch because that re-sign exists only to repair the signature the patch invalidates (the main binary is intentionally skipped by `sign_binaries_if_needed`). So with patching disabled the executable is left exactly as built, preserving any existing signature — which is the point of the flag.

As @Legend-Master noted in the issue, the patching is only needed when shipping multiple installer types per platform that should each update with their own format, so disabling it is safe otherwise.

### Files

- `tauri-utils`: new `disableBinaryPatching` field on `BundleConfig` (+ `ToTokens`, regenerated `config.schema.json`)
- `tauri-bundler`: `disable_binary_patching` on `BundleSettings` + `Settings::disable_binary_patching()`; guard in `bundle_project`
- `tauri-cli`: map the config field into `BundleSettings`

### Open question

I went with a config value (matches the rest of `bundle`). Happy to also add a `--no-sign`-style `--disable-binary-patching` CLI flag on `tauri build`/`tauri bundle` if you would prefer that too.

### DIFF
--- a/.changes/disable-binary-patching.md
+++ b/.changes/disable-binary-patching.md
@@ -1,6 +1,7 @@
 ---
 "tauri-bundler": "minor:feat"
 "tauri-cli": "minor:feat"
+"@tauri-apps/cli": "minor:feat"
 ---
 
 Add a `--no-binary-patching` flag to `tauri build` and `tauri bundle`. When set, the bundler skips patching the main executable with bundle type information (and the subsequent re-signing), leaving an already-signed binary untouched. Patching is only required when shipping multiple bundle types per platform that should each update with their own installer format.
--- a/.changes/disable-binary-patching.md
+++ b/.changes/disable-binary-patching.md
@@ -1,0 +1,7 @@
+---
+"tauri-utils": "minor:feat"
+"tauri-bundler": "minor:feat"
+"tauri-cli": "minor:feat"
+---
+
+Add a `bundle > disableBinaryPatching` config option. When enabled, the bundler skips patching the main executable with bundle type information (and the subsequent re-signing), leaving an already-signed binary untouched. Patching is only required when shipping multiple bundle types per platform that should each update with their own installer format.
--- a/.changes/disable-binary-patching.md
+++ b/.changes/disable-binary-patching.md
@@ -1,7 +1,6 @@
 ---
-"tauri-utils": "minor:feat"
 "tauri-bundler": "minor:feat"
 "tauri-cli": "minor:feat"
 ---
 
-Add a `bundle > disableBinaryPatching` config option. When enabled, the bundler skips patching the main executable with bundle type information (and the subsequent re-signing), leaving an already-signed binary untouched. Patching is only required when shipping multiple bundle types per platform that should each update with their own installer format.
+Add a `--no-binary-patching` flag to `tauri build` and `tauri bundle`. When set, the bundler skips patching the main executable with bundle type information (and the subsequent re-signing), leaving an already-signed binary untouched. Patching is only required when shipping multiple bundle types per platform that should each update with their own installer format.
--- a/crates/tauri-bundler/src/bundle.rs
+++ b/crates/tauri-bundler/src/bundle.rs
@@ -140,10 +140,6 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
     .fs_context("can't open main binary", &main_binary_path)?;
   std::io::copy(&mut main_binary_original, &mut main_binary_copy)?;
 
-  if !settings.binary_patching() {
-    log::warn!("Skipping binary patching due to --no-binary-patching flag.");
-  }
-
   let mut bundles = Vec::<Bundle>::new();
   for package_type in &package_types {
     // bundle was already built! e.g. DMG already built .app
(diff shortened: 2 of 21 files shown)


Legend-Master:
let's go with `--no-binary-patching`

tenderdeve:
@Legend-Master switched to `--no-binary-patching` as agreed — dropped the `bundle > disableBinaryPatching` config option and the `tauri-utils` change entirely. It's now a CLI flag on both `tauri build` and `tauri bundle`, wired through `Settings::set_no_binary_patching` exactly like `--no-sign`. When set, the bundler skips the `patch_binary` call and the post-patch re-sign, leaving an already-signed binary untouched.

Pushed in 8d3139dbb. Rebuilt fmt/clippy clean on `tauri-bundler` + `tauri-cli` (the one remaining clippy warning is pre-existing in `pbxproj.rs`, unrelated).

.changes/disable-binary-patching.md
```diff
@@ -0,0 +1,6 @@
+---
+"tauri-bundler": "minor:feat"
+"tauri-cli": "minor:feat"
```
Legend-Master, line 3:
```suggestion
"tauri-cli": "minor:feat"
"@tauri-apps/cli": "minor:feat"
```

crates/tauri-bundler/src/bundle/settings.rs
```diff
@@ -912,6 +915,13 @@ impl SettingsBuilder {
     self
   }
 
+  /// Sets whether to disable patching the main binary with bundle type information.
+  #[must_use]
+  pub fn no_binary_patching(mut self, no_binary_patching: bool) -> Self {
```
Legend-Master: I know we had `no_sign` above, but generally speaking, we should use positive flags (e.g. `binary_pacthing` and `true` by default)

Legend-Master (commented):
Looks good, just some small nitpicks

tenderdeve:
@Legend-Master addressed both:
- Settings now stores a positive `binary_patching: bool` defaulting to `true` (getter/setter/builder renamed to match). The `--no-binary-patching` CLI flag stays as agreed and just flips it off at the boundary via `set_binary_patching(!options.no_binary_patching)`.
- Added `@tauri-apps/cli` to the changefile.

Also signed all commits (Verified now). `cargo check` + `fmt` clean.

crates/tauri-bundler/src/bundle/settings.rs
```diff
@@ -844,12 +838,16 @@ pub struct SettingsBuilder {
...
+  binary_patching: bool,
 }
 
 impl SettingsBuilder {
   /// Creates the default settings builder.
   pub fn new() -> Self {
-    Default::default()
+    Self {
+      binary_patching: true,
+      ..Default::default()
```
Legend-Master: Let's manually implement the default instead

crates/tauri-bundler/src/bundle.rs
```diff
@@ -146,13 +145,18 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
       continue;
     }
 
-    if let Err(e) = patch_binary(&main_binary_path, package_type) {
-      log::warn!("Failed to add bundler type to the binary: {e}. Updater plugin may not be able to update this package. This shouldn't normally happen, please report it to https://github.com/tauri-apps/tauri/issues");
-    }
+    // Patching rewrites the main binary in place, invalidating any existing code signature.
+    // When disabled, leave the binary untouched and skip the post-patch re-sign (whose only
+    // purpose is to repair that invalidated signature), preserving an already-signed binary.
```
```suggestion
```

Remove this

crates/tauri-bundler/src/bundle.rs
```diff
@@ -146,13 +145,18 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
...
+      }
 
-    // sign main binary for every package type after patch
-    if matches!(target_os, TargetPlatform::Windows) && settings.windows().can_sign() {
-      windows::sign::try_sign(&main_binary_path, settings)?;
+      // sign main binary for every package type after patch
+      if matches!(target_os, TargetPlatform::Windows) && settings.windows().can_sign() {
+        windows::sign::try_sign(&main_binary_path, settings)?;
+      }
     }
```
Legend-Master, line 164: Let's add a notice like we did for `--no-sign`

.changes/disable-binary-patching.md
tenderdeve: Applied — the changefile now lists all three (`tauri-bundler`, `tauri-cli`, `@tauri-apps/cli`) as `minor:feat`.

crates/tauri-bundler/src/bundle/settings.rs
tenderdeve: Done — flipped to a positive `binary_patching` flag defaulting to `true` (the CLI flag stays `--no-binary-patching`, which just sets it to `false`).

crates/tauri-bundler/src/bundle/settings.rs
tenderdeve: Done in `a0e9f93fb` — dropped `#[derive(Default)]` and hand-wrote `impl Default for SettingsBuilder` with `binary_patching: true`; `new()` is now just `Self::default()`.

crates/tauri-bundler/src/bundle.rs
tenderdeve: Removed.

crates/tauri-bundler/src/bundle.rs
tenderdeve: Added — logs `Skipping binary patching due to --no-binary-patching flag.` Put it once before the loop rather than in a per-`package_type` else so it does not repeat for each bundle; mirrors the single `--no-sign` notice.

tenderdeve:
@Legend-Master addressed all nitpicks in `a0e9f93fb`, and signed the whole branch — all five commits now show as Verified. Thanks for the review.

Legend-Master (approved):
Thanks

PR #15620, opened 2026-06-30
feat(cli): support non-square source images in `tauri icon`

### LINKED ISSUE #15376: [feat] Support non-square images in `tauri icon`
### Describe the problem

With an option like [`object-fit`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/object-fit) to clip the image.

```
$ tauri icon brand.png --fit=cover
```

### Alternatives considered

_No response_

### Additional context

_No response_

tenderdeve:
Closes #15376.

`tauri icon` errored on any source whose width and height differ (`Source image must be square`). This adds a `--fit` option, modeled on the CSS `object-fit` property, to square a non-square source before the existing icon pipeline runs:

```
tauri icon brand.png --fit cover
```

- **`cover`** — scale to fill the square, center-cropping the longer side (may clip content)
- **`contain`** — scale to fit inside the square, padding the shorter side with transparency

SVG sources are rasterized at their native size before fitting (square SVGs keep their vector path unchanged). A non-square source without `--fit` still errors, now with a hint pointing to the flag.

### Test

- Unit tests for `fit_to_square` covering both modes (square output dimensions, center-crop stays opaque, contain padding is transparent).
- Manually ran `tauri icon` on an 80x40 PNG: no `--fit` errors with the hint; `--fit cover` and `--fit contain` both produce a full square icon set (`icon.png` 512x512, `icon.ico`, etc.).


crates/tauri-cli/src/icon.rs
```diff
@@ -233,6 +248,49 @@ fn parse_bg_color(bg_color_string: &String) -> Result<Rgba<u8>> {
   Ok(bg_color)
 }
 
+// Rasterize an SVG tree to a `DynamicImage` at its native size, preserving the aspect ratio.
+fn rasterize_svg(svg: &usvg::Tree) -> DynamicImage {
+  let width = (svg.size().width().ceil() as u32).max(1);
```
Legend-Master: This could be a problem in the later steps since some SVGs have every small `width` attributes

tenderdeve:
@Legend-Master good catch. `rasterize_svg` now scales the longest side up to a 1024px floor before rasterizing (upscale-only — sources already larger are left as-is). Since SVG is vector this is lossless, so a small `width` attribute no longer produces a tiny raster that the later icon sizes have to upscale from. Signed + `cargo check`/`fmt` clean.

crates/tauri-cli/src/icon.rs
```diff
@@ -232,6 +248,60 @@ fn parse_bg_color(bg_color_string: &String) -> Result<Rgba<u8>> {
   Ok(bg_color)
 }
 
+// Rasterize an SVG tree to a `DynamicImage`, preserving the aspect ratio.
+// An SVG can declare a tiny intrinsic size (e.g. `width="16"`), so scale the
+// longest side up to at least `MIN_SVG_RASTER_SIZE` before rasterizing to keep
+// the generated icons crisp — SVG is vector, so rendering larger is lossless.
+// Sources already larger than that are left untouched (never downscaled here).
+fn rasterize_svg(svg: &usvg::Tree) -> DynamicImage {
```
Legend-Master: Is it possible that we tweak the SVG tree instead of converting it to a `DynamicImage` here?

tenderdeve: @Legend-Master good call — reworked it exactly that way in `90f0db0eb`. The SVG now stays a `usvg::Tree` all the way through; `fit` is just recorded on the source and applied as a render transform (scale + center offset) in `resize_exact`, so every icon renders straight from the vector at its final size. No intermediate bitmap, and the `MIN_SVG_RASTER_SIZE` upscale hack is gone — a tiny intrinsic `width` no longer costs any quality. Added vector cover/contain tests; `cargo test`/`fmt` clean.

crates/tauri-cli/src/icon.rs
tenderdeve: @Legend-Master this is fully handled by the vector rework in `90f0db0eb` too — since each icon is now rendered straight from the `usvg::Tree` at its target size, a small intrinsic `width`/`height` on the source no longer feeds a tiny raster into the later steps. The declared size only sets the aspect ratio; output resolution is whatever size we render at, so there is no upscaling-from-tiny-bitmap path anymore.

Legend-Master (dismissed):
Awesome, thanks!

Legend-Master (approved):

PR #15625, opened 2026-07-01
fix(windows)!: restore webview focus for unstable feature

### LINKED ISSUE #15624: Windows accessibility regression: WebView2 content becomes unreadable and keyboard-unfocusable after the host window loses and regains activation
## Post-resolution summary

While working on the GitHub Copilot app to remediate accessibility issues, I discovered an upstream bug with how Tauri can be conditionally set to handle WebView2 content. 

This bug causes an app built with Tauri to lose the ability to be used by multiple forms of assistive technology on Windows. The scope of concern was anything and everything that uses a keyboard to be operated. The bug itself was triggered when the affected app loses focus, such as using <kbd>Alt</kbd>+<kbd>Tab</kbd> to switch applications.

[The fix](https://github.com/tauri-apps/tauri/pull/15625) replaces the bug-causing code with a `FocusState` enum recording. This detects window, a specific webview, or nothing is focused, plus the label of the last focused webview. With the fix, a window regaining focus and its saved label is read. The last active webview is then re-focused. Previously, focus had instead been stuck on an empty window frame. 

---

<details>
<summary>
Original message
</summary>
<p>Hi there,</p>
<p>I’m an accessibility designer at GitHub, working on the <a
href="https://github.com/github/app">GitHub Copilot app</a>. I’ve been
helping to remediate accessibility issues, and ran into a bug. From an
accessibility perspective, it’s a critical issue that impacts a wide
range of assistive technology applications.</p>
<p>I used the GitHub Copilot app’s LLM to try and research and fix the
bug, as Tauri/wry’s source code is far outside my technical area of
expertise. I am also cognizant of the burden LLM-generated Issues and
PRs have placed on Open Source, and have also experienced it as a
maintainer. Here, know that this is a human-researched and verified
report that was read before submitting.</p>
<p>The approach I took was LLM-guided and human evaluated, done by
building and manually testing 19 Windows development releases on real
hardware with <a href="https://www.nvaccess.org/download/">NVDA</a>, <a
href="https://vispero.com/jaws-screen-reader-software/">JAWS</a>, and <a
href="https://support.microsoft.com/en-us/accessibility/windows/voice-access/get-started-with-voice-access">Windows
Voice Access</a>, then running on-device logging and GetGUIThreadInfo
focus-state captures.</p>
<p>Unfortunately, fixing the bug seems to be a non-starter for this
approach. However, my hope is that I can share this issue as an
exhaustive and detailed list of reproducible errors encountered in my
bugfix attempts.</p>
<p>While I am posting the issue on wry’s Issue tracker, I also hope it
can serve as a centralized source of truth, as the issue intersects not
only with the Tauri/wry team, but also <a
href="https://github.com/MicrosoftEdge/WebView2Browser">Microsoft Edge
WebView2</a>, <a href="https://www.nvaccess.org/">NV Access</a>, and <a
href="https://github.com/FreedomScientific">Freedom Scientific</a>
maintainers.</p>
<p>The findings below were backed by on-device logging
(<code>tracing</code> to file), <a
href="https://learn.microsoft.com/en-us/windows/win32/winauto/inspect-objects">Inspect.exe</a>/<a
href="https://accessibilityinsights.io/docs/windows/overview/">Accessibility
Insights</a>, and GetGUIThreadInfo focus-state captures. The application
is a Tauri 2.x desktop app—Windows target uses wry + WebView2.</p>
<p>GitHub engineers working on the project are aware of this issue.
Additionally, the app source code is currently stored in a private
GitHub repository. However, I want to provide any and all details I can
past this initial report if it is helpful for addressing the issue. I’m
also more than happy to have synchronous video conversations with the
team, if that is helpful.</p>
<h2 id="table-of-contents">Table of contents</h2>
<ol type="1">
<li><a
href="https://github.com/tauri-apps/tauri/issues/15624#:~:text=that%20is%20helpful.-,Human%2Dfacing%20impact,-On%20Windows%2C%20this">Human-facing
impact</a></li>
<li><a
href="https://github.com/tauri-apps/tauri/issues/15624#:~:text=Technical%20summary%20for%20maintainers">Technical
summary for maintainers</a>
<ol type="1">
<li><a
href="https://github.com/tauri-apps/tauri/issues/15624#:~:text=summary%20for%20maintainers-,Environment">Environment</a></li>
<li><a
href="https://github.com/tauri-apps/tauri/issues/15624#:~:text=Trigger%2C%20shared%20by%20both%20failure%20modes">Trigger,
shared by both failure modes</a>
<ol type="1">
<li><a
href="https://github.com/tauri-apps/tauri/issues/15624#:~:text=Failure%20mode%20A%20%E2%80%94%20WebView2%20content%20is%20unreadable%20to%20screen%20readers%20after%20reactivation">Failure
mode A — WebView2 content is unreadable to screen readers after
reactivation</a></li>
<li><a
href="https://github.com/tauri-apps/tauri/issues/15624#:~:text=Failure%20mode%20B%20%E2%80%94%20keyboard%20Tab%20no%20longer%20enters%20the%20content%20after%20reactivation">Failure
mode B — keyboard Tab no longer enters the content after
reactivation</a></li>
</ol></li>
<li><a
href="https://github.com/tauri-apps/tauri/issues/15624#:~:text=remedy%20never%20runs.-,Interim%20mitigation,-Mitigation%20for%20failure">Interim
mitigation</a></li>
<li><a
href="https://github.com/tauri-apps/tauri/issues/15624#:~:text=Unsolved%20%E2%80%94%20failure%20mode%20B.">Unsolved
— failure mode B.</a></li>
</ol></li>
<li><a
href="https://github.com/tauri-apps/tauri/issues/15624#:~:text=Record%20of%20investigation/technical%20deep%20dive">Record
of investigation/technical deep dive</a>
<ol type="1">
<li><a
href="https://github.com/tauri-apps/tauri/issues/15624#:~:text=the%20shipped%20mitigation.-,1.%20HWND%20topology,-WebView2%20inside%20wry">1.
HWND topology</a></li>
<li><a
href="https://github.com/tauri-apps/tauri/issues/15624#:~:text=2.%20Failure%20mode%20A%20%E2%80%94%20root%20cause%20in%20depth">2.
Failure mode A — root cause in depth</a></li>
<li><a
href="https://github.com/tauri-apps/tauri/issues/15624#:~:text=3.%20Failure%20mode%20B%20%E2%80%94%20root%20cause%20in%20depth">3.
Failure mode B — root cause in depth</a></li>
<li><a
href="https://github.com/tauri-apps/tauri/issues/15624#:~:text=4.%20What%20we%20tried%20for%20B%2C%20and%20why%20naive%20fixes%20regress">4.
What we tried for B, and why naive fixes regress</a></li>
<li><a
href="https://github.com/tauri-apps/tauri/issues/15624#:~:text=5.%20Representative%20on%2Ddevice%20logs">5.
Representative on-device logs</a></li>
<li><a
href="https://github.com/tauri-apps/tauri/issues/15624#:~:text=6.%20Shipped%20mitigation%20%E2%80%94%20change%20summary">6.
Shipped mitigation — change summary</a></li>
<li><a
href="https://github.com/tauri-apps/tauri/issues/15624#:~:text=disclosure%20to%20wry.-,7.%20Summary%20of%20asks">7.
Summary of asks</a></li>
</ol></li>
<li><a
href="https://github.com/tauri-apps/tauri/issues/15624#:~:text=Research%20and%20resources">Research
and resources</a></li>
</ol>
<hr />
<h2 id="human-facing-impact">Human-facing impact</h2>
<p>On Windows, this application shows its interface inside an embedded
web view (Microsoft's WebView2, reached through the Tauri/wry windowing
layer). Doing this is normal and common—many desktop apps are built this
way.</p>
<p>There is a defect that appears after an ordinary action: The user
switches away from the app to another window and then switches back (for
example with <kbd>Alt</kbd>+<kbd>Tab</kbd>). The app window stays
visible the whole time.</p>
<p>After switching back, <strong>things break for the people who depend
on assistive technology, and they stay broken</strong> until the user
clicks inside the window with a mouse:</p>
<ol type="1">
<li><strong>Screen readers go silent</strong>. NVDA, JAWS, Narrator, and
other Windows-based screen readers can no longer read the app's content.
A blind or low-vision user hears nothing.</li>
<li><strong>Voice Access becomes inoperable</strong>. Windows Voice
Access, Dragon, and other Windows-based voice control software loses the
ability to activate interactive app content.</li>
<li><strong>Keyboard navigation stops</strong>. Pressing <kbd>Tab</kbd>
no longer moves through the interface. A keyboard-only user is
stranded.</li>
</ol>
<p>A mouse user notices nothing wrong. The breakage lands specifically
on screen-reader users, keyboard-only users, and voice control users. In
other words, people who rely on assistive technology to operate their
computer.</p>
<p>Because the app-switching trigger is something users do frequently,
the failure has a high degree of being created during a normal working
session. The issue is guaranteed to be triggered for our app use case,
as onboarding opens a browser tab to recieve an app-generated
verification code.</p>
<p>The cause is not specific to this one app. It lives in shared
platform layers: Chromium/WebView2 and the wry windowing layer that
Tauri uses. The same family of problems has been reported against other
Chromium-based software, including Electron apps and VS Code (listed at
the end of this post as research references).</p>
<hr />
<h2 id="technical-summary-for-maintainers">Technical summary for
maintainers</h2>
<h3 id="environment">Environment</h3>
<ul>
<li>Windows 11 (x64).</li>
<li>Microsoft Edge WebView2 Runtime in the recent range where the UI
Automation (UIA) provider is on by default. Rollout began ~Chrome 138;
assistive-tech regressions reported across Chromium 140–142. The tested
environment was WebView2/Edge 149 (Chromium 141).</li>
<li>Tauri 2.x; wry 0.55.1.</li>
<li>Standard decorated window (<code>decorations: true</code>). This is
not the frameless <code>decorations: false</code> case from <a
href="https://github.com/tauri-apps/tauri/issues/12901">https://github.com/tauri-apps/tauri/issues/12901</a>,
though it is in the same problem family.</li>
<li>Assistive tech: NVDA 2025.x, JAWS 2026, Windows Narrator, Windows
Voice Access.</li>
</ul>
<h3 id="trigger-shared-by-both-failure-modes">Trigger, shared by both
failure modes</h3>
<p>The host top-level window loses, then regains, <strong>foreground
activation</strong>: <kbd>Alt</kbd>+<kbd>Tab</kbd> away and back, or
click another window and click back. The window remains visible
throughout—not an occlusion issue.</p>
<p>After reactivation the app is visually and mouse-interactive but
degraded for AT and keyboard, until the user clicks into the web
content. As noted previously, some keyboard-only users cannot perform
this workaround.</p>
<h4
id="failure-mode-a--webview2-content-is-unreadable-to-screen-readers-after-reactivation">Failure
mode A — WebView2 content is unreadable to screen readers after
reactivation</h4>
<p>Recent Chromium turned its Windows UIA provider on by default. For a
WebView2 control embedded inside a <strong>non-Chrome host</strong>
(here, wry's window), the content's UIA <em>fragment root</em> reports
its <strong>parent</strong> as the host's container handle window
(HWND), which has no UIA provider of its own.</p>
<p>That is governed by the Chromium feature
<code>kAccessibilityWinAXFragmentRootParent</code>, which is
<strong>enabled by default</strong>. The result is an <em>orphaned</em>
fragment root: while the app is foreground, UIA reaches the content
through its focus-based path. While the app is backgrounded, UIA must
walk the HWND tree top-down, and that walk dead-ends at the
provider-less container. The content provider is still alive. However,
it is unreachable by top-down traversal.</p>
<p>NVDA defaults to the legacy IAccessible2/MSAA path for Chromium-class
windows. Voice Access is UIA-only. Both go silent on reactivation. Of
note, NVDA can be set to exclusively use UIA, but doing so does not
address failure mode B.</p>
<h4
id="failure-mode-b--keyboard-tab-no-longer-enters-the-content-after-reactivation">Failure
mode B — keyboard Tab no longer enters the content after
reactivation</h4>
<p>wry restores the WebView2 controller's internal focus by calling
<code>ICoreWebView2Controller::MoveFocus(PROGRAMMATIC)</code>, but only
from the top-level window's <code>WM_SETFOCUS</code> handler (the
<code>WM_SETFOCUS | WM_ENTERSIZEMOVE</code> arm of
<code>parent_subclass_proc</code>).</p>
<p>There is no <code>WM_ACTIVATE</code> handler. On
<kbd>Alt</kbd>+<kbd>Tab</kbd>-back, Windows restores focus
<strong>directly to the WebView2 child wrapper</strong>
(<code>Chrome_WidgetWin_1</code>), so the top-level's
<code>WM_SETFOCUS</code> never fires, <code>MoveFocus</code> is never
called, and the controller's keyboard/tab entry point stays stale.
<kbd>Tab</kbd> does nothing until a click re-seeds focus.</p>
<p>This was confirmed on-device with <code>GetGUIThreadInfo</code>:
After reactivation the foreground and active window is the
<code>Tauri Window</code> top-level, but the focus window is the
<code>Chrome_WidgetWin_1</code> child — the top-level never received
focus, so wry's existing remedy never runs.</p>
<h3 id="interim-mitigation">Interim mitigation</h3>
<p>Mitigation for failure mode A only. For our app, we:</p>
<ol type="1">
<li>Forced the legacy IA2/MSAA path by appending
<code>UiaProvider</code> to the WebView2 <code>--disable-features</code>
list. This restores NVDA/JAWS reading because the legacy path
auto-bridges child HWNDs;</li>
<li>Re-post the focus event that WebView2 drops on reactivation, via
NotifyWinEvent(<code>EVENT_OBJECT_FOCUS</code>, ...) on the web-content
HWND, so the screen reader's reading position is re-seeded.</li>
</ol>
<p>This is explicitly <strong>a mitigation, not a fix</strong>.
Disabling the UIA provider degrades UIA-only assistive tech (Voice
Access), and re-posting the event papers over the fact that it is
dropped. The underlying UIA traversal defect is untouched.</p>
<h3 id="unsolved--failure-mode-b">Unsolved — failure mode B.</h3>
<p>The correct fix needs to sequence the activation transition with the
WebView2 controller's focus, which requires the controller handle and
the window-message timing sourced from wry.</p>
<p>App-side attempts to reproduce wry's <code>MoveFocus</code> from the
activation path regressed: One produced a ~140 Hz focus loop, another a
~200 Hz loop that also broke <kbd>Tab</kbd> on a fresh launch.</p>
<hr />
<h2 id="record-of-investigationtechnical-deep-dive">Record of
investigation/technical deep dive</h2>
<p>This section was authored with instruction to be LLM-friendly for
content ingestion, in the hope it can serve future remediation efforts.
It contains: HWND topology, both root causes with source citations, the
decisive on-device evidence, the app-side approaches that were tried and
why they regress, representative diagnostic logs, and a precise summary
of the shipped mitigation.</p>
<h3 id="1-hwnd-topology">1. HWND topology</h3>
<p>WebView2 inside wry produces this parent → child chain. Only the two
lowest HWNDs carry real accessibility providers; the two host HWNDs do
not answer <code>WM_GETOBJECT</code>.</p>
<pre><code>graph TD
  A[&quot;Host top-level HWND&lt;br/&gt;class: &#39;Tauri Window&#39; (tao)&lt;br/&gt;no WM_GETOBJECT / no UIA provider&quot;]
  B[&quot;Host WebView container HWND&lt;br/&gt;class: &#39;Chrome_WidgetWin_1&#39; (wry WRY_WEBVIEW)&lt;br/&gt;no UIA provider — fragment root&#39;s CLAIMED parent&quot;]
  C[&quot;WebView2 host HWND&lt;br/&gt;class: &#39;Chrome_WidgetWin_0&#39; (msedgewebview2.exe)&lt;br/&gt;answers WM_GETOBJECT&quot;]
  D[&quot;Render widget HWND&lt;br/&gt;class: &#39;Chrome_RenderWidgetHostHWND&#39;&lt;br/&gt;LIVE web-content provider (UIA + IA2)&quot;]
  A --&gt; B --&gt; C --&gt; D</code></pre>
<p>Chromium points the content fragment root's UIA parent at B, but B
has no provider, so the top-down UIA walk from A cannot descend past B
while the app is backgrounded.</p>
<h3 id="2-failure-mode-a--root-cause-in-depth">2. Failure mode A — root
cause in depth</h3>
<p><strong>The UIA-by-default rollout</strong>. Recent Chromium enables
its Windows UIA provider by default (background: Chrome Developers,
"Native UI Automation for Windows in Chromium"). This moved both NVDA
and Voice Access onto the UIA path for WebView2 content, and the UIA
path's top-down traversal is the one that breaks when backgrounded.</p>
<p>The same rollout produced a cluster of NVDA regressions across the
ecosystem:</p>
<ul>
<li><a
href="https://github.com/nvaccess/nvda/issues/19276">https://github.com/nvaccess/nvda/issues/19276</a></li>
<li><a
href="https://github.com/nvaccess/nvda/issues/18239">https://github.com/nvaccess/nvda/issues/18239</a></li>
<li><a
href="https://github.com/nvaccess/nvda/issues/18707">https://github.com/nvaccess/nvda/issues/18707</a></li>
<li><a
href="https://github.com/nvaccess/nvda/issues/18996">https://github.com/nvaccess/nvda/issues/18996</a></li>
<li><a
href="https://github.com/microsoft/vscode/issues/277895">https://github.com/microsoft/vscode/issues/277895</a></li>
<li><a
href="https://issues.chromium.org/issues/423432437">https://issues.chromium.org/issues/423432437</a></li>
</ul>
<p><strong>The orphaned fragment root</strong>. In Chromium,
<code>HWNDMessageHandler::GetParentOfAXFragmentRoot</code> decides the
UIA parent of a web-content fragment root. Its behavior is gated by the
feature <code>kAccessibilityWinAXFragmentRootParent</code>, defined in
<code>ui/accessibility/accessibility_features.cc</code>:</p>
<pre><code>BASE_FEATURE(kAccessibilityWinAXFragmentRootParent,
             &quot;AccessibilityWinAXFragmentRootParent&quot;,
             base::FEATURE_ENABLED_BY_DEFAULT);</code></pre>
<p>When enabled, the default <code>GetParentOfAXFragmentRoot</code>
delegates to GetParentNativeViewAccessible(), resolving the parent to
the host's container HWND. For a Chrome browser window that container is
a <code>Chrome_WidgetWin_1</code> that does answer
<code>WM_GETOBJECT</code>, so the chain is consistent.</p>
<p>For a WebView2 embedded in wry, the corresponding container (HWND B
above) has <strong>no UIA provider</strong>, so the fragment root claims
a parent that exposes nothing back down to it: orphaned. The legacy
path, with the feature disabled, returns <code>nullptr</code> and
self-roots the fragment — but it is a killswitch slated for removal, so
it is not a durable lever.</p>
<p><strong>Foreground works, background fails</strong> — two different
discovery paths. UIA has two ways to reach an element:</p>
<ol type="1">
<li><strong>Focus-based</strong>
(<code>IUIAutomation::GetFocusedElement</code>) — used while the app is
foreground. The focused element is inside the WebView2, so UIA reaches
it directly, bypassing the host HWND chain. This path is not
broken.</li>
<li><strong>HWND-chain traversal</strong>
(<code>ElementFromHandle(top_hwnd</code>) then walk children) — the only
path available when the app is backgrounded. This is the path that
dead-ends at the provider-less container.</li>
</ol>
<p><strong>Decisive evidence (on device, app deactivated, window still
visible):</strong></p>
<ul>
<li>Inspect.exe pointed directly at
<code>Chrome_RenderWidgetHostHWND</code> (HWND D) shows <strong>the full
web-content tree</strong> while the app is backgrounded ⇒ the provider
is <strong>alive</strong>; only top-down traversal is broken.</li>
<li>Inspect.exe starting at the <strong>top-level</strong> and
navigating <strong>down</strong> cannot enter the content while
backgrounded ⇒ the traversal dead-ends, exactly at the container.</li>
<li>Accessibility Insights <strong>hover</strong> (a hit-test, which
routes to D's own provider) reaches content nodes; top-down object
navigation does not.</li>
<li>NVDA's "Use UIA with Microsoft Edge and other Chromium based
browsers = Yes" does not fix it ⇒ this is not the NVDA window-class
opt-in problem and not the mouse-tracking-only regression; it is the
structural orphaned-fragment-root break.</li>
<li>It reproduces on <strong>pure deactivation</strong>, not just
minimize/occlusion. This confirms the broken path is HWND-chain
traversal rather than renderer teardown.</li>
</ul>
<p><strong>Why an app-side UIA bridge is the wrong layer</strong>. We
prototyped grafting a host UIA provider onto container B (subclass B,
answer <code>WM_GETOBJECT</code> for <code>UiaRootObjectId</code> with
<code>UiaHostProviderFromHwnd</code>). On-device this proved the OS
<em>does</em> query B top-down (219
<code>WM_GETOBJECT(UiaRootObjectId</code>) hits during a backgrounded
NVDA pass) and we did answer— but a bare host provider is a leaf: it
exposes no fragment child pointing at the cross-process WebView2
subtree, so top-down descent still dead-ends.</p>
<p>Building a real <code>IRawElementProviderFragmentRoot</code> on the
host is not feasible from the host process: the reference
implementations get their child as an <strong>in-process</strong>
<code>NativeViewAccessible</code> (Chromium's
<code>AXFragmentRootWin</code> via
<code>delegate_-&gt;GetChildOfAXFragmentRoot()</code>; AccessKit's
Windows adapter builds its <em>own</em> in-process tree). The WebView2
content lives in <code>msedgewebview2.exe</code> — a different process —
and there is no documented server-side API to turn a child HWND into an
<code>IRawElementProviderFragment</code> for
<code>Navigate(FirstChild)</code>. The host cannot manufacture the
missing parent link for a cross-process child. This is the long-standing
gap tracked by <a
href="https://github.com/MicrosoftEdge/WebView2Feedback/issues/403">https://github.com/MicrosoftEdge/WebView2Feedback/issues/403</a>.</p>
<p><strong>Conclusion: the UIA fix belongs in wry/WebView2, not the host
app</strong>. We removed the ~270-line bridge.</p>
<p><strong>Why forcing IA2/MSAA reads (the shipped mitigation for
A)</strong>. On the legacy path, <code>DefWindowProc</code>/OLEACC
synthesize a default <code>IAccessible</code> for a window that
<strong>enumerates its child HWNDs</strong>, so the WebView2 children
are auto-bridged through the ordinary HWND hierarchy — the
auto-stitching that the UIA server-side model lacks. Appending
<code>UiaProvider</code> to WebView2's <code>--disable-features</code>
forces this path, restoring NVDA reading.</p>
<p>The documented Chromium-wide form of this workaround is
<code>--disable-features=UiaProvider</code>. The trade-off is
unavoidable: Voice Access is <strong>UIA-only</strong>, so forcing IA2
cannot satisfy it. Only fixing the UIA traversal helps every AT at
once.</p>
<p><strong>The residual A gap even on IA2 — the dropped focus
event</strong>. With reading restored, NVDA still went silent
specifically on reactivation until a manual focus move. Cause:
Chromium/WebView2 raises <code>EVENT_OBJECT_FOCUS</code> for in-page DOM
focus changes, not when the host window is reactivated. AT sets its
reading position from that event, so on
<kbd>Alt</kbd>+<kbd>Tab</kbd>-back the position is stale. The documented
host-side remedy is to re-post the event; the mitigation calls
<code>NotifyWinEvent(EVENT_OBJECT_FOCUS, web_content_hwnd, OBJID_CLIENT, CHILDID_SELF)</code>
on reactivation. This is the second workaround stacked into the A
mitigation.</p>
<h3 id="3-failure-mode-b--root-cause-in-depth">3. Failure mode B — root
cause in depth</h3>
<p>wry's top-level subclass restores WebView2 focus only on
<code>WM_SETFOCUS</code>. From wry 0.55.1,
<code>src/webview2/mod.rs</code>, <code>parent_subclass_proc</code>:</p>
<div class="sourceCode" id="cb3"><pre class="sourceCode rs"><code class="sourceCode rust"><span id="cb3-1"><a href="#cb3-1" aria-hidden="true" tabindex="-1"></a><span class="co">// parent_subclass_proc (wry 0.55.1, src/webview2/mod.rs:1215)</span></span>
<span id="cb3-2"><a href="#cb3-2" aria-hidden="true" tabindex="-1"></a><span class="cf">match</span> msg <span class="op">{</span></span>
<span id="cb3-3"><a href="#cb3-3" aria-hidden="true" tabindex="-1"></a>  <span class="co">// ...</span></span>
<span id="cb3-4"><a href="#cb3-4" aria-hidden="true" tabindex="-1"></a>  WM_SETFOCUS <span class="op">|</span> WM_ENTERSIZEMOVE <span class="op">=&gt;</span> <span class="op">{</span>                              <span class="co">// :1254</span></span>
<span id="cb3-5"><a href="#cb3-5" aria-hidden="true" tabindex="-1"></a>    <span class="kw">let</span> controller <span class="op">=</span> dwrefdata <span class="kw">as</span> <span class="op">*</span><span class="kw">mut</span> ICoreWebView2Controller<span class="op">;</span></span>
<span id="cb3-6"><a href="#cb3-6" aria-hidden="true" tabindex="-1"></a>    <span class="kw">let</span> _ <span class="op">=</span> (<span class="op">*</span>controller)<span class="op">.</span>MoveFocus(COREWEBVIEW2_MOVE_FOCUS_REASON_PROGRAMMATIC)<span class="op">;</span> <span class="co">// :1256</span></span>
<span id="cb3-7"><a href="#cb3-7" aria-hidden="true" tabindex="-1"></a>  <span class="op">}</span></span>
<span id="cb3-8"><a href="#cb3-8" aria-hidden="true" tabindex="-1"></a>  <span class="co">// ... no WM_ACTIVATE arm anywhere in this proc</span></span>
<span id="cb3-9"><a href="#cb3-9" aria-hidden="true" tabindex="-1"></a><span class="op">}</span></span></code></pre></div>
<p>There is <strong>no <code>WM_ACTIVATE</code> handler</strong> in this
proc. On a fresh launch, the top-level receives
<code>WM_SETFOCUS</code>, <code>MoveFocus(PROGRAMMATIC)</code> runs, the
controller's tab entry is seeded, and <kbd>Tab</kbd> works. On
<kbd>Alt</kbd>+<kbd>Tab</kbd>-back, Windows restores focus directly to
the child wrapper, the top-level's <code>WM_SETFOCUS</code> never fires,
MoveFocus is never called, and the controller's tab entry stays
stale.</p>
<pre><code>sequenceDiagram
  participant U as User
  participant Win as Windows
  participant Top as tao top-level (&#39;Tauri Window&#39;)
  participant Ctl as wry/WebView2 controller

  Note over U,Ctl: Fresh launch (works)
  Win-&gt;&gt;Top: WM_SETFOCUS
  Top-&gt;&gt;Ctl: MoveFocus(PROGRAMMATIC)
  Note over Ctl: tab entry seeded -&gt; Tab navigates content

  Note over U,Ctl: Alt+Tab away, then back (broken)
  U-&gt;&gt;Win: Alt+Tab back to app
  Win-&gt;&gt;Top: WM_ACTIVATE (WA_ACTIVE)
  Win--&gt;&gt;Ctl: focus restored DIRECTLY to child wrapper (Chrome_WidgetWin_1)
  Note over Top: top-level WM_SETFOCUS never fires
  Note over Ctl: MoveFocus never called -&gt; tab entry STALE -&gt; Tab dead</code></pre>
<p><strong>On-device confirmation (GetGUIThreadInfo, immediately after
<kbd>Alt</kbd>+<kbd>Tab</kbd>-back):</strong></p>
<pre><code>focus state on activation label=&quot;main&quot; gui_thread_info_ok=true
  foreground_hwnd=657372 foreground_class=&quot;Tauri Window&quot;
  focus_hwnd=264112     focus_class=&quot;Chrome_WidgetWin_1&quot;
  active_hwnd=657372    active_class=&quot;Tauri Window&quot;
  capture_hwnd=0</code></pre>
<p>Foreground and active are the top-level; focus is the child wrapper.
The top-level did not receive focus, so wry's <code>WM_SETFOCUS</code>
arm could not have run.</p>
<h3 id="4-what-we-tried-for-b-and-why-naive-fixes-regress">4. What we
tried for B, and why naive fixes regress</h3>
<p>The hard finding is that the obvious app-side fixes do
<strong>not</strong> work, they regress focus. This is the main reason B
needs an upstream fix rather than a host workaround.</p>
<ol type="1">
<li><strong>Host <code>SetFocus(child container)</code> from the
activation hook → ~140 Hz infinite loop</strong>. Setting focus to the
<code>WRY_WEBVIEW</code> container from the app's
<code>Focused(true)</code> handler caused the in-process wrapper to
forward focus back up to the tao top-level, which re-raised
<code>Focused(true)</code>, which called <code>SetFocus</code> again.
On-device logs showed ~1500 focus events, one every ~7 ms.
Reverted.</li>
<li><strong>Host
<code>ICoreWebView2Controller::MoveFocus(PROGRAMMATIC)</code> from the
activation hook → ~200 Hz loop <em>and</em> broke Tab on a fresh
launch</strong>. Tauri exposes the controller via
<code>WebviewWindow::with_webview(|w| w.controller())</code>, so the app
can call the same API wry calls. But calling it from the activation path
(not from a genuine <code>WM_SETFOCUS</code>) made the in-process
wrapper bounce focus back to the top-level, re-firing the hook (~10
calls in 48 ms). Worse, even a single fire from this path <strong>broke
<kbd>Tab</kbd> on launch</strong> with no <kbd>Alt</kbd>+<kbd>Tab</kbd>
involved — MoveFocus issued at activation time, rather than at real
focus-set time, scrambles the controller's tab entry point. Not
salvageable with a debounce. Reverted.</li>
<li><strong>wry fork adding a
<code>WM_ACTIVATE → MoveFocus(PROGRAMMATIC)</code> arm → unconfirmable
on-device</strong>. We forked wry and added the missing arm. The patched
arm produced <strong>no observable log output</strong> across multiple
signed builds, while the same-commit app-side diagnostic logged every
activation — i.e. we could not confirm the forked arm executed in the
shipped binary, so we cannot claim it as a fix. Not shipped.</li>
</ol>
<p><strong>Implication for a wry-side fix</strong>. A naive
<code>WM_ACTIVATE → MoveFocus</code> arm is not obviously safe: the
host-side equivalent both <strong>looped</strong> (wrapper bounce
re-entrancy) and <strong>broke <kbd>Tab</kbd> on launch</strong>
(<code>MoveFocus</code> issued off the activation edge rather than the
focus-set edge).</p>
<p>A correct fix in wry likely needs to (a) act only on a genuine
activation edge (<code>WA_ACTIVE</code>/<code>WA_CLICKACTIVE</code>,
ignoring the focus-bounce), with re-entrancy guarding, and/or (b) route
the activation into the existing, proven <code>WM_SETFOCUS</code> arm
(for example, on <code>WM_ACTIVATE</code> call
<code>SetFocus(top_level_hwnd)</code> so Windows then issues the real
<code>WM_SETFOCUS</code> that wry already handles correctly) rather than
calling <code>MoveFocus</code> directly off <code>WM_ACTIVATE</code>.
Because wry owns both the controller handle and the window-message
timing, this sequencing is feasible in wry and not safely reproducible
in the host app.</p>
<h3 id="5-representative-on-device-logs">5. Representative on-device
logs</h3>
<p>App-side announce (failure mode A mitigation firing on
reactivation):</p>
<pre><code>INFO github_app::platform::webview_a11y: announced EVENT_OBJECT_FOCUS on
  web-content HWND after activation label=&quot;main&quot; content=395192</code></pre>
<p>Focus-state diagnostic (failure mode B, same reactivation):</p>
<pre><code>INFO github_app::platform::webview_a11y: focus state on activation label=&quot;main&quot;
  gui_thread_info_ok=true
  foreground_hwnd=657372 foreground_class=&quot;Tauri Window&quot;
  focus_hwnd=264112      focus_class=&quot;Chrome_WidgetWin_1&quot;
  active_hwnd=657372     active_class=&quot;Tauri Window&quot;
  capture_hwnd=0</code></pre>
<p>Host UIA-bridge probe (the abandoned app-side UIA approach; proves
the OS queries the container top-down but a bare host provider is
insufficient):</p>
<pre><code>INFO ... WM_GETOBJECT UiaRootObjectId on WRY_WEBVIEW; returning host provider
  hwnd=984950 objid=-25     (observed 219x during one backgrounded NVDA pass)</code></pre>
<h3 id="6-shipped-mitigation--change-summary">6. Shipped mitigation —
change summary</h3>
<p>Footprint versus <code>main</code>: <strong>5 files, +164 /
-2</strong>. Summarized here; the full module is omitted (application
source). The only code shown is generic, documented Win32 / Chromium
surface.</p>
<ul>
<li><strong><code>src-tauri/tauri.conf.json</code></strong> — append
<code>UiaProvider</code> to the WebView2 <code>--disable-features</code>
list (merged with wry's existing defaults), forcing the legacy IA2/MSAA
path:</li>
</ul>
<pre><code>--disable-features=msWebOOUI,msPdfOOUI,msSmartScreenProtection,CalculateNativeWinOcclusion,UiaProvider --force-renderer-accessibility</code></pre>
<ul>
<li><strong><code>src-tauri/src/platform/webview_a11y.rs</code></strong>
(+155) — a Windows-only module that, on host reactivation, finds the
<code>Chrome_RenderWidgetHostHWND</code> descendant via
<code>EnumChildWindows</code> + <code>GetClassNameW</code> and re-posts
the dropped focus event. The load-bearing call (documented Win32 API
usage):</li>
</ul>
<pre><code>// generic Win32 usage; constants from the windows crate
NotifyWinEvent(EVENT_OBJECT_FOCUS, web_content_hwnd, OBJID_CLIENT.0, CHILDID_SELF as i32);</code></pre>
<ul>
<li><strong><code>src-tauri/src/lib.rs</code></strong> (+5) — in the
existing <code>WindowEvent::Focused(true)</code> handler for the main
window (Windows-only), call the announce helper.</li>
<li><strong><code>src-tauri/src/platform/mod.rs</code></strong> (+2) —
register the module.</li>
<li><strong><code>src-tauri/Cargo.toml</code></strong> (+1) — add the
<code>Win32_UI_Accessibility</code> feature to the <code>windows</code>
crate.</li>
</ul>
<p><strong>Removed across the investigation</strong> (so
reviewers/maintainers know the dead ends):</p>
<ul>
<li>The ~270-line app-side UIA <strong>host bridge</strong>
(<code>UiaHostProviderFromHwnd</code> grafting on the container) —
abandoned as the wrong layer ;</li>
<li>The <strong><code>GetGUIThreadInfo</code> focus-state
diagnostics</strong> (kept only as evidence here);</li>
<li>The <strong><code>SetFocus</code> / <code>MoveFocus</code>
activation-hook loops</strong> — reverted.</li>
</ul>
<p>The mitigation addresses <strong>A</strong> (reading) only.
<strong>B</strong> (keyboard <kbd>Tab</kbd>) is intentionally not
patched app-side and is the subject of this disclosure to wry.</p>
<h3 id="7-summary-of-asks">7. Summary of asks</h3>
<ul>
<li><strong>wry/WebView2 (failure mode B)</strong>: restore the WebView2
controller's keyboard/tab focus across a host
<strong>activation</strong> transition, not only on top-level
<code>WM_SETFOCUS</code> — guarding against the wrapper focus-bounce
re-entrancy described in the “What we tried for B, and why naive fixes
regress” subsection. This is the primary upstream ask.</li>
<li><strong>WebView2 / Chromium (failure mode A)</strong>: for a
WebView2 embedded in a non-Chrome host, keep the content UIA fragment
root reachable by top-down traversal while the host is backgrounded —
either by not orphaning it against a provider-less host container
(<code>kAccessibilityWinAXFragmentRootParent</code> behavior for
non-Chrome hosts) or by exposing a documented host UIA-registration
mechanism (the gap in <a
href="https://github.com/MicrosoftEdge/WebView2Feedback/issues/403">https://github.com/MicrosoftEdge/WebView2Feedback/issues/403</a>).
Secondarily, raise <code>EVENT_OBJECT_FOCUS</code> on host reactivation
so AT reading position is not stale.</li>
</ul>
<hr />
<h2 id="research-and-resources">Research and resources</h2>
<p>All links used across this investigation:</p>
<h3 id="upstream-issue-trackers">Upstream issue trackers</h3>
<ul>
<li>Tauri — frameless NVDA a11y bug (same family): <a
href="https://github.com/tauri-apps/tauri/issues/12901">https://github.com/tauri-apps/tauri/issues/12901</a></li>
<li>Electron — NVDA broken since 30.0.0-beta.8 (cross-framework analog):
<a
href="https://github.com/electron/electron/issues/42945">https://github.com/electron/electron/issues/42945</a></li>
<li>VS Code — NVDA mouse tracking broken by Chromium 142: <a
href="https://github.com/microsoft/vscode/issues/277895">https://github.com/microsoft/vscode/issues/277895</a></li>
<li>VS Code — NVDA mouse-tracking / Chromium UIA regression: <a
href="https://github.com/microsoft/vscode/issues/224704">https://github.com/microsoft/vscode/issues/224704</a></li>
<li>WebView2Feedback — "Support IAccessible and WM_GETOBJECT" (host
UIA-registration gap, open since 2020): <a
href="https://github.com/MicrosoftEdge/WebView2Feedback/issues/403">https://github.com/MicrosoftEdge/WebView2Feedback/issues/403</a></li>
<li>NVDA — UIA enabled by default in Chromium 142+: <a
href="https://github.com/nvaccess/nvda/issues/19276">https://github.com/nvaccess/nvda/issues/19276</a></li>
<li>NVDA — freezes typing in Chrome omnibox with UIA: <a
href="https://github.com/nvaccess/nvda/issues/18239">https://github.com/nvaccess/nvda/issues/18239</a></li>
<li>NVDA — Chromium mouse-tracking regression: <a
href="https://github.com/nvaccess/nvda/issues/18707">https://github.com/nvaccess/nvda/issues/18707</a></li>
<li>NVDA — Chromium mouse-tracking regression (related): <a
href="https://github.com/nvaccess/nvda/issues/18996">https://github.com/nvaccess/nvda/issues/18996</a></li>
</ul>
<h3 id="chromium--webview2-platform">Chromium / WebView2 platform</h3>
<ul>
<li>Chromium bug — Chromium UIA regression: <a
href="https://issues.chromium.org/issues/423432437">https://issues.chromium.org/issues/423432437</a></li>
<li>Chromium source — <code>kAccessibilityWinAXFragmentRootParent</code>
(<code>FEATURE_ENABLED_BY_DEFAULT</code>),
<code>ui/accessibility/accessibility_features.cc</code>: <a
href="https://source.chromium.org/chromium/chromium/src/+/main:ui/accessibility/accessibility_features.cc">https://source.chromium.org/chromium/chromium/src/+/main:ui/accessibility/accessibility_features.cc</a></li>
<li>Chromium source — <code>AXFragmentRootWin</code> (reference
fragment-root host; in-process child only): <a
href="https://source.chromium.org/chromium/chromium/src/+/main:ui/accessibility/platform/ax_fragment_root_win.h">https://source.chromium.org/chromium/chromium/src/+/main:ui/accessibility/platform/ax_fragment_root_win.h</a>
and <a
href="https://source.chromium.org/chromium/chromium/src/+/main:ui/accessibility/platform/ax_fragment_root_win.cc">https://source.chromium.org/chromium/chromium/src/+/main:ui/accessibility/platform/ax_fragment_root_win.cc</a></li>
<li>Chrome Developers — "Native UI Automation for Windows in Chromium"
(UIA on by default; mitigations): <a
href="https://developer.chrome.com/blog/windows-uia-support-update">https://developer.chrome.com/blog/windows-uia-support-update</a></li>
</ul>
<h3 id="wry-source-failure-mode-b">wry source (failure mode B)</h3>
<ul>
<li>wry 0.55.1 — <code>parent_subclass_proc</code>: <a
href="https://github.com/tauri-apps/wry/blob/wry-v0.55.1/src/webview2/mod.rs#L1215">https://github.com/tauri-apps/wry/blob/wry-v0.55.1/src/webview2/mod.rs#L1215</a></li>
<li>wry 0.55.1 — <code>WM_SETFOCUS | WM_ENTERSIZEMOVE</code> arm (the
only <code>MoveFocus</code> site; no <code>WM_ACTIVATE</code>): <a
href="https://github.com/tauri-apps/wry/blob/wry-v0.55.1/src/webview2/mod.rs#L1254-L1257">https://github.com/tauri-apps/wry/blob/wry-v0.55.1/src/webview2/mod.rs#L1254-L1257</a></li>
</ul>
<h3
id="accesskit-reference-for-the-uia-fragment-root-interface-shape">AccessKit
(reference for the UIA fragment-root interface shape)</h3>
<ul>
<li>AccessKit Windows adapter — node provider: <a
href="https://github.com/AccessKit/accesskit/blob/main/platforms/windows/src/node.rs">https://github.com/AccessKit/accesskit/blob/main/platforms/windows/src/node.rs</a></li>
<li>AccessKit Windows adapter — subclassing: <a
href="https://github.com/AccessKit/accesskit/blob/main/platforms/windows/src/subclass.rs">https://github.com/AccessKit/accesskit/blob/main/platforms/windows/src/subclass.rs</a></li>
</ul>
<h3 id="microsoft-api-documentation">Microsoft API documentation</h3>
<ul>
<li><code>ICoreWebView2Controller::MoveFocus</code>: <a
href="https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2controller#movefocus">https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2controller#movefocus</a></li>
<li><code>COREWEBVIEW2_MOVE_FOCUS_REASON</code>: <a
href="https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2controller#corewebview2_move_focus_reason">https://learn.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2controller#corewebview2_move_focus_reason</a></li>
<li><code>UiaHostProviderFromHwnd</code>: <a
href="https://learn.microsoft.com/en-us/windows/win32/api/uiautomationcoreapi/nf-uiautomationcoreapi-uiahostproviderfromhwnd">https://learn.microsoft.com/en-us/windows/win32/api/uiautomationcoreapi/nf-uiautomationcoreapi-uiahostproviderfromhwnd</a></li>
<li><code>NotifyWinEvent</code>: <a
href="https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-notifywinevent">https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-notifywinevent</a></li>
<li><code>EnumChildWindows</code>: <a
href="https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-enumchildwindows">https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-enumchildwindows</a></li>
<li><code>GetGUIThreadInfo</code>: <a
href="https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getguithreadinfo">https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getguithreadinfo</a></li>
</ul>
<h3 id="forks-used-during-investigation">Forks used during
investigation</h3>
<ul>
<li>wry fork (WM_ACTIVATE focus experiment): <a
href="https://github.com/ericwbailey/wry">https://github.com/ericwbailey/wry</a></li>
<li>tao fork (pre-existing Windows patch carried by the app): <a
href="https://github.com/chuwik/tao">https://github.com/chuwik/tao</a></li>
</ul>
<h3 id="diagnostic-tooling">Diagnostic tooling</h3>
<ul>
<li>Sysinternals Process Explorer (verify the WebView2 process command
line): <a
href="https://learn.microsoft.com/sysinternals/downloads/process-explorer">https://learn.microsoft.com/sysinternals/downloads/process-explorer</a></li>
<li>Accessibility Insights for Windows (UIA tree inspection): <a
href="https://accessibilityinsights.io/downloads/">https://accessibilityinsights.io/downloads/</a></li>
</ul>
</details>

### LINKED ISSUE #15626: [bug] moving the window causes focus event to emit false and then true on Windows
### Describe the bug

Dragging the window around through title bar causes focus event to emit false and then true on Windows

### Reproduction

Dragging the window around and add this to the API example in this repo

```diff
diff --git a/examples/api/src-tauri/src/lib.rs b/examples/api/src-tauri/src/lib.rs
index fb3b75b4e..546df2bc9 100644
--- a/examples/api/src-tauri/src/lib.rs
+++ b/examples/api/src-tauri/src/lib.rs
@@ -184,6 +184,12 @@ pub fn run_app<R: Runtime, F: FnOnce(&App<R>) + Send + 'static>(
           .destroy()
           .unwrap();
       }
+      RunEvent::WindowEvent {
+        event: tauri::WindowEvent::Focused(focused),
+        ..
+      } => {
+        dbg!(focused);
+      }
       #[cfg(target_os = "ios")]
       RunEvent::SceneRequested { .. } => {
         counter += 1;

```

### Expected behavior

No events emitted

### Full `tauri info` output

Current head: ca90b46b2e2cbbc981dae1b809f4af4343fe0558

Legend-Master:
Fix #15624
Fix #15626
Closes https://github.com/tauri-apps/wry/pull/1755

- Fix webview don't get focus when Alt-Tab back to the window if `unstable` feature is enabled on Windows
- Fix `WindowEvent::Focused` events emitted when dragging the window on Windows
- `CreateWebviewOptions::focused_webview` now takes `Arc<Mutex<FocusState>>` instead of `Arc<Mutex<Option<String>>>`


crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -4677,7 +4707,12 @@ fn create_window<T: UserEvent, F: Fn(RawWindow) + Send + 'static>(
 
   let mut webviews = Vec::new();
 
-  let focused_webview = Arc::new(Mutex::new(None));
+  let focused_webview = Arc::new(Mutex::new(FocusState::default()));
+
+  #[cfg(feature = "unstable")]
+  let has_children = webview.is_some();
```
Legend-Master, line 4619: @lucasfernog You might want to double check if this is correct

amrbashir:
does this PR in wry https://github.com/tauri-apps/wry/pull/1755/ tries to fix the same issue? also should we maybe fix it in wry or not?

Legend-Master:
> does this PR in wry [tauri-apps/wry#1755](https://github.com/tauri-apps/wry/pull/1755) tries to fix the same issue?

https://github.com/tauri-apps/wry/pull/1755 doesn't make any sense to me at least at a glance, it just tries to attach to the subclass to the first webview in the window.

> also should we maybe fix it in wry or not?

If `build_as_child` is supposed to allow the caller to control the sizing and focusing states, doing it in tauri is correct I think (we should document `build_as_child` better by the way)

zendy00:
We hit this exact bug in our app (`unstable` multi-webview on Windows) and spent a while diagnosing it, so sharing one concern about the trigger path.

The restore here runs on tao's `WindowEvent::Focused(true)`, which tao emits only from `WM_SETFOCUS` on the top-level window (tao 0.35 `event_loop.rs` — there is no `WM_ACTIVATE` arm). But in the scenario we reproduced (recent WebView2, ~Edge 149 / Chromium 141+, same family as tauri-apps/wry#1754), reactivating the window (alt-tab **or** clicking the taskbar icon / another app and back) restores OS focus **directly to the WebView2 child HWND** (`Chrome_WidgetWin_1`), so the top-level window never receives `WM_SETFOCUS`. We measured this with `GetGUIThreadInfo`: foreground/active = the top-level Tauri window, focus window = `Chrome_WidgetWin_1`, and keyboard input stays dead until a physical click. In that state tao won't emit `Focused(true)`, so the restore path in this PR wouldn't run.

That's why our wry-side fix (tauri-apps/wry#1755) hooks `WM_ACTIVATE` instead — it is always delivered to the top-level on activation regardless of where focus lands. One trap we hit there, in case it's useful: calling `MoveFocus` synchronously during `WM_ACTIVATE` handling gets overwritten when Windows subsequently restores focus to the child, so it has to be deferred (post a message to self and call `MoveFocus` after activation routing settles).

Has this been tested against a WebView2 runtime new enough to show the direct-to-child focus restore behavior? If `Focused(true)` still fires there, great — and tracking the *last focused* webview here is a nice improvement over restoring the first webview (which is what #1755 does). The two fixes could compose well.

Legend-Master:
@zendy00 This is tested on Webview2 149, and alt-tab does give the focus to the tao window for me

zendy00:
Thanks for checking this.

I re-tested in our app by applying this PR through `[patch.crates-io]`, pinning the Tauri crates to `a4c387e67dd0e672d22e6471413bfaa2cdce7c29`. Importantly, I removed my `wry` patch for the test, so `wry` resolves back to crates.io `0.55.1`.

With that setup, Alt+Tab focus restoration works correctly for our Windows `unstable` multi-webview case. So my concern above about `WindowEvent::Focused(true)` not being enough was wrong for this PR/test case.

Really glad to see this fixed on the Tauri side. Thanks for the quick clarification and for working on this.

FabianLars (approved):

PR #15628, opened 2026-07-01
fix(runtime-wry): reapply custom traffic light position on macOS

### LINKED ISSUE #13044: [bug] Setting window title resets traffic light position
### Describe the bug

Thanks so much for https://github.com/tauri-apps/tauri/pull/12366! Unfortunately, setting the window title from JS or Rust resets the traffic light position:

```ts
getCurrentWebviewWindow().setTitle(newTitle)
```

```rust
window.set_title(title).unwrap();
```

Note: Resizing the window will set the traffic lights back to the custom value.

https://github.com/user-attachments/assets/4bec4a38-321d-43cb-886f-530ef4793ca9

### Reproduction

_No response_

### Expected behavior

The traffic lights should remain in the same position

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 15.3.2 arm64 (X64)
    ✔ rustc: 1.85.1 (4eb161250 2025-03-15)
    ✔ cargo: 1.85.1 (d73d2caf9 2024-12-31)
    - node: 22.8.0
    - npm: 10.9.1
    - deno: deno 2.1.6

[-] Packages
    - tauri 🦀: 2.4.0
    - tauri-build 🦀: 2.1.0
    - wry 🦀: 0.50.5
    - tao 🦀: 0.32.8
    - @tauri-apps/api : 2.3.0 (outdated, latest: 2.4.0)
    - @tauri-apps/cli : 2.2.7 (outdated, latest: 2.4.0)

[-] Plugins
    - tauri-plugin-clipboard-manager 🦀: 2.2.2
    - @tauri-apps/plugin-clipboard-manager : 2.2.2
    - @tauri-apps/plugin-opener : 2.2.6
    - @tauri-apps/plugin-dialog : 2.2.0
    - tauri-plugin-window-state 🦀: 2.2.1
    - @tauri-apps/plugin-window-state : not installed!
    - tauri-plugin-shell 🦀: 2.2.0
    - @tauri-apps/plugin-shell : 2.2.0
    - tauri-plugin-log 🦀: 2.3.1
    - @tauri-apps/plugin-log : 2.3.1
    - tauri-plugin-updater 🦀: 2.6.1
    - tauri-plugin-fs 🦀: 2.2.0
    - @tauri-apps/plugin-fs : 2.2.0
    - tauri-plugin-single-instance 🦀: 2.2.2
    - tauri-plugin-os 🦀: 2.2.1
    - @tauri-apps/plugin-os : 2.2.1

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - framework: React
```

### Stack trace

```text

```

### Additional context

_No response_

### LINKED ISSUE #15451: [bug] [macOS] Custom traffic light position reset after quitting fullscreen of maximized window
### Describe the bug

When you set a custom traffic light position via `tauri.conf.json` but also with `.traffic_light_position(LogicalPosition::new(12.0, 21.0))`, double-click the window to maximize, go into fullscreen (green dot) and then quit it, the position is reset to the macOS default.

It seems to work if you resize the window again. Then it will indeed set the correct custom position again, just not when it stays maximized.

### Reproduction

https://github.com/lockinmarv/tauri-traffic-lights-bug

### Expected behavior

The traffic light position should stay the custom one without having to resize the window.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 26.5.0 arm64 (X64)
    ✔ Xcode: 26.5
    - node: 24.15.0
    - pnpm: 11.4.0
    - npm: 11.12.1

[-] Packages
    - tauri 🦀: 2.11.2
    - tauri-build 🦀: 2.6.2
    - wry 🦀: 0.55.1
    - tao 🦀: 0.35.3
    - @tauri-apps/api  ⱼₛ: 2.11.0
    - @tauri-apps/cli  ⱼₛ: 2.11.2

[-] Plugins
    - tauri-plugin-single-instance 🦀: 2.4.2
    - @tauri-apps/plugin-single-instance  ⱼₛ: not installed!

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - framework: React
    - bundler: Vite
```

### Stack trace

```text

```

### Additional context

_No response_

tenderdeve:
Closes #15451
Closes #13044

macOS resets a window's custom traffic light inset back to the system default in two cases, and it stays wrong until the window is resized:

- exiting fullscreen from a maximized window (#15451)
- changing the window title from JS/Rust (#13044)

Both come down to the same thing — AppKit re-lays-out the title bar buttons and tao doesn't reapply our inset. @FabianLars flagged these should be fixed together.

Fix: keep the configured position on the window and reapply it right after a title change, and on resize — a resize is the only event tao emits for a green-button fullscreen exit, and it matches the existing `resize the window and it comes back` workaround. `set_traffic_light_position` also updates the stored value so later reapplies use the latest.

All of it is macOS-gated. Raw windows (`CreateRawWindow`) don't carry the config so there's nothing to reapply for them.

### DIFF
--- a/.changes/fix-traffic-light-position-reset.md
+++ b/.changes/fix-traffic-light-position-reset.md
@@ -1,0 +1,5 @@
+---
+'tauri-runtime-wry': 'patch:bug'
+---
+
+Reapply a window's custom traffic light position on macOS after it gets reset by a title change or by exiting fullscreen on a maximized window, so the inset no longer reverts to the system default until the window is resized.
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -717,6 +717,8 @@ pub struct WindowBuilderWrapper {
   prevent_overflow: Option<Size>,
   #[cfg(target_os = "macos")]
   tabbing_identifier: Option<String>,
+  #[cfg(target_os = "macos")]
+  traffic_light_position: Option<Position>,
 }
 
 impl std::fmt::Debug for WindowBuilderWrapper {
@@ -1129,7 +1131,9 @@ impl WindowBuilder for WindowBuilderWrapper {
 
   #[cfg(target_os = "macos")]
   fn traffic_light_position<P: Into<Position>>(mut self, position: P) -> Self {
-    self.inner = self.inner.with_traffic_light_inset(position.into());
+    let position = position.into();
+    self.traffic_light_position = Some(position);
+    self.inner = self.inner.with_traffic_light_inset(position);
     self
   }
 
@@ -2520,6 +2524,10 @@ pub struct WindowWrapper {
   #[cfg(windows)]
   surface: Option<softbuffer::Surface<Arc<Window>, Arc<Window>>>,
   focused_webview: Arc<Mutex<Option<String>>>,
+  // Custom traffic light inset from the window config, kept so we can reapply it
+  // after macOS resets it on title changes (#13044) and fullscreen exit (#15451).
+  #[cfg(target_os = "macos")]
+  traffic_light_position: Option<Position>,
 }
 
 impl WindowWrapper {
@@ -3381,7 +3389,20 @@ fn handle_user_message<T: UserEvent>(
           WindowMessage::SetMaximizable(maximizable) => window.set_maximizable(maximizable),
           WindowMessage::SetMinimizable(minimizable) => window.set_minimizable(minimizable),
           WindowMessage::SetClosable(closable) => window.set_closable(closable),
-          WindowMessage::SetTitle(title) => window.set_title(&title),
+          WindowMessage::SetTitle(title) => {
+            window.set_title(&title);
+            // macOS resets the traffic light inset when the title changes, so
+            // reapply the configured position to keep it in place (#13044).
+            #[cfg(target_os = "macos")]
+            if let Some(position) = windows
+              .0
+              .borrow()
+              .get(&id)
+              .and_then(|w| w.traffic_light_position)
+            {
+              window.set_traffic_light_inset(position);
+            }
+          }
           WindowMessage::Maximize => window.set_maximized(true),
           WindowMessage::Unmaximize => window.set_maximized(false),
           WindowMessage::Minimize => window.set_minimized(true),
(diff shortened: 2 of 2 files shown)


FabianLars (changes requested):
this pr also contains https://github.com/tauri-apps/tauri/pull/15604 - i think that will resolve itself when the other one is merged but just blocking the merge on that to not do an oopsie

FabianLars:
In general i am not sure if this should be fixed in tauri. For similar cases i added the workarounds into tao and wry directly and hoped we could do the same here.

tenderdeve:
@FabianLars rebased onto `dev` — the #15604 commits are gone, so this PR now only touches `crates/tauri-runtime-wry/src/lib.rs` + its change file. Merge order no longer matters.

On whether it belongs in tauri vs tao/wry: fair point. The reapply itself is just `window.set_traffic_light_inset(pos)` (already a tao API), so tao could own this by caching the last inset and re-applying it internally after `set_title` and on fullscreen-exit — that'd fix it for non-tauri tao users too. The only reason it's here is that tauri-runtime-wry is where the configured position currently lives; tao's setter is one-shot and doesn't persist.

Happy to move it upstream instead — close this and open a tao PR making the inset sticky across those events. Your call on which layer you'd prefer; if you'd rather keep the fix tauri-side for now I'll leave it as is.

FabianLars:
> The only reason it's here is that tauri-runtime-wry is where the configured position currently lives; tao's setter is one-shot and doesn't persist.

This isn't the case actually, we store it in an objc Ivar as well https://github.com/tauri-apps/tao/blob/14a6dec20897904c9041c52cb92111b81db20e13/src/platform_impl/macos/view.rs#L343-L349

But i just remembered i had a tao branch related to one of the 2 github issues https://github.com/tauri-apps/tao/tree/refresh-lights (that's a bad way to do it i think - and its not using the ivar state as well)

Let me go through things, i'll get back to you.

tenderdeve:
@FabianLars opened the tao-side version: tauri-apps/tao#1254. It reapplies the inset from the existing `taoState` ivar after `setTitle:` **and** after leaving fullscreen, so it covers both #13044 and #15451 with no new state (addresses the "not using the ivar" point on your `refresh-lights` branch).

Happy to close this PR (#15628) once #1254 lands and the tao bump flows through, or keep this one as the interim runtime-wry fix if you would rather not wait on a tao release — your call.

tenderdeve:
@FabianLars #15604 merged, so I rebased on `dev` — the overlapping commits dropped out and this PR is now just the traffic-light change (`crates/tauri-runtime-wry/src/lib.rs` + changefile). Commit `4a857199c`, signed/Verified. Should be safe to merge now.

FabianLars:
i'll test the tao PR and if it works we'll go with that instead :) kinda forgot about this, sry

PR #15629, opened 2026-07-01
fix(build): copy newly added files from resource directories

### LINKED ISSUE #14992: [bug] tauri does not copy new files from resource directories
### Describe the bug

Given a `tauri.conf.json` like this:
```json
{
  "bundle": {
    "resources": ["my_resources/"]
  }
}
```

`tauri dev` will copy files from `src-tauri/my_resources/` to `src-tauri/target/debug/my_resources/`.

But this only occurs when either:
- A resource file that tauri already knows about, gets modified
- The build script needs to be re-run for some other reason, e.g. `tauri.conf.json` is modified or `build.rs` is modified.

Otherwise, newly added files get ignored, which is very confusing during development.

One workaround is to explicitly add `println!("cargo:rerun-if-changed=my_resources");` to `build.rs`.

### Reproduction

https://github.com/montyc1999/tauri-repro-new-files-not-picked-up

### Expected behavior

When new files are added to resource directories, tauri should copy them.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 14.8.1 arm64 (X64)
    ✔ rustc: 1.93.1 (01f6ddf75 2026-02-11)
    ✔ cargo: 1.93.1 (083ac5135 2025-12-15)
```

### Stack trace

```text

```

### Additional context

_No response_

### LINKED ISSUE #15390: [feat] Support skip resource watching

tenderdeve:
Closes #14992
Closes #15390

`copy_resources` only emitted `cargo:rerun-if-changed` for the individual files it resolved at build time. Since resource dirs/globs are expanded once during the build, a file added to a resource directory afterwards is unknown to cargo — the build script never re-runs, so the new file isn't copied until some unrelated change (editing `tauri.conf.json`, `build.rs`, or an already-tracked file) forces a rebuild. The documented workaround is to manually add `cargo:rerun-if-changed=<dir>`.

Fix: `ResourcePathsIter` now records each directory it walks and the fixed base of each glob, exposed via `rerun_if_changed()`. `copy_resources` emits a `rerun-if-changed` for each after copying, so adding or removing a file inside a resource directory re-runs the script and copies it.

Added a unit test covering directory, glob, and single-file patterns. Existing resource-path tests still pass.


Legend-Master (commented):
Please sign your past commits before I start reviewing so we don't lose much history going forwards.

tenderdeve:
@Legend-Master signed and re-pushed — the commit is now `c8911c4ec` (GitHub shows it Verified). Ready for review whenever you are.

Legend-Master (approved):
Nice one!

PR #15630, opened 2026-07-01
fix(runtime-wry)!: query monitors on main thread from runtime handle

### LINKED ISSUE #15170: [bug] Multiple app handle functions can crash Tauri, including available_monitors and cursor_position
### Describe the bug

I have been using Tauri on a new project for a few months. However, I have noticed that Tauri can crash from time to time. I decided to try to narrow down the possible causes.

Here is a basic program (only the file lib.rs was modified after running `pnpm create tauri-app`) that can crash on both Windows and Linux, within 1s to 60s:
```rust
pub fn run() {
    tauri::Builder::default()
        .plugin(tauri_plugin_opener::init())
        .invoke_handler(tauri::generate_handler![greet])
        .setup(|app| {
            let app_handle = app.handle().clone();

                loop {
                    let _handle = app_handle.clone();
                    let func_a = |app_handle: AppHandle| {
                        for _ in 0..1000 {
                            let _ = app_handle.available_monitors();
                        }
                    };
                    let func_b = |app_handle: AppHandle| {
                        for _ in 0..1000 {
                            let _ = app_handle.cursor_position();
                        }
                    };
                    let mut handles = vec!();
                    for _ in 0..10 {
                        let a_handle = app_handle.clone();
                        let b_handle = app_handle.clone();
                        handles.push(std::thread::spawn(move || {func_a(a_handle)}));
                        handles.push(std::thread::spawn(move || {func_b(b_handle)}));
                    }

                    for (_i, h) in handles.into_iter().enumerate() {
                        let _ = h.join();
                    }
                }
            });

            Ok(())
        })
}
```

Basically, the functions available_monitors and cursor_position are not thread safe. There might be other problematic functions, but I have only tested those. The crashes are more frequent when `func_a` and `func_b` are called, but can occur with only `func_a` or `func_b`.

Thanks in advance.

### Reproduction

The error can occur both in debug or release mode. Here is an example:

1. `pnpm install`
2. `pnpm tauri build --debug --no-bundle`
3. `./src-tauri/target/debug/crashtauri`

Here is the project (without node_modules or release folders):
[crashtauri.zip](https://github.com/user-attachments/files/26326421/crashtauri.zip)

### Expected behavior

The functions called from app.handle should not cause the program to crash.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Linux Mint 22.2.0 x86_64 (X64) (cinnamon on x11)
    ✔ webkit2gtk-4.1: 2.48.7
    ✔ rsvg2: 2.58.0
    - node: 22.21.1
    - pnpm: 10.20.0
    - npm: 10.9.4

[-] Packages
    - tauri 🦀: 2.10.3
    - tauri-build 🦀: 2.5.6
    - wry 🦀: 0.54.4, (outdated, latest: 0.55.0)
    - tao 🦀: 0.34.8, (outdated, latest: 0.35.0)
    - @tauri-apps/api  ⱼₛ: 2.10.1
    - @tauri-apps/cli  ⱼₛ: 2.10.1

[-] Plugins
    - tauri-plugin-opener 🦀: 2.5.3
    - @tauri-apps/plugin-opener  ⱼₛ: 2.5.3

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - framework: React
    - bundler: Vite
```

### Stack trace

```text
Multiple errors can occur. For examples: 
Aborted (core dumped)
or
Segmentation fault (core dumped)
or
malloc(): unaligned fastbin chunk detected
Aborted (core dumped)
```

### Additional context

_No response_

tenderdeve:
Closes #15170

`WryHandle` (the app/`AppHandle`-level runtime handle) implemented `primary_monitor`, `monitor_from_point` and `available_monitors` by reading `self.context.main_thread.window_target` **directly on the calling thread**. The tao `EventLoopWindowTarget` is not thread safe (GTK on Linux, Win32 on Windows), so calling these from a background thread races with the UI thread — the reporter sees `Aborted (core dumped)`, `Segmentation fault`, and `malloc(): unaligned fastbin chunk` within seconds.

`cursor_position` on the same handle already avoids this by marshalling through the event loop. This routes the three monitor queries the same way, via new `EventLoopWindowTargetMessage` variants handled on the main thread. `send_user_message` runs inline when already on the main thread, so no deadlock and no behavior change for main-thread callers.

The main-thread `Runtime` impl keeps its direct access — it already runs on the UI thread.

No repro on my machine (crash is Linux/Windows only), but the fix removes the off-thread `window_target` access that is the documented cause, mirroring the existing `cursor_position` path.


crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -2785,31 +2788,51 @@ impl<T: UserEvent> RuntimeHandle<T> for WryHandle<T> {
...
+    let (tx, rx) = channel();
+
+    let _ = send_user_message(
+      &self.context,
+      Message::EventLoopWindowTarget(EventLoopWindowTargetMessage::PrimaryMonitor(tx)),
+    );
+
+    rx.recv()
+      .ok()
+      .flatten()
```
FabianLars: it should be possible to use `event_loop_window_getter!()` here (and in the other changed functions) like in `cursor_position()` - looks a bit cleaner i think

tenderdeve:
@FabianLars done in fc2078b3f — switched all three (`primary_monitor`, `monitor_from_point`, `available_monitors`) to `event_loop_window_getter!` like `cursor_position`.

One wrinkle: the macro expands to `send_user_message(..)?`, so it needs a `Result`/`Option`-returning scope. `cursor_position` returns `Result`, but these three `RuntimeHandle` methods return a bare `Option<Monitor>` / `Vec<Monitor>` (the `Dispatcher` versions return `Result<..>`, which is why they use it directly). So I wrapped each call in a small `(|| ..)()` closure to absorb the `?`, then `.ok().flatten()`. `monitor_from_point` also passes a `|tx| ..` closure for the extra point arg.

Reads cleaner for `primary_monitor`/`available_monitors`; `monitor_from_point` ends up slightly nested. If you'd rather keep that one explicit (or want the trait methods changed to return `Result` — bigger change), say the word.

FabianLars:
Ah shit i didn't see the `?` in the macro. I think the previous version was better than the closure wrapper then - sorry for the back and forth. Please revert the last commit then, and while you're at it i think it'd be better to return None if send_user_message returns an error, otherwise there's a chance we're getting stuck on the recv() call.

tenderdeve:
@FabianLars done in a989173a4 (revert in 57bd57ce8) — no worries on the back-and-forth. Back to the explicit `send_user_message` + `recv()` version, and now each of the three bails out early (`None` / empty `Vec`) if the send fails, so a dropped message can no longer leave the caller blocked on `recv()`. `cargo check -p tauri-runtime-wry` is green.

Legend-Master:
Please sign your commits.

tenderdeve:
@Legend-Master done — re-signed all four commits with SSH and force-pushed, they're showing Verified now. Also rebased onto latest `dev` while I was at it.

Legend-Master:
@FabianLars Since the runtime and runtime-wry crates are not stable, I have gone ahead made them return results. I think this should be fine since the only consumer of the runtime crates is tauri and the implementers of the runtime crate need updates in respond to the added functions we do in minor functions anyways. Let me know your takes on this.

Legend-Master (approved):

FabianLars:
> Let me know your takes on this.

this can always cause issues if packages are updated independently but since i've been pushing breaking changes for tauri-utils i can't complain about that here 😂 so lgtm

Legend-Master:
I'm thinking to depend on minor versions after we get #15412 in for tauri

PR #15644, opened 2026-07-02
feat(bundler): show bundle sizes in the finished output

### LINKED ISSUE #15341: [feat] Show build sizes in console
### Describe the problem

I want to see how big or small the builds are in the console after running `cargo tauri build`. This is what Tauri is good at, after all! Right now, I need to find it with a command.

Image

Show the sizes of the bundles in the console.

### Alternatives considered

_No response_

### Additional context

_No response_

tenderdeve:
Closes #15341

After `tauri build`, the sizes of the generated bundles aren't shown — you have to look them up manually. This appends a human-readable size to each path in the `Finished N bundles at:` output:

```
    Finished 2 bundles at:
        .../myapp.app (24.6 MB)
        .../myapp_1.0.0_aarch64.dmg (25.1 MB)
```

Directory bundles (macOS `.app`, etc.) are measured recursively; size lookup is best-effort and falls back to no size annotation if a path can't be read, so it never fails the build.

Added a unit test for the size formatting.


crates/tauri-bundler/src/bundle.rs
```diff
@@ -298,6 +302,41 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
...
+      }
+    }
+    Some(total)
+  } else {
+    Some(metadata.len())
+  }
+}
+
+/// Formats a byte count into a human-readable string, e.g. `12.3 MB`.
+fn human_readable_size(bytes: u64) -> String {
```
Legend-Master: let's use something like https://crates.io/crates/human_bytes instead, since we might share it in tauri-cli for the future uses

crates/tauri-bundler/src/bundle.rs
```diff
@@ -298,6 +302,41 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
   Ok(bundles)
 }
 
+/// Total size in bytes of a bundle path, recursing into directories (e.g. macOS `.app`).
+///
+/// Returns `None` if the path can't be read.
+fn bundle_size(path: &std::path::Path) -> Option<u64> {
```
Legend-Master: Could you give this a quick test see how much time it usually takes to run?

.changes/bundler-show-sizes.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri-bundler": "minor:feat"
```
Legend-Master, line 2: let's also include tauri-cli so the changelog appears there as well

tenderdeve:
@Legend-Master both addressed:
- Swapped the hand-rolled formatter for the `human_bytes` crate (`human_bytes(bytes as f64)`), dropped the custom fn + its test. Output looks like `0 B` / `512 B` / `1 KB` / `12.3 MB`.
- Added `tauri-cli` to the changefile.

On timing: `bundle_size` just walks the bundle once with `walkdir` and sums `metadata.len()` — no reads. Benchmarked an equivalent tree (~700 files) at well under 30ms cold, and a real `.app` is typically far fewer files, so it's negligible next to the bundling/signing itself. Signed + `cargo check`/`fmt` clean.

Legend-Master (approved):
Thanks.

I have changed `human_size` to `bytesize` so we can display 2 digit precisions and made it fail (don't display size) instead of partial/wrong sizes

PR #15649, opened 2026-07-04
fix(runtime): clamp oversized window dimensions (fix #15648)

### LINKED ISSUE #15648: [bug] panic in 'main' when tauri.conf.json app.window.width > i32 max value
### Describe the bug

When trying to set app.window.width to a value greater than i32 max value,  the program panics. 

```
thread 'main' (13376312) panicked at /rustc/31fca3adb283cc9dfd56b49cdee9a96eb9c96ffd/library/core/src/panicking.rs:225:5:
panic in a function that cannot unwind
thread caused non-unwinding panic. aborting.
```

If this should be expected behaviour, please close.

### Reproduction

`pnpm create tauri-app@latest`

in `src-tauri/tauri.conf.json` set  `app.window.width` to 2147483648

`pnpm tauri dev`

### Expected behavior

I expected it to clamp the value to the width of the screen as it would with a number within i32 range.

### Full `tauri info` output

```text
[✘] Environment
    - OS: Mac OS 15.7.7 x86_64 (X64)
    ✘ Xcode: not installed!
    ✔ rustc: 1.96.1 (31fca3adb 2026-06-26)
    ✔ cargo: 1.96.1 (356927216 2026-06-26)
    ✔ Rust toolchain: stable-x86_64-apple-darwin (default)
    - node: 24.17.0
    - pnpm: 10.28.1
    - npm: 11.13.0

[-] Packages
    - tauri 🦀: 2.11.5
    - tauri-build 🦀: 2.6.3
    - wry 🦀: 0.55.1
    - tao 🦀: 0.35.3
    - @tauri-apps/api  ⱼₛ: 2.11.1
    - @tauri-apps/cli  ⱼₛ: 2.11.4

[-] Plugins
    - @tauri-apps/plugin-opener  ⱼₛ: 2.5.4

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - framework: React
    - bundler: Vite
```

### Stack trace

```text

```

### Additional context

I wanted the window to be max width at launch, so I set it to a large number

justjavac:
## What changed

- Clamp initial window width and height before passing them to the WRY window builder.
- Add a patch changeset for `tauri-runtime-wry`.

Closes #15648.

## Testing

- `cargo fmt -p tauri-runtime-wry`
- `cargo check -p tauri-runtime-wry`
- `git diff --check`

### DIFF
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -925,7 +925,10 @@ impl WindowBuilder for WindowBuilderWrapper {
   }
 
   fn inner_size(mut self, width: f64, height: f64) -> Self {
-    self.inner = self.inner.with_inner_size(LogicalSize::new(width, height));
+    self.inner = self.inner.with_inner_size(LogicalSize::new(
+      width.min(i32::MAX as f64),
+      height.min(i32::MAX as f64),
+    ));
     self
   }
 
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -925,10 +925,7 @@ impl WindowBuilder for WindowBuilderWrapper {
   }
 
   fn inner_size(mut self, width: f64, height: f64) -> Self {
-    self.inner = self.inner.with_inner_size(LogicalSize::new(
-      width.min(i32::MAX as f64),
-      height.min(i32::MAX as f64),
-    ));
+    self.inner = self.inner.with_inner_size(LogicalSize::new(width, height));
     self
   }
 
@@ -4455,6 +4452,18 @@ fn create_window<T: UserEvent, F: Fn(RawWindow) + Send + 'static>(
     }
   }
 
+  #[cfg(target_os = "macos")]
+  if let Some(size) = window_builder.inner.window.inner_size {
+    let monitor = if let Some(window_position) = &window_builder.inner.window.position {
+      find_monitor_for_position(event_loop.available_monitors(), *window_position)
+    } else {
+      event_loop.primary_monitor()
+    };
+    let scale_factor = monitor.map_or(1., |m| m.scale_factor());
+    window_builder.inner.window.inner_size =
+      Some(clamp_initial_window_size_for_macos(size, scale_factor));
+  }
+
   #[cfg(desktop)]
   if window_builder.prevent_overflow.is_some() || window_builder.center {
     let monitor = if let Some(window_position) = &window_builder.inner.window.position {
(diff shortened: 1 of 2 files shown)


crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -925,7 +925,10 @@ impl WindowBuilder for WindowBuilderWrapper {
   }
 
   fn inner_size(mut self, width: f64, height: f64) -> Self {
-    self.inner = self.inner.with_inner_size(LogicalSize::new(width, height));
+    self.inner = self.inner.with_inner_size(LogicalSize::new(
```
Legend-Master: It would be good to know where exactly crashed and fixing that instead of clamping the values here (there're another bug and this just makes it harder to find)

Legend-Master (changes requested):

crates/tauri-runtime-wry/src/lib.rs
justjavac: Thanks, updated in aed6a5dca.

I moved the clamp out of `WindowBuilderWrapper::inner_size`; that builder method now preserves the caller-provided size. The size is now capped only on macOS immediately before creating the window, using the monitor scale factor and capping the physical initial size to `i32::MAX`. This is the path that feeds tao/AppKit window creation instead of clamping every builder input.

Validation I ran:
- Reproduced the original crash on an unpatched checkout by setting `examples/helloworld` width to `2147483648` and running `RUST_BACKTRACE=full cargo run -p tauri --example helloworld`. It aborts with `panic in a function that cannot unwind`, at `tao::platform_impl::platform::app_delegate::did_finish_launching`.
- Ran the same oversized config on this branch; the app entered the event loop with no panic, then I stopped it manually.
- Added unit coverage that `inner_size` preserves oversized input, safe macOS values are left unchanged, oversized values are capped, and scale factor is accounted for.
- `cargo fmt --check`
- `cargo test -p tauri-runtime-wry --lib`
- `cargo check -p tauri-runtime-wry`
- `git diff --check`

Legend-Master: It sounds like a tao issue to me? We should probably fix it there

justjavac:
Moved the fix upstream to tao as suggested: tauri-apps/tao#1259.

The tao PR clamps oversized macOS initial inner window dimensions before constructing the native `NSWindow`, and uses the same clamped value for tao's initial `inner_rect`.

Validation run on the tao branch:
- `cargo fmt --check` (passes; only existing `imports_granularity` nightly warnings)
- `cargo test initial_inner_size --lib`
- `cargo test --lib`
- `cargo check` (passes with existing warnings)
- `git diff --check`

Earlier Tauri-side reproduction: setting the helloworld example width to `2147483648` reproduced the macOS abort during `did_finish_launching`; applying this clamp before native window creation allowed the app to enter the event loop.

PR #15655, opened 2026-07-06
chore: ran `typos --write-changes`

Legend-Master:


crates/tauri-bundler/src/bundle/macos/dmg/bundle_dmg
```diff
@@ -520,7 +520,7 @@ else
 	fi
 fi
 
-# Make sure it's not world writeable
+# Make sure it's not world writable
```
FabianLars, line 523: english is so confusing. something you have to keep the e and sometimes you don't.

FabianLars (approved):

PR #15668, opened 2026-07-07
chore: run clippy on all platforms

Legend-Master:


Legend-Master:
I don't really understand why it says `counter` not unused, it seems to be used to me 🤔

Edit: ah, `#[cfg(not(test))]`

FabianLars (approved):

PR #15674, opened 2026-07-08
docs: `menu_on_left_click` no longer works since v2.2

Legend-Master:
It's always overriden by `show_menu_on_left_click` since v2.2

Alternatively we can change `show_menu_on_left_click` to `Option<bool>` (I don't think it's worth fixing anymore that time has passed)


FabianLars (approved):
i feel like i say this all the time, but crazy that nobody noticed until now

PR #15675, opened 2026-07-08
refactor: migrate to svelte ts for API example

Legend-Master:
Since we already have the checkJs enabled, might as well just migrate to typescript so I don't get vscode marking every svelte file as having type errors

disclosure: initial migration was done through codex to save me some typings


Legend-Master:
I feel like every HTML based custom language I have used all suffer from more or less a subpar editor experience, some syntaxes don't work sometimes and all that, and the formatters are especially bad. And we are on Astro and Svelte 😂 I must say I had a much better experience with JSX based frameworks like React and SolidJS

FabianLars (approved):
i don't want to look at svelte "code" so enjoy a blind approval :)

PR #15683, opened 2026-07-09
fix(core): don't resend IPC over postMessage on page unload (fix #14154)

### LINKED ISSUE #14154: [bug] The asynchronous command is re-invoked when the WebView reloads.
### Describe the bug

When you reload a WebView while an asynchronous command is running, the command you initiated is automatically re-invoked.

### Reproduction

https://github.com/kanatapple/tauri-app-command

1. Run app
2. Open Dev Tools
3. Click greet button
4. Execute location.reload() in Dev Tools(while the greet command is running)

### Expected behavior

Unfinished commands are not re-invoked when the WebView reloads.

### Full `tauri info` output

```text
[✔] Environment
    ✔ WebView2: 139.0.3405.125
    ✔ MSVC:
        - Visual Studio Professional 2017
        - Visual Studio Professional 2019
    - node: 22.14.0
    - yarn: 1.22.22
    - npm: 10.9.2

[-] Packages
    - tauri 🦀: 2.8.4
    - tauri-build 🦀: 2.4.0
    - wry 🦀: 0.53.3
    - tao 🦀: 0.34.3
    - @tauri-apps/api : 2.2.0
    - @tauri-apps/cli : 2.0.0

[-] Plugins
    - tauri-plugin-http 🦀: 2.5.2
    - @tauri-apps/plugin-http : not installed!
    - tauri-plugin-log 🦀: 2.6.0
    - tauri-plugin-dialog 🦀: 2.3.3
    - @tauri-apps/plugin-dialog : 2.0.0
    - tauri-plugin-fs 🦀: 2.4.2

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - framework: Vue.js
    - bundler: Vite
```

### Stack trace

```text

```

### Additional context

_No response_

Sahilverse:
## Summary

Fixes #14154 - The asynchronous command is re-invoked when the WebView reloads

**Root cause:** `crates/tauri/scripts/ipc-protocol.js` sends invokes as a `fetch()` POST to the `ipc` custom protocol. If that fetch rejects, the code assumes the custom protocol is unsupported/blocked and falls back by resending the identical request over `window.ipc.postMessage`. A fetch aborted by the browser because the page is navigating away (reload) rejects the same way, so the reload was misclassified as a protocol failure, and the still-in-flight command got dispatched and executed a second time on the Rust side (there's no request-level idempotency in `Webview::on_message`).

**Fix:** track document teardown with a `pagehide` listener; when the fetch rejection happens during unload, drop the request instead of resending it. Genuine protocol failures (CSP block, unsupported webview) are unaffected and still fall back to postMessage as before.

## Reproduction / verification

Built a temporary local harness on `examples/helloworld` (not included in this PR): an async `greet` command with an invocation counter + 5s delay, and a page that auto-invokes once then calls `location.reload()` at 1.5s.

- **Before fix:** `greet` entered twice. once via the custom-protocol dispatch, once via a postMessage dispatch carrying the same `callback`/`error` ids, timed exactly at the reload.
- **After fix:** `greet` enters exactly once across the reload.
- **Regression check:** with the custom-protocol scheme blocked via CSP (`connect-src 'none'`, no reload involved), the postMessage fallback still triggers correctly and the command still resolves confirming the genuine-failure fallback path is untouched.

Tested on Windows 11 / WebView2. The affected code path (`fetch` rejection on navigation) is not Windows-specific, so the fix should apply the same way on macOS/Linux, though I only verified on Windows.

## Testing

- `cargo test -p tauri --lib` --> 57 passed
- `cargo clippy -p tauri --lib` --> clean
- `pnpm format:check` (prettier) -->  clean

## Notes for reviewers

- This fixes the *re-invocation*, not the underlying "orphaned in-flight command keeps running after reload" behavior (already flagged to users via a console warning in `core.js`)  that's a separate, larger design question (commands aren't cancellable today) and out of scope here.
- Considered a Rust-side in-flight-request dedup in `Webview::on_message` as a more defense-in-depth fix, but that touches core IPC dispatch for every command and adds state/cleanup for a bug whose root cause is fully explained and fixed at the JS transport layer. Happy to discuss if maintainers would prefer that approach instead/also.


Legend-Master:
Hey, thanks for contributing. As discussed in https://github.com/tauri-apps/tauri/issues/14154#issuecomment-3356088365, this only works on Windows so we can't take it as a solution here. Feel free to keep the conversation going though.

PR #15679, opened 2026-07-08
fix: menu channels are never cleaned up

Legend-Master:
Fix #13263

PR #15684, opened 2026-07-09
feat(example): rework menu examples

Legend-Master:
Depends on #15675

The primary reason for building this is to understand where our docs are lacking and which part of the APIs are not ergonomic enough.

I feel like the ownerships of menus are quite confusing, especially from the JavaScript side. The menus are reference counted that means if you have:

- Menu
  - Item 1
  - Item 2

when you close `Menu`, `Item 1` and `Item 2` are also released, but if you somehow got a reference to one of them left, either it's shared to another menu, or you forget the one you get from a `Menu.get`, you have a leak.

This is something we should definitely improve the docs on

---

preview of the change:

https://github.com/user-attachments/assets/28a18324-f522-450c-8e6f-bd9dec97e914


Legend-Master:
Svelte seems to insert comments to the DOM as anchors that simply can't work well with SortableJS...

Edit: did a workaround in d6b562d to always re-create the menus after sorting (referencing https://github.com/sveltejs/svelte/issues/11826 in case someone else stumbles onto this as well) (I still want to stick with SortableJS since it is the best working drag and drop library I have tried so far)

Legend-Master:
I'm still seeing constant memory increases when I spam re-creating the menus, even with the changes from #15679 applied 🤔

From some testing, the handles should be properly freed as well as the resources in the resource table 🤔

The code I used to test

```ts
const sleep = (ms: number) =>
  new Promise((resolve) => setTimeout(resolve, ms))

const items = [await MenuItem.new({ text: 'test' })]

for (let index = 0; index < 10_000; index++) {
  const menu = await Submenu.new({
    text: 'app',
    items
  })
  await menu.close()
  await sleep(10)
}

for (const item of items) {
  await item.close()
}
```

the rust equivalent doesn't reproduce the leak

```rust
let item = MenuItemBuilder::new("test").build(&app)?;
for _ in 0..1_000_000 {
  SubmenuBuilder::new(&app, "submenu").item(&item).build()?;
}
```

Legend-Master:
Anyways, we can do that later, the rework itself is ready and should not leak when we fix the underlying issues.

FabianLars:
The styling is off in dark mode:
grafik

Just looks like plain text to me and took me a second to realize those are elements i can drag

Legend-Master:
It's even weirder from my side 😂

image

I'm so used to the css variable just take care of these but I guess I was the wrong ones 🤦‍♂️

Legend-Master:
@FabianLars I am so not used to tailwind, am I supposed to add in CSS variables like we normally do or just spam `dark:*` like we did right now?

Legend-Master:
Spammed a bunch of `dark:*` for now... I thought sematic color names like `accent` and `neutral` would have the dark variant set automatically.

image

FabianLars:
idk either 😂

FabianLars (approved):

PR #15691, opened 2026-07-09
fix(menu): return error instead of panicking on unexpected ItemKind

### LINKED ISSUE #15686: [bug] menu plugin can panic
### Describe the bug

Many menu plugin APIs can panic due to the use of `do_menu_item!` which panics on unexpected `ItemKind`

### Reproduction

This is possible to do in JavaScript even the TypeScript types are not correct

```js
await Menu.new({
  items: [await Menu.new()]
})
```

### Expected behavior

can't panic

### Full `tauri info` output

```text
Current head 925faaafd1143ecaa95d2e722db290f3954f81f6
```

### Stack trace

```text

```

### Additional context

_No response_

krishpranav:
fix(menu): return error instead of panicking on unexpected ItemKind

## Problem
Menu plugin commands take `kind` straight from the IPC payload and dispatch on it through the `do_menu_item!` macro, whose accepted kind set never includes the root `Menu` variant (and some call sites narrow it further). Anything outside that set fell through to `_ => unreachable!()` and took the whole process down.

Reachable from plain JS — the TS types don't stop it, and raw IPC ignores them:

```js
await Menu.new({ items: [await Menu.new()] })
```

## Fix
- Changed the `do_menu_item!` fallback arm from `unreachable!()` to a recoverable error (`anyhow::anyhow!("unexpected menu item kind").into()`), matching the hand-written `_ => return Err(...)` arms already in `menu/plugin.rs`. One macro change covers all nine call sites.
- Added a regression test that drives the affected commands with mismatched kinds and asserts they return `Err` instead of panicking.
- Added `.changes/fix-menu-plugin-panic.md` with valid covector front matter (`"tauri": patch:bug`, `"tauri-macros": patch:bug`).

## Testing
```bash
cargo fmt --check
cargo clippy -p tauri -p tauri-macros --all-features -- -D warnings
cargo test -p tauri --lib menu::plugin::tests
```

The regression test panics on `dev` and passes with this change.

**Before** (fallback arm still `unreachable!()`):
```text
test menu::plugin::tests::unexpected_item_kind_errors_instead_of_panicking ... FAILED

thread '...' panicked at crates/tauri/src/menu/plugin.rs:751:3:
internal error: entered unreachable code

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 57 filtered out
```

**After** (this PR):
```text
test menu::plugin::tests::unexpected_item_kind_errors_instead_of_panicking ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 57 filtered out; finished in 0.01s
```

### App survives the repro (`examples/api`)
Ran the exact repro from the issue (`Menu.new({ items: [await Menu.new()] })`). On `dev` the process aborts; with this change the call rejects with `unexpected menu item kind` and the window stays alive.

demo

Fixes #15686

### DIFF
--- a/crates/tauri-macros/src/menu.rs
+++ b/crates/tauri-macros/src/menu.rs
@@ -75,7 +75,18 @@ impl Parse for DoMenuItemInput {
   }
 }
 
-pub fn do_menu_item(input: DoMenuItemInput) -> TokenStream {
+/// How the generated `match` should handle a `kind` that falls outside the
+/// accepted set.
+pub enum UnexpectedKind {
+  /// Emit `unreachable!()`. Kept for callers that have already narrowed `kind`
+  /// and treat any other value as a programming error.
+  Unreachable,
+  /// Return a recoverable `crate::Error`. Used when `kind` comes from untrusted
+  /// input (e.g. the IPC payload) and must not be able to abort the process.
+  Error,
+}
+
+pub fn do_menu_item(input: DoMenuItemInput, on_unexpected: UnexpectedKind) -> TokenStream {
   let DoMenuItemInput {
     rid,
     resources_table,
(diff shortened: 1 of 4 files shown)


crates/tauri-macros/src/menu.rs
```diff
@@ -130,7 +130,10 @@ pub fn do_menu_item(input: DoMenuItemInput) -> TokenStream {
         #expr
       }
       )*
-      _ => unreachable!(),
+      // `kind` comes straight from the (untrusted) IPC payload, so a kind
+      // outside the accepted set is reachable and must be a recoverable error
+      // rather than a panic. See tauri-apps/tauri#15686.
+      _ => Err(anyhow::anyhow!("unexpected menu item kind").into()),
```
Legend-Master: This is a breaking change, we'll need to either add a new macro or add an optional parameter to this macro

PR #15694, opened 2026-07-10
fix(android): Migrate Gradle scripts to `compilerOptions` DSL

Turbo87:
Kotlin Gradle Plugin deprecated the string-based `kotlinOptions` DSL and 2.3+ rejects it with a hard error. This PR changes the code to use the `compilerOptions` DSL with `JvmTarget.JVM_1_8` instead, which is accepted on both 1.9.x and 2.x, so projects can move to Kotlin 2.x without editing the vendored `:tauri-android` scripts.

This has a bit of overlap with #15335, but is a more focused non-breaking change, since this new syntax is supported by both major versions of the Gradle plugin.

### DIFF
--- a/.changes/android-compileroptions-dsl.md
+++ b/.changes/android-compileroptions-dsl.md
@@ -1,0 +1,7 @@
+---
+"tauri": patch:enhance
+"tauri-cli": patch:enhance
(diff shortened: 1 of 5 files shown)


Legend-Master (approved):
Thanks!

PR #15701, opened 2026-07-11
fix(linux): use resize cursor when within the defined inset for an undecorated webkitgtk window

yobson1:
while `undecorated_resizing.rs` handled the event to actually enable the resizing within the defined 5 pixel inset around the window it didn't handle changing the cursor to the appropriate type like `w-resize` or whichever cursor type for the respective side/corner the cursor is hovering.

this PR fixes that by listening to the `connect_motion_notify_event`, using the same logic to determine when we should be resizing as the existing code did and then using `window.set_cursor` to set the appropriate cursor for the resize.

additionally in `connect_button_press_event` we return `Propagation::Stop` instead of `Propagation::Proceed` because we have already determined this event is for resizing the window and claimed it, allowing it to continue to propagate can cause conflicts where the app tries to use the same event for something else like dragging via a custom top bar decoration.

included is also some basic refactors to extract logic that was being shared between the 3 webview event listeners and changing some ifs to guard clauses instead of nesting them a bunch.

i tested this on [Kesomannen/gale](https://github.com/Kesomannen/gale) which is a tauri app that exhibited this behaviour of no cursor type change near the edge for resizing due to it being undecorated and using this patched version of tauri addresses the issue nicely.

before:
the window was able to be resized if you were within the 5 pixel inset but there was no indication that was the case by the cursor changing (the cursor changes you can see in the clip are only AFTER i hold down LMB and begin the resize)

https://github.com/user-attachments/assets/afa3ca63-4bea-4ecd-815a-8c3712a1d281

after:
resizing works as normal and the cursor now correctly updates to reflect where you can resize from

https://github.com/user-attachments/assets/6c6de26f-622c-41f4-9f24-17eb9002e65a

ref: https://github.com/Kesomannen/gale/issues/226


.changes/webkitgtk-resize-cursor.md
```diff
@@ -0,0 +1,5 @@
+---
+'tauri-runtime-wry': 'patch:bug'
+---
+
+Fix the resize cursor not being set correctly on Linux/WebKitGTK and prevent resize operations being overridden
```
amrbashir:
```suggestion
---
'tauri-runtime-wry': 'patch:bug'
'tauri': 'patch:bug'
---

On Linux, fix resize cursor for undecorated Window not being set correctly when mouse is over a resize edge.
```

amrbashir (approved):
Thank you

PR #15716, opened 2026-07-15
chore(deps): update pnpm to v11

Legend-Master:

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,5 +28,5 @@
     "prettier": "^3.8.1"
   },
   "minimumReleaseAge": 4320,
-  "packageManager": "pnpm@10.30.3"
+  "packageManager": "pnpm@11.13.0"
 }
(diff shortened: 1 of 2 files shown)


FabianLars (approved):

package.json
```diff
@@ -28,5 +28,5 @@
     "prettier": "^3.8.1"
   },
   "minimumReleaseAge": 4320,
```
FabianLars, line 30: actually not sure if it still reads this, pnpm 11 moved all pnpm configs to pnpm-workspace.yaml

v11 now uses a default value of 1440 though so we're still good.

Legend-Master: Not sure, they said

> Moves settings out of package.json#pnpm into pnpm-workspace.yaml. In v11, pnpm no longer reads configuration from the pnpm field in package.json.

So technically not this field but who knows 😂

PR #15723, opened 2026-07-15
chore(deps): bump create-pull-request to v8

Legend-Master:
This should fix https://github.com/tauri-apps/tauri/actions/runs/29422184676/job/87382328317

### DIFF
--- a/.github/workflows/covector-version-or-publish.yml
+++ b/.github/workflows/covector-version-or-publish.yml
@@ -107,7 +107,7 @@ jobs:
 
       - name: Create Pull Request With Versions Bumped
         if: steps.covector.outputs.commandRan == 'version'
-        uses: peter-evans/create-pull-request@67ccf781d68cd99b580ae25a5c18a1cc84ffff1f # 7.0.6
+        uses: peter-evans/create-pull-request@v8.1.1
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/covector-version-or-publish.yml
```diff
@@ -107,7 +107,7 @@ jobs:
 
       - name: Create Pull Request With Versions Bumped
         if: steps.covector.outputs.commandRan == 'version'
-        uses: peter-evans/create-pull-request@67ccf781d68cd99b580ae25a5c18a1cc84ffff1f # 7.0.6
+        uses: peter-evans/create-pull-request@v8.1.1
```
Legend-Master, line 110: This is immutable so no need to pin

> https://github.com/peter-evans/create-pull-request/releases/tag/v8.1.1

FabianLars: is that a new feature? never saw that but it was damn overdue!

Legend-Master: Yeah, I think I saw that first somewhere a few months ago, seems like a feature from the end of last year

> https://github.blog/changelog/2025-10-28-immutable-releases-are-now-generally-available/

FabianLars (approved):

PR #15735, opened 2026-07-16
fix: menu regression from #15679

lucasfernog:


FabianLars (dismissed):

Legend-Master (dismissed):
I just knew for a fact calling something in drop would break something 😂

crates/tauri/src/menu/plugin.rs
```diff
@@ -889,7 +889,11 @@ struct MenuChannels(Mutex<HashMap<MenuId, Channel<MenuId>>>);
 
 // Called in `Menu`'s `Drop` to clean up the event handlers
 pub(crate) fn remove_menu_channel<R: Runtime>(app: &AppHandle<R>, id: &MenuId) {
-  app.state::<MenuChannels>().0.lock().unwrap().remove(id);
+  // The state is only managed once this plugin is initialized, which happens after
+  // the `Builder::menu` closure runs, so menus dropped before that have no channel to remove.
```
Legend-Master:
```suggestion
  // the `crate::Builder::menu` closure runs, so menus dropped before that have no channel to remove.
```

Legend-Master (commented):
I wonder if we should move the menu creation after `app.register_core_plugins()?;` like we did for initializing tray icons? (we should probably still keep the `try_state` to keep it safe though)

https://github.com/tauri-apps/tauri/blob/e9e88a7b3dff78fa2fe841c07d1d927282c4883a/crates/tauri/src/app.rs#L2383-L2396

FabianLars (approved):

sftse:
> I wonder if we should move the menu creation after `app.register_core_plugins()?;` like we did for initializing tray icons? (we should probably still keep the `try_state` to keep it safe though)
> 
> https://github.com/tauri-apps/tauri/blob/e9e88a7b3dff78fa2fe841c07d1d927282c4883a/crates/tauri/src/app.rs#L2383-L2396

This is probably not the right conversation to unpack this, but I think the state management should be strictly split between user and Tauri state. State initialization order is a bit non-obvious, and could be made more type safe.

Legend-Master:
I do think there's room for improvement, but not sure how, do you have some ideas (no rush though)?

Issue #15028, opened 2026-03-03
[docs] updater signer environment variables inconsistency


Legend-Master:
_Copied from the message I posted on discord_

Our doc says

> `export TAURI_SIGNING_PRIVATE_KEY="Path or content of your private key"`

while `tauri signer generate` outputs

> ```
> Environment variables used to sign:
> - `TAURI_SIGNING_PRIVATE_KEY`: String of your private key
> - `TAURI_SIGNING_PRIVATE_KEY_PATH`: Path to your private key file
> - `TAURI_SIGNING_PRIVATE_KEY_PASSWORD`:  Your private key password (optional if key has no password)
> ```

The `signer` command uses what it says while the `bundle` command only uses `TAURI_SIGNING_PRIVATE_KEY` (try as path first and if it's not a file, use it as literal)

I don't mind either of them, we just need to be consistent about it, worst case we do both (keep both `TAURI_SIGNING_PRIVATE_KEY` and `TAURI_SIGNING_PRIVATE_KEY_PATH` and allow path or content for `TAURI_SIGNING_PRIVATE_KEY`)

mnardit:
While you are updating the documentation for the signer environment variables (specifically TAURI_SIGNING_PRIVATE_KEY_PASSWORD), I highly recommend adding a warning/note about shell escaping behavior.

Many users run into a cryptic Wrong password for that key error because their terminal, CI pipeline, or CLI wrappers (like AI coding agents) silently escape special characters in the password. For example, if a password contains an exclamation mark (!), underlying tools like shell-quote often mutate it to \! during key generation or when exporting the env var, completely breaking the signing process downstream.

I recently spent hours debugging this exact scenario where the environment variable was silently mutated, causing 20+ releases to fail. I wrote a detailed post-mortem about it here:

https://max.nardit.com/articles/the-invisible-backslash

Adding a small tip in the docs like "Warning: Ensure your password does not contain shell metacharacters that might be silently escaped by your terminal or CI environment when using TAURI_SIGNING_PRIVATE_KEY_PASSWORD" would save developers a massive amount of debugging time!

nandanpugalia:
I dug into the current v2 docs and CLI source because I recently spent time around updater signing while shipping a Tauri updater flow.

From what I can tell:

- `tauri build` / updater artifact signing currently reads `TAURI_SIGNING_PRIVATE_KEY`, and treats its value as either the key contents or a path if that path exists.
[...]

PR #15737, opened 2026-07-17
docs(cli): document TAURI_SIGNING_PRIVATE_KEY_PATH and clarify key vars

lazerg:
`TAURI_SIGNING_PRIVATE_KEY_PATH` was missing from the reference even though `tauri signer sign` reads it. The old wording also glossed over how `build`/`bundle` accept the key as a string or a path, while `signer sign` only takes the literal string. So this adds the var and spells out that difference. #15028

### DIFF
--- a/.changes/signing-env-vars.md
+++ b/.changes/signing-env-vars.md
@@ -1,0 +1,6 @@
+---
+"tauri-cli": patch:enhance
+"@tauri-apps/cli": patch:enhance
+---
+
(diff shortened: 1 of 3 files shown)


Legend-Master (commented):
Thanks! Could you also update the `sign generate` command's output and add a change file?

lazerg:
Done, updated the `signer generate` output to match and added a change file.

Legend-Master (approved):
Thanks. Please upload your signing public key to GitHub so I can merge this.

Image

lazerg:
✅

PR #15749, opened 2026-07-20
fix(cef): preserve mapped custom protocol authorities

### LINKED ISSUE #15748: [bug] CEF runtime: mapped custom protocol does not match Wry's WebView2/Android behaviour
### Describe the bug

The CEF runtime maps custom protocol URLs onto `https` the same way Wry does for WebView2 and Android, but the mapping is not reverted and the routing is narrower, so the two runtimes do not behave the same.

Wry maps `scheme://authority/path` to `https://scheme.authority/path` and reverts it before calling the handler (`wry::custom_protocol_workaround::revert_uri_work_around`). On CEF:

1. The scheme handler factory is registered against the fixed domain `<scheme>.localhost` (`request_context.rs`), so only the bare `localhost` authority matches. `app://a.localhost/index.html` maps to `https://app.a.localhost/index.html`, misses the fixed domain, and falls through to the network stack. Wry does not restrict the authority: `is_work_around_uri` matches on the `<scheme>.` prefix alone and WebView2 filters on `https://<scheme>.*`, so this same URL already works on WebView2 and Android.

2. There is no revert step. `process_request` passes `request.url()` straight to the `UriSchemeProtocolHandler`, so the callback receives `https://app.localhost/index.html`. WebView2 reverts in `prepare_request` and Android rewrites `request.uri_mut()` before dispatch, so both give the handler `app://localhost/index.html`. On macOS and Linux the real scheme is registered with WebKit, so handlers see `app://` there too. CEF is the only runtime where a protocol handler has to special-case the URI it is given.

The practical effect of (1) is that a scheme is limited to one origin on CEF. `app://a.localhost` and `app://b.localhost` are two distinct browser origins with separate storage on WebView2 and Android; on CEF only `localhost` is reachable.

### Reproduction

1. Register a custom protocol, e.g. `app`.
2. Create a webview pointing at `app://a.localhost/index.html` (any authority that is not exactly `localhost`).
3. The page fails to load; the registered protocol handler is never invoked. The same URL loads on WebView2 and Android.

For the second half: point a webview at `app://localhost/index.html` and log the URI inside the protocol handler. It is `https://app.localhost/index.html` on CEF and `app://localhost/index.html` on every Wry runtime.

### Expected behavior

CEF should behave the same as the WebView2/Android mapped-protocol path: a custom protocol URL with any authority routes to the registered handler, and the handler receives the logical `scheme://authority/path` URL. Requests that are not mapped custom protocol URLs should keep falling through to CEF's normal network stack.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 26.5.2 arm64 (X64)
    ✔ Xcode: 26.6
    ✔ Rust toolchain: 1.96.0-aarch64-apple-darwin (overridden by environment variable RUSTUP_TOOLCHAIN)
    - node: 26.4.0
    - pnpm: 11.9.0
    - npm: 11.4.2

[-] Packages
    - tauri 🦀: git+https://github.com/nclave-app/tauri-cef.git?rev=d034b86ae#d034b86ae00058f3dd39b009ca095689389f2e65 (2.11.5)
    - tauri-build 🦀: git+https://github.com/nclave-app/tauri-cef.git?rev=d034b86ae#d034b86ae00058f3dd39b009ca095689389f2e65 (2.6.3)
    - wry 🦀: 0.55.1
    - tao 🦀: 0.35.3
    - tauri-cli 🦀: 2.11.4

[-] Plugins
    - tauri-plugin-dialog 🦀: git+https://github.com/tauri-apps/plugins-workspace.git?rev=1423992771e0b57582ca3b06a6adc46ec26aa784#1423992771e0b57582ca3b06a6adc46ec26aa784 (2.7.1), (outdated, latest: 2.7.2)
    - tauri-plugin-opener 🦀: git+https://github.com/tauri-apps/plugins-workspace.git?rev=1423992771e0b57582ca3b06a6adc46ec26aa784#1423992771e0b57582ca3b06a6adc46ec26aa784 (2.5.4)
    - tauri-plugin-fs 🦀: git+https://github.com/tauri-apps/plugins-workspace.git?rev=1423992771e0b57582ca3b06a6adc46ec26aa784#1423992771e0b57582ca3b06a6adc46ec26aa784 (2.5.1)

[-] App
    - build-type: bundle
    - CSP: connect-src 'self' ipc: http://ipc.localhost nclave-app: https://nclave-app.localhost; media-src 'self' data: blob:; style-src 'self' 'unsafe-inline'; webrtc 'none'; img-src 'self' data: blob:; script-src 'self'; form-action 'none'; worker-src 'none'; frame-src nclave-app: https://*.localhost; manifest-src 'self'; default-src 'self'; font-src 'self' data:; object-src 'none'; base-uri 'none'
    - frontendDist: ../../../web/shell/dist

Just a note, the tauri cli / code is from my repo, which is an exact fork of feat/cef branch
```

### Stack trace

```text

```

### Additional context

I have a fix working and can open a PR. It is running in an unreleased app of mine that relies on per-authority origins on CEF, so it has had real use, though nothing public yet.

karteekiitg:
Closes #15748

Brings the CEF mapped custom protocol in line with what Wry already does for WebView2 and Android. The mapping onto `https` was already there, but the factory was registered against the fixed domain `<scheme>.localhost`, so mapped URLs with any other authority were never intercepted, and there was no revert step, so the Tauri protocol callback got the mapped `https` URL instead of the logical `scheme://` one.

Changes:

- Register a single factory per mapped scheme with a NULL domain, which for a standard scheme matches every host. CEF cannot match a wildcard domain, and one registration per authority is not workable when authorities are created at runtime.
- Add `cef_impl/mapped_scheme.rs`, which decides what that factory claims. It routes only hosts of the form `<scheme>.*.localhost` where the leading label is a registered custom scheme for that browser, and returns `None` for everything else. Because `https`/`http` is a built-in scheme, `None` falls through to CEF's normal network handler, so unrelated origins are untouched.
- Normalize the intercepted URL back to `scheme://authority/path` before calling the handler, mirroring `revert_uri_work_around` in WebView2's `prepare_request` and the `request.uri_mut()` rewrite on Android, so handlers see the same request shape on every runtime.
- Update the devtools init script guard to match the scheme label plus a `.localhost` suffix rather than a fixed host set, so per-authority origins are detected too.

Net effect: distinct logical authorities on one scheme become distinct browser origins with their own storage, which is already how WebView2 and Android behave.

Note this is deliberately stricter than Wry: `is_work_around_uri` matches any host beginning with `<scheme>.`, and WebView2's filter is `https://<scheme>.*`. The CEF factory has no per-protocol filter to sit behind, so it is consulted for every http(s) request, and claiming everything under `<scheme>.` felt too broad. Restricting to `.localhost` keeps the intercepted namespace reserved. Happy to widen it to match Wry exactly if you'd rather the two be identical.

### Testing

`cargo test -p tauri-runtime-cef`, 9 tests covering namespace matching, rejection of unrelated and malformed hosts, URL reversal, and case normalization.

Also verified live over CDP: origin and storage isolation across authorities, non-mapped HTTPS falling through to the network, and separate renderer processes per origin.

First time contributing here, so happy to adjust the approach or the naming. This has been running in an unreleased app of mine that relies on per-authority origins on CEF, so it has had real use, but nothing public yet.

### DIFF
--- a/crates/tauri-runtime-cef/src/cef_impl/mapped_scheme.rs
+++ b/crates/tauri-runtime-cef/src/cef_impl/mapped_scheme.rs
@@ -1,0 +1,254 @@
+// Copyright 2019-2024 Tauri Programme within The Commons Conservancy
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: MIT
+
+//! Host-preserving mapped custom-protocol routing for CEF.
+//!
+//! CEF — like WebView2 and Android WebView — cannot expose a non-standard
+//! custom scheme to `fetch`/storage/etc. without registering that scheme in
+//! every process. Tauri instead maps a logical custom-protocol URL onto the
+//! built-in `https` (or `http`) scheme, exactly like Wry's WebView2/Android
+//! workaround (`wry::custom_protocol_workaround`):
+//!
+//! ```text
+//! logical   scheme://authority/path
+//! mapped    https://scheme.authority/path
+//! ```
+//!
+//! The `scheme.` label is prepended to the authority, so the *logical authority
+//! is preserved* as a browser-visible subdomain rather than collapsed to a
+//! single fixed host. That is what lets distinct logical authorities on one
+//! scheme (`app://a.localhost/…`, `app://b.localhost/…`) become distinct
+//! browser origins (`https://app.a.localhost/…`, `https://app.b.localhost/…`),
+//! each with its own same-origin storage.
+//!
+//! Unlike Wry — whose per-protocol interception filter is scoped by the host
+//! runtime — the CEF factory is registered against the whole built-in scheme
+//! (see [`crate::cef_impl::request_context`]). Routing therefore has to do two
+//! extra jobs here:
+//!
+//! 1. [`split_mapped_host`] restricts interception to the reserved
+//!    `<scheme>.….localhost` namespace, so unrelated HTTPS is never claimed.
+//! 2. [`to_logical_url`] normalizes an intercepted mapped URL back to the
+//!    logical custom-protocol request before it reaches the Tauri handler, so
+//!    the handler always receives one logical URL shape across every runtime.
+
+use url::Url;
+
+#[derive(Debug, PartialEq, Eq)]
+pub(crate) struct MappedRequest {
+  pub(crate) custom_scheme: String,
+  pub(crate) logical_url: String,
+}
+
+/// Splits a mapped work-around host `<scheme>.<authority>` into its logical
+/// scheme label and logical authority, but only for the reserved
+/// `*.localhost` namespace.
+///
+/// Returns `None` for any host that is not `<non-empty-label>.….localhost`, so
+/// a factory registered against the whole `https`/`http` scheme leaves
+/// unrelated origins untouched.
+///
+/// ```text
+/// app.localhost                -> ("app", "localhost")
+/// app.a.localhost              -> ("app", "a.localhost")
+/// localhost | example.com      -> None
+/// app.a.localhost.example.com  -> None  (does not end in .localhost)
+/// ```
+pub(crate) fn split_mapped_host(host: &str) -> Option<(&str, &str)> {
+  // A bare `localhost` is not part of the reserved namespace: the scheme label
+  // is mandatory, so the host must have at least `<label>.localhost`.
+  if !host.ends_with(".localhost") {
+    return None;
+  }
+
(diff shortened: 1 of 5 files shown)


karteekiitg:
@FabianLars just fyi, i kept the checks intentionally scoped to this particular fix, so that the pr remains light. I can fix them if you want me to, but I prefer not to touch the code beyond the scope - would need to understand the context of code I wouldn't know as well

PR #15760, opened 2026-07-22
feat: add e2e tests for API

lucasfernog:


crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -4916,15 +4984,54 @@ You may have it installed on another user account, but it is not available for t
...
+      });
+    }
+
     webview_builder = webview_builder.with_proxy_config(config);
   }
 
   #[cfg(windows)]
   {
-    if let Some(additional_browser_args) = webview_attributes.additional_browser_args {
+    // WebView2 is meant to apply `WEBVIEW2_ADDITIONAL_BROWSER_ARGUMENTS` on top of the
```
Legend-Master: Please don't do this work around, see https://github.com/tauri-apps/wry/pull/1783#issuecomment-5057319548, we can try https://github.com/tauri-apps/wry/issues/1782#issuecomment-5566579268

.github/workflows/test-api-e2e.yml
```diff
@@ -0,0 +1,164 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - { name: Linux, os: ubuntu-latest }
+          - { name: Windows, os: windows-latest }
+          - { name: macOS, os: macos-latest }
+
+    steps:
+      - uses: actions/checkout@v4
```
Legend-Master: We want to update these before merging

Legend-Master (changes requested):

crates/tauri-runtime-wry/src/lib.rs
FabianLars: +1 - this cannot land. if we really want this than maybeee put it behind a feature flag, an env var guard is not good enough imo. though i'd still not really like this tbh 🤔 

also this feels like a wry change, not a tauri(-runtime-wry) one, no?

.github/workflows/test-api-e2e.yml
```diff
@@ -0,0 +1,179 @@
...
+        # See https://github.com/tauri-apps/wry/issues/1782 and
+        # https://github.com/MicrosoftEdge/WebView2Feedback/issues/5645.
+        # The runner has an interactive desktop session, so window-state tests run as-is;
+        # the tray module is skipped to match the Linux job.
+        env:
+          E2E_SKIP: tray
+          E2E_SPEC_RETRIES: '1'
+        run: |
+          winget install gerardog.gsudo --accept-source-agreements --accept-package-agreements
+          $env:PATH = "$env:LOCALAPPDATA\Microsoft\WinGet\Links;$env:PATH"
```
Legend-Master:
```suggestion
```

This should not be needed

examples/api/src-tauri/Cargo.toml
```diff
@@ -23,6 +23,11 @@ log = "0.4.21"
 tauri-plugin-sample = { path = "./tauri-plugin-sample/" }
 tauri-plugin-log = "2"
 
+# WebDriver automation bridge, used by the `@tauri-apps/api` e2e suite (packages/api-e2e).
+# Desktop-only and gated behind `debug_assertions` at the call site so it never ships in release.
+[target."cfg(not(any(target_os = \"android\", target_os = \"ios\")))".dependencies]
+tauri-plugin-automation = "0.1.4"
```
Legend-Master: I wonder if this should be gated in a feature flag like we had in https://github.com/crabnebula-dev/tauri-e2e-example/blob/22decccb6ed7a1292d78a4a1274d75d746066119/src-tauri/Cargo.toml#L28

.github/workflows/test-api-e2e.yml
```diff
@@ -0,0 +1,179 @@
...
+        env:
+          CN_API_KEY: ${{ secrets.TAURI_E2E_CN_API_KEY }}
+
+      # The CrabNebula Webdriver proxies every command to a server running *inside* the app
+      # process, so when the app dies the suite only ever reports `connection refused` — the
+      # actual cause is never in this log. The app's stdout/stderr is inherited (its `println!`
+      # output shows up above) and no panic message is printed, which means it goes down on a
+      # signal. macOS records the reason in a crash report, and that is the only place it
+      # survives the run.
+      - name: collect app crash reports (macOS)
```
Legend-Master, line 133: Ideally we don't have such a complicated script in the CI file

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -4033,6 +4033,17 @@ fn handle_user_message<T: UserEvent>(
       }
       EventLoopWindowTargetMessage::SetTheme(theme) => {
         event_loop.set_theme(to_tao_theme(theme));
+        // On macOS tao caches each window's theme and only refreshes it from the
+        // system-wide appearance change notification, which the app-level
+        // `NSApp.setAppearance` call above never posts, so `Window::theme()`
+        // would keep reporting the previous value. tao's window-level setter
+        // does update the cache, so push the theme through it as well.
+        #[cfg(target_os = "macos")]
```
Legend-Master, line 4041: Maybe this should be upstreamed to tao?

packages/api-e2e/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "@tauri-apps/api": "../api/dist",
+    "@types/node": "^22.10.0",
+    "@wdio/cli": "^9.19.0",
+    "@wdio/globals": "^9.19.0",
+    "@wdio/local-runner": "^9.19.0",
+    "@wdio/mocha-framework": "^9.19.0",
+    "@wdio/spec-reporter": "^9.19.0",
+    "expect-webdriverio": "^5.0.0",
+    "tsx": "^4.20.0",
+    "typescript": "^5.9.0",
```
Legend-Master: We're on ts 6 for all other places

packages/api-e2e/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "@crabnebula/test-runner-backend": "^0.2.9",
+    "@tauri-apps/api": "../api/dist",
+    "@types/node": "^22.10.0",
+    "@wdio/cli": "^9.19.0",
+    "@wdio/globals": "^9.19.0",
+    "@wdio/local-runner": "^9.19.0",
+    "@wdio/mocha-framework": "^9.19.0",
+    "@wdio/spec-reporter": "^9.19.0",
+    "expect-webdriverio": "^5.0.0",
+    "tsx": "^4.20.0",
```
Legend-Master: I'm not sure I see anywhere this was used

PR #15777, opened 2026-07-25
fix: make embedded assets codegen and config serialization deterministic

VityaSchel:
Hi, I'm developing an app with Tauri that requires reproducible builds. I found a non-determenistic build part that has to do with CSP. I have seen #14978 and #14986 but those didn't fix it fully.

1. tauri-codegen: `ToTokens for EmbeddedAssets` emits the embedded asset map and the CSP script/style hash lists in the order the assets were collected, `WalkDir` traversal without sorting, i.e. raw readdir order. It differs on different machines with e.g. ext4 hashed directories so `CspHash` slice is different
2. tauri-utils: the derived `Serialize` for `Csp::DirectiveMap`, `HeaderSource::Map` and `PluginConfig` writes `HashMap` key order, which is random per process. Tauri cli serializes processed config into the generated android project (`assets/tauri.conf.json`) so every build of the same source produces a differently ordered file

With this patched I was able to produce identical unsigned apks with Tauri with CSP enabled.


VityaSchel:
Please let me know if you want any changes to this PR. Would love to have this shipped in coming versions as right now I have to either host a whole tauri-codegen source in my project's tree, add a patch-cargo dependency that works like patch-package in JS world or write complex shell scripts that do pretty much the same thing as patch-cargo but without third party tool.

crates/tauri-codegen/src/embedded_assets.rs
```diff
@@ -391,7 +391,11 @@ impl EmbeddedAssets {
 impl ToTokens for EmbeddedAssets {
   fn to_tokens(&self, tokens: &mut TokenStream) {
     let mut assets = TokenStream::new();
-    for (key, (input, output)) in &self.assets {
+    // Iterate sorted so the generated code is deterministic
+    // see: https://github.com/tauri-apps/tauri/issues/14978
+    let mut sorted_assets: Vec<_> = self.assets.iter().collect();
```
Legend-Master: I think we can change `assets` to `BTreeMap` since it's private? (as well as `csp_hashes.inline_scripts`)

Legend-Master (approved):
Thanks!

PR #15778, opened 2026-07-25
fix(cli): respect CARGO_BUILD_TARGET when resolving build target

### LINKED ISSUE #15768: feat/cef: dev CEF assembly ignores CARGO_BUILD_TARGET, reads only .cargo/config.toml
On branch `feat/cef` (rev be0f6b5), `cargo tauri dev` on macOS builds the app with the target cargo resolves from the `CARGO_BUILD_TARGET` environment variable, but the CEF bundling/dev-assembly step resolves the target from `.cargo/config.toml`'s `build.target` only. In a repo whose cargo config defaults to a cross deploy target, the run ends with:

```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 45.31s
       Error invalid CEF target x86_64-unknown-linux-musl: Unsupported target triplet: x86_64-unknown-linux-musl
```

even though `CARGO_BUILD_TARGET=aarch64-apple-darwin` was set (and cargo compiled for it, as the successful build shows). Cargo gives the env var precedence over the config file, so the CLI's target resolution should too. Workaround: pass `--target` explicitly.

(reported by an AI agent via Claude Code on behalf of a user)

lazerg:
`cargo tauri dev` compiles for the target Cargo resolves from `CARGO_BUILD_TARGET`, but the CEF dev assembly read the target from `.cargo/config.toml`'s `build.target` only. When the config file points at a different target, the app builds for the env target while CEF bundling fails on the config one, e.g. `invalid CEF target x86_64-unknown-linux-musl`.

Cargo gives `CARGO_BUILD_TARGET` precedence over `build.target` in the config file, so `CargoConfig::load` now does the same.

Closes #15768

### DIFF
--- a/.changes/cargo-build-target-env.md
+++ b/.changes/cargo-build-target-env.md
@@ -1,0 +1,6 @@
+---
+"tauri-cli": patch:bug
+"@tauri-apps/cli": patch:bug
+---
+
+Respect the `CARGO_BUILD_TARGET` environment variable when resolving the build target, matching Cargo's precedence over `build.target` in `.cargo/config.toml`.
--- a/crates/tauri-cli/src/interface/rust/cargo_config.rs
+++ b/crates/tauri-cli/src/interface/rust/cargo_config.rs
@@ -60,6 +60,13 @@ impl Config {
   pub fn load(path: &Path) -> Result<Self> {
     let mut config = Self::default();
 
+    if let Ok(target) = std::env::var("CARGO_BUILD_TARGET")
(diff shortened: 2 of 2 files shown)


lazerg:
@lucasfernog could you take a look at this when you get a chance.

FabianLars (approved):
we should probably use https://docs.rs/cargo-config2/latest/cargo_config2/ in the future but let's got with this for now to not diverge feat/cef further. thanks for the pr :)

PR #15779, opened 2026-07-26
fix(core): carry filter through pending emit queue

### LINKED ISSUE #15759: [bug] `emit_filter` emits to all targets if inserted to pending
https://github.com/tauri-apps/tauri/blob/17717757b75581d54de4c8fa2c8c450d10c3a257/crates/tauri/src/event/listener.rs#L215

I tried this but it seems like we can't do this since our public API didn't require `Send`

```diff
diff --git a/crates/tauri/src/event/listener.rs b/crates/tauri/src/event/listener.rs
index afcc3ca5b..b72bb982d 100644
--- a/crates/tauri/src/event/listener.rs
+++ b/crates/tauri/src/event/listener.rs
@@ -24,7 +24,10 @@ enum Pending {
     event: crate::EventName,
     handler: Handler,
   },
-  Emit(EmitArgs),
+  Emit {
+    args: EmitArgs,
+    filter: Option<Box<dyn Fn(&EventTarget) -> bool + Send>>,
+  },
   RemoveForTarget(EventTarget),
 }
 
@@ -128,7 +131,7 @@ impl Listeners {
       match action {
         Pending::Unlisten(id) => self.unlisten(id),
         Pending::Listen { id, event, handler } => self.listen_with_id(id, event, handler),
-        Pending::Emit(args) => self.emit(args)?,
+        Pending::Emit { args, filter } => self.emit_filter(args, filter)?,
         Pending::RemoveForTarget(target) => self.remove_listeners_for_target(target),
       }
     }
@@ -212,7 +215,10 @@ impl Listeners {
     let mut maybe_pending = false;
 
     match self.inner.handlers.try_lock() {
-      Err(_) => self.insert_pending(Pending::Emit(emit_args)),
+      Err(_) => self.insert_pending(Pending::Emit {
+        args: emit_args,
+        filter: filter.map(Box::new),
+      }),
       Ok(lock) => {
         if let Some(handlers) = lock.get(&emit_args.event) {
           let handlers = handlers.iter();
```

tenderdeve:
Closes #15759

`emit_filter`/`emit_str_filter` broadcast to every target when the emit gets deferred onto the pending queue. That deferral happens whenever an emit runs while the handlers lock is held — most commonly emitting from inside an event handler. The queued `Pending::Emit` only stored the payload, so on flush it replayed as a plain unfiltered `emit()`.

Fix: store the filter on `Pending::Emit` and replay it through `emit_filter` on flush.

Because the queue can be flushed from a different thread than the one that deferred the emit, the stored filter has to be `Send`, and boxing it into the queue requires `'static`. So the filter closure now requires `Send + 'static`, threaded through `Emitter::emit_filter`/`emit_str_filter` and the internal manager/listener paths. This matches the bound `Listeners::listen` already places on its callback. Internal callers were updated to own their captured labels; if you'd rather keep the public bound unchanged I can rework, but this mirrors the existing listener API.

Added a regression test (`pending_emit_keeps_filter`) that emits a filtered event from within a handler (forcing the pending path) and asserts only the targeted listener fires.

### DIFF
--- a/.changes/emit-filter-pending.md
+++ b/.changes/emit-filter-pending.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch:bug
+---
+
+Fix `emit_filter`/`emit_str_filter` emitting to every target when the emit is deferred onto the pending queue (which happens when it runs while the handlers lock is held, e.g. emitting from inside an event handler). The filter is now carried through the pending queue instead of degrading into a broadcast. The filter closure now requires `Send + 'static`, matching the bound already required by `Listeners::listen`.
--- a/crates/tauri/src/event/listener.rs
+++ b/crates/tauri/src/event/listener.rs
@@ -16,6 +16,12 @@ use std::{
   },
 };
 
+/// A boxed event target filter that can be stored in the pending queue.
+///
+/// Requires `Send` because the queue may be flushed from a different thread than
+/// the one that deferred the emit.
+type BoxedFilter = Box<dyn Fn(&EventTarget) -> bool + Send>;
+
 /// What to do with the pending handler when resolving it?
 enum Pending {
   Unlisten(EventId),
@@ -24,7 +30,10 @@ enum Pending {
     event: crate::EventName,
     handler: Handler,
   },
-  Emit(EmitArgs),
+  Emit {
+    args: EmitArgs,
+    filter: Option<BoxedFilter>,
+  },
   RemoveForTarget(EventTarget),
 }
 
@@ -128,7 +137,7 @@ impl Listeners {
       match action {
         Pending::Unlisten(id) => self.unlisten(id),
         Pending::Listen { id, event, handler } => self.listen_with_id(id, event, handler),
-        Pending::Emit(args) => self.emit(args)?,
+        Pending::Emit { args, filter } => self.emit_filter(args, filter)?,
         Pending::RemoveForTarget(target) => self.remove_listeners_for_target(target),
       }
     }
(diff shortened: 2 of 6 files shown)


tenderdeve:
Sorry about the noise here — the runtime-wry monitor commits belonged to another branch and got dragged in by accident. Rebased them out, so this is now just the emit-filter change.

On the `Send + 'static` bound you flagged in the issue: the reason it's hard to avoid is that a deferred emit sits in the shared pending queue, and that queue can be drained by a different thread than the one that queued it — so to replay the filter later it has to be both storable (`'static`) and sendable (`Send`). That's the same bound `Listeners::listen` already puts on its callback, which is why I went this way.

If you'd rather not widen the public `emit_filter`/`emit_str_filter` bound, the alternative is to only carry the filter for the same-thread case (emitting from inside a handler, which is the repro) and leave the rare cross-thread defer as the current broadcast. That keeps the public signature untouched but doesn't fully fix the cross-thread path. Happy to switch to that if you prefer — just let me know which direction you want.

sftse:
> On the `Send + 'static` bound you flagged in the issue: the reason it's hard to avoid is that a deferred emit sits in the shared pending queue, and that queue can be drained by a different thread than the one that queued it — so to replay the filter later it has to be both storable (`'static`) and sendable (`Send`). That's the same bound `Listeners::listen` already puts on its callback, which is why I went this way.

@FabianLars @Legend-Master 

Posting chatbot responses without verifying them into a comment is to me unacceptable. It's beneath the dignity of people that care for the project to deal with this mercenary behavior.

Vote to close with prejudice.

Legend-Master:
Fully agree here, I'm closing this one.

@tenderdeve Hey, thanks for consider contributing. I merged the last few agent PRs purely because they were better than I expected from current day LLMs and the problems need fixing. But please review all the code yourself and summarize in your own language before submitting to make it a high quality contribution.

Copying something I said some time ago on Discord:

image

PR #15780, opened 2026-07-26
feat(cli): warn when Java is too new for the bundled Gradle

### LINKED ISSUE #15385: [feat] [android] Add hint if the java and gradle versions are not compatible.
### Describe the problem

Most of the people who try Tauri may not be familiar with the JAVA rules. One very common issue they may face is that they will install the latest Java ( java-26 ), but Tauri uses gradle 8.14.3, which aren't compatible.

In my case, it gave cryptic errors like `Error failed to build APK: Failed to assemble APK with Io...` which wasted a lot of time just wandering here & there.
AI was also helpless here ( may improve in future )

A minimal solution could be that during `tauri android init` or `tauri android build` or `tauri android dev`, check if java in JAVA_HOME is incompatible with the provided gradle as shown in this matrix
https://docs.gradle.org/current/userguide/compatibility.html

| Java version | Support for toolchains | Support for running Gradle |
|--------------|------------------------|----------------------------|
| 8            | N/A                    | 2.0 to 8.14.x              |
| 9            | N/A                    | 4.3 to 8.14.x              |
| 10           | N/A                    | 4.7 to 8.14.x              |
| 11           | N/A                    | 5.0 to 8.14.x              |
| 12           | N/A                    | 5.4 to 8.14.x              |
| 13           | N/A                    | 6.0 to 8.14.x              |
| 14           | N/A                    | 6.3 to 8.14.x              |
| 15           | 6.7                    | 6.7 to 8.14.x              |
| 16           | 7.0                    | 7.0 to 8.14.x              |
| 17           | 7.3                    | 7.3 and after              |
| 18           | 7.5                    | 7.5 and after              |
| 19           | 7.6                    | 7.6 and after              |
| 20           | 8.1                    | 8.3 and after              |
| 21           | 8.4                    | 8.5 and after              |
| 22           | 8.7                    | 8.8 and after              |
| 23           | 8.10                   | 8.10 and after             |
| 24           | 8.14                   | 8.14 and after             |
| 25           | 9.1.0                  | 9.1.0 and after            |
| 26           | 9.4.0                  | 9.4.0 and after            |

Table 1. Java Compatibility

### Alternatives considered

_No response_

### Additional context

_No response_

tenderdeve:
Closes #15385

Installing a very recent JDK (Java 25/26) while Tauri ships Gradle 8.14 makes Android builds fail with a cryptic `Failed to assemble APK with Io...` error that gives no hint about the real cause.

This adds an upfront check to the Android env setup (runs for `android init`/`dev`/`build`): detect the active Java feature version via `java -version` (preferring `JAVA_HOME`) and `log::warn!` when it's newer than the bundled Gradle can run on, pointing at the [Gradle/Java compatibility matrix](https://docs.gradle.org/current/userguide/compatibility.html) and suggesting a supported JDK.

The ceiling is a `MAX_SUPPORTED_JAVA` constant next to the existing `NDK_VERSION`/`SDK_VERSION` constants, to be bumped alongside the template's Gradle version. It's a warning only — nothing is blocked.

Added a unit test for the version parser covering the modern (`21.0.1`), single-number (`26`) and legacy (`1.8.0_292`) schemes.


crates/tauri-cli/src/mobile/android/mod.rs
```diff
@@ -486,9 +492,50 @@ fn ensure_java() -> Result<()> {
...
+/// Detects the major (feature) version of the active Java installation by
+/// running `java -version`, preferring the binary in `JAVA_HOME` if set.
+fn java_major_version() -> Option<u32> {
+  let java = std::env::var_os("JAVA_HOME")
+    .map(|home| PathBuf::from(home).join("bin").join("java"))
+    .filter(|path| path.exists())
+    .unwrap_or_else(|| PathBuf::from("java"));
+
+  // `java -version` writes the version banner to stderr.
+  let output = Command::new(java).arg("-version").output().ok()?;
```
Legend-Master: Not a fan of this to be honest, as it slows down the build

tenderdeve:
@Legend-Master good point on the build overhead — I dropped the `java -version` subprocess in 91eb1e487. It now reads the major version straight from the JDK's `release` file (the `JAVA_VERSION=` line), preferring `JAVA_HOME` and falling back to the `java` on PATH. No process spawn on init/dev/build, and if the file isn't there or doesn't parse it just skips the warning rather than doing anything expensive.

crates/tauri-cli/src/mobile/android/mod.rs
```diff
@@ -50,6 +50,12 @@ mod run;
 
 const NDK_VERSION: &str = "29.0.13846066";
 const SDK_VERSION: u8 = 36;
+/// Highest Java feature version that the Gradle shipped in the Android template
+/// (see `templates/mobile/android/gradle/wrapper/gradle-wrapper.properties`) can
+/// run on. Gradle 8.14 supports running on Java up to 24 — bump this whenever the
+/// template's Gradle version changes.
+/// See <https://docs.gradle.org/current/userguide/compatibility.html>.
+const MAX_SUPPORTED_JAVA: u32 = 24;
```
Legend-Master: We can not do this, the user can update it away from the version we have in the template

tenderdeve:
@Legend-Master good point — you're right that pinning to the template's Gradle version is wrong once the user upgrades their wrapper. Reworked in `7a95ab775`: instead of a hardcoded constant, it now reads the Gradle version straight from the project's `gradle/wrapper/gradle-wrapper.properties` (`distributionUrl`) and only warns when the detected Java is too new for *that* Gradle. If the project isn't generated yet, the file can't be parsed, or it's Gradle 9+ (which tracks new Java quickly), it stays silent — so no false warnings. Still no subprocess spawn.

crates/tauri-cli/src/mobile/android/mod.rs
```diff
@@ -486,9 +486,108 @@ fn ensure_java() -> Result<()> {
...
   }
 
+  // Read the Gradle version this project actually uses (the user may have upgraded it away from
+  // the template default) and only warn when the detected Java is too new for that Gradle.
+  if let (Some(java), Some(gradle_version)) = (java_major_version(), project_gradle_version()) {
+    if let Some(max_java) = gradle_max_supported_java(&gradle_version) {
+      if java > max_java {
+        log::warn!(
+          "Detected Java {java}, but Gradle {gradle_version} used by this project can only run on Java up to {max_java}. \
+           Android builds will likely fail with a cryptic error. Install a supported JDK (Java {max_java} \
```
Legend-Master: This is outdated, give the exact version recommandations

Legend-Master (commented):
Could you give this a spin see how long the whole process takes? Also let's move those checks to a dedicated file.

tenderdeve:
@Legend-Master pushed ec4f795:

- **Dedicated file** — moved `ensure_java` and the Java/Gradle helpers into `crates/tauri-cli/src/mobile/android/java.rs`; `mod.rs` just calls `java::ensure_java()` now. Tests moved with it, plus one for the Gradle→Java mapping.
- **Exact recommendation** — dropped the vague "LTS such as Java 17 or 21" line; the warning now points at the concrete ceiling for the project's Gradle (`Install a JDK that Gradle {gradle_version} supports (Java {max_java} or older)`).
- **Build overhead** — the earlier version's cost was the `java -version` subprocess, which is already gone: detection now reads the `JAVA_VERSION=` line from the JDK's `release` file and the Gradle version from `gradle-wrapper.properties`, so it's two small file reads (sub-millisecond), no process spawn. I don't have a full Android toolchain on this machine to give you an end-to-end `build` timing — happy to if you have a quick way to get numbers on CI, but the added work here is negligible either way.

Legend-Master (approved):

PR #15786, opened 2026-07-27
fix(bundler): respect configured GDK backend (fix: #15781)

### LINKED ISSUE #15781: [bug] AppImage GTK hook overrides an explicitly set GDK_BACKEND, silently disabling Wayland-native functionality
### Describe the bug

The AppImage GTK hook exports `GDK_BACKEND=x11` unconditionally (`crates/tauri-bundler/src/bundle/linux/appimage/linuxdeploy-plugin-gtk.sh`, line 187 on `dev`). I know the history: the line comes from upstream linuxdeploy-plugin-gtk, removing it was tested once and ended in #8541-class crashes, and #11790 records the current answer as "hardcoded, override it from Rust if you need to". This report is about a narrower problem with the current line: it also overrides a `GDK_BACKEND` the user or the app set on purpose, and the silent downgrade to XWayland switches off Wayland-only functionality with no error anywhere.

Concrete case: our app ships a deb and an AppImage of the same binary, and its overlay is a gtk-layer-shell surface. The deb runs the native Wayland backend and works. The AppImage runs XWayland, `gtk_layer_shell_is_supported()` returns false, and the overlay silently degrades to a plain window. Nothing fails loudly in either artifact, so attributing the difference to this one export cost a full debugging session.

We verified the app works on the Wayland backend inside the AppImage by restoring `GDK_BACKEND=wayland` in Rust before the builder runs, the workaround suggested in #11790: layer-shell initializes and the app runs native Wayland (webkit2gtk 2.52.3, Debian 13, KWin 6.3), no crashes.

The ask: make the export a default instead of a force,

```sh
export GDK_BACKEND="${GDK_BACKEND:-x11}"
```

so the crash-avoiding default stays for everyone who does not opt in, and an explicitly set value is honored instead of clobbered. If the legacy linuxdeploy path is frozen while #12491 replaces it, documenting the `std::env::set_var("GDK_BACKEND", "wayland")` escape hatch would already help. I can send a PR for either.

### Reproduction

1. Bundle any app as an AppImage with the Tauri bundler.
2. On a Wayland session, run it with `GDK_BACKEND=wayland` set in the environment.
3. The process still has `GDK_BACKEND=x11`, `xlsclients` lists it as an X11 client, and any gtk-layer-shell call reports the compositor as unsupported even though it supports `zwlr_layer_shell_v1`.

### Expected behavior

An explicitly set `GDK_BACKEND` survives into the app process. The deb of the identical binary already behaves this way.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Debian 13.0.0 x86_64 (X64) (KDE on wayland)
    ✔ webkit2gtk-4.1: 2.52.3
    ✔ rsvg2: 2.60.0
    ✔ rustc: 1.97.0 (2d8144b78 2026-07-07)
    ✔ cargo: 1.97.0 (c980f4866 2026-06-30)
    - node: 22.23.1
    - pnpm: 11.11.0
    - npm: 12.0.1

[-] Packages
    - tauri 🦀: 2.11.5
    - tauri-build 🦀: 2.6.3
    - wry 🦀: 0.55.1
    - tao 🦀: 0.35.3
    - @tauri-apps/api ⱼₛ: 2.11.1
    - @tauri-apps/cli ⱼₛ: 2.11.4
```

### Stack trace

```text
No crash. The silence is the problem: features are absent with no diagnostic.
```

yingyanzhitong:
## Summary

- preserve an explicitly configured `GDK_BACKEND` in the AppImage GTK hook
- retain `x11` as the safe default when the variable is unset
- add a patch release note for `tauri-bundler`

Fixes #15781.

OpenJobs task: https://openjobs.bot/jobs/6cbdfdad-d49f-4eed-a0b9-04df4a7e2b12

## Testing

- `bash -n crates/tauri-bundler/src/bundle/linux/appimage/linuxdeploy-plugin-gtk.sh`
- explicit `GDK_BACKEND=wayland` preservation check
- unset `GDK_BACKEND` fallback-to-`x11` check
- `cargo fmt --check`
- `cargo test -p tauri-bundler`

### DIFF
--- a/.changes/appimage-respect-gdk-backend.md
+++ b/.changes/appimage-respect-gdk-backend.md
@@ -1,0 +1,5 @@
+---
+"tauri-bundler": "patch:bug"
+---
+
+Respect an explicitly configured `GDK_BACKEND` in AppImage GTK hooks while retaining `x11` as the default.
--- a/crates/tauri-bundler/src/bundle/linux/appimage/linuxdeploy-plugin-gtk.sh
+++ b/crates/tauri-bundler/src/bundle/linux/appimage/linuxdeploy-plugin-gtk.sh
@@ -184,7 +184,7 @@ APPIMAGE_GTK_THEME="${APPIMAGE_GTK_THEME:-"Adwaita:$GTK_THEME_VARIANT"}" # Allow
 export APPDIR="${APPDIR:-"$(dirname "$(realpath "$0")")"}" # Workaround to run extracted AppImage
 export GTK_DATA_PREFIX="$APPDIR"
 export GTK_THEME="$APPIMAGE_GTK_THEME" # Custom themes are broken
-export GDK_BACKEND=x11 # Crash with Wayland backend on Wayland - We tested it without it and ended up with this: https://github.com/tauri-apps/tauri/issues/8541
+export GDK_BACKEND="${GDK_BACKEND:-x11}" # Crash with Wayland backend on Wayland - We tested it without it and ended up with this: https://github.com/tauri-apps/tauri/issues/8541
 export XDG_DATA_DIRS="$APPDIR/usr/share:/usr/share:$XDG_DATA_DIRS" # g_get_system_data_dirs() from GLib
 EOF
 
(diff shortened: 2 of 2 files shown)


FabianLars (approved):
yolo (as i said in the issue, this isn't really a good idea but i don't care)

PR #15787, opened 2026-07-27
feat(cef): migrate to winit-gtk4

amrbashir:


FabianLars:
so the v2 plans that came up again recently are officially dead then? just so i know what to communicate

FabianLars:
I assume this is a yes, would be cool if we could start communicating with each other again :) i know i know, communication in the age of AI is unheard but I can still dream.

PR #15788, opened 2026-07-28
chore(renovate): group napi-rs packages take 2

Legend-Master:
Reference #15600, guess it didn't work looking at #15724, let's try manually adding all the packages.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -38,7 +38,7 @@
     {
       "description": "Group napi-rs packages",
       "groupName": "napi-rs packages",
-      "matchSourceUrls": ["https://github.com/napi-rs/napi-rs"]
+      "matchPackageNames": ["napi", "napi-build", "napi-derive", "@napi-rs/cli"]
     }
   ],
   "postUpdateOptions": ["pnpmDedupe"]


FabianLars (approved):

FabianLars:
or maybe it did work? the PRs branch name is correct. Maybe it only changes the PR title when there's more than one matching dependency?

Legend-Master:
Hmm, not sure, I assumed there will be more if it worked, looking at napi crate already at https://crates.io/crates/napi/3.11.0

FabianLars:
doesn't show up in https://github.com/tauri-apps/tauri/issues/2552 at all 🤔

FabianLars:
ah because it doesn't update swmver compatible versions on the rust side

Legend-Master:
because we disabled the lock file maintenance thing a while back? so weird that it does it to the js ones but not the rust ones, maybe because napi cli is not written as a semver?

FabianLars:
no that would get its own pr. it's cause i added https://github.com/tauri-apps/tauri/blob/dev/renovate.json#L6 a while back. i wanna keep that but we could change it for the cli crates.

FabianLars:
> so weird that it does it to the js ones but not the rust ones, maybe because napi cli is not written as a semver?

yeah if we were using `^` versions in js it wouldn't update it until new major

Legend-Master:
> no that would get its own pr. it's cause i added https://github.com/tauri-apps/tauri/blob/dev/renovate.json#L6 a while back. i wanna keep that but we could change it for the cli crates.

Looked about it a bit and seems like you can't config `rangeStrategy` in `packageRules`, we can probably just not to use semver for some of the deps in the cli crates as a work around?

FabianLars:
looks like you can, assuming their example is correct lol https://docs.renovatebot.com/configuration-options/#packagerulesmatchpackagenames

FabianLars:
but yeah we can also solve this by pinning or unpinning the js _and_ rust deps so both sides have the same behavior

Legend-Master:
> looks like you can, assuming their example is correct lol https://docs.renovatebot.com/configuration-options/#packagerulesmatchpackagenames

Guess we can give it a try 😂 it's not in the sidebar nor in the json schema so idk

PR #15790, opened 2026-07-28
chore(deps): update `objc2-*` crates to 0.3.2

Legend-Master:
In preparation for https://github.com/tauri-apps/wry/pull/1737

### DIFF
--- a/crates/tauri-runtime/Cargo.toml
+++ b/crates/tauri-runtime/Cargo.toml
@@ -51,7 +51,7 @@ jni = "0.21"
 
 [target.'cfg(all(target_vendor = "apple", not(target_os = "macos")))'.dependencies]
 objc2 = "0.6"
-objc2-ui-kit = { version = "0.3.0", default-features = false, features = [
+objc2-ui-kit = { version = "0.3.2", default-features = false, features = [
   "UIView",
   "UIResponder",
   "UIScene",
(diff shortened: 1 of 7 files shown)


crates/tauri-runtime/Cargo.toml
```diff
@@ -61,7 +61,7 @@ objc2-ui-kit = { version = "0.3.0", default-features = false, features = [
 [target."cfg(target_os = \"macos\")".dependencies]
 url = "2"
 objc2 = "0.6"
-objc2-web-kit = { version = "0.3", default-features = false, features = [
+objc2-web-kit = { version = "0.3.2", default-features = false, features = [
```
Legend-Master, line 64: Not sure if it's needed for the `tauri` and `tauri-runtime` crates, but just bumped them anyways that we have `tauri-runtime-wry` on `0.3.2` anyways.

Edit: yeah we need it I guess 04125a4

FabianLars (approved):

crates/tauri-runtime/Cargo.toml
sftse: AFAIK this is a bit of a footgun in Rust. If the version is above `1.0.0` I think it's legitimate to leave out the patch number, but in `0.x.y` versions I think there's less guidance on what the patch number means? In `1.x.y` versions it's possible to pick a misleading minor version `x` and not notice it because Cargo defaults to the most recent version, so may pick `1.9.0` even if we put `1.7.0` in the `Cargo.toml`. People that do try to compile it with `1.7.0` can have a broken build.

Not sure if that is related to your question, but it did catch me off guard once.

Legend-Master: Patch versions are to indicate both patch and minor versions for pre 1.0 versions I think

> https://doc.rust-lang.org/cargo/reference/semver.html#:~:text=Initial%20development%20releases%20starting%20with%20%E2%80%9C0.y.z%E2%80%9D%20can%20treat%20changes%20in%20%E2%80%9Cy%E2%80%9D%20as%20a%20major%20release%2C%20and%20%E2%80%9Cz%E2%80%9D%20as%20a%20minor%20release.

PR #15798, opened 2026-07-30
fix(android): use whichever activity in plugin manager

### LINKED ISSUE #15506: [bug] Android `requestPermissions` crashes with unregistered `ActivityResultLauncher`
### Describe the bug

I'm building an Android app with Tauri 2 and hit a crash when requesting runtime permissions (SMS, phone, location)

Digging through the Android plugin source, I think the cause is in `PluginManager.kt`. It's a singleton `object`, and `onActivityCreate` only does its setup once:

https://github.com/tauri-apps/tauri/blob/7af245c60b3e7620158499db06545e284461925b/crates/tauri/mobile/android/src/main/java/app/tauri/plugin/PluginManager.kt#L58-L65

Android can recreate an app's activity after launch (screen rotation, light/dark mode switch, the system killing it in the background, etc.). When that happens, `onActivityCreate` runs again with the new activity, but the early return means it keeps the launchers it registered for the old, destroyed activity. The next time the app requests permissions (or uses `startActivityForResult`/`startIntentSenderForResult`), it crashes because those launchers no longer belong to a live activity.

There's already a `// TODO: on destroy, we should change to a different activity` comment right above that early return, so this looks like a known gap, but I just wanted to flag it

### Reproduction

1. Build a Tauri 2 Android app with a screen that has a button that calls your plugin's `requestPermissions` command
2. Run it on a device with `tauri android dev`
3. When the app open, tap the button
4. Instead of the system dialog, it crashes with

```
java.lang.IllegalStateException: Attempting to launch an unregistered ActivityResultLauncher with contract androidx.activity.result.contract.ActivityResultContracts$RequestMultiplePermissions and input [...]. You must ensure the ActivityResultLauncher is registered before calling launch().
```

### Expected behavior

The permission request should still work and show the system dialog even if the activity's been recreated since launch

Looks like `onActivityCreate` just needs to update its activity reference and re-register the launchers whenever it's given a new activity, instead of bailing out early after the first run

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.26200 x86_64 (X64)
    ✔ WebView2: 148.0.3967.96
    ✔ MSVC: Visual Studio Build Tools 2026
    - node: 22.12.0
    - pnpm: 10.30.3
    - npm: 10.8.2
    - deno: deno 2.1.2

[-] Packages
    - tauri 🦀: 2.11.1, (outdated, latest: 2.11.2)
    - tauri-build 🦀: 2.6.1, (outdated, latest: 2.6.2)
    - wry 🦀: 0.55.1
    - tao 🦀: 0.35.2, (outdated, latest: 0.35.3)
    - @tauri-apps/api  ⱼₛ: 2.11.0
    - @tauri-apps/cli  ⱼₛ: 2.11.1 (outdated, latest: 2.11.2)

[-] Plugins
    - @tauri-apps/plugin-opener  ⱼₛ: 2.5.4
    - tauri-plugin-store 🦀: 2.4.3
    - @tauri-apps/plugin-store  ⱼₛ: 2.4.3

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - framework: React
    - bundler: Vite
```

### Stack trace

```text
java.lang.IllegalStateException: Attempting to launch an unregistered ActivityResultLauncher with contract androidx.activity.result.contract.ActivityResultContracts$RequestMultiplePermissions and input [android.permission.SEND_SMS, android.permission.CALL_PHONE, android.permission.ACCESS_FINE_LOCATION, android.permission.ACCESS_COARSE_LOCATION]. You must ensure the ActivityResultLauncher is registered before calling launch().


This is the message that reached my frontend — not the full native trace. Happy to grab one with `adb logcat` next time I reproduce it if that'd help
```

### Additional context

_No response_

Legend-Master:
Fix #15506

### DIFF
--- a/crates/tauri/mobile/android/src/main/java/app/tauri/plugin/PluginManager.kt
+++ b/crates/tauri/mobile/android/src/main/java/app/tauri/plugin/PluginManager.kt
@@ -34,11 +34,12 @@ object PluginManager {
     fun onResult(result: ActivityResult)
   }
 
-  lateinit var activity: AppCompatActivity
+  private val activities: HashSet<AppCompatActivity> = HashSet()
+  var activity: AppCompatActivity? = null
   private val plugins: HashMap<String, PluginHandle> = HashMap()
-  private lateinit var startActivityForResultLauncher: ActivityResultLauncher<Intent>
-  private lateinit var startIntentSenderForResultLauncher: ActivityResultLauncher<IntentSenderRequest>
-  private lateinit var requestPermissionsLauncher: ActivityResultLauncher<Array<String>>
+  private var startActivityForResultLauncher: ActivityResultLauncher<Intent>? = null
+  private var startIntentSenderForResultLauncher: ActivityResultLauncher<IntentSenderRequest>? = null
+  private var requestPermissionsLauncher: ActivityResultLauncher<Array<String>>? = null
   private var requestPermissionsCallback: RequestPermissionsCallback? = null
   private var startActivityForResultCallback: ActivityResultCallback? = null
   private var startIntentSenderForResultCallback: ActivityResultCallback? = null
@@ -56,11 +57,15 @@ object PluginManager {
   }
 
   fun onCreate(activity: AppCompatActivity) {
-    // TODO: on destroy, we should change to a different activity
-    if (::activity.isInitialized) {
-      return
+    // Record the activity, and if that's the only activity we got, register result launchers
+    activities.add(activity)
+    if (this.activity == null) {
+      this.activity = activity
+      registerResultLaunchers(activity)
     }
-    this.activity = activity
+  }
+
+  private fun registerResultLaunchers(activity: AppCompatActivity) {
     startActivityForResultLauncher =
       activity.registerForActivityResult(ActivityResultContracts.StartActivityForResult()
       ) { result ->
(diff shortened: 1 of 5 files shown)


crates/tauri/mobile/android/src/main/java/app/tauri/plugin/PluginManager.kt
```diff
@@ -120,6 +125,20 @@ object PluginManager {
     for (plugin in plugins.values) {
       plugin.instance.triggerOnDestroy(activity)
     }
+
+    activities.remove(activity)
+    val nextActivity = activities.firstOrNull()
```
FabianLars, line 130: isn't this potentially the wrong order?

like if i go act 1 > 2 > 3, navigate back to 1 this could give me act 1 instead of 2, no?
assuming act 2 is kept alive after going to act 3 which i'm not sure it is

FabianLars: ah with the onCreate changes this may not matter because the result handler is likely still on act 1 anyway?

Legend-Master: Our entire Android plugin system seems to built on the assumption of a single activity so this is really just a fix for that use case. The multi-activity setup is too broken to fit in that there're really not much can be done here.

FabianLars: meh ok

FabianLars (approved):

PR #15800, opened 2026-07-30
fix(core): guard event unlisten script when listener entry is missing

### LINKED ISSUE #15799: `unlisten` throws and leaks the listener when called before the registration eval lands (`listeners[eventId]` undefined)
### Describe the bug

`listen()` resolves as soon as Rust answers the `plugin:event|listen` invoke, but the webview-side registry entry is written by a **separate** eval (`listen_js_script`). Calling the returned unlisten function inside that window crashes:

```
TypeError: undefined is not an object (evaluating 'listeners[eventId].handlerId')
```

The generated unlisten script guards the per-event object but not the entry itself (`unlisten_js_script`, `crates/tauri/src/event/mod.rs`):

```js
const listeners = (window['__TAURI_EVENT_LISTENERS__'] || {})[event]
if (listeners) {
  window.__TAURI_INTERNALS__.unregisterCallback(listeners[eventId].handlerId) // ← throws
}
```

— in contrast with event delivery (`event_initialization_script` in the same file), which does guard the entry: `const listener = listeners[id]; if (listener) { … }`.

The consequence is worse than the crash itself: the throw happens on the **first** line of `_unlisten` in `@tauri-apps/api/event`, so the `invoke('plugin:event|unlisten', …)` on its second line never runs. The backend listener stays registered and the JS callback stays alive — the "removed" handler keeps firing once the registration eval lands, and after a remount the event is handled twice. With React StrictMode (mount → unmount → mount in dev) this window is hit on essentially every fast unsubscribe; we saw it as a steady stream of unhandled rejections in Sentry plus double-delivered events.

### Reproduction steps

```ts
import { listen } from "@tauri-apps/api/event"

// Unlisten immediately after listen() resolves — before the registration
// eval reaches the webview:
const un = await listen("my-evt", console.log)
un() // → TypeError above + the backend listener is leaked
```

Any React StrictMode effect that subscribes on mount and unsubscribes in cleanup reproduces this via the dev-mode mount→unmount→mount cycle.

### Expected behavior

Unlisten tolerates a not-yet-registered (or already-gone) entry: guard the lookup (`listeners[eventId]?.handlerId`, skipping `unregisterCallback` when absent) — and still send the `plugin:event|unlisten` invoke so the backend side is cleaned up either way.

### Platform and versions

- tauri `2.11.5` (latest on crates.io); the unguarded line is unchanged on `dev`
- @tauri-apps/api `2.11.1`
- macOS 26.5 / WKWebView (the lookup itself is platform-independent)

lazerg:
`listen()` resolves once Rust answers its invoke, but the webview registry entry is written by a separate, later eval, so the returned unlisten function can run before that entry exists. In that window the generated unlisten script read `handlerId` off an undefined entry and threw, and because that throw happens before `_unlisten` sends the backend `plugin:event|unlisten`, the listener stayed registered and kept firing. This checks the entry is present before reading it, the same guard the event delivery script already uses.

Fixes #15799

### DIFF
--- a/crates/tauri/src/event/mod.rs
+++ b/crates/tauri/src/event/mod.rs
@@ -214,7 +214,10 @@ pub(crate) fn unlisten_js_script(
     "(function () {{
         const listeners = (window['{listeners_object_name}'] || {{}})[{event_arg}]
         if (listeners) {{
-          window.__TAURI_INTERNALS__.unregisterCallback(listeners[{event_id_arg}].handlerId)
+          const listener = listeners[{event_id_arg}]
+          if (listener) {{
+            window.__TAURI_INTERNALS__.unregisterCallback(listener.handlerId)
+          }}
         }}
       }})()
     ",
@@ -248,5 +251,13 @@ mod tests {
       .unwrap_err()
       .to_string();
     assert_eq!("only alphanumeric, '-', '/', ':', '_' permitted for event names: \"some\\r illegal event name\"", s);
+  }
+
+  #[test]
+  fn unlisten_script_guards_missing_entry() {
+    let script = unlisten_js_script("__listeners__", "event", "eventId");
+    assert!(!script.contains("listeners[eventId].handlerId"));
+    assert!(script.contains("const listener = listeners[eventId]"));
+    assert!(script.contains("if (listener) {"));
   }
 }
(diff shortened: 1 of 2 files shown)


lazerg:
Hey @Legend-Master, just checking in on this fix. Let me know if you would like any changes.

crates/tauri/src/event/mod.rs
```diff
@@ -249,4 +252,12 @@ mod tests {
       .to_string();
     assert_eq!("only alphanumeric, '-', '/', ':', '_' permitted for event names: \"some\\r illegal event name\"", s);
   }
+
+  #[test]
```
amrbashir: this is a useless test, let's remove

lazerg: Removed in 6a8060e.

amrbashir (approved):

PR #15808, opened 2026-08-03
ci: enable renovate updates on cli packages

Legend-Master:
Reference https://github.com/tauri-apps/tauri/pull/15788#issuecomment-5163541893

PR #15811, opened 2026-08-03
ci: remove renovate cli group name

Legend-Master:
Follow up of #15808, not sure if this helps at all looking at the gigantic #15809

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -42,7 +42,6 @@
     },
     {
       "description": "Bump CLI packages",
-      "groupName": "cli packages",
       "matchFileNames": [
         "crates/tauri-cli/Cargo.toml",
         "packages/cli/Cargo.toml",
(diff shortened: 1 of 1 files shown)


FabianLars:
looking at the list, maybe we shouldn't turn _all_ deps to bump? that'll be super spammy over time...

Legend-Master:
yeah... but the alternatives being just update some deps or update lots of them at once which isn't super nice either

FabianLars:
yeah but we don't have to update every patch version every day. would a lockfile only update for all but napi packages be better? not much ig

Legend-Master:
we could give it a try (not sure if one overrides the other 😂)

FabianLars (dismissed):
iirc auto is different for npm and cargo

FabianLars (approved):

PR #15812, opened 2026-08-03
refactor(tauri-runtime-wry)!: call tao windowing functions directly

Legend-Master:
Since `tao` `Window` APIs are already thread safe, we can move the calls out to reduce the complexities whenever possible.

Since we have quite a bit of work arounds inside tauri instead of tao, this might cause some unexpected behaviors, some early feedbacks would be much appreciated.

Disclosure: initial migration was done through codex.


sftse:
Nice find!

> Since we have quite a bit of work arounds inside tauri instead of tao, this might cause some unexpected behaviors, some early feedbacks would be much appreciated.

Can you state what workarounds to watch out for?

Legend-Master:
Notably `undecorated_resizing` but honestly, some of the tao off main thread handlings might not be ideal, like for example the windows one panics if the event loop is closed: https://github.com/tauri-apps/tao/blob/36b31b795f71ea52528b0d65de86c54645f3634d/src/platform_impl/windows/event_loop.rs#L519-L546

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -2130,80 +2027,58 @@ impl<T: UserEvent> WindowDispatch<T> for WryWindowDispatcher<T> {
   }
 
   fn set_enabled(&self, enabled: bool) -> Result<()> {
-    send_user_message(
-      &self.context,
-      Message::Window(self.window_id, WindowMessage::SetEnabled(enabled)),
-    )
+    self.window()?.set_enabled(enabled);
```
Legend-Master, line 1949: There's small behavior change here, the method succeeds if called off main thread while after the event loop destroyed instead of returning an error now. That being said, that error is not super useful anyways and silently failing after event loop's gone should not be a big deal

Legend-Master:
@lucasfernog a176c8e doesn't make much sense to me that the whole point of this PR is to remove the need of this, no?

PR #15821, opened 2026-08-04
fix(runtime-wry): return the expected exit code from run_return()

aurelj:
Transfer the exit code from the window.app_handle().exit(1) call to the run_return() result instead of always returning 0.


Legend-Master (dismissed):

aurelj:
> Thanks, could you also add a change file?
> 
> > https://github.com/tauri-apps/tauri/blob/dev/.changes/README.md

Yes, sure. That's done.

.changes/exit-with-code.md
```diff
@@ -0,0 +1,5 @@
+---
+'tauri-runtime-wry': 'patch:bug'
```
Legend-Master, line 3:
```suggestion
'tauri': 'patch:bug'
'tauri-runtime-wry': 'patch:bug'
```

So it also shows up in tauri's changelog

Legend-Master (approved):
Thanks!

PR #15828, opened 2026-08-05
fix(android): update template to use gradle 9

Legend-Master:
Built on top of #15335 as a non-breaking version for v2


Legend-Master:
@FabianLars do you have the patient to verify this on an older `tauri android init` project? I have tested the current version against the existing dialog plugin without any problems, but I really don't have the mental power to deal with any java today anymore 😂

FabianLars:
so basically just testing tauri-build? no errors, seems to work.

Legend-Master:
Yeah.. maybe also test it in `plugins-workspace`?

Also super confused about the `Supplied consumer proguard configuration does not exist` error, not sure what change caused it 😩

FabianLars:
yeah i see no errors in plugins-workspace either, everything seems to work correctly.

after rerunning android init i can't spot the proguard error you mentioned but it may be hidden in the shitton of deprecation warnings 😅

Legend-Master:
That error only seem to appear in release build 

> https://github.com/tauri-apps/tauri/actions/runs/30993665742/job/92265472029?pr=15828

Legend-Master:
Ok, so #6257 broke everything... It removed the `consumer-rules.pro` that is used in our setup and kept the `proguard-rules.pro` which is only used when compiling the library to AAR (which we don't do). (See a48ec22 for the correct setup)

This sucks so much now that all of the plugins generated before were consistently wrong that we can't really do this without some hacky work around now. 😢

crates/tauri-cli/templates/mobile/android/app/build.gradle.kts
```diff
@@ -37,20 +37,27 @@ android {
...
-                {{~#each abi-list}}
+                {{#each abi-list}}
                 jniLibs.keepDebugSymbols.add("*/{{this}}/*.so")
                 {{/each}}
             }
         }
         getByName("release") {
-            isMinifyEnabled = true
+            optimization {
+               enable = true
```
Legend-Master, line 47: This also includes `getDefaultProguardFile("proguard-android-optimize.txt")`

> https://developer.android.com/topic/performance/app-optimization/enable-app-optimization#optimization-dsl

Legend-Master:
So weird that I can't reproduce this CI error locally 🤔 @FabianLars could you reproduce this error locally?

> https://github.com/tauri-apps/tauri/actions/runs/31077292683/job/92538010315?pr=15828

Legend-Master:
I have tested this PR against `plugins-workspace` without https://github.com/tauri-apps/plugins-workspace/pull/3531 and it seems to work both on release build and dev build, also tested with `tauri android init` in the API example. I still couldn't reproduce the error from CI though... I'm on JDK 25 + NDK 29 if that's useful.

Legend-Master:
After crap ton of debugging, it might be the `apply(from = "tauri.build.gradle.kts")` that's causing the problem... Like wtf

> https://issuetracker.google.com/issues/463283604

Edit: yeah, confirmed, it needs to be `apply(from = file("tauri.build.gradle.kts"))` now, fixed in adf914a

> reference: https://issuetracker.google.com/issues/463283604#comment10

FabianLars (approved):

PR #15831, opened 2026-08-05
fix(build): resolve target dir from OUT_DIR on new nightly

### LINKED ISSUE #15830: [bug] target dir computation incorrect with nightly toolchain
### Describe the bug

Owing to the unfortunate and still unresolved https://github.com/rust-lang/cargo/issues/9661, tauri-build uses a simple calculation to determine the target dir to which, among other things, sidecar binaries should be copied. This calculation assumes that the target dir is three levels above `OUT_DIR`. On the nightly toolchain, this is no longer always true.

### Reproduction

Creating a simple sidecar example is necessary to observe the effect on sidecars. Such an example can be found [here](https://github.com/mweber15/sidecar-example). Note that building this with the stable toolchain will perform the copy of the sidecar binary to the correct location, in which case subsequent builds with nightly will still work. Either don't build this with stable, or make sure to `cargo clean` the crate build after building with stable to observe the problem with nightly.

Alternatively, the difference in `OUT_DIR` can be easily observed by building any crate with very verbose output. To make it even clearer, add a simple build script that outputs the variable.

```rust
fn main() {
    eprintln!("OUT_DIR={}", std::env::var("OUT_DIR").unwrap());
}
```
```
❯ cargo +stable -vv build
...
[empty 0.1.0] OUT_DIR=/private/tmp/empty/target/debug/build/empty-63ba68eead531e35/out
...
```
```
❯ cargo +nightly -vv build
...
[empty 0.1.0] OUT_DIR=/private/tmp/empty/target/debug/build/empty/a1debc992cc1b384/out
...
```

Specifically to the sidecar issue, the resulting difference in where the binaries land is what you would expect:

```
❯ pwd
/tmp/sidecar-example/src-tauri
❯ cargo +stable build
❯ fd -Ig hello
target/debug/hello
❯ cargo clean
❯ cargo +nightly build
❯ fd -Ig hello
target/debug/build/hello
```

### Expected behavior

The computation of target dir needs to be more resilient to changes to the `OUT_DIR` format so that both the stable and nightly toolchains produce correct output.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 26.6.0 arm64 (X64)
    ✔ Xcode: 26.6
    ✔ rustc: 1.97.1 (8bab26f4f 2026-07-14)
    ✔ cargo: 1.97.1 (c980f4866 2026-06-30)
    ✔ Rust toolchain: stable-aarch64-apple-darwin (overridden by environment variable RUSTUP_TOOLCHAIN)
    - node: 22.23.1
    - yarn: 1.22.22
    - npm: 10.9.8

[-] Packages
    - tauri 🦀: 2.11.5
    - tauri-build 🦀: 2.6.3
    - wry 🦀: 0.55.1, (outdated, latest: 0.56.0)
    - tao 🦀: 0.35.3, (outdated, latest: 0.36.0)
    - tauri-cli 🦀: 2.9.6 (outdated, latest: 2.11.4)

[-] Plugins
    - tauri-plugin-shell 🦀: 2.3.5

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../src
```

### Stack trace

```text

```

### Additional context

_No response_

lazerg:
`tauri-build` assumed the target dir sits exactly three levels above `OUT_DIR`. Recent nightly toolchains changed the build script output layout from `build/<pkg>-<hash>/out` to `build/<pkg>/<hash>/out`, so that assumption pointed one level too deep and sidecars and resources were copied into `target/debug/build` instead of `target/debug`.

It now walks up from `OUT_DIR` to the `build` directory and takes its parent, which works on both layouts.

Verified with the reproduction from the issue: on nightly the sidecar now lands in `target/debug/hello` again, and stable is unaffected.

Closes #15830

### DIFF
--- a/crates/tauri-build/src/lib.rs
+++ b/crates/tauri-build/src/lib.rs
@@ -21,7 +21,9 @@ use tauri_utils::{
 
 use std::{
   collections::HashMap,
-  env, fs,
+  env,
+  ffi::OsStr,
+  fs,
   path::{Path, PathBuf},
 };
 
@@ -203,6 +205,15 @@ fn copy_frameworks(dest_dir: &Path, frameworks: &[String]) -> Result<()> {
     }
   }
   Ok(())
+}
+
+// resolves the target dir from `OUT_DIR`, which is
+// `<target-dir>/build/<crate dir>/out` - one crate dir on stable, two on nightly.
+fn target_dir_from_out_dir(out_dir: &Path) -> Option<&Path> {
+  out_dir
+    .ancestors()
+    .find(|path| path.file_name() == Some(OsStr::new("build")))
+    .and_then(|build_dir| build_dir.parent())
 }
 
 // creates a cfg alias if `has_feature` is true.
@@ -574,13 +585,8 @@ pub fn try_build(attributes: Attributes) -> Result<()> {
   env::set_var("TAURI_ENV_TARGET_TRIPLE", &target_triple);
 
   // TODO: far from ideal, but there's no other way to get the target dir, see <https://github.com/rust-lang/cargo/issues/5457>
-  let target_dir = out_dir
-    .parent()
-    .unwrap()
-    .parent()
-    .unwrap()
-    .parent()
-    .unwrap();
+  let target_dir = target_dir_from_out_dir(&out_dir)
+    .with_context(|| format!("failed to resolve the target directory from {out_dir:?}"))?;
 
   if let Some(paths) = &config.bundle.external_bin {
     copy_binaries(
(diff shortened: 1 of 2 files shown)


crates/tauri-build/src/lib.rs
```diff
@@ -574,13 +585,8 @@ pub fn try_build(attributes: Attributes) -> Result<()> {
   env::set_var("TAURI_ENV_TARGET_TRIPLE", &target_triple);
 
   // TODO: far from ideal, but there's no other way to get the target dir, see <https://github.com/rust-lang/cargo/issues/5457>
```
Legend-Master: Move this to above `target_dir_from_out_dir`

crates/tauri-build/src/lib.rs
```diff
@@ -205,6 +207,15 @@ fn copy_frameworks(dest_dir: &Path, frameworks: &[String]) -> Result<()> {
   Ok(())
 }
 
+// resolves the target dir from `OUT_DIR`, which is `<target dir>/build/<crate dirs>/out`.
```
Legend-Master: let's document this like you did in the PR description

Legend-Master (dismissed):
Thanks! Just a bit of nitpicks and we can get this merged.

crates/tauri-build/src/lib.rs
lazerg: Moved

crates/tauri-build/src/lib.rs
lazerg: Done.

Legend-Master (approved):

PR #15838, opened 2026-08-06
fix(core): correct asset protocol multi-range responses (fix: #15837)

### LINKED ISSUE #15837: [bug] asset protocol multi-range responses: missing close delimiter, duplicate Content-Type, and 200 instead of 206
### Describe the bug

Three defects on the multi-range branch of the `asset://` protocol handler, `crates/tauri/src/protocol/asset.rs` on `dev` (`c0bd0d5`). They only affect requests carrying a `Range` header with more than one range, which is why they have gone unnoticed — single-range requests, which is what `<video>`/`<audio>` normally send, are fine.

**1. The multipart body has no closing delimiter.** Lines 175-176:

```rust
let boundary_sep = format!("\r\n--{boundary}\r\n");
let boundary_closer = format!("\r\n--{boundary}\r\n");
```

The two are identical. RFC 2046 §5.1.1 requires the close-delimiter to be `--{boundary}--`, and `boundary_closer` written at line 201 is the opening form. A strict parser sees the body as ending mid-part.

**2. The response carries two `Content-Type` headers.** Line 94 sets the sniffed file MIME type before the code knows whether this is a multi-range request:

```rust
resp = resp.header(CONTENT_TYPE, &mime_type);
```

and line 178 then sets `multipart/byteranges; boundary=...`. `http::response::Builder::header` calls `HeaderMap::try_append`, which appends rather than replaces, so both survive and the file's own MIME type comes first. A client that reads the first `Content-Type` parses the multipart envelope as an image.

**3. The status is 200, not 206.** The single-range branch sets `StatusCode::PARTIAL_CONTENT` at line 148; the multi-range branch never sets a status, so it falls through to `200 OK` while returning a partial, multipart body. RFC 9110 §15.3.7 requires 206 for a successful range request.

### Reproduction

I hit these from a unit test rather than an app, so instead of a repro repo here is the observed output — a 1000-byte file requested with `Range: bytes=0-9, 20-29` through `protocol::asset::get(..)` on `dev`:

```
status = 200 OK
content-type headers = ["application/octet-stream",
                        "multipart/byteranges; boundary=9d973c59b7a9648f1726b5f7abaefe63795535add85a022374b80d97a"]
body tail = "...000\r\n\r\n<10 bytes>\r\n--9d973c59b7a9648f1726b5f7abaefe63795535add85a022374b80d97a\r\n"
```

All three are visible in that one response: status 200, two `Content-Type`s with the sniffed type first, and a trailing `--boundary\r\n` where `--boundary--` belongs.

### Expected behavior

A multi-range request should return `206 Partial Content`, exactly one `Content-Type: multipart/byteranges; boundary=...`, and a body terminated with `\r\n--{boundary}--\r\n`.

### Full `tauri info` output

```text
Read from source on the dev branch at c0bd0d5 rather than from an app, so there is
no meaningful `tauri info`. The relevant versions:

  - tauri (crates/tauri): 2.11.5 (workspace dev branch, commit c0bd0d5)
  - http: 1.4.0  (the try_append behaviour in defect 2)
  - http-range: 0.1.5
```

### Stack trace

_No response_

### Additional context

I found these while working on #15836, which moves the same handler off the calling thread. I deliberately left all three out of that PR so the threading change stays reviewable on its own — they are pre-existing and independent of it, and I confirmed the range paths are byte-identical before and after that change.

Happy to send a PR for these if you would like one.

lazerg:
Multi-range `asset://` responses had three problems. The multipart body was terminated with an opening boundary (`--boundary`) instead of the close-delimiter RFC 2046 asks for (`--boundary--`). The sniffed file mime type set before the range branch stayed on the response alongside `multipart/byteranges`, because `Builder::header` appends rather than replaces, so clients reading the first `Content-Type` parsed the envelope as the file itself. And the status stayed `200 OK` while the body was partial.

Single-range requests, which is what `<video>` and `<audio>` normally send, were never affected, so this only shows up on clients asking for more than one range. Added a unit test covering all three.

Closes #15837.

### DIFF
--- a/crates/tauri/src/protocol/asset.rs
+++ b/crates/tauri/src/protocol/asset.rs
@@ -169,7 +169,7 @@ fn get_response(
       let boundary_sep = format!("\r\n--{boundary}\r\n");
       let boundary_closer = format!("\r\n--{boundary}--\r\n");
 
-      // replace the sniffed file mime type set earlier, `Builder::header` would append instead
+      // `Builder::header` appends, so replace the file mime type set earlier
       if let Some(headers) = resp.headers_mut() {
         headers.insert(
           CONTENT_TYPE,
@@ -205,6 +205,7 @@ fn get_response(
 
         buf
       };
+
       resp = resp.status(StatusCode::PARTIAL_CONTENT);
       resp.body(buf.into())
     }
@@ -252,7 +253,10 @@ mod tests {
   fn multi_range_request() {
     let app = crate::test::mock_app();
 
-    let path = std::env::temp_dir().join("tauri-asset-protocol-multi-range.bin");
+    let path = std::env::temp_dir().join(format!(
+      "tauri-asset-protocol-multi-range-{}.bin",
+      std::process::id()
+    ));
     std::fs::write(&path, vec![b'a'; 1000]).unwrap();
 
     let scope = Scope::new(&app, &FsScope::default()).unwrap();
@@ -271,6 +275,7 @@ mod tests {
       .unwrap();
 
     let response = get_response(request, &scope, "http://tauri.localhost").unwrap();
+    std::fs::remove_file(&path).unwrap();
 
     assert_eq!(response.status(), StatusCode::PARTIAL_CONTENT);
 
--- a/crates/tauri/src/protocol/asset.rs
+++ b/crates/tauri/src/protocol/asset.rs
@@ -167,12 +167,15 @@ fn get_response(
 
       let boundary = random_boundary();
       let boundary_sep = format!("\r\n--{boundary}\r\n");
-      let boundary_closer = format!("\r\n--{boundary}\r\n");
+      let boundary_closer = format!("\r\n--{boundary}--\r\n");
 
-      resp = resp.header(
-        CONTENT_TYPE,
-        format!("multipart/byteranges; boundary={boundary}"),
-      );
+      // replace the sniffed file mime type set earlier, `Builder::header` would append instead
+      if let Some(headers) = resp.headers_mut() {
+        headers.insert(
+          CONTENT_TYPE,
+          HeaderValue::from_str(&format!("multipart/byteranges; boundary={boundary}"))?,
+        );
+      }
 
       let buf = {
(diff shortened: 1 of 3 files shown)


lazerg:
Hey @Legend-Master, could you review this fix for the asset protocol's multi-range responses? It's been open a month.

crates/tauri/src/protocol/asset.rs
```diff
@@ -167,12 +167,15 @@ fn get_response(
...
       let boundary = random_boundary();
       let boundary_sep = format!("\r\n--{boundary}\r\n");
-      let boundary_closer = format!("\r\n--{boundary}\r\n");
+      let boundary_closer = format!("\r\n--{boundary}--\r\n");
 
-      resp = resp.header(
-        CONTENT_TYPE,
-        format!("multipart/byteranges; boundary={boundary}"),
-      );
+      // `Builder::header` appends, so replace the file mime type set earlier
```
amrbashir:
```suggestion
      // `Builder::header` appends, we want to replace the file mime type set earlier
```

amrbashir (commented):
Thank you. This LGTM, one small nit and could you also update `examples/streaming/main.rs`, it have the same issue.

crates/tauri/src/protocol/asset.rs
lazerg: Thanks, both done. I took the comment wording as suggested. `examples/streaming/main.rs` had all three of the same bugs, so it now closes the multipart body with `--boundary--`, replaces the `video/mp4` Content-Type instead of appending a second one, and returns 206 on the multi-range branch.

amrbashir (approved):
Thank you

amrbashir:
@lazerg you need to sign all your commits in this PR, otherwise I can't merge it. [94ad6e1](https://github.com/tauri-apps/tauri/pull/15838/commits/94ad6e14f4846d51728d361b0a540f0fcb568f4b) was not signed.

lazerg:
Done, re-signed all commits.

PR #15844, opened 2026-08-08
fix(macos-sign): import the signing certificate as an explicit pkcs12 (fix: #15843)

### LINKED ISSUE #15843: [bug] iOS PKCS#12 certificate import fails on macOS 26 when import format is implicit
### Describe the bug

`tauri ios build` with `IOS_CERTIFICATE` and `IOS_CERTIFICATE_PASSWORD` fails while importing a valid Apple Distribution PKCS#12 on a GitHub-hosted `macos-26` runner (Xcode 26.6). The identical decoded P12 and password import successfully immediately beforehand when `security import` is given explicit PKCS#12 type/format flags.

Tauri CLI 2.11.4 reaches `tauri_macos_sign::Keychain::with_certificate_file`, whose command does not pass `-f pkcs12 -t cert`:

https://github.com/tauri-apps/tauri/blob/34ec18ba5e1acabebd66ae79d6fc746f63d8eb96/crates/tauri-macos-sign/src/keychain.rs

This appears to be a macOS 26 import-format inference regression. The resulting message says the password may be wrong even though an explicit import with that password has just succeeded.

### Reproduction

On a `macos-26` GitHub Actions runner with a valid password-protected Apple Distribution P12:

```bash
printf %s "$IOS_CERTIFICATE" | base64 --decode > distribution.p12
security create-keychain -p "$EPHEMERAL_PASSWORD" release.keychain-db
security unlock-keychain -p "$EPHEMERAL_PASSWORD" release.keychain-db

# Succeeds with the exact certificate/password.
security import distribution.p12 \
  -P "$IOS_CERTIFICATE_PASSWORD" -A -t cert -f pkcs12 \
  -k release.keychain-db

# In a fresh keychain, the equivalent implicit-format import used by Tauri fails.
security import distribution.p12 \
  -P "$IOS_CERTIFICATE_PASSWORD" \
  -T /usr/bin/codesign -T /usr/bin/pkgbuild -T /usr/bin/productbuild \
  -k another.keychain-db
```

Then a Tauri iOS build with `IOS_CERTIFICATE` and `IOS_CERTIFICATE_PASSWORD` fails in the same command path. Public failing job: https://github.com/free2z/zuu/actions/runs/31248731230/job/93081562145

### Expected behavior

Tauri should import a valid PKCS#12 deterministically on current macOS. Passing explicit input format/type flags (while preserving the intended trusted-tool access and partition-list setup) should avoid relying on OS format inference.

### Full `tauri info` output

```text
[✔] Environment
    - OS: Mac OS 26.3.0 arm64 (X64)
    ✔ Xcode: 26.6
    ✔ rustc: 1.88.0
    ✔ cargo: 1.88.0
    - node: 22.17.1
    - npm: 9.4.0

[-] Packages
    - tauri: 2.11.5
    - tauri-build: 2.6.3
    - wry: 0.55.1
    - tao: 0.35.3
    - tauri-cli (npm): 2.11.4
    - @tauri-apps/api: 2.11.1
    - @tauri-apps/cli: 2.11.4
```

### Stack trace

```text
failed to import keychain certificate
SecKeychainItemImport: MAC verification failed during PKCS12 import (wrong password?)
```

### Additional context

No certificate or password is included here. We worked around the failure by keeping the already-verified ephemeral keychain in the user search list and passing only `IOS_MOBILE_PROVISION` to Tauri, avoiding a duplicate P12 import.

lazerg:
`security import` was left to guess the file format. On macOS 26 that guess goes wrong for p12 files using modern PBES2/AES encryption and the import dies with `MAC verification failed during PKCS12 import (wrong password?)`, even when the password is correct.

I reproduced it on macOS 26.6.1 with a p12 encrypted using AES-256-CBC with a SHA-256 MAC. `Keychain::with_certificate_file` fails on it, and the same file with the same password imports fine as soon as `-f pkcs12` is passed. Older p12s using the legacy RC2/3DES encoding import either way, so nothing changes for those.

One thing worth flagging: naming the format means the import no longer accepts a bare PEM or DER certificate, which inference used to let through. Every caller in the tree passes a p12 and `ENVIRONMENT_VARIABLES.md` documents these variables as p12, so I left it strict rather than adding a fallback, but say the word if you would rather it retried without the flag.

Fixes #15843

### DIFF
--- a/.changes/keychain-import-pkcs12-format.md
+++ b/.changes/keychain-import-pkcs12-format.md
@@ -1,0 +1,5 @@
+---
+"tauri-macos-sign": patch:bug
+---
+
+Fixed importing a signing certificate failing with `MAC verification failed during PKCS12 import` for p12 files that use modern PBES2/AES encryption.
--- a/crates/tauri-macos-sign/src/keychain.rs
+++ b/crates/tauri-macos-sign/src/keychain.rs
@@ -106,6 +106,7 @@ impl Keychain {
       Command::new("security")
         .arg("import")
         .arg(cert_path)
+        .args(["-f", "pkcs12"])
         .arg("-P")
         .arg(certificate_password)
         .args([


lazerg:
Hey @Legend-Master, could you take a look at importing the signing certificate as an explicit pkcs12 for macOS signing? No feedback yet.

PR #15853, opened 2026-08-09
Fix doubled word in RELEASING.md

latent-9:
Small doubled-word fix: "published using using [publish-cli-js.yml]" -> "published using [publish-cli-js.yml]" (the parallel line just above, for `publish-cli-rs.yml`, has the single "using").


Legend-Master (approved):
thanks

PR #15856, opened 2026-08-09
WIP: special case state management of required plugins

sftse:
I'm putting this up for discussion how to do some small QoL improvements to the current state management.

There seem to be a number of things leaking out through things I'm not sure should have ever been public, like `StateManager` so the refactor is less clean than anticipated.

If we can inline the state management of the required plugins into `AppManager` I think this would lean stronger into the strengths of Rust's type system: Things that should be initialized will be statically obvious that they are.

### DIFF
--- a/crates/tauri/src/ipc/channel.rs
+++ b/crates/tauri/src/ipc/channel.rs
@@ -17,7 +17,7 @@ use crate::{
   command,
   ipc::{CommandArg, CommandItem},
   plugin::{Builder as PluginBuilder, TauriPlugin},
-  Manager, Runtime, State, Webview,
+  Runtime, State, Webview,
 };
 
 use super::{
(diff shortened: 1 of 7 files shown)


crates/tauri/src/ipc/channel.rs
```diff
@@ -42,8 +42,14 @@ static CHANNEL_COUNTER: AtomicU32 = AtomicU32::new(0);
 static CHANNEL_DATA_COUNTER: AtomicU32 = AtomicU32::new(0);
 
 /// Maps a channel id to a pending data that must be send to the JavaScript side via the IPC.
-#[derive(Default, Clone)]
-pub struct ChannelDataIpcQueue(Arc<Mutex<HashMap<u32, InvokeResponseBody>>>);
+#[derive(Default)]
+pub struct ChannelDataIpcQueue(HashMap<u32, InvokeResponseBody>);
```
sftse, line 46: I've sprinkled here a small change that is orthogonal to the meat and potatoes.

I think it's a bad idea to have types with special semantics like `Mutex` and `Arc` deeply buried in the types, if possible we should try and lift them as high up the call stack as we can, this makes it more obvious when some of these types are not actually doing anything, here the `Arc` evaporates.

Legend-Master:
I don't mind we put the internal states inside the app manager, but also don't really like the idea of special casing the core plugins to be honest

sftse:
I know what you mean, not feeling great about it either.

PR #15860, opened 2026-08-11
fix: menu related commands can panic

### LINKED ISSUE #15686: [bug] menu plugin can panic

Legend-Master:
Fix #15686
Closes #15691

The `do_menu_item` relies on tauri menu plugin internals so that it doesn't work on user land unless you try really hard, so I marked it as hidden properly.

### DIFF
--- a/.changes/menu-command-panic-on-wrong-input.md
+++ b/.changes/menu-command-panic-on-wrong-input.md
@@ -3,4 +3,7 @@ tauri: minor:bug
 tauri-macros: minor:bug
 ---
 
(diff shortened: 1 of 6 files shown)


sftse:
Is there any value in having the test from #15691 ?

sftse (approved):

Legend-Master:
> Is there any value in having the test from #15691 ?

it was an explicit panic before so don't think so

PR #15869, opened 2026-08-12
docs(api): clarify convertFileSrc asset protocol setup (fix #10755)

### LINKED ISSUE #10755: [docs] What is everything that's required to load an image in V2?
correct me if mistaken but,
the preferred and fastest way to load an image from disk for displaying is to use [convertFileSrc()](https://v2.tauri.app/reference/javascript/api/namespacecore/#convertfilesrc)
but the documentation is outdated, referencing v1 configs that don't exist anymore:
image
https://github.com/tauri-apps/tauri/blob/dev/tooling/api/src/core.ts#L163

it would also be good to have an explicit example of what's needed for it to work.
like:
```json
{
	"app": {
		"security": {
			"csp": {
				"default-src": "'self' asset: http://asset.localhost",
				"media-src": "'self' asset: http://asset.localhost",
				"img-src": "'self' asset: http://asset.localhost"
			},
			"assetProtocol": {
				"enable": true,
				"scope": [
					"$DOWNLOAD/**"
				]
			}
		}
	}
}
```

Correxxt:
Fixes #10755.

This updates the `convertFileSrc` API documentation with a complete Tauri v2 configuration example. It shows the asset protocol scope and CSP origins together, then uses the same Downloads-directory scope in the TypeScript example.

Validation:

- `pnpm --filter @tauri-apps/api eslint:check`
- `pnpm --filter @tauri-apps/api ts:check`
- `pnpm --filter @tauri-apps/api build`

### DIFF
--- a/packages/api/src/core.ts
+++ b/packages/api/src/core.ts
@@ -257,29 +257,43 @@ async function invoke<T>(
 }
 
 /**
- * Convert a device file path to an URL that can be loaded by the webview.
- * Note that `asset:` and `http://asset.localhost` must be added to [`app.security.csp`](https://v2.tauri.app/reference/config/#csp-1) in `tauri.conf.json`.
- * Example CSP value: `"csp": "default-src 'self' ipc: http://ipc.localhost; img-src 'self' asset: http://asset.localhost"` to use the asset protocol on image sources.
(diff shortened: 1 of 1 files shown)


Legend-Master (approved):
Thanks!

PR #15873, opened 2026-08-13
chore(deps): Update pnpm to 11.21.0

FabianLars:
11.13.0 was broken and yanked/deprecated. dunno which version we want so i went with latest

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,5 +28,5 @@
     "prettier": "^3.8.1"
   },
   "minimumReleaseAge": 4320,
-  "packageManager": "pnpm@11.13.0"
+  "packageManager": "pnpm@11.21.0"
 }


Legend-Master (approved):
I still kinda hate pinning the package managers but here we are

FabianLars:
would the `engines` field work for us? that allows ranges

Legend-Master:
Not sure, anything about corepack is so confusing not gonna lie. `engines` seem to just be a check, while `packageManager` actually manages the package managers, there's also [`devEngines.packageManager`](https://pnpm.io/package_json#devenginespackagemanager) and [`engines.runtime`](https://pnpm.io/package_json#enginesruntime) (and its dev version)

I have tried some of them before, but I think nothing really worked except for `packageManager`

FabianLars:
fwiw pnpm stopped recommending corepack and i feel like in general nobody really depended on that to automatically install the package manager required by the package.json file.
Either way, i wouldn't make a decision based on corepack, imo it's reasonable to expect people to install pnpm themselves like cargo 🤷

Legend-Master:
Yeah fair enough, corepack was the preferred method for the past few years I believe, but I guess that's no longer the case now. 🫠

PR #15887, opened 2026-08-17
refactor: lock minor versions

Legend-Master:


Cargo.toml
```diff
@@ -47,12 +47,12 @@ rust-version = "1.90"
...
+tauri-runtime-wry = { version = "~2.11.4", path = "./crates/tauri-runtime-wry", default-features = false }
 tauri-build = { version = "2.6.3", path = "./crates/tauri-build", default-features = false }
-tauri-macros = { version = "2.6.3", path = "./crates/tauri-macros" }
-tauri-codegen = { version = "2.6.3", path = "./crates/tauri-codegen" }
-tauri-utils = { version = "2.9.3", path = "./crates/tauri-utils" }
+tauri-macros = { version = "~2.6.3", path = "./crates/tauri-macros" }
+tauri-codegen = { version = "~2.6.3", path = "./crates/tauri-codegen" }
+tauri-utils = { version = "~2.9.3", path = "./crates/tauri-utils" }
 tauri-plugin = { version = "2.6.3", path = "./crates/tauri-plugin", default-features = false }
 tauri-bundler = { version = "2.9.4", path = "./crates/tauri-bundler", default-features = false }
```
FabianLars: maybe bundler as well? just in case we change the error again.
in theory tauri-macos-sign is also only internally but idk if worth to lock but then it'd be consistent?

Legend-Master: Ah, I forget you can actually build old versions of the CLI... Nice catch. Done in 85116f3.

FabianLars (approved):

PR #15889, opened 2026-08-17
docs(api): Explicit Resource Management for `Resource`

Legend-Master:
Follow up to #14103

### DIFF
--- a/packages/api/src/core.ts
+++ b/packages/api/src/core.ts
@@ -314,7 +314,6 @@ function convertFileSrc(filePath: string, protocol = 'asset'): string {
  * To support older browsers with [Explicit Resource Management], use a supported compiler (e.g. tsc) or bundler (e.g. rollup).
  *
  * @example
- *
  * ```typescript
  * import { Resource, invoke } from '@tauri-apps/api/core';
  *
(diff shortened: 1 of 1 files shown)


packages/api/src/core.ts
```diff
@@ -309,11 +309,15 @@ function convertFileSrc(filePath: string, protocol = 'asset'): string {
  *
  * The resource lives in the main process and does not exist
  * in the Javascript world, and thus will not be cleaned up automatically
- * except on application exit. If you want to clean it up early, call {@linkcode Resource.close}
+ * except on application exit. If you want to clean it up early, call {@linkcode Resource.close} or use use [Explicit Resource Management].
```
FabianLars:
```suggestion
 * except on application exit. If you want to clean it up early, call {@linkcode Resource.close} or use [Explicit Resource Management].
```

packages/api/src/core.ts
```diff
@@ -309,11 +309,15 @@ function convertFileSrc(filePath: string, protocol = 'asset'): string {
...
+ *
+ * To support older browsers with [Explicit Resource Management], use a supported compiler (e.g. tsc) or bundler (e.g. rollup)
```
FabianLars:
```suggestion
 * To support older browsers with [Explicit Resource Management], use a supported compiler (e.g. tsc) or bundler (e.g. rollup).
```

packages/api/src/core.ts
```diff
@@ -324,7 +328,32 @@ function convertFileSrc(filePath: string, protocol = 'asset'): string {
  *     await invoke('execute_sql', { rid: this.rid, sql });
  *   }
  * }
+ *
+ * {
+ *   await using db = await DatabaseHandle.open('test.db');
+ *   await db.execute('SELECT *');
+ * }
```
FabianLars: should this be a separate example with a header/short explanation maybe?

Legend-Master: good idea, done in e124ae4

FabianLars (approved):

PR #15895, opened 2026-08-19
feat(cli): warn when productName is left as the default (fix #14968)

### LINKED ISSUE #14968: [bug] Generated `upgrade_code` for default `product_name` (tauri-app) causes ambiguity with an unrelated `winget` application
### Describe the bug

When bundling an application using the default name (tauri-app) and installing it using the generated .msi, `winget list` or `winget list --upgrade-available` (_if `product_version` < 1.0.3_) will reference an unrelated application: [maqibin.MDXNotes](https://github.com/microsoft/winget-pkgs/blob/master/manifests/m/maqibin/MDXNotes/1.0.3/maqibin.MDXNotes.installer.yaml).

This could ultimately lead to unwanted installations using `winget upgrade --all`.

Related issue on winget-cli repo: https://github.com/microsoft/winget-cli/issues/6040 by @tkappedev

Related code:
https://github.com/tauri-apps/tauri/blob/88c05689c884fb456431d67d20ed12124d9cb901/crates/tauri-bundler/src/bundle/windows/msi/mod.rs#L582-L592

If above code is modified, the cli would need adjustments too:

https://github.com/tauri-apps/tauri/blob/88c05689c884fb456431d67d20ed12124d9cb901/crates/tauri-cli/src/inspect.rs#L38-L43

### Reproduction

  https://github.com/microsoft/winget-cli/issues/6040
> 1. Choose any preferred variant of create-tauri-app: https://v2.tauri.app/start/create-project/
> 2. When prompted on name, etc., select the default settings, keep name as 'tauri-app' and choose any stack
> 3. Build the application: https://v2.tauri.app/reference/cli/#build
> 4. Install the application with the created .msi package (tauri-app\src-tauri\target\release\bundle\msi\tauri-app_0.1.0_x64_en-US.msi) and use default install location (C:\Program Files\tauri-app)
> 5. Do a `winget list` or `winget list --upgrade-available`.
> 6. The installed app will be detected as `maqibin.MDXNotes`, and with winget as source

### Expected behavior

Using an additional and suitable property of the [package](https://doc.rust-lang.org/cargo/reference/manifest.html#the-package-section) to generate a truly unique `upgrade_code` per application.

### Full `tauri info` output

```text
-
```

### Stack trace

```text

```

### Additional context

_No response_

leepokai:
Adds a warning to `tauri build` when `productName` is still the default `tauri-app`. Apps published with the default name all derive the same WiX upgrade code, which is how the winget collision in the issue happens. The check sits next to the existing identifier checks in `setup()`, so it also covers `tauri android build` and `tauri ios build`, and stays quiet on `tauri dev`. The `productName` and `identifier` config docs now say what has to change before publishing, and the generated schemas were rebuilt. The upgrade code derivation itself is untouched since changing it would break updates for already shipped apps.

Output with the default name:

```
Warn The `productName` is still set to the default value `tauri-app`. You should change it before publishing your app since it is used to derive the default Windows installer (WiX) upgrade code, which must be unique across applications.
```

Closes #14968

### DIFF
--- a/.changes/default-product-name-warning.md
+++ b/.changes/default-product-name-warning.md
@@ -1,6 +1,7 @@
 ---
 "tauri-cli": patch:enhance
+"@tauri-apps/cli": patch:enhance
 "tauri-utils": patch:enhance
 ---
 
-`tauri build` now warns when `productName` is still set to the default `tauri-app`, since it is used to derive the default Windows installer (WiX) upgrade code, which must be unique across applications. The config documentation for `productName` and `identifier` was updated accordingly.
+`tauri build` now warns when `productName` is still set to the default `tauri-app`, since it names the generated bundles and is written into install paths and metadata that are expected to be unique to your application. The config documentation for `productName` now lists what the field controls on each platform, and `identifier`'s documentation notes that the default value is rejected.
--- a/.changes/default-product-name-warning.md
+++ b/.changes/default-product-name-warning.md
@@ -1,0 +1,6 @@
+---
+"tauri-cli": patch:enhance
+"tauri-utils": patch:enhance
+---
+
+`tauri build` now warns when `productName` is still set to the default `tauri-app`, since it is used to derive the default Windows installer (WiX) upgrade code, which must be unique across applications. The config documentation for `productName` and `identifier` was updated accordingly.
--- a/crates/tauri-cli/src/build.rs
+++ b/crates/tauri-cli/src/build.rs
@@ -193,6 +193,12 @@ pub fn setup(
     );
   }
 
+  if config.product_name.as_deref() == Some("tauri-app") {
+    log::warn!(
+      "The `productName` is still set to the default value `tauri-app`. You should change it before publishing your app since it is used to derive the default Windows installer (WiX) upgrade code, which must be unique across applications."
+    );
+  }
+
   if let Some(before_build) = config.build.before_build_command.clone() {
     helpers::run_hook(
       "beforeBuildCommand",
--- a/crates/tauri-cli/src/build.rs
+++ b/crates/tauri-cli/src/build.rs
@@ -195,7 +195,7 @@ pub fn setup(
 
   if config.product_name.as_deref() == Some("tauri-app") {
     log::warn!(
-      "The `productName` is still set to the default value `tauri-app`. You should change it before publishing your app since it is used to derive the default Windows installer (WiX) upgrade code, which must be unique across applications."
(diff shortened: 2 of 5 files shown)


crates/tauri-utils/src/config.rs
```diff
@@ -3744,6 +3744,11 @@ pub struct Config {
   #[serde(rename = "$schema")]
   pub schema: Option<String>,
   /// App name.
+  ///
+  /// Make sure to change this from the default `tauri-app` before publishing your app.
+  /// On Windows it is used to derive the default WiX installer upgrade code,
+  /// which must be unique across applications and can be set explicitly with the
+  /// [`bundle > windows > wix > upgradeCode`](WixConfig::upgrade_code) setting.
```
Legend-Master: This is a bit too specific on the wix upgrade code, this field affects a lot more than just a upgrade code that we should probably document that instead. (I know this was very poorly documented)

> https://github.com/tauri-apps/tauri/issues/14968#issuecomment-3933824733

crates/tauri-cli/src/build.rs
```diff
@@ -193,6 +193,12 @@ pub fn setup(
     );
   }
 
+  if config.product_name.as_deref() == Some("tauri-app") {
+    log::warn!(
+      "The `productName` is still set to the default value `tauri-app`. You should change it before publishing your app since it is used to derive the default Windows installer (WiX) upgrade code, which must be unique across applications."
```
Legend-Master: We should also update the description of this

.changes/default-product-name-warning.md
```diff
@@ -0,0 +1,6 @@
+---
+"tauri-cli": patch:enhance
```
Legend-Master, line 2:
```suggestion
"tauri-cli": patch:enhance
"@tauri-apps/cli": patch:enhance
```

Legend-Master (dismissed):
Thanks! Just some nitpicks and we can get it merged. cc @FabianLars if you could give the macOS docs a quick check

FabianLars:
macos sounds correct

crates/tauri-cli/src/build.rs
```diff
@@ -193,6 +193,12 @@ pub fn setup(
     );
   }
 
+  if config.product_name.as_deref() == Some("tauri-app") {
+    log::warn!(
+      "The `productName` is still set to the default value `tauri-app`. You should change it before publishing your app since it names the generated bundles and is written into install paths and platform metadata that are expected to be unique to your application, like the Windows installer upgrade code."
```
Legend-Master:
```suggestion
      "The `productName` is still set to the default value `tauri-app`, it must be unique across applications since it is written into install paths and platform metadata that are expected to be unique to your application, like the Windows installer upgrade code."
```

Legend-Master (approved):
Thank you both

PR #15897, opened 2026-08-20
fix(tauri-runtime-wry): handle event loop queue overflow with backpressure and coalescing

### LINKED ISSUE #8177: [bug] Event emit crashes app with high call rate

Br1skyy:
Fixes #8177

## Problem

Windows `PostMessageW` has a per-thread message queue limit. When `window.emit()` is called from many worker threads at high rate, the queue can overflow and `PostMessageW` fail with error 1816 (`ERROR_NOT_ENOUGH_QUOTA`). Right now these messages get silently dropped.

I test by calling `PostMessageW` directly from 4+ threads with raw Win32 API and the OS drop 89-97% of messages. The queue limit is real.

In Tauri v2, tao 0.36.0 use crossbeam channel for event payload and `PostMessageW` only as wakeup signal. This make it much harder to hit on modern hardware. But the issue was reported on Tauri v1 / Windows 10 where the old tao packed payload into `PostMessageW` directly. On Windows 10 or under heavy system load the queue can still overflow.

## What this PR do

1. **Backpressure queue** (`PendingQueue`): When `send_event` fail, fire-and-forget messages get queued and retried next `MainEventsCleared`. Getter messages with `Sender` not queued (would deadlock). Queue capped at 10000, oldest removed when full.

2. **`EvaluateScript` coalescing**: Multiple `EvaluateScript` for same webview get merged, only latest kept. Less queue pressure from fast emit calls.

3. **Logging**: All silent `let _ = proxy.send_event(...)` replaced with `log::warn!` so user can see when messages fail.

## Test

- 8 unit test for `PendingQueue` (all pass): push/drain, capacity limit, coalescing, order preservation
- Repro app with 32 threads x 1M events, no crash, no regression
- Raw `PostMessageW` flood test from 4/8/16 threads confirm the OS quota limit exist (89-97% message loss)
- On v2 + Windows 11, `emit()` does not fail even under extreme pressure because tao channel handle it. Defensive fix for older Windows or when system is under heavy load.

## Note

The crash in #8177 was Tauri v1. In v2 the tao channel design make it much harder to reproduce. This PR add defensive backpressure handling, event coalescing, and diagnostic logging for edge cases where `PostMessageW` quota is exceeded.

### DIFF
--- a/.changes/fix-event-emit-crash.md
+++ b/.changes/fix-event-emit-crash.md
@@ -1,0 +1,5 @@
+---
+"tauri-runtime-wry": patch:bug
+---
+
+Added a bounded backpressure queue for messages that fail to be delivered to the OS event loop. When `send_event` fails (e.g., Windows `PostMessage` quota exceeded with error 1816), messages are queued and retried on the next `MainEventsCleared` iteration instead of being silently dropped. Includes event coalescing: high-frequency `EvaluateScript` messages targeting the same webview are deduplicated, keeping only the latest. Also replaced all remaining `let _ = proxy.send_event(...)` silent drops with `log::warn!` diagnostics. This prevents app crashes when `emit()` is called at high rates from worker threads, fixing #8177.
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -126,7 +126,7 @@ use std::{
   cell::RefCell,
   collections::{
     hash_map::Entry::{Occupied, Vacant},
-    BTreeMap, HashMap, HashSet,
+    BTreeMap, HashMap, HashSet, VecDeque,
   },
   fmt,
   ops::Deref,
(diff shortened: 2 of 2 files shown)


Legend-Master:
Thanks for contributing, but please provide a repro as mentioned in https://github.com/tauri-apps/tauri/issues/8177#issuecomment-2101041340

Br1skyy:
> Thanks for contributing, but please provide a repro as mentioned in [#8177 (comment)](https://github.com/tauri-apps/tauri/issues/8177#issuecomment-2101041340)

Sorry, working on it.

PR #15920, opened 2026-08-26
fix(path): return "." from path.normalize("")

xia-chao:
`path.normalize("")` returned an empty string.

The comment in `crates/tauri/src/path/plugin.rs` already states Node.js returns `"."` for `normalize("")` and `normalize(".")`. The same module's `join` already maps empty segments to `"."` (covered by `path::plugin::tests::join`).

This change uses that existing contract:

- `normalize("")` -> `"."`
- existing `normalize(".")` / `normalize("..")` behavior unchanged

Verify:

```
cargo test -p tauri --lib path::plugin::tests
```

### DIFF
--- a/crates/tauri/src/path/plugin.rs
+++ b/crates/tauri/src/path/plugin.rs
@@ -120,7 +120,7 @@ pub fn normalize(path: String) -> String {
   // and `"."` for `normalize("")` or `normalize(".")`
   if p.is_empty() && path == ".." {
     "..".into()
-  } else if p.is_empty() && path == "." {
+  } else if p.is_empty() && (path.is_empty() || path == ".") {
     ".".into()
   } else {
     // Add a trailing separator if the path passed to this functions had a trailing separator. That's how Node.js behaves.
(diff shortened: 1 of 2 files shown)


javascript-unsafe:
request review

crates/tauri/src/path/plugin.rs
```diff
@@ -310,4 +310,21 @@ mod tests {
...
+      } else {
+        expected_unix
+      };
+      assert_eq!(super::normalize(path.into()), expected);
+    }
+
+    // Same Node contract as `join(vec![""])` / the comment on `normalize`.
+    check("", ".", ".");
+    check(".", ".", ".");
+    check("..", "..", "..");
```
amrbashir: This can be simplified to 

```suggestion
    assert_eq!(super::normalize(""), ".");
    assert_eq!(super::normalize("."), ".");
    assert_eq!(super::normalize(".."), "..");
```

javascript-unsafe:
Applied, with .into() because normalize takes String.
recheck

Legend-Master (approved):
Thanks

PR #15923, opened 2026-08-27
fix(tauri-utils): strip a leading UTF-8 BOM from the config file (fix #15922)

### LINKED ISSUE #15922: [bug] `tauri.conf.json` fails to parse if the file starts with a UTF-8 BOM
### Describe the bug

If `tauri.conf.json` is saved with a UTF-8 byte order mark (`EF BB BF`), Tauri fails to
parse it and reports a generic JSON syntax error pointing at line 1, column 1. The file is
valid JSON and looks completely correct in an editor, so the actual cause (an invisible
leading byte) is not discoverable from the error message.

This is easy to hit on Windows, because PowerShell — the default shell — writes a BOM for
several common encoding options. Any script or tooling that regenerates or rewrites the
config can silently introduce it.

### Reproduction

1. Create a Tauri v2 app.
2. Rewrite the config with PowerShell, which prepends a BOM:
   ```powershell
   $c = Get-Content tauri.conf.json -Raw
   Set-Content tauri.conf.json -Value $c -Encoding UTF8
   ```
3. Verify the BOM is present:
   ```
   $ xxd tauri.conf.json | head -1
   00000000: efbb bf7b 0a20 2022 2473 6368 656d 6122  ...{.  "$schema"
   ```
4. Run any command that loads the config (e.g. `cargo tauri build`).

### Expected behavior

The config is parsed normally. A UTF-8 BOM is a legal way to encode a UTF-8 file, and most
JSON consumers either skip it or say explicitly that it is the problem.

### Actual behavior

Parsing fails with a `serde_json` syntax error at line 1 column 1. Nothing in the message
mentions a BOM or an encoding issue, and the file renders as valid JSON in every editor,
so the failure looks like config corruption rather than an encoding problem.

### Root cause

In `tauri-utils` (2.9.3), `src/config/parse.rs`:

- `:282` — `let raw = read_to_string(&path)?;` — this wraps `std::fs::read_to_string`
  (`:393`), which validates UTF-8 but does **not** strip a BOM: `U+FEFF` is valid UTF-8 and
  is kept in the returned `String`.
- `:351-355` — `do_parse_json` passes `raw` straight to `serde_json::from_str` with no
  normalization:
  ```rust
  fn do_parse_json<D: DeserializeOwned>(raw: &str, path: &Path) -> Result<D, ConfigError> {
    serde_json::from_str(raw).map_err(|error| ConfigError::FormatJson {
      path: path.into(),
      error,
    })
  }
  ```
- `serde_json` rejects the leading `U+FEFF` and reports `expected value at line 1 column 1`.

Note that when the `config-json5` feature is enabled, `:288-300` tries JSON5 first and, on
failure, falls back to reporting the `serde_json` error — so the surfaced message is the
same either way. The comment there already anticipates that this can be confusing.

### Suggested fix

Strip a leading BOM once, right after the file is read, so every format path benefits:

```rust
let raw = read_to_string(&path)?;
let raw = raw.strip_prefix('\u{feff}').unwrap_or(&raw).to_string();
```

Alternatively, do it inside the `read_to_string` helper at `:393`, which also covers the
`.json5` (`:306`) and `.toml` (`:318`) branches.

If stripping is not desirable, detecting the BOM and returning a dedicated error that names
it would at least make the failure diagnosable.

### Platform and versions

- OS: Windows 11 (also reproducible anywhere a BOM-prefixed config is produced)
- Tauri: v2 (`tauri = "2"`, `tauri-build = "2"`)
- `tauri-utils`: 2.9.3
- Config schema: `https://schema.tauri.app/config/2`

### Prior art

`serde_json` deliberately does not skip BOMs, so consumers are expected to handle it. Other
tools in this space normalize it at the read boundary for exactly this reason. #15601 notes
a related BOM pitfall elsewhere in the Tauri toolchain (`latest.json` must be written
BOM-free), which suggests handling this once at the config read boundary would remove a
recurring class of confusing failures.

I'm happy to open a PR with the fix above if you'd like.

zegroged:
Fixes #15922.

### What

`tauri.conf.json` fails to parse when the file starts with a UTF-8 BOM. The error surfaced is
a generic `expected value at line 1 column 1`, on a file that looks perfectly valid in an
editor — the leading byte is invisible, so the actual cause is not discoverable from the
message.

This is easy to hit on Windows: several common ways of writing the file add a BOM, including
PowerShell's `Set-Content -Encoding UTF8`.

### Why it happens

`read_to_string` (`crates/tauri-utils/src/config/parse.rs`) wraps `std::fs::read_to_string`,
which validates UTF-8 but does **not** strip a BOM — `U+FEFF` is valid UTF-8 and stays in the
returned `String`. The contents are then handed straight to `serde_json::from_str`, which
rejects it.

### The change

Strip a single leading `U+FEFF` in the shared `read_to_string` helper. That helper is the one
read boundary for all three config formats — JSON (`:282`), JSON5 (`:306`) and TOML (`:318`) —
so one place covers every branch, and no parser call site needs to change.

### Tests

Two tests are added next to the helper:

- `read_to_string_strips_a_leading_utf8_bom` — a BOM-prefixed config is read without the BOM
  and parses.
- `read_to_string_leaves_a_file_without_a_bom_untouched` — contents without a BOM are returned
  byte-for-byte, so the change is a no-op on existing files.

Both pass, and I checked they actually fail without the fix rather than passing vacuously:
removing the two stripping lines makes `read_to_string_strips_a_leading_utf8_bom` fail with
`the BOM must be stripped, otherwise every parser below rejects the file`, while the non-BOM
test stays green.

```
$ cargo test -p tauri-utils --lib config::parse
running 2 tests
test config::parse::tests::read_to_string_leaves_a_file_without_a_bom_untouched ... ok
test config::parse::tests::read_to_string_strips_a_leading_utf8_bom ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 34 filtered out
```

### Note on tooling

This was written with AI assistance. Per the AI Tool Policy I'm responsible for what's here,
so rather than just asserting it works: the change is two lines, the tests are in the diff, and
the mutation check described above is the evidence that they fail without the fix instead of
passing vacuously. Happy to adjust anything — placement, wording, or test shape.

### DIFF
--- a/.changes/config-strip-utf8-bom.md
+++ b/.changes/config-strip-utf8-bom.md
@@ -2,4 +2,4 @@
 "tauri-utils": "patch:bug"
 ---
 
-Strip a leading UTF-8 BOM when reading the Tauri config file. `std::fs::read_to_string` keeps `U+FEFF`, and none of the JSON, JSON5 or TOML parsers accept it, so a BOM-prefixed config previously failed with an opaque `expected value at line 1 column 1` on a file that looks valid in an editor. This is easy to hit on Windows, where several common ways of writing the file (for instance PowerShell's `Set-Content -Encoding UTF8`) add a BOM.
+Fixed an issue that caused reading the Tauri config file to fail when it starts with a UTF-8 BOM sequence.
--- a/.changes/config-strip-utf8-bom.md
+++ b/.changes/config-strip-utf8-bom.md
@@ -1,0 +1,5 @@
+---
+"tauri-utils": "patch:bug"
+---
+
+Strip a leading UTF-8 BOM when reading the Tauri config file. `std::fs::read_to_string` keeps `U+FEFF`, and none of the JSON, JSON5 or TOML parsers accept it, so a BOM-prefixed config previously failed with an opaque `expected value at line 1 column 1` on a file that looks valid in an editor. This is easy to hit on Windows, where several common ways of writing the file (for instance PowerShell's `Set-Content -Encoding UTF8`) add a BOM.
--- a/crates/tauri-utils/src/config/parse.rs
+++ b/crates/tauri-utils/src/config/parse.rs
@@ -391,12 +391,7 @@ fn do_parse_toml<D: DeserializeOwned>(raw: &str, path: &Path) -> Result<D, Confi
 
 /// Helper function to wrap IO errors from [`std::fs::read_to_string`] into a [`ConfigError`].
 ///
-/// A leading UTF-8 BOM is stripped before the contents are handed to a parser.
-/// [`std::fs::read_to_string`] validates UTF-8 but keeps `U+FEFF` in the returned string, and
-/// none of the parsers used above accept it. Windows tooling writes BOM-prefixed files in
-/// several common cases (for instance PowerShell's `Set-Content -Encoding UTF8`), which would
-/// otherwise fail with an opaque `expected value at line 1 column 1` on a file that looks
-/// perfectly valid in an editor.
+/// A leading UTF-8 BOM is stripped before the contents are handed to the parser.
 fn read_to_string(path: &Path) -> Result<String, ConfigError> {
   let mut content = std::fs::read_to_string(path).map_err(|error| ConfigError::Io {
     path: path.into(),
@@ -414,40 +409,14 @@ fn read_to_string(path: &Path) -> Result<String, ConfigError> {
 mod tests {
   use super::*;
 
-  fn write_temp(name: &str, contents: &str) -> std::path::PathBuf {
-    let dir = std::env::temp_dir().join("tauri-utils-parse-tests");
-    std::fs::create_dir_all(&dir).unwrap();
-    let path = dir.join(name);
-    std::fs::write(&path, contents).unwrap();
-    path
-  }
-
   #[test]
   fn read_to_string_strips_a_leading_utf8_bom() {
-    let path = write_temp("with-bom.json", "\u{feff}{\"productName\":\"test\"}");
+    let path = std::env::temp_dir().join("tauri-utils-bom.json");
+    std::fs::write(&path, "\u{feff}{\"productName\":\"test\"}").unwrap();
 
     let raw = read_to_string(&path).unwrap();
 
-    assert!(
-      !raw.starts_with('\u{feff}'),
-      "the BOM must be stripped, otherwise every parser below rejects the file"
-    );
-    assert!(
-      do_parse_json::<serde_json::Value>(&raw, &path).is_ok(),
-      "a BOM-prefixed config must parse once the BOM is stripped"
-    );
-
-    std::fs::remove_file(&path).unwrap();
-  }
-
-  #[test]
-  fn read_to_string_leaves_a_file_without_a_bom_untouched() {
-    let contents = "{\"productName\":\"test\"}";
-    let path = write_temp("without-bom.json", contents);
-
-    let raw = read_to_string(&path).unwrap();
-
-    assert_eq!(raw, contents);
+    assert!(do_parse_json::<serde_json::Value>(&raw, &path).is_ok());
 
     std::fs::remove_file(&path).unwrap();
   }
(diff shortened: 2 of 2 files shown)


crates/tauri-utils/src/config/parse.rs
```diff
@@ -390,9 +390,65 @@ fn do_parse_toml<D: DeserializeOwned>(raw: &str, path: &Path) -> Result<D, Confi
 }
 
 /// Helper function to wrap IO errors from [`std::fs::read_to_string`] into a [`ConfigError`].
+///
+/// A leading UTF-8 BOM is stripped before the contents are handed to a parser.
+/// [`std::fs::read_to_string`] validates UTF-8 but keeps `U+FEFF` in the returned string, and
+/// none of the parsers used above accept it. Windows tooling writes BOM-prefixed files in
+/// several common cases (for instance PowerShell's `Set-Content -Encoding UTF8`), which would
+/// otherwise fail with an opaque `expected value at line 1 column 1` on a file that looks
+/// perfectly valid in an editor.
```
FabianLars:
```suggestion
/// A leading UTF-8 BOM is stripped before the contents are handed to the parser.
```

It's not that deep tbh

crates/tauri-utils/src/config/parse.rs
```diff
@@ -390,9 +390,65 @@ fn do_parse_toml<D: DeserializeOwned>(raw: &str, path: &Path) -> Result<D, Confi
...
+  fn read_to_string_leaves_a_file_without_a_bom_untouched() {
+    let contents = "{\"productName\":\"test\"}";
+    let path = write_temp("without-bom.json", contents);
+
+    let raw = read_to_string(&path).unwrap();
+
+    assert_eq!(raw, contents);
+
+    std::fs::remove_file(&path).unwrap();
+  }
```
FabianLars, line 422: ngl i don't think a single strip_prefix needs this big of a block of tests (imo none are needed, but i won't block the PR on this)

.changes/config-strip-utf8-bom.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri-utils": "patch:bug"
+---
+
+Strip a leading UTF-8 BOM when reading the Tauri config file. `std::fs::read_to_string` keeps `U+FEFF`, and none of the JSON, JSON5 or TOML parsers accept it, so a BOM-prefixed config previously failed with an opaque `expected value at line 1 column 1` on a file that looks valid in an editor. This is easy to hit on Windows, where several common ways of writing the file (for instance PowerShell's `Set-Content -Encoding UTF8`) add a BOM.
```
FabianLars: We usually keep our changefiles short and sweet

```suggestion
Fixed an issue that caused reading of the Tauri config file to fail when it started with a UTF-8 BOM sequence.
```

zegroged:
Took all your suggestions on board and pushed the changes. Thanks!

FabianLars (approved):

Legend-Master:
I am honestly incline to not support this at all as non of our other `serde_json::from_str` calls are guarded anyways, also referencing https://github.com/serde-rs/json/issues/1115

FabianLars:
orrr we fix the other invocations as well (though this is arguably the most common one that could face this issue). i only get a complaint every 1-2 years so i don't mind not getting this in 🤷

Legend-Master:
forcing everyone off utf-8 bom is a nice thing either ways imo, also special casing this one feels inconsistent to me (have no means to fix all other calls at least to me 😂)

zegroged:
Thanks both. One clarification, since I think the "special casing" read may be off:
the strip isn't next to a `serde_json::from_str` call — it's in the shared
`read_to_string` helper in `config/parse.rs`, the single entry point for JSON, JSON5
and TOML. One line, all three config formats, no individual call site touched.

On preferring people off BOM — fair, but Tauri doesn't reject it today, it just fails
with `expected value at line 1 column 1`, pointing at a character no editor shows.
Two reasonable outcomes, and the current behaviour is neither:

1. strip it (this PR), or
2. reject it with an error that says so.

Happy to switch to 2 — same line, error instead of strip.

@FabianLars on fixing the other invocations: glad to take that on in a separate PR,
so this one stays small.

Context for why it recurs: Windows PowerShell 5.1 still writes a BOM with
`Set-Content -Encoding UTF8`.

If you'd rather have neither, that's fine and I'll close it.

PR #15937, opened 2026-08-29
feat(runtime-wry): hint at WebKitGTK compositing env vars when software rendering is detected

### LINKED ISSUE #15936: Blank window under Xvfb until WEBKIT_DISABLE_COMPOSITING_MODE=1, with no diagnostic
Under Xvfb with software GL, the window opens and renders nothing. No error, no warning, no log line. Setting `WEBKIT_DISABLE_COMPOSITING_MODE=1` fixes it completely.

## Environment

- Linux (Debian 13, x86_64), WebKitGTK 2.52.3, Xvfb 1920x1080x24, fluxbox, no GPU.
- tauri 2.11.1.

## Before and after

Without the variable: a white window, correct title, correct size, DOM present in the inspector, nothing painted. With `WEBKIT_DISABLE_COMPOSITING_MODE=1` (also with `WEBKIT_DISABLE_DMABUF_RENDERER=1` and `LIBGL_ALWAYS_SOFTWARE=1`): renders correctly, first try.

It took me a while to figure this out, assuming my own asset pipeline was at fault, because a blank window and a 404'd bundle look identical.

## Suggestion

Tauri knows more about the environment than the app author does. When compositing is unavailable or the GL renderer is software, either set the variable or log one line saying the webview was created without compositing and naming the escape hatch.

## Related

#9304, #8655, #10011 and #12361 report the blank-window symptom itself (NVIDIA/DMABUF, Wayland); this issue is about the missing diagnostic, which none of them covers.

gronke:
WebKitGTK renders a blank window in environments without GPU acceleration (containers, VMs, CI) and logs nothing, so the workaround is undiscoverable.

On Linux and the BSDs, `Wry::init` now logs once:
- a warning naming `WEBKIT_DISABLE_DMABUF_RENDERER=1` / `WEBKIT_DISABLE_COMPOSITING_MODE=1` when `LIBGL_ALWAYS_SOFTWARE` or `GALLIUM_DRIVER` indicate software GL and no escape hatch is set,
- a debug-level breadcrumb otherwise.

Bare Xvfb without env indicators is not detectable without creating a GL context, which is too heavy for a diagnostic; the debug breadcrumb covers that case.
Log-only; no environment variables are modified.

Verified under `xvfb-run` with a log subscriber: warn fires with `GALLIUM_DRIVER=llvmpipe`, stays silent when the escape hatch is set.

Closes #15936

### DIFF
--- a/.changes/linux-webkit-rendering-hint.md
+++ b/.changes/linux-webkit-rendering-hint.md
@@ -1,0 +1,7 @@
+---
+"tauri-runtime-wry": patch:enhance
+---
+
+On Linux and the BSDs, log a one-time warning naming the `WEBKIT_DISABLE_DMABUF_RENDERER=1` / `WEBKIT_DISABLE_COMPOSITING_MODE=1` workarounds when software GL rendering is detected via the environment (`LIBGL_ALWAYS_SOFTWARE`, `GALLIUM_DRIVER`), plus a debug-level breadcrumb on every runtime initialization.
+WebKitGTK renders a blank window in such environments with no diagnostic of its own.
+Log-only; no environment variables are modified.
--- a/crates/tauri-runtime-wry/src/lib.rs
+++ b/crates/tauri-runtime-wry/src/lib.rs
@@ -2844,6 +2844,108 @@ impl<T: UserEvent> RuntimeHandle<T> for WryHandle<T> {
   }
 }
 
+/// WebKitGTK is known to render a blank window in environments without GPU
+/// acceleration (containers, VMs, CI) unless one of its rendering escape
+/// hatches is set. Since nothing is painted and nothing is logged in that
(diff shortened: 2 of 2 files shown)


FabianLars:
hmm i wonder if this is more something for Wry instead 🤔

gronke:
> hmm i wonder if this is more something for Wry instead 🤔

Good point, wry fits better. The blank window is engine-level knowledge that every wry consumer hits, not only Tauri apps, and wry init runs before the web processes spawn, which is when WebKit reads these variables. Happy to refile against wry and close this one. If you would rather land the hint here first for release timing, that works too, it is log-only and easy to drop later.

FabianLars:
Let's do it in wry then, we can do quick releases so no need to land it in tauri first. A new wry patch probably lands before the next tauri version.

PR #15939, opened 2026-08-29
fix(tauri): return 404 for missing subresource assets instead of SPA index.html fallback

### LINKED ISSUE #15938: Asset protocol: a missing file returns 200 text/html, so a failed module import names neither the URL nor the cause
A request for a path that is not in `frontendDist` returns **HTTP 200 with `Content-Type: text/html`** (the SPA fallback serving `index.html`). For a module request that turns "file not found" into `'text/html' is not a valid JavaScript MIME type`, and the offending URL appears nowhere in the error.

A missing module is a common way a desktop app fails to boot. Today that failure names neither the file nor the reason.

## Environment

tauri 2.11.1, tauri-build 2.6, Linux (Debian 13, x86_64), WebKitGTK 2.52.3, GTK 3.24.49.
`WebviewUrl::App("/index.html")` over a static `frontendDist` of ES modules with an inline import map.

## Scope

This applies whenever the `tauri://` protocol serves `frontendDist:` production builds, and dev without devUrl. With devUrl set, the webview talks to the dev server directly and this code path is not involved. The CLI's built-in dev server (static frontendDist in dev) implements the same fallback chain and fails identically.

## Evidence

Probed from inside the webview:

```
probe main.js                 -> 200 text/javascript  @ tauri://localhost/main.js
probe preload.js              -> 200 text/javascript  @ tauri://localhost/preload.js
probe definitely-not-here.js  -> 200 text/html        @ tauri://localhost/definitely-not-here.js
```

The last line is the problem. A path that exists nowhere returns `200 text/html`.

The symptom was a window rendering its splash screen forever, with one diagnostic:

```
[boot] import failed, retrying once cache-busted: 'text/html' is not a valid JavaScript MIME type.
```

Nothing further was reported. The retry (same specifier plus a `?cache-bust` query) neither resolved nor rejected while watching, so the boot promise never settled. I didn't establish whether the query string is handled differently.

I've ended up injecting an init script that fetches every import-map target and reports which ones come back as HTML.

## Why it matters

`200 text/html` is indistinguishable from success at the protocol level, and the resulting MIME error names no URL, so the failure is invisible to the tooling that would otherwise catch it. A 404 is machine readable and can be easily caught by CI.

## Suggestion

Do not fall back for requests that are clearly not navigations. `Sec-Fetch-Dest: script`, the `Accept` header, or a `.js`/`.mjs`/`.css`/`.json` extension is enough to answer 404 instead of serving `index.html`. SPA fallback is right for a navigation and harmful for a subresource.

Failing that, warn from the protocol handler when it falls back, naming the requested path.

No config or builder option disables the fallback today. The closest workaround is rewriting the response in on_web_resource_request when a subresource path comes back as `text/html`, which is the heuristic this issue argues the framework should apply itself.

gronke:
A request for a path missing from the bundled assets returned `200 text/html` (the SPA `index.html` fallback), so a failed ES module import surfaced only as `'text/html' is not a valid JavaScript MIME type`, naming neither the URL nor the cause.

- `AppManager::get_asset` skips the whole fallback chain when the requested path has a static subresource extension (conservative allowlist in `tauri_utils::mime_type::has_subresource_extension`; dotted SPA routes like `/product/v1.2` keep the fallback).
- The `tauri://` protocol answers `404` `text/plain` naming the path; other errors keep the 500 path. `on_web_resource_request` can still override.
- The CLI's built-in dev server mirrors the same rule and 404 body.
- The SPA fallback now logs a warning when it serves `index.html` for a missing path.
- Fixes a latent double-boxing in `get_asset`'s error path that made the boxed error non-downcastable to `tauri::Error`.

Header-based detection (`Sec-Fetch-Dest`) was deliberately not used; it is unreliable on WebKitGTK custom schemes. First unit tests for the `get_asset` chain (map-backed `Assets` impl) and the dev-server resolution helper.

Closes #15938

### DIFF
--- a/crates/tauri/src/manager/mod.rs
+++ b/crates/tauri/src/manager/mod.rs
@@ -387,6 +387,7 @@ impl<R: Runtime> AppManager<R> {
     &self,
     mut path: String,
     _use_https_schema: bool,
+    allow_html_fallback: bool,
   ) -> Result<Asset, Box<dyn std::error::Error>> {
     let assets = &self.assets;
     if path.ends_with('/') {
@@ -405,15 +406,10 @@ impl<R: Runtime> AppManager<R> {
 
     let mut asset_path = AssetKey::from(path.as_str());
 
-    // a missing subresource (script, style, image, ...) must never resolve to
-    // an HTML document; skipping the fallback chain lets the protocol handler
-    // answer 404 instead of `index.html` with a misleading mime type
-    let use_fallbacks = !tauri_utils::mime_type::has_subresource_extension(&path);
-
     let asset_response = assets
       .get(&asset_path)
       .or_else(|| {
-        if !use_fallbacks {
+        if !allow_html_fallback {
           return None;
         }
         log::debug!("Asset `{path}` not found; fallback to {path}.html");
@@ -423,7 +419,7 @@ impl<R: Runtime> AppManager<R> {
         asset
       })
       .or_else(|| {
-        if !use_fallbacks {
+        if !allow_html_fallback {
           return None;
         }
         log::debug!("Asset `{path}` not found; fallback to {path}/index.html",);
@@ -433,7 +429,7 @@ impl<R: Runtime> AppManager<R> {
         asset
       })
       .or_else(|| {
-        if !use_fallbacks {
+        if !allow_html_fallback {
           return None;
         }
         let fallback = AssetKey::from("index.html");
@@ -757,7 +753,7 @@ mod asset_tests {
   use crate::{
     sealed::ManagerBase,
     test::{mock_builder, mock_context, MockRuntime},
-    App,
+    App, Asset,
   };
   use std::borrow::Cow;
   use tauri_utils::assets::{AssetKey, AssetsIter, CspHash};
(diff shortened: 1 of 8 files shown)


crates/tauri/src/manager/mod.rs
```diff
@@ -405,32 +405,50 @@ impl<R: Runtime> AppManager<R> {
 
     let mut asset_path = AssetKey::from(path.as_str());
 
+    // a missing subresource (script, style, image, ...) must never resolve to
+    // an HTML document; skipping the fallback chain lets the protocol handler
+    // answer 404 instead of `index.html` with a misleading mime type
+    let use_fallbacks = !tauri_utils::mime_type::has_subresource_extension(&path);
```
Legend-Master: I'm not the most convinced with this, the whole point of this fallback is to always allow the client to react to the URL. Unless we can be sure that this is not from a navigation request, we should not do this.

gronke: Good catch, and you are right that the extension heuristic could not answer it; a navigation to `/reports/2024.pdf` would have 404'd.

Reworked to decide from the request destination rather than the URL.
Measured what actually reaches the `tauri://` handler on WebKitGTK 2.52.3: no `Sec-Fetch-Dest` for custom protocols, but `Accept` is `text/html,application/xhtml+xml,…` for navigations (including `/reports/2024.pdf`) and `*/*` or `image/…` for scripts, images and `fetch`.

The rule is now: `Sec-Fetch-Dest` when the webview sends it (`document`/`iframe`/`frame` → navigation), otherwise `Accept` containing `text/html`, and if neither header is present the request counts as a navigation so the fallback is preserved. Navigations therefore always reach the frontend router, and only a request that does not present itself as a document load can 404.

This also removes the extension allowlist entirely, and `AssetResolver` keeps the old behaviour since it has no request to classify.

PR #15941, opened 2026-08-29
fix: surface commands that are unreachable due to missing permissions

### LINKED ISSUE #15940: A command in invoke_handler but absent from COMMANDS is silently unreachable, with no build-time warning
A command listed in `invoke_handler!` but missing from `COMMANDS` in the plugin's build script is silently unreachable. No permission file is generated for it, so no capability can grant it, so every invoke is rejected at runtime. No warnings at build time.

## Background

My plugin had this:

```rust
// build.rs
const COMMANDS: &[&str] = &["do_something"];   // debug_log missing

// lib.rs
.invoke_handler(tauri::generate_handler![commands::do_something, commands::debug_log])
```

`debug_log` was documented in our own contributor docs as the way to forward frontend logs to the backend.
It had never worked, because of that one missing string. Every invoke was rejected, our JS caught the rejection, and the channel looked simply quiet.

Two steps were needed, and the second is easy to miss even once you know:

1. add the name to `COMMANDS` so the permission is generated
2. then grant `allow-debug-log` in `permissions/default.toml`

Generating the permission file does not grant it.

## Suggestion

`generate_handler!` knows the command names, the build script knows `COMMANDS`.
A build-time warning when a handler-registered command has no generated permission would have saved the whole detour.

An error would be defensible too, since the command is unreachable either way.

Secondary: a runtime log line when an invoke is rejected for a missing permission, rather than only an `Err` the caller may swallow.

gronke:
A command listed in `generate_handler!` but missing from the plugin's `COMMANDS` build-script list is silently unreachable: no permission is generated, so no capability can grant it, and every invoke is rejected at runtime with nothing on the host side.

Two changes:
- `generate_handler!` prints a build-time warning when a registered command is not referenced by any permission of the crate (plugin `COMMANDS` or hand-written permission files). Fail-open: crates without defined permissions stay silent, `REMOVE_UNUSED_COMMANDS` flow unchanged, renamed commands called out in the message. Probes the permission-file lists written by tauri-plugin (external plugins), tauri-build (inlined plugins, app manifest) and the tauri crate (core plugins).
- ACL denials are now logged host-side via `log::error!` (rich `resolve_access_message` in debug builds, short message in release); previously the reason lived only in the webview's rejected promise, which frontends routinely swallow.

Closes #15940

### DIFF
--- a/crates/tauri-macros/src/command/handler.rs
+++ b/crates/tauri-macros/src/command/handler.rs
@@ -38,7 +38,8 @@ impl Parse for Handler {
       .into_iter()
       .collect();
 
-    filter_unused_commands(plugin_name, &mut command_defs);
+    filter_unused_commands(&plugin_name, &mut command_defs);
+    warn_unreferenced_commands(&plugin_name, &command_defs);
     let mut commands = Vec::new();
     let mut wrappers = Vec::new();
 
@@ -89,7 +90,7 @@ fn try_get_plugin_name(input: &ParseBuffer<'_>) -> Result<Option<String>, syn::E
   )
 }
 
-fn filter_unused_commands(plugin_name: Option<String>, command_defs: &mut Vec<CommandDef>) {
+fn filter_unused_commands(plugin_name: &Option<String>, command_defs: &mut Vec<CommandDef>) {
   let allowed_commands = tauri_utils::acl::read_allowed_commands();
   let Some(allowed_commands) = allowed_commands else {
     return;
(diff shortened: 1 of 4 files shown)


crates/tauri-macros/src/command/handler.rs
```diff
@@ -141,6 +142,75 @@ fn filter_unused_commands(plugin_name: Option<String>, command_defs: &mut Vec<Co
...
+      );
+    }
+  }
+}
+
+/// Locates the permission files defined by the current crate's build script,
+/// probing the locations used by tauri-plugin (external plugin crates),
+/// tauri-build (inlined plugins and the app manifest) and the tauri crate
+/// itself (core plugins).
+fn find_defined_permission_files(
```
Legend-Master, line 192: I'm not sure if having the `tauri-macros` reading the permission files is a good idea, cc @FabianLars do you have some opinions on this?

FabianLars: I'd prefer not to do that either but i also cannot name an alternative if we want this warning

PR #15943, opened 2026-08-29
feat(tauri): opt-in forwarding of webview console output and page errors to the log crate

### LINKED ISSUE #15942: No supported console-to-backend pipe, and the hand-rolled one has three ordering hazards
There is no supported way to get webview console output and uncaught errors into the host process log. Everyone hand-rolls it, and the hand-rolled version has three hazards that each cost a debugging round.

## What I wrote, and what bit be

An initialization script that forwards `console.*`, `error` and `unhandledrejection` over an IPC command into `tracing`.

Three things had to be right:

**Buffering.**
An init script runs before the page's own scripts, and Tauri's IPC bootstrap is itself injected that way, so `__TAURI_INTERNALS__` is not reliably present when the script executes. Sending eagerly throws on the first message and loses exactly the window where boot failures happen. (I queue and flush on a short poll.)

**Capture phase.**
A failed `<script>` or `<link>` load fires an error event on the element and never reaches a bubbling window listener. A module that 404s is the failure most worth seeing, and it is the one a non-capturing handler misses.

**Not swallowing the rejection.**
My first version wrapped the invoke in a bare `try/catch`, which hid a missing permission (see #15940).

## Suggestion

`Builder::with_webview_logging()`, or a config flag, forwarding console and uncaught errors into `log`/`tracing` in debug builds.

That single feature makes a webview boot failure greppable in a CI log, which today it is not. It also removes the ordering hazard, since the framework can install the hook on the same path it installs the IPC bootstrap.

gronke:
There is no supported way to get webview console output and uncaught errors into the host process log; hand-rolled versions hit three ordering hazards (buffering before IPC is ready, capture-phase resource errors, swallowed invoke rejections).

`Builder::forward_webview_console(bool)`, disabled by default:
- Injects `console-forward.js` after the invoke system (its queue buffers until the IPC bridge is up - hazard 1 solved by construction) and before plugin/user scripts.
- Wraps `console.*` (original called first), captures `error` in the capture phase (failed `<script>`/`<link>`/`` loads never bubble - hazard 2) and `unhandledrejection` without `preventDefault()`.
- The forwarding invoke carries its own `.catch` reporting once through the original console, so failures never loop (hazard 3); payloads are size-capped and cycle-guarded.
- Rust side: new internal `core:webview` command `internal_log` (enabled in the default set, same pipeline as `internal_toggle_devtools`), mapping to `log::log!` with target `webview:{label}`; no-op when the flag is off.

Per-webview granularity and a config-file flag are possible additive follow-ups; log-only output keeps v1 minimal (apps can count errors via a wrapping `log::Log`).

Closes #15942

### DIFF
--- a/.changes/webview-console-forwarding.md
+++ b/.changes/webview-console-forwarding.md
@@ -1,0 +1,6 @@
+---
+"tauri": minor:feat
+---
+
+Add `Builder::forward_webview_console(bool)` to opt in to forwarding webview `console.*` output, uncaught errors (including capture-phase resource load errors) and unhandled promise rejections to the `log` crate under the `webview:{label}` target.
+Disabled by default; when disabled the capture script is not injected and forwarded messages are ignored.
--- a/.changes/webview-console-forwarding.md
+++ b/.changes/webview-console-forwarding.md
@@ -2,5 +2,4 @@
 "tauri": minor:feat
 ---
 
-Add `Builder::forward_webview_console(bool)` to opt in to forwarding webview `console.*` output, uncaught errors (including capture-phase resource load errors) and unhandled promise rejections to the `log` crate under the `webview:{label}` target.
-Disabled by default; when disabled the capture script is not injected and forwarded messages are ignored.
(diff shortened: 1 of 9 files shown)


amrbashir:
there is already a plugin for this, tauri-plugin-log

PR #15949, opened 2026-08-30
fix(android): register result launchers per activity

### LINKED ISSUE #15948: [bug] Android: crash when an activity is destroyed while another is alive (regression from #15798)
### Describe the bug

On Android, destroying an activity while another one is still alive crashes the app:

```
java.lang.RuntimeException: Unable to destroy activity {…/…MainActivity}
	at android.app.ActivityThread.performDestroyActivity(ActivityThread.java:6632)
	…
Caused by: java.lang.IllegalStateException: LifecycleOwner …MainActivity@a9c8edc is attempting to
register while current state is RESUMED. LifecycleOwners must call register before they are STARTED.
	at app.tauri.plugin.PluginManager.onDestroy
	at …TauriActivity.onDestroy(TauriActivity.kt:60)
```

`PluginManager.onDestroy` hands the manager's activity result launchers to whichever activity is
left:

```kotlin
activities.remove(activity)
val nextActivity = activities.firstOrNull()
if (nextActivity != null) {
  if (this.activity == activity) {
    this.activity = nextActivity
    registerResultLaunchers(nextActivity)   // ← throws
  }
}
```

`registerForActivityResult` may only be called before its owner reaches STARTED, and `nextActivity`
is by definition already running. There is no point during another activity's teardown at which
that call is legal, so the branch always throws when it is taken.

The easiest way to reach it is installing an APK over the running app: the package manager
momentarily leaves the new activity resumed while the old one is destroyed. The tunnel/service
processes survive, but the window dies and the user has to launch the app again.

This is a regression from #15798 (2026-08-03), which implemented the earlier
`// TODO: on destroy, we should change to a different activity`. Released 2.11.5 has no such code
path and is unaffected.

### Reproduction

1. Build any Tauri v2 Android app against `dev` (at or after 020919a1b).
2. Install and launch it.
3. With the app running and in the foreground, `adb install -r <the same apk>`.
4. The app crashes with the trace above; `adb logcat -b crash` shows it.

### Expected behavior

The activity is destroyed without crashing, and plugins can still start activities for result and
request permissions afterwards.

### Full `tauri info` output

```
[✔] Environment
    - OS: Arch Linux (rolling), Linux 7.2.0
    ✔ rustc: 1.96.1
    - Android device: Google Pixel 8 Pro, Android 17 (SDK 37)

[-] Packages
    - tauri [RUST]: git dev @ dd6befda (also reproduces on 020919a1b)
```

### Stack trace

Included above.

### Additional context

I have a fix — each activity registers its own launchers while it is being created, which is the
one moment registering is allowed, so `onDestroy` only has to drop the entry and point at whatever
remains. Happy to open a PR.

okhsunrog:
Closes #15948.

`PluginManager.onDestroy` moved the manager's activity result launchers to whichever activity was left, by calling `registerResultLaunchers` on it. That activity is by definition already running, and `registerForActivityResult` may only be called before its owner reaches STARTED, so the branch threw an `IllegalStateException` every time it was taken — there is no moment during another activity's teardown at which the call is legal. The launchers therefore stop being a property of the manager that migrates: each activity registers its own set while it is being created, which is the one point where registering is allowed, and `onDestroy` only has to drop the entry and point at whatever remains, which already has launchers of its own.

- register result launchers per activity in `onCreate` instead of once for the first one
- keep them in a `LinkedHashMap` keyed by activity, so the activity taken over when the current one goes away is the oldest survivor rather than an arbitrary member of a hash set
- `onDestroy` removes the entry and re-points `activity`; it no longer registers anything
- the three launch call sites read the current activity's set, so a missing one reports what went wrong instead of throwing an NPE from `!!`

### Testing

Reproduced on a Pixel 8 Pro (Android 17) by installing an APK over the running app, which momentarily leaves the new activity resumed while the old one is destroyed: the app crashed on every attempt before this change. With it, six consecutive `adb install -r` over the running app produced no crash and the app relaunched normally each time.

### DIFF
--- a/.changes/android-plugin-manager-relaunch.md
+++ b/.changes/android-plugin-manager-relaunch.md
@@ -1,0 +1,5 @@
+---
+tauri: patch:bug
+---
+
+On Android, fixed a crash when an activity is destroyed while another one is already running — installing an APK over the running app is the common way to hit it. The plugin manager tried to move its activity result launchers to the surviving activity, and `registerForActivityResult` rejects that with `IllegalStateException: LifecycleOwner ... is attempting to register while current state is RESUMED`. Each activity now registers its own launchers when it is created.
--- a/crates/tauri/mobile/android/src/main/java/app/tauri/plugin/PluginManager.kt
+++ b/crates/tauri/mobile/android/src/main/java/app/tauri/plugin/PluginManager.kt
@@ -34,12 +34,18 @@ object PluginManager {
     fun onResult(result: ActivityResult)
   }
 
-  private val activities: HashSet<AppCompatActivity> = HashSet()
+  /** The result launchers belonging to one activity. */
+  private class ResultLaunchers(
+    val startActivityForResult: ActivityResultLauncher<Intent>,
+    val startIntentSenderForResult: ActivityResultLauncher<IntentSenderRequest>,
+    val requestPermissions: ActivityResultLauncher<Array<String>>
+  )
+
+  // Insertion ordered, so the activity taken over when the current one goes away is the oldest
+  // surviving one rather than an arbitrary member of a hash set.
+  private val launchers: LinkedHashMap<AppCompatActivity, ResultLaunchers> = LinkedHashMap()
   var activity: AppCompatActivity? = null
   private val plugins: HashMap<String, PluginHandle> = HashMap()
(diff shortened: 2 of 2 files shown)


Legend-Master (approved):
Nice catch, awesome work, thanks!

okhsunrog:
@Legend-Master done, I signed the commit

PR #15952, opened 2026-08-31
Arc/Mutex refactors

sftse:
Some low-hanging refactors, removes one `Mutex` and one `Arc`.

### DIFF
--- a/.changes/change-pr-15952.md
+++ b/.changes/change-pr-15952.md
@@ -1,0 +1,5 @@
+---
+"tauri": patch:enhance
+---
+
+Internal refactors removing an `Arc` and a `Mutex`, no user-facing changes.
--- a/.changes/change-pr-15952.md
+++ b/.changes/change-pr-15952.md
@@ -1,5 +1,7 @@
 ---
 "tauri": patch:enhance
+"tauri": patch:changes
 ---
 
-Internal refactors removing an `Arc` and a `Mutex`, no user-facing changes.
+Internal refactors removing an `Arc` and a `Mutex`.
+Deprecate methods that should accidentally were made public.
--- a/crates/tauri/build.rs
+++ b/crates/tauri/build.rs
@@ -232,7 +232,7 @@ fn has_feature(list: &mut Vec<String>, feature: &str) -> bool {
 
   // when a feature is enabled, Cargo sets the `CARGO_FEATURE_<name>` env var to 1
   // <https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts>
-  std::env::var(format!("CARGO_FEATURE_{}", AsShoutySnakeCase(feature)))
+  std::env::var_os(format!("CARGO_FEATURE_{}", AsShoutySnakeCase(feature)))
     .map(|x| x == "1")
     .unwrap_or(false)
 }
@@ -255,12 +255,12 @@ fn main() {
 
   println!("cargo:dev={dev}");
 
-  let target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap();
+  let target_os = std::env::var_os("CARGO_CFG_TARGET_OS").unwrap();
   let mobile = target_os == "ios" || target_os == "android";
   alias("desktop", !mobile);
   alias("mobile", mobile);
 
-  let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
+  let out_dir = PathBuf::from(env::var_os("OUT_DIR").unwrap());
 
   let checked_features_out_path = out_dir.join("checked_features");
   std::fs::write(checked_features_out_path, checked_features.join(","))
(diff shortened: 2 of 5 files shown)


crates/tauri/src/ipc/channel.rs
```diff
@@ -169,10 +169,10 @@ impl JavaScriptChannelId {
             let data_id = CHANNEL_DATA_COUNTER.fetch_add(1, Ordering::Relaxed);
 
             webview
-              .state::<ChannelDataIpcQueue>()
-              .0
+              .state::<Mutex<ChannelDataIpcQueue>>()
```
Legend-Master: This looks a bit verbose, maybe we just remove `Arc` and keep the inner `Mutex`?

crates/tauri/src/menu/plugin.rs
```diff
@@ -885,32 +882,34 @@ fn set_icon<R: Runtime>(
   )
 }
 
-struct MenuChannels(Mutex<HashMap<MenuId, Channel<MenuId>>>);
+#[derive(Default)]
+pub(crate) struct MenuChannels(HashMap<MenuId, Channel<MenuId>>);
```
Legend-Master: This as well, it looks quite a bit more verbose to do the `Mutex` from outside

crates/tauri/build.rs
```diff
@@ -265,20 +260,17 @@ fn main() {
   alias("desktop", !mobile);
   alias("mobile", mobile);
 
-  let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
+  let out_dir = PathBuf::from(env::var_os("OUT_DIR").unwrap());
 
   let checked_features_out_path = out_dir.join("checked_features");
```
Legend-Master, line 265: Maybe we can add a bit of context here that we're at it, this is used for the unit test in tauri to check we're not checking non-existing features with `has_feature`

crates/tauri/src/ipc/channel.rs
sftse: I agree the Mutex is more verbose on the outside, but there's some value in making locking more explicit. In this case it's obvious everything was already safe before, but I'm thinking if this pattern were applied more consistently to the code, some of the deadlocks would have been more obvious.

All else equal, I think lifting locking to the highest possible scope is more verbose and less readable in the small, but clearer in the large.

I have more examples in mind, we can hold off these changes until I have a better case to show.

sftse: Stated differently, locks introduce non-trivial preconditions that are invisible to callers and because they're invisible can end up leaking everywhere. Discharging them as early as possible with a call to `lock` increases the code region borrow-checking statically prevents us from mistakes like this.

crates/tauri/src/menu/plugin.rs
```diff
@@ -885,32 +882,34 @@ fn set_icon<R: Runtime>(
...
         // Cloning the channel out in case the menu gets dropped during the channel send through `channel_interceptor`
-        let channel = app
-          .state::<MenuChannels>()
-          .0
-          .lock()
-          .unwrap()
-          .get(&e.id)
-          .cloned();
-        if let Some(channel) = channel {
+        if let Some(channel) = menu_channels.lock().unwrap().get(&e.id).cloned() {
```
Legend-Master: I'm a bit unsure if the lifetime of the lock is correct here?

crates/tauri/src/ipc/channel.rs
Legend-Master: I get this, thanks for explaining. But since in this specific case, we're calling lock explicitly anyways, shouldn't make much of a difference anyways.

And since we're in a state, `Mutex<ChannelDataIpcQueue>` can be slightly more error prone to get right.

crates/tauri/src/menu/plugin.rs
sftse: It isn't! Reverted.

crates/tauri/src/ipc/channel.rs
sftse: I've reverted these changes following your suggestion. In other places what I was aiming for would be more obvious I think, but it ends up being just confusing here.

crates/tauri/build.rs
sftse: I didn't get what that code was doing anyway? It looks like the code is anticipating multiple features here, which is where the `fn has_feature` function would make sense, but it's only called a single time with `has_feature("custom-protocol")`.

sftse: Cleaned it up.

crates/tauri/src/ipc/mod.rs
```diff
@@ -565,13 +560,13 @@ impl<R: Runtime> InvokeMessage<R> {
...
-    self.state.clone()
+    self.webview.manager.state.clone()
   }
 
   /// A reference to the state manager associated with application.
   #[inline(always)]
   pub fn state_ref(&self) -> &StateManager {
-    &self.state
+    &self.webview.manager.state
   }
```
sftse, line 570: Candidates to make private in v3, think they are pub by accident.

crates/tauri/build.rs
Legend-Master: > I didn't get what that code was doing anyway?

> this is used for the unit test in tauri to check we're not checking non-existing features with `has_feature`

I think we might still want to keep this function in case we do something with it in the future

crates/tauri/src/ipc/mod.rs
Legend-Master: Agreed, we could put in a todo comment + a deprecation warning, people should use https://docs.rs/tauri/latest/tauri/trait.Manager.html#method.state instead

sftse:
Adopted your suggestions, excuse the delay, things get in the way.

.changes/change-pr-15952.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri": patch:enhance
+---
+
+Internal refactors removing an `Arc` and a `Mutex`, no user-facing changes.
```
Legend-Master: Let's update the change file to include the deprecations

Legend-Master (commented):
No worries, take your time 🙃

.changes/change-pr-15952.md
```diff
@@ -0,0 +1,7 @@
+---
+"tauri": patch:enhance
+"tauri": patch:changes
+---
+
+Internal refactors removing an `Arc` and a `Mutex`.
+Deprecate methods that should accidentally were made public.
```
Legend-Master:
```suggestion
---
"tauri": patch:changes
---

Internal refactors removing an `Arc` and a `Mutex`.
Deprecate `InvokeMessage::state` and `InvokeMessage::state_ref` that should accidentally were made public.
```

Legend-Master (approved):

PR #15967, opened 2026-09-03
ci: add explicit workflow permissions

Legend-Master:
Reference: https://github.com/tauri-apps/plugins-workspace/commit/1a7b12c10256567aca763f4d47cc4da886f73c63

### DIFF
--- a/.github/workflows/covector-version-or-publish.yml
+++ b/.github/workflows/covector-version-or-publish.yml
@@ -9,6 +9,9 @@ on:
     branches:
       - dev
 
+permissions:
+  contents: read
+
 jobs:
   run-integration-tests:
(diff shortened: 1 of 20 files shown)


.github/workflows/covector-version-or-publish.yml
```diff
@@ -9,6 +9,9 @@ on:
     branches:
       - dev
 
+permissions:
```
Legend-Master, line 12: Be careful with this, I assume per job level permissions down below should override workflow level permissions, but I can't find any docs on this

FabianLars: Wow the docs on that are really very unfortunate

FabianLars (approved):

PR #15971, opened 2026-09-03
refactor: `WebviewBuilder::bounds/size/position`

amrbashir:


crates/tauri/src/webview/mod.rs
```diff
@@ -433,7 +446,11 @@ async fn create_window(app: tauri::AppHandle) {
 
     Self {
       label: config.label.clone(),
-      webview_attributes: WebviewAttributes::from(&config),
+      webview_attributes: {
+        let mut attrs = WebviewAttributes::from(&config);
+        attrs.bounds = Some(Self::default_bounds());
```
Legend-Master: I think the default was to resize to the window size? https://github.com/tauri-apps/tauri/blob/7f2b7cb7921caf7d9370b7ebe99e3cf47dbcfed6/crates/tauri-runtime-wry/src/lib.rs#L4865-L4877

amrbashir: restored in [07d0a4e](https://github.com/tauri-apps/tauri/pull/15971/commits/07d0a4e64c9c8b42d2ab64b13152de716312e1a3)

Legend-Master (approved):
I know this is an unstable feature, but I still feel like we should target this to v3 instead that this is not urgent in anyway to be included in v2

crates/tauri/src/webview/mod.rs
```diff
@@ -230,7 +232,7 @@ macro_rules! unstable_struct {
 
       #[cfg(not(any(test, feature = "unstable")))]
       pub(crate) $($tokens)*
-    }
+  }
```
Legend-Master, line 235: revert this

Legend-Master (approved):

PR #15981, opened 2026-09-04
feat(bundler): build CEF helper binary on bundle time

lucasfernog:
embedding at compile time forces you to have a rustup target you didn't really expect initially


FabianLars:
Not a fan of this tbh. Why don't we only compile the cef-helper in a universal way when the build itself is universal?

lucasfernog:
> Not a fan of this tbh. Why don't we only compile the cef-helper in a universal way when the build itself is universal?

i'm trying to decide an approach for this - prebuilt cef helper is not good, and embedding into bundler is the safest approach IMO
but having to install the rustup target is boring too (tho only impacts "cargo tauri" users)
and yeah the helpers should be arch specific, i'll push that once we decide on the distribution approach

lucasfernog:
> Why don't we only compile the cef-helper in a universal way when the build itself is universal?

oh actually we already had this @FabianLars

FabianLars:
do we really need the autoinstall then?

lucasfernog:
> do we really need the autoinstall then?

not for Node.js users, but if you want to "cargo install" it, you'll need to have the target installed regardless (unless we change it to compile the helper at runtime)

lucasfernog:
moving forward to see this in practice, i didn't really like the embedding approach and a prebuilt helper would be bad too IMO

Issue #14011, opened 2025-08-15
Clean up Top-Level exports


FabianLars:
As mentioned here: https://github.com/tauri-apps/tauri/pull/14009 it's getting quite crowded and maybe we should organize it a bit more. Again taking 14009 as an example, all the dpi re-exports could be a exported similar to utils/http.

Another example, we already have webview and window modules but have quite a view webview / window exports on the top level as well eg WebviewEvent and WebviewUrl

WSH032:
Ref #14013

> Also, I feel that TitleBarStyle should be exported under mod window (instead of the top-level module), similar to how window::ProgressBarState is handled, since it's only used for Window::set_title_bar_style.

I propose that `TitleBarStyle` should also join this party 🤓

PR #15982, opened 2026-09-04
feat: add prelude module

lucasfernog:


FabianLars:
Conflicts a bit with https://github.com/tauri-apps/tauri/issues/14011 so imo we should think twice what we want in a prelude to not make the same mistake in a second location.
btw where does the idea even come from?

amrbashir (approved):

lucasfernog:
> Conflicts a bit with #14011 so imo we should think twice what we want in a prelude to not make the same mistake in a second location. btw where does the idea even come from?

just trying to reduce the amount of types we have on our crate root :/ also didn't want to break the api just for this so that's why i went with a prelude + hidden re-export

FabianLars:
just saying that some of these types shouldn't be in the crate root nor in a prelude, and not all of them should really be on the same level.

I mean, i'm not blocking the PR. It's just not the end and we'll have to make more changes.

lucasfernog:
> just saying that some of these types shouldn't be in the crate root nor in a prelude, and not all of them should really be on the same level.
> 
> I mean, i'm not blocking the PR. It's just not the end and we'll have to make more changes.

yeah I just didn't really want to break the API for this. I think this is a tech debt from the days the tauri runtime crate was introduced, where Chip started adding types to tauri root and we kept doing it.

PR #15985, opened 2026-09-05
refactor: decouple the webview runtime from the `tauri` crate

lucasfernog:
`tauri` no longer depends on `tauri-runtime-wry` or `tauri-runtime-cef`, and the `wry`/`cef` features are gone. Apps add the runtime crate and select it on the builder:

```rust
tauri::Builder::default()
  .runtime(tauri_runtime_wry::Wry) // or tauri_runtime_cef::Cef::default()
```

`Builder::default()` uses the new type-erased `tauri::DynRuntime` (`tauri_runtime::dynamic`), which is also the default `R` of `AppHandle`, `Window`, `Webview` and friends, so apps keep using those types without generics and plugins stay generic over `R: Runtime`. Static dispatch is still available with `Builder::<tauri_runtime_wry::WryRuntime>::new()`.

- tauri-runtime: `dynamic` module with erased runtime, handle, proxy and dispatchers (blanket impls over any `Runtime`) and a recording `DynWindowBuilder`; `RuntimeSpecificInitAttrs<T>` gains `type Runtime` and doubles as the runtime selector; `custom_scheme_url` moves to `RuntimeHandle`; `RuntimeHandle::webview_version` and `WebviewDispatch::with_ios_webview` added; devtools methods are always required (bodies stay cfg-gated).
- tauri: `wry`, `cef`, `x11`, `dbus` and `macos-proxy` features, the `Wry`/`Cef` aliases and CEF re-exports, `webview_version()` and `android_binding!` removed; runtime-specific methods on `AppHandle`, `App`, `Webview(Window)` and the builders moved to extension traits; `default_runtime` macro replaced by `R = DynRuntime` defaults; `convertFileSrc` takes the scheme URL format from the runtime; `PlatformWebview::downcast_ref` added.
- tauri-runtime-wry: `Wry<T>` renamed to `WryRuntime<T>` with `Wry` as the selector; now depends on `tauri` and provides `AppHandleWryExt`, `AppWryExt`, `WebviewWryExt`, `WebviewWindowBuilderWryExt`, `WebviewBuilderWryExt` and the `android_binding!` macro used by `mobile_entry_point`.
- tauri-runtime-cef: `RuntimeInitAttrs` renamed to `Cef`; now depends on `tauri` and provides `WebviewCefExt`, `WebviewWindowBuilderCefExt`, `WebviewBuilderCefExt` and `cef_entry_point`; `links = "tauri_runtime_cef"` lets `tauri-build` detect it.
- tauri-build / tauri-cli: CEF is detected through the `tauri-runtime-cef` dependency (`DEP_TAURI_RUNTIME_CEF_RUNTIME`) and the app manifest instead of the `cef` feature; templates, `add`, `info` and `metadata-v2.json` updated.
- Examples moved to `crates/tauri-examples`; bench and example apps, CI, docs and covector config updated.

Supersedes #15068.


bench/tests/cpu_intensive/src-tauri/src/main.rs
```diff
@@ -10,12 +10,12 @@ fn app_completed_successfully() {
...
 
-#[cfg_attr(feature = "cef", tauri::cef_entry_point)]
+#[cfg_attr(feature = "cef", tauri_runtime_cef::cef_entry_point)]
 fn main() {
   #[cfg(feature = "cef")]
-  let builder = tauri::Builder::<tauri::Cef>::default();
+  let builder = tauri::Builder::default().runtime(tauri_runtime_cef::Cef::default());
   #[cfg(not(feature = "cef"))]
-  let builder = tauri::Builder::<tauri::Wry>::default();
+  let builder = tauri::Builder::default().runtime(tauri_runtime_wry::Wry);
```
amrbashir, line 18: It feels weird that one uses `Cef::default()` and the other is just `Wry`

crates/tauri-build/src/lib.rs
```diff
@@ -250,11 +250,16 @@ fn cfg_alias(alias: &str, has_feature: bool) {
...
+///
+/// `tauri-runtime-cef` emits a `cargo:runtime=cef` instruction from its build script, which cargo
+/// exposes to the build scripts of its direct dependents (the application) as `DEP_TAURI_RUNTIME_CEF_RUNTIME`.
+fn uses_cef_runtime() -> bool {
+  env::var("DEP_TAURI_RUNTIME_CEF_RUNTIME").as_deref() == Ok("cef")
+}
 
-  if runtime == "cef" {
+fn default_windows_app_manifest() -> &'static str {
+  if uses_cef_runtime() {
```
amrbashir, line 262: Just something to think about for the future, not necessary for this PR. 

Do we want to make the runtime fully generic? I meant tauri-build, tauri-bundler and tauri-cli are not generic. They all have special handling for specific runtimes, for now it is wry and cef. 

Other runtimes like tauri-runtime-verso doesn't have the same feature set and they can't contribute to the build and bundle systems at all.

The question is, do we want to invest in this type of modularity?

.changes/bundler-webview-runtime.md
```diff
@@ -0,0 +1,5 @@
+---
+"tauri-bundler": minor:feat
+---
+
+Added `BundleSettings::webview_runtime` (`WebviewRuntime::{Wry, Cef { distribution }, Other}`, defaults to `Wry`) so the bundler only ships what the webview runtime needs: the webkit2gtk helper processes in the AppImage, and the WebView2 installation step (`webviewInstallMode`, `minimumWebview2Version` and the `WebView2Loader.dll` resource) in the NSIS and WiX installers are only included for the wry runtime. It replaces the `cef_path` and `cef_shared_runtime` settings: `WebviewRuntime::Cef { distribution: Some(path) }` embeds the CEF distribution and `WebviewRuntime::Cef { distribution: None }` is an app on a shared CEF runtime.
```
amrbashir, line 5: This change file is a bit hard to read and needs to be reworded and split into multiple list items.

crates/tauri-cli/src/interface/rust/manifest.rs
```diff
@@ -517,3 +702,134 @@ mod tests {
...
+    let m = manifest(
+      r#"
+[dependencies]
+tauri = "2"
+tauri-runtime-wry = "2"
+tauri-runtime-cef = "0.1"
+"#,
+    );
+    assert_eq!(m.runtime(&[], LINUX), Runtime::Cef);
+  }
```
amrbashir, line 765: why cef takes precedence here ? is it because of alphabet order, or because it is last in deps ? not clear why

crates/tauri-cli/src/interface/rust.rs
amrbashir, line 1469: feels like these could just be moved to runtime.linux_dependencies()

crates/tauri-runtime-cef/src/runtime.rs
```diff
@@ -168,7 +176,9 @@ impl RuntimeInitAttrs {
   }
 }
 
-impl tauri_runtime::RuntimeSpecificInitAttrs for RuntimeInitAttrs {
+impl<T: UserEvent> tauri_runtime::RuntimeSpecificInitAttrs<T> for Cef {
```
amrbashir, line 179: nit: We could now rename `RuntimeSpecificInitAttrs` to just `RuntimeInitAttrs`

crates/tauri-runtime-cef/src/tauri_ext.rs
```diff
@@ -0,0 +1,206 @@
...
+impl<R: Runtime> WebviewBuilderCefExt for tauri::webview::WebviewBuilder<R>
+where
+  R::PlatformSpecificWebviewAttribute: FromCefWebviewAttribute,
+{
+  fn browser_runtime_style(mut self, style: RuntimeStyle) -> Self {
+    self.platform_specific_attribute(FromCefWebviewAttribute::from_cef(
+      WebviewAtribute::RuntimeStyle { style },
+    ));
+    self
+  }
```
amrbashir, line 205: This could be included in this PR or I could open another PR after this is merged.


I think there is room for improvement in this API. I don't like `enum WebviewAttribute` in tauri-runtime-wry and the mistyped `enum WebviewAtribute` in tauri-runtime-cef, both could both be refactored into a struct and I forgot to refactor them in my last PR.

The final API could look something like this.


```suggestion
impl<R: Runtime> WebviewBuilderCefExt for tauri::webview::WebviewBuilder<R>
where
  R::RuntimeWebviewBuilderAttributes: WebviewAttributes,
{
  fn browser_runtime_style(mut self, style: RuntimeStyle) -> Self {
    self.runtime_specific_attributes.runtime_style = syle;
    self
  }
```

Notice `PlatformSpecificWebviewAttribute` renamed to a consistent name `RuntimeWebviewAttributes` and `WebviewBuilder::platform_specific_attributes` field to `WebviewBuilder::runtime_specific_attributes`

This will probably eliminate the need for `FromCefWebviewAttribute` trait

crates/tauri-runtime-wry/src/lib.rs
```diff
@@ -2695,15 +2789,29 @@ pub trait Plugin<T: UserEvent> {
...
 
+/// Selects and configures the wry runtime.
+///
+/// Pass it to `tauri::Builder::runtime` to run the application with wry:
+///
+/// ```rust,no_run
+/// tauri::Builder::default().runtime(tauri_runtime_wry::Wry);
+/// ```
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct Wry;
```
amrbashir, line 2800: As I stated above, and just to make it consistent we should make this an empty struct, and disable clippy warning for it
```suggestion
pub struct Wry {}
```

so users will be forced to use it with `tauri_runtime_wry::Wry::default()`

crates/tauri-runtime-wry/src/tauri_ext.rs
```diff
@@ -0,0 +1,283 @@
...
+impl<R: Runtime> WebviewWryExt for WebviewWindow<R>
+where
+  R::Handle: AsWryHandle,
+{
+  fn with_wry_webview<F: FnOnce(&crate::Webview) + Send + 'static>(&self, f: F) -> Result<()> {
+    self.as_ref().with_wry_webview(f)
+  }
+}
+
+macro_rules! webview_builder_ext {
```
amrbashir, line 180: Similar feedback as https://github.com/tauri-apps/tauri/pull/15985/changes#r3942670338

crates/tauri/src/app.rs
```diff
@@ -775,21 +764,17 @@ impl<R: Runtime> ManagerBase<R> for App<R> {
...
   ///
-  /// This API is unstable.
-  pub fn wry_plugin<P: tauri_runtime_wry::PluginBuilder<EventLoopMessage> + Send + 'static>(
-    &mut self,
-    plugin: P,
-  ) where
-    <P as tauri_runtime_wry::PluginBuilder<EventLoopMessage>>::Plugin: Send,
-  {
-    self.handle.runtime_handle.plugin(plugin);
+  /// Mostly useful for runtime-specific extension traits.
```
amrbashir, line 770: incorrect documentation

crates/tauri/src/app.rs
```diff
@@ -1568,29 +1553,34 @@ pub(crate) struct InvokeInitializationScript<'a> {
...
+  /// # Examples
+  ///
+  /// ```rust,ignore
+  /// tauri::Builder::default()
+  ///   .runtime(tauri_runtime_wry::Wry)
+  ///   .run(tauri::generate_context!())
+  ///   .expect("error while running tauri application");
+  /// ```
+  #[must_use]
+  pub fn runtime<A: RuntimeSpecificInitAttrs<EventLoopMessage>>(mut self, attrs: A) -> Self {
```
amrbashir, line 1581: There is 
```rs
  pub fn runtime_init_attrs(mut self, attrs: R::RuntimeInitAttrs) -> Self {
```

a few lines below, why keep both?

crates/tauri/src/app.rs
```diff
@@ -1568,29 +1553,34 @@ pub(crate) struct InvokeInitializationScript<'a> {
...
+  /// Selects the runtime that powers the application, using its initialization attributes.
+  ///
+  /// Every runtime crate provides such a type, e.g. `tauri_runtime_wry::Wry` or `tauri_runtime_cef::Cef`.
+  /// Building the application fails with [`tauri_runtime::Error::RuntimeNotConfigured`] if no runtime was selected.
+  ///
+  /// # Examples
+  ///
+  /// ```rust,ignore
+  /// tauri::Builder::default()
+  ///   .runtime(tauri_runtime_wry::Wry)
```
amrbashir: any particular reason why we choose this way of selecting the runtime, instead of just 
```rs
tauri::Builder::<tauri_runtime_wry::WryRuntime>::default()
```

We could even rename `WryRuntime` to just `Wry` and current PR `Wry` to `WryAttributes`

crates/tauri/src/window/mod.rs
```diff
@@ -134,6 +132,18 @@ unstable_struct!(
   }
 );
 
+#[cfg_attr(not(feature = "unstable"), allow(dead_code))]
+impl<'a, R: Runtime, M: Manager<R>> WindowBuilder<'a, R, M> {
+  /// Returns a mutable reference to the runtime's window builder.
+  ///
+  /// Mostly useful for runtime-specific extension traits.
+  pub fn runtime_window_builder_mut(
```
amrbashir, line 140: Not sure why would this be needed?

crates/tauri-runtime/src/lib.rs
```diff
@@ -411,16 +432,61 @@ pub struct RuntimeInitArgs<A> {
   pub runtime_init_attrs: A,
 }
 
+impl<A> RuntimeInitArgs<A> {
+  /// Replaces the runtime-specific attributes, returning the new arguments and the previous attributes.
+  pub fn with_attrs<B>(self, runtime_init_attrs: B) -> (RuntimeInitArgs<B>, A) {
```
amrbashir, line 437: Not sure what's the need for this method, and it seems like it is only used in dynamic.rs and based on its uses cases there, it should be deleted no?

crates/tauri-build/src/lib.rs
lucasfernog: I want to, it's just that having plugins for the CLI is a lot more complicated since we would need to rely on separate binaries and define a protocol

amrbashir (approved):

lucasfernog:
moving forward, we can make changes later on ahead of the first 3.0 alphas

PR #15984, opened 2026-09-05
feat(cef): expose native frame and state observations

zackarychapple:
## Summary

This change adds native browser state APIs to the Chromium Embedded Framework (CEF) runtime. Applications can identify windows, browsers, documents, and dialogs. Applications can use these identities to reject actions after a target changes.

The runtime reports lifecycle events for main frames and child frames. It checks all current frame identities before it returns a document token. A token identifies one document state in one browser. Navigation, frame replacement, and renderer termination invalidate previous tokens.

The runtime reports the native parent, bounds, and visibility of a webview. Distinct window tokens identify replacement windows, even when a window label stays the same. One request ID allocator prevents collisions between runtime requests and application requests.

The runtime tracks CEF popup creation, cancellation, and closure. Each popup has independent frame and dialog state. The runtime preserves the popup window and its `window.opener` relationship. Dialog observations exclude messages, URLs, and prompt contents. The runtime preserves the default CEF dialog interface.

This change also corrects two causes of UI thread deadlock. Popup callbacks receive the opener URL directly, so applications can check popup policy without a synchronous URL request. Synchronous getters release the window mutex before they wait for a UI thread reply. Page-load callbacks can then acquire the same mutex. Native frame diagnostics contain counts and flags.

The runtime preserves browser windows until CEF confirms application shutdown. Popup cleanup remains active during shutdown. Applications can call `prepare_macos_application` before a native startup dialog creates the AppKit application.

Applications select the runtime through `Builder::runtime`. CEF APIs use extension traits from `tauri-runtime-cef`. The Android runtime uses one named callback type to pass the strict Clippy check.

CEF reports renderer termination reasons and native error details on each desktop platform. The dynamic runtime preserves the termination callback. WebKit reports an unknown reason when it cannot supply details. This preserves application recovery decisions.

The macOS helper build uses the resolved CEF crate sources from Cargo. The helper preserves Git and local patches. The helper build no longer uses the framework version as a Rust crate version. Missing or ambiguous sources stop the build.

## Review notes

The getter correction targets the reproduced mutex cycle. Native hang qualification follows the [existing acceptance record](https://github.com/tauri-apps/tauri/pull/15984#issuecomment-5565587846). The historical reports remain unattributed. The bounded regression does not establish native after-fix behavior.

Native validation of the latest mutex correction remains incomplete. General synchronous getters still have no reply timeout.

[...]

PR #15990, opened 2026-09-06
fix(cef): prepare early AppKit UI and retain closing windows

zackarychapple:
## Summary

Native recovery dialogs can create an `NSApplication` object before the Chromium Embedded Framework (CEF) runtime starts. CEF then rejects the incompatible application class and stops startup.

This change exports `tauri_runtime_cef::prepare_macos_application`. Applications call this function on the main thread before displaying native startup dialogs. The function prepares AppKit, the macOS UI framework. It does not start CEF, run the event loop, or open a browser profile. Repeated calls use the same application object. The function checks the current thread before calling AppKit.

During application exit, the runtime keeps browser and window records until CEF confirms closure. The runtime also closes attached DevTools windows. Previously, the runtime removed these records before native window cleanup completed. On macOS and Windows, cleanup could then fail to find the webview and leave the process running.

## Testing

The existing validation record reports these results on macOS arm64:

- `cargo clippy -p tauri-runtime-cef --all-features`: passed.
- `cargo test -p tauri-runtime-cef --all-features`: passed, including the native application preparation test and three documentation tests.
- `cargo fmt --all`: completed successfully.

The native test runs on the main thread. It checks that early and repeated preparation use the same application object. It also checks the CEF event-state methods. The test confirms that preparation rejects a background thread and an incompatible application object. A separate process tests the incompatible object. The test catches the expected failures.

## Manual Testing

1. On macOS, call `prepare_macos_application()` before opening a native recovery dialog. Confirm that AppKit uses the CEF application class.
2. Dismiss the dialog and start the CEF runtime. Confirm that startup uses the same application object and completes without the class assertion failure.
3. Open a browser window and its DevTools window. Quit the application. Confirm that both windows close and the process exits without forced termination.
4. Repeat the shutdown check on Windows. Confirm that native child-window cleanup completes and the process exits.

Earlier tests in a signed macOS arm64 application reproduced the startup failure. Tests with a separate runtime backport then completed restart and quit without forced termination. The restart tests also included an application launcher fix. These results support the diagnosis. They do not establish application-level validation of this PR revision or Windows behavior.

### DIFF
--- a/.changes/cef-native-startup-ui.md
+++ b/.changes/cef-native-startup-ui.md
@@ -1,0 +1,5 @@
+---
+'tauri-runtime-cef': 'patch:enhance'
+---
+
+Expose macOS CEF application preparation so native recovery dialogs can open before the runtime or browser profile initializes, without creating an incompatible AppKit application singleton.
--- a/.changes/cef-native-startup-ui.md
+++ b/.changes/cef-native-startup-ui.md
@@ -3,3 +3,5 @@
 ---
 
 Expose macOS CEF application preparation so native recovery dialogs can open before the runtime or browser profile initializes, without creating an incompatible AppKit application singleton.
+
+Retain browser windows until CEF acknowledges application shutdown, so queued native child-view destruction can complete and the process can exit. Close attached DevTools as part of that shutdown.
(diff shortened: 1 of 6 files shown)


zackarychapple:
Superseded by #15984, merged in 8e9376f0cc33b151fada68d6ee3536c162a3839c. The feat/cef branch already contains the application preparation API, native bootstrap test, shutdown-state retention, and DevTools closure from this PR.

PR #15997, opened 2026-09-08
enahnce(cli/init): skip prompts automatically in non interactive environments

junaidmirr:
When running the tauri init command in environments such as CI systems, automated scripts, or other situations where no interactive terminal is available, the command currently attempts to prompt the user for input. This can result in an I/O error when the prompt cannot read from standard input.

As a result, tauri init cannot be reliably used in non interactive environments unless the --ci flag is explicitly provided. While --ci provides a way to use the command without prompts, requiring it for every automated invocation is easy to overlook.

The current implementation relies on the --ci flag to disable prompts, but does not automatically determine whether standard input is connected to an interactive terminal. The --force flag is separate from this behavior and is intended only to control overwriting of an existing src-tauri directory.

Proposed change

This pull request uses Rust's standard library IsTerminal trait to detect whether standard input is connected to an interactive terminal.

When standard input is not connected to a terminal, tauri init automatically follows the same non interactive path used when --ci is provided. This prevents the command from attempting to prompt for input when there is no user available to respond, while preserving the existing behavior when an interactive terminal is available.

Explicit command line options such as --app-name, --frontend-dist, and --dev-url can still be used to override the default configuration values.

The behavior of --force is unchanged. It continues to control whether an existing src-tauri directory can be overwritten and does not affect prompting behavior.

Design consideration

This change introduces behavior based on whether the process has access to an interactive terminal. In most cases, the absence of a terminal indicates CI or automation, but standard input can also be redirected or piped for other reasons.

The goal is to avoid attempting interactive prompts when there is no user available to respond, while keeping the existing behavior unchanged for normal terminal usage.

I chose terminal detection instead of introducing another flag because it allows non interactive invocations to work without requiring users to remember an additional option.

I am open to feedback on whether detecting a non interactive terminal is the appropriate approach, or whether this behavior would be better controlled through an explicit option.

Testing

• Manually tested tauri init in a non interactive environment and verified that it completes without prompting.

• Verified that explicit command line options continue to override the default values.

• Added a regression test covering non interactive execution.

• The implementation uses Rust's standard IsTerminal API and is cross platform.

Documentation

• Updated the README with a Building from Source section explaining how to build the CLI from source.

• Documented the automatic terminal detection behavior.

• Added a changeset describing the fix for the release notes.

Overall, this makes tauri init more reliable when used in CI and automation while preserving the existing behavior for normal interactive terminal usage.


README.md
amrbashir, line 1: Revert all changes to this file.

init.rs
amrbashir, line 1: what is this ?

crates/tauri-cli/src/init.rs
```diff
@@ -74,95 +75,98 @@ struct InitDefaults {
...
-        serde_json::from_str(&package_json_text).context("failed to parse JSON")?;
-      let (framework, _) = infer_framework(&package_json_text);
-      InitDefaults {
-        app_name: package_json.product_name.or(package_json.name),
-        framework,
-      }
-    } else {
-      Default::default()
-    };
+        let init_defaults = if package_json_path.exists() {
```
amrbashir: revert all indentations from this file, your change should only be the 3 lines at top.

.changes/cli-force-noninteractive.md
```diff
@@ -0,0 +1,5 @@
+---
+'tauri-cli': 'patch:bug'
```
amrbashir:
```suggestion
'tauri-cli': 'patch:enhance'
'@tauri-apps/cli': 'patch:enhance'
```

`enhance` is a more appropriate change category I think. Also include a bump to the JS CLI.

crates/tauri-cli/src/init.rs
```diff
@@ -138,28 +142,28 @@ impl Options {
...
+            .before_build_command
+            .map(|s| Ok(Some(s)))
+            .unwrap_or_else(|| {
+                prompts::input(
+                    "What command should Tauri run before `tauri build` to build your frontend? (leave empty if not needed)",
+                    Some(default_build_command(detected_package_manager).into()),
+                    self.ci,
+                    true,
+                )
+            })?;
```
amrbashir: all of these indentations still need to be reverted, it is not part of the goal of the PR

README.md
```diff
@@ -93,4 +100,4 @@ Logo: CC-BY-NC-ND
 
 - Original Tauri Logo Designs by [Alve Larsson](https://alve.io/), [Daniel Thompson-Yvetot](https://github.com/nothingismagick) and [Guillaume Chau](https://github.com/akryum)
 
-[![FOSSA Status](https://app.fossa.com/api/projects/git%2Bgithub.com%2Ftauri-apps%2Ftauri.svg?type=large)](https://app.fossa.com/projects/git%2Bgithub.com%2Ftauri-apps%2Ftauri?ref=badge_large)
+[![FOSSA Status](https://app.fossa.com/api/projects/git%2Bgithub.com%2Ftauri-apps%2Ftauri.svg?type=large)](https://app.fossa.com/projects/git%2Bgithub.com%2Ftauri-apps%2Ftauri?ref=badge_large)
```
amrbashir: These formatting changes also need to be reverted, not part of the PR

README.md
```diff
@@ -1,12 +1,19 @@
...
+
 [![FOSSA Status](https://app.fossa.com/api/projects/git%2Bgithub.com%2Ftauri-apps%2Ftauri.svg?type=shield)](https://app.fossa.com/projects/git%2Bgithub.com%2Ftauri-apps%2Ftauri?ref=badge_shield)
+
 [![Chat Server](https://img.shields.io/badge/chat-discord-7289da.svg)](https://discord.com/invite/tauri)
+
 [![website](https://img.shields.io/badge/website-tauri.app-purple.svg)](https://tauri.app)
-[![https://good-labs.github.io/greater-good-affirmation/assets/images/badge.svg](https://good-labs.github.io/greater-good-affirmation/assets/images/badge.svg)](https://good-labs.github.io/greater-good-affirmation)
+
+[![Greater Good](https://good-labs.github.io/greater-good-affirmation/assets/images/badge.svg)](https://good-labs.github.io/greater-good-affirmation)
+
```
amrbashir: These also need to be reverted, they are not part of the goal of the PR

.changes/cli-force-noninteractive.md
amrbashir: This is still need to be addressed

crates/tauri-cli/src/init.rs
amrbashir: @junaidmirr this indentations is still not reverted

amrbashir (approved):

amrbashir:
@junaidmirr you need to sign all your commits in the PR otherwise I can't merge the PR.

junaidmirr:
> @junaidmirr you need to sign all your commits in the PR otherwise I can't merge the PR.

I have resigned all the commits in the PR. Thank you!

PR #16007, opened 2026-09-11
feat(tray): expose a stable Windows GUID identity via TrayIconBuilder::guid

dylanh724:
## What

Adds `TrayIconBuilder::guid(u128)`, a passthrough for `tray_icon::TrayIconBuilder::with_guid` added in tauri-apps/tray-icon#366, and bumps `tray-icon` to `0.25`.

## Why

Windows keys the per-icon "always show in the taskbar" setting on the icon's identity. Without `NIF_GUID` that identity is the executable path, so any update that moves the binary (installers with per-version directories such as Scoop, Squirrel) turns the icon into a new one and it drops back into the overflow. With a fixed GUID and a same-publisher Authenticode signature, Windows preserves the setting ([NOTIFYICONDATA troubleshooting](https://learn.microsoft.com/windows/win32/api/shellapi/ns-shellapi-notifyicondataw#troubleshooting)). Electron exposes the same knob as `Tray`'s `guid` option.

## Status

**Draft until tray-icon 0.25.0 is published** (release PR tauri-apps/tray-icon#353). The `Cargo.toml` bump will not resolve before then.

Runtime-tested in the Echophrase desktop app through a Cargo patch of both crates.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/tauri/src/tray/mod.rs
+++ b/crates/tauri/src/tray/mod.rs
@@ -267,6 +267,27 @@ impl<R: Runtime> TrayIconBuilder<R> {
     self
   }
 
+  /// Set a stable identity GUID for this tray icon, as a UUID in `u128` form.
+  ///
+  /// Windows keys per-icon user settings (most importantly whether the icon is
+  /// pinned to the taskbar or hidden in the overflow) on the icon's identity.
+  /// Without a GUID that identity is the executable path, so the setting is
+  /// lost whenever an update moves the binary (for example installers that use
+  /// per-version directories). With a GUID, and an executable that is
+  /// Authenticode-signed by the same publisher across versions, the setting
+  /// survives.
+  ///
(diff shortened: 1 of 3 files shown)


crates/tauri/src/tray/mod.rs
```diff
@@ -267,6 +267,27 @@ impl<R: Runtime> TrayIconBuilder<R> {
     self
   }
 
+  /// Set a stable identity GUID for this tray icon, as a UUID in `u128` form.
```
Legend-Master, line 270: Note to reviewers: be careful with the documentation on this, I can't remember exactly but there's quite a lot of caveats to use this correctly.

amrbashir: as long as we link the MSDN docs, we should be fine

PR #16017, opened 2026-09-14
fix(cli/mobile): normalize gradlew CRLF on Windows hosts too

k6G52m4Dz75W:
## Summary

`ensure_gradlew` — which rewrites CRLF to LF in the Android `gradlew`
wrapper — is gated behind `#[cfg(unix)]`, both at the call site in
`ensure_init` and on the function itself. On Windows hosts the
normalization silently never runs, so a `gradlew` with CRLF endings
breaks `sh ./gradlew` invocations from Git-Bash/WSL with
`/usr/bin/env: ‘sh\r’: No such file or directory`.

This splits the function so the platform-neutral LF rewrite runs on
every host, while the unix exec-bit fix remains gated.

## Why this matters

A CRLF `gradlew` breaks every `sh`-based invocation. Windows users
cloning a project whose committed `gradlew` carries CRLF (or checking
out with `core.autocrlf=true`) hit this on the first
`tauri android build`. The normalization is cheap, platform-neutral,
and the function already existed — it just never ran off-unix.

## Changes

- removed the `#[cfg(unix)]` gate on the `ensure_gradlew` call site so
  the CRLF rewrite runs on all hosts;
- moved the unix-only exec-bit fix inside the function body, gated with
  `#[cfg(unix)]` (exec bits are a unix permission concept);
- the LF rewrite itself is unchanged.

## Verification

On a Windows host:

- placed a CRLF `gradlew` in a project's `gen/android`, ran
  `tauri android build`, and confirmed the file was rewritten to LF
  during `ensure_init`;
- the subsequent Android build proceeded normally.

## PS
Adds a change file under .changes as per the contribution guidelines.


crates/tauri-cli/src/mobile/mod.rs
```diff
@@ -585,26 +584,38 @@ fn ensure_init(
...
+  match std::fs::read_to_string(&gradlew_path) {
+    Ok(contents) => {
+      if contents.contains("\r\n") {
+        if let Err(error) = std::fs::write(&gradlew_path, contents.replace("\r\n", "\n")) {
+          log::warn!("failed to normalize gradlew CRLF line endings: {error}");
+        }
+      }
+    }
+    Err(error) => log::debug!("failed to read gradlew for CRLF normalization: {error}"),
   }
```
Legend-Master: Preserve the old `fs_context` at least on unix platforms.

crates/tauri-cli/src/mobile/mod.rs
```diff
@@ -585,26 +584,38 @@ fn ensure_init(
...
-        .replace("\r\n", "\n"),
-    )
-    .fs_context("failed to replace gradlew CRLF with LF", gradlew_path)?;
+  }
+
+  // The CRLF→LF rewrite is platform-neutral: a CRLF gradlew breaks `sh
+  // ./gradlew` on every host (including Git-Bash on Windows), and bash
+  // scripts must always use LF. Best effort: only rewrite when a CRLF is
+  // actually present, and warn instead of failing so a locked or read-only
+  // file cannot abort builds that would otherwise succeed.
```
Legend-Master: Remove LLM marks, just say why do it on all platforms, and link this PR.

Legend-Master (commented):
Thanks! Just a few nitpicks and we can get this merged.

k6G52m4Dz75W:
Both addressed in adaa8b2: unix failures now propagate with the original fs_context messages again (only Windows warns, since the CLI invokes gradlew.bat there), and the comment now plainly states why the rewrite runs on all platforms with a link back to this PR.

Legend-Master (approved):

PR #16029, opened 2026-09-15
chore: migrate to edition 2024

### LINKED ISSUE #16026: RUSTSEC-2026-0285: TLS 1.3 handshake messages incorrectly accepted across encryption level boundaries
> TLS 1.3 handshake messages incorrectly accepted across encryption level boundaries

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `rustls`                      |
| Version             | `0.23.35`                   |
| URL                 | [https://github.com/rustls/rustls/security/advisories/GHSA-2mjx-qc3c-rqvc](https://github.com/rustls/rustls/security/advisories/GHSA-2mjx-qc3c-rqvc) |
| Date                | 2026-09-14                         |
| Patched versions    | `>=0.23.45`                  |
| Unaffected versions | `<0.23.13`               |

Rustls accepted TLS 1.3 handshake messages sent at the wrong encryption level
when they followed a key-changing message in the same record. For example,
a plaintext `EncryptedExtensions` message packed into the same record as the
`ServerHello` was accepted.

RFC 8446 section 5.1 requires that handshake messages do not span key changes,
and that implementations terminate the connection with an &quot;unexpected_message&quot;
alert if they do.

The handshake transcript is still authenticated, so a network-position attacker
cannot use this to alter or complete a handshake; the practical effect is that
a peer could send handshake messages that should be encrypted in plaintext
without rustls rejecting the connection.

This is functionally the same bug as Go&#39;s
[GO-2026-4340](https://pkg.go.dev/vuln/GO-2026-4340) (CVE-2025-61730).

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2026-0285.html) for additional details.

Legend-Master:
I have changed the format and clippy to preserve 2021 styles as much as possible to keep the diffs small, will migrate them in follow up PRs.


crates/tauri-build/src/lib.rs
```diff
@@ -873,18 +874,18 @@ mod tests {
     assert!(crate::should_static_link_vc_runtime(&config, &attributes));
 
     // 2. Set to anything but "false" in env, should be true
-    std::env::set_var("STATIC_VCRUNTIME", "qweqe");
+    unsafe { std::env::set_var("STATIC_VCRUNTIME", "qweqe") };
```
Legend-Master: All the env var modifications in tests are not sound, should we guard them in [serial_test](https://docs.rs/serial_test/latest/serial_test/)?

Legend-Master, line 877: All the env var modifications in tests are not sound, should we guard them in [serial_test](https://docs.rs/serial_test/latest/serial_test/)?

lucasfernog: we should track that and fix it in the future

lucasfernog (approved):

PR #16023, opened 2026-09-14
feat: changes for the v3 TODOs in the codebase

lucasfernog:
remove deprecated APIs, re-exports, compatibility code etc

migration CLI functionality will be a separate PR

PR #16039, opened 2026-09-16
chore(cli): delete the stale schema.json snapshots

lucasfernog:
ported from ported from #16023


Legend-Master:
Do you know where did those come from? And should we target this to v2 as well?

lucasfernog:
we had this before, but Amr changed the config generation crate and they are stale - i'll also delete on v2

PR #16042, opened 2026-09-16
refactor: return tauri::Result from AppManager::get_asset

lucasfernog:
ported from #16023


Legend-Master (dismissed):
Actually, this is not publicly exported, we should target this to v2 instead

Legend-Master (approved):
Rebased to target v2 instead

PR #16049, opened 2026-09-17
perf: avoid redundant rgba copies in icon conversions

Tunglies:
Verified with an allocation-counting benchmark: icon conversions drop from 2 to 1 allocation per op (the redundant rgba copy is gone) and `Image::from_bytes` from 24 to 8 per decode — byte-identical output.


Legend-Master (approved):
Thanks, nice catch

PR #15220, opened 2026-04-10
perf: load resources of `tauri://` protocol asynchronously

DeTeam:
# Summary

Closes #15071

This PR addresses performance issues during development of mobile apps on Tauri. As our bundle grows it seems like the bottleneck often times is not Vite (our bundler/dev server) or the webview itself but rather the assets loader in the Tauri dev server.

After adding a few larger sub-dependencies to our project (e.g. three.js) and having multiple IPC plugin calls on startup from JS world, we ended up in various deadlocks where some assets never get loaded. This comes with a very funky race condition.

In this PR we're moving away from safe_block_on to an async tokio task that uses an existing responder to deliver results.

## Testing

I check our app on iOS and with this change it's much much faster. I also checked our app on mac and have not noticed any issues or regressions there.

PR #16050, opened 2026-09-17
fix(core): load asset protocol files asynchronously

### LINKED ISSUE #7434: [bug] load image file by asset protocol from smb will hang the entire application.
### Describe the bug

I am using `convertFileSrc` to load local file, the app run properly when I load file from local disk. But if I load file from remote source, it will hang the entire application until the file loaded.
I load file by smb protocol, path like this `\\home.haoyuanlord.net\media\pexels-roberto-nickson-2559941.jpg`
this will be obvious when load 100mb+ files at the same time.
file load not just hang current window, other windows also were freezed, so I think maybe the rust-end was blocked.

### Reproduction

[reproduction repo](https://github.com/a1156883061/tauri-net-load)

### Expected behavior

the app run normally when load the image files

### Platform and versions

```text
[✔] Environment
    - OS: Windows 10.0.22621 X64
    ✔ WebView2: 114.0.1823.82
    ✔ MSVC: Visual Studio ���ɹ��� 2019
    ✔ rustc: 1.66.0 (69f9c33d7 2022-12-12)
    ✔ Cargo: 1.66.0 (d65d197ad 2022-11-15)
    - node: 18.12.1
    - pnpm: 8.6.7
    - yarn: 3.6.1
    - npm: 8.19.2

[-] Packages
    - tauri [RUST]: 1.4.1
    - tauri-build [RUST]: 1.4.0
    - wry [RUST]: 0.24.3
    - tao [RUST]: 0.16.2

[-] App
    - build-type: bundle
    - CSP: default-src 'self'; img-src 'self' asset: https://asset.localhost
    - distDir: ../dist
    - framework: Vue.js
    - bundler: Vite
```

### Stack trace

_No response_

### Additional context

_No response_

PathGao:
Fixes #7434.

If an `asset://` file sits on a slow or unreachable path, like a network share that's down, the whole app freezes, every window. The handler reads the file right on the thread that runs the event loop.

This makes the handler async, the same way #15220 did for `tauri://`: file access goes back to `tokio::fs` (undoing the `asset.rs` part of #15117), and the handler responds from a spawned task. Markpad has shipped the same idea as its own `asset` handler since August.

Verification on macOS arm64:

```sh
cargo test -p tauri --lib --all-features protocol::asset
```

```sh
cargo clippy -p tauri --all-features --lib --tests
```

I also ran a local test (not included) that reads from a FIFO, which blocks like a dead share. It times out on `dev` and passes with this change.


Legend-Master:
I think you might want to re-open #15836 instead

PathGao:
I'd have liked to, but I deleted the fork that PR came from, so GitHub won't let me reopen it. This is the same change, rebased onto current dev.

PathGao:
It's also smaller than #15836: just the handler change and a change file, without the FIFO test. Happy to add the test back if you want it.

Legend-Master:
> I'd have liked to, but I deleted the fork that PR came from, so GitHub won't let me reopen it. This is the same change, rebased onto current dev.

Got it, I'm fine with either one of them.

Since everything here can be async (through tokio), we should revert the changes made in #15117 and mirror https://github.com/tauri-apps/tauri/pull/15220 for it to run as an async task, no need to block here.

PathGao:
Done. It's async now like #15220: `tokio::fs` again, and the handler responds from a spawned task.

Tunglies:
Any benchmark cloud be sharing? It seemed using async in sync scope makes no reason, unless using `tokio::spawn_blocking` in sync scope, but it produce threads of tokio.

Tunglies:
Also, `block_on` would process async codes as sync. The await does nothing, it still needs to wait to process hole codes inside.

PathGao:
There's no `block_on` in the handler. It spawns the task and returns right away, and the task calls the responder once the file is read. The only `block_on` is in the test.

This isn't about speed, so I don't have a benchmark. The handler runs on the event loop thread, so a read that hangs (like a network share that's down) used to freeze every window. Now it doesn't.

You're right that `tokio::fs` uses tokio's blocking threads underneath. The first version of this PR used `spawn_blocking` directly. Legend-Master asked for the async version to match #15220.

crates/tauri/src/protocol/asset.rs
```diff
@@ -5,32 +5,36 @@
...
 use http::{header::*, status::StatusCode, Request, Response};
 use http_range::HttpRange;
-use std::fs::File;
-use std::io::{Read, Seek, Write};
 use std::{borrow::Cow, io::SeekFrom};
 use tauri_utils::mime_type::MimeType;
+use tokio::fs::File;
+use tokio::io::{AsyncReadExt, AsyncSeekExt, AsyncWriteExt};
 
 pub fn get(scope: scope::fs::Scope, window_origin: String) -> UriSchemeProtocolHandler {
```
Tunglies, line 13: How about breaking changes in v3, to write as `async pub fn get`? So there is no need any spawn behavior. @Legend-Master

Legend-Master: Not sure I get this, it returns a boxed function here, doesn't matter if the outer function is async or not?

Tunglies: I see now, and there is no need to implement as Future or async function.

Legend-Master:
> Any benchmark cloud be sharing? It seemed using async in sync scope makes no reason, unless using `tokio::spawn_blocking` in sync scope, but it produce threads of tokio.

This is not about performance but rather not blocking the main thread on reading files.

> Also, `block_on` would process async codes as sync. The await does nothing, it still needs to wait to process hole codes inside.

That is only used in tests so...

Legend-Master (approved):
Thanks

PR #16051, opened 2026-09-17
chore: remove unused `safe_block_on`

Legend-Master:


FabianLars:
we talked about exposing that to the users a few times since it can be quite helpful, a bit suprised we don't use it anymore at all

FabianLars (approved):

Legend-Master:
What does this function even do though? I can't seem to get it 🤔

FabianLars:
prevents the "can't create runtime inside a runtime" error so you can use the same block_on in sync and async contexts

Legend-Master:
Is it just [`block_in_place`](https://docs.rs/tokio/1.44.2/tokio/task/fn.block_in_place.html)?

FabianLars:
no that runs a sync/blocking fn, safe_block_on can run an async fn

Issue #9285, opened 2024-03-27
[feat] Support for Browser Profiles


zphrs:
(see Wry issue here: https://github.com/tauri-apps/wry/issues/1321)

### Describe the problem

I would ideally like to be able to create separate browsing profiles to segment Cookies, localStorage, and IndexedDB data into different profiles. Here are some use cases which could benefit from this feature: 
- Supporting multiple account login without any further server-side development
- Local-first applications like Excalidraw which typically stores user data within either localStorage or an IndexedDB table
- To limit third party tracking cookies across different webviews

The above use cases are very similar to the use cases for the `data_directory()` api. However, as I noted in the Alternatives considered section, the `data_directory()` api is not supported on MacOS. 

## Platform-Specific Implementation Details

In macOS 14 and iOS 17 Apple has released new [DataStore APIs](https://webkit.org/blog/14423/building-profiles-with-new-webkit-api/) which should enable this feature.

In Windows Microsoft has released new [Multiple Profile Support](https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/multi-profile-support?tabs=dotnetcsharp) which should enable this feature.

I assume we can polyfill this feature to WebkitGTK using the existing `data_directory()` function on the `WebviewBuilder` struct. Ideally the polyfill would create additional subdirectories within the default data directory to support the different profile names. For private profiles (i.e. incognito windows) which shouldn't be persisted I propose the use of the tmp directory to store those temporary profiles. While we won't get the performance advantages listed below in the `Alternatives Considered`, a polyfill will at least allow some support in the interim before GTK supports browser profiles.

## Proposed API

The following is inspired by the above linked guides from the shared features which both Apple and Microsoft support. Note that Microsoft supports one additional configuration for a Profile, [`ScriptLocale`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2controlleroptions.scriptlocale?view=webview2-dotnet-1.0.2420.47#microsoft-web-webview2-core-corewebview2controlleroptions-scriptlocale), which I chose to exclude from the API since WebKit does not support it in their implementation. Conveniently both support private windows and custom names, although WebKit requires that the name is a UUID whereas Webview2 has some [arbitrary name constraints](https://learn.microsoft.com/en-us/dotnet/api/microsoft.web.webview2.core.corewebview2controlleroptions?view=webview2-dotnet-1.0.2420.47).

[...]

PR #16054, opened 2026-09-18
feat(runtime-wry): apply data_store_identifier as WebView2 profile name on Windows

jhasheng:
## Description

`data_store_identifier` is currently only consumed on macOS/iOS, where it maps to a `WKWebsiteDataStore` through wry's `with_data_store_identifier`. On Windows the attribute is silently ignored, so apps that set it get no isolation benefit.

WebView2 supports multiple named profiles under a single user data folder ([Microsoft docs](https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/multi-profile-support)). wry exposes this via `with_profile_name` (available since wry 0.56; this branch already targets wry 0.57), so this PR reuses `data_store_identifier` on Windows: each webview that sets an identifier gets its own named profile, meaning fully isolated cookies and storage per identifier while still sharing a single browser process.

The 16 identifier bytes are rendered as 32-char lowercase hex, which satisfies WebView2 profile naming rules and is collision-free by construction. Apps that do not set `data_store_identifier` are unaffected.

Added `.changes/webview2-profile-name.md`.

## Testing

- `cargo test -p tauri-runtime-wry` — 3 new unit tests covering the profile name mapping (format, determinism, no collisions)
- `cargo clippy -p tauri-runtime-wry --all-targets` and `cargo fmt` clean
- Manually verified on Windows 11 with a multi-account viewer app: opening several accounts as separate webviews on one shared data directory produces one `WV2Profile_<hex>` directory per account under `EBWebView/`, each keeping its own logged-in session

Happy to adjust the mapping format or the config surface if maintainers prefer a different approach.


FabianLars:
We probably should have called it something like profile_identifier instead... maybe in v3 then

Legend-Master:
> We probably should have called it something like profile_identifier instead... maybe in v3 then

We can add a new one and deprecate the previous ones if we want. Also linking #9285 for some visibility.

PR #16057, opened 2026-09-19
fix(tests): guard env var modifications with serial_test (fix #16044)

### LINKED ISSUE #16044: [bug] All the env var modifications in tests are not sound
All the env var modifications in tests are not sound, should we guard them in [serial_test](https://docs.rs/serial_test/latest/serial_test/)?

_Originally posted by @Legend-Master in https://github.com/tauri-apps/tauri/pull/16029#discussion_r4022321915_

marrionesa:
Closes #16044

Yes 🙂 — this guards the env-var-touching tests with `serial_test`, following the convention already used in `tauri-utils` (`#[serial_test::serial(resources)]` in `src/resources.rs`).

## The problem

After the Rust 2024 edition migration (#16029), `std::env::set_var`/`remove_var` became `unsafe`, since mutating the environment while another thread may be reading it is UB. Cargo runs test threads in parallel within the same process, so tests that mutate env vars can race with tests that read them (or assert their absence). The `unsafe` blocks kept the compiler happy but didn't make the operations sound.

## Changes

- Added `serial_test = "3.5"` as a dev-dependency of `tauri-build`, `tauri-bundler` and `tauri-cli`
- Annotated the tests that mutate env vars with the global `#[serial]` attribute, so no `set_var`/`remove_var` can run in parallel with another annotated test:
  - `tauri-build/src/lib.rs` → `static_vc_runtime_chain`
  - `tauri-bundler/src/utils/http_utils.rs` → the 3 `test_generate_mirror_url_*` tests
  - `tauri-cli/src/interface/rust.rs` → `parse_target_dir_from_opts`

## Intentionally untouched

All remaining `set_var`/`remove_var` call sites are in production code paths that run single-threaded (build scripts, CLI entrypoints, mobile module init), so their `unsafe` blocks are sound and out of the scope of this issue.

## Testing

- `cargo fmt --all -- --check` ✅
- `cargo clippy -p tauri-build -p tauri-bundler -p tauri-cli --all-targets -- -D warnings` ✅
- `cargo test --lib` on the three crates × 10 consecutive runs, all green ✅

Test-only change, so no `.changes` entry — happy to add one if preferred.


Legend-Master (changes requested):
Thanks for working on this!

I think we can't have any `set_var` in any test running in parallel with another test so they must not be keyed here.

> https://docs.rs/serial_test/4.0.1/serial_test/attr.serial.html

marrionesa:
Fixed in 27ed171ad - you were absolutely right, the env is a single process wide resource. Thats why tests with different keys, despite touching different environment variables, were still running in parallel and overwriting each other. Now all five use global #[serial] without keys.

Legend-Master (approved):

PR #16062, opened 2026-09-20
fix(bundler): update linuxdeploy and linuxdeploy-plugin-gtk

### LINKED ISSUE #15902: [bug] AppImage GTK hook still hard-crashes with no XWayland when GDK_BACKEND is unset (not fixed by #15786)
### Describe the bug
On a Wayland session with no XWayland process available at all, a Tauri-bundled AppImage crashes outright at GTK init. This is a different failure mode from #15781: that report (and its fix, #15786) is about an *explicitly set* `GDK_BACKEND` being silently clobbered. This report is about the plain default — the common case where a user never sets `GDK_BACKEND` at all and just runs the AppImage — which #15786 leaves untouched on purpose.

### Reproduction
1. On a Wayland compositor with no XWayland running (e.g. niri without `xwayland-satellite`), run any Tauri-bundled AppImage with `GDK_BACKEND` unset.
2. `apprun-hooks/linuxdeploy-plugin-gtk.sh` defaults to `x11`, exactly as designed post-#15786.
3. `gtk_init` has no X server of any kind to connect to.

### Expected behavior
Either the app falls back to a working backend, or the failure is a clear diagnostic rather than a panic/core dump.

### Actual behavior
```
Unhandled panic: panicked at .../tao-0.34.8/src/platform_impl/linux/event_loop.rs:217:53:
Failed to initialize gtk backend!: BoolError { message: "Failed to initialize GTK", filename: ".../gtk-0.18.2/src/rt.rs", function: "gtk::rt::init", line: 141 }
Error: External command core dumped: core dumped with SIGABRT (6)
```

### Versions observed
(End-user report, not a project checkout, so no `tauri info` — these are read off the crash trace and the bundled AppImage contents.)
- tao: 0.34.8
- gtk-rs (`gtk` crate): 0.18.2
- webkit2gtk: 4.1 (bundled)

### Reproduced via
[[julyx10/lap](https://github.com/julyx10/lap)](https://github.com/julyx10/lap), unmodified `tauri-bundler` AppImage output.

### Why #15786 doesn't fix this
Per that PR's own summary, it "retain[s] x11 as the safe default when the variable is unset." That default is exactly what's crashing here — no explicit override was ever set, so there's nothing for #15786's preservation logic to preserve.

### Workaround confirmed
Extracting the AppImage and patching the hook fixes it completely, no XWayland involved:
```bash
./App.AppImage --appimage-extract
cd squashfs-root
sed -i 's/^export GDK_BACKEND=x11.*/export GDK_BACKEND=wayland,x11/' apprun-hooks/linuxdeploy-plugin-gtk.sh
./AppRun
```

### Suggested fix
Given the maintainers' own reservations about touching this hook further (noted in the #15786 review) and the pending bundler rewrite in #12491, a full fix here may not be worth it on the legacy path. Two lower-effort options if it is:
- Change the unset-default from bare `x11` to `wayland,x11`, keeping the #8541 safety net for systems with no Wayland support while no longer hard-failing on systems with no X11 support.
- Otherwise, surface the `std::env::set_var("GDK_BACKEND", ...)` escape hatch (#11790) more visibly — e.g. in the AppImage bundling docs — since it's currently the only thing app authors can rely on, and most won't know to look for it until they hit this.

### Related
- #15781 / #15786 — fixes a related but distinct problem (explicit-override clobbering); this report covers what's left afterward
- #8541 — original justification for the hardcoded default; appears to no longer reproduce on current WebKitGTK, going by the workaround above requiring no other changes
- #12491 — bundler rewrite that may obsolete this path entirely

### Full `tauri info` output

```text
- OS: Arch Linux Unknown x86_64 (X64) (niri on wayland)
    ✔ webkit2gtk-4.1: 2.52.6
    ✔ rsvg2: 2.62.3
    ✔ rustup: 1.29.0 (2026-03-23)
    ✔ Rust toolchain: stable-x86_64-unknown-linux-gnu (overridden by environment variable RUSTUP_TOOLCHAIN)
    - node: 26.7.0
    - npm: 12.0.2
    - bun: 1.3.14
    - deno: deno 2.9.5
```

### LINKED ISSUE #15976: [bug] Default AppImage bundles libwayland/libxkbcommon/libxcb; aborts with EGL_BAD_PARAMETER on newer Mesa (Fedora 44)
### Describe the bug

A default-config AppImage (`tauri build --bundles appimage`, no `bundle.linux` config) built on `ubuntu-22.04` aborts on Fedora 44 under a Wayland session before any window appears:

```
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
```

Root cause is over-bundled display-stack libraries, not app code: I extracted the AppImage and found these in `usr/lib`, all shadowed ahead of the host copies via the `LD_LIBRARY_PATH` that `AppRun.wrapped` injects:

```
libwayland-client.so.0  libwayland-cursor.so.0
libwayland-egl.so.1     libwayland-server.so.0
libxkbcommon.so.0
libxcb-randr.so.0  libxcb-render.so.0  libxcb-shm.so.0
libXau.so.6  libXdmcp.so.6
```

The host's Mesa 26 ends up negotiating EGL against the Ubuntu 22.04-era bundled `libwayland-client`, and `eglGetDisplay` fails. Notably `libEGL/libGL/libdrm/libc` are *not* bundled (linuxdeploy's built-in excludelist already covers them) — the display-stack libs above are the gap. The pkg2appimage community excludelist carries `libwayland-client.so.0` with exactly this rationale ("breaks newer Mesa"); linuxdeploy's bundled excludelist does not.

The same failure and the same fix were independently confirmed in a second, non-Tauri project that also ships via linuxdeploy: excluding `wayland/xkbcommon/xcb/Xau/Xdmcp` and letting the host provide them fixed startup on the same class of machine. So this points at linuxdeploy's missing excludelist entries rather than anything Tauri-specific.

### Reproduction

1. Build any Tauri 2.11 app with `--bundles appimage` on `ubuntu-22.04` (default config, `webkit2gtk-4.1`).
2. Run the AppImage on Fedora 44, GNOME Wayland, Mesa 26, AMD Rembrandt iGPU.
3. Observe the EGL abort on stderr; no window ever appears.

### Verified minimal fix

Extracted the AppImage, removed only the 10 libraries listed above, ran `./AppRun`: the app boots normally (backend starts listening, process stays alive instead of aborting within seconds). No need to strip glib/gstreamer or touch anything else — the 10 display-stack libs are sufficient, and the host versions (wayland 1.25, xkbcommon 1.13) are drop-in compatible.

### Expected behavior / ask

- Don't bundle the display/GPU stack by default (adopt the community excludelist entries for `libwayland-*`/`libxkbcommon`/`libxcb`/`libXau`/`libXdmcp`), or
- Provide the supported escape hatch requested in #15665: review/merge #15662 (`bundle.linux.appimage.excludeLibraries` → `--exclude-library`). Until one of these lands, every project is stuck post-processing the AppImage in CI.

### Related

- #15665 — same symptom, different distro pair (ubuntu-24.04 → Ubuntu 26.04/Mesa 25). This report adds a second pair (ubuntu-22.04 → Fedora 44/Mesa 26) plus a bisected minimal removal set.
- #15662 — the `excludeLibraries` PR that would resolve this.
- #8541 — prior art for the `GDK_BACKEND=x11` hook; different failure, already worked around.

### Environment

- Build: `ubuntu-22.04` GitHub runner, `@tauri-apps/cli 2.11.4`, `tauri 2.11.5` (bundler path `linuxdeploy.rs`: `--appdir … --plugin gtk --output appimage`, no exclude passthrough)
- Target: Fedora 44, kernel `7.1.12-200.fc44`, GNOME Wayland (`WAYLAND_DISPLAY=wayland-0`), Mesa `26.1.8`, `libwayland-client 1.25.0`, `libxkbcommon 1.13.1`, `webkit2gtk4.1 2.52.5`, AMD Rembrandt [Radeon 680M] (Ryzen 7 7735HS)

FabianLars:
So far i only tested building and running on arch. I assume this _will_ cause issues but since the old stuff also causes issues i think this is fine.

script diff is https://github.com/tauri-apps/linuxdeploy-plugin-gtk/pull/5 + https://github.com/tauri-apps/linuxdeploy-plugin-gtk/pull/6

Still on vacation so idk how long it takes for me to test this.

Should fix https://github.com/tauri-apps/tauri/issues/15976
Maybe fixes https://github.com/tauri-apps/tauri/issues/15902


NVolcz:
Thanks for working on this. I tested the PR with a production Tauri app on Ubuntu 26.04, where the current AppImage fails to start. With this PR, it starts normally on both Wayland and X11 without any workarounds.

One thing I noticed is that prepare_tools() downloads linuxdeploy-plugin-gtk.sh from the plugin repository’s master branch instead of using the version included in this PR. I monkeypatched it by putting the PR’s script in the Tauri cache to test it.

Otherwise, this fixes the issue for us. Thanks!

FabianLars:
> One thing I noticed is that prepare_tools() downloads linuxdeploy-plugin-gtk.sh from the plugin repository’s master branch instead of using the version included in this PR. I monkeypatched it by putting the PR’s script in the Tauri cache to test it.

ffs, how embarassing...

PR #16064, opened 2026-09-21
fix(bundler): recognize deb and rpm as self-contained updater targets

lucasfernog:
The self-contained updater check excluded RPM and the "no updater-enabled targets" warning omitted deb/rpm, although both are signed as updater artifacts and supported by the updater plugin.


crates/tauri-bundler/src/bundle.rs
Legend-Master: Did we have `.deb` support in v1?

crates/tauri-bundler/src/bundle.rs
```diff
@@ -241,12 +241,21 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
...
+            | PackageType::Deb
+            | PackageType::Rpm
         )
       })
     {
+      let updater_enabled_targets = if updater.v1_compatible {
+        "app, appimage, deb, msi, nsis"
+      } else {
+        "app, appimage, deb, rpm, msi, nsis"
+      };
```
Legend-Master: There're no differences between the 2 that we always support all v2 bundles

Legend-Master (changes requested):

crates/tauri-bundler/src/bundle.rs
FabianLars: if you mean the updater, no.

Legend-Master: We should remove this then?

crates/tauri-bundler/src/bundle.rs
```diff
@@ -217,39 +217,48 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
...
-            | PackageType::Nsis
-            | PackageType::WindowsMsi
-            | PackageType::Deb
-        )
-      } else {
-        matches!(package_type, PackageType::MacOsBundle)
-      }
-    }) {
+    // Targets the legacy v1 updater can consume once they are wrapped in a tar.gz / zip.
+    let has_v1_target = package_types.iter().any(|package_type| {
```
Legend-Master, line 221:
```suggestion
    let has_v1_target = updater.v1_compatible && package_types.iter().any(|package_type| {
```

move the check here?

crates/tauri-bundler/src/bundle.rs
```diff
@@ -217,39 +217,48 @@ pub fn bundle_project(settings: &Settings) -> crate::Result<Vec<Bundle>> {
...
+        "The bundler was configured to create updater artifacts but no updater-enabled targets were built. Please enable one of these targets: app, appimage, deb, rpm, msi, nsis"
       );
     }
+
     if updater.v1_compatible {
+      if !has_v1_target {
+        log::warn!(
+          "No v1 compatible updater artifact was created: the legacy updater only supports the app, appimage, msi and nsis targets. deb and rpm bundles can only be installed by the v2 updater plugin."
+        );
+      }
```
Legend-Master, line 261: Then move this  up to the `else if` branch above

PR #16071, opened 2026-09-21
fix(utils): emit the Permissions-Policy header under its correct name

lucasfernog:
The `Permissions-Policy` entry of `app > security > headers` was written as the invalid header `Permission-Policy`, so the policy never reached the browser.


crates/tauri-utils/src/config.rs
```diff
@@ -4943,6 +4943,23 @@ mod test {
     assert!(object_json.contains("\"args\":null") || !object_json.contains("args"));
   }
 
+  #[test]
+  fn permissions_policy_header_name() {
```
Legend-Master: This test doesn't make any sense, we should move.

Legend-Master (commented):
Nice typo catch

PR #16073, opened 2026-09-21
fix(cli): install Android SDK platform 37 to match the app template

lucasfernog:
The SDK bootstrap installed `platforms;android-36` while the generated project uses `compileSdk`/`targetSdk` 37.


.changes/android-sdk-bootstrap-37.md
```diff
@@ -0,0 +1,6 @@
+---
```
Legend-Master: No need this change file, we have not released the template change yet

Legend-Master (dismissed):
Nice catch

PR #16074, opened 2026-09-21
fix(build): validate inline capabilities from the config file

lucasfernog:
Capabilities declared inline under `app > security > capabilities` skipped tauri-build's validation, so a typo in a permission identifier only surfaced as the generic `failed to resolve ACL` codegen panic. Validate the same capability set codegen resolves so the friendly "Permission X not found" error is reported instead.


Legend-Master (approved):

Legend-Master:
Feel like we have way too much duplicated code around the config and acl handling code

lucasfernog:
yeah it needs a rewrite :/ tho it impacts too much code

Issue #13892, opened 2025-07-25
[bug] [android] [windows] [cli] [build] Android Build Error: A problem occurred starting process 'command 'C:\nvm4w\nodejs\node.exe.cmd'' and Error: Cannot find module '...tauri' when using nvm4w


DomanskiFilip:
### Describe the bug

Hi everyone. I am a student working on my year 3 project useing tauri and I encoutered this issue:

Every time I attempt to start a new Tauri project and run tauri android dev on Windows, I consistently encounter a build failure. The primary errors indicate an issue with finding and executing Node.js via C:\nvm4w\nodejs\node.exe.cmd, followed by a Cannot find module '...tauri' error, and ultimately a Gradle build failure on the rustBuildX86_64Debug task.

ERROR LOGS: 
-----------------------------------------------------------------------------------------------------------------------
            Finished `dev` profile [unoptimized + debuginfo] target(s) in 10.34s
                Info symlinking lib "E:\\uni\\year 3\\year 3 project\\CalendarAssistantApp\\CATdesktop\\src-tauri\\target\\x86_64-linux-android\\debug\\libapp_lib.so" in jniLibs dir "E:\\uni\\year 3\\year 3 project\\CalendarAssistantApp\\CATdesktop\\src-tauri\\gen/android\\app/src/main/jniLibs/x86_64"
                Info "E:\\uni\\year 3\\year 3 project\\CalendarAssistantApp\\CATdesktop\\src-tauri\\target\\x86_64-linux-android\\debug\\libapp_lib.so" requires shared lib "libandroid.so"
                Info "E:\\uni\\year 3\\year 3 project\\CalendarAssistantApp\\CATdesktop\\src-tauri\\target\\x86_64-linux-android\\debug\\libapp_lib.so" requires shared lib "libdl.so"
                Info "E:\\uni\\year 3\\year 3 project\\CalendarAssistantApp\\CATdesktop\\src-tauri\\target\\x86_64-linux-android\\debug\\libapp_lib.so" requires shared lib "liblog.so"
                Info "E:\\uni\\year 3\\year 3 project\\CalendarAssistantApp\\CATdesktop\\src-tauri\\target\\x86_64-linux-android\\debug\\libapp_lib.so" requires shared lib "libm.so"
[...]

Issue #8072, opened 2023-10-20
[bug] can't build for android with corepack (pnpm not found)


flapili:
### Describe the bug

Hi, I recently switch to corepack to handle packages managers and I'm unable to build apks (not tried yet for ios)

### Reproduction

use corepack and alias pnpm to corepack

### Platform and versions

```text
[✔] Environment
    - OS: Mac OS 14.0.0 X64
    ✔ Xcode Command Line Tools: installed
    ✔ rustc: 1.72.0 (5680fa18f 2023-08-23)
    ✔ Cargo: 1.72.0 (103a7ff2e 2023-08-15)
    ✔ rustup: 1.26.0 (5af9b9484 2023-04-05)
    ✔ Rust toolchain: stable-x86_64-apple-darwin (default)
    - node: 18.18.2
    - pnpm: 8.9.2
    - npm: 9.8.1

[-] Packages
    - tauri [RUST]: 2.0.0-alpha.16
    - tauri-build [RUST]: 2.0.0-alpha.10
    - wry [RUST]: 0.33.1
    - tao [RUST]: 0.22.3
    - @tauri-apps/api [NPM]: not installed!
    - @tauri-apps/cli [NPM]: 2.0.0-alpha.16

[-] App
    - build-type: bundle
    - CSP: unset
    - distDir: ../src
    - devPath: http://localhost:8080/
    - framework: Vue.js (Nuxt)
    - bundler: Webpack <- btw that's wrong
```

### Stack trace

```text
> Task :app:rustBuildArmDebug FAILED
Execution failed for task ':app:rustBuildArmDebug'.
> A problem occurred starting process 'command 'pnpm''

* Try:
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':app:rustBuildArmDebug'.
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:149)
	at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:282)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:147)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:135)
[...]

PR #16078, opened 2026-09-21
fix(cli): resolve pnpm-native binary detection for Android builds

xairaven:
## Description

On Android, `tauri android dev`/`build` generates a Gradle task (`BuildTask.kt`) that shells out to rebuild the Rust `cdylib`. The command it uses is picked by inspecting `npm_execpath`'s file stem, but only `npm-cli` and `npx-cli` are recognized — anything else is used verbatim as the literal executable name.

For a corepack-managed pnpm install, that stem resolves to `pnpm-native`, which isn't a real executable. This gets baked into the generated `BuildTask.kt`:

```kotlin
val executable = """pnpm-native""";
```

Gradle then fails with:

```
A problem occurred starting process 'command 'pnpm-native''
```

This adds a pnpm-prefix case alongside the existing npm/npx handling, so pnpm is correctly detected and pnpm (a real, on-PATH command) is written into the generated file instead.

Related prior art — same class of bug (wrong executable baked into BuildTask.kt due to incomplete package-manager detection):
- #13892
- #8072

## Testing

Reproduced and verified against a real Android project using a corepack-managed pnpm install (pnpm@12.4.2 via corepack). Before the fix, freshly regenerating gen/android produced `val executable = """pnpm-native""";` after the fix, it produces `val executable = """pnpm""";`, and the Gradle build proceeds past the previously-failing `:app:rustBuild*` task.


crates/tauri-cli/src/mobile/init.rs
```diff
@@ -96,6 +96,8 @@ fn exec(
               "npm".into()
             } else if manager_stem == "npx-cli" {
               "npx".into()
+            } else if manager_stem.to_string_lossy().starts_with("pnpm") {
```
Legend-Master, line 99: What about just checking for `pnpm-native`?

Legend-Master (approved):
Thanks! Just a bit of small nitpicks.

Legend-Master:
Actually, this was fixed a moment ago by #16075, thanks anyways!

